PR #26, opened 2016-07-18
Set router subdomain to <ip>.xip.io by default

### LINKED ISSUE #18: allow the subdomain to be configured and default it to an xip.io default OOTB
When you create a Route on MiniShift with an empty Host it gets defaulted to: `$(name)-$(namespace).router.default.svc.cluster.local` which doesn't work on the host and there's no way to change it.

It'd be nice if `minishift start` had a `subdomain` parameter which defaulted to `$(minishift ip).xip.io` so that if the Router was used with MiniShift that they work.

On our old Vagrant VM we configured the subdomain using a master yaml file change here: 
https://github.com/fabric8io/fabric8-installer/blob/master/vagrant/openshift/Vagrantfile#L100-L101

I had a peek into the MiniShift VM and wasn't sure where the master yaml config file is; am guessing we're not using one right now?

jimmidyson:
Fixes #18

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -85,7 +85,12 @@ func runStart(cmd *cobra.Command, args []string) {
 		os.Exit(1)
 	}
 
-	if err := cluster.StartCluster(host); err != nil {
+	kubeIP, err := host.Driver.GetIP()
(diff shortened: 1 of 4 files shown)


jimmidyson:
LGTM

PR #49, opened 2016-08-23
Fixes #39: You can use minishift start

### LINKED ISSUE #39: lets create an openshift registry OOTB so that ImageStreams and S2I can be used
the default way to build things on OpenShift is via ImageStreams which requires the openshift registry; so we should probably default to creating that (but make it so folks could opt out maybe?)

chirino:
--internal-registry=true --router=true

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -44,6 +44,8 @@ var (
 	insecureRegistry []string
 	registryMirror   []string
 	hostOnlyCIDR     string
+	deployRouter     bool
+	deployRegistry   bool
 )
 
 // startCmd represents the start command
@@ -70,6 +72,8 @@ func runStart(cmd *cobra.Command, args []string) {
 		InsecureRegistry: insecureRegistry,
 		RegistryMirror:   registryMirror,
 		HostOnlyCIDR:     hostOnlyCIDR,
+		DeployRouter:     deployRouter,
+		DeployRegistry:   deployRegistry,
 	}
 
 	var host *host.Host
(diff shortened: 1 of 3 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -182,5 +187,8 @@ func init() {
 	startCmd.Flags().StringSliceVar(&insecureRegistry, "insecure-registry", nil, "Insecure Docker registries to pass to the Docker daemon")
 	startCmd.Flags().StringSliceVar(&registryMirror, "registry-mirror", nil, "Registry mirrors to pass to the Docker daemon")
 
+	startCmd.Flags().BoolVarP(&startDockerRegistry, "internal-registry", "", false, "Should the OpenShift internal Docker registry be started?")
```
jimmidyson: Just registry will do for the flag I think. Alternatively, deploy-registry & deploy-router for consistency?

jimmidyson:
From travis:

```
Checking gofmt...
--- /dev/fd/63	2016-08-24 09:24:52.120691328 +0000
+++ /dev/fd/62	2016-08-24 09:24:52.120691328 +0000
@@ -0,0 +1,2 @@
+cmd/minikube/cmd/start.go
+pkg/minikube/cluster/cluster.go
make: *** [test] Error 1
```

Please make sure you run gofmt on all files. If you're using an IDE then you should set this up on save (go-vim is awesome for this btw).

Also I think you'll need to run `make gendocs` to ensure docs are up to date.

jimmidyson:
With #52 this means that it the created registry works ootb.

jimmidyson:
Awesome thanks!

PR #69, opened 2016-09-08
make update feature generic so it can be reused

rawlingsj:
@jimmidyson not sure if this is the best approach but; I've vendored the minishift update.go package in gofabric8 and reused the `MaybeUpdate`  function which works great now that gofabric8 releases are in the same form as minikube and minishift!  - by "the same form" I mean they have a checksum and are not tarballs.

I've also tested `update.MaybeUpdateFromGithub(os.Stdout)` which still works to update minishift as the default.

I had to make these small changes so that the update function is generic enough for both gofabric8 and minishift.

I wasn't sure if vendoring this package is the correct way or if we should split out the update into its  own repo but figured I'd start here as it was easier and it works.

Let me know what you think.

### DIFF
--- a/pkg/minikube/update/update.go
+++ b/pkg/minikube/update/update.go
@@ -43,24 +43,26 @@ import (
 )
 
 const (
-	updateLinkPrefix   = "https://github.com/jimmidyson/minishift/releases/tag/" + version.VersionPrefix
-	downloadLinkFormat = "https://github.com/jimmidyson/minishift/releases/download/v%s/%s"
-	githubOwner        = "jimmidyson"
-	githubRepo         = "minishift"
-	timeLayout         = time.RFC1123
+	timeLayout  = time.RFC1123
+	githubOwner = "jimmidyson"
+	githubRepo  = "minishift"
 )
 
 var (
 	lastUpdateCheckFilePath = constants.MakeMiniPath("last_update_check")
 	githubClient            *github.Client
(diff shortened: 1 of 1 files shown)


rawlingsj:
whoa teats!

ok so looks like the github related tests are being rate limiting. FWIW `make test` is working locally.

jimmidyson:
That rate limiting is a problem for real users too... That's why I want to look at tuf or similar on our own servers (dedicated perhaps?). I'll look at these changes later. Thanks for the update!

rawlingsj:
Great - thanks

pkg/minikube/update/update.go
```diff
@@ -43,24 +43,26 @@ import (
...
-	downloadBinary          = "minishift-" + runtime.GOOS + "-" + runtime.GOARCH
 )
 
 func MaybeUpdateFromGithub(output io.Writer) {
 	MaybeUpdate(output, githubOwner, githubRepo, lastUpdateCheckFilePath)
 }
 
 func MaybeUpdate(output io.Writer, githubOwner, githubRepo, lastUpdatePath string) {
+
+	downloadBinary := githubRepo + "-" + runtime.GOOS + "-" + runtime.GOARCH
```
jimmidyson: Binary name will not always be githubRepo - can you make binary name configurable please.

rawlingsj: Sure thing

PR #78, opened 2016-09-09
Show current version in update prompt

### LINKED ISSUE #73: Tell current version in auto update
So the user know what he/she current runs and what the new version is. For example if there is patch releases then it can be assumed safter to update from 1.0.2 to 1.0.3 than from 1.0.2 to 1.2.0 etc.

jimmidyson:
Fixes #73

How does that look @davsclaus?


davsclaus:
+1

PR #103, opened 2016-09-22
Updated quickstart example with hello-openshift example

budhrg:


README.md
```diff
@@ -59,30 +59,30 @@ Creating machine...
 Starting local OpenShift cluster...
 OpenShift is available at https://192.168.99.100:8443.
 
-$ oc run hello-minishift --image=gcr.io/google_containers/echoserver:1.4 --port=8080 --expose --service-overrides='{"apiVersion": "v1", "spec": {"type": "NodePort"}}'
-service "hello-minishift" created
-deploymentconfig "hello-minishift" created
+$ oc create -f hello-openshift.json
```
jimmidyson: How about using `oc new-app https://github.com/openshift/ruby-hello-world.git` instead?

budhrg: @jimmidyson That would a nice example but I would go with `hello-openshift` demo as quick start demo as `openshift/hello-openshift` is around `7.6 MB` and will enable users to start quickly.

However, `ruby-22-centos7` is around `440 MB` and may not be easy for users with less bandwidth.

Moreover, I found `ruby-hello-world` repo is not update with service type `NodePort` which may adds additional steps in the accessing application like exposing service.

Let me know what do you think.

budhrg: @jimmidyson any update?

jimmidyson: Sorry for not looking at this in a while. OK you've convinced me :) Much slicker& easier.

Do you think you could turn that hello-openshift.json into a `new-app`able repo then we could use the oc new-app & follow on the normal openshift flow from there?

LalatenduMohanty:
@budhrg Are you planning to work on this? or we can close this PR for now?

budhrg:
@LalatenduMohanty Will work on it as it needs some updates and then we can merge it.

README.md
budhrg: @jimmidyson Sorry could not follow up this PR due to busy schedule.

I am bit curious to know the advantage of having `oc new-app` here against `oc create -f <example>.json`.

I believe to get `oc new-app`, I need to create a `<template-example>.json` and then it is two commands set up as 
- `oc create -f <template-example>.json`
- `oc new-app template-example`

WDYT ?

budhrg: @LalatenduMohanty @jimmidyson @hferentschik WDYT ?

hferentschik:
The plan from my side was to address this as part of pull request #165.

budhrg:
> The plan from my side was to address this as part of pull request #165.

Good to know that. I just wanted to make sure we don't need to use two commands to get set up with demo example. :smile:

hferentschik:
> Good to know that. I just wanted to make sure we don't need to use two commands to get set up with demo example. 😄

Fair enough. I am planning to work with the docs today. I'll let you know when I make changes to this.

hferentschik:
Closing the pull request, since the plan is to change the example as part of the doc changes for issue #138

Issue #147, opened 2016-11-07
Remove the vendor directory and add it to .gitignore


hferentschik:
Personally, I'd like to remove the _vendor_ directory and just keep _glide.yaml_ and _glide.yaml_. As a developer this means you need after a first checkout run _glide install --strip-vendor_, but this takes only a few seconds (provided you have the dependencies in the Glide cache).

The benefits are that it becomes clear that one should not change anything in the _vendor_ directory. Also one cannot accidentally check in changes to this directory. It becomes also clear how dependencies are managed and how new ones are added.

One drawback might be that the very first time you use Glide to get the dependencies, it takes time to download everything. For me that's acceptable.

Thoughts?

jimmidyson:
As discussed, committing vendor dir to source control is the geneal approach of both Kubernetes, OpenShift & a lot of other go projects. It means that dependencies are always up to date once a developer does a clone or pull without the need for another `glide install` step & continuously checking whether glide.yaml/lock have changed.

> As a developer this means you need after a first checkout run `glide install --strip-vendor`

No you don't need to run that at all - as a contributor you clone/pull the repo & can run the build immediately. Why do you think you need to run `glide install --strip-vendor`? Adding new dependencies is a case of running `glide get --strip-vendor your/new/dependency`. You then commit the changed files as you would any other file.

hferentschik:
> As discussed, committing vendor dir to source control is the geneal approach of both Kubernetes, OpenShift & a lot of other go projects.

Sure, that does not make it right though and there are other projects which don't commit it, for example Almighty.

> It means that dependencies are always up to date once a developer does a clone or pull without the need for another glide install step

With the risk that whatever is in _vendor_ is not in sync with the current Glide config which can lead to problems when trying to update the _vendor_ directory, for example when adding a new dependency. 

> continuously checking whether glide.yaml/lock have changed.

> Why do you think you need to run glide install --strip-vendor

If we were to remove the _vendor_ directory. Then a developer would first need to run _glide install_. Right now he does not have to.

You really don't have to check. If you want to be on the safe side, you can just run _glide install_. If nothing has changed, it is pretty much a noop. Right now you have to check the commits to whether someone modifies the _vendor_ directory. Also not optimal.

jimmidyson:
These kind of discussions really have no winner - both sides have what they want to do due to personal experience & both have their merits. I can't personally be bothered to waste any time discussing this so if you feel strongly about not committing vendor directory go for it.

[...]

PR #164, opened 2016-11-29
Issue #147 Removing the vendor directory

hferentschik:
Addresses issue #147 

Removing the vendor directory. Dependencies managed by Glide.


hferentschik:
Closing and re-opening to trigger a new TravisCI build

budhrg:
@hferentschik 

Error

```
The command "curl https://glide.sh/get | sh" failed and exited with 1 during .
```

hferentschik:
> The command "curl https://glide.sh/get | sh" failed and exited with 1 during .

I know. It is a GitHub issue. glide.sh downloads from api.github.com which has a rate limit. Given that I repushed now so often, I exceeded the rate limit. One could just wait for an hour or so, but I think I found a workaround - https://github.com/Masterminds/glide/issues/639

BTW, you can get the same working with the minishift codebase where the download of the binaries is also via api.github.com. You can work around this by setting an environment variable specifying a github token. Authenticated client have a higher rate. 

In fact the latest CI build actually worked now for Glide, but failed for the openshift download. Nothing much to do there right now. Need to leave it alone for a bit.

hferentschik:
Seems the CI builds are happy now. On both pull requests :-)

hferentschik:
> The travis config can probably be tidied up a bit but LGTM for now.

Ok. More changes to the Travis config in the next pull request. Work in progress in this area as well :-)

jimmidyson:
BTW also not a big fan personally of automating squashing commits via github UI - feel like that should be a human decision based on whether commits should actually be squashed (related to same functionality). I'm personally happy to have more than 1 commit pushed for a PR if those commits show distinct changes. WDYT?

LalatenduMohanty:
> BTW also not a big fan personally of automating squashing commits via github UI - feel like that should be a human decision based on whether commits should actually be squashed (related to same functionality). I'm personally happy to have more than 1 commit pushed for a PR if those commits show distinct changes.

Agree. IMO the squashing if required should be done by the PR contributor.

PR #188, opened 2016-12-12
Fix #186 Building the default b2d ISO fails

### LINKED ISSUE #186: Building the default b2d ISO fails
$ make deploy/iso/minishift.iso

    Step 13 : RUN /make_iso.sh
     ---> Running in cc348157ad10
     /bin/sh: 1: /make_iso.sh: not found
    The command '/bin/sh -c /make_iso.sh' returned a non-zero code: 127
    make: *** [deploy/iso/minishift.iso] Error 1

praveenkumar:
This patch will fix b2d iso creation. https://github.com/boot2docker/boot2docker/commit/2d5e77def87a9b9daa156c44bc17d8af183a4744 what caused this issue.

### DIFF
--- a/deploy/iso/Dockerfile
+++ b/deploy/iso/Dockerfile
@@ -42,7 +42,7 @@ RUN curl -fSL -o /tmp/dockerbin.tgz https://get.docker.com/builds/Linux/x86_64/d
 RUN $ROOTFS/usr/local/bin/docker -v
 
 # Build the actual iso image
-RUN /make_iso.sh
+RUN /tmp/make_iso.sh
 
 # Output it to transfer back to the host
 CMD ["cat", "boot2docker.iso"]


hferentschik:
thanks @praveenkumar. Great work to figure this one out. Any idea whether this is a recent change? I am trying to understand how the release would have worked without it.

@jimmidyson, ooi, did you actually build the ISO during the last view minishift releases, or did you sneakingly copy over the old one. If you actually build the ISO, I am trying to understand how you did it. Also, I am wondering whether it is "safe" to build and upload the latest.

LalatenduMohanty:
@praveenkumar thanks for the fix.  Any reason it was sent to cluster-up branch and not to `master` branch directly. cc @hferentschik

hferentschik:
Because for simplicity reason it made more sense to have it on _cluster-up_. It was discovered on the _cluster-up_ branch as well. We did all work coming building up to this point, on cluster-up. Once we release the beta1, we will be back on master and moving forward there won't be a need for a different development branch.

LalatenduMohanty:
@hferentschik  IMO for simplicity reason it should be on master. Because cluster-up branch is a feature branch and should be only used for changes related to "oc cluster up" feature and this change has nothing to do with "cluster up" feature.

hferentschik:
>  IMO for simplicity reason it should be on master.

I disagree. This changes were necessary to test and prepare the next release from the _cluster-up_ branch. If we'd applied it on  master, I would have needed to merge it into cluster-up as well. Or at least rebase cluster-up. Given that _cluster-up_ is the next release branch, it made sense to merge into cluster-up.

> Because cluster-up branch is a feature branch and should be only used for changes related to "oc cluster up" feature

So what is your issue really? Why to go through so much trouble a day before release and a day before the whole discussion becomes obsolete anyways?

LalatenduMohanty:
> So what is your issue really? 

Because we are tweaking the general process around how feature branches work and setting a wrong example.

Issue #197, opened 2016-12-19
Allow for interactive selection of the OpenShift version during 'minishift start'


hferentschik:
Following up on issue #68.

It would be nice, if the user by specifying a flag, could be interactively presented with the (compatible) OpenShift versions which can be used as part of 'minishift start'.

LalatenduMohanty:
I think this is a must for CDK use case.

praveenkumar:
@hferentschik Shouldn't we have it already now, I think we can close it?

hferentschik:
No we have a way to list versions in a separate command, but no way to interactively chose a version. Whether this really is so important is another question.

jeremyeder:
whats the command to list versions?

jorgemoralespou:
@jeremyeder  

````
$ minishift openshift version list
The following OpenShift versions are available:
	- v1.5.1
	- v3.6.0-alpha.0
	- v3.6.0-alpha.1
	- v3.6.0-alpha.2
````

budhrg:
Revisit during discussion of https://github.com/minishift/minishift/issues/1417.

LalatenduMohanty:
@minishift/minishift-dev WDYT about this issue. I think we are pretty good with current user experience i.e. `minishift openshift version list` and we do not need change this.

budhrg:
>  I think we are pretty good with current user experience i.e. minishift openshift version list and we do not need change this.

Yes. Better to close it instead of updating milestones each time.

Issue #182, opened 2016-12-08
Unable to start minishift with --iso-url option on Windows


budhrg:
The `minishift start` have option `--iso-url` which can be used for customized iso instead of default boot2docker iso.

However, I am not able to start VM using this option in Windows.

It might be due to path issue in Windows.

Tried following different combinations:
- `minishift start --iso-url file:D:/minishift-centos.iso`
- `minishift start --iso-url file:D:\minishift-centos.iso`
- `minishift start --iso-url file:///D:/minishift-centos.iso`
- `minishift start --iso-url file:///D:\minishift-centos.iso` 
- `minishift start --iso-url D:\minishift-centos.iso`
- `minishift start --iso-url D:/minishift-centos.iso`

Note: Also, checked with url enclosed within double quotes.

Everytime I am getting following error:
```
PS C:\Users\budhram> minishift-windows-amd64.exe start --iso-url file:///d:\minishift-centos.iso --v=5 --show-libmachine-logs
Starting local OpenShift cluster...
Found binary path at D:\my_bin\minishift-windows-amd64.exe
Launching plugin server for driver hyperv
Plugin server listening at address 127.0.0.1:59674
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(minishift) Calling .GetState
(minishift) DBG | [executing ==>] : C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive ( Get-VM minishift ).state
(minishift) DBG | [stdout =====>] :
(minishift) DBG | [stderr =====>] : Get-VM : Hyper-V was unable to find a virtual machine with name "minishift".
(minishift) DBG | At line:1 char:3
(minishift) Calling .Start
(minishift) DBG | + ( Get-VM minishift ).state
(minishift) DBG | +   ~~~~~~~~~~~~~~~~
(minishift) DBG |     + CategoryInfo          : InvalidArgument: (minishift:String) [Get-VM], VirtualizationException
(minishift) DBG |     + FullyQualifiedErrorId : InvalidParameter,Microsoft.HyperV.PowerShell.Commands.GetVM
(minishift) DBG |
(minishift) DBG |
(minishift) DBG | [executing ==>] : C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive Start-VM minishift
(minishift) DBG | [stdout =====>] :
[...]

PR #198, opened 2016-12-19
[WIP] Issue #190 Editing Cobra doc strings and other log messages

### LINKED ISSUE #190: Review and update command doc strings
While going over the docs to standardize the Minishift name (issue #189 ) I ran into some typos and other issues with the doc strings in the command files.

For the sake of avoiding cross-purposes, I kept the PR focused on the name fixes in doc files, and opened this issue to track improvements and corrections to the doc strings that are important and better done sooner than later.

thatdocslady:
Fixes #190 

This PR includes an overall review and edit of all doc strings (at least all relevant ones that I could find). Review/feedback welcome!

Please note the following during review:

- Search for "NEEDINFO" strings, as I added questions in comments about different aspects of the functions, some are asking for clarification and some expressing need for expanded info

- There is a batch of options in https://github.com/minishift/minishift/blob/master/docs/minishift.md that I can't find anywhere in the code, I tried searching for the option names as well as text, and I can't find their source. Help appreciated.

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -105,9 +105,14 @@ var settings []Setting = []Setting{
 
 var ConfigCmd = &cobra.Command{
 	Use:   "config SUBCOMMAND [flags]",
-	Short: "Modify minishift config",
-	Long: `config modifies minishift config files using subcommands like "minishift config set vm-driver kvm"
-Configurable fields: ` + "\n\n" + configurableFields(),
+	Short: "Modifies Minishift configuration properties.",
+	//NEEDINFO: can we do this any time? or do we need to stop the cluster first? how many config files can we have?
+	Long: `Modifies Minishift configuration properties. The configuration properties in the file are parallel to
+	the options that you set when you run the `minishift start` command. You can set the configuration in
+	this file and point to it during cluster start instead of setting the options directly in the command.
+
+Configurable properties (enter as SUBCOMMAND): ` + "\n\n" + configurableFields(),
+//NEEDINFO: There are properties here that aren't in the start command and I can't find them anywhere. Also, is there a way to link to the start cmd help file in the "see also"?
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
 	},
@@ -128,12 +133,12 @@ func ReadConfig() (MinikubeConfig, error) {
 		if os.IsNotExist(err) {
 			return make(map[string]interface{}), nil
 		}
-		return nil, fmt.Errorf("Could not open file %s: %s", constants.ConfigFile, err)
+		return nil, fmt.Errorf("Cannot open file %s: %s", constants.ConfigFile, err)
 	}
 	var m MinikubeConfig
 	m, err = decode(f)
 	if err != nil {
-		return nil, fmt.Errorf("Could not decode config %s: %s", constants.ConfigFile, err)
+		return nil, fmt.Errorf("Cannot read config %s: %s", constants.ConfigFile, err)
 	}
 
 	return m, nil
@@ -143,12 +148,12 @@ func ReadConfig() (MinikubeConfig, error) {
 func WriteConfig(m MinikubeConfig) error {
 	f, err := os.Create(constants.ConfigFile)
 	if err != nil {
-		return fmt.Errorf("Could not open file %s: %s", constants.ConfigFile, err)
+		return fmt.Errorf("Cannot open file %s: %s", constants.ConfigFile, err)
 	}
 	defer f.Close()
 	err = encode(f, m)
 	if err != nil {
-		return fmt.Errorf("Error encoding config %s: %s", constants.ConfigFile, err)
+		return fmt.Errorf("Cannot write config %s: %s", constants.ConfigFile, err)
 	}
 	return nil
 }
--- a/cmd/minikube/cmd/config/get.go
+++ b/cmd/minikube/cmd/config/get.go
@@ -25,8 +25,9 @@ import (
 
 var configGetCmd = &cobra.Command{
 	Use:   "get PROPERTY_NAME",
-	Short: "Gets the value of PROPERTY_NAME from the minishift config file",
-	Long:  "Returns the value of PROPERTY_NAME from the minishift config file.  Can be overwritten at runtime by flags or environmental variables.",
+	Short: "Gets the value of a configuration property from the Minishift configuration file",
+	Long:  "Gets the value of a configuration property from the minishift configuration file. This value can be overwritten at runtime by flags or environmental variables.",
+	//NEEDINFO: Why the runtime clarification? Is this to warn that the value might not be as expected? This cmd doesn't set or change anything, why does this matter? 
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(args) != 1 {
 			fmt.Fprintln(os.Stderr, "usage: minishift config get PROPERTY_NAME")
--- a/cmd/minikube/cmd/config/set_test.go
+++ b/cmd/minikube/cmd/config/set_test.go
@@ -22,5 +22,6 @@ func TestNotFound(t *testing.T) {
 	err := set("nonexistant", "10")
 	if err == nil {
 		t.Fatalf("Set did not return error for unknown property")
+		//NEEDINFO: So it did or didn't? What is the error here? :)
 	}
 }
--- a/cmd/minikube/cmd/config/unset.go
+++ b/cmd/minikube/cmd/config/unset.go
@@ -25,8 +25,9 @@ import (
 
 var configUnsetCmd = &cobra.Command{
 	Use:   "unset PROPERTY_NAME",
-	Short: "unsets an individual value in a minishift config file",
-	Long:  "unsets PROPERTY_NAME from the minishift config file.  Can be overwritten by flags or environmental variables",
+	Short: "Clears the value of a configuration property in the Minishift configuration file.",
+	Long:  "Clears the value of a configuration property in the Minishift configuration file. The value can be overwritten at runtime by flags or environmental variables",
+	//NEEDINFO: can I run this any time? what supercedes if I do this at runtime?
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(args) != 1 {
 			fmt.Fprintf(os.Stdout, "usage: minishift config unset PROPERTY_NAME")
--- a/cmd/minikube/cmd/config/view.go
+++ b/cmd/minikube/cmd/config/view.go
@@ -36,8 +36,9 @@ type ConfigViewTemplate struct {
 
 var configViewCmd = &cobra.Command{
 	Use:   "view",
-	Short: "Display values currently set in the minishift config file",
-	Long:  "Display values currently set in the minishift  config file.",
+	Short: "Display the properties and values of the Minishift configuration file.",
+	Long:  "Display the properties and values of the Minishift configuration file. You can set the output format from one of the available Go templates.",
+	//NEEDINFO: is the format option required? what happens if we don't set this option? STDOUT?
 	Run: func(cmd *cobra.Command, args []string) {
 		err := configView()
 		if err != nil {
@@ -49,8 +50,8 @@ var configViewCmd = &cobra.Command{
 
 func init() {
 	configViewCmd.Flags().StringVar(&configViewFormat, "format", constants.DefaultConfigViewFormat,
-		`Go template format string for the config view output.  The format for Go templates can be found here: https://golang.org/pkg/text/template/
-For the list of accessible variables for the template, see the struct values here: https://godoc.org/github.com/jimmidyson/minishift/cmd/minikube/cmd/config#ConfigViewTemplate`)
+		`Go template format to apply to the configuration file. For more information about Go templates, see: https://golang.org/pkg/text/template/
+For the list of configurable variables for the template, see the struct values section of ConfigViewTemplate at: https://godoc.org/github.com/minishift/minishift/cmd/minikube/cmd/config#ConfigViewTemplate`)
 }
 
 func init() {
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -30,18 +30,18 @@ import (
 var deleteCmd = &cobra.Command{
 	Use:   "delete",
 	Short: "Deletes a local OpenShift cluster.",
-	Long: `Deletes a local OpenShift cluster. This command deletes the VM, and removes all
-associated files.`,
+	Long: `Deletes a local OpenShift cluster, including the Minishift VM and all associated files.`,
+	//MOREINFO: Can we delete more than one? Is there a way to exclude files? Do we want to elaborate on usage or examples?
 	Run: func(cmd *cobra.Command, args []string) {
-		fmt.Println("Deleting local OpenShift cluster...")
+		fmt.Println("Deleting the local OpenShift cluster...")
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
 
 		if err := cluster.DeleteHost(api); err != nil {
-			fmt.Println("Errors occurred deleting machine: ", err)
+			fmt.Println("Error deleting cluster: ", err)
 			os.Exit(1)
 		}
-		fmt.Println("Machine deleted.")
+		fmt.Println("Cluster deleted.")
 	},
 }
 
--- a/cmd/minikube/cmd/get_openshift_versions.go
+++ b/cmd/minikube/cmd/get_openshift_versions.go
@@ -26,8 +26,9 @@ import (
 // getVersionsCmd represents the ip command
 var getVersionsCmd = &cobra.Command{
 	Use:   "get-openshift-versions",
-	Short: "Gets the list of available OpenShift versions available for minishift.",
-	Long:  `Gets the list of available OpenShift versions available for minishift.`,
+	Short: "Gets the list of OpenShift versions available for Minishift.",
+	Long:  `Gets the list of OpenShift versions available for Minishift.`,
+	//MOREINFO: Do we want to elaborate on usage or examples? Like is this STDOUT only, can we output to a file, do we want to pass this onto something else?
 	Run: func(cmd *cobra.Command, args []string) {
 		openshiftversions.PrintOpenShiftVersionsFromGitHub(os.Stdout)
 	},
--- a/cmd/minikube/cmd/logs.go
+++ b/cmd/minikube/cmd/logs.go
@@ -30,14 +30,16 @@ import (
 // logsCmd represents the logs command
 var logsCmd = &cobra.Command{
 	Use:   "logs",
-	Short: "Gets the logs of the running OpenShift instance, used for debugging minishift, not user code.",
-	Long:  `Gets the logs of the running OpenShift instance, used for debugging minishift, not user code.`,
+	Short: "Gets the logs of the running Minishift instance.",
+	Long:  `Gets the logs of the running Minishift instance and prints them to the standard output. You use these log messages to debug Minishift.
+	The logs do not contain information about your application code.`,
+	//NEEDINFO: Is it stdout? What if I want to output to a file? How can users debug their apps? Anything else? What about log levels (i.e. debug, warning, info)?
 	Run: func(cmd *cobra.Command, args []string) {
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
 		s, err := cluster.GetHostLogs(api)
 		if err != nil {
-			log.Println("Error getting machine logs:", err)
+			log.Println("Error getting logs:", err)
 			os.Exit(1)
 		}
 		fmt.Fprintln(os.Stdout, s)
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -75,8 +75,10 @@ var (
 var startCmd = &cobra.Command{
 	Use:   commandName,
 	Short: "Starts a local OpenShift cluster.",
-	Long: `Starts a local OpenShift cluster using Virtualbox. This command
-assumes you already have Virtualbox installed.`,
+	Long: `Starts a local OpenShift cluster using VirtualBox. You must have VirtualBox installed
+	to start the cluster. If you have an existing configuration file, you can use it by setting the
+	--use-existing-config option.`,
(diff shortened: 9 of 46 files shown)


hferentschik:
@thatdocslady, you need to make sure that the code still compiles when you change these strings:

    $ make

Right now it seems you broke the build.

Also, if you change the Cobra CLI strings, you need to run:

   $ make prerelease

Since you need to check in the updated markdown files as well.

Also, could you please split this into two commits? I thought you only wanted to change the Cobra strings?

cmd/minikube/cmd/config/config.go
```diff
@@ -105,9 +105,14 @@ var settings []Setting = []Setting{
 
 var ConfigCmd = &cobra.Command{
 	Use:   "config SUBCOMMAND [flags]",
-	Short: "Modify minishift config",
-	Long: `config modifies minishift config files using subcommands like "minishift config set vm-driver kvm"
-Configurable fields: ` + "\n\n" + configurableFields(),
+	Short: "Modifies Minishift configuration properties.",
+	//NEEDINFO: can we do this any time? or do we need to stop the cluster first? how many config files can we have?
+	Long: `Modifies Minishift configuration properties. The configuration properties in the file are parallel to
```
hferentschik: " The configuration properties in the file are parallel" - what does "in parallel" mean in this context?

cmd/minikube/cmd/config/config.go
```diff
@@ -105,9 +105,14 @@ var settings []Setting = []Setting{
...
+	the options that you set when you run the `minishift start` command. You can set the configuration in
+	this file and point to it during cluster start instead of setting the options directly in the command.
```
hferentschik: > You can set the configuration in this file and point to it during cluster start instead of setting the options directly in the command

Where did you get this from?

cmd/minikube/cmd/config/config.go
```diff
@@ -128,12 +133,12 @@ func ReadConfig() (MinikubeConfig, error) {
 		if os.IsNotExist(err) {
 			return make(map[string]interface{}), nil
 		}
-		return nil, fmt.Errorf("Could not open file %s: %s", constants.ConfigFile, err)
+		return nil, fmt.Errorf("Cannot open file %s: %s", constants.ConfigFile, err)
```
hferentschik: Can we separate the changes towards the CLI switches from the error message changes?

cmd/minikube/cmd/config/config.go
```diff
@@ -143,12 +148,12 @@ func ReadConfig() (MinikubeConfig, error) {
...
 	f, err := os.Create(constants.ConfigFile)
 	if err != nil {
-		return fmt.Errorf("Could not open file %s: %s", constants.ConfigFile, err)
+		return fmt.Errorf("Cannot open file %s: %s", constants.ConfigFile, err)
 	}
 	defer f.Close()
 	err = encode(f, m)
 	if err != nil {
-		return fmt.Errorf("Error encoding config %s: %s", constants.ConfigFile, err)
+		return fmt.Errorf("Cannot write config %s: %s", constants.ConfigFile, err)
```
hferentschik: So why did you switch from "encoding" to "cannot write". This seems to be loosing some information. In the context of error message I think we should keep as much context as possible.

cmd/minikube/cmd/config/get.go
```diff
@@ -25,8 +25,9 @@ import (
 
 var configGetCmd = &cobra.Command{
 	Use:   "get PROPERTY_NAME",
-	Short: "Gets the value of PROPERTY_NAME from the minishift config file",
-	Long:  "Returns the value of PROPERTY_NAME from the minishift config file.  Can be overwritten at runtime by flags or environmental variables.",
+	Short: "Gets the value of a configuration property from the Minishift configuration file",
+	Long:  "Gets the value of a configuration property from the minishift configuration file. This value can be overwritten at runtime by flags or environmental variables.",
```
hferentschik: Why would you use lowercase "minishift"?

hferentschik: It should be "Minishift" right?

cmd/minikube/cmd/config/set_test.go
```diff
@@ -22,5 +22,6 @@ func TestNotFound(t *testing.T) {
 	err := set("nonexistant", "10")
 	if err == nil {
 		t.Fatalf("Set did not return error for unknown property")
+		//NEEDINFO: So it did or didn't? What is the error here? :)
```
hferentschik: This is a test. I don't think it matters so much here

cmd/minikube/cmd/config/unset.go
```diff
@@ -25,8 +25,9 @@ import (
 
 var configUnsetCmd = &cobra.Command{
 	Use:   "unset PROPERTY_NAME",
-	Short: "unsets an individual value in a minishift config file",
-	Long:  "unsets PROPERTY_NAME from the minishift config file.  Can be overwritten by flags or environmental variables",
+	Short: "Clears the value of a configuration property in the Minishift configuration file.",
+	Long:  "Clears the value of a configuration property in the Minishift configuration file. The value can be overwritten at runtime by flags or environmental variables",
```
hferentschik: I know this is the intention, but did you confirm that works? This is something I actually wanted to investigate, but have not come around yet.

cmd/minikube/cmd/config/view.go
```diff
@@ -36,8 +36,9 @@ type ConfigViewTemplate struct {
 
 var configViewCmd = &cobra.Command{
 	Use:   "view",
-	Short: "Display values currently set in the minishift config file",
-	Long:  "Display values currently set in the minishift  config file.",
+	Short: "Display the properties and values of the Minishift configuration file.",
+	Long:  "Display the properties and values of the Minishift configuration file. You can set the output format from one of the available Go templates.",
```
hferentschik: How does the template stuff work?

cmd/minikube/cmd/delete.go
```diff
@@ -30,18 +30,18 @@ import (
 var deleteCmd = &cobra.Command{
 	Use:   "delete",
 	Short: "Deletes a local OpenShift cluster.",
-	Long: `Deletes a local OpenShift cluster. This command deletes the VM, and removes all
-associated files.`,
+	Long: `Deletes a local OpenShift cluster, including the Minishift VM and all associated files.`,
```
hferentschik: This is actually not correct. _delete_ really just deletes the VM.

cmd/minikube/cmd/get_openshift_versions.go
```diff
@@ -26,8 +26,9 @@ import (
 // getVersionsCmd represents the ip command
 var getVersionsCmd = &cobra.Command{
 	Use:   "get-openshift-versions",
-	Short: "Gets the list of available OpenShift versions available for minishift.",
-	Long:  `Gets the list of available OpenShift versions available for minishift.`,
+	Short: "Gets the list of OpenShift versions available for Minishift.",
+	Long:  `Gets the list of OpenShift versions available for Minishift.`,
+	//MOREINFO: Do we want to elaborate on usage or examples? Like is this STDOUT only, can we output to a file, do we want to pass this onto something else?
```
hferentschik: How do you want to handle these _MOREINFO_s? Some of these things have some related issues. Is the plan to merge and keep some of these comments? Or are you planning to address them all prior to merging? I think for some of your comments, it makes more sense to create trackable issues, instead of having some comments in the code. Regarding this particular issue, it maps to https://github.com/minishift/minishift/issues/197

cmd/minikube/cmd/logs.go
```diff
@@ -30,14 +30,16 @@ import (
 // logsCmd represents the logs command
 var logsCmd = &cobra.Command{
 	Use:   "logs",
-	Short: "Gets the logs of the running OpenShift instance, used for debugging minishift, not user code.",
-	Long:  `Gets the logs of the running OpenShift instance, used for debugging minishift, not user code.`,
+	Short: "Gets the logs of the running Minishift instance.",
+	Long:  `Gets the logs of the running Minishift instance and prints them to the standard output. You use these log messages to debug Minishift.
```
hferentschik: "You use these log messages to debug Minishift" - Do we really need this?

cmd/minikube/cmd/start.go
```diff
@@ -75,8 +75,10 @@ var (
 var startCmd = &cobra.Command{
 	Use:   commandName,
 	Short: "Starts a local OpenShift cluster.",
-	Long: `Starts a local OpenShift cluster using Virtualbox. This command
-assumes you already have Virtualbox installed.`,
+	Long: `Starts a local OpenShift cluster using VirtualBox. You must have VirtualBox installed
```
hferentschik: I think this is not quite correct. I need to double check. I think the default varies depends on the OS.

cmd/minikube/cmd/start.go
```diff
@@ -171,32 +173,43 @@ func init() {
...
+	//NEEDINFO: Is this the same as the binary? we should be consistent in how we refer to it. Also, we need to change the default from jdyson.
+	startFlagSet.String(vmDriver, constants.DefaultVMDriver, fmt.Sprintf("The driver to use for the Minishift VM. Possible values: %v", constants.SupportedVMDrivers))
+	//NEEDINFO: If virtualbox is the virt provider how come kvm is the default driver and how are drivers related to this at all?
+	startFlagSet.Int(memory, constants.DefaultMemory, "Amount of RAM to allocate to the Minishift VM.")
+	startFlagSet.Int(cpus, constants.DefaultCPUS, "Number of CPU cores to allocate to the Minishift VM.")
+	startFlagSet.String(humanReadableDiskSize, constants.DefaultDiskSize, "Disk size to allocate to the Minishift VM. Use the format <size><unit>, where unit = b, k, m or g.")
 	startFlagSet.String(hostOnlyCIDR, "192.168.99.1/24", "The CIDR to be used for the minishift VM (only supported with Virtualbox driver)")
-	startFlagSet.StringSliceVar(&dockerEnv, "docker-env", nil, "Environment variables to pass to the Docker daemon. (format: key=value)")
+	//NEEDINFO: What is a CIDR?
```
hferentschik: https://en.wikipedia.org/wiki/Classless_Inter-Domain_Routing

cmd/minikube/cmd/start.go
```diff
@@ -171,32 +173,43 @@ func init() {
 
 // initClusterUpFlags creates the CLI flags which needs to be passed on to 'oc cluster up'
 func initStartFlags() {
-	startFlagSet.String(isoURL, constants.DefaultIsoUrl, "Location of the minishift iso")
-	startFlagSet.String(vmDriver, constants.DefaultVMDriver, fmt.Sprintf("VM driver is one of: %v", constants.SupportedVMDrivers))
-	startFlagSet.Int(memory, constants.DefaultMemory, "Amount of RAM allocated to the minishift VM")
-	startFlagSet.Int(cpus, constants.DefaultCPUS, "Number of CPUs allocated to the minishift VM")
-	startFlagSet.String(humanReadableDiskSize, constants.DefaultDiskSize, "Disk size allocated to the minishift VM (format: <number>[<unit>], where unit = b, k, m or g)")
+	startFlagSet.String(isoURL, constants.DefaultIsoUrl, "Location of the minishift ISO.")
+	//NEEDINFO: Is this the same as the binary? we should be consistent in how we refer to it. Also, we need to change the default from jdyson.
```
hferentschik: which binary? This is the ISO image. The image in which we run OpenShift.

cmd/minikube/cmd/start.go
```diff
@@ -171,32 +173,43 @@ func init() {
...
+	//NEEDINFO: If virtualbox is the virt provider how come kvm is the default driver and how are drivers related to this at all?
+	startFlagSet.Int(memory, constants.DefaultMemory, "Amount of RAM to allocate to the Minishift VM.")
+	startFlagSet.Int(cpus, constants.DefaultCPUS, "Number of CPU cores to allocate to the Minishift VM.")
+	startFlagSet.String(humanReadableDiskSize, constants.DefaultDiskSize, "Disk size to allocate to the Minishift VM. Use the format <size><unit>, where unit = b, k, m or g.")
 	startFlagSet.String(hostOnlyCIDR, "192.168.99.1/24", "The CIDR to be used for the minishift VM (only supported with Virtualbox driver)")
-	startFlagSet.StringSliceVar(&dockerEnv, "docker-env", nil, "Environment variables to pass to the Docker daemon. (format: key=value)")
+	//NEEDINFO: What is a CIDR?
+	startFlagSet.StringSliceVar(&dockerEnv, "docker-env", nil, "Environment variables to pass to the Docker daemon. Use the format <key>=<value>.")
+	//NEEDINFO: How do you separate multiple env vars? Comma? Space + comma? etc..
```
hferentschik: Comma I think, but needs to be confirmed.

hferentschik: This might be a good one for which we need to have a follow up issue. There are several things we need to test and verify yet, before we can change some of these things. Right now it does not have a high priority imo.

cmd/minikube/cmd/start.go
```diff
@@ -171,32 +173,43 @@ func init() {
...
 	startFlagSet.StringSliceVar(&insecureRegistry, "insecure-registry", []string{"172.30.0.0/16"}, "Insecure Docker registries to pass to the Docker daemon")
-	startFlagSet.StringSliceVar(&registryMirror, "registry-mirror", nil, "Registry mirrors to pass to the Docker daemon")
-	startFlagSet.String(openshiftVersion, version.GetOpenShiftVersion(), "The OpenShift version that the minishift VM will run (ex: v1.2.3)")
+	//NEEDINFO: "insecure" is a really inappropriate word, not sure what it's supposed to mean but we really cannot use it.
+	startFlagSet.StringSliceVar(&registryMirror, "registry-mirror", nil, "Registry mirrors to pass to the Docker daemon.")
+	//NEEDINFO: Is this URLs of the mirrors? hostnames? also, there's a blank default and that looks weird, either remove it or populate it
```
hferentschik: We cannot just remove this. In this case empty string is the default, I think.

cmd/minikube/cmd/start.go
```diff
@@ -171,32 +173,43 @@ func init() {
...
-	clusterUpFlagSet.Int(serverLogLevel, 0, "Log level for OpenShift server")
-	clusterUpFlagSet.StringSliceVarP(&openShiftEnv, openshiftEnv, "e", []string{}, "Specify key value pairs of environment variables to set on OpenShift container")
+	clusterUpFlagSet.Bool(skipRegistryCheck, false, "Skip the Docker daemon registry check.")
+	clusterUpFlagSet.String(publicHostname, "", "Public host name of the OpenShift cluster.")
+	clusterUpFlagSet.String(routingSuffix, "", "Default suffix for the server routes.")
+	clusterUpFlagSet.Bool(useExistingConfig, false, "Use existing configuration if available.")
+	clusterUpFlagSet.String(hostConfigDir, dockerhost.DefaultConfigDir, "Location of the OpenShift configuration files on the host.")
+	clusterUpFlagSet.String(hostVolumesDir, dockerhost.DefaultVolumesDir, "Location of the OpenShift volumes on the host.")
+	clusterUpFlagSet.String(hostDataDir, "", "Location of the OpenShift data on the host. If not specified, etcd data will not be persisted on the host.")
+	//NEEDINFO: what does etcd data mean? persistent?
```
hferentschik: This is the cluster state information. Everything which tells OpenShift how to operate. OpenShift cluster communication is built upon _etcd_

cmd/minikube/cmd/start.go
```diff
@@ -171,32 +173,43 @@ func init() {
...
+	clusterUpFlagSet.Bool(forwardPorts, false, "Use Docker port forwarding to communicate with origin container. Requires 'socat' locally.")
+	//NEEDINFO: origin or original? origin might not be the right word here. what is socat?
+	clusterUpFlagSet.Int(serverLogLevel, 0, "Set the log level for the OpenShift server.")
+	//NEEDINFO: do we need to default and possible values here? what is this used for, debugging openshift? the app? minishift?
+	clusterUpFlagSet.StringSliceVarP(&openShiftEnv, openshiftEnv, "e", []string{}, "Specify key-value pairs of environment variables to pass to the OpenShift container.")
+	//NEEDINFO: This is the first time I see OpenShift container, isn't this the same as Minishift VM? or app? and what's the difference between this and the Docker daemon env vars?
```
hferentschik: > OpenShift container, isn't this the same as Minishift VM?

Absolutely not. The VM is the thing which runs a Docker daemon. On this Docker daemon we are running a containerized version of OpenShift. It is this container to which these settings are relevant.

cmd/minikube/cmd/start.go
```diff
@@ -229,7 +242,7 @@ func clusterUp(config *cluster.MachineConfig) {
 	err = runner.Run(cmdName, cmdArgs...)
 	if err != nil {
 		// TODO glog is probably not right here. Need some sort of logging wrapper
-		glog.Errorln("Error starting 'cluster up': ", err)
+		glog.Errorln("Error starting the cluster: ", err)
```
hferentschik: Well, I intentionally used 'cluster up', since this is the command which gets executed.

cmd/minikube/cmd/stop.go
```diff
@@ -30,18 +30,20 @@ import (
...
+	VM but does not delete any associated files. To start the cluster again, use the start command.`,
+	//NEEDINFO: is this a graceful stop? what would make it impossible to stop? can you force stop somehow? and what does virtualbox have to do with it?
 	Run: func(cmd *cobra.Command, args []string) {
 		fmt.Println("Stopping local OpenShift cluster...")
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
 
 		if err := cluster.StopHost(api); err != nil {
-			fmt.Println("Error stopping machine: ", err)
+			fmt.Println("Error stopping cluster: ", err)
```
hferentschik: Actually I want to get away from "cluster". This is in our context a very misleading term. A cluster implies to me that there are multiple nodes which is not the case for us. I think we should just say "Starting VM and provisioning OpenShift" and then "Stopping virtual machine". As said, let's try to get away from cluster or at least be more specific like "single node cluster" or something.

hferentschik:
> There is a batch of options in https://github.com/minishift/minishift/blob/master/docs/minishift.md that I can't find anywhere in the code, I tried searching for the option names as well as text, and I can't find their source. Help appreciated.

Which one in particular

cmd/minikube/cmd/start.go
hferentschik: "Starts a single node OpenShift cluster running on the selected hypervisor"

VirtualBox is only needed if it is the selected hypervisor.

cmd/minikube/cmd/config/unset.go
```diff
@@ -25,8 +25,9 @@ import (
...
+	//NEEDINFO: can I run this any time? what supercedes if I do this at runtime?
```
hferentschik: This is something we need a full tutorial around. Let's create an issue for this. I don't think we can do much here in the context of this help text.

cmd/minikube/cmd/start.go
hferentschik: This is an expert settings for OpenShift config. The ones who need it will know what it means.

cmd/minikube/cmd/start.go
hferentschik: btw, we should make clear that this is an URI, so either an http URL or file URI, see https://github.com/minishift/minishift/issues/182. Preferably the help text should have examples for both.

cmd/minikube/cmd/config/config.go
thatdocslady: the power of deduction? :) in the start command there's an option to use existing configuration, and a lot of the config properties in the file are parallel to the start command options, so it made sense that you can do this, but please tell me if my deduction is off the cliff here..

cmd/minikube/cmd/config/config.go
thatdocslady: "parallel" is probably the wrong word here, I want to say that these config properties are equivalent to options in the minishift start command.

cmd/minikube/cmd/config/config.go
thatdocslady: encoding means changing something from one format to another, while in this context the error message for encoding and decoding look like messages for getting/setting. please correct me if I'm off here, or explain what encoding means in this context (I've never seen encode/decode used in a config context)

cmd/minikube/cmd/config/get.go
thatdocslady: aha, missed one :)

cmd/minikube/cmd/config/unset.go
thatdocslady: And either way I need to document expected behavior, if there's a bug in this function we should make a note of it somewhere but the intention/expected behavior should be documented here.

cmd/minikube/cmd/get_openshift_versions.go
thatdocslady: The NEEDINFO comments were basically me asking for clarification or extra info about the content I was editing, with the hope of resolving them in the scope of this PR. Of course if we need to create related issues (for example, if the answer to my question is too big for the scope of the PR) then sure let's go ahead and do that, but I'd like to keep them in the code until resolved if it doesn't get in the way.

cmd/minikube/cmd/logs.go
thatdocslady: This text was already there, and I didn't remove it. My guess would be that we should differentiate what these logs are for, since we're dealing with Minishift>VM>OpenShift>app.

cmd/minikube/cmd/config/config.go
hferentschik: parallel is not the right word I think. This is also one of the things we need to actually test first.

cmd/minikube/cmd/config/config.go
hferentschik: thanks

cmd/minikube/cmd/config/config.go
hferentschik: This is encoding from the Go _MinikubeConfig_ data structure into a JSON document. So yes, formats are changed here.

cmd/minikube/cmd/get_openshift_versions.go
hferentschik: Sounds like a reasonable approach

hferentschik:
I am closing this pull request as it stands now. It does not compile and it won't apply onto master anymore. I think it is better to create a new pull request once things are sorted out and commits are rebased on top of master. Having a clean slate will also make it easier to review then.

thatdocslady:
@hferentschik you beat me to this by about 1hr. I just cherry-picked each of the 3 commits into new branches and I'm in the process of applying your feedback from this PR and submitting 3 new PRs :)

PR #200, opened 2016-12-19
Issue #137 Allow registration for RHEL VM

### LINKED ISSUE #137: Allow for registration of RHEL VM
This ties into #136, since registration needs to happen prior to provisioning.

praveenkumar:
This is skeleton about how registration detector and provisioner going to work. This is currently for early review if something we need to change in the design. 

TODO
=====
- Changes according to review.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -30,6 +30,7 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/cache"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
+	"github.com/minishift/minishift/pkg/minishift/registration"
 	"github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/version"
 	dockerhost "github.com/openshift/origin/pkg/bootstrap/docker/host"
@@ -85,6 +86,9 @@ var startFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
 
 // clusterUpFlagSet contains the command line switches which needs to be passed on to 'cluster up'
 var clusterUpFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
+
+// registerFlagSet contains registration specific switches
+var registerFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
 
 // minishiftToClusterUp is a mapping between falg names used in minishift CLI and flag name as passed to 'cluster up'
 var minishiftToClusterUp = map[string]string{
@@ -143,6 +147,20 @@ func runStart(cmd *cobra.Command, args []string) {
 		os.Setenv(k, v)
 	}
 
+	registrator, err := registration.DetectRegistrator(host.Driver)
+	if err != nil {
+		fmt.Println("Distribution doesn't support registration")
+	} else {
+		fmt.Printf("Registration with %s...", registrator)
+		m := make(map[string]string)
+		m["username"] = viper.GetString(username)
+		m["password"] = viper.GetString(password)
+		if err := registrator.Register(m); err != nil {
+			fmt.Errorf("Error running registration: %s", err)
+			os.Exit(1)
+		}
+	}
+
 	clusterUp(&config)
 }
 
@@ -162,6 +180,7 @@ func init() {
 
 	startCmd.Flags().AddFlagSet(startFlagSet)
 	startCmd.Flags().AddFlagSet(clusterUpFlagSet)
+	startCmd.Flags().AddFlagSet(registerFlagSet)
 
 	viper.BindPFlags(startCmd.Flags())
 	RootCmd.AddCommand(startCmd)
--- a/cmd/minikube/cmd/stop.go
+++ b/cmd/minikube/cmd/stop.go
@@ -23,6 +23,7 @@ import (
 	"github.com/docker/machine/libmachine"
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/registration"
 	"github.com/spf13/cobra"
 )
 
@@ -36,6 +37,22 @@ itself, leaving all files intact. The cluster can be started again with the "sta
 		fmt.Println("Stopping local OpenShift cluster...")
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
+		// Unregister VM when stop
+		host, err := api.Load(constants.MachineName)
+		if err != nil {
+			fmt.Println("Error stopping machine: ", err)
+			os.Exit(1)
+		}
+		registrator, err := registration.DetectRegistrator(host.Driver)
+		if err != nil {
+			fmt.Println("Distribution doesn't support unregister")
+		} else {
+			fmt.Println("Unregistering distribution...")
+			if err := registrator.Unregister(); err != nil {
+				fmt.Println("Error Unregistering distribution ", err)
+				os.Exit(1)
+			}
+		}
 
 		if err := cluster.StopHost(api); err != nil {
 			fmt.Println("Error stopping machine: ", err)
--- a/pkg/minishift/registration/redhat.go
+++ b/pkg/minishift/registration/redhat.go
@@ -1,0 +1,82 @@
+/*
+Copyright (C) 2016 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package registration
+
+import (
+	"fmt"
+	"github.com/docker/machine/libmachine/drivers"
+	"github.com/docker/machine/libmachine/provision"
+	"strings"
+)
+
+func init() {
+	Register("Redhat", &RegisteredRegistrator{
+		New: NewRedhatRegistrator,
+	})
+}
+
+func NewRedhatRegistrator(d drivers.Driver) Registrator {
+	return &RedhatRegistrator{
+		SSHCommander: provision.GenericSSHCommander{Driver: d},
+	}
+}
+
+type RedhatRegistrator struct {
+	provision.SSHCommander
+}
+
+func (registrator *RedhatRegistrator) CompatibleWithDistribution(osReleaseInfo *provision.OsRelease) bool {
+	if osReleaseInfo.ID != "rhel" {
+		return false
+	}
+	if _, err := registrator.SSHCommand("sudo subscription-manager version"); err != nil {
+		return false
+	} else {
+		return true
+	}
+}
+
+func (registrator *RedhatRegistrator) Register(register_options map[string]string) error {
+	if output, err := registrator.SSHCommand("sudo subscription-manager version"); err != nil {
+		return err
+	} else {
+		if strings.Contains(output, "not registered") {
+			subscriptionCommand := "sudo subscription-manager register --auto-attach"
+			for key, value := range register_options {
+				subscriptionCommand += fmt.Sprintf(" --%s %s", key, value)
+			}
+			if _, err := registrator.SSHCommand(subscriptionCommand); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+func (registrator *RedhatRegistrator) Unregister() error {
+	if output, err := registrator.SSHCommand("sudo subscription-manager version"); err != nil {
+		return err
+	} else {
+		if !strings.Contains(output, "not registered") {
+			if _, err := registrator.SSHCommand(
+				"sudo subscription-manager unregister"); err != nil {
+				return err
+			}
+		}
+	}
+	return nil
+}
--- a/pkg/minishift/registration/redhat_test.go
+++ b/pkg/minishift/registration/redhat_test.go
@@ -1,0 +1,115 @@
+/*
+Copyright (C) 2016 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package registration
+
+import (
+	"github.com/docker/machine/libmachine/drivers"
+	"github.com/minishift/minishift/pkg/minikube/tests"
+	"testing"
+	"github.com/docker/machine/libmachine/provision"
+)
+
+func TestRedhatRegistratorCompatibleWithDistribution(t *testing.T) {
+	s, _ := tests.NewSSHServer()
+	s.CommandToOutput = make(map[string]string)
+	s.CommandToOutput["sudo subscription-manager version"] = `server type: This system is currently not registered.
+subscription management server: 0.9.51.11-1
+subscription management rules: 5.15
+subscription-manager: 1.17.15-1.el7
+python-rhsm: 1.17.9-1.el7
+`
+	port, err := s.Start()
+	if err != nil {
+		t.Fatalf("Error starting ssh server: %s", err)
+	}
+	d := &tests.MockDriver{
+		Port: port,
+		BaseDriver: drivers.BaseDriver{
+			IPAddress:  "127.0.0.1",
+			SSHKeyPath: "",
+		},
+	}
+
+	registrator := NewRedhatRegistrator(d)
+	info := &provision.OsRelease{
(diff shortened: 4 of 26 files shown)


pkg/minishift/registration/redhat.go
```diff
@@ -0,0 +1,110 @@
+/*
```
praveenkumar: This is for Redhat specific registration provisioner.

pkg/minishift/registration/redhat.go
```diff
@@ -0,0 +1,110 @@
...
+
+type GenericSSHCommander struct {
+	Driver drivers.Driver
+}
+
+func (sshCmder GenericSSHCommander) SSHCommand(args string) (string, error) {
+	return drivers.RunSSHCommandFromDriver(sshCmder.Driver, args)
+}
+
+func credentials() (string, string) {
```
praveenkumar: This function will not be part of final PR, currently it's there to test the registration works as expected and we have plan to pass username/password from the command line

pkg/minishift/registration/registration.go
```diff
@@ -0,0 +1,91 @@
...
+func (detector StandardRegistrator) DetectRegistrator(d drivers.Driver) (Registrator, error) {
+	log.Info("Waiting for SSH to be available...")
+	if err := drivers.WaitForSSH(d); err != nil {
+		return nil, err
+	}
+
+	log.Info("Detecting the Registration Distro...")
+
+	for _, r := range registrators {
+		registration := r.New(d)
```
hferentschik: "registrar" =! "registration"

Again, I am not saying that "registrar" is the best word here and I am happy to find am more fitting term, but going from "registrators" to "registration" is wrong. Let's try to use correct and telling variable names.

pkg/minishift/registration/registration.go
```diff
@@ -0,0 +1,91 @@
...
+
+	// Return the auth options used to configure remote connection for the daemon.
+	GetUnregister() error
+
+	// Figure out if this is the right registration tool to use based on linux distro
+	CompatibleWithDistribution() bool
+}
+
+// RegisteredProvisioner creates a new provisioner
+type RegisteredRegistrator struct {
```
hferentschik: And why do you need this type?

hferentschik: Why can it not just be "func Register(name string, r *Registrar)"?

pkg/minishift/registration/registration.go
```diff
@@ -0,0 +1,91 @@
...
+
+func SetDetector(newDetector Detector) {
+	detector = newDetector
+}
+
+// Registration defines distribution specific actions
+type Registrator interface {
+	SSHCommander
+	// Get the directory where the settings files for docker are to be found
+	GetRegister() error
```
hferentschik: Also, this is really where you need to pass the credentials, right? If you want to mock this up, you should add the missing parameters at least, instead of internally just calling "credentials()". We want to get the interface right, no?

cmd/minikube/cmd/start.go
```diff
@@ -143,6 +144,16 @@ func runStart(cmd *cobra.Command, args []string) {
 		os.Setenv(k, v)
 	}
 
+	registrator, err := registration.DetectRegistrator(host.Driver)
+	if err != nil {
+		glog.Info("Distro doesn't support registration")
```
hferentschik: Don't use glog at this level. If you want the user to see this, then use ftm.Println for now until we have a proper logging facade.

cmd/minikube/cmd/start.go
```diff
@@ -143,6 +144,16 @@ func runStart(cmd *cobra.Command, args []string) {
...
+	} else {
+		glog.Infof("Registration with %s...", registrator)
+		if err := registrator.GetRegister(); err != nil {
+			fmt.Errorf("Error running registration: %s", err)
```
hferentschik: don't we want to abort the bootstrap at this point?

hferentschik:
@praveenkumar, could be btw align on "Issue #<key>" in the commit message? The branch name I don't care so much, but we are really referencing an issue key, not a "fix".

pkg/minishift/registration/registration.go
```diff
@@ -0,0 +1,91 @@
...
+}
+
+type StandardRegistrator struct{}
+
+func SetDetector(newDetector Detector) {
+	detector = newDetector
+}
+
+// Registration defines distribution specific actions
+type Registrator interface {
```
praveenkumar: @hferentschik https://github.com/gliderlabs/registrator as per description I think we can use it but l am happy to take @Preeticp or @thatdocslady suggestion here.

pkg/minishift/registration/registration.go
```diff
@@ -0,0 +1,91 @@
...
+	SSHCommander
+	// Get the directory where the settings files for docker are to be found
+	GetRegister() error
+
+	// Return the auth options used to configure remote connection for the daemon.
+	GetUnregister() error
+
+	// Figure out if this is the right registration tool to use based on linux distro
+	CompatibleWithDistribution() bool
```
praveenkumar: @hferentschik this method is just return bool as it implemented like for rhel it will check if `subscription-manager` command available, if someone want to implement for enterprise suse then they can implement this method to check if `smt` command available. For others it just return false and we put a nice message as we discussed `Distro don't support registration` and continue with it.

pkg/minishift/registration/registration.go
praveenkumar: @hferentschik This `struct` type we need to have to get driver in place and used it across registrars.

pkg/minishift/registration/registration.go
```diff
@@ -0,0 +1,91 @@
...
+)
+
+var (
+	ErrDetectionFailed = errors.New("Distro type not recognized to Registration")
+	registrators       = make(map[string]*RegisteredRegistrator)
+	detector  Detector = &StandardRegistrator{}
+)
+
+
+type SSHCommander interface {
```
praveenkumar: @hferentschik here is only interface and actual implementation happen in the different registrator/registrar side.

pkg/minishift/registration/registration.go
praveenkumar: @hferentschik Agree. I will include those steps.

pkg/minishift/registration/registration.go
```diff
@@ -0,0 +1,91 @@
...
+	if err := drivers.WaitForSSH(d); err != nil {
+		return nil, err
+	}
+
+	log.Info("Detecting the Registration Distro...")
+
+	for _, r := range registrators {
+		registration := r.New(d)
+
+		if registration.CompatibleWithDistribution() {
```
praveenkumar: @hferentschik this method is just return bool as it implemented like for rhel it will check if subscription-manager command available, if someone want to implement for enterprise suse then they can implement this method to check if smt command available. For others it just return false and we put a nice message as we discussed Distro don't support registration and continue with it.

cmd/minikube/cmd/start.go
praveenkumar: @hferentschik Ack

LalatenduMohanty: We should replace `Distro` with `Distribution`

pkg/minishift/registration/redhat.go
```diff
@@ -0,0 +1,110 @@
...
+
+func (registrator *RedhatRegistrator) GetRegister() error {
+	if output, err := registrator.SSHCommand("sudo subscription-manager version"); err != nil {
+		return err
+	} else
+	{
+		if strings.Contains(output, "not registered") {
+			username, password := credentials()
+			if _, err := registrator.SSHCommand(fmt.Sprintf(
+				"sudo subscription-manager register --username %s --password %s --auto-attach",
```
LalatenduMohanty: We should proxy options as well. But should be optional to the user.

pkg/minishift/registration/redhat.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+func (registrator *RedhatRegistrator) GetUnregister() error {
+	if output, err := registrator.SSHCommand("sudo subscription-manager version"); err != nil {
+		return err
+	} else
+	{
+		if ! strings.Contains(output, "not registered") {
+			if _, err := registrator.SSHCommand(
+				"sudo subscription-manager unregister"); err != nil {
```
LalatenduMohanty: Same as above, we should use proxy options, so that user can register and un-register with proxy server.

praveenkumar: @LalatenduMohanty yes once we have proper way of handling proxy for different part we will implement those stuff.

pkg/minishift/registration/redhat.go
LalatenduMohanty: @praveenkumar right, just cross check and also found https://github.com/openshift/origin/issues/11323

pkg/minishift/registration/redhat.go
LalatenduMohanty: @praveenkumar agree

pkg/minishift/registration/registration.go
Preeticp: As per the Oxford dictionary https://en.oxforddictionaries.com/definition/registrator Registrator is a valid word. (Again that is British English and I think we use American English for docs). So I suppose we can use it, if we wish, though commonly we do use Registrar and most people would identify with it. 
Also Registrar conjures a person whereas Registrator seems impersonal and machine like.
As per https://en.wiktionary.org/wiki/registrator;  'registrator ‎: A registrar or registrant, person or device which registers data.' I find this part 'device which registers data' especially pertinent to this case.

Preeticp: I thought of "Registration" too but it does fit  in all of the scenarios here, for eg. registrators       = make(map[string]*RegisteredRegistrator)

pkg/minishift/registration/registration.go
hferentschik: but why do we need to define a new one. There is already such a thing, right? I am planning to do  a bit of a deeper review later today. Maybe it becomes apparent to me then.

pkg/minikube/constants/constants.go
```diff
@@ -39,6 +40,9 @@ const MinikubeContext = "minishift"
 // MiniShiftEnvPrefix is the prefix for the environmental variables
 const MiniShiftEnvPrefix = "MINISHIFT"
 
+// Variable for registrator
+var Registrator registration.Registrator
```
praveenkumar: This is something I want to try so it's easy to use this `Registrator` variable to unregister a VM without initiate it again but looks like it not working as expected.

cmd/minikube/cmd/stop.go
```diff
@@ -36,7 +36,12 @@ itself, leaving all files intact. The cluster can be started again with the "sta
 		fmt.Println("Stopping local OpenShift cluster...")
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
-
+		if constants.Registrator.CompatibleWithDistribution(){
```
praveenkumar: Currently this is something which is not working during the stop process and giving a `panic`. I need to find how it is possible to use same registrator object for unregister for stop/delete.

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x64a14]

goroutine 1 [running]:
panic(0xc51ac0, 0xc4200100a0)
	/usr/local/go/src/runtime/panic.go:500 +0x1a1
github.com/minishift/minishift/cmd/minikube/cmd.glob..func13(0x14eb680, 0x1515eb0, 0x0, 0x0)
	/Users/prkumar/work/github/go_practice/src/github.com/minishift/minishift/cmd/minikube/cmd/stop.go:39 +0x334
github.com/minishift/minishift/vendor/github.com/spf13/cobra.(*Command).execute(0x14eb680, 0x1515eb0, 0x0, 0x0, 0x14eb680, 0x1515eb0)
	/Users/prkumar/work/github/go_practice/src/github.com/minishift/minishift/vendor/github.com/spf13/cobra/command.go:603 +0x439
github.com/minishift/minishift/vendor/github.com/spf13/cobra.(*Command).ExecuteC(0x14ea9c0, 0x0, 0x0, 0x14b8c20)
	/Users/prkumar/work/github/go_practice/src/github.com/minishift/minishift/vendor/github.com/spf13/cobra/command.go:689 +0x367
github.com/minishift/minishift/vendor/github.com/spf13/cobra.(*Command).Execute(0x14ea9c0, 0x0, 0x2a)
	/Users/prkumar/work/github/go_practice/src/github.com/minishift/minishift/vendor/github.com/spf13/cobra/command.go:648 +0x2b
github.com/minishift/minishift/cmd/minikube/cmd.Execute()
	/Users/prkumar/work/github/go_practice/src/github.com/minishift/minishift/cmd/minikube/cmd/root.go:101 +0x31
main.main()
	/Users/prkumar/work/github/go_practice/src/github.com/minishift/minishift/cmd/minikube/main.go:26 +0x19
```

pkg/minishift/registration/registration.go
thatdocslady: Wow, does nobody realize how strange something like "RegisteredRegistrator" looks? I thought that registration interface does the registration, and not registers itself?

hferentschik: > the word "Registrator", even though it is grammatically correct, sounds wrong to me as well

+1, When I was looking this up in the dictionary I could find _registrar_, but not _registrator_.

>  I would recommend something more simple like "Registration"

-1 Registration is the what is going to be executed, but not the name of the actor on this. In this case I recommend something more technical like _RegistrationManager_.

pkg/minishift/registration/redhat.go
hferentschik: I suggest to defer the proxy stuff into a separate issue. They will apply at multiple levels. Let's try to get basic registration right first.

pkg/minikube/constants/constants.go
hferentschik: @praveenkumar this cannot work. The _start_ and _delete_ are in different executions. Different processes. And this is not a constant by any mean.

cmd/minikube/cmd/stop.go
hferentschik: The problem with using configuration is that atm I am not 100% yet how this works. This probably needs some side investigations first. Happy to do this, if you want to go down this path. It would also help us in the case where you call _delete_ on a stopped VM. In this case, we would like to know whether this VM is registered. We could get this from the config. If it is, we can print out a warning that you first need to start the VM, before deleting it so that un-registration can happen. From this perspective config setting makes most sense.

To move this along though, I think it is easiest to run the detector at this point in time again.

pkg/minishift/registration/registration.go
Preeticp: Agreed, +1

pkg/minishift/registration/redhat_test.go
```diff
@@ -0,0 +1,101 @@
+/*
```
praveenkumar: Test file to make sure we have some test in place.

pkg/minishift/registration/registration.go
praveenkumar: @hferentschik this is required here to define the interface but I used existing one in the redhat registrator and no duplication there.

pkg/minishift/registration/redhat.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+}
+
+type RedhatRegistrator struct {
+	SSHCommander
+	Driver drivers.Driver
+}
+
+func (registrator *RedhatRegistrator) CompatibleWithDistribution() bool {
+	osReleaseOut, err := registrator.SSHCommand("sudo cat /etc/os-release")
```
praveenkumar: We are taking os-release info here without passing it registrator because anyway it will be on ssh command which need to be executed so better here.

pkg/minishift/registration/redhat_test.go
```diff
@@ -0,0 +1,127 @@
...
+			SSHKeyPath: "",
+		},
+	}
+
+	registrator := NewRedhatRegistrator(d)
+	m := make(map[string]string)
+	m["username"] = "foo"
+	m["password"] = "foo"
+	if err := registrator.Register(m); err != nil {
+		t.Fatal("Distribution should able to register")
```
hferentschik: You really do not assert anything here. There is an essential bit in the mock ssh server missing. The ability to ask the mock server whether it has received a given message. Try changing "s.CommandToOutput["sudo subscription-manager version"]" to whatever. The test will still pass. This type of testing is of the worst type. It looks like something gets tested and in reality we have a self fulfilling test. This is even worse than no test

hferentschik:
@praveenkumar IMO there are still some major issues in this code. Some could maybe put down as weak design, but others need to be fixed. 

I recommend we close this pull request, since with all the various comments, it gets really unwieldy to find the right spots to comment. Let's close this pull request and re-open with some hopefully improved code.

cmd/minikube/cmd/start.go
```diff
@@ -143,6 +147,20 @@ func runStart(cmd *cobra.Command, args []string) {
 		os.Setenv(k, v)
 	}
 
+	registrator, err := registration.DetectRegistrator(host.Driver)
+	if err != nil {
+		fmt.Println("Distribution doesn't support registration")
+	} else {
+		fmt.Printf("Registration with %s...", registrator)
+		m := make(map[string]string)
+		m["username"] = viper.GetString(username)
```
hferentschik: You probably need to check here as well whether the parameters are set or not. If you find a registrar and username/password are not specified, you can abort with an appropriate message

cmd/minikube/cmd/start.go
```diff
@@ -143,6 +147,20 @@ func runStart(cmd *cobra.Command, args []string) {
 		os.Setenv(k, v)
 	}
 
+	registrator, err := registration.DetectRegistrator(host.Driver)
```
hferentschik: if you pull the whole registration code into a dedicated method it becomes testable. This way you can write tests for the various cases of parameters where the credentials are or are not specified. See also _start_test.go_

praveenkumar:
Created #222 for finial review. Closing this now.

Issue #134, opened 2016-11-04
Split ISO creation out into its own repository


hferentschik:
The ISO creation is really a orthogonal problem to the minishift CLI. Moving forward we want to offer multiple pluggable ISO images to work with minishift. Having dedicated repositories for the various ISO will create a clear split. 

Suggested repo name for the default ISO - _minishift-iso_.

budhrg:
> Moving forward we want to offer multiple pluggable ISO images to work with minishift

+1 and name is also fine.

hferentschik:
After discussing a bit with @LalatenduMohanty around names, my vote would be atm for something like _minishift-b2d-iso_. Then we move our current CentOS based ISO to the minishift org as _minishift-centos-iso_ and hopefully some time down the road (after the technical issues on the Atomic side have been solved) we can add a _minishift-atomic-iso_ which will be our ultimate goal.

LalatenduMohanty:
> minishift-b2d-iso

+1. sounds good.

budhrg:
> minishift-b2d-iso

What is 'minishift-b2d-iso'?

LalatenduMohanty:
@budhrg minishift-b2d-iso is for the boot2docker ISO

praveenkumar:
I just created repo `minishift-b2d-iso` and we can start moving deploy part to that repo and add Makefile and README and also remove those bits from minishift once we complete this switch.

LalatenduMohanty:
@praveenkumar I am on it. I was trying to understand how Viper config can be changed to make minishift-centos-iso or minishift-rhel-iso default.

praveenkumar:
@LalatenduMohanty I am not sure why you need that right now to separate the repo, minishift-rhel-iso as default we need downstream and that something we will take care. For upstream once we have that separation we just need to update the ISO url to fetch by default image, IIRC.

- https://github.com/minishift/minishift/blob/7b93c2c80a2421d09cdce17b9c437f01ef42cf52/pkg/minikube/constants/constants.go#L61

LalatenduMohanty:
I think there are two things we need to fix. 

1.  At this point of time the boot2docker ISO used in minishift is strongly tied with the version of minishift [1] [2] , when we break the ISO in to separate repository we need to fix it, so that it can be configured (through Viper) to use a particular release of the ISO or other ISOs. 

[...]

PR #225, opened 2016-12-24
Issue #230 Update Docker version in boot2docker ISO

### LINKED ISSUE #230: Update Docker version in boot2docker ISO
Hi,

Currently, the boot2docker image version used to generate [`boot2docker.iso` is `1.12.3`](https://github.com/minishift/minishift/blob/master/deploy/iso/Dockerfile#L17).

However, Docker version is still lower than it i.e [`1.11.2`](https://github.com/minishift/minishift/blob/master/deploy/iso/Dockerfile#L20).

Upstream [boot2docker repo](https://github.com/boot2docker/boot2docker/releases/tag/v1.12.3) also maintain the same version for both.
One definite reason would be "less confusion".

Unless we have any issues with docker version `1.12.3`, we should update it.

Moreover, I found interesting facts today as:
```
#       Boot2docker version 1.12.3 boot2docker.iso 
Boot2Docker version 1.12.3, build HEAD : 7fc7575 - Thu Oct 27 17:23:17 UTC 2016
Docker version 1.12.3, build 6b644ec

#       Minishift master boot2docker.iso
Boot2Docker version 1.11.2, build HEAD : 7fc7575 - Thu Oct 27 17:23:17 UTC 2016
Docker version 1.11.2, build b9f10c9
```
Clearly, minishift master boot2docker.iso is showing wrong info of Boot2docker version for `build HEAD 7fc7575`.

budhrg:
Fix #230 .

### DIFF
--- a/deploy/iso/Dockerfile
+++ b/deploy/iso/Dockerfile
@@ -17,8 +17,8 @@
 FROM boot2docker/boot2docker:1.12.3
 
 # Set the version of Docker and the expected sha.
-RUN echo "1.11.2" > $ROOTFS/etc/version
-ENV DOCKER_SHA 1bfd065784e0f422c000d86da4feda87cd63ced8
+RUN echo "1.12.3" > $ROOTFS/etc/version
+ENV DOCKER_SHA a3bb88649bcc47775c29dab7d95709f4dda04d0e
 
 # Add other dependencies here to $ROOTFS/
 ADD nsenter $ROOTFS/usr/bin/


deploy/iso/Dockerfile
```diff
@@ -17,7 +17,7 @@
 FROM boot2docker/boot2docker:1.12.3
 
 # Set the version of Docker and the expected sha.
-RUN echo "1.11.2" > $ROOTFS/etc/version
+RUN echo "1.11.3" > $ROOTFS/etc/version
```
LalatenduMohanty: @budhrg shouldn't it be 1.12.3 instead of 1.11.3? Also can you please open an issue for it?

budhrg: Ya @LalatenduMohanty :smile: Updated now

praveenkumar:
LGTM :+1: 

@LalatenduMohanty looks like change which you requested is now part of this patch.

LalatenduMohanty:
@praveenkumar @budhrg Manually updating the same variable in two places is not a good idea IMO. Can we use a variable here , so that when we update the version variable in gets updated in required places?

budhrg:
> @praveenkumar @budhrg Manually updating the same value in two places is not a good idea IMO. Can we use a variable here , so that when we update the version variable it gets updated in required places?

Agreed :+1:

budhrg:
@LalatenduMohanty But it looks like `FROM` must be the [first statement in Dockerfile](https://docs.docker.com/engine/reference/builder/#from). Hence, storing anything before make the Dockerfile invalid.

praveenkumar:
@budhrg @LalatenduMohanty why don't we put something as build args in the build.sh file because there it create the iso, docker support build time args.

build.sh file: https://github.com/minishift/minishift/blob/master/deploy/iso/build.sh

https://docs.docker.com/engine/reference/commandline/build/#/set-build-time-variables---build-arg

praveenkumar:
@budhrg alright so looks like currently it's not possible https://github.com/docker/docker/issues/18119 and https://github.com/faradayio/cage/issues/56 is rfc for same. I can now think of using shell power and do some magic like https://github.com/minishift/minishift-centos-iso/blob/master/Makefile#L36

budhrg:
> I can now think of using shell power and do some magic like https://github.com/minishift/minishift-centos-iso/blob/master/Makefile#L36

That's looks bit complex and I am letting it like that.

jimmidyson:
You could pipe the Dockerfile to stdin of docker build perhaps?

praveenkumar:
@jimmidyson Good idea 👍 , @budhrg try that I think that will work.

hferentschik:
Can we track this stuff in an issue please.

hferentschik:
guys, what makes you believe that this is the same things and that it should be aligned? In the FROM clause we are dealing with the boot2docker/boot2docker image version, in the second case we are dealing with the Docker version to be installed. Why do you think that this should be the same. And even if, looking at the rest of the _Dockerfile_, just updating the docker version without the SHA will most likely fail, right? 

So I am still wondering what this is about. Are we addressing a concrete bug? The pull request title says "Fix boot2docker pinned version" which seems wrong? Is the aim to update the Docker daemon version?

praveenkumar:
@hferentschik FROM section tag represent which boot2docker base image we are going to consume and `$ROOTFS/etc/version` represent which docker version we want in our iso and both are completely different and shouldn't be hard requirement to match as you said. In past openshift had issue like it only work with specific version of docker and this allow us to chose which version we should embedded in our iso. 

@budhrg Please change your commit message and issue tracker details also if that patch will go in, it also need update on SHA and this need test after build that iso before we merge.

LalatenduMohanty:
@hferentschik You are right , I had missed https://github.com/minishift/minishift/blob/master/deploy/iso/Dockerfile#L35-L40 and I was under an impression that docker version ind boot2docker versions stays same as per https://github.com/boot2docker/boot2docker/blob/master/Dockerfile#L288-L292

LalatenduMohanty:
@budhrg I am not sure if we want to keep the docker version same as boot2docker version unless you have a strong reason for this.

budhrg:
Right now I am building iso from Dockerfile with suggested changes to see any regression.

Some changes highlighted here around [docker version 1.12.3](https://github.com/docker/docker/blob/master/CHANGELOG.md#1123-2016-10-26).

budhrg:
Tested the ISO generated with the suggested changes and `minishift start` working fine with proper boot2docker version and docker version.
```
Boot2Docker version 1.12.3, build HEAD : 7fc7575 - Thu Oct 27 17:23:17 UTC 2016
Docker version 1.12.3, build 6b644ec
```

Full logs - https://gist.github.com/budhrg/2858071ec621271b9084622a99a8ad9f

hferentschik:
I am not opposed to upgrading docker provided the ISO build works now and there are not other regressions when it comes to provisioning OpenShift. What I don't understand it why we look into this atm where we for example have also issue #134 which would initially be more valuable. If we had the boot2docker ISO generated extracted, we could add CentOS CI for testing the ISO build and run smoke tests for OpenShift provisioning. This would have caught the issue here from the start.

budhrg:
@hferentschik We can add this change while migrating to new repo. I am ok to wait for that .

hferentschik:
Closing this issue in favor of minishift/minishift-b2d-iso#6. Once minishift-b2d-iso is working we can add a pull request there as a first functional change. Then we can cut a release we can use to test integration with minishift.

Issue #214, opened 2016-12-22
Document xhyve driver known issues


hferentschik:
Let's update the troubleshooting and driver installation guides - https://github.com/minishift/minishift/blob/master/docs/troubleshooting.md resp https://github.com/minishift/minishift/blob/master/docs/docker-machine-drivers.md with some of the known issues around xhyve.

* The requirement to use the latest version
* Verify the brew install of the driver works
* Add manual installation instructions 
* Problem around vmnet creation which manifests itself in something like: "Could not create vmnet interface, permission denied or no entitlement?"

hferentschik:
Merged via pull #226

Issue #203, opened 2016-12-20
Improve installation documentation


LalatenduMohanty:
We should improve the documentation for installation section [1] . Here are some suggestion.

 * We should mention respective docker-machine-driver for respective virtualization driver e.g. KVM section should have link to install instruction for docker-machine-kvm driver
* Exact steps with example, user needs to do for installation of minishift.
* Example of minishift command  with `--vm-driver=xxx`

[1] https://github.com/minishift/minishift#installation
[2] https://github.com/minishift/minishift/blob/master/docs/docker-machine-drivers.md

hferentschik:
Merged via pull request #245

PR #226, opened 2016-12-26
Issue #214 Document xhyve driver known issues

praveenkumar:
This patch will add xhyve driver required version, manual installation steps and troubleshoot steps.

### DIFF
--- a/docs/docker-machine-drivers.md
+++ b/docs/docker-machine-drivers.md
@@ -55,6 +55,18 @@ $ newgrp libvirt
 <a name="xhyve-driver"></a>
 ## xhyve driver
 
+Minishift is currently tested against `docker-machine-driver-xhyve` 0.3.1
+
+```
+$ brew info --installed docker-machine-driver-xhyve
+docker-machine-driver-xhyve: stable 0.3.1 (bottled), HEAD
+Docker Machine driver for xhyve
+https://github.com/zchee/docker-machine-driver-xhyve
+/usr/local/Cellar/docker-machine-driver-xhyve/0.3.1 (3 files, 13.2M) *
+  Poured from bottle on 2016-12-20 at 17:44:35
+From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/docker-machine-driver-xhyve.rb
+```
+
 From https://github.com/zchee/docker-machine-driver-xhyve#install:
 
 ```
@@ -63,4 +75,18 @@ $ brew install docker-machine-driver-xhyve
 # docker-machine-driver-xhyve need root owner and uid
 $ sudo chown root:wheel $(brew --prefix)/opt/docker-machine-driver-xhyve/bin/docker-machine-driver-xhyve
 $ sudo chmod u+s $(brew --prefix)/opt/docker-machine-driver-xhyve/bin/docker-machine-driver-xhyve
+```
+
+Manual Installation
+
+```
+$ go get -u -d github.com/zchee/docker-machine-driver-xhyve
+$ cd $GOPATH/src/github.com/zchee/docker-machine-driver-xhyve
+
+# Install docker-machine-driver-xhyve binary into /usr/local/bin
+$ make install
+
+# docker-machine-driver-xhyve need root owner and uid
+$ sudo chown root:wheel /usr/local/bin/docker-machine-driver-xhyve
+$ sudo chmod u+s /usr/local/bin/docker-machine-driver-xhyve
 ```
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -4,12 +4,13 @@ This section contains solutions to common problems that you might encounter whil
 using Minishift.
 
 - [KVM drivers](#kvm-drivers)
+- [XHYVE drivers](#xhyve-drivers)
 
 ## KVM drivers
 
-- Minishift fails to start because of ```Error creating new host: dial tcp: missing address```
+- Minishift fails to start because of ```Error creating new host: dial tcp: missing address```.
 
-  the problem is likely to be `libvirtd` servce not running, you can check it with
+  The problem is likely to be that the `libvirtd` service is not running, you can check it with
 
   ```
   systemctl status libvirtd
@@ -22,9 +23,9 @@ using Minishift.
   ```
 
 
-- Minishift fails to start because of ```Failed to connect socket to '/var/run/libvirt/virtlogd-sock'```
+- Minishift fails to start because of ```Failed to connect socket to '/var/run/libvirt/virtlogd-sock'```.
 
-  the problem is likely to be `virtlogd` service not running, you can check it with
+  The problem is likely to be that the `virtlogd` service not running, you can check it with
 
   ```
   systemctl status virtlogd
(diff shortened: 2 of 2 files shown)


docs/troubleshooting.md
```diff
@@ -35,3 +36,28 @@ using Minishift.
   systemctl start virtlogd
   systemctl enable virtlogd
   ```
+
+
+## XHYVE drivers
+
+- Could not create vmnet interface, permission denied or no entitlement
+
```
Preeticp: We need some context here, something like, Minishift fails to start because of `the error mssg` will keep it consistent with the above issues, but whatever is the case here needs to be added.

docs/troubleshooting.md
```diff
@@ -35,3 +36,28 @@ using Minishift.
...
+    The problem is xhyve driver is not able to clean up vmnet when remove a VM.
+
```
Preeticp: The problem is likely to be that the xhyve driver is not able to clean up vmnet when a VM is removed.

docs/troubleshooting.md
```diff
@@ -35,3 +36,28 @@ using Minishift.
...
+    Workaround:
+        vmnet.framework seems to have decide the IP based on below files
```
Preeticp: The vmnet.framework decides the IP based on the following files:

docs/docker-machine-drivers.md
```diff
@@ -55,6 +55,18 @@ $ newgrp libvirt
 <a name="xhyve-driver"></a>
 ## xhyve driver
 
+Minishift is currently tested against `docker-machine-driver-xhyve` 0.3.1
```
budhrg: Probably "with version" ahead of `0.3.1` would make it more readable.

`Minishift is currently tested against 'docker-machine-driver-xhyve' with version 0.3.1`.

docs/troubleshooting.md
budhrg: There is no clear separation of where the problem is being stated and where the solution is beginning. This become more in [KVM driver section](https://github.com/praveenkumar/minishift/blob/05b42f4cd8f505dc2e58a0d54564fe86985a576b/docs/troubleshooting.md#kvm-drivers) where the solution start with small caps. 

I suggest to add full stop or `?` to end the problem statement and use Capital letter to begin the solution

docs/troubleshooting.md
```diff
@@ -35,3 +36,28 @@ using Minishift.
...
+
+            /var/db/dhcpd_leases
+            /Library/Preferences/SystemConfiguration/com.apple.vmnet.plist
+
+    So, If you want to reset `minishift` specific IP database, then make sure you remove `minishift` entry section from `dhcpd_leases` file and reboot your system.
+
```
Preeticp: When a user faces the above error he will have to do this step right? So we can drop the 'If' and say: Reset the `minishift` specific IP database and make sure you remove the `minishift` entry section from the `dhcpd_leases` file. Finally, reboot your system.

docs/troubleshooting.md
```diff
@@ -35,3 +36,28 @@ using Minishift.
...
+            {
+                ip_address=192.168.64.2
+                hw_address=1,2:51:8:22:87:a6
+                identifier=1,2:51:8:22:87:a6
+                lease=0x585e6e70
+                name=minishift
+            }
+
+    If you want to complete reset IP database, please remove both files manually but very **risky**.
```
Preeticp: **Warning:** You can do a complete reset of the IP database by removing both the files manually, however this is very risky.

docs/troubleshooting.md
Preeticp: Budhram, the whole document needs to be given a formatting and language cleanup. Currently I have focused on getting the basics right and keeping it consistent.

docs/troubleshooting.md
Preeticp: On line 13,  a minor change, servce>service, though this is not a part of your PR.

docs/docker-machine-drivers.md
Preeticp: The documentation part is being looked into as part of: https://github.com/minishift/minishift/issues/203 where the whole document will be reviewed in detail.

praveenkumar: @budhrg I thinks when you specify `0.3.1` that means referring to version. Also as @Preeticp said there is another issue which will take care of all version info against minishift and more changes in documentation.

docs/troubleshooting.md
```diff
@@ -35,3 +36,28 @@ using Minishift.
...
   systemctl enable virtlogd
   ```
+
+
+## XHYVE drivers
+
+- Minishift failed to start because of ```Error: could not create vmnet interface, permission denied or no entitlement```.
+
+    The problem is likely to be that the xhyve driver is not able to clean up vmnet when a VM is removed.
+
```
Preeticp: not able to clean up vmnet>not able to clean vmnet/not able to clean the `vmnet.framework`
Do we need back ticks for vmnet.framework on line 45 and 48? either ways let us keep it consistent everywhere.

docs/troubleshooting.md
```diff
@@ -35,3 +36,28 @@ using Minishift.
...
+    Workaround:
+        vmnet.framework decides the IP based on following files:
+
+            /var/db/dhcpd_leases
+            /Library/Preferences/SystemConfiguration/com.apple.vmnet.plist
+
+    Reset `minishift` specific IP database and make sure you remove `minishift` entry section from `dhcpd_leases` file. Finally, reboot your system.
+
```
Preeticp: Add 'the' > Reset the `minishift` specific IP database and make sure you remove the `minishift` entry section from the `dhcpd_leases` file.

docs/troubleshooting.md
```diff
@@ -35,3 +36,28 @@ using Minishift.
...
+            {
+                ip_address=192.168.64.2
+                hw_address=1,2:51:8:22:87:a6
+                identifier=1,2:51:8:22:87:a6
+                lease=0x585e6e70
+                name=minishift
+            }
+
+    You can do a complete reset IP database by removing both files manually but this is very **risky**.
```
Preeticp: You can do a complete reset of the IP database by removing both the files

I think it would be better to add a  **Note** or **Warning** before the sentence.

docs/troubleshooting.md
```diff
@@ -4,12 +4,13 @@ This section contains solutions to common problems that you might encounter whil
...
 - [KVM drivers](#kvm-drivers)
+- [XHYVE drivers](#xhyve-drivers)
 
 ## KVM drivers
 
-- Minishift fails to start because of ```Error creating new host: dial tcp: missing address```
+- Minishift fails to start because of ```Error creating new host: dial tcp: missing address```.
 
-  the problem is likely to be `libvirtd` servce not running, you can check it with
+  The problem is likely to be `libvirtd` service is not running, you can check it with
```
Preeticp: likely to be `libvirtd` > likely to be that `libvirtd`

docs/troubleshooting.md
```diff
@@ -35,3 +36,28 @@ using Minishift.
   systemctl start virtlogd
   systemctl enable virtlogd
   ```
+
```
Preeticp: Not a part of this PR, Line 26, full stop after ending of problem statement, and capital T for the problem. 
the problem is likely to be `virtlogd` service not running, you can check it with>  The problem is likely to be that the `virtlogd` service is not running, you can check it with....

docs/troubleshooting.md
praveenkumar: @Preeticp `vmnet.framework` is API which allow virtual machines to read and write packets and `vmnet` is interface which created using this API so here vmnet is right because that interface is not deleted.

docs/docker-machine-drivers.md
budhrg: :+1:

PR #228, opened 2016-12-27
Refactor phony targets

budhrg:
Grouped phony targets. This will reduce lines and easy to maintain IMO.


hferentschik:
-1, there is a reason why the phony definition is next to the target. It shall be immediately obvious that this is a phony target and also reduce the risk that the target and phony definition get out of sync.

hferentschik:
Also, I am missing an issue here.

Issue #223, opened 2016-12-23
Improve code around virtual machine regsitration


hferentschik:
Follow up issue for issue #137 and pull request #222:

* Improve registrar detection algorithm. Don't use error (_ErrDetectionFailed_) for handling normal program flow
* Review _TestRedHatRegistratorRegister_ and _TestRedHatRegistratorUnregister_
   * Why are register resp. unregister called twice?
   * One seems to be still able to change the cmd string (at least the first time around, why?)
* Add test around providing credentials, either via flags or environment variables

praveenkumar:
> Add test around providing credentials, either via flags or environment variables

Credentials are part of flag/env for viper and looks like that is already tested as part of `root_test` imo.

hferentschik:
Resolved via pull request #231

PR #231, opened 2016-12-28
Issue #223 Improve code around VM registration

praveenkumar:
This patch is try to resolve below issues.

>  Improve registrar detection algorithm. Don't use error (ErrDetectionFailed) for handling normal program flow

Resolved, added a bool to check registration functionality instead err.

> Review TestRedHatRegistratorRegister and TestRedHatRegistratorUnregister
        Why are register resp. unregister called twice?
        One seems to be still able to change the cmd string (at least the first time around, why?)

Resolved, fixed a error message around the command. Mock ssh only handle command output doesn't have any provision for command error or returned value.

> Add test around providing credentials, either via flags or environment variables

Tests for flags/env are already there as part of `root_test.go`. Adding `username/password` there doesn't mean any sense because that already tested against a generic way.

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -25,6 +25,8 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/registration"
 	"github.com/spf13/cobra"
+	"github.com/docker/machine/libmachine/drivers"
+	"github.com/docker/machine/libmachine/state"
 )
 
 // deleteCmd represents the delete command
@@ -39,14 +41,16 @@ associated files.`,
 		defer api.Close()
 		host, err := api.Load(constants.MachineName)
 		if err != nil {
-			fmt.Println("Errors occurred deleting machine: ", err)
+			fmt.Println("Error occurred deleting machine: ", err)
 			os.Exit(1)
 		}
 
-		// Unregister Host VM
-		if err := registration.UnregisterHostVM(host, RegistrationParameters); err != nil {
-			fmt.Printf("Error unregistring machine: %s", err)
-			os.Exit(1)
+		if ! drivers.MachineInState(host.Driver, state.Stopped)() {
+			// Unregister Host VM
+			if err := registration.UnregisterHostVM(host, RegistrationParameters); err != nil {
+				fmt.Printf("Error unregistring machine: %s", err)
+				os.Exit(1)
+			}
 		}
 
 		if err := cluster.DeleteHost(api); err != nil {
--- a/cmd/minikube/cmd/stop.go
+++ b/cmd/minikube/cmd/stop.go
@@ -25,6 +25,8 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/registration"
 	"github.com/spf13/cobra"
+	"github.com/docker/machine/libmachine/state"
+	"github.com/docker/machine/libmachine/drivers"
 )
 
 // stopCmd represents the stop command
@@ -40,14 +42,16 @@ itself, leaving all files intact. The cluster can be started again with the "sta
 
 		host, err := api.Load(constants.MachineName)
 		if err != nil {
-			fmt.Println("Errors occurred deleting machine: ", err)
+			fmt.Println("Error occurred stopping machine: ", err)
 			os.Exit(1)
 		}
 
-		// Unregister Host VM
-		if err := registration.UnregisterHostVM(host, RegistrationParameters); err != nil {
-			fmt.Printf("Error unregistring machine: %s", err)
-			os.Exit(1)
+		if ! drivers.MachineInState(host.Driver, state.Stopped)() {
+			// Unregister Host VM
+			if err := registration.UnregisterHostVM(host, RegistrationParameters); err != nil {
+				fmt.Printf("Error unregistring machine: %s", err)
+				os.Exit(1)
+			}
 		}
 
 		if err := cluster.StopHost(api); err != nil {
--- a/pkg/minishift/registration/redhat_test.go
+++ b/pkg/minishift/registration/redhat_test.go
@@ -21,6 +21,7 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/tests"
 	"testing"
 	"github.com/docker/machine/libmachine/provision"
+	"fmt"
 )
 
 var (
@@ -28,9 +29,12 @@ var (
 		Username: "foo",
 		Password: "foo",
 	}
+	expectedCMDRegistration = fmt.Sprintf("sudo subscription-manager register --auto-attach --username %s --password %s ",
+		param.Username, param.Password)
+	expectedCMDUnregistration = "sudo subscription-manager unregister"
 )
 
-func TestRedHatRegistratorCompatibleWithDistribution(t *testing.T) {
+func setup(t *testing.T) (registrator Registrator){
 	s, _ := tests.NewSSHServer()
 	s.CommandToOutput = make(map[string]string)
 	s.CommandToOutput["sudo subscription-manager version"] = `server type: This system is currently not registered.`
@@ -48,7 +52,12 @@ func TestRedHatRegistratorCompatibleWithDistribution(t *testing.T) {
 
 	commander := provision.GenericSSHCommander{Driver: d}
 
-	registrator := NewRedHatRegistrator(commander)
+	registrator = NewRedHatRegistrator(commander)
+	return registrator
+}
+
+func TestRedHatRegistratorCompatibleWithDistribution(t *testing.T) {
+	registrator := setup(t)
 	info := &provision.OsRelease{
 		Name:      "Red Hat Enterprise Linux Server",
 		ID: 	   "rhel",
@@ -56,6 +65,18 @@ func TestRedHatRegistratorCompatibleWithDistribution(t *testing.T) {
 	}
 	if !registrator.CompatibleWithDistribution(info) {
 		t.Fatal("Registration capability should be in the Distribution")
+	}
+}
+
+func TestRedHatRegistratorNotCompatibleWithDistribution(t *testing.T) {
+	registrator := setup(t)
+	info := &provision.OsRelease{
+		Name:      "CentOS",
+		ID: 	   "centos",
+		VersionID: "7.3",
+	}
+	if registrator.CompatibleWithDistribution(info) {
+		t.Fatal("Registration capability shouldn't be in the Distribution")
 	}
 }
 
@@ -75,22 +96,13 @@ func TestRedHatRegistratorRegister(t *testing.T) {
 	}
 	commander := provision.GenericSSHCommander{Driver: d}
 	registrator := NewRedHatRegistrator(commander)
-	s.CommandToOutput["sudo subscription-manager version"] = `server type: RedHat Subscription Management`
-	if err := registrator.Register(param); err != nil {
-		t.Fatal("Distribution shouldn't have to register")
-	} else {
-		cmd := "sudo subscription-manager register --auto-attach --username foo --password foo"
-		if _, ok := s.Commands[cmd]; ok{
-			t.Fatalf("Expected command: %s", cmd)
-		}
-	}
+
 	s.CommandToOutput["sudo subscription-manager version"] = `server type: This system is currently not registered.`
 	if err := registrator.Register(param); err != nil {
 		t.Fatal("Distribution should able to register")
 	} else {
-		cmd := "sudo subscription-manager register --auto-attach --username foo --password foo "
-		if _, ok := s.Commands[cmd]; !ok{
-			t.Fatalf("Expected command: %s", cmd)
+		if _, ok := s.Commands[expectedCMDRegistration]; !ok{
+			t.Fatalf("Expected command: %s", expectedCMDRegistration)
 		}
 	}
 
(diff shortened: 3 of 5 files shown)


cmd/minikube/cmd/delete.go
```diff
@@ -39,14 +41,16 @@ associated files.`,
...
-			fmt.Println("Errors occurred deleting machine: ", err)
+			fmt.Println("Error occurred deleting machine: ", err)
 			os.Exit(1)
 		}
 
-		// Unregister Host VM
-		if err := registration.UnregisterHostVM(host, RegistrationParameters); err != nil {
-			fmt.Printf("Error unregistring machine: %s", err)
-			os.Exit(1)
+		if ! drivers.MachineInState(host.Driver, state.Stopped)() {
```
hferentschik: +1 Very good

cmd/minikube/cmd/stop.go
```diff
@@ -40,14 +42,16 @@ itself, leaving all files intact. The cluster can be started again with the "sta
...
-			fmt.Println("Errors occurred deleting machine: ", err)
+			fmt.Println("Error occurred stopping machine: ", err)
 			os.Exit(1)
 		}
 
-		// Unregister Host VM
-		if err := registration.UnregisterHostVM(host, RegistrationParameters); err != nil {
-			fmt.Printf("Error unregistring machine: %s", err)
-			os.Exit(1)
+		if ! drivers.MachineInState(host.Driver, state.Stopped)() {
```
hferentschik: +1

pkg/minishift/registration/registration.go
```diff
@@ -24,13 +24,13 @@ import (
 )
 
 var (
-	ErrDetectionFailed          = errors.New("Distro type not recognized to Registration")
+	ErrDetectionFailed          = errors.New("Distribution type not recognized to Registration")
```
hferentschik: +1

pkg/minishift/registration/registration.go
```diff
@@ -62,18 +62,18 @@ func Register(name string, r *RegisteredRegistrator) {
 	registrators[name] = r
 }
 
-func DetectRegistrator(c provision.SSHCommander) (Registrator, error) {
+func DetectRegistrator(c provision.SSHCommander) (Registrator, bool, error) {
```
hferentschik: +1, that works, adding an additional boolean return value. I think this is quite idiomatic in Go.

pkg/minishift/registration/registration.go
```diff
@@ -62,18 +62,18 @@ func Register(name string, r *RegisteredRegistrator) {
...
-		return nil, err
+		return nil, false, err
 	}
 	osReleaseInfo, err := provision.NewOsRelease([]byte(osReleaseOut))
 	if err != nil {
-		return nil, fmt.Errorf("Error parsing /etc/os-release file: %s", err)
+		return nil, false, fmt.Errorf("Error parsing /etc/os-release file: %s", err)
 	}
 
 	log.Info("Detecting Host VM is managed by registration manager...")
```
hferentschik: I don't think we need this. I think this might be a bit confusing for most of the community folks where registration is nothing they know about. I suggest to change this to some sort of debug trace. Also it should be something like:"Detecting whether virtual machine is managed by a registration manager..."

pkg/minishift/registration/utils.go
```diff
@@ -35,21 +34,24 @@ type RegistrationParameters struct {
 // Register host VM
 func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
 	commander := provision.GenericSSHCommander{Driver: host.Driver}
-	registrator, err := DetectRegistrator(commander)
-	if err == ErrDetectionFailed {
+	registrator, supportRegistration, err := DetectRegistrator(commander)
+	if ! supportRegistration {
 		fmt.Println("Distribution doesn't support registration")
```
hferentschik: I think this should be debug at best. In the default case with the b2d or CentOS image, there won't be registration, still the user will see this everytime he stops the machine. That might throw him off.

pkg/minishift/registration/utils.go
```diff
@@ -35,21 +34,24 @@ type RegistrationParameters struct {
...
+
+	if err != nil && err != ErrDetectionFailed {
+		return err
 	}
 
 	if registrator != nil {
 		if param.Username == "" || param.Password == "" {
-			return errors.New(fmt.Sprint("This virutal machine requires registration. Credentials must either be passed via the environment variables MINISHIFT_USERNAME and MINISHIFT_PASSWORD " +
-				" or the --username and --password flags\n"))
+			return errors.New("This virutal machine requires registration. " +
```
hferentschik: nice

pkg/minishift/registration/utils.go
```diff
@@ -35,21 +34,24 @@ type RegistrationParameters struct {
 // Register host VM
 func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
 	commander := provision.GenericSSHCommander{Driver: host.Driver}
-	registrator, err := DetectRegistrator(commander)
-	if err == ErrDetectionFailed {
+	registrator, supportRegistration, err := DetectRegistrator(commander)
+	if ! supportRegistration {
```
hferentschik: Same as comment below. I don't think we should log this on info level. If anything debug level.

pkg/minishift/registration/redhat_test.go
```diff
@@ -75,15 +75,7 @@ func TestRedHatRegistratorRegister(t *testing.T) {
...
-	s.CommandToOutput["sudo subscription-manager version"] = `server type: RedHat Subscription Management`
-	if err := registrator.Register(param); err != nil {
-		t.Fatal("Distribution shouldn't have to register")
-	} else {
-		cmd := "sudo subscription-manager register --auto-attach --username foo --password foo"
-		if _, ok := s.Commands[cmd]; ok{
-			t.Fatalf("Expected command: %s", cmd)
-		}
-	}
+
```
hferentschik: Actually, this is a comment for line 51-60. I like the test. I think in these cases, it is always nice to show both sides of the test. So in case we dont have 'rhel' the distribution is indeed not compatible. Especially with boolean conditions you most of the times want to check both sides.

pkg/minishift/registration/redhat_test.go
```diff
@@ -112,22 +104,14 @@ func TestRedHatRegistratorUnregister(t *testing.T) {
...
-	} else {
-		cmd := "sudo subscription-manager unregister"
-		if _, ok := s.Commands[cmd]; ok{
-			t.Fatalf("Expected command: %s", cmd)
-		}
-	}
+
 	s.CommandToOutput["sudo subscription-manager version"] = `server type: RedHat Subscription Management`
 	if err := registrator.Unregister(param); err != nil {
 		t.Fatal("Distribution should able to unregister")
```
hferentschik: "...should be able to..."

pkg/minishift/registration/redhat_test.go
```diff
@@ -112,22 +104,14 @@ func TestRedHatRegistratorUnregister(t *testing.T) {
...
 	} else {
 		cmd := "sudo subscription-manager unregister"
 		if _, ok := s.Commands[cmd]; !ok{
-			t.Fatalf("Expected command: %s", cmd)
+			t.Fatalf("Expected command: sudo subscription-manager unregister")
```
hferentschik: I think what you wanted to do is: "t.Fatalf("Expected command: %s", cmd)"

praveenkumar: @hferentschik no, so I want to check here what command suppose to get executed so if we change `cmd` variable it will let us show what actual command it should have.

hferentschik: ??? 

the content of _cmd_ is your expected command! This is the key you go into the commands map. If you want to show what actually got received you need to change the test. The expected command as per your test is whatever you put into 'cmd'.

hferentschik: > if we change cmd variable it will let us show what actual command it should have.

how does that make sense? When would you update _cmd_ without updating the expected message in the error. What is indeed missing and would be interesting would be the list of actual commands executed.

hferentschik:
> Tests for flags/env are already there as part of root_test.go.

These are different tests. They server a different purpose.

> Adding username/password there doesn't mean any sense because that already tested against a generic way.

That's not true. None of these tests verifies in any way that registration would work via command line falgs, environment variables or config settings. The tests you are referring to just test a very small subset of the globally available flags. None of the tests proofs the existence of the registration falgs and none of them show/proofs that the flags/values get properly picked up in the appropriate sub-commands (start/stop/delete).

I am happy to merge this and move on, but I disagree that this is tested on the flags/env level. I guess we let this go for now and take testing in this area up at a later point in time.

pkg/minishift/registration/redhat_test.go
```diff
@@ -112,22 +121,14 @@ func TestRedHatRegistratorUnregister(t *testing.T) {
...
+
 	s.CommandToOutput["sudo subscription-manager version"] = `server type: RedHat Subscription Management`
 	if err := registrator.Unregister(param); err != nil {
-		t.Fatal("Distribution should able to unregister")
+		t.Fatal("Distribution should be able to unregister")
 	} else {
 		cmd := "sudo subscription-manager unregister"
 		if _, ok := s.Commands[cmd]; !ok{
-			t.Fatalf("Expected command: %s", cmd)
+			t.Fatal("Expected command: sudo subscription-manager unregister")
```
hferentschik: Your expected command is whatever is in _cmd_. This is how your test is setup. I have not idea why you type duplicate the string in the t.Fatal call.

praveenkumar: @hferentschik s.Commands contain all commands executed on the mock ssh server, during unregister what happen is it execute `sudo subscription-manager unregister` and Commands map have that entry. Now if we put `cmd` variable then error message would be misleading. I got what is confusion here so I did some more modification.

hferentschik: > s.Commands contain all commands executed on the mock ssh server, 

Exactly, this is why your test is actually quite weak. Really a better test would assert the total number of executed commands (which we already could) and the order (which afaik we cannot do). Hence, I said once, that we should extend the mock ssh server. Different issue though. Happy to accept the "simpler" form of the test"

>  Now if we put cmd variable then error message would be misleading

No, how can it be? You use the value of _cmd_ as key into the commands map. If the command is there, you say the test passed, if not you say the test fails. Hence the expected commands for your test is whatever the value is in _cmd_.

> I got what is confusion here so I did some more modification.

You are making it worse.

pkg/minishift/registration/redhat_test.go
```diff
@@ -75,22 +94,14 @@ func TestRedHatRegistratorRegister(t *testing.T) {
...
-		if _, ok := s.Commands[cmd]; ok{
-			t.Fatalf("Expected command: %s", cmd)
-		}
-	}
+
 	s.CommandToOutput["sudo subscription-manager version"] = `server type: This system is currently not registered.`
 	if err := registrator.Register(param); err != nil {
 		t.Fatal("Distribution should able to register")
 	} else {
 		cmd := "sudo subscription-manager register --auto-attach --username foo --password foo "
```
hferentschik: Ok, let's try this a different way. Say your production code changes and the registration command indeed needs changing. I think we can agree that in this case the test will fail. What will you have to update to make the test pass again? The value of _cmd_, right? You can change _expectedCMDRegistration_ as much as you like, but to get the test pass again, you need to change _cmd_ to match the **expectation** of _Register_ in _redhat.go_.

praveenkumar: @hferentschik I think I will remove `cmd` altogether and use `expectedCMDRegistration` and `expectedCMDUnregister` variable for test cases.

```
if _, ok := s.Commands[expectedCMDRegistration]; !ok{
            t.Fatalf("Expected command: %s", expectedCMDRegistration)
}
```

pkg/minishift/registration/redhat_test.go
```diff
@@ -75,22 +96,13 @@ func TestRedHatRegistratorRegister(t *testing.T) {
...
+
 	s.CommandToOutput["sudo subscription-manager version"] = `server type: This system is currently not registered.`
 	if err := registrator.Register(param); err != nil {
 		t.Fatal("Distribution should able to register")
 	} else {
-		cmd := "sudo subscription-manager register --auto-attach --username foo --password foo "
-		if _, ok := s.Commands[cmd]; !ok{
-			t.Fatalf("Expected command: %s", cmd)
+		if _, ok := s.Commands[expectedCMDRegistration]; !ok{
+			t.Fatalf("Expected command: %s", expectedCMDRegistration)
```
hferentschik: Yes :-) 

I hope this makes sense to you as well and you understand where I was coming from

pkg/minishift/registration/redhat_test.go
```diff
@@ -21,16 +21,20 @@ import (
...
 		Username: "foo",
 		Password: "foo",
 	}
+	expectedCMDRegistration = fmt.Sprintf("sudo subscription-manager register --auto-attach --username %s --password %s ",
+		param.Username, param.Password)
+	expectedCMDUnregistration = "sudo subscription-manager unregister"
 )
 
-func TestRedHatRegistratorCompatibleWithDistribution(t *testing.T) {
+func setup(t *testing.T) (registrator Registrator){
```
hferentschik: +1

pkg/minishift/registration/redhat_test.go
```diff
@@ -59,6 +68,18 @@ func TestRedHatRegistratorCompatibleWithDistribution(t *testing.T) {
 	}
 }
 
+func TestRedHatRegistratorNotCompatibleWithDistribution(t *testing.T) {
```
hferentschik: nice with testing both sides of the equation now

hferentschik:
Merged, thanks.

Issue #233, opened 2016-12-28
Add overlay as default storage driver for minishift provisioners


praveenkumar:
Currently we are using `devicemapper` as default storage driver for our minishift provisioners which should be changed to `overlay` as it is now available to linux kernel by default and have better performance then devicemapper.

Refer: https://github.com/minishift/minishift-centos-iso/issues/73

hferentschik:
> Currently we are using devicemapper as default storage driver for our minishift provisioners which should be changed to overlay 

why? I am not generally opposed to switch to another storage driver, but it needs to be backed up by some proper reasons.

> as it is now available to linux kernel by default 

hmm, devicemapper is a kernel tech as well right?

> and have better performance then devicemapper.

Is this relevant to us? Do have any performance data to back this up? Looking at 

* https://docs.docker.com/engine/userguide/storagedriver/selectadriver/
* https://docs.docker.com/engine/userguide/storagedriver/device-mapper-driver/
* https://success.docker.com/Policies/Compatibility_Matrix

and in particular the compatibility matrix, devicemapper still seems to be the recommended drvier for CentOS/RHEL.

praveenkumar:
> why? I am not generally opposed to switch to another storage driver, but it needs to be backed up by some proper reasons.

https://batmat.net/2015/08/26/docker-storage-driver-dont-use-devicemapper/ end of this post have lot of reference about not using devicemapper in loopback mode.

> devicemapper still seems to be the recommended driver for CentOS/RHEL.

I agree but we never tell our user to use devicemapper as loopback mode (which we have currently for iso) the way we recommend our users to create a lvm first and then use docker-stroage-setup for which use devicemapper but not as loopback. http://www.projectatomic.io/blog/2015/06/notes-on-fedora-centos-and-docker-storage-drivers/ have a good details around it. 

We are not able to use devicemapper currently as we used for our CDK-2.x/ADB because live image don't support partition for specific use case like https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/build_tools/kickstarts/centos-7-adb-vagrant.ks#L21-L25

hferentschik:
@praveenkumar, thanks, that was the information I was looking for.

Looking at http://www.projectatomic.io/blog/2015/06/notes-on-fedora-centos-and-docker-storage-drivers/, I am wondering about the following:

* "so we have to disable SELinux support within our containers" - does that mean containers with SELinux won't work
[...]

PR #234, opened 2016-12-28
Issue #233 Add overlay as default storage driver for minishift provisioners

praveenkumar:
This patch will add `overlay` as default storage driver for minishift custom provisioners.


hferentschik:
@praveenkumar How does this relate to https://github.com/minishift/minishift-centos-iso/pull/74? Are we really in a position yet to switch or do we need more investigations? Seems overlay driver has also some drawbacks. I also don't fully understand yet why we cannot use a "proper" devicemapper setup.

hferentschik:
I am going to close this for now. I think we need more investigation here and there might be an alternative of using devicemapper + lvm. See discussion on issue #233

PR #239, opened 2016-12-30
Issue #238 Issue template to guide users

### LINKED ISSUE #238: Issue template to guide users to quickly file issue/feature request
Hi,
 
Issue template will help users to have the guidelines in place to file issue or feature request with proper details about the issue like affecting OS / version / how to repro / what is expected etc.

We can follow similar to [minikube way](https://github.com/kubernetes/minikube/issues/new).

budhrg:
Fix #238

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -1,0 +1,29 @@
+<!--
+Thanks for showing interest in filing an issue!
+However, make sure that we do not have any duplicates already open. You can ensure
+this by searching the issue list here https://github.com/minishift/minishift/issues.
+-->
+
+**Minishift version** (use `minishift version`):
+
+
+**Environment**:
+- **OS** (e.g. from /etc/os-release):
+- **VM Driver** :
+- **Others**:
+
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE
```diff
@@ -0,0 +1,41 @@
...
+**What happened**:
+
+
+**What you expected to happen**:
+
+
+**How to reproduce it** (as minimally and precisely as possible):
+
+
+**Anything else do we need to know**:
```
praveenkumar: Can we also add logs information with `--show-libmachine-logs --v=5` so that when someone open a issue can provide those info also.

hferentschik:
-1, imo this gets more in the way than anything else.

.github/ISSUE_TEMPLATE
hferentschik: -1, in my experience you quite often need more time to clean up a issue created via the template, than it takes to ask information which is appropriate in the context. And it absolutely gets in the way when creating non bug reports.

PR #248, opened 2017-01-04
Issue #190 edited error and log messages

thatdocslady:
This PR is a subset of the original big PR for issue #190 

Originally this failed on unit tests, but after fixing build errors I managed to debug and then ran make test locally successfully. Hopefully the Travis tests will reflect this :)

### DIFF
--- a/deploy/minikube/schema_check.go
+++ b/deploy/minikube/schema_check.go
@@ -47,7 +47,7 @@ func validateSchema(schemaPathString string, docPathString string) {
 	if result.Valid() {
 		fmt.Printf("The document %s is valid\n", docPathString)
 	} else {
-		fmt.Printf("The document %s is not valid. see errors :\n", docPathString)
+		fmt.Printf("The document %s is not valid. Error :\n", docPathString)
 		for _, desc := range result.Errors() {
 			fmt.Printf("- %s\n", desc)
 		}
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -69,7 +69,7 @@ func init() {
 func StartHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 	exists, err := api.Exists(constants.MachineName)
 	if err != nil {
-		return nil, fmt.Errorf("Error checking if host exists: %s", err)
+		return nil, fmt.Errorf("Error checking if the host exists: %s", err)
 	}
 	if !exists {
 		return createHost(api, config)
@@ -79,13 +79,13 @@ func StartHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 	h, err := api.Load(constants.MachineName)
 	if err != nil {
 		return nil, fmt.Errorf(
-			"Error loading existing host: %s. Please try running [minikube delete], then run [minikube start] again.", err)
+			"Error loading existing host: %s. Try running `minishift delete` and then run `minishift start` again.", err)
 	}
 
 	s, err := h.Driver.GetState()
 	glog.Infoln("Machine state: ", s)
 	if err != nil {
-		return nil, fmt.Errorf("Error getting state for host: %s", err)
+		return nil, fmt.Errorf("Error getting the state for host: %s", err)
 	}
 
 	if s != state.Running {
@@ -98,7 +98,7 @@ func StartHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 	}
 
 	if err := h.ConfigureAuth(); err != nil {
-		return nil, fmt.Errorf("Error configuring auth on host: %s", err)
+		return nil, fmt.Errorf("Error configuring authorization on host: %s", err)
 	}
 	return h, nil
 }
@@ -198,7 +198,7 @@ func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 	defer response.Body.Close()
 
 	if response.StatusCode != http.StatusOK {
-		return fmt.Errorf("Received %d response from %s while trying to download minikube.iso", response.StatusCode, m.MinikubeISO)
+		return fmt.Errorf("Received %d response from %s while trying to download the ISO", response.StatusCode, m.MinikubeISO)
 	}
 
 	iso := response.Body
@@ -305,7 +305,7 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 	if err := api.Create(h); err != nil {
 		// Wait for all the logs to reach the client
 		time.Sleep(2 * time.Second)
-		return nil, fmt.Errorf("Error creating. %s", err)
+		return nil, fmt.Errorf("Error creating the VM. %s", err)
 	}
 
 	if err := api.Save(h); err != nil {
@@ -356,7 +356,7 @@ func checkIfApiExistsAndLoad(api libmachine.API) (*host.Host, error) {
 		return nil, err
 	}
 	if !exists {
-		return nil, fmt.Errorf("Machine does not exist for api.Exists(%s)", constants.MachineName)
+		return nil, fmt.Errorf("Machine does not exist for API. Exists(%s)", constants.MachineName)
 	}
 
 	host, err := api.Load(constants.MachineName)
@@ -473,7 +473,7 @@ type MissingNodePortError struct {
 }
 
 func (e MissingNodePortError) Error() string {
-	return fmt.Sprintf("Service %s/%s does not have a node port. To have one assigned automatically, the service type must be NodePort or LoadBalancer, but this service is of type %s.", e.service.Namespace, e.service.Name, e.service.Spec.Type)
+	return fmt.Sprintf("Service %s/%s does not have a node port. To assign a port automatically, the service type must be NodePort or LoadBalancer, but this service is of type %s.", e.service.Namespace, e.service.Name, e.service.Spec.Type)
 }
 
 func getServiceFromServiceGetter(services serviceGetter, service string) (*kubeapi.Service, error) {
--- a/pkg/minikube/openshiftversions/openshift_versions_test.go
+++ b/pkg/minikube/openshiftversions/openshift_versions_test.go
@@ -66,7 +66,7 @@ func TestGetVersionsCorrect(t *testing.T) {
 	}
 	if len(releases) != 2 { // TODO(aprindle) change to len(handler....)
 		//Check values here as well?  Write eq method?
-		t.Fatalf("Expected two OpenShift releases, it was instead %s", len(releases))
+		t.Fatalf("Expected two OpenShift releases, received instead %s", len(releases))
 	}
 }
 
@@ -88,7 +88,7 @@ func TestGetVersionsNone(t *testing.T) {
 
 	_, err := getVersions()
 	if err == nil {
-		t.Fatal("No kubernetes versions were returned from URL but no error was thrown")
+		t.Fatal("No OpenShift versions returned from URL but no error reported.")
 	}
 }
 
@@ -112,7 +112,7 @@ func TestGetVersionsMalformed(t *testing.T) {
 
 	_, err := getVersions()
 	if err == nil {
-		t.Fatal("Malformed version value was returned from URL but no error was thrown")
+		t.Fatal("Malformed version value returned from URL but no error reported.")
 	}
 }
 
@@ -131,7 +131,7 @@ func TestPrintOpenShiftVersions(t *testing.T) {
 	var outputBuffer bytes.Buffer
 	PrintOpenShiftVersions(&outputBuffer)
 	if len(outputBuffer.String()) != 0 {
-		t.Fatalf("Expected PrintOpenShiftVersions to not output text as there are no versioned served at the current URL but output was [%s]", outputBuffer.String())
+		t.Fatalf("Expected no output from PrintOpenShiftVersions because no versions exist at the current URL but the output was [%s]", outputBuffer.String())
 	}
 
 	// test that update text is printed if the latest version is greater than the current version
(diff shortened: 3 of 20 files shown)


thatdocslady:
@hferentschik I think Travis hates me :( what is this error for data rate limit from OpenShift? it seems to be failing one of the tests.

deploy/minikube/schema_check.go
```diff
@@ -45,9 +45,9 @@ func validateSchema(schemaPathString string, docPathString string) {
 	}
 
 	if result.Valid() {
-		fmt.Printf("The document %s is valid\n", docPathString)
+		fmt.Printf("The document %s is valid \n", docPathString)
```
praveenkumar: -1  In term of programing this was expected. there is no need any space between word and `\n`

deploy/minikube/schema_check.go
```diff
@@ -45,9 +45,9 @@ func validateSchema(schemaPathString string, docPathString string) {
...
 	} else {
-		fmt.Printf("The document %s is not valid. see errors :\n", docPathString)
+		fmt.Printf("The document %s is not valid. Error: \n", docPathString)
```
praveenkumar: Wording is expected but don't make sure you put`\n` as soon as wording finish without any space.

pkg/minikube/cluster/cluster.go
```diff
@@ -305,7 +305,7 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 	if err := api.Create(h); err != nil {
 		// Wait for all the logs to reach the client
 		time.Sleep(2 * time.Second)
-		return nil, fmt.Errorf("Error creating. %s", err)
+		return nil, fmt.Errorf("Error creating store. %s", err)
```
praveenkumar: What do you mean store here?

deploy/minikube/schema_check.go
hferentschik: +1

pkg/minikube/cluster/cluster.go
hferentschik: +1

pkg/util/github/github.go
```diff
@@ -159,17 +159,17 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
...
 	if err != nil {
-		return errors.Wrapf(err, "Unable to create file %s", assetTmpFile)
+		return errors.Wrapf(err, "Cannot create file %s", assetTmpFile)
 	}
 
 	// Copy the asset and verify its hash
 	_, err = io.Copy(out, asset)
 	if err != nil {
-		return errors.Wrapf(err, "Failed to copy %s to %s", assetTmpFile, tmpDir)
+		return errors.Wrapf(err, "Cannot copy %s to %s", assetTmpFile, tmpDir)
```
hferentschik: For me there is actually a difference between "cannot" and "failed". This is one of the changes which imo go too far. Failing means I tried to do something and an actual unexpected error occurred. This is different from I cannot do something, where my regular program flow checks something and decides it cannot do something.

pkg/util/github/github.go
```diff
@@ -193,18 +193,18 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
 		tarFile := assetTmpFile[:len(assetTmpFile)-3]
 		err = archive.Ungzip(assetTmpFile, tarFile)
 		if err != nil {
-			return errors.Wrapf(err, "Unable to ungzip %s", assetTmpFile)
+			return errors.Wrapf(err, "Cannot run ungzip %s", assetTmpFile)
 		}
 		err = archive.Untar(tarFile, tmpDir)
 		if err != nil {
-			return errors.Wrapf(err, "Unable to untar %s", tarFile)
+			return errors.Wrapf(err, "Cannot run untar %s", tarFile)
```
hferentschik: I am not "running" 'tar' the binary. I "untar". I am running the inverse tart algorithm.

pkg/util/utils_test.go
```diff
@@ -38,24 +38,24 @@ func TestErrorGenerator(t *testing.T) {
...
+		t.Fatalf("Error should not have been reported by this call.")
 	}
 }
 
 func TestRetry(t *testing.T) {
 
 	f := errorGenerator(4)
 	if err := Retry(5, f); err != nil {
-		t.Fatalf("Error should not have been raised by retry.")
+		t.Fatalf("Error should not have been reported during retry.")
```
hferentschik: What's wrong with "raised" and "thrown"? Definitely in the context of a test.

pkg/minikube/openshiftversions/openshift_versions_test.go
```diff
@@ -112,7 +112,7 @@ func TestGetVersionsMalformed(t *testing.T) {
 
 	_, err := getVersions()
 	if err == nil {
-		t.Fatal("Malformed version value was returned from URL but no error was thrown")
+		t.Fatal("Incorrect version value returned from URL but no error reported.")
```
hferentschik: this is based on "malformed URL" which is the right expression

hferentschik:
TBH, I think we are overshooting a bit here. Issue #190 was supposed to verify the help context as part of the Cobra help system. At least this is how I understood it. This changes go beyond this and even though overall they might be good, there seem to be several changes which go too far and actually are against the developers original intend. Not quite sure what do say about this.

pkg/minikube/cluster/cluster.go
```diff
@@ -198,7 +198,7 @@ func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 	defer response.Body.Close()
 
 	if response.StatusCode != http.StatusOK {
-		return fmt.Errorf("Received %d response from %s while trying to download minikube.iso", response.StatusCode, m.MinikubeISO)
+		return fmt.Errorf("Received %d response from %s while trying to download minishift.iso", response.StatusCode, m.MinikubeISO)
```
hferentschik: One might think we are downloading minishift.iso, but it is still wrong. It's neither in this case.

pkg/minishift/provisioner/minishift_detector_test.go
```diff
@@ -71,7 +71,7 @@ VARIANT_VERSION="1.0.0-alpha.1"
 
 	osRelease, _ := provisioner.GetOsReleaseInfo()
 	if osRelease.Variant != "minishift" {
-		t.Fatal("Release info should contain 'minishift' variant")
+		t.Fatal("Release info must contain 'minishift' variant")
```
hferentschik: Actually in the context of this test, "should" is better. I setup a test harness and had an expectation that something happened. Then I executed the test and my expectation was not met. In this case I think it makes sense to say: "It should have contained". It should be there, but for some reason I don't know it was not.

hferentschik:
I feel just throwing away all changes does not make sense either. Let's try to find some minimal common ground and get on with this.

deploy/minikube/schema_check.go
budhrg: Ya, the text is suppose to be printed on stdout and adding space will not matter here.

pkg/util/github/github.go
budhrg: +1

pkg/minikube/cluster/cluster.go
thatdocslady: This error text is right before another error text that talks about saving a store, I assumed that they are related.
If I'm wrong please let me know but we need to qualify the action, what are we creating here? Before the error was vague.

pkg/minikube/openshiftversions/openshift_versions_test.go
thatdocslady: Well I don't think that "malformed URL" is a good phrase, and when I look it up I see tht it means basically the same thing. I'll choose my battles though, if you insist on this I'll go with what you say.

pkg/util/github/github.go
thatdocslady: On a grammatical level the two phrases are equivalent, but in this case what you're saying means that we're actually saying the wrong thing. We should say something like "Unexpected error occured while coping x to y", to differentiate between that and expected errors.

pkg/util/github/github.go
thatdocslady: ugh, I was hoping to avoid a "verbize-noun" (i.e. unzip is not a real verb either). I was thinking if it's a command called "untar" then we would run it. I can't find a lot of usage on this "untar", but if you say it's the accepted verb for it then I guess there is no choice.

pkg/util/utils_test.go
thatdocslady: Nothing is technically "wrong" but it's a matter of naming conventions and RH style guide to use as clear terms as possible, where raised can be ambiguous and thrown is aggressive.

pkg/minikube/cluster/cluster.go
thatdocslady: So what is it? Orginally this said minikube, so I changed it to minishift. We should specify what it is that we're trying to download.

pkg/minishift/provisioner/minishift_detector_test.go
thatdocslady: generally we try to avoid using the word "should", from the understanding that something is either in or out. Should is ambiguious, evasive, and human-associated, plus it doesn't translate well. In this case it's quite simple, something was supposed to be there but it isn't. If the error message appears when you miss a required piece of data, then that piece of data must be there.

thatdocslady:
@hferentschik I don't think we should be throwing all of the changes away just because there are a few places that require discussion. You've already done this one-sided with my iso readme fixed in #247, and I know you have a lot on your plate but this is exactly what I'm here for, my job is to contribute my skills to help improve the UX of the project, is to ensure that the written content is clear and useful. If I contributed something incorrect this is why we have feedback cycles, but I wouldn't throw away the baby with the bathwater here. Maybe we can set an additional reviewer for doc/message issues who can provide focused feedback if you are unavailable?

pkg/util/github/github.go
budhrg: I am fine with `Unexpected error occured while coping x to y`.

pkg/minishift/provisioner/minishift_detector_test.go
budhrg: > In this case it's quite simple, something was supposed to be there but it isn't. If the error message appears when you miss a required piece of data, then that piece of data must be there.

+1

pkg/minikube/cluster/cluster.go
hferentschik: The best I can find is  _h.HostOptions.AuthOptions.StorePath_, which is not really about creating a _store_. If anything the VM is created here.

pkg/minikube/openshiftversions/openshift_versions_test.go
hferentschik: Well, it comes down to semantics again. Malformed URL implies that the URL was wrong and hence something happened which prevented me form retrieving the version information. Your message is more to interpret that I hit an URL which is part of an API and it returned me something, but it was not what I expected. 

So I agree, I also don't want to be pedantic, but there is more to things than just making something sounds good. Given this is a test and the message never being intended to be passed to a user, I'd argue that changing this is not a good idea. I can live with both versions.

pkg/util/github/github.go
hferentschik: > On a grammatical level the two phrases are equivalent,

How can two sentences be equivalent on a grammatical level? They might be grammatically correct in both cases, but imo they differ semantically. 

> "Unexpected error occured while coping x to y",

Sure, something like this works

pkg/util/utils_test.go
hferentschik: Again, this is a test. This is not a user facing message, never. Maybe we should just not touch test files (*_test.go)?

pkg/minikube/cluster/cluster.go
hferentschik: This is in flux and in fact dynamic. The default will be _minishift-b2d.iso_, but it is configurable. So there is more needed than just a typo fix.

pkg/minishift/provisioner/minishift_detector_test.go
hferentschik: Again this is a test.

hferentschik:
> I don't think we should be throwing all of the changes away just because there are a few places that require discussion. 

I agree. But maybe we start with excluding changes to test files. I really don't get what is gained by changing them.

> You've already done this one-sided with my iso readme fixed in #247,

Because this change was obsolete. The whole directory will be removed in one of the upcoming pull requests. Also the whole process described in this document will be updated. 

> and I know you have a lot on your plate but this is exactly what I'm here for, my job is to contribute my skills to help improve the UX of the project, is to ensure that the written content is clear and useful. 

Right, the focus should be on the UX. That's not really the case for tests imo.

thatdocslady:
Updated the PR with feedback. Regarding the _test files, I think that if these strings form some log messages and our users are basically developers, then we cannot assume they will never look at these. If we encourage people to develop and extend minishift, we should treat each string as potentially visible, and therefore it should be clear and useful. I accept that I might not be an expert in the inner workings of the code but again, this is why we all work together to combine our knowledge. This is one of those tasks that you do once in the very beginning and then you don't have to worry about it, so hopefully we can resolve whatever open issues and move on with our lives :)

hferentschik:
@thatdocslady, lets move this along. Could you please create a new issues regarding this actual change. Something along the line of "reviewing info and error message" and create this pull request against this issue instead. Then you can also rebase once more onto master.

thatdocslady:
@hferentschik done, see PR #270 and issue #269. Closing this PR.

PR #253, opened 2017-01-05
Issue #206 oc downloaded on each start of Minishift

### LINKED ISSUE #206: oc downloaded on each start of Minishift
**Minishift  1.0.0-beta.1 + Windows + VirtualBox**

Hi 
I'm starting and stopping minishift on window 10 using the following OC commands:
```
minishift start
...
...
minishift stop
```

Every start up I see a download of 18.42MB and everything is reset again and basically loose all my projects. If this is usual/expected behaviour then what is the oc delete used for in that case?

`6.52 MB / 18.42 MB [======================================================================>---------------------------------------------------------------------------------------------------------------------------------]  35.41% 1m37`

budhrg:
Fix #206 

Tested in Windows and Linux :
```
PS C:\gowork\src\github.com\minishift\minishift> go test -v .\pkg\minishift\cache
=== RUN   TestIsCached
--- PASS: TestIsCached (0.01s)
=== RUN   TestCacheOc
--- PASS: TestCacheOc (3.30s)
PASS
ok      github.com/minishift/minishift/pkg/minishift/cache      3.431s
```

### DIFF
--- a/cmd/minikube/cmd/start_test.go
+++ b/cmd/minikube/cmd/start_test.go
@@ -29,6 +29,7 @@ import (
 	"path/filepath"
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"net/http"
+	"github.com/minishift/minishift/pkg/minikube/constants"
 )
 
 type RecordingRunner struct {
@@ -57,7 +58,7 @@ func TestStartClusterUpNoFlags(t *testing.T) {
 
 	clusterUp(&testMachineConfig)
 
-	expectedOc := filepath.Join(testDir, "cache", "oc", testMachineConfig.OpenShiftVersion, "oc")
+	expectedOc := filepath.Join(testDir, "cache", "oc", testMachineConfig.OpenShiftVersion, constants.OC_BINARY_NAME)
 	if testRunner.Cmd != expectedOc {
 		t.Errorf("Expected command '%s'. Got '%s'", expectedOc, testRunner.Cmd)
 	}
--- a/pkg/minishift/cache/oc_caching.go
+++ b/pkg/minishift/cache/oc_caching.go
@@ -19,14 +19,12 @@ package cache
 import (
 	"os"
 	"path/filepath"
-
+	"strings"
 	"github.com/pkg/errors"
-
 	"github.com/minishift/minishift/pkg/util/github"
 	minishiftos "github.com/minishift/minishift/pkg/util/os"
+	"github.com/minishift/minishift/pkg/minikube/constants"
 )
-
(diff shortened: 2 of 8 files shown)


pkg/minishift/cache/oc_caching.go
```diff
@@ -50,7 +49,12 @@ func (oc *Oc) GetCacheFilepath() string {
 }
 
 func (oc *Oc) isCached() bool {
-	if _, err := os.Stat(filepath.Join(oc.GetCacheFilepath(), OC_BINARY_NAME)); os.IsNotExist(err) {
+	binaryName := OC_BINARY_NAME
+	if runtime.GOOS == "windows" {
```
hferentschik: Why do you do the check here? Why not set the _OC_BINARY_NAME_ correctly depending on the OS. What if other code uses 'OC_BINARY_NAME' in a different context. They will run into the same issue.

hferentschik: Could catch otherwise

budhrg:
@hferentschik Updated with constants OC_BINARY_NAME approach as we discussed.

cmd/minikube/cmd/start_test.go
```diff
@@ -57,7 +58,7 @@ func TestStartClusterUpNoFlags(t *testing.T) {
 
 	clusterUp(&testMachineConfig)
 
-	expectedOc := filepath.Join(testDir, "cache", "oc", testMachineConfig.OpenShiftVersion, "oc")
+	expectedOc := filepath.Join(testDir, "cache", "oc", testMachineConfig.OpenShiftVersion, constants.OC_BINARY_NAME)
```
hferentschik: ahh, that's why the test passed on Windows.

pkg/minishift/cache/oc_caching.go
```diff
@@ -46,11 +44,11 @@ func (oc *Oc) EnsureIsCached() error {
 }
 
 func (oc *Oc) GetCacheFilepath() string {
-	return filepath.Join(oc.MinishiftCacheDir, OC_BINARY_NAME, oc.OpenShiftVersion)
+	return filepath.Join(oc.MinishiftCacheDir, strings.TrimRight(constants.OC_BINARY_NAME, ".exe"), oc.OpenShiftVersion)
```
hferentschik: I am going to replace this with a const - "const OC_CACHE_DIR = "oc"

hferentschik:
Looking good. I will make a small change to _oc_caching.go_, but otherwise ok.

hferentschik:
Thanks. Merged.

PR #256, opened 2017-01-06
Issue #216 Improve startup message to include selected driver

### LINKED ISSUE #216: Improve startup message to include selected driver
I think we should add the selected driver to the startup message to make it more explicit which driver is selected. See also question #215.

Currently we have:

> Starting local OpenShift cluster...

I suggest we have something like this

> Starting local OpenShift instance using \<driver\> hypervisor ...

budhrg:
Fix #216

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,7 +100,7 @@ The interaction with OpenShift is via the command line tool _oc_ which is copied
 1. Assuming you have put _minishift_  on the _PATH_ as described in [Installing Minishift](#installing-minshift) you can start Minishift via:
 
         $ minishift start
-        Starting local OpenShift cluster...
+        Starting local OpenShift instance using 'kvm' hypervisor...
         ...
            OpenShift server started.
            The server is accessible via web console at:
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
Tested on OSX+xhyve and Virtualbox and it works fine.

cmd/minikube/cmd/start.go
```diff
@@ -126,6 +124,7 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	var host *host.Host
 	start := func() (err error) {
+		fmt.Printf("Starting local OpenShift instance using '%s' hypervisor...\n", config.VMDriver)
```
praveenkumar: LGTM :+1: , I am not sure we can add any test case for this change because that's a print statement.

budhrg: Ya, I am not able to think what approach to use to write any test case. Let me know @hferentschik , if you have any approach.

hferentschik: >  I am not able to think what approach to use to write any test case for above change. Let me know @hferentschik , if you have any approach.

This is very much testable :-), however, I think in this case it is not worth the effort. 

One way to test this, is to simply capture the stdout and make sure the log statement comes out. Eventually we will need some helper utility for that. For now we should be good.

hferentschik: @budhrg any reason why you moved the log statement. You put it at a place where it potentially will be printed several times. I am inclined to move it back.

README.md
```diff
@@ -113,7 +113,8 @@ The interaction with OpenShift is via the command line tool _oc_ which is copied
            To login as administrator:
                oc login -u system:admin
 
-    Note that, the IP seen above is dynamic and can change. It can be retrieved with `minishift ip`.
+    Note that, the IP seen above is dynamic and can change. It can be retrieved with `minishift ip`. Also,
+    instead of 'kvm', you will see 'xhyve' in Mac OS and 'hyperv' in Windows.
```
hferentschik: 'on'

cmd/minikube/cmd/start.go
hferentschik: I guess you need the config, but you still don't have to put it into the function which gets retried.

hferentschik:
Merged with some minor modifications by adding commit 7c32e0b67cfad65c446c12160a666797f2bc0b4f

PR #260, opened 2017-01-09
Issue #259 run single test/test case/package

### LINKED ISSUE #259: Document how to run single test/test case/package in development section
Currently, in the [development section#running-tests](https://github.com/minishift/minishift/blob/master/docs/developing.md#running-tests), there is no instruction how one can run single test/ test case or package.

It is needed for faster development as running entire tests with `make test` may take time.

Document the possible ways to test single test/test case or package.

budhrg:
Fix #259

### DIFF
--- a/docs/developing.md
+++ b/docs/developing.md
@@ -151,11 +151,32 @@ For more options refer to the _minishift_ [synopsis](./minishift.md).
 <a name="unit-tests"></a>
 #### Unit Tests
 
-Unit tests are run on Travis before code is merged. To run as part of a development cycle:
+Unit tests get run on Travis before code is merged. To run as part of a development cycle:
 
 ```shell
 make test
 ```
+
+To run specific tests, use one of the following approach:
+
+- Run tests within entire package
+```
+go test -v <relative path of package>
+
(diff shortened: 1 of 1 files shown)


docs/developing.md
```diff
@@ -151,12 +151,31 @@ For more options refer to the _minishift_ [synopsis](./minishift.md).
...
+- Run specific test case
+```
+go test -v <relative path of package> -run <Testcase Name>
+```
+
+- Run tests matching pattern
+```
+go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
+```
+
```
praveenkumar: I would suggest to add `go test --help` at the end for reference if someone want to know what other flags they can use.

budhrg: Added @praveenkumar :+1:

hferentschik: You probably would like to indent this as part of the enum, right? Now your code blocks are outside. I'll take care of this.

budhrg: ok @hferentschik

hferentschik:
Merged, thanks. I adjusted the code block formatting though.

Issue #235, opened 2016-12-29
Change mentions of jimmidyson to minishift in the code and documents


LalatenduMohanty:
We have moved from https://github.com/jimmidyson/minishift to https://github.com/minishift/minishift. However there are still mentions of jimmidyson in the code and it does not make sense now.
```
$ git grep jimmidyson

cmd/minikube/cmd/config/view.go:For the list of accessible variables for the template, see the struct values here: https://godoc.org/github.com/jimmidyson/minishift/cmd/minikube/cmd/config#ConfigViewTemplate`)
docs/minishift_config_view.md:For the list of accessible variables for the template, see the struct values here: https://godoc.org/github.com/jimmidyson/minishift/cmd/minikube/cmd/config#ConfigViewTemplate (default "- {{.ConfigKey}}: {{.ConfigValue}}\n")
glide.lock:  repo: https://github.com/jimmidyson/viper.git
glide.yaml:  repo: https://github.com/jimmidyson/viper.git
pkg/minikube/update/update.go:  githubOwner = "jimmidyson"
pkg/version/version.go:// This is a private field and should be set when compiling with --ldflags="-X github.com/jimmidyson/minishift/pkg/version.version=vX.Y.Z"
pkg/version/version.go:// This is a private field and should be set when compiling with --ldflags="-X github.com/jimmidyson/minishift/pkg/version.openshiftVersion=vX.Y.Z"
```

hferentschik:
Not all of this can just be changed. There is more to it than that. Some of these mentions are legit.

LalatenduMohanty:
@hferentschik 
Apart from 
```
glide.lock:  repo: https://github.com/jimmidyson/viper.git
glide.yaml:  repo: https://github.com/jimmidyson/viper.git
```
These lines can be fixed I think.

```
cmd/minikube/cmd/config/view.go:For the list of accessible variables for the template, see the struct values here: https://godoc.org/github.com/jimmidyson/minishift/cmd/minikube/cmd/config#ConfigViewTemplate`)
docs/minishift_config_view.md:For the list of accessible variables for the template, see the struct values here: https://godoc.org/github.com/jimmidyson/minishift/cmd/minikube/cmd/config#ConfigViewTemplate (default "- {{.ConfigKey}}: {{.ConfigValue}}\n")
pkg/minikube/update/update.go:  githubOwner = "jimmidyson"
[...]

PR #264, opened 2017-01-10
[Do not merge] Issue #235 Removing mentions of jimmidyson

LalatenduMohanty:

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: f0f9906056562c5ccea818ad27d50572a73d18ff3ef61c4b389ff12672925272
-updated: 2016-11-29T11:56:31.490022127+01:00
+hash: 8e3c08a2cb219c817d013aa8cd4da680ae5228f5f3cc6d73fdad721417fa3080
+updated: 2017-01-10T22:55:45.011971272+05:30
 imports:
 - name: github.com/beorn7/perks
   version: 3ac7bf7a47d159a033b107610db8a1b6575507a4
@@ -7,8 +7,6 @@ imports:
   - quantile
 - name: github.com/blang/semver
   version: 31b736133b98f26d5e078ec9eb591666edfd091f
-- name: github.com/BurntSushi/toml
-  version: 99064174e013895bbd9b025c31100bd1d9b590ca
 - name: github.com/coreos/go-oidc
   version: 5cf2aa52da8c574d3aa4458f471ad6ae2240fe6b
   subpackages:
@@ -61,6 +59,7 @@ imports:
   subpackages:
   - commands/mcndirs
   - drivers/errdriver
+  - drivers/fakedriver
   - drivers/hyperv
   - drivers/none
   - drivers/virtualbox
@@ -83,6 +82,7 @@ imports:
   - libmachine/persist
   - libmachine/provision
   - libmachine/provision/pkgaction
+  - libmachine/provision/provisiontest
   - libmachine/provision/serviceaction
   - libmachine/shell
   - libmachine/ssh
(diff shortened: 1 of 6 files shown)


LalatenduMohanty:
Steps done for testing the PR

```
```
$ rm -rf vendor
$ make vendor
$ make
$ make test
$ minishift start
$ minishift delete
```

**Comment by budhrg** [comment]
Working fine for me.
Tested as:
```
$ make clean
$ make 
$ make test
$ make start
$ # deploy app
$ make stop
$ make start
$ # check app
```

LGTM :+1:

File: glide.lock [hunk 1]
```diff
@@ -341,7 +341,7 @@ imports:
   version: 1560c1005499d61b80f865c04d39ca7505bf7f0b
 - name: github.com/spf13/viper
   version: c975dc1b4eacf4ec7fdbf0873638de5d090ba323
-  repo: https://github.com/jimmidyson/viper.git
+  repo: https://github.com/spf13/viper.git
```
hferentschik: this does not look right. Did you remove the lock file and regenerate it after updating _glide.yaml_ or did you just do a find and replace.

glide.yaml
```diff
@@ -31,7 +31,7 @@ import:
 - package: github.com/olekukonko/tablewriter
 - package: github.com/pkg/browser
 - package: github.com/spf13/viper
-  repo: https://github.com/jimmidyson/viper.git
+  repo: https://github.com/spf13/viper.git
```
hferentschik: If we don't use a custom repo we don't need the explicit _repo_ part. But we might need to set a version. See also https://github.com/minishift/minishift/blob/master/docs/developing.md#updating-dependencies

glide.lock
hferentschik: I am also suspicious since we are referencing a commit sha here, but this sha might not be in the main repo.

budhrg: @hferentschik Oh ya, I tested with removing glide.lock. but found that `spf13/viper.git` version is same as ` c975dc1b4eacf4ec7fdbf0873638de5d090ba323`.

hferentschik:
@budhrg One cannot just "change" _glide.lock_, you want to make the change to _glide.yaml_ and then re-generate the lock file.

budhrg:
@hferentschik , I did remove `glide.lock` file and regenerated with the lala's `glide.yaml` changes. It looks fine.

glide.yaml
budhrg: > If we don't use a custom repo we don't need the explicit repo part. 

Yes

Setting version will help to get to stable point.

hferentschik:
So if you actually follow the proper procedure to do switch to the official Viper repo, you get:

    CGO_ENABLED=0 GOARCH=amd64 GOOS=darwin go build --installsuffix cgo -ldflags="-X github.com/minishift/minishift/    pkg/version.version=1.0.0-beta.2 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v1.3.1 -s -w     -extldflags '-static'" -o ./out/darwin-amd64/minishift ./cmd/minikube
    # github.com/minishift/minishift/vendor/github.com/spf13/viper
    vendor/github.com/spf13/viper/viper.go:741: undefined: mapstructure.StringToTimeDurationHookFunc
    make: *** [out/darwin-amd64/minishift] Error 2

So something is still astray. If we want to remove the usage of jimmy's repo, we need to dig a bit deeper here and see whether we can resolve this in any other way (understand what jimmy's repo actually changes and whether there some alternative in the official Viper repo or whether we can inline a helper or so)

LalatenduMohanty:
@hferentschik Right, I had manually modified glide.lock , now as per your comment I could reproduce the error.

LalatenduMohanty:
Closing the PR now. Will resend after fixing the issue.

PR #272, opened 2017-01-10
Issue #190 Editing cobra doc strings

### LINKED ISSUE #190: Review and update command doc strings

thatdocslady:
Fixes #190 

This PR was originally submitted as PR #198 but following feedback and request from @hferentschik was split into a new PR.

Notes about this PR:

- All feedback implemented from previous PR 
- Rebased against master, tested locally, and generated MD docs for the check-in
- Cleaned up all NEEDINFO comments and will create follow-up issues for open questions or issues that need further investigation
- Consulted with OpenShift doc writers and based on their style guide was advised that we should use "cluster" when describing the OpenShift that we run inside the VM, because the term refers to the combined elements required to run the instance, and the term "instance" is too generic (OpenShift is also working to standardize this in their docs). So we can use "local OpenShift cluster", "single-node OpenShift cluster", etc. See the style guide for clarification: https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/term_glossary.adoc#cluster

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -75,8 +75,9 @@ var (
 var startCmd = &cobra.Command{
 	Use:   commandName,
 	Short: "Starts a local OpenShift cluster.",
-	Long: `Starts a local OpenShift cluster using Virtualbox. This command
-assumes you already have Virtualbox installed.`,
+	Long: `Starts a local single-node OpenShift cluster on the specified hypervisor.
+If you are restarting the cluster and you use persistant storage for the host data,
+you can specify --use-existing-config to maintain the cluster state.`,
 	Run: runStart,
 }
 
@@ -111,6 +112,7 @@ func SetMinishiftDir(newDir string) {
 
 // runStart is executed as part of the start command
 func runStart(cmd *cobra.Command, args []string) {
+	fmt.Println("Starting the local OpenShift cluster...")
 	libMachineClient := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer libMachineClient.Close()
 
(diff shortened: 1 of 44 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -181,30 +183,30 @@ func init() {
...
+	startFlagSet.Int(memory, constants.DefaultMemory, "Amount of RAM to allocate to the Minishift VM.")
+	startFlagSet.Int(cpus, constants.DefaultCPUS, "Number of CPU cores to allocate to the Minishift VM.")
+	startFlagSet.String(humanReadableDiskSize, constants.DefaultDiskSize, "Disk size to allocate to the Minishift VM. Use the format <size><unit>, where unit = b, k, m or g.")
+	startFlagSet.String(hostOnlyCIDR, "192.168.99.1/24", "The CIDR to be used for the minishift VM. (Only supported with VirtualBox driver.)")
+	startFlagSet.StringSliceVar(&dockerEnv, "docker-env", nil, "Environment variables to pass to the Docker daemon. Use the format <key>=<value>.")
+	startFlagSet.StringSliceVar(&insecureRegistry, "insecure-registry", []string{"172.30.0.0/16"}, "Non-secure Docker registries to pass to the Docker daemon.")
+	startFlagSet.StringSliceVar(&registryMirror, "registry-mirror", nil, "Registry mirrors to pass to the Docker daemon.")
+	startFlagSet.String(openshiftVersion, version.GetOpenShiftVersion(), "The OpenShift version to run. Use the format v<n.n.n>")
```
hferentschik: 'v<n.n.n>' is not correct. First it implies that the same digit is repeated and secondly, there is no guarantee that the format will always be like this.

cmd/minikube/cmd/start.go
```diff
@@ -181,30 +183,30 @@ func init() {
...
-	clusterUpFlagSet.String(publicHostname, "", "Public hostname for OpenShift cluster")
-	clusterUpFlagSet.String(routingSuffix, "", "Default suffix for server routes")
-	clusterUpFlagSet.String(hostConfigDir, hostConfigDirectory, "Directory on Docker host for OpenShift configuration")
-	clusterUpFlagSet.String(hostVolumesDir, dockerhost.DefaultVolumesDir, "Directory on Docker host for OpenShift volumes")
-	clusterUpFlagSet.String(hostDataDir, hostDataDirectory, "Directory on Docker host for OpenShift data. If not specified, etcd data will not be persisted on the host.")
-	clusterUpFlagSet.Bool(forwardPorts, false, "Use Docker port-forwarding to communicate with origin container. Requires 'socat' locally.")
-	clusterUpFlagSet.Int(serverLogLevel, 0, "Log level for OpenShift server")
-	clusterUpFlagSet.StringSliceVarP(&openShiftEnv, openshiftEnv, "e", []string{}, "Specify key value pairs of environment variables to set on OpenShift container")
+	clusterUpFlagSet.Bool(skipRegistryCheck, false, "Skip the Docker daemon registry check.")
+	clusterUpFlagSet.String(publicHostname, "", "Public host name of the OpenShift cluster.")
```
hferentschik: It is actually 'hostname'

hferentschik:
thanks, merged. I applied Go formatting and made some additional changes via commit 78df115166125f33287d5d4ab225bacb2b24f88f

Issue #278, opened 2017-01-11
Ensure major flags can be configured via persistent options on config.json


hferentschik:
For CDK reasons we need to be at least able to set the default driver and iso-url. Not sure whether we can and should extend this to all flags, similar to the MINISHIFT_XYZ environment variables. Maybe this can even be supported out of the box.

hferentschik:
the following flags are currently supported via _config.json_:

	{
		name:        "vm-driver",
		set:         SetString,
		validations: []setFn{IsValidDriver},
		callbacks:   []setFn{RequiresRestartMsg},
	},
	{
		name:        "v",
		set:         SetInt,
		validations: []setFn{IsPositive},
	},
	{
		name:        "cpus",
		set:         SetInt,
		validations: []setFn{IsPositive},
		callbacks:   []setFn{RequiresRestartMsg},
	},
	{
		name:        "disk-size",
		set:         SetString,
		validations: []setFn{IsValidDiskSize},
		callbacks:   []setFn{RequiresRestartMsg},
	},
	{
		name:        "host-only-cidr",
		set:         SetString,
		validations: []setFn{IsValidCIDR},
	},
	{
		name:        "memory",
		set:         SetInt,
		validations: []setFn{IsPositive},
		callbacks:   []setFn{RequiresRestartMsg},
	},
	{
		name: "show-libmachine-logs",
		set:  SetBool,
	},
	{
		name:        "log_dir",
		set:         SetString,
		validations: []setFn{IsValidPath},
	},
	{
		name: "openshift-version",
		set:  SetString,
	},
	{
		name: config.WantUpdateNotification,
		set:  SetBool,
	},
	{
		name: config.ReminderWaitPeriodInHours,
		set:  SetInt,
	},

See _config.go_

hferentschik:
For now we need at least the iso-url as well.

hferentschik:
See also https://issues.jboss.org/browse/CDK-46 for which this is required.

hferentschik:
@praveenkumar when you add some docs around this, have a look at issue #279. The questions around the config command listed there, give you an idea what we probably should cover as part of the docs.

hferentschik:
Merged via pull request #304

PR #286, opened 2017-01-13
Issue #134: Removing boot2docker ISO creation

LalatenduMohanty:
Now the the default ISO would be from github.com/minishift/minishift-b2d-iso


LalatenduMohanty:
@hferentschik Actually I wanted to cover part of https://github.com/minishift/minishift/issues/278 in this PR. But I am still do not have solution for that and also due to lack of time (as I am going PTO soon) trying to just fix #134.

pkg/minikube/constants/constants.go
```diff
@@ -44,7 +43,7 @@ const (
 	DefaultDiskSize = "20g"
 )
 
-var DefaultIsoUrl = "https://github.com/minishift/minishift/releases/download/v" + version.GetVersion() + "/boot2docker.iso"
+var DefaultIsoUrl = "https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.0/minishift-b2d.iso"
```
praveenkumar: Does hardcoded verison make sense can't we add a go function to get latest release url? I have prototype in python which I can put in go also : https://gist.github.com/praveenkumar/f5e26a20f2b6a6640b86bc2934371bf9

@hferentschik wdyt?

LalatenduMohanty: +1 to the idea of automatically fetching the latest release. I am only concerned with the call to api.github.com. Users are already facing some issues with it with respect to number of calls being made.

hferentschik: -1 for hard coded the version like this. If we hard code it (which I am ok doing initially), we need to do it in the Makefile (inlining the main Minishift version and the OpenShift version).

If we want to GitHub, we want to define a base version say 1.x and then use semantic versioning to get the latest version which is compatible. Not just the latest version. The functionality is already there to do this. So on sets the base version 1.x (eg in the Makefile). Then you go out to GitHub and get the latest version which is compatible in a semantic versioning sense. This is what I had in mind.

There is no risk to hit the Git Hub API rate limit here. We only do this when the ISO is not cached. There is no automatic update right now.

Makefile
```diff
@@ -14,6 +14,8 @@
 
 VERSION ?= $(shell cat VERSION)
 OPENSHIFT_VERSION ?= $(shell cat OPENSHIFT_VERSION)
+OPENSHIFT_VERSION ?= $(shell cat OPENSHIFT_VERSION)
+ISO_VERSION ?= $(shell cat ISO_VERSION)
```
hferentschik: Can we inline this version files? I find we have a bit of a proliferation of version files here.

LalatenduMohanty:
@hferentschik I did not create the GetIsoSemverVersion because I do not see the use of the function in the current code. We can create one when we need to use it. 
I am done with my testing and it is ready for review.

hferentschik:
>  I did not create the GetIsoSemverVersion because I do not see the use of the function in the current code

Sure. I guess for now we can go with a single hard coded version. Let's see whether can refine this for this sprint or whether we pick this up later. 

> I am done with my testing and it is ready for review.

ok

hferentschik:
Created issue #292 to follow up on this.

Also added a commit to add some more logging and a commit to get completely rid of the _deploy_ directory.

pkg/version/version.go
```diff
@@ -47,3 +48,7 @@ func GetOpenShiftVersion() string {
 func GetOpenShiftSemverVersion() (semver.Version, error) {
 	return semver.Make(strings.TrimPrefix(GetOpenShiftVersion(), VersionPrefix))
```
hferentschik: @LalatenduMohanty here is your code which makes use of semantic versioning. You added your code right below ;-) This is using https://github.com/blang/semver which allows you to compare versions and determine whether they are compatible in a semantic versioning sense.

LalatenduMohanty: @hferentschik I have seen this but I guess I missed https://github.com/minishift/minishift/blob/master/pkg/minikube/update/update.go#L57

Issue #294, opened 2017-01-14
Wrong instructions on developing doc


cmoulliard:
On MacOS with `go version go1.6.2 darwin/amd64`, the commands to be used to prepare the project in order to build are 

1) Glide 

```
glide install -v --strip-vcs
```

and not 

```
make vendor
glide install -v
[ERROR]	--strip-vendor cannot be used without --strip-vcs
make: *** [vendor] Error 1
```

2) Run

```
./out/darwin-amd64/minishift version
Minishift version: 1.0.0-beta.2
```

and not 

```
./out/minishift start
```

gbraad:
Thanks for the report. Yes, the current documentation does not reflect the current process. We are in the process of updating this and adding instructions for platform other than OSX.

hferentschik:
@cmoulliard, regarding Glide, the target you want to execute is 'make vendor' which under the hood calls glide. As @gbraad pointed out, you seem to have an outdated development environment. What is your version of Glide? Mine is v0.12.3 and if I set the '--strop-vcs' flag I get:

     [WARN]	The --strip-vcs flag is deprecated. This now works by default.

Maybe we want to add a note, about expected versions in this case. The target itself is correct though.

The minishift binary, though, is in indeed in a different directory now. This changed when we switched to releasing archive bundles instead of single binaries. I am going to update this, since I also have a couple of other things I want to add.

gbraad:
@hferentschik I did not have this problem with glide. Was pushing the other change already

hferentschik:
Resolved via pull request #296

PR #296, opened 2017-01-16
Issue #294 Update development instructions

gbraad:
Changes related to the platform specific output of binaries.


hferentschik:
Merged, thanks. I added another commit with some more info and updates to the release process as well.

Issue #299, opened 2017-01-16
Fix broken links and minor formatting and language issues in the Using.md


Preeticp:

hferentschik:
Merged via pull request #300

PR #300, opened 2017-01-16
Issue #299 Corrected a few broken links and minor formatting …

Preeticp:
…and language corrections.

Addresses issue #299

### DIFF
--- a/docs/minishift.md
+++ b/docs/minishift.md
@@ -24,17 +24,16 @@ Minishift is a command-line tool that provisions and manages single-node OpenShi
 ```
 
 ### SEE ALSO
-* [minishift config](minishift_config.md)	 - Modifies Minishift configuration properties.
-* [minishift console](minishift_console.md)	 - Opens or displays the OpenShift Web Console URL.
-* [minishift delete](minishift_delete.md)	 - Deletes the Minishift VM.
-* [minishift docker-env](minishift_docker-env.md)	 - Sets Docker environment variables.
-* [minishift get-openshift-versions](minishift_get-openshift-versions.md)	 - Gets the list of OpenShift versions available for Minishift.
-* [minishift ip](minishift_ip.md)	 - Gets the IP address of the running cluster.
-* [minishift logs](minishift_logs.md)	 - Gets the logs of the running OpenShift cluster.
-* [minishift service](minishift_service.md)	 - Prints the URL for the specified service to the console.
-* [minishift ssh](minishift_ssh.md)	 - Log in to or run a command on a Minishift VM with SSH.
-* [minishift start](minishift_start.md)	 - Starts a local OpenShift cluster.
-* [minishift status](minishift_status.md)	 - Gets the status of the local OpenShift cluster.
-* [minishift stop](minishift_stop.md)	 - Stops the running local OpenShift cluster.
-* [minishift version](minishift_version.md)	 - Gets the version of Minishift.
-
+* [minishift config](./minishift_config.md)	 - Modifies Minishift configuration properties.
+* [minishift console](./minishift_console.md)	 - Opens or displays the OpenShift Web Console URL.
+* [minishift delete](./minishift_delete.md)	 - Deletes the Minishift VM.
+* [minishift docker-env](./minishift_docker-env.md)	 - Sets Docker environment variables.
+* [minishift get-openshift-versions](./minishift_get-openshift-versions.md)	 - Gets the list of OpenShift versions available for Minishift.
+* [minishift ip](./minishift_ip.md)	 - Gets the IP address of the running cluster.
+* [minishift logs](./minishift_logs.md)	 - Gets the logs of the running OpenShift cluster.
+* [minishift service](./minishift_service.md)	 - Prints the URL for the specified service to the console.
+* [minishift ssh](./minishift_ssh.md)	 - Log in to or run a command on a Minishift VM with SSH.
+* [minishift start](./minishift_start.md)	 - Starts a local OpenShift cluster.
+* [minishift status](./minishift_status.md)	 - Gets the status of the local OpenShift cluster.
+* [minishift stop](./minishift_stop.md)	 - Stops the running local OpenShift cluster.
+* [minishift version](./minishift_version.md)	 - Gets the version of Minishift.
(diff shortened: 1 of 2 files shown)


docs/using.md
```diff
@@ -29,20 +29,20 @@ This section contains information about basic virtual machine and OpenShift mana
...
+The [minishift start](minishift_start.md) command is used to start your OpenShift instance.
 This command creates and configures a virtual machine that runs a single-node OpenShift instance.
 
 <a name="stopping-openshift"></a>
 ### Stopping OpenShift
 
-The [minishift stop](./docs/minishift_stop.md) command can be used to stop your OpenShift instance.
-This command shuts down the Minishift virtual machine, but preserves all instance state and data.
+The [minishift stop](minishift_stop.md) command is used to stop your OpenShift instance.
+This command shuts down the Minishift virtual machine, but preserves the instance state and data.
```
hferentschik: hmm, is it not 'instance's' in this case? Genitive!? Personally I feel 'all' read better.

docs/using.md
```diff
@@ -29,20 +29,20 @@ This section contains information about basic virtual machine and OpenShift mana
...
 Starting the instance again will restore it to it's previous state.
 
 <a name="deleting-openshift"></a>
 ### Deleting OpenShift
 
-The [minishift delete](./docs/minishift_delete.md) command can be used to delete the OpenShift instance.
+The [minishift delete](minishift_delete.md) command is used to delete the OpenShift instance.
```
hferentschik: Same as above

docs/using.md
```diff
@@ -51,17 +51,17 @@ This command shuts down and deletes the Minishift virtual machine. No data or st
 Minishift allows you to specify command line flags you commonly use via environment variables.
 To do so, apply the following rules to the flag you want to set via an environment variable.
 
-* Apply the _MINISHIFT__ as a prefix to your environment variable, for example the _vm-driver_ flag
-  of the [start](./docs/minishift_start.md) command becomes _MINISHIFT_vm-driver_.
-* Uppercase the flag, _MINISHIFT_vm-driver_ becomes _MINISHIFT_VM-DRIVER_.
-* Last but not least, replace _-_ with _\__, _MINISHIFT_VM-DRIVER_ becomes _MINISHIFT_VM_DRIVER_
+* Apply the `MINISHIFT_` as a prefix to your environment variable, for example the `vm-driver` flag
```
hferentschik: Actually, we need to drop the 'the': 'Apply `MINISHIFT_` as a prefix'

docs/using.md
```diff
@@ -51,17 +51,17 @@ This command shuts down and deletes the Minishift virtual machine. No data or st
...
+  of the [start](minishift_start.md) command becomes `MINISHIFT_vm-driver`.
```
hferentschik: same as above, current directory

docs/using.md
```diff
@@ -51,17 +51,17 @@ This command shuts down and deletes the Minishift virtual machine. No data or st
...
+* Uppercase the flag, `MINISHIFT_vm-driver` becomes `MINISHIFT_VM-DRIVER`.
+* Last but not least, replace `-` with `_`, `MINISHIFT_VM-DRIVER` becomes `MINISHIFT_VM_DRIVER`
 
 Another common example might be the URL of the ISO to be used. Usually you specify it via
-_iso-url_ of the [start](./docs/minishift_start.md) command. Applying the rules from above, you can
-also specify this URL by setting the environment variable _MINISHIFT_ISO_URL_.
+`iso-url` of the [start](minishift_start.md) command. Applying the rules from above, you can
```
hferentschik: same as above

docs/using.md
```diff
@@ -104,8 +104,8 @@ minishift service [-n NAMESPACE] [--url] NAME
 <a name="mounted-host-folders"></a>
 ## Mounted host folders
 
-Some of drivers will mount a host folder within the VM so that you can easily share files between the VM and host.
-These are not configurable at the moment and are different for each driver and OS that you use.
+Some of the drivers will mount a host folder within the VM so that you can easily share files between the VM and the host.
```
hferentschik: I think it is almost better to drop the 'of the'

docs/using.md
Preeticp: You are right, misunderstood this.

docs/using.md
```diff
@@ -29,20 +29,20 @@ This section contains information about basic virtual machine and OpenShift mana
 <a name="starting-openshift"></a>
 ### Starting OpenShift
 
-The [minishift start](./docs/minishift_start.md) command can be used to start your OpenShift instance.
+The [minishift start](minishift_start.md) command is used to start your OpenShift instance.
```
Preeticp: In that case, I will make these corrections to the minishift .md too. I picked this style from that doc, so will change it there too to keep it consistent.

hferentschik: +1

docs/using.md
hferentschik: Two things, we most likely need to say "preserves the cluster state", since I think we agreed that we are talking in terms of (single node) cluster. 

> What does 'state' exactly mean here?

Second, what we persist is the state of the VM, the state of the OpenShift config, the OpenShift runtime state and any persistent Minishift config. All basically. Everything needed to allow the user to re-start and continue where we was.

> In case we are starting/stopping multiple instances we can simply say: 'preserves all instances and data.'

We are dealing with a single instance or as said 'cluster'

docs/minishift.md
```diff
@@ -24,17 +24,16 @@ Minishift is a command-line tool that provisions and manages single-node OpenShi
...
-* [minishift ip](minishift_ip.md)	 - Gets the IP address of the running cluster.
-* [minishift logs](minishift_logs.md)	 - Gets the logs of the running OpenShift cluster.
-* [minishift service](minishift_service.md)	 - Prints the URL for the specified service to the console.
-* [minishift ssh](minishift_ssh.md)	 - Log in to or run a command on a Minishift VM with SSH.
-* [minishift start](minishift_start.md)	 - Starts a local OpenShift cluster.
-* [minishift status](minishift_status.md)	 - Gets the status of the local OpenShift cluster.
-* [minishift stop](minishift_stop.md)	 - Stops the running local OpenShift cluster.
-* [minishift version](minishift_version.md)	 - Gets the version of Minishift.
-
+* [minishift config](./minishift_config.md)	 - Modifies Minishift configuration properties.
```
hferentschik: This is really a generated page. It won't help to edit this by hand. I will drop this when merging

hferentschik:
Merged, dropping the changes to _minishift.md_

PR #301, opened 2017-01-16
Issue #278 Ensure major flags can be configured via persistent options on config.json

praveenkumar:
This patch will add some document around config.json and also add `--iso-url` to make it persistent.

### DIFF
--- a/cmd/minikube/cmd/config/config.go
+++ b/cmd/minikube/cmd/config/config.go
@@ -94,6 +94,12 @@ var settings []Setting = []Setting{
 		set:  SetString,
 	},
 	{
+		name:        "iso-url",
+		set:         SetString,
+		validations: []setFn{IsValidUrl},
+		callbacks:   []setFn{RequiresRestartMsg},
+	},
+	{
 		name: config.WantUpdateNotification,
 		set:  SetBool,
 	},
--- a/cmd/minikube/cmd/config/config_test.go
+++ b/cmd/minikube/cmd/config/config_test.go
@@ -41,12 +41,14 @@ var configTestCases = []configTestCase{
     "ReminderWaitPeriodInHours": 99,
     "cpus": 4,
     "disk-size": "20g",
+    "iso-url": "http://foo.bar/minishift-centos.iso",
     "log_dir": "/etc/hosts",
     "show-libmachine-logs": true,
     "v": 5,
     "vm-driver": "kvm"
 }`,
 		config: map[string]interface{}{
+			"iso-url": "http://foo.bar/minishift-centos.iso",
 			"vm-driver": "kvm",
 			"cpus":      4,
 			"disk-size": "20g",
--- a/cmd/minikube/cmd/config/set_test.go
+++ b/cmd/minikube/cmd/config/set_test.go
@@ -16,11 +16,36 @@ limitations under the License.
 
 package config
 
-import "testing"
+import (
+	"testing"
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	"os"
+)
 
 func TestNotFound(t *testing.T) {
 	err := set("nonexistant", "10")
 	if err == nil {
 		t.Fatalf("Set did not return error for unknown property")
+	}
+}
+
+func TestModifyData(t *testing.T) {
+	constants.ConfigFile = "/tmp/config.json"
+	defer os.Remove("/tmp/config.json")
+	err := set("cpus", "4")
+	if err != nil {
+		t.Fatalf("Error setting value %s", err)
+	}
+	// override existing value and check if that persistent
+	err = set("cpus", "10")
+	if err != nil {
+		t.Fatalf("Error setting value %s", err)
+	}
+	getValue, err := get("cpus")
+	if err != nil {
+		t.Fatalf("Error getting value %s", err)
+	}
+	if getValue != "10" {
+		t.Fatal("Not able to update data to config file")
 	}
 }
--- a/cmd/minikube/cmd/config/validations.go
+++ b/cmd/minikube/cmd/config/validations.go
@@ -21,6 +21,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"net/url"
 
 	units "github.com/docker/go-units"
 
@@ -72,6 +73,14 @@ func IsValidPath(name string, path string) error {
 	_, err := os.Stat(path)
 	if err != nil {
 		return fmt.Errorf("%s path is not valid: %v", name, err)
+	}
+	return nil
+}
+
+func IsValidUrl(string, isoURL string) error {
+	_, err := url.ParseRequestURI(isoURL)
+	if err != nil {
+		return fmt.Errorf("%s url is not valid: %v", isoURL, err)
 	}
 	return nil
 }
--- a/docs/using.md
+++ b/docs/using.md
@@ -10,6 +10,7 @@ overview of different components and services.
   - [Stopping OpenShift](#stopping-openshift)
   - [Deleting OpenShift](#deleting-openshift)
 - [Environment variables](#environment-variables)
+- [Config file](#config-file)
 - [Interacting with OpenShift](#interacting-with-openshift)
   - [OpenShift client binary \(oc\)](#openshift-client-binary-oc)
   - [Console](#console)
@@ -64,6 +65,33 @@ also specify this URL by setting the environment variable _MINISHIFT_ISO_URL_.
 its runtime state into _~/.minishift_. Using _MINISHIFT_HOME_, you can choose a different directory
 as Minishift's home directory. This is currently experimental and semantics might change in
 future releases.
+
+<a name="config-file"></a>
+## Config file
+
+Minishift also maintains a config file (_~/.minishift/config/config.json_) which can be used to set some
+variables like (CPU, memory ...etc.) and can be used without using different parameter in start command.
+This can be used to control some of the default behavior similar to using
+[environment variables](#environment-variables).
+
+**Note:**
+
+* config.json file is not generated automatic until user use `set` sub-command as described below.
+* Manual edit to this file is discouraged because it might be error-prone, use defined sub-command for required use-case.
+
+```shell
+$ minishift config -h
+```
+
+You can set using `set` sub-command provided by config and it expect `PROPERTY_NAME PROPERTY_VALUE`
+
+    # Set default memory 4096 MB
+    $ minishift config set memory 4096
+
+To view what already set and available you can use `view` sub-command
+
+    $ minishift config view
+    - memory: 4096
 
 <a name="interacting-with-openshift"></a>
 ## Interacting with OpenShift
--- a/cmd/minikube/cmd/config/validations_test.go
+++ b/cmd/minikube/cmd/config/validations_test.go
@@ -94,3 +94,27 @@ func TestValidCIDR(t *testing.T) {
 
 	runValidations(t, tests, "cidr", IsValidCIDR)
 }
+
+func TestValidURL(t *testing.T) {
+
+	var tests = []validationTest{
+		{
+			value:	   "http://foo.com/minishift.tar.gz",
+			shouldErr: false,
+		},
+		{
+			value:     "http/foo.com/minishift.tar.gz",
+			shouldErr: true,
+		},
+		{
+			value:     "file:///foo/download/minishift.tar.gz",
+			shouldErr: false,
+
+		},
+		{
+			value:      "file/foo/download/minishift.tar.gz",
+			shouldErr:  true,
+		},
+	}
+	runValidations(t, tests, "iso-url", IsValidUrl)
+}
--- a/docs/minishift_config.md
+++ b/docs/minishift_config.md
@@ -19,6 +19,7 @@ Configurable properties (enter as SUBCOMMAND):
  * show-libmachine-logs
  * log_dir
  * openshift-version
+ * iso-url
  * WantUpdateNotification
  * ReminderWaitPeriodInHours
 


cmd/minikube/cmd/config/validations.go
```diff
@@ -75,3 +76,11 @@ func IsValidPath(name string, path string) error {
...
 	return nil
 }
+
+func IsValidUrl(string, isoURL string) error {
+	_, err := url.ParseRequestURI(isoURL)
+	if err != nil {
+		return fmt.Errorf("%s url is not valid: %v", isoURL, err)
+	}
+	return nil
+}
```
budhrg: Looks like empty line is missing here.

praveenkumar: Added new line

cmd/minikube/cmd/config/config.go
```diff
@@ -94,6 +94,12 @@ var settings []Setting = []Setting{
 		set:  SetString,
 	},
 	{
+		name:        "iso-url",
```
hferentschik: +1

cmd/minikube/cmd/config/validations.go
```diff
@@ -75,3 +76,11 @@ func IsValidPath(name string, path string) error {
 	}
 	return nil
 }
+
+func IsValidUrl(string, isoURL string) error {
```
hferentschik: and this works for file based URIs as well? Can we add a test for this?

cmd/minikube/cmd/config/config_test.go
```diff
@@ -41,12 +41,14 @@ var configTestCases = []configTestCase{
...
     "cpus": 4,
     "disk-size": "20g",
+    "iso-url": "http://foo.bar/minishift-centos.iso",
     "log_dir": "/etc/hosts",
     "show-libmachine-logs": true,
     "v": 5,
     "vm-driver": "kvm"
 }`,
 		config: map[string]interface{}{
+			"iso-url": "http://foo.bar/minishift-centos.iso",
```
hferentschik: Good with tests, but they only seem to exercise the default case. What I am interested in as well, is whether I can have other settings in the config, eg _foo_ which basically get ignored then. The set/get functionality seems to just deal with _MinikubeConfig_. Can there be other data? If not, does this code throw an error? What if there is a mis-typed flag? Can we extend these tests a bit to cover some border cases? This will be a learning exercise as  well, since we might want to store other data in the config file which goes beyond _MinikubeConfig_. I would like to know whether this is possible. Also we need to better understand what happens, if the user manually edits these files and maybe introduces typos.
Also some tests around providing a config with not all settings in the config would be nice. One would then assert that default values apply.

hferentschik: I'd like to see some tests also persisting actually to disk where on then verifies the content via reading from disk (potentially utilizing _MINISHIFT_HOME_ to make sure the config file gets somewhere into a known location. Maybe not a must have right now, but it would be nice. Right now we just do everything in memory.

docs/using.md
```diff
@@ -60,6 +60,21 @@ Another common example might be the URL of the ISO to be used. Usually you speci
 _iso-url_ of the [start](./docs/minishift_start.md) command. Applying the rules from above, you can
 also specify this URL by setting the environment variable _MINISHIFT_ISO_URL_.
 
+Minishift also maintain a config file (_~/.minishift/config/config.json_) which can be used to set some
```
hferentschik: 'maintains'

docs/using.md
```diff
@@ -60,6 +60,21 @@ Another common example might be the URL of the ISO to be used. Usually you speci
...
+variable like (CPU, memory ...etc.) and can be used without using different parameter in start command.
```
hferentschik: variables

hferentschik: > and can be used without using different parameter in start command

". This can be used to control some of the default behavior similar to using environment variables. "

Also link to the environment variable section

docs/using.md
hferentschik: On a side note, right now, _config.json_ does not get generated unless you execute a _set_ command. This is a bit confusing for people looking for this file. So either we need to document that this file does not exist unless you add some custom configuration or we make sure that an empty config file gets generated in case it does not exist.

budhrg: +1 to generate empty config file on `minishift start`.

hferentschik:
@praveenkumar I am wondering whether we should make subscription credentials persistent as well. WDYT? Also maybe we should rename _username_ and _pasword_ to _subscription_username_ and _subscription_password_!?

budhrg:
> @praveenkumar I am wondering whether we should make subscription credentials persistent as well. WDYT? 

Might be security issue if we allow it.

cmd/minikube/cmd/config/config_test.go
praveenkumar: > The set/get functionality seems to just deal with MinikubeConfig. Can there be other data? If not, does this code throw an error?

What do you mean other data, can you provide some pointer?

> What if there is a mis-typed flag? Can we extend these tests a bit to cover some border cases? 

Set method already have test which tell if we put a unknown property and it does throw an error.

```
$ ./minikube config set foo bar
Cannot find property name foo
```

> Also we need to better understand what happens, if the user manually edits these files and maybe introduces typos.

If a user manually edit then minishift doesn't care about extra flags or if by mistake user introduced a typo in config that also ignored by minishift.

```
# Manually edited config file and introduced a typo and also another variable which is not
# part of property
$ ./minikube config view
- memory: 4096
- vm-drive: virtualbox
- foo: bar
- iso-url: file:/Users/prkumar/minishift-centos.iso

# minishift started ignoring typo and extra variable and used default values.
$ ./minishift start --show-libmachine-logs
Starting local OpenShift cluster using 'xhyve' hypervisor...
Running pre-create checks...
Creating machine...
...
```

praveenkumar: I think we can add a comment on top of that file and warn user not to edit it manually and use config set command.

praveenkumar:
> I am wondering whether we should make subscription credentials persistent as well. WDYT? Also maybe we should rename username and pasword to subscription_username and subscription_password!?

Agree on both, you want those changes to be part of this PR or separate one. I think a separate one will make more sense.

praveenkumar:
> Might be security issue if we allow it.

@budhrg what security issue are you talking about here?

docs/using.md
praveenkumar: I would prefer to document it instead auto-generating it then user try to edit manually instead of `set` sub-command but if we document it how to generate it using `set` sub-command then user have mindset to use it to setup a property and it will be less error-prone.

budhrg:
> @budhrg what security issue are you talking about here?

I believe, adding the username/password as persistent flags, it will add to config.json which anyone can access, right?

Correct me if I understood wrongly here.

docs/using.md
```diff
@@ -65,6 +66,33 @@ its runtime state into _~/.minishift_. Using _MINISHIFT_HOME_, you can choose a
...
+Minishift also maintains a config file (_~/.minishift/config/config.json_) which can be used to set some
+variables like (CPU, memory ...etc.) and can be used without using different parameter in start command.
+This can be used to control some of the default behavior similar to using
+[environment variables](#environment-variables).
+
+**Note:**
+
+* config.json file is not generated automatic until user use `set` sub-command as described below.
+* Manual edit to this file is discouraged because it might be error-prone, use defined sub-command for required use-case.
+
```
budhrg: Three extra lines here.

cmd/minikube/cmd/config/validations.go
praveenkumar: Yes it works with file based URI also https://golang.org/src/net/url/url_test.go  (check L#271) but it not identify OS specific rule like for windows file://<path> works and for Mac file:/<path> and linux file:///<path> works that something user should know and add it accordingly.

I will add a generic test case around it.

praveenkumar:
> I believe, adding the username/password as persistent flags, it will add to config.json which anyone can access, right?

anyone? only person who have access to that machine can view it and that even for environment variable use case where user can list set variable.

docs/using.md
praveenkumar: @budhrg It's already taken care.

docs/using.md
```diff
@@ -65,6 +66,33 @@ its runtime state into _~/.minishift_. Using _MINISHIFT_HOME_, you can choose a
...
+## Config file
+
+Minishift also maintains a config file (_~/.minishift/config/config.json_) which can be used to set some
+variables like (CPU, memory ...etc.) and can be used without using different parameter in start command.
+This can be used to control some of the default behavior similar to using
+[environment variables](#environment-variables).
+
+**Note:**
+
+* config.json file is not generated automatic until user use `set` sub-command as described below.
```
budhrg: probably `_config.json_` to differentiate then other text.

and sentence should be `config.json file is not generated automatically until user `set` sub-command ad described [below](link to "Set Configuration")

docs/using.md
```diff
@@ -65,6 +66,33 @@ its runtime state into _~/.minishift_. Using _MINISHIFT_HOME_, you can choose a
...
+* Manual edit to this file is discouraged because it might be error-prone, use defined sub-command for required use-case.
+
+```shell
+$ minishift config -h
+```
+
```
budhrg: Need a sub-heading as "Set Configuration" which can be referenced above https://github.com/minishift/minishift/pull/301/files#r96368747

docs/using.md
```diff
@@ -65,6 +66,33 @@ its runtime state into _~/.minishift_. Using _MINISHIFT_HOME_, you can choose a
...
+You can set using `set` sub-command provided by config and it expect `PROPERTY_NAME PROPERTY_VALUE`
+
+    # Set default memory 4096 MB
+    $ minishift config set memory 4096
+
```
budhrg: Need a sub-heading as "View Configuration"

cmd/minikube/cmd/config/validations.go
hferentschik: > but it not identify OS specific rule?

What do you mean? A file URI is a file URI right? It's something of the form 'file://<path>. I guess the _path_ part might differ a bit due to the drive character in Windows, but did we not have an issue around this and in the end we said that it also behaved correctly on Windows.

> I will add a generic test case around it.

+1

cmd/minikube/cmd/config/config_test.go
hferentschik: > What do you mean other data, can you provide some pointer?

Right now we don't have anything, but for some of the things I have in mind, we need to store some additional state information. I'd like to now whether we can add this to _config.json_ or whether we need a different config file.

> Set method already have test which tell if we put a unknown property and it does throw an error.

Sure, but what if the user edits the file or we add other custom data. Is this code smart enough to deal with this. Also there is a disconnect here. 'minishift set' will make sure to persist the properties listed here. However, Viper does not care about this and reads the plain config file. So if I manually add another flag (which is not listed here) to the config, I think it will be picked up fine by Viper. I think there might be some things we need to think through here.

> If a user manually edit then minishift doesn't care about extra flags or if by mistake user introduced a typo in config that also ignored by minishift.

See this is what I mean. So this code seems to be able to just filter out the settings for _MinikubeConfig_, ignoring other values. This means we could also store other data in _config.json_. What's about config writes? Say I have additional data and read the config, modify it and then save it again (using minishift get/set), does the additional data persist? This are the things I would like to see tests around to define the expected behavior. 

> I think we can add a comment on top of that file and warn user not to edit it manually and use config set command.

-1, I think it is a valid use case to edit the json file by hand.

hferentschik:
> Agree on both, you want those changes to be part of this PR or separate one. I think a separate one will make more sense.

Another pull request makes sense. Maybe we should even wait with the implementation. What we really need is this change. Whether we add the credentials can be discussed and if agreed on scheduled for another sprint.

Regarding security, it is not ideal, but neither are environment variable. To make this work, you would add the environment variables somewhere in your bashrc or user profile which is even worse. Now you have the password in clear text somewhere completely different. If we put them into _config.json_ we could even encrypt them (in a simple way). 

In the case of DevStudio using the env variables makes a bit more sense, since they are only set once and then stored in the Eclipse secure storage. So this works I guess. This is more about making the CDK more usable w/o DevStudio.

Also, whether you store the credentials in _config.json_ would be optional. The same way using environment variables is optional. Everyone can make his own decision how critical this password is.

hferentschik:
> anyone? only person who have access to that machine can view it and that even for environment variable use case where user can list set variable.

+1, using environment variables is not more secure at all.

cmd/minikube/cmd/config/validations.go
praveenkumar: > What do you mean? 

@hferentschik I meant if we use "file://" in mac or in Linux that didn't work to get file downloaded using file URI. so as per parsing concern file:/, file:// and file:/// all are valid from ParseURI method.

cmd/minikube/cmd/config/config_test.go
praveenkumar: > I'd like to now whether we can add this to config.json or whether we need a different config file.

I would say we should use a different config file and add some intelligent there to verify key-value pair which and that config can be modified by manually. Currently config.json file used by viper and I am not sure if we can add some magic there to validate what key and respective value and that's why `set` and `get` make more sense here to modify that file instead manually where validator functions are already in place.

> So if I manually add another flag (which is not listed here) to the config, I think it will be picked up fine by Viper. I think there might be some things we need to think through here.

Yes if you add manually then Viper will pick it up that's how viper works, it just read that config file  https://github.com/minishift/minishift/blob/master/cmd/minikube/cmd/root.go#L129-L138 

> What's about config writes? Say I have additional data and read the config, modify it and then save it again (using minishift get/set), does the additional data persist?

Yes, I added a set test for that it will update the key value and then make it persistent.

cmd/minikube/cmd/config/validations.go
hferentschik: > I meant if we use "file://" in mac or in Linux that didn't work to get file downloaded using file URI.

I am pretty sure this works

> so as per parsing concern file:/, file:// and file:/// all are valid from ParseURI method.

I think this is not quite correct implementation wise, but if it is like this we don't have to touch this right now.

praveenkumar: > but if it is like this we don't have to touch this right now.

Yes agree on that part, let's not touch it.

cmd/minikube/cmd/config/config_test.go
hferentschik: > Yes if you add manually then Viper will pick it up that's how viper works, it just read that config file

Exactly, so the minishift get/set is just something on top to allow the validation and setting of a specific set of flags. TBH, this bugs me a bit, since it is not consistent in behavior and it prevents me to set other flags which are not listed there. That's also why we initially struggled with understanding why Viper config file was not working as expected.

> I would say we should use a different config file and add some intelligent there to verify key-value pair which and that config can be modified by manually. 

I don't think another config file is needed. I think we can nicely solve this with a single file as well, but we can cross this bridge when we have to

> Currently config.json file used by viper and I am not sure if we can add some magic there to validate what key and respective value

We don't have to. However, if we wanted to we could.

> and that's why set and get make more sense here to modify that file instead manually where validator functions are already in place.

The get/set approach is what we are going to document for sure. 

One thing I don't like at all is that this _config set/get_ approach is not self documenting. One has no idea which values are supported. Either view should show the full MinikubeConfig applying defaults where appropriate or we need another command or something in the help which lists the available flags

cmd/minikube/cmd/config/set_test.go
```diff
@@ -16,11 +16,36 @@ limitations under the License.
...
 
 func TestNotFound(t *testing.T) {
 	err := set("nonexistant", "10")
 	if err == nil {
 		t.Fatalf("Set did not return error for unknown property")
 	}
 }
+
+func TestModifyData(t *testing.T) {
+	constants.ConfigFile = "/tmp/config.json"
```
hferentschik: This is not going to work cross platform. You really want to use ioutil.TempDir - https://github.com/minishift/minishift/blob/26a1df15a2d75ee014bce270de5bf620a46d8e4a/pkg/minishift/cache/oc_caching_test.go#L73-L73

docs/using.md
praveenkumar: I don't want to cater small details into sub-heading I think when user check config file then he should able to get all the required details.

docs/using.md
praveenkumar: In this way we have to sub-heading all the options which provided by `config`.

docs/using.md
hferentschik: We should reference to the generated help page for the config command, since there the available properties are listed. Otherwise we need to list it somewhere in here as well

hferentschik:
Closing this pull request. It will be re-opened to allow for a final review.

PR #304, opened 2017-01-17
Issue #278 Ensure major flags can be configured via persistent options on config.json

praveenkumar:
This patch will add some document around config.json and also add --iso-url to make it persistent.

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -75,6 +75,19 @@ var RootCmd = &cobra.Command{
 			}
 		}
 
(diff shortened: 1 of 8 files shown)


cmd/minikube/cmd/root.go
```diff
@@ -75,6 +75,19 @@ var RootCmd = &cobra.Command{
 			}
 		}
```
praveenkumar: We need to add this here instead L#129 because this is the place where all required directory get initialized irrespective to what command we provide.

hferentschik:
Merged. I added another commit to apply some minor refactorings and some updated to the docs. In particular the links were broken :-(

Issue #306, opened 2017-01-18
Creation of config file fails when there is no existing Minishift home directory


praveenkumar:
Due to https://github.com/minishift/minishift/blob/master/cmd/minikube/cmd/root.go#L131 inclusion minishift now break if we build from master because initconfig try to create config file to a uninitialized directory location.

```
$ ./minishift delete
F0118 12:09:05.549463   16996 root.go:164] Cannot create file /home/prkumar/.minishift/config/config.json: open /home/prkumar/.minishift/config/config.json: no such file or directory
```

hferentschik:
Merged. Thanks. Good catch.

PR #307, opened 2017-01-18
Issue #306 Master tip break for minishift

praveenkumar:
This patch will enable master tip to work again.


hferentschik:
Merged. Thanks. Good catch.

Issue #297, opened 2017-01-16
Provide ability to retrieve OpenShift URL in machine readable format.


praveenkumar:
Minishift offers currently offers:

```
minishift console --url
```

This, however, prints just the URL of OpenShift to the console. This is not in the export variable format as it used to be in CDK 2.x. For CDK-3.x we need to extend the console command to provide something like:

```
$  minishift console --machine-readable
HOST=192.168.99.110
PORT=8443
CONSOLE_URL=https://192.168.99.110:8443
```

hferentschik:
See also https://issues.jboss.org/browse/CDK-47

hferentschik:
Merged via pull request #310

PR #310, opened 2017-01-19
Issue #297 Provide ability to retrieve OpenShift URL in machine readable format

praveenkumar:
This patch will enable get machine readable data from minishift cli.

```
10:54 $ ./minikube console --url
https://192.168.99.103:8443

✔ ~/work/github/go_practice 
10:57 $ ./minikube console --machine-readable
HOST=https://192.168.99.103
PORT=8443
CONSOLE_URL=https://192.168.99.103:8443
```

### DIFF
--- a/cmd/minikube/cmd/console.go
+++ b/cmd/minikube/cmd/console.go
@@ -30,6 +30,10 @@ import (
 
 var (
 	consoleURLMode bool
+	machineReadAble bool
+	machineDetails = `HOST=%s
+PORT=%d
+CONSOLE_URL=%s`
 )
 
 // consoleCmd represents the console command
@@ -47,6 +51,12 @@ var consoleCmd = &cobra.Command{
 		}
 		if consoleURLMode {
 			fmt.Fprintln(os.Stdout, url)
+		} else if machineReadAble{
+			hostIP, err := cluster.GetHostIP(api)
+			if err != nil {
+				glog.Errorln("Cannot get Host IP. Verify that Minishift is running. Error: ", err)
+			}
+			displayConsoleInMachineReadable(hostIP, url)
 		} else {
 			fmt.Fprintln(os.Stdout, "Opening the OpenShift Web console in the default browser...")
 			browser.OpenURL(url)
@@ -54,7 +64,13 @@ var consoleCmd = &cobra.Command{
 	},
 }
 
+func displayConsoleInMachineReadable(hostIP string, url string) {
+	machineDetails = fmt.Sprintf(machineDetails, hostIP, constants.APIServerPort, url)
+	fmt.Fprintln(os.Stdout, machineDetails)
+}
+
 func init() {
 	consoleCmd.Flags().BoolVar(&consoleURLMode, "url", false, "Prints the OpenShift Web Console URL to the console.")
+	consoleCmd.Flags().BoolVar(&machineReadAble, "machine-readable", false, "Prints OpenShift's IP, port and Web Console URL in Machine readable format")
 	RootCmd.AddCommand(consoleCmd)
 }
--- a/cmd/minikube/cmd/console_test.go
+++ b/cmd/minikube/cmd/console_test.go
@@ -1,0 +1,58 @@
+/*
+Copyright (C) 2016 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"testing"
+	"os"
+	"io/ioutil"
+	"strings"
+)
+
+func TestDisplayConsoleInMachineReadable(t *testing.T) {
+	testDir, err := ioutil.TempDir("", "minishift-config-")
+	if err != nil {
+		t.Error()
+	}
+	defer os.RemoveAll(testDir)
+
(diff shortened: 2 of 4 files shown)


hferentschik:
> HOST=https://192.168.99.103

I think we said, this should be the host, aka IP, not a URL

cmd/minikube/cmd/console.go
```diff
@@ -47,14 +52,24 @@ var consoleCmd = &cobra.Command{
...
+			fmt.Fprintln(os.Stdout, machineDetails)
 		} else {
 			fmt.Fprintln(os.Stdout, "Opening the OpenShift Web console in the default browser...")
 			browser.OpenURL(url)
 		}
 	},
 }
 
+func getHostAndPort(consoleUrl string) (string, string) {
+	return strings.Join(strings.Split(consoleUrl, ":")[:2], ":"),
```
hferentschik: Personally I feel this is a bit cryptic. I'd suggest to add some comments on what you do or write this in a bit more self descriptive way. 

Also, this seems to fall short on what we defined in the issue. The host should just be the IP, not a URL.

cmd/minikube/cmd/console.go
```diff
@@ -47,14 +52,24 @@ var consoleCmd = &cobra.Command{
...
+		strings.Split(consoleUrl, ":")[2]
+}
+
 func init() {
 	consoleCmd.Flags().BoolVar(&consoleURLMode, "url", false, "Prints the OpenShift Web Console URL to the console.")
+	consoleCmd.Flags().BoolVar(&machineReadAble, "machine-readable", false, "Prints OpenShift Web Console details in Machine readable format.")
```
hferentschik: "Prints OpenShift's IP, port and Web Console URL in Machine readable format."

cmd/minikube/cmd/console.go
```diff
@@ -47,14 +52,24 @@ var consoleCmd = &cobra.Command{
 		}
 		if consoleURLMode {
 			fmt.Fprintln(os.Stdout, url)
+		} else if machineReadAble{
```
hferentschik: Testing would get much easier if you take as much as possible out of 'func(cmd *cobra.Command, args []string) '. Split this up into multiple functions. For example split this into two methods. One to get the URL using libmachine client and _cluster.GetConsoleURL_. This part won't be testable unless you mock. Second method is _displayConsoleInformation(url string)_. This just gets the URL and is concerned about putting this to screen in the right format. You can test the right format gets printed now without mocking anything. You can test _displayConsoleInformation_. Just capture standard out and make sure the result is correct.

cmd/minikube/cmd/console.go
```diff
@@ -30,6 +30,10 @@ import (
 
 var (
 	consoleURLMode bool
+	machineReadAble bool
```
hferentschik: +1

cmd/minikube/cmd/console.go
```diff
@@ -47,14 +51,26 @@ var consoleCmd = &cobra.Command{
 		}
 		if consoleURLMode {
 			fmt.Fprintln(os.Stdout, url)
+		} else if machineReadAble{
+			hostIP, err := cluster.GetHostIP(api)
```
hferentschik: Design wise, I find it odd that in this method you read the url above the if/switch and the ip in this sub-branch. But never mind

cmd/minikube/cmd/console_test.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+func TestDisplayConsoleInMachineReadable(t *testing.T) {
+	testDir, err := ioutil.TempDir("", "minishift-config-")
+	if err != nil {
+		t.Error()
+	}
+	defer os.RemoveAll(testDir)
+
+	f, err := os.Create(testDir + "out.txt")
```
hferentschik: This works :-) Writing standard out to file and re-read. Even nicer imo is to use a Pipe. Then you need no additional file.

cmd/minikube/cmd/console_test.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		t.Fatalf("Error creating test file", err)
```
hferentschik: If you don't interpolate something into the string, _t.Fatal_ is better. If nothing else your IDE shuts up ;-)

pkg/minikube/cluster/cluster.go
```diff
@@ -402,6 +402,19 @@ func GetConsoleURL(api libmachine.API) (string, error) {
 	return fmt.Sprintf("https://%s:%d", ip, constants.APIServerPort), nil
 }
 
+func GetHostIP(api libmachine.API) (string, error) {
```
hferentschik: +1

cmd/minikube/cmd/console.go
```diff
@@ -47,14 +51,26 @@ var consoleCmd = &cobra.Command{
...
+			displayConsoleInMachineReadable(hostIP, url)
 		} else {
 			fmt.Fprintln(os.Stdout, "Opening the OpenShift Web console in the default browser...")
 			browser.OpenURL(url)
 		}
 	},
 }
 
+func displayConsoleInMachineReadable(hostIP string, url string) {
+	machineDetails = fmt.Sprintf(machineDetails, hostIP, constants.APIServerPort, url)
```
hferentschik: I actually found it quite interesting the other day as well, that the port is a constant. I guess there is no big need to change it, but somehow I expected the port to be configurable and then retrievable in some way.

cmd/minikube/cmd/console_test.go
```diff
@@ -0,0 +1,58 @@
...
+		t.Fatalf("Error setting offset back", err)
+	}
+
+	data, err := ioutil.ReadAll(f)
+	if err != nil {
+		t.Fatalf("Error reading file", err)
+	}
+
+	if strings.TrimSpace(string(data)) != expectedStdout {
+		t.Fatalf("expected hello got %s", string(data))
```
hferentschik: I guess 'hello' is a typo

hferentschik:
Merged, thanks

PR #311, opened 2017-01-19
Issue #173 Setup CI build on CentOS CI

### LINKED ISSUE #173: Setup CI build on CentOS CI
Using CentOS CI will give us nested virtualization, so that we will be able to run proper integration tests. See also:

* https://wiki.centos.org/QaWiki/CI/Duffy
* https://github.com/projectatomic/vagrant-service-manager/tree/master/.ci

budhrg:
Fix #173

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -1,0 +1,70 @@
+#!/bin/bash
+
+# Copyright (C) 2016 Red Hat, Inc.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Output command before executing
+set -x
+
+# Exit on error
+set -e
+
+# Source environment variables of the jenkins slave
+# that might interest this worker.
+if [ -e "jenkins-env" ]; then
+  cat jenkins-env \
+    | grep -E "(JENKINS_URL|GIT_BRANCH|GIT_COMMIT|BUILD_NUMBER|ghprbSourceBranch|ghprbActualCommit|BUILD_URL|ghprbPullId)=" \
+    | sed 's/^/export /g' \
+    > ~/.jenkins-env
+  source ~/.jenkins-env
+fi
+
+# We need to disable selinux for now, XXX
+/usr/sbin/setenforce 0
+
+# Get all the deps in
+yum -y install \
+  gcc \
+  make \
+  git \
+  curl
+
+# Install Go 1.7
+curl -LO https://storage.googleapis.com/golang/go1.7.linux-amd64.tar.gz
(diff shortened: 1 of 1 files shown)


hack/boilerplate/boilerplate.py
```diff
@@ -105,7 +105,7 @@ def file_passes(filename, refs, regexs):
 def file_extension(filename):
     return os.path.splitext(filename)[1].split(".")[-1].lower()
 
-skipped_dirs = ['Godeps', 'third_party', '_gopath', '_output', '.git', 'cluster/env.sh', "vendor", "test/e2e/generated/bindata.go"]
+skipped_dirs = ['Godeps', 'third_party', '_gopath', '_output', '.git', 'cluster/env.sh', "vendor", "test/e2e/generated/bindata.go", "centos_ci.sh"]
```
hferentschik: Why don't we just add the required boilerplate code?

centos_ci.sh
```diff
@@ -0,0 +1,59 @@
...
+mkdir $HOME/gopath $HOME/gopath/src $HOME/gopath/bin $HOME/gopath/pkg
+export GOPATH=$HOME/gopath
+export PATH=$GOROOT/bin:$GOPATH/bin:$PATH
+
+echo $GOPATH
+echo $GOROOT
+
+# Setup minishift repo
+mkdir -p $GOPATH/src/github.com/minishift/minishift
+cp -r `ls` $GOPATH/src/github.com/minishift/minishift/
```
hferentschik: instead of copying a symbolic link should do

centos_ci.sh
```diff
@@ -0,0 +1,59 @@
...
+cd $GOPATH/src/github.com/minishift/minishift/
+
+# Glide setup
+GLIDE_OS_ARCH=`go env GOHOSTOS`-`go env GOHOSTARCH`
```
hferentschik: Instead of all this stuff below have you tried: "curl https://glide.sh/get | sh"

budhrg: @hferentschik  Looks like Github API rate limit hit here too https://github.com/Masterminds/glide/issues/708

budhrg:
centos ci is passing here, with travis ci failed due to openshift oc binary github api rate limit.

cc @hferentschik

centos_ci.sh
```diff
@@ -0,0 +1,70 @@
...
+GLIDE_TAG=$(curl -s https://glide.sh/version)
+GLIDE_LATEST_RELEASE_URL="https://github.com/Masterminds/glide/releases/download/${GLIDE_TAG}/glide-${GLIDE_TAG}-${GLIDE_OS_ARCH}.tar.gz"
+curl -LO ${GLIDE_LATEST_RELEASE_URL}
+mkdir /tmp/glide
+tar --directory=/tmp/glide -xvf glide-${GLIDE_TAG}-${GLIDE_OS_ARCH}.tar.gz
+export PATH=$PATH:/tmp/glide/${GLIDE_OS_ARCH}
+
+# Test
+make test cross fmtcheck prerelease
```
hferentschik: Looking good. I was wondering whether you could look into one more thing though :-) The value proposition of using CentOS CI is to be able to run integration tests (under the assumption that we have nested virtualization on CentOS CI). To confirm this we could run the current integration tests as well.

To make this work you would need to add the _integration_ task here. But you also need to either provision VirtualBox (something I would like to investigate as well) or probably easier make a change to _flags.go_ to read the driver from say an environment variable and if not set defaults to _virtualbox_.

PR #315, opened 2017-01-23
Fix #302 minishift logs doesn't work

### LINKED ISSUE #302: "minishift logs" doesn't work
"minishift logs" with 1.0.0-beta-2 always returns a single empty line. 

This worked with "minishift 0.9.0" prefectly, also with setting the log level.

I'm running minishift on OS X with xhyve.

praveenkumar:
I didn't edited https://github.com/minishift/minishift/blob/master/pkg/minikube/cluster/commands.go#L49 because eventually we can get rid of this file completely.

### DIFF
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -47,7 +47,7 @@ import (
 )
 
 var (
-	certs = []string{"apiserver.crt", "apiserver.key"}
+	logsCmd = "docker logs origin"
 )
 
 const (
@@ -343,7 +343,7 @@ func GetHostLogs(api libmachine.API) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	s, err := host.RunSSHCommand(logsCommand)
+	s, err := host.RunSSHCommand(logsCmd)
 	if err != nil {
 		return "", nil
 	}


pkg/minikube/cluster/cluster.go
```diff
@@ -48,6 +48,7 @@ import (
 
 var (
 	certs = []string{"apiserver.crt", "apiserver.key"}
+	logsCmd = "docker logs origin"
```
hferentschik: Will this work with OCP as well? Is the container there also named 'origin'? Or do we some smarter way to determine the right container?

pkg/minikube/cluster/cluster.go
```diff
@@ -48,6 +48,7 @@ import (
 
 var (
 	certs = []string{"apiserver.crt", "apiserver.key"}
```
hferentschik: I think we can get rid of this. Can you have a look?

pkg/minikube/cluster/cluster.go
praveenkumar: @hferentschik That was the first check which I did before this PR :) yes it have same name as `origin` there.

pkg/minikube/cluster/cluster.go
praveenkumar: @hferentschik sure I will take a look and see if this is consumed somewhere.

pkg/minikube/cluster/cluster.go
hferentschik: +1 :-) Otherwise we could do some name independent things via the container metadata. Eventually I think we need something like this. I think I mentioned that I'd like to build an abstraction layer in Minishift to interact and control the Docker daemon in the VM. Something which is a bit "smarter" than executing plain Docker commands. 

However, if the name of the container is always 'origin' we can go for this approach.

BTW, would there be an issue with exposing the '-f' flag as well to allow following the log?

pkg/minikube/cluster/cluster.go
praveenkumar: @hferentschik Checked and it's not used anywhere in codebase now.

praveenkumar:
> I think there is no reason for keeping commands.go. AFAICT we can get rid of it.

@hferentschik Can we just create another issue for that clean up stuff because variables which are used in this file also no logger used and can be removed from constants.go. It will make our git history little bit clear, wdyt?

praveenkumar:
@hferentschik I checked `--follow` option and it will not work because `s, err := host.RunSSHCommand(LogsCmd)` actually return the command output once command is success and --follow will block that command without returning anything. So let's we go with this right now.

hferentschik:
>  I checked --follow option and it will not work because s, err := host.RunSSHCommand(LogsCmd) actually return the command output once command is success and --follow will block that command without returning anything

Got you. Thanks for investigating this. I think in this case we maybe just create a new issue. It's something which would be quite nice imo, but also something which has a low priority right now.

>  So let's we go with this right now.

+1

praveenkumar:
@hferentschik https://github.com/minishift/minishift/issues/319 follow up issue created for clean up part so I think we can merge this and then will work on clean up part.

praveenkumar:
> I think in this case we maybe just create a new issue.

https://github.com/minishift/minishift/issues/320 this will track for `--follow` stuff.

hferentschik:
Merged, thanks.

Issue #319, opened 2017-01-23
Delete commands.go and related test file


praveenkumar:
Since we moved ahead with `oc cluster up` approach now so better to clean up older way of running minishift which was using openshift binary. Also we should clean up unused variables from different file.

hferentschik:
This is a follow up of issue #302

hferentschik:
Resolved via pull request #319

PR #321, opened 2017-01-23
Issue #319 Removing some obsolete code

hferentschik:


pkg/minikube/cluster/cluster.go
```diff
@@ -584,7 +584,7 @@ func GetServiceURLs(api libmachine.API, namespace string, t *template.Template)
 }
 
 func GetCA(h sshAble) (string, error) {
```
praveenkumar: I think we can remove this method completely because I didn't find this is used anywhere in the codebase now.

hferentschik: I know, I thought about this as well. It is not used right now, but it actually works and I thought it could be useful at some stage. But I guess when we have the need for it, we can get it back. ATM, I am also not seeing the exact use case where we need it. 

I'll remove it

Issue #322, opened 2017-01-23
Upgrade to latest OpenShift version


hferentschik:
Upgrade to use OpenShift 1.4.0 pre default.

hferentschik:
Resolved via pull request #323

PR #323, opened 2017-01-23
Issue #322 Upgrading to OpenShift 1.4.0

hferentschik:


hferentschik:
retest this please

docs/minishift_start.md
```diff
@@ -27,7 +27,7 @@ minishift start
       --memory int                 Amount of RAM to allocate to the Minishift VM. (default 2048)
       --metrics                    Install metrics (experimental)
   -e, --openshift-env value        Specify key-value pairs of environment variables to set on the OpenShift container. (default [])
-      --openshift-version string   The OpenShift version to run, eg. v1.3.1 (default "v1.3.1")
+      --openshift-version string   The OpenShift version to run, eg. v1.3.1 (default "v1.4.0")
```
budhrg: Why not update the `eg` piece to `1.4.0` ?

praveenkumar: @budhrg it just a place holder to tell user how can define the version tag. we can update to 1.4.0 and leave it as it is also.

praveenkumar:
@hferentschik looks like appveyor is stuck, can you check and manually trigger build for same?

hferentschik:
> looks like appveyor is stuck, can you check and manually trigger build for same?

It's odd, I tried, but I don't even seem to see the matching pull request build. Maybe it recovers when I repush the branch ...

docs/minishift_start.md
hferentschik: Right, so it really does not matter, but I agree it is a bit odd. Maybe we should interpolate the the version specified into the Makefile, twice. So it effectively mentions the same version twice. Might be better though!? @budhrg @praveenkumar - WDYT?

 If would be

     "The OpenShift version to run, eg. v1.4.0 (default "v1.4.0")"

And if we change the version number it would be for example:

    The OpenShift version to run, eg. v1.5.0 (default "v1.5.0")

PR #332, opened 2017-01-25
Issue #331 Upgrade to openshift 1.4.0+ breaks downloading caching 'oc'

### LINKED ISSUE #331: Upgrade to openshift 1.4.0+ breaks downloading caching 'oc' binary
It looks like regression of upgrading to `1.4.0`, `minishift start` fails on caching `oc` binary.

```
minishift start --vm-driver virtualbox
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.0/minishift-b2d.iso'
 40.00 MB / 40.00 MB [============================================================================================================================] 100.00% 0s
 22.37 MB / 22.37 MB [============================================================================================================================] 100.00% 0s
E0125 10:46:57.786410   26875 start.go:219] Error starting the cluster:  Error attempting to download and cache oc: Cannot untar /tmp/minishift-asset-download-280664917/openshift-origin-client-tools-v1.4.0-208f053-linux-64bit.tar: open /tmp/minishift-asset-download-280664917/openshift-origin-client-tools-v1.4.0+208f053-linux-64bit/oc: no such file or directory
```

Found that now from 1.4.1 onwards, tar directory `openshift-origin-client-tools-v1.4.0+208f053-linux-64bit` includes `+` instead of `-` which is not the case in windows or mac.

Might be regression from origin itself and filing issue there.

budhrg:
Fix #331

### DIFF
--- a/cmd/minikube/cmd/start_test.go
+++ b/cmd/minikube/cmd/start_test.go
@@ -46,7 +46,7 @@ func (r *RecordingRunner) Run(command string, args ...string) error {
 var testDir string
 var testRunner *RecordingRunner
 var testMachineConfig = cluster.MachineConfig{
-	OpenShiftVersion: "v1.3.1",
+	OpenShiftVersion: tests.OPENSHIFT_VERSION,
 }
 
 func TestStartClusterUpNoFlags(t *testing.T) {
--- a/pkg/minikube/tests/constants.go
+++ b/pkg/minikube/tests/constants.go
@@ -1,0 +1,20 @@
+/*
+Copyright (C) 2016 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+// Version to be tested
+const OPENSHIFT_VERSION = "v1.4.1"
--- a/pkg/minishift/cache/oc_caching_test.go
+++ b/pkg/minishift/cache/oc_caching_test.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 	"testing"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minikube/tests"
 )
 
 var testDir string
@@ -33,7 +34,7 @@ func TestIsCached(t *testing.T) {
 	setUp(t)
 	defer os.RemoveAll(testDir)
 
-	ocDir := filepath.Join(testDir, "cache", "oc", "v1.3.1")
+	ocDir := filepath.Join(testDir, "cache", "oc", tests.OPENSHIFT_VERSION)
 	os.MkdirAll(ocDir, os.ModePerm)
 
 	if testOc.isCached() != false {
@@ -59,7 +60,7 @@ func TestCacheOc(t *testing.T) {
 	client.Transport = minitesting.NewMockRoundTripper()
 	defer minitesting.ResetDefaultRoundTripper()
 
-	ocDir := filepath.Join(testDir, "cache", "oc", "v1.3.1")
+	ocDir := filepath.Join(testDir, "cache", "oc", tests.OPENSHIFT_VERSION)
 	os.MkdirAll(ocDir, os.ModePerm)
 
 	err := testOc.cacheOc()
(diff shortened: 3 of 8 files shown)


pkg/util/github/github.go
```diff
@@ -190,7 +190,9 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
 	binaryPath := ""
 	switch {
 	case strings.HasSuffix(assetTmpFile, TAR):
-		tarFile := assetTmpFile[:len(assetTmpFile)-3]
+		tokens := strings.Split(assetFilename, fmt.Sprintf("%s-", version))
```
hferentschik: I am reluctant to put anything like this in for now. Let's first wait what the response is from the OpenShift side.

If we need to handle this ourselves, I think we need to write some code which is a bit  better to understand than this.

hferentschik: Instead of hacking directory names, we can work around this by just unpacking the tarball and reading the actual created directory from the file system. Much more resilient. This will work regardless of what Origin will do, since it won't be based on assumptions.

hferentschik: This change also seems to break some tests.

budhrg:
Looks like CI is also passing now except travis ci failing due to github rate limit.

cmd/minikube/cmd/start_test.go
```diff
@@ -46,7 +46,7 @@ func (r *RecordingRunner) Run(command string, args ...string) error {
 var testDir string
 var testRunner *RecordingRunner
 var testMachineConfig = cluster.MachineConfig{
-	OpenShiftVersion: "v1.3.1",
+	OpenShiftVersion: tests.OPENSHIFT_VERSION,
```
hferentschik: -1, you cannot just upgrade the version. Now you broke it for 1.3.1. Not saying that it makes much sense to use an older version of OpenShift, but we don't have to break it intentionally either.

pkg/minikube/tests/constants.go
```diff
@@ -0,0 +1,20 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package tests
+
+// Version to be tested
+const OPENSHIFT_VERSION = "v1.4.1"
```
hferentschik: -1 for a test constants. What do you want to achieve here?

pkg/minishift/cache/oc_caching_test.go
```diff
@@ -33,7 +34,7 @@ func TestIsCached(t *testing.T) {
 	setUp(t)
 	defer os.RemoveAll(testDir)
 
-	ocDir := filepath.Join(testDir, "cache", "oc", "v1.3.1")
+	ocDir := filepath.Join(testDir, "cache", "oc", tests.OPENSHIFT_VERSION)
```
hferentschik: again, you make it work by changing the base version. The test would still fail with 1.3.1 now right?

pkg/util/github/github.go
```diff
@@ -190,7 +190,16 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
 	binaryPath := ""
 	switch {
 	case strings.HasSuffix(assetTmpFile, TAR):
-		tarFile := assetTmpFile[:len(assetTmpFile)-3]
+		tarFile := ""
+		if osType == minishiftos.LINUX {
+			// https://github.com/openshift/origin/issues/12663
```
hferentschik: -1, this is just putting a bandage on the problem. It break backwards compatibility, is heard to understand and does not tackle the problem in its core. As said, we should be a bit smarter here and don't rely on the what we think is in the tar ball.

pkg/util/github/github_test.go
```diff
@@ -37,18 +38,17 @@ var (
...
 	os               minishiftos.OS
 	version          string
 	expectedAssetId  int
 	expectedFilename string
 }{
-	{OPENSHIFT, minishiftos.LINUX, testVersion, 2489309, "openshift-origin-server-v1.3.1-dad658de7465ba8a234a4fb40b5b446a45a4cee1-linux-64bit.tar.gz"},
-	{OC, minishiftos.LINUX, testVersion, 2489310, "openshift-origin-client-tools-v1.3.1-dad658de7465ba8a234a4fb40b5b446a45a4cee1-linux-64bit.tar.gz"},
-	{OC, minishiftos.DARWIN, testVersion, 2586147, "openshift-origin-client-tools-v1.3.1-2748423-mac.zip"},
-	{OC, minishiftos.WINDOWS, testVersion, 2489312, "openshift-origin-client-tools-v1.3.1-dad658de7465ba8a234a4fb40b5b446a45a4cee1-windows.zip"},
+	{OPENSHIFT, minishiftos.LINUX, tests.OPENSHIFT_VERSION, 3053311, "openshift-origin-server-v1.4.1-3f9807a-linux-64bit.tar.gz"},
```
hferentschik: I really like the existing tests to keep passing

pkg/util/github/github.go
hferentschik: What if https://github.com/openshift/origin/issues/12576 gets "fixed" and the filename contains '-' again, then we have to adjust the code again. This will result in versions where our code will break and versions where it will work.

hferentschik:
Superseded by pull request #333

pkg/util/github/github.go
budhrg: Make sense :+1:

PR #333, opened 2017-01-25
Issue #331 Determining directory names within archive dynamically

hferentschik:
Addresses issue #331

### DIFF
--- a/pkg/util/archive/archive.go
+++ b/pkg/util/archive/archive.go
@@ -49,11 +49,6 @@ func Ungzip(source, target string) error {
 }
 
 func Untar(tarball, targetDir string) error {
-	err := os.MkdirAll(tarball[:len(tarball)-4], 0755)
-	if err != nil {
-		return err
-	}
-
 	reader, err := os.Open(tarball)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


pkg/util/github/github.go
```diff
@@ -235,6 +248,28 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
 	return nil
 }
 
+func listDirExcluding(dir string, excludeRegexp string) ([]string, error) {
```
budhrg: `func listDirExcluding(dir, excludeRegexp string)` format here?

pkg/util/archive/archive.go
```diff
@@ -70,15 +65,13 @@ func Untar(tarball, targetDir string) error {
...
 		path := filepath.Join(targetDir, header.Name)
-		info := header.FileInfo()
-		if info.IsDir() {
-			if err = os.MkdirAll(path, info.Mode()); err != nil {
-				return err
-			}
-			continue
+
+		// tar.Next() will externally only iterate files, so we might have to create intermediate directories here
+		if err = os.MkdirAll(filepath.Dir(path), 0770); err != nil {
```
budhrg: :+1:

budhrg:
LGTM :+1: if you take care one best practice related comment :wink:

pkg/util/github/github.go
hferentschik: format? what do you mean?

praveenkumar: @budhrg I also don't understand what do you mean format here. function definition seems good to me.

budhrg: @hferentschik @praveenkumar , I mean as per Go, if you have same datatype of parameters in the function(like for `dir` and `excludeRegexp` have `string` as datatype) then you keep the datatype of leftmost parameter and omit rest as : 
```
func listdirExcluding(dir, excludeRegexp string) ([]string, error) {
....
}
```

Its a common idiom in Go.
Sorry I didn't elaborate my comment previously.
```

hferentschik: hmm, one can for sure do so. I am not sure whether this is part of best practices or idiomatic in Go. Personally I am not a big fan.

Issue #334, opened 2017-01-27
Updates needed for CentOS custom provisioner to accomodate changes for latest Docker version


praveenkumar:
We need to update https://github.com/minishift/minishift/blob/master/pkg/minishift/provisioner/utils.go#L65 to accommodate latest changes of docker-1.12 because now we also need to have docker-runc-current as daemon option to work with it. Below snip is what system unit file provide right now with update docker package.

```
ExecStart=/usr/bin/dockerd-current \
          --add-runtime docker-runc=/usr/libexec/docker/docker-runc-current \
          --default-runtime=docker-runc \
          --exec-opt native.cgroupdriver=systemd \
          --userland-proxy-path=/usr/libexec/docker/docker-proxy-current \
          $OPTIONS \
          $DOCKER_STORAGE_OPTIONS \
          $DOCKER_NETWORK_OPTIONS \
          $ADD_REGISTRY \
          $BLOCK_REGISTRY \
          $INSECURE_REGISTRY
```

Depend on: https://github.com/minishift/minishift-centos-iso/issues/89

hferentschik:
Merged via pull request #335

PR #335, opened 2017-01-27
Issue #334 Updates needed for CentOS custom provisioner to accomodate changes for latest Docker version

praveenkumar:
This patch adds addition options which required for latest docker for CentOS - issue #334


hferentschik:
Merged, thanks

PR #341, opened 2017-01-30
Issue #339 Fix 'minishift ssh' help message

### LINKED ISSUE #339: "minishift ssh" doesn't behave like "docker-machine ssh" as mentioned in doc
As per `minishift ssh` help, it should behave like [`docker-machine ssh`](https://docs.docker.com/machine/reference/ssh/) where one can run command on VM.

```
$ minishift ssh df -h
Log in to or run a command on a Minishift VM with SSH. This command is similar to 'docker-machine ssh'.

Usage:
  minishift ssh [flags]

Global Flags:
      --alsologtostderr value          log to standard error as well as files
      --log-flush-frequency duration   Maximum number of seconds between log flushes (default 5s)
      --log_backtrace_at value         when logging hits line file:N, emit a stack trace (default :0)
      --log_dir value                  If non-empty, write log files in this directory
.....
```

The help usage mention that one can use command on a Minishift VM which seems not working at the moment.

Looking at the ["minishift" code for `ssh` command](https://github.com/minishift/minishift/blob/master/cmd/minikube/cmd/ssh.go#L35-L41), it looks like it doesn't provide any mechanism to support it.

@jimmidyson Please confirm.

budhrg:
Fix #339 

Now, we can run `minishift ssh 'COMMAND'` or `minishift ssh -- COMMAND`:

```
minishift ssh 'df -h'
Filesystem                Size      Used Available Use% Mounted on
tmpfs                     1.8G    196.6M      1.6G  11% /
tmpfs                  1001.0M    172.0K   1000.9M   0% /dev/shm
/dev/sda1                17.9G    785.1M     16.2G   5% /mnt/sda1
cgroup                 1001.0M         0   1001.0M   0% /sys/fs/cgroup
/dev/sda1                17.9G    785.1M     16.2G   5% /mnt/sda1/var/lib/docker/aufs
tmpfs                     1.8G    196.6M      1.6G  11% /var/lib/origin/openshift.local.volumes
.....
```
### Or
```
minishift ssh -- df -h                             
Filesystem                Size      Used Available Use% Mounted on
tmpfs                     1.8G    196.6M      1.6G  11% /
tmpfs                  1001.0M    172.0K   1000.9M   0% /dev/shm
/dev/sda1                17.9G    785.1M     16.2G   5% /mnt/sda1
cgroup                 1001.0M         0   1001.0M   0% /sys/fs/cgroup
/dev/sda1                17.9G    785.1M     16.2G   5% /mnt/sda1/var/lib/docker/aufs
tmpfs                     1.8G    196.6M      1.6G  11% /var/lib/origin/openshift.local.volumes
........
```

### DIFF
--- a/cmd/minikube/cmd/ssh.go
+++ b/cmd/minikube/cmd/ssh.go
@@ -28,7 +28,7 @@ import (
 
 // sshCmd represents the docker-machine ssh command
 var sshCmd = &cobra.Command{
-	Use:   "ssh",
+	Use:   "ssh [-- COMMAND]",
 	Short: "Log in to or run a command on a Minishift VM with SSH.",
 	Long:  "Log in to or run a command on a Minishift VM with SSH. This command is similar to 'docker-machine ssh'.",
 	Run: func(cmd *cobra.Command, args []string) {
--- a/docs/minishift_ssh.md
+++ b/docs/minishift_ssh.md
@@ -8,7 +8,7 @@ Log in to or run a command on a Minishift VM with SSH.
 Log in to or run a command on a Minishift VM with SSH. This command is similar to 'docker-machine ssh'.
 
 ```
-minishift ssh
+minishift ssh [-- COMMAND]
 ```
 
 ### Options inherited from parent commands


cmd/minikube/cmd/ssh.go
```diff
@@ -28,7 +28,7 @@ import (
 
 // sshCmd represents the docker-ssh command
 var sshCmd = &cobra.Command{
-	Use:   "ssh",
+	Use:   "ssh ['COMMAND' or -- COMMAND]",
```
hferentschik: I don't think this is quite right. Single and double quotes make a difference and depending what you want are valid as well. Using single quotes will not interpolate anything into the command which might or might not be what you want - http://www.howtogeek.com/howto/29980/whats-the-difference-between-single-and-double-quotes-in-the-bash-shell. It will also depend on the shell type. I think the best we can do it generally use [command].

budhrg: @hferentschik 

so can we use [ -- COMMAND ]  ? 

Here, main issue is when we try to run command containing `-` inside the vm as
`minishift ssh df -h`. Here, cobra treat `-h` as the root level help and hence print help instead of actual command.

So, here we need something to escape that and looks like cobra allow `-- COMMAND` approach.
Check this https://github.com/spf13/cobra/blob/master/cobra_test.go#L128.

I actually added quote since it was working locally for me but may not work in different shell as you mentioned.

WDYT ?

hferentschik:
retest this please

cmd/minikube/cmd/ssh.go
```diff
@@ -28,7 +28,7 @@ import (
 
 // sshCmd represents the docker-machine ssh command
 var sshCmd = &cobra.Command{
-	Use:   "ssh",
+	Use:   "ssh [-- COMMAND]",
```
hferentschik: hmm, this is one way of doing this, but depending on the command you want to run the '--' might or might not be needed.

hferentschik:
Merged, thanks.

PR #354, opened 2017-02-01
[Not to be Merged] DEBUG CI

budhrg:


hferentschik:
@budhrg Git Hub token not working?

praveenkumar:
@budhrg what you wanted to test here I still don't see any token value from CI job's log?

budhrg:
> @budhrg what you wanted to test here I still don't see any token value from CI job's log?

Me too, might be wrong way. Any idea?

hferentschik:
> Me too, might be wrong way. Any idea?

I'd like to add something to the tests which makes a request to the GitHub API and then inspects the results of whether the request was authenticated or not. Just to see what's going on.

Also adding some trace in this method could help - https://github.com/minishift/minishift/blob/56a5723e3c3d3ebea82fe8029b5658a62d5238f8/pkg/util/github/github.go

Basically I would like to know whether a token is found or not. 

We need to narrow down what's going on. Do the CI environments screw up the injection of the variables or is something wrong in the code?

budhrg:
@hferentschik , 
> Basically I would like to know whether a token is found or not.

Yes added few debug lines and checking again.

hferentschik:
@budhrg can we close this for now. Or do you still need this around?

budhrg:
We can close this now and we found its resolution.

Issue #298, opened 2017-01-16
Unset all DOCKER_* variables as part of initial setup of minishift


praveenkumar:
Currently if we already have docker env set on the same shell session where we are running `minishift start <options>` it fails  because minishift consume those available variable to do the initial check for docker daemon configuration and failed with below error.

```
Checking Docker daemon configuration ... FAIL
   Error: Error response from daemon: client is newer than server (client API version: 1.23, server API version: 1.22)
E0113 12:49:08.408930    2148 start.go:246] Error starting 'cluster up':  exit status 1
```

Another improvement should be done in same direction #252

praveenkumar:
@LalatenduMohanty Have a look to https://github.com/minishift/minishift/blob/master/cmd/minikube/cmd/env.go#L158-L197 it have complete logic around this.

LalatenduMohanty:
I am not able to reproduce the issue now. I was able to reproduce it once when I assigned this to me but can not anymore. Looking in to what has changed.

LalatenduMohanty:
Still not able to reproduce. Took a boot2docker ISO version of 1.12.2 and created the docker env with the version. But minishift starts fine. 
```
$ docker version
Client:
 Version:      1.13.0
 API version:  1.24 (downgraded from 1.25)
 Go version:   go1.7.3
 Git commit:   49bf474
 Built:        Wed Jan 18 16:20:26 2017
 OS/Arch:      darwin/amd64

Server:
 Version:      1.12.2
 API version:  1.24 (minimum version )
 Go version:   go1.6.3
 Git commit:   bb80604
 Built:        Tue Oct 11 17:00:50 2016
 OS/Arch:      linux/amd64
 Experimental: false
```

```
Lalatendus-MacBook-Pro:1.12.2 lmohanty$ minishift start
Starting local OpenShift cluster using 'xhyve' hypervisor...
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.0/minishift-b2d.iso'
 40.00 MB / 40.00 MB [======================================================================================================================================================================] 100.00% 0s
 21.34 MB / 21.34 MB [======================================================================================================================================================================] 100.00% 0s
Provisioning OpenShift via '/Users/lmohanty/.minishift/cache/oc/v1.4.1/oc [cluster up --use-existing-config --host-config-dir /var/lib/minishift/openshift.local.config --host-data-dir /var/lib/minishift/hostdata]'
-- Checking OpenShift client ... OK
[...]

PR #360, opened 2017-02-02
Issue #298 Removing docker environment variables

LalatenduMohanty:
Removing existing docker environment variables before
setting them by minishift.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -113,6 +113,11 @@ func runStart(cmd *cobra.Command, args []string) {
 	libMachineClient := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer libMachineClient.Close()
 
+	if _, err := shellCfgUnset(libMachineClient); err != nil {
+		glog.Errorln("Error unsetting existing docker environment variables:", err)
+		os.Exit(1)
+	}
+
 	config := cluster.MachineConfig{
 		MinikubeISO:      viper.GetString(isoURL),
 		Memory:           viper.GetInt(memory),


cmd/minikube/cmd/start.go
```diff
@@ -113,6 +113,12 @@ func runStart(cmd *cobra.Command, args []string) {
 	libMachineClient := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer libMachineClient.Close()
 
+	_, err := shellCfgUnset(libMachineClient)
+	if err != nil {
+		glog.Errorln("Error unsetting existing docker environment variables:", err)
+                os.Exit(1)
```
budhrg: extra space before `os.Exit(1)`

praveenkumar: @LalatenduMohanty better to do in a single line so you don't have to change below anything.

```
if _, err := shellCfgUnset(libMachineClient); err != nil {
        glog.Errorln("Error unsetting existing docker environment variables:", err)
        os.Exit(1)
}
```

LalatenduMohanty: @praveenkumar @budhrg Ack.

budhrg: Also, now `err` is local to that `if` block and will not be available IMO.

cmd/minikube/cmd/start.go
```diff
@@ -113,6 +113,11 @@ func runStart(cmd *cobra.Command, args []string) {
 	libMachineClient := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer libMachineClient.Close()
 
+	if _, err := shellCfgUnset(libMachineClient); err != nil {
```
budhrg: :+1:

budhrg:
LGTM :+1:

cmd/minikube/cmd/start.go
hferentschik: -1, this does not do anything. All you get back is a reference to ShellConfig. There is nothing happening here. 

So either we really unset the env variables via os.Unsetenv or we don't do anything.

hferentschik:
> Removing existing docker environment variables before setting them by minishift.

That's not what this code does.

praveenkumar:
It will be my mistake again I put that snip to the issue and @LalatenduMohanty used it, I didn't look closely about it's return value (which is shellcfg template and need to be processed). We should've a unit test case for same which is easy I think

budhrg:
Need rebase to leverage https://github.com/minishift/minishift/issues/362

hferentschik:
> It will be my mistake again I put that snip to the issue and @LalatenduMohanty used it, I didn't look closely about it's return value

we need to actually check the docs resp code what something is doing ;-)

> which is shellcfg template and need to be processed

that's not going to help. This methods gives you some unset commands based on OS and shell, but we want to unset for the process which is going to run cluster up, right? We don't want to dump an eval command here, right? Unless there is a misunderstanding here on what we want to do. 

What is it you want to do?

LalatenduMohanty:
@hferentschik @praveenkumar I misunderstood the code.  I am going to close the PR and resend once I have better calrity. We can continue the discussion on the issue.

Issue #287, opened 2017-01-13
Implement proxy support for VM registration 


hferentschik:
As a preparation for issue #90 which is about enabling proxies for Docker/OpenShift, we also need to be able to register a VM via a proxy. Adding this, will provide the Minishift flags and infrastructure needed for enabling proxies for Docker/OpenShift.

jorgemoralespou:
@praveenkumar @hferentschik please, align with minikube to the options/commands that are implemented in minikube, as we need to have the same experience, even if the code would not be shared.

https://github.com/kubernetes/minikube#using-minikube-with-an-http-proxy

praveenkumar:
@jorgemoralespou In minikube it just set docker proxy variable not VM specific which we need for registration and for oc binary. Now if we want to align it then we need to extract docker env and iterate to get proxy key then set it for oc and registration I think we should send a patch for minikube to allign with minishift for proxy stuff wdyt?

hferentschik:
> In minikube it just set docker proxy variable not VM specific which we need for registration and for oc binary

Right in Minikube they use the already existing 'docker-env' command. That won't cut it for us. We do actually a fair bit more here.

>  I think we should send a patch for minikube to allign with minishift for proxy stuff wdyt?

+1 Let's at least create an issue and discuss. Makes sense to share this code and having the explicit global proxy flags makes sense for Minikube as well.

jorgemoralespou:
@praveenkumar @hferentschik I think that as closer we get the experience with minishift, the easier will be for people to adopt minishift. Many people will hate having some switches for minikube and totally different for minishift.
If we can alter minikube, then it's awesome, but if not I would consider maybe looking into how to best provide the proxy experience with the current minikube experience. Maybe it's not optimal and for this one specific we shouldn't go further, but in the future I would love to see alignment in usage, which means that users would be able to move back and forth between both tools easily. That might mean we need to monitor more minikube and possibly get more involved in that community moving forward.

My 2 cents.

hferentschik:
Resolved via pull request #365

Issue #318, opened 2017-01-23
Incomplete handling of VM status for un-registration


hferentschik:
Currently the code for un-registrating the VM in _stop.go_ / _delete.go_ looks like this:

    if !drivers.MachineInState(host.Driver, state.Stopped)() {
        ...
    }

This means that un-registration will occur in all other states than stopped. However, there are other states in which un-regsitration won't work. We have the following states:

	"Running",
	"Paused",
	"Saved",
	"Stopped",
	"Stopping",
	"Starting",
	"Error",
	"Timeout",

Even more, the code for un-registration should not be in _stop.go_ / _delete.go_, but rather in _cluster.StopHost_ resp. _cluster.DeleteHost_, so that other callers of _StopHost_ can un-register as well.

praveenkumar:
@hferentschik Agree on that behavior. In past we agreed on that piece of code should be independent of `cluster.go` file because that's something comes from minikube and it will be easily rebase if we need to get changes from minikube repo. If that is what we still follow then I think better we should handle all state in same code base instead making changes in `cluster.go`

praveenkumar:
@hferentschik I think we only have to switch our condition to `if drivers.MachineInState(host.Driver, state.Running)()` because for any other state ssh command will not be able to execute on the VM side which handle unregister. wdyt?

amitkrout:
Adding more execution result to it.
``
$ minishift.exe version
minishift version: 1.0.0-beta.2
Virtualbox : 5.1.10
Platform : Windows 10
```
Randomly i was executing sanity tests includes docker images pull and run within the box. Then i exit from the box and found the error while running minishift command includes ssh, console and stop

```
$ minishift.exe start --vm-driver=virtualbox --iso-url https://github.com/minishift/minishift-centos-iso/releases/download/v1.0.0-beta.1/minishift-centos.iso
Starting local OpenShift instance using 'virtualbox' hypervisor...
Downloading ISO
 316.00 MB / 316.00 MB  100.00% 0sssss
E0123 10:07:53.048934    3936 start.go:135] Error starting machine: Error creating. Error creating machine: Error checking the host: Error checking and/or regenerating the certs: There was an error validating certificates for host "192.168.99.100:2376": dial tcp 192.168.99.100:2376: i/o timeout
You can attempt to regenerate them using 'docker-machine regenerate-certs [name]'.
Be advised that this will trigger a Docker daemon restart which might stop running containers.
. Retrying.
[...]

PR #365, opened 2017-02-03
Issue #287 Implement proxy support for VM registration

praveenkumar:
This patch will enable proxy support for VM and docker + openshift.

### DIFF
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -115,8 +115,8 @@ func setFlagsUsingViper() {
 
 func init() {
 	RootCmd.PersistentFlags().Bool(showLibmachineLogs, false, "Show logs from libmachine.")
-	RootCmd.PersistentFlags().String(username, "", "User name for the virtual machine.")
-	RootCmd.PersistentFlags().String(password, "", "Password for the virtual machine.")
+	RootCmd.PersistentFlags().String(username, "", "User name for the virtual machine registration.")
+	RootCmd.PersistentFlags().String(password, "", "Password for the virtual machine registration.")
 	RootCmd.AddCommand(configCmd.ConfigCmd)
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	logDir := pflag.Lookup("log_dir")
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -31,6 +31,7 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/cache"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
 	"github.com/minishift/minishift/pkg/minishift/registration"
+	minishiftUtil "github.com/minishift/minishift/pkg/minishift/util"
 	"github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/version"
 	dockerhost "github.com/openshift/origin/pkg/bootstrap/docker/host"
@@ -62,6 +63,11 @@ const (
 	serverLogLevel    = "server-loglevel"
 	openshiftEnv      = "openshift-env"
 	metrics           = "metrics"
+
+	// Setting proxy
+	httpProxy   = "http-proxy"
+	httpsProxy  = "https-proxy"
+	noProxyList = "no-proxy"
 )
 
 var (
@@ -69,6 +75,7 @@ var (
 	insecureRegistry []string
 	registryMirror   []string
 	openShiftEnv     []string
+	shellProxyEnv    string
 )
 
 // startCmd represents the start command
@@ -91,6 +98,9 @@ var startFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
 // clusterUpFlagSet contains the command line switches which needs to be passed on to 'cluster up'
 var clusterUpFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
 
+// proxyFlagSet contains the command line switches for proxy
+var proxyFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
+
 // minishiftToClusterUp is a mapping between falg names used in minishift CLI and flag name as passed to 'cluster up'
 var minishiftToClusterUp = map[string]string{
 	"openshift-env": "env",
@@ -112,6 +122,12 @@ func SetMinishiftDir(newDir string) {
 func runStart(cmd *cobra.Command, args []string) {
 	libMachineClient := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer libMachineClient.Close()
+
+	validateProxyArgs()
+
+	setDockerProxy()
+	setOcProxy()
+	setShellProxy()
 
 	config := cluster.MachineConfig{
 		MinikubeISO:      viper.GetString(isoURL),
@@ -141,6 +157,15 @@ func runStart(cmd *cobra.Command, args []string) {
 		glog.Errorln("Error starting the VM: ", err)
 		os.Exit(1)
 	}
+
+	// Set Proxy to as Shell Env
+	if viper.IsSet("http-proxy") || viper.IsSet("https-proxy") {
+		if err := minishiftUtil.SetProxyToShellEnv(host, shellProxyEnv); err != nil {
+			fmt.Printf("Error setting proxy to VM: %s", err)
+			os.Exit(1)
+		}
+	}
+
 	// Register Host VM
 	if err := registration.RegisterHostVM(host, RegistrationParameters); err != nil {
 		fmt.Printf("Error registering the VM: %s", err)
@@ -156,6 +181,56 @@ func runStart(cmd *cobra.Command, args []string) {
 	clusterUp(&config)
 }
 
+// Set Docker Proxy
+func setDockerProxy() {
+	if viper.IsSet(httpProxy) {
+		dockerEnv = append(dockerEnv, fmt.Sprintf("HTTP_PROXY=%s", viper.GetString(httpProxy)))
+	}
+	if viper.IsSet(httpsProxy) {
+		dockerEnv = append(dockerEnv, fmt.Sprintf("HTTPS_PROXY=%s", viper.GetString(httpsProxy)))
+	}
+	if viper.IsSet(noProxyList) {
+		dockerEnv = append(dockerEnv, fmt.Sprintf("NO_PROXY=%s,%s", updateNoProxyForDocker(),
+			viper.GetString(noProxyList)))
+	} else if viper.IsSet(httpProxy) || viper.IsSet(httpsProxy) {
+		dockerEnv = append(dockerEnv, fmt.Sprintf("NO_PROXY=%s", updateNoProxyForDocker()))
+	}
+
+}
+
+// Set OpenShiftProxy
+func setOcProxy() {
+	if viper.IsSet(httpProxy) {
+		clusterUpFlagSet.String(httpProxy, viper.GetString(httpProxy), "HTTP proxy to use for master and builds")
+	}
+	if viper.IsSet(httpsProxy) {
+		clusterUpFlagSet.String(httpsProxy, viper.GetString(httpsProxy), "HTTPS proxy to use for master and builds")
+	}
+	if viper.IsSet(noProxyList) {
+		clusterUpFlagSet.String(noProxyList, viper.GetString(noProxyList), "List of hosts or subnets for which a proxy shouldn't be use")
+	}
+}
+
+// Set shell proxy
+func setShellProxy() {
+	if viper.IsSet(httpProxy) {
+		shellProxyEnv += fmt.Sprintf("http_proxy=%s ", viper.GetString(httpProxy))
+	}
+	if viper.IsSet(httpsProxy) {
+		shellProxyEnv += fmt.Sprintf("https_proxy=%s ", viper.GetString(httpsProxy))
+	}
+	if viper.IsSet(noProxyList) {
+		shellProxyEnv += fmt.Sprintf("no_proxy=%s", viper.GetString(noProxyList))
+	} else if viper.IsSet(httpProxy) || viper.IsSet(httpsProxy) {
+		dockerEnv = append(dockerEnv, fmt.Sprintf("NO_PROXY=%s", updateNoProxyForDocker()))
+	}
+}
+
+// update default no-proxy for docker
+func updateNoProxyForDocker() string {
+	return "localhost,127.0.0.1,172.30.1.1"
+}
+
 // calculateDiskSizeInMB converts a human specified disk size like "1000MB" or "1GB" and converts it into Megabits
 func calculateDiskSizeInMB(humanReadableDiskSize string) int {
 	diskSize, err := units.FromHumanSize(humanReadableDiskSize)
@@ -169,9 +244,11 @@ func calculateDiskSizeInMB(humanReadableDiskSize string) int {
 func init() {
 	initStartFlags()
 	initClusterUpFlags()
+	initProxyFlags()
 
 	startCmd.Flags().AddFlagSet(startFlagSet)
 	startCmd.Flags().AddFlagSet(clusterUpFlagSet)
+	startCmd.Flags().AddFlagSet(proxyFlagSet)
 
 	viper.BindPFlags(startCmd.Flags())
 	RootCmd.AddCommand(startCmd)
@@ -208,6 +285,13 @@ func initClusterUpFlags() {
 	clusterUpFlagSet.Bool(metrics, false, "Install metrics (experimental)")
 }
 
+// initProxyFlags create the CLI flags which needs to be passed for proxy
+func initProxyFlags() {
+	proxyFlagSet.String(httpProxy, "", "HTTP proxy for virtual machine (In the format of http://<username>:<password>@<proxy_host>:<proxy_port>)")
+	proxyFlagSet.String(httpsProxy, "", "HTTPS proxy for virtual machine (In the format of https://<username>:<password>@<proxy_host>:<proxy_port>)")
+	proxyFlagSet.String(noProxyList, "", "List of hosts or subnets for which proxy should not be used.")
+}
+
 // clusterUp downloads and installs the oc binary in order to run 'cluster up'
 func clusterUp(config *cluster.MachineConfig) {
 	oc := cache.Oc{
@@ -235,6 +319,10 @@ func clusterUp(config *cluster.MachineConfig) {
 			if exists {
 				key = minishiftToClusterUp[key]
 			}
+			if !ocSupportFlag(cmdName, key) {
+				glog.Errorf("Flag %s is not supported for current oc verison", flag.Name)
+				os.Exit(1)
+			}
 			cmdArgs = append(cmdArgs, "--"+key)
 			cmdArgs = append(cmdArgs, value)
 		}
@@ -246,4 +334,31 @@ func clusterUp(config *cluster.MachineConfig) {
 		glog.Errorln("Error starting the cluster: ", err)
 		os.Exit(1)
 	}
+}
+
+func validateProxyArgs() {
+	if viper.IsSet(httpProxy) {
+		if !util.ValidateProxyURI(viper.GetString(httpProxy)) {
+			glog.Exitf("HTTP Proxy URL is not valid, Please check help message")
+		}
+	}
+	if viper.IsSet(httpsProxy) {
+		if !util.ValidateProxyURI(viper.GetString(httpsProxy)) {
+			glog.Exitf("HTTPS Proxy URL is not valid, Please check help message")
+		}
+	}
+}
+
+func ocSupportFlag(cmdName string, flag string) bool {
+	cmdArgs := []string{"cluster", "up", "-h"}
+	cmdOut, err := runner.Output(cmdName, cmdArgs...)
+	if err != nil {
+		glog.Errorf("Not able to get output of 'oc -h' Error: %s", err)
+		os.Exit(1)
+	}
+	ocCommandOptions := util.ParseOcHelpCommand(cmdOut)
+	if ocCommandOptions != nil {
+		return util.FlagExist(ocCommandOptions, flag)
+	}
+	return false
 }
--- a/cmd/minikube/cmd/start_test.go
+++ b/cmd/minikube/cmd/start_test.go
@@ -30,6 +30,69 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"fmt"
+)
+
+var (
+	ocClusterUpHelp = []byte(`Starts an OpenShift cluster using Docker containers, provisioning a registry, router, initial templates, and a default
+project.
+
+This command will attempt to use an existing connection to a Docker daemon. Before running the command, ensure that you
+can execure docker commands successfully (i.e. 'docker ps').
+
+Optionally, the command can create a new Docker machine for OpenShift using the VirtualBox driver when the
+--create-machine argument is specified. The machine will be named 'openshift' by default. To name the machine
+differently, use the --docker-machine=NAME argument. If the --docker-machine=NAME argument is specified, but
+--create-machine is not, the command will attempt to find an existing docker machine with that name and start it if it's
+not running.
+
+By default, the OpenShift cluster will be setup to use a routing suffix that ends in xip.io. This is to allow dynamic
+host names to be created for routes. An alternate routing suffix can be specified using the --routing-suffix flag.
+
+A public hostname can also be specified for the server with the --public-hostname flag.
+
+Usage:
+  oc cluster up [options]
+
+Examples:
+  # Start OpenShift on a new docker machine named 'openshift'
+  oc cluster up --create-machine
+
+  # Start OpenShift using a specific public host name
+  oc cluster up --public-hostname=my.address.example.com
+
+  # Start OpenShift and preserve data and config between restarts
+  oc cluster up --host-data-dir=/mydata --use-existing-config
+
+  # Use a different set of images
+  oc cluster up --image="registry.example.com/origin" --version="v1.1"
+
+Options:
+      --create-machine=false: If true, create a Docker machine if one doesn't exist
+      --docker-machine='': Specify the Docker machine to use
+  -e, --env=[]: Specify a key-value pair for an environment variable to set on OpenShift container
+      --forward-ports=true: If true, use Docker port-forwarding to communicate with origin container. Requires 'socat'
+locally.
+      --host-config-dir='/var/lib/origin/openshift.local.config': Directory on Docker host for OpenShift configuration
+      --host-data-dir='': Directory on Docker host for OpenShift data. If not specified, etcd data will not be persisted
+on the host.
+      --host-pv-dir='/var/lib/origin/openshift.local.pv': Directory on host for OpenShift persistent volumes
+      --host-volumes-dir='/var/lib/origin/openshift.local.volumes': Directory on Docker host for OpenShift volumes
+      --http-proxy='': HTTP proxy to use for master and builds
+      --https-proxy='': HTTPS proxy to use for master and builds
+      --image='openshift/origin': Specify the images to use for OpenShift
+      --image-streams='centos7': Specify which image streams to use, centos7|rhel7
+      --logging=false: If true, install logging (experimental)
+      --metrics=false: If true, install metrics (experimental)
+      --no-proxy=[]: List of hosts or subnets for which a proxy should not be used
+      --public-hostname='': Public hostname for OpenShift cluster
+      --routing-suffix='': Default suffix for server routes
+      --server-loglevel=0: Log level for OpenShift server
+      --skip-registry-check=false: If true, skip Docker daemon registry check
+      --use-existing-config=false: If true, use existing configuration if present
+      --version='': Specify the tag for OpenShift images
+
+Use "oc options" for a list of global command-line options (applies to all commands).`)
 )
 
 type RecordingRunner struct {
@@ -41,6 +104,12 @@ func (r *RecordingRunner) Run(command string, args ...string) error {
 	r.Cmd = command
 	r.Args = args
 	return nil
+}
+
+func (r *RecordingRunner) Output(command string, args ...string) ([]byte, error) {
+	r.Cmd = command
+	r.Args = args
+	return ocClusterUpHelp, nil
 }
 
 var testDir string
@@ -126,6 +195,60 @@ func TestStartClusterUpWithFlag(t *testing.T) {
 	assertCommandLineArguments(expectedArguments, t)
 }
 
+func TestShellProxyVariableWithProxyFlag(t *testing.T) {
+	defer viper.Reset()
+
+	expectedShellProxyVariable := "http_proxy=http://localhost:3128 https_proxy=https://localhost:3128 no_proxy=localhost,127.0.0.1"
+
+	viper.Set("http-proxy", "http://localhost:3128")
+	viper.Set("https-proxy", "https://localhost:3128")
+	viper.Set("no-proxy", "localhost,127.0.0.1")
+
+	setShellProxy()
+
+	if shellProxyEnv != expectedShellProxyVariable {
+		t.Fatalf("Shell variable doesn't set properly expected %s got %s", expectedShellProxyVariable, shellProxyEnv)
+	}
+}
+
+func TestDockerEnvWithProxyFlag(t *testing.T) {
+	defer viper.Reset()
+
+	expectedDockerEnv := []string{"HTTP_PROXY=http://localhost:3128",
+		fmt.Sprintf("NO_PROXY=%s", updateNoProxyForDocker())}
+
+	viper.Set("http-proxy", "http://localhost:3128")
+
+	setDockerProxy()
+
+	assertCompareSlice(expectedDockerEnv, dockerEnv, t)
+}
+
+func TestClusterUpWithProxyFlag(t *testing.T) {
+	setUp(t)
+	defer os.RemoveAll(testDir)
+	defer minitesting.ResetDefaultRoundTripper()
+	defer SetRunner(util.RealRunner{})
+	defer viper.Reset()
+
+	viper.Set("http-proxy", "http://localhost:3128")
+	viper.Set("https-proxy", "https://localhost:3128")
+	viper.Set("no-proxy", "10.0.0.1")
+
+	setOcProxy()
+
+	clusterUp(&testMachineConfig)
+
+	expectedArguments := []string{"cluster", "up", "--use-existing-config",
+		"--host-config-dir", hostConfigDirectory,
+		"--host-data-dir", hostDataDirectory,
+		"--http-proxy", "http://localhost:3128",
+		"--https-proxy", "https://localhost:3128",
+		"--no-proxy", "10.0.0.1"}
+	assertCommandLineArguments(expectedArguments, t)
+
+}
+
 func TestStartClusterUpWithOpenShiftEnv(t *testing.T) {
 	setUp(t)
 	defer os.RemoveAll(testDir)
@@ -141,6 +264,22 @@ func TestStartClusterUpWithOpenShiftEnv(t *testing.T) {
 		"--host-data-dir", hostDataDirectory,
 		"--env", "HTTP_PROXY=http://localhost:3128,HTTP_PROXY_USER=foo,HTTP_PROXY_PASS=bar"}
 	assertCommandLineArguments(expectedArguments, t)
+}
+
+func assertCompareSlice(expectedArguments []string, recievedArguments []string, t *testing.T) {
+	if len(expectedArguments) > len(recievedArguments) {
+		t.Errorf("Expected more arguments than received. Expected: '%s'. Got: '%s'", expectedArguments, recievedArguments)
+	}
+
+	if len(expectedArguments) < len(recievedArguments) {
+		t.Errorf("Received more arguments than expected. Expected: '%s'. Got '%s'", expectedArguments, recievedArguments)
+	}
+
+	for i, v := range recievedArguments {
+		if v != expectedArguments[i] {
+			t.Errorf("Expected argument '%s'. Received '%s'", expectedArguments[i], v)
+		}
+	}
 }
 
 func assertCommandLineArguments(expectedArguments []string, t *testing.T) {
--- a/docs/minishift_docker-env.md
+++ b/docs/minishift_docker-env.md
@@ -26,10 +26,10 @@ minishift docker-env
       --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
       --log_dir string                   If non-empty, write log files in this directory (default "")
       --logtostderr                      log to standard error instead of files
-      --password string                  Password for the virtual machine.
+      --password string                  Password for the virtual machine registration.
       --show-libmachine-logs             Show logs from libmachine.
       --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
-      --username string                  User name for the virtual machine.
+      --username string                  User name for the virtual machine registration.
   -v, --v Level                          log level for V logs
       --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
 ```
--- a/glide.lock
+++ b/glide.lock
@@ -1,8 +1,10 @@
-hash: b22433552234edad060646b980eebe603a7f98b8447c1b81bc831e49791f1fc1
-updated: 2017-02-02T17:05:27.616783744+05:30
+hash: 6c46de8510c27a5b6b56122ac96feda1ad9651b9f6a1ec89bd4ed1bbf272e9fd
+updated: 2017-02-07T11:55:45.327429678+05:30
 imports:
+- name: github.com/asaskevich/govalidator
+  version: 7b3beb6df3c42abd3509abfc3bcacc0fbfb7c877
 - name: github.com/blang/semver
-  version: ff9bd0e2be32754107fc7f37531b104b6219eab8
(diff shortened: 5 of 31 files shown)


pkg/minishift/registration/utils.go
```diff
@@ -27,9 +27,6 @@ import (
 type RegistrationParameters struct {
 	Username      string
 	Password      string
-	Proxy         string
-	ProxyUser     string
-	ProxyPassword string
```
praveenkumar: Since for subscription we are not using individual flag for proxy so removing those variable from struct.

pkg/util/runner.go
```diff
@@ -33,6 +33,7 @@ func (r RealRunner) Run(command string, args ...string) error {
 	fmt.Println(fmt.Sprintf("Provisioning OpenShift via '%s %s'", command, args))
 	cmd := exec.Command(command, args...)
 	cmd.Stdout = os.Stdout
+	cmd.Stderr = os.Stdout
```
praveenkumar: This will make sure if we pass any unsupported flag to oc like http proxy not support less than version 1.5.0 then it will tell user unsupported option. As per doc https://golang.org/pkg/os/exec/

```
        // If Stdout and Stderr are the same writer, at most one
        // goroutine at a time will call Write.
        Stdout io.Writer
        Stderr io.Writer
```

cmd/minikube/cmd/start.go
```diff
@@ -156,6 +169,36 @@ func runStart(cmd *cobra.Command, args []string) {
...
+		viper.Set(noProxyList, viper.GetString("no-proxy"))
+		shellProxyEnv += fmt.Sprintf("no_proxy=%s", viper.GetString("no-proxy"))
+	} else if viper.IsSet("http-proxy") || viper.IsSet("https-proxy") {
+		dockerEnv = append(dockerEnv, fmt.Sprintf("NO_PROXY=%s", updateNoProxyForDocker()))
+	}
+}
+
+// update default no-proxy for docker
+func updateNoProxyForDocker() string {
+	return "localhost,127.0.0.1,172.30.1.1"
```
budhrg: IMO, `constants.go` is good place to store it. Hard coded values doesn't looks good.

pkg/minishift/util/setproxytoshell.go
```diff
@@ -0,0 +1,32 @@
+/*
```
budhrg: Let's follow Go file naming convention here `set_proxy_to_shell.go`

praveenkumar: @budhrg In go for file name and variable we don't use "_" but use Mixed-case IIRC.

cmd/minikube/cmd/start.go
praveenkumar: @budhrg it is hardcoded for oc cluster up also https://github.com/openshift/origin/pull/12483/files#diff-289dd62730ab075b34eb10a03a9687e6R671 I just used same formatting.

pkg/minishift/util/setproxytoshell.go
```diff
@@ -0,0 +1,32 @@
...
+
+func SetProxyToShellEnv(host *host.Host, shellProxyEnv string) error {
+	if _, err := host.RunSSHCommand(fmt.Sprintf("export %s", shellProxyEnv)); err !=nil {
+		return err
+	}
+	if _, err := host.RunSSHCommand(fmt.Sprintf(`sudo su -c 'echo "export %s" > /etc/profile.d/proxy.sh'`, shellProxyEnv)); err !=nil {
+		return err
+	}
+	return nil
+}
```
budhrg: one line is missing here, `fmtcheck` might fail

cmd/minikube/cmd/start.go
budhrg: There I don't think they can put as constant as they are building array but I think here we have a scope.

pkg/minishift/util/setproxytoshell.go
praveenkumar: @budhrg looks like in file name can have '_' but http://stackoverflow.com/a/25162021 should also need to consider.

budhrg: @praveenkumar Looks like no such convention, only convenctions are http://stackoverflow.com/a/25162021/1120530.

But IMO, adding `_` give more readability. Like this https://github.com/minishift/minishift/blob/master/cmd/minikube/cmd/get_openshift_versions.go

cmd/minikube/cmd/root.go
```diff
@@ -117,6 +120,9 @@ func init() {
 	RootCmd.PersistentFlags().Bool(showLibmachineLogs, false, "Show logs from libmachine.")
 	RootCmd.PersistentFlags().String(username, "", "User name for the virtual machine.")
```
hferentschik: Wanted to comment before on this, but this might be the better scope. "User name for the virtual machine." seems wrong and could be misinterpreted as some sort of credentials to log onto the VM. Really this should be something like "User name for virtual machine registration"

cmd/minikube/cmd/root.go
```diff
@@ -117,6 +120,9 @@ func init() {
...
 	RootCmd.PersistentFlags().String(password, "", "Password for the virtual machine.")
```
hferentschik: Same as with the username description

cmd/minikube/cmd/start.go
```diff
@@ -113,6 +115,8 @@ func runStart(cmd *cobra.Command, args []string) {
 	libMachineClient := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer libMachineClient.Close()
 
+	setProxy()
```
hferentschik: Instead of this _setProxy_ here, I rather split the content of this method into two parts, one which affects Docker and one which affects OpenShift. something like: _determineDockerEnv_ and _determineOpenShiftEnv_. The former can then be used within the cluster.MachineConfig setup.

cmd/minikube/cmd/start.go
```diff
@@ -141,6 +145,15 @@ func runStart(cmd *cobra.Command, args []string) {
 		glog.Errorln("Error starting the VM: ", err)
 		os.Exit(1)
 	}
+
+	// Set Proxy to as Shell Env
+	if viper.IsSet("http-proxy") || viper.IsSet("https-proxy") {
```
hferentschik: Wouldn't it make sense to move this into cluster.StartHost? In fact, we might want to add the proxy options to _cluster.MachineConfig_? WDYT?

cmd/minikube/cmd/start.go
```diff
@@ -141,6 +145,15 @@ func runStart(cmd *cobra.Command, args []string) {
...
+		if err := minishiftUtil.SetProxyToShellEnv(host, shellProxyEnv); err != nil {
+			fmt.Printf("Error setting proxy to VM: %s", err)
+			os.Exit(1)
+		}
+	}
+
 	// Register Host VM
 	if err := registration.RegisterHostVM(host, RegistrationParameters); err != nil {
```
hferentschik: Maybe we can move this as well now. This needs to go into StartHost.

cmd/minikube/cmd/start.go
hferentschik: I don't think this belongs to constants.go. It could be a const in this file for sure, but constants.go seems wrong. Personally I would also make it a const, but I am fine with a method as well.

pkg/minishift/registration/utils.go
hferentschik: +1

pkg/minishift/util/setproxytoshell.go
hferentschik: > Let's follow Go file naming convention here set_proxy_to_shell.go

I don't think this is a Go convention, but the current code base uses underscores in file names, so I think we should stick to it. In actual code we of course use camel case.

pkg/util/runner.go
hferentschik: So what will be the error message in the end? For the case the flag is not supported?

cmd/minikube/cmd/start.go
```diff
@@ -156,6 +169,36 @@ func runStart(cmd *cobra.Command, args []string) {
 	clusterUp(&config)
 }
 
+// Set docker and oc cluster proxy in place
+func setProxy() {
+	if viper.IsSet("http-proxy") {
```
hferentschik: Should we not add some basic validation to the parameter? This could save the user a quite some pain in case of typos.

cmd/minikube/cmd/start.go
praveenkumar: @hferentschik  Why I am actually not in favor to move registration bits and this one to StartHost because https://github.com/minishift/minishift/blob/master/cmd/minikube/cmd/start.go#L139 so whatever we move cluster.StartHost it will going to provision 3 time in case of any error. Since registration take around 1-2 mins if that fail first time the it start retry (assume user provided invalid username/password). 

MachineConfig already have Docker Env which actually responsible ( https://github.com/minishift/minishift/blob/master/pkg/minikube/cluster/cluster.go#L172-L180 ) for set across different provisioner => https://github.com/minishift/minishift/blob/master/cmd/minikube/cmd/start.go#L139

cmd/minikube/cmd/start.go
praveenkumar: @hferentschik As I described  https://github.com/minishift/minishift/pull/365#discussion_r99322076 I think we shouldn't add any our own param here because this is coming from docker-machine side.

cmd/minikube/cmd/root.go
praveenkumar: Ack

cmd/minikube/cmd/root.go
praveenkumar: Agree

cmd/minikube/cmd/start.go
praveenkumar: What kind of validation we are talking here valid proxy url?

pkg/util/runner.go
praveenkumar:
```
Provisioning OpenShift via '/Users/prkumar/.minishift/cache/oc/v1.4.1/oc [cluster up --use-existing-config --host-config-dir /var/lib/minishift/openshift.local.config --host-data-dir /var/lib/minishift/hostdata --http-proxy http://fedora:fedora123@10.70.49.109:3128]'
Error: unknown flag: --http-proxy


Usage:
  oc cluster up [options]

Examples:
  # Start OpenShift on a new docker machine named 'openshift'
  oc cluster up --create-machine
  
  # Start OpenShift using a specific public host name
  oc cluster up --public-hostname=my.address.example.com
  
  # Start OpenShift and preserve data and config between restarts
  oc cluster up --host-data-dir=/mydata --use-existing-config
  
  # Use a different set of images
  oc cluster up --image="registry.example.com/origin" --version="v1.1"

Options:
      --create-machine=false: Create a Docker machine if one doesn't exist
      --docker-machine='': Specify the Docker machine to use
  -e, --env=[]: Specify a key-value pair for an environment variable to set on OpenShift container
      --forward-ports=false: Use Docker port-forwarding to communicate with origin container. Requires 'socat' locally.
      --host-config-dir='/var/lib/origin/openshift.local.config': Directory on Docker host for OpenShift configuration
      --host-data-dir='': Directory on Docker host for OpenShift data. If not specified, etcd data will not be persisted on the host.
      --host-volumes-dir='/var/lib/origin/openshift.local.volumes': Directory on Docker host for OpenShift volumes
      --image='openshift/origin': Specify the images to use for OpenShift
      --logging=false: Install logging (experimental)
      --metrics=false: Install metrics (experimental)
      --public-hostname='': Public hostname for OpenShift cluster
      --routing-suffix='': Default suffix for server routes
      --server-loglevel=0: Log level for OpenShift server
      --skip-registry-check=false: Skip Docker daemon registry check
      --use-existing-config=false: Use existing configuration if present
      --version='': Specify the tag for OpenShift images

Use "oc options" for a list of global command-line options (applies to all commands).

E0202 17:35:50.329889   86962 start.go:276] Error starting the cluster:  exit status 1
```

hferentschik: > Provisioning OpenShift via '/Users/prkumar/.minishift/cache/oc/v1.4.1/oc [cluster up --use-existing-config --host-config-dir /var/lib/minishift/openshift.local.config --host-data-dir /var/lib/minishift/hostdata --http-proxy http://fedora:fedora123@10.70.49.109:3128]'
Error: unknown flag: --http-proxy

I think we should do better than that and have a nicer error message. There is a lot of 'oc cluster' in there which a novice might not understand. Let's do the mapping approach and determine beforehand whether we can pass the proxy. If not, we can say something along the line:

"The requested Openshift version <version> does not support HTTP proxies. You will need to use a newer version of OpenSift".

BTW, do we want to halt the VM at this stage? I guess it is running right?

cmd/minikube/cmd/root.go
hferentschik: cool

cmd/minikube/cmd/start.go
hferentschik: >  Why I am actually not in favor to move registration bits and this one to StartHost because https://github.com/minishift/minishift/blob/master/cmd/minikube/cmd/start.go#L139 so whatever we move cluster.StartHost it will going to provision 3 time in case of any error. 

Ok, we might change the retry then. I don't see much use in this anyway. Unless we are covering for a error condition which occurs sporadically. 

Otherwise, we at least need to move the functionality of registration into a public method in cluster.

hferentschik: > MachineConfig already have Docker Env

Sure, but not proxy settings, right? It's a machine config thing, or not?

cmd/minikube/cmd/start.go
hferentschik: That the proxy matches the expected format - http://fedora:fedora123@10.70.49.109:3128
In fact, the help string should mention mention this very format, so that the user knows how to set the proxy.

cmd/minikube/cmd/start.go
hferentschik: There might be a misunderstanding here. I don't want to change the DockerEnv property, but instead of just assigning _dockerEnv_, I'd make this a function call at this point (which returns the correct "proxified" string array if required).

Also, what comes from docker-machine? cluster.MachineConfig is from us right? So we can easily for example add the proxy to it.

pkg/util/runner.go
praveenkumar: @hferentschik That is something we should handle as a separate PR I can remove this bit for this and then we can have separate issue for this.

cmd/minikube/cmd/start.go
praveenkumar: > Ok, we might change the retry then. I don't see much use in this anyway. Unless we are covering for a error condition which occurs sporadically.

It's required because many time cert creation for ssh need time and that time it do the retry (Which I observed some time back).

> Otherwise, we at least need to move the functionality of registration into a public method in cluster.

That I can take a look but I would make this part of change in https://github.com/minishift/minishift/issues/318

cmd/minikube/cmd/start.go
praveenkumar: @hferentschik we can add proxy to cluster.MachineConfig struct but to pass it till docker daemon we should have such option in the docker-machine side also https://github.com/minishift/minishift/blob/master/pkg/minikube/cluster/cluster.go#L174 here engine.Options doesn't have proxy argument but it is only supported by dockerEnv which take care of daemon options.

praveenkumar:
> Where are the tests? There should be ample of opportunities for tests.

Agree, added test cases for proxy checks.

cmd/minikube/cmd/root.go
```diff
@@ -115,8 +118,11 @@ func setFlagsUsingViper() {
 
 func init() {
 	RootCmd.PersistentFlags().Bool(showLibmachineLogs, false, "Show logs from libmachine.")
-	RootCmd.PersistentFlags().String(username, "", "User name for the virtual machine.")
-	RootCmd.PersistentFlags().String(password, "", "Password for the virtual machine.")
+	RootCmd.PersistentFlags().String(username, "", "User name for the virtual machine registration.")
+	RootCmd.PersistentFlags().String(password, "", "Password for the virtual machine registration.")
+	RootCmd.PersistentFlags().String(httpProxy, "", "HTTP proxy for virtual machine.")
+	RootCmd.PersistentFlags().String(httpsProxy, "", "HTTPS proxy for virtual machine.")
```
hferentschik: Do we really need two proxy options? Given the format we require, the distinction between HTTP and HTTPS is implicit right? Different if you just allow the user to do something like HTTPS_PROXY=<hostname>, then the protocol cannot be derived. WDYT?

cmd/minikube/cmd/start.go
hferentschik: > we can add proxy to cluster.MachineConfig struct but to pass it till docker daemon we should have such option in the docker-machine side also

I am not after passing this through to docker-machine. I am trying to make things a bit more explicit. setProxy does all and I would like to split things up a bit. 

* First get the proxy settings from Viper and process/validate them
* Then apply them for example to MachineConfig, which also allows you to pass this options into StartHost or any other helper method. Really this are machine config options, right? Also it might make things more testable. Right now, you need to test proxy settings at the same time as command line options, because setProxy deals with both. By separating these two things (processing the falgs and doing something with them), you open up more/better ways of testing this

cmd/minikube/cmd/start.go
hferentschik: > It's required because many time cert creation for ssh need time and that time it do the retry (Which I observed some time back).

And why can we not wait for the cert creation to be finished. Something does not sounds right here. And if that is so, we might have to bite into the sour apple and have the registration in the loop. Correctness wins in this case.

hferentschik: > That I can take a look but I would make this part of change in #318

fair enough

cmd/minikube/cmd/start_test.go
```diff
@@ -126,6 +127,71 @@ func TestStartClusterUpWithFlag(t *testing.T) {
...
+func TestShellProxyVariableWithProxyFlag(t *testing.T) {
+	setUp(t)
+	defer os.RemoveAll(testDir)
+	defer minitesting.ResetDefaultRoundTripper()
+	defer SetRunner(util.RealRunner{})
+	defer viper.Reset()
+
+	expectedShellProxyVariable := "http_proxy=http://localhost:3128 https_proxy=https://localhost:3128 no_proxy=localhost,127.0.0.1"
+
+	viper.Set("http-proxy", "http://localhost:3128")
```
hferentschik: does this even make sense? Can one set both values? Are these proxies dynamically chosen?

cmd/minikube/cmd/start.go
```diff
@@ -141,6 +145,15 @@ func runStart(cmd *cobra.Command, args []string) {
 		glog.Errorln("Error starting the VM: ", err)
 		os.Exit(1)
 	}
+
+	// Set Proxy to as Shell Env
+	if viper.IsSet("http-proxy") || viper.IsSet("https-proxy") {
+		if err := minishiftUtil.SetProxyToShellEnv(host, shellProxyEnv); err != nil {
```
hferentschik: You are not testing anywhere that anything happens to the VM

cmd/minikube/cmd/start_test.go
```diff
@@ -126,6 +127,71 @@ func TestStartClusterUpWithFlag(t *testing.T) {
...
+}
+
+func TestClusterUpWithProxyFlag(t *testing.T) {
+	setUp(t)
+	defer os.RemoveAll(testDir)
+	defer minitesting.ResetDefaultRoundTripper()
+	defer SetRunner(util.RealRunner{})
+	defer viper.Reset()
+
+	viper.Set("http-proxy", "http://localhost:3128")
```
hferentschik: See, here you need to tests flags and proxy configuration code together. If you separate this you can so the Viper/command line thing independent from the configuration part. 

Otherwise good step forward with the tests

budhrg:
Error reported by centos-ci looks valid for me:
```
00:04:38.255 start flag redefined: http-proxy
00:04:38.256 --- FAIL: TestDockerEnvWithProxyFlag (0.00s)
00:04:38.256 panic: start flag redefined: http-proxy [recovered]
00:04:38.256 	panic: start flag redefined: http-proxy
```

cmd/minikube/cmd/root.go
hferentschik: Forget this one, I was thinking wrong. I think I understand now. Tools needs to select the proxy depending on what they want to proxy, right?

praveenkumar: @hferentschik yes that's right.

cmd/minikube/cmd/start.go
praveenkumar: I did try to find a resource which can actually validate user input for proxy URL but I didn't find any standard/non-standard way of handling it. url.Parse is something people use but it's not parse the proxy url which we want it to do.  I agree on the help string side changes.

hferentschik: > I did try to find a resource which can actually validate user input for proxy URL but I didn't find any standard/non-standard way of handling it.

We can do it ourself, right? Using a regexp?

cmd/minikube/cmd/start.go
```diff
@@ -156,6 +169,36 @@ func runStart(cmd *cobra.Command, args []string) {
...
+// Set docker and oc cluster proxy in place
+func setProxy() {
+	if viper.IsSet("http-proxy") {
+		dockerEnv = append(dockerEnv, fmt.Sprintf("HTTP_PROXY=%s", viper.GetString("http-proxy")))
+		clusterUpFlagSet.String(httpProxy, "", "HTTP proxy to use for master and builds")
+		viper.Set(httpProxy, viper.GetString("http-proxy"))
+		shellProxyEnv += fmt.Sprintf("http_proxy=%s ", viper.GetString("http-proxy"))
+	}
+	if viper.IsSet("https-proxy") {
+		dockerEnv = append(dockerEnv, fmt.Sprintf("HTTPS_PROXY=%s", viper.GetString("https-proxy")))
```
hferentschik: Maybe what "bugs" me is that in this method you are setting up three things in one go - Docker, OpenShift and the env for the VM in general. That's quite intertwined and not really clear to the user. Opposed to something like setOpenShiftProxyEnv, setDockerProxyEnv, setVmProxyEnv.

cmd/minikube/cmd/start_test.go
praveenkumar: @hferentschik we can set both values also something like if same server serve http and https request then it's dynamically chosen by application.

```
export {http,https}_proxy="http://proxy-server:port"
```

cmd/minikube/cmd/root.go
```diff
@@ -115,8 +120,11 @@ func setFlagsUsingViper() {
 
 func init() {
 	RootCmd.PersistentFlags().Bool(showLibmachineLogs, false, "Show logs from libmachine.")
-	RootCmd.PersistentFlags().String(username, "", "User name for the virtual machine.")
-	RootCmd.PersistentFlags().String(password, "", "Password for the virtual machine.")
+	RootCmd.PersistentFlags().String(username, "", "User name for the virtual machine registration.")
+	RootCmd.PersistentFlags().String(password, "", "Password for the virtual machine registration.")
+	RootCmd.PersistentFlags().String(httpProxy, "", "HTTP proxy for virtual machine (ex. http://<username>:<password>@foo.com:<proxy_port>)")
```
hferentschik: Shouldn't it be \<proxy_host\> instead of foo.com?

hferentschik: and maybe instead of "ex" "in the format ..."

cmd/minikube/cmd/start.go
```diff
@@ -156,6 +171,56 @@ func runStart(cmd *cobra.Command, args []string) {
 	clusterUp(&config)
 }
 
+// Set Docker Proxy
+func setDockerProxy () {
```
hferentschik: +1,  I think this is better and easier to understand, even though you have to repeat the if statements

cmd/minikube/cmd/root.go
```diff
@@ -171,3 +179,16 @@ func ensureConfigFileExists(configPath string) {
 		}
 	}
 }
+
+func validateProxyArgs() {
+	if viper.IsSet(httpProxy) {
+		if ! util.ValidateProxyURI(viper.GetString(httpProxy)){
+			glog.Exitf("HTTP Proxy URL is not valid, Please check help message")
```
hferentschik: Since this is a user feedback message, I think we should use ftm.println. Otherwise the user needs to deal with the "strange" log format

cmd/minikube/cmd/start_test.go
```diff
@@ -126,6 +127,68 @@ func TestStartClusterUpWithFlag(t *testing.T) {
 	assertCommandLineArguments(expectedArguments, t)
 }
 
+func TestShellProxyVariableWithProxyFlag(t *testing.T) {
+	setUp(t)
+	defer os.RemoveAll(testDir)
```
hferentschik: technically a lot of this setup is not needed here. I don't think this tests needs a temp dir, nor does it need to HTTP round tripper nor the runner. Since you call _setup_ you will need to add this defer methods though. 

I think for this test, you can just not call setup, just a call to 'defer viper.Reset()' should do imo

cmd/minikube/cmd/start_test.go
```diff
@@ -126,6 +127,68 @@ func TestStartClusterUpWithFlag(t *testing.T) {
...
+	viper.Set("no-proxy", "localhost,127.0.0.1")
+
+	setShellProxy()
+
+	if shellProxyEnv != expectedShellProxyVariable {
+		t.Fatalf("Shell variable doesn't set properly expected %s got %s", expectedShellProxyVariable, shellProxyEnv)
+	}
+}
+
+func TestDockerEnvWithProxyFlag(t *testing.T) {
```
hferentschik: +1, nice test. Same comment regarding the setup as above though

docs/minishift_docker-env.md
```diff
@@ -23,13 +23,15 @@ minishift docker-env
 
 ```
+      --http-proxy string                HTTP proxy for virtual machine (ex. http://<username>:<password>@foo.com:<proxy_port>)
+      --https-proxy string               HTTPS proxy for virtual machine (ex. https://<username>:<password>@foo.com:<proxy_port>)
-      --password string                  Password for the virtual machine.
+      --password string                  Password for the virtual machine registration.
```
hferentschik: hmm, here the global --no-proxy does not get added. I guess that's because the command itself has a no-proxy option. Do we have a conflict there?

pkg/util/utils_test.go
```diff
@@ -60,6 +60,22 @@ func TestRetry(t *testing.T) {
 
 }
 
+func TestValidateProxyURI(t *testing.T) {
```
hferentschik: :+1:

glide.yaml
```diff
@@ -53,6 +53,8 @@ import:
   - ssh
 - package: golang.org/x/oauth2
 - package: gopkg.in/cheggaaa/pb.v1
+- package: github.com/asaskevich/govalidator/
```
hferentschik: Nice one. Seems to do what we want. Not that we could not have done this ourselves ;-)

glide.lock
```diff
@@ -1,8 +1,10 @@
-hash: b22433552234edad060646b980eebe603a7f98b8447c1b81bc831e49791f1fc1
-updated: 2017-02-02T17:05:27.616783744+05:30
+hash: 6c46de8510c27a5b6b56122ac96feda1ad9651b9f6a1ec89bd4ed1bbf272e9fd
+updated: 2017-02-07T11:55:45.327429678+05:30
 imports:
+- name: github.com/asaskevich/govalidator
+  version: 7b3beb6df3c42abd3509abfc3bcacc0fbfb7c877
 - name: github.com/blang/semver
-  version: ff9bd0e2be32754107fc7f37531b104b6219eab8
+  version: 31b736133b98f26d5e078ec9eb591666edfd091f
```
hferentschik: It keeps worrying me that adding a new dependency keeps also changing things which are effectively unrelated. This is due to the fact that several of our dependencies are still not specifying versions. At some stage we should do a walk-through the dependencies and see how many we can lock down.

hferentschik:
retest this please

hferentschik:
@praveenkumar - I tried to test this and have a couple of questions. 

* Have you tried proxy registration end to end?
* Will this implicitly fix #287 since we are setting the proxy settings in the environment. I know we talked about this, but I cannot remember what we said.

So here is what I tried, using the RHEL ISO and a locally running proxy.

    minishift --username service-manager@mailinator.com --password service-manager start --vm-driver virtualbox     --iso-url file:///users/hardy/downloads/minishift-rhel7.iso --http-proxy http://192.168.1.82:8080
    starting local openshift cluster using 'virtualbox' hypervisor...
    provisioning openshift via '/users/hardy/.minishift/cache/oc/v1.4.1/oc [cluster up --use-existing-config     --host-config-dir /var/lib/minishift/openshift.local.config --host-data-dir /var/lib/minishift/hostdata     --http-proxy http://192.168.1.82:8080]'
    E0207 19:54:09.783520   32691 start.go:330] Error starting the cluster:  exit status 1

So this fails. My guess is that OpenShift 1.4.1 does not know about proxy option. That said, it reminded me that we wanted to do some proper error handling here as well. Seems we forgot about this completely. 

Next I tried:

      minishift --username service-manager@mailinator.com --password service-manager start   --openshift-version  v1.5.0-alpha.2 --vm-driver virtualbox --iso-url file:///Users/hardy/Downloads/  minishift-rhel7.iso  --http-proxy http://192.168.1.82:8080
       tarting local OpenShift cluster using 'virtualbox' hypervisor...
       rovisioning OpenShift via '/Users/hardy/.minishift/cache/oc/v1.5.0-alpha.2/oc [cluster up        -use-existing-config --host-config-dir /var/lib/minishift/openshift.local.config --host-data-dir /var/   lib/    inishift/hostdata --http-proxy http://192.168.1.82:8080]'
       - Checking OpenShift client ... OK
       - Checking Docker client ... OK
       - Checking Docker version ... OK
       - Checking for existing OpenShift container ... OK
       - Checking for openshift/origin:v1.5.0-alpha.2 image ...
         Pulling image openshift/origin:v1.5.0-alpha.2
       AIL
         Error: error pulling Docker image openshift/origin:v1.5.0-alpha.2
         Caused By:
           Error: Error while pulling image: Get https://index.docker.io/v1/repositories/openshift/origin/   images: x509: certificate signed by unknown authority
       0207 20:28:15.364821   34029 start.go:330] Error starting the cluster:  exit status 1   

Better I guess. Something is happening. I am willing to believe that something does not work with my proxy I am trying locally. Need to dig a bit deeper. Something which is deeply missing here is whether registration occurs and whether or not it is successful. I think we discussed this, right? If there is no other issue for that we should create one.

Last but not least, I tried:

    minishift --username service-manager@mailinator.com --password service-manager start -- openshift-version v1.5.0-alpha.2 --vm-driver virtualbox --iso-url file:///Users/hardy/Downloads/minishift-rhel7.iso

That actually worked, so ISO and OpenShift version don't seem to be the problem. 

So right now we for seem to miss the code which handles the availability of the http-proxy option on _oc_. That we already discussed. 

For my second case, I am not sure whether there is something wrong in the way we set everything up or whether I just do something wrong in my proxy setup. Have you tried this end to end?

hferentschik:
@praveenkumar ahh, and I think you need to run _gendocs_

hferentschik:
For what its worth, if I ssh into the VM and execute _curl http://example.com_ I can see that this goes through the proxy. So this part of setting the proxy env on the VM seems to work.

hferentschik:
$sudo systemctl show --property Environment docker
    Environment=HTTP_PROXY=http://192.168.1.82:8080 NO_PROXY=localhost,127.0.0.1,172.30.1.1

So it seems the proxy setting make it into the Docker config

praveenkumar:
@hferentschik I also tried with proxy and looks like everything worked as expected (shell, docker, oc and registration)

```
$ ./minikube start --http-proxy=http://fedora:fedora123@10.70.49.109:3128 --iso-url=file:/Users/prkumar/minishift-rhel7.iso --show-libmachine-logs --cpus 4 --memory 4096 --openshift-version v1.5.0-alpha.2 
Starting local OpenShift cluster using 'xhyve' hypervisor...
Running pre-create checks...
...
Provisioning OpenShift via '/Users/prkumar/.minishift/cache/oc/v1.5.0-alpha.2/oc [cluster up --use-existing-config --host-config-dir /var/lib/minishift/openshift.local.config --host-data-dir /var/lib/minishift/hostdata --http-proxy http://fedora:fedora123@10.70.49.109:3128]'
-- Checking for openshift/origin:v1.5.0-alpha.2 image ... 
   Pulling image openshift/origin:v1.5.0-alpha.2
   Pulled 0/3 layers, 3% complete
   Pulled 0/3 layers, 55% complete
   Pulled 1/3 layers, 72% complete
   Pulled 2/3 layers, 100% complete
   Pulled 3/3 layers, 100% complete
   Extracting
...

$ ./minishift ssh
Last login: Wed Feb  8 00:48:57 2017 from gateway
[docker@minishift ~]$ env | grep proxy
http_proxy=http://fedora:fedora123@10.70.49.109:3128

$ sudo systemctl show --property Environment docker
Environment=HTTP_PROXY=http://fedora:fedora123@10.70.49.109:3128 NO_PROXY=localhost,127.0.0.1,172.30.1.1

$ sudo subscription-manager version
server type: Red Hat Subscription Management
subscription management server: 0.9.51.11-1
subscription management rules: 5.15
subscription-manager: 1.17.15-1.el7
python-rhsm: 1.17.9-1.el7
```
$ cat /var/lib/minishift/openshift.local.config/master/master-config.yaml | grep http 
        - name: https_proxy
          value: http://fedora:fedora123@10.70.49.109:3128
        - name: http_proxy
          value: http://fedora:fedora123@10.70.49.109:3128
        gitHTTPProxy: http://fedora:fedora123@10.70.49.109:3128
```

**Comment by praveenkumar** [comment]
> Have you tried proxy registration end to end?

Yes and it do work.

> Will this implicitly fix #287 since we are setting the proxy settings in the environment. I know we talked about this, but I cannot remember what we said.

Yes it will implicitly fix #287 

> That said, it reminded me that we wanted to do some proper error handling here as well. Seems we forgot about this completely.

I didn't forget but that is something will handle in a separate issue. Also today morning I did some experiment about how we can implement a map/list structure out of `oc cluster -h` but looks like it need lot of parsing and not reliable, let me know if you have any idea regarding it.

> ahh, and I think you need to run gendocs

Thanks, forgot to run it for last change. Done now.

**Comment by hferentschik** [comment]
>> Have you tried proxy registration end to end?
> Yes and it do work.

hmm, wonder what the issue is then. Might be my proxy setup. I would love to get an integration test for this, running some sort of mock proxy. Even if we use the current integration test setup. Porting should be straight forward later.

> Yes it will implicitly fix #287

Cool

> I didn't forget but that is something will handle in a separate issue. 

which one? Really it belongs here. It is part of providing a complete/consistent solution 


> Also today morning I did some experiment about how we can implement a map/list structure out of oc cluster -h but looks like it need lot of parsing and not reliable

Hmm, the output of 'oc cluster up -h' seems straight forward. Why do you think extracting the options is a problem?

**Comment by hferentschik** [comment]
> oc cluster -h but looks like it need lot of parsing and not reliable

So why not just take everything under _Options:_ and then for each line match what's between '--' and '='. I am not saying that is is super stable, but I also don't see this breaking so easily.

**Comment by hferentschik** [comment]
OOI, how will openshift.org link into it? How does one get to http://openshift.org/minishift if not via the redirect from minishift.io

**Comment by hferentschik** [comment]
Also note, https://www.openshift.org/vm/ has changed after the deprecation of the all-in-one. So there is some duplication here. IMO,  https://www.openshift.org/vm/ should change as well and basically just mention that the old Vagrant based approach is not longer supported. Then it adds a link to https://www.openshift.org/minishift.

**Comment by hferentschik** [comment]
@praveenkumar

I tried:

    $ minishift --username service-manager@mailinator.com --password service-manager start --openshift-version v1.5.0-alpha.2 --vm-driver virtualbox --iso-url file:///Users/hardy/Downloads/minishift-rhel7.iso --http-proxy=http://fedora:fedora123@10.70.49.109:3128

That worked for me as well. If I get some time, I will figure out what's wrong with my local approach. I really would like to see/proof that the traffic goes through the proxy. That this command worked it not a proof per se yet.

I can also confirm that all the various tests mentioned above worked as well.

File: pkg/util/utils.go [hunk 26]
```diff
@@ -101,3 +107,34 @@ func (m MultiError) ToError() error {
 	}
 	return fmt.Errorf(strings.Join(errStrings, "\n"))
 }
+
+func ParseOcHelpCommand(cmdOut []byte) []string {
```
hferentschik: Instead of putting this into a generic _utils_ package, we maybe should put this into a new oc package?

pkg/util/utils.go
```diff
@@ -101,3 +107,34 @@ func (m MultiError) ToError() error {
...
+	ocOptions := []string{}
```
hferentschik: Here we go. Looks quite simple, no? Just a couple of regular expressions :-)

pkg/util/utils.go
```diff
@@ -101,3 +107,34 @@ func (m MultiError) ToError() error {
...
+	ocOptionRegex := regexp.MustCompile(`(?s)Options(.*)OpenShift images`)
+	matches := ocOptionRegex.FindSubmatch(cmdOut)
+	if matches != nil {
+		tmpOptionsList := string(matches[0])
+		for _, value := range strings.Split(tmpOptionsList, "\n")[1:] {
+			tmpOption := strings.Split(strings.Split(strings.TrimSpace(value), "=")[0], "--")
+			if len(tmpOption) > 1 {
+				if tmpOption[1] == "env" {
```
hferentschik: this is not so good imo. This creates a dependency to the start command. Instead of hard coding this, this method should take a map with the existing mappings (basically _minishiftToClusterUp_) and do this dynamically. This way, we can for example add another option re-map to the start command and all will continue to work.

pkg/util/utils.go
```diff
@@ -101,3 +107,34 @@ func (m MultiError) ToError() error {
...
+				}
+			}
+		}
+	} else {
+		return nil
+	}
+	return ocOptions
+}
+
+func IsFlagExist(ocCommandOptions []string, flag string) bool {
```
hferentschik: _FlagExist_ or _IsFlag_, _IsFalgExist_ sounds odd to me

hferentschik: this method would also be in this _oc_ package

agajdosi:
Tried today on Fedora:
```
[agajdosi@dhcp bin]$ ./minishift start --http-proxy "http://fedora:fedora123@10.70.49.109"
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.0/minishift-b2d.iso'
 40.00 MB / 40.00 MB [=============================================================================================================================================] 100.00% 0
 22.37 MB / 22.37 MB [=============================================================================================================================================] 100.00% 0
E0209 14:00:05.119716   32132 start.go:317] Flag http-proxy is not supported for current oc verison
```
version of oc in .minishift/cache/oc is v1.4.1.

hferentschik:
@agajdosi that's expected. 1.4.1 does not have the proxy support. You need _v1.5.0-alpha.2_:

     $ minishift --username service-manager@mailinator.com --password service-manager start --openshift-version v1.5.0-alpha.2 --vm-driver virtualbox --iso-url file:///Users/hardy/Downloads/minishift-rhel7.iso --http-proxy=http://fedora:fedora123@10.70.49.109:3128 --https-proxy=http://fedora:fedora123@10.70.49.109:3128

cmd/minikube/cmd/start.go
```diff
@@ -235,6 +319,10 @@ func clusterUp(config *cluster.MachineConfig) {
 			if exists {
 				key = minishiftToClusterUp[key]
 			}
+			if !ocSupportFlag(cmdName, key) {
+				glog.Errorf("Flag %s is not supported for current oc verison", flag.Name)
```
hferentschik: Typo in 'version'. I am going to take care of this while merging

cmd/minikube/cmd/start.go
```diff
@@ -235,6 +319,10 @@ func clusterUp(config *cluster.MachineConfig) {
 			if exists {
 				key = minishiftToClusterUp[key]
 			}
+			if !ocSupportFlag(cmdName, key) {
```
hferentschik: as we discussed, it would be much nicer for the user to fail earlier. So I'd have preferred to first download oc and do some validation, but let's go with this for now. Hopefully once we align  with 1.5 not so many people will run into this anyways.

pkg/util/oc_parse_test.go
```diff
@@ -0,0 +1,113 @@
...
+Use "oc options" for a list of global command-line options (applies to all commands).`)
+
+	expectedOptions = []string {"create-machine", "docker-machine", "env",
+		"forward-ports", "host-config-dir", "host-data-dir", "host-volumes-dir",
+		"image", "logging", "metrics", "public-hostname", "routing-suffix", "server-loglevel",
+		"skip-registry-check", "use-existing-config", "version"}
+)
+
+func TestParseOcHelpCommand(t *testing.T) {
+	recievedOptions := ParseOcHelpCommand(ocClusterUpHelp)
```
hferentschik: :+1:

pkg/util/oc_parse_test.go
```diff
@@ -0,0 +1,113 @@
...
+	assertCompareSlice(expectedOptions, recievedOptions, t)
+}
+
+func TestIsFlagExist(t *testing.T) {
+	if ! FlagExist(expectedOptions, "image") {
```
hferentschik: :+1:

hferentschik:
Merged, thanks. Created issue #374 to add the doc changes.

PR #377, opened 2017-02-09
Issue #378 Add virsh commands to allow cleaning-up the VMs

BrunoVernay:
It would have helped me a lot


hferentschik:
ok to test

hferentschik:
Awesome, thanks for the feedback. Much appreciated. Merged into master.
FYI, I created an issue and adjusted the commit message to make sure this will be properly tracked.

Issue #276, opened 2017-01-11
Provide command to customize OpenShift master configuration


hferentschik:
This will allow a generic solution for issue #257 as well

hferentschik:
My current thinking around this is to utilise the 'openshift ex config patch' - https://github.com/openshift/origin/pull/9165/files. Basically we pass the ability to use this out to the user via a

    $ minishift openshift patch -f <patch-file>

Internally we will then use the SSH communicator to communicate with the VM and execute appropriate docker commands to apply the patch and potentially restart all required containers.

For that we probably need a bit of infrastructure to make this docker command execution easy to use and re-usable for other use cases.

jorgemoralespou:
@hferentschick this is what we do in the wrapper script. Please take a look
for example at blob/master/plugins.d/pipelines.local.plugin at the install
method.

hferentschik:
See also issue #25

jorgemoralespou:
@hferentschik I would try to encapsulate that configuration that a
developer would do and possibly provide a way for them to apply any config.

In the following example, set would have a set of well known
customizations, and patch would apply any (which could potentially break
the config).

E.g
minishift config set oauth=htpasswd
minishift config patch --file=mycustomconfig.json

hferentschik:
> I would try to encapsulate that configuration that a
developer would do and possibly provide a way for them to apply any config.

But that puts us into the position of knowing all the possible customisation and keeping in sync with whatever OpenShift does. We then can end up in situations where the user ask why customisation xyz is not supported yet. In the end one would need the generic approach as well as a "power user" feature.

I think, I'd prefer the generic approach first and then maybe add on on this and provide common known customisations as sort of "alias" where we keep/maintain the required patch under the hoods.

> which could potentially break the config

true, I think we can alleviate this a bit by providing a rollback function of sorts. I have seen the code in your wrapper scripts. Something along these lines, but with the ability to revert a applied patch by rolling back the modified config file. WDYT?

hferentschik:
from https://github.com/minishift/minishift/issues/207#issuecomment-274168431
> My testing also seems to have uncovered an issue with openshift ex config patch command run 
> inside of a container. If MiniShift was going to use that to patch the master config file, there is a 
> chance could trip the same problem depending on what you are doing. The issue appears to be in > upstream code, so don't know at this point when/if might be fixed or whether need to start doing > yaml parsing in the wrapper to make changes to master config file.

hferentschik:
[...]

Issue #254, opened 2017-01-05
Expose Openshift Docker Registry


glennodickson:
Environment:
minishift 0.9.0
OS : Windows 10
Docker client : 1.9.1

I'm trying to login/push into Openshift Integrated Docker Registry. Below I have describe both of these and their errors. **If anyone can help me through this I would be most appreciative.**

# Started Minishift
I've started minishift (not is used correctly) from the window command prompt using:
`minishift start --insecure-registry 172.30.0.0/16    --insecure-registry minishift 
`

# Registry Service 
```
c:\..\..\>oc describe service docker-registry
Name:                   docker-registry
Namespace:              default
Labels:                 docker-registry=default
Selector:               docker-registry=default
Type:                   NodePort
IP:                     172.30.225.22
Port:                   5000-tcp        5000/TCP
NodePort:               5000-tcp        30832/TCP
Endpoints:              172.17.0.2:5000
Session Affinity:       ClientIP
No events.
```

# Registry Route:
```
apiVersion: v1
kind: Route
metadata:
  name: docker-registry
  namespace: default
  selfLink: /oapi/v1/namespaces/default/routes/docker-registry
  uid: 9f7eee66-d299-11e6-89fb-42527179d16d
  resourceVersion: '88756'
  creationTimestamp: '2017-01-04T16:19:52Z'
  labels:
    docker-registry: default
  annotations:
    openshift.io/host.generated: 'true'
spec:
  host: docker-registry-default.192.168.99.100.xip.io
  to:
    kind: Service
    name: docker-registry
    weight: 100
  port:
    targetPort: 5000-tcp
  tls:
    termination: edge
    insecureEdgeTerminationPolicy: Allow
status:
  ingress:
    -
      host: docker-registry-default.192.168.99.100.xip.io
      routerName: router
      conditions:
        -
          type: Admitted
          status: 'True'
          lastTransitionTime: '2017-01-04T16:19:52Z'
```
# Login
Attempt to login from the window command prompt using:
 ` docker login  -u admin -p admin -e email@email.com docker-registry-default.192.168.99.100.xip.io`

Get this error:
 
> Error response from daemon: Get **https**://docker-registry-default.192.168.99.100.xip.io/v1/users/: x509: certificate is valid for router.default.svc, router.default.svc.cluster.local, not docker-registry-default.192.168.99.100.xip.io

# Push
List Docker images from WIndows command prompt:
```
c::\....\..>docker images
[...]

PR #379, opened 2017-02-09
Issue #276 Creating 'minishift openshift [config|patch|restart]' commands

hferentschik:
This is just a preview of where things are heading. Most is in place, but there are several loose ends. 

This pull request will add _minishift openshift [config|patch|restart]_ with various flags. Most likely there will in the end be also a _status_ and I think we can eventually also add _registry_  (issue #254).

The pull request should compile and successfully pass the tests. Also _config_ and _restart_ work. _patch_ works when one knows what to do ;-) (WIP)

Also I added a atexit handler allowing us to collect all os.exit calls into one place and potentially add some hooks (useful for testing).


cmd/minishift/cmd/root.go
```diff
@@ -27,8 +27,8 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
-
-	configCmd "github.com/minishift/minishift/cmd/minikube/cmd/config"
+
```
praveenkumar: Looks like this extra line already have a tab or space.

docs/using.md
```diff
@@ -50,6 +51,36 @@ Starting Minishift again will restore the cluster, allowing you to continue work
...
+
+```
+$  minishift openshift patch --patch '{"corsAllowedOrigins": [".*"]}'
+```
+
+Per default the master configuration is targeted, but you can also path the node config
+by specifying the `target` flag. For more details about the
+`patch` command refer to its [synopsis](./minishift_openshift_patch.md).
+
 <a name="environment-variables"></a>
```
LalatenduMohanty: I tried below commands and diff does not show me much difference. Not sure if it is expected. 
```
$ minishift openshift config > config1
$ minishift openshift patch --patch '{"corsAllowedOrigins": [".*"]}'
Patching OpenShift configuration /var/lib/origin/openshift.local.config/master/master-config.yaml with {"corsAllowedOrigins": [".*"]}

$ minishift openshift config > config2
```

```
$ diff config1 config2
65,67c65
< - 127.0.0.1
< - 192.168.64.7:8443
< - localhost
---
> - .*
225a224
> 
```

LalatenduMohanty: As we are restarting OpenShift after patching it , we should mention that in the documentation and also in output of `minishift openshift patch --patch <>`. As this would keep the operation we are doing in the cluster transparent to user IMO.

pkg/minishift/openshift/openshift.go
```diff
@@ -0,0 +1,221 @@
...
+
+	// Tweak the result configuration, we need to escape single quotes
+	result = strings.Replace(result, "'", "'\\''", -1)
+	writeConfig(target, result, commander)
+	if err != nil {
+		rollback(target, commander, patchId)
+		return false, nil
+	}
+
+	_, err = commander.Restart(OPENSHIFT_CONTAINER_NAME)
```
LalatenduMohanty: I think this should be `openshift.RestartOpenShift(dockerCommander)`

LalatenduMohanty:
As an user I would love if I can save output of `openshift config` -> Modify it (as required) -> then use the file to patch the config. WDYT?

cmd/minishift/cmd/root_test.go
```diff
@@ -77,9 +78,13 @@ var configTests = []cli.TestOption{
 
 func TestPreRunDirectories(t *testing.T) {
 	// Make sure we create the required directories.
-	tempDir := tests.MakeTempDir()
-	defer os.RemoveAll(tempDir)
+	testDir, err := ioutil.TempDir("", "minishift-test-start-cmd-")
```
hferentschik: no - https://golang.org/pkg/io/ioutil/#TempDir
Same behaviour as in Ruby

cmd/minishift/cmd/openshift/config.go
```diff
@@ -0,0 +1,89 @@
...
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/docker"
+	"github.com/minishift/minishift/pkg/minishift/openshift"
+	"github.com/minishift/minishift/pkg/minishift/util"
+)
+
+const (
+	configTargetFlag         = "target"
+	unknownConfigTargetError = "Unkown config target. Only 'master' and 'node' are supported."
+	unableToRetrieveIpErrror = "Unable to retrieve IP of VM."
```
hferentschik: ok.

docs/using.md
hferentschik: > I tried below commands and diff does not show me much difference. Not sure if it is expected.

LOL, what do you expect? You effectively changed the default CORS from allowing only requests from localhost and 192.168.64.7:8443 to all domains (.*). And yes, you just changed a single config option :-)

hferentschik: >  we are restarting OpenShift after patching it , we should mention that in the documentation and also in output of minishift openshift patch --patch <>

Ok. Will do.

pkg/minishift/openshift/openshift.go
hferentschik: Because you want to have the output to the stdout? I guess for the initial attempt that's ok, in the rollback I intentionally call the commander directly, since the output gets confusing otherwise.

hferentschik:
> As an user I would love if I can save output of openshift config -> Modify it (as required) -> then use the modified file to patch the config. WDYT?

This would require to convert it to JSON. I see where you are coming from. I guess that would be a valid extension of what we have. I would see that as a new feature though. At some stage we need to get the basics in place.

LalatenduMohanty:
> This would require to convert it to JSON. I see where you are coming from. I guess that would be a valid extension of what we have. I would see that as a new feature though. At some stage we need to get the basics in place.

Sure. Lets create an issue track it there.

cmd/minishift/cmd/root_test.go
LalatenduMohanty: Ok. cool

docs/using.md
LalatenduMohanty: > LOL, what do you expect? You effectively changed the default CORS from allowing only requests from localhost and 192.168.64.7:8443 to all domains (.*). And yes, you just changed a single config option :-)
> @hferentschik

Good to know.

pkg/minishift/openshift/openshift.go
LalatenduMohanty: Yes, because I think printing we are restarting the cluster to stdout will be better.

hferentschik:
Merged, thanks

PR #382, opened 2017-02-09
About "successfully" login as system/admin

BrunoVernay:
The "Login successful" message is error prone.

### DIFF
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -99,3 +99,13 @@ If you are using Windows, the problem is likely to be either an outdated version
 of Virtual Box or you forgot to use `--vm-driver virtualbox` option when starting minishift.
 
 We recommend to use `Virtualbox >= 5.1.12` to avoid this issue.
+
+## oc login:  User "system" cannot ...
+
+You may succesfully login like this:
+```
+> oc login -u system -p admin
+Login successful.
+```
+and later find out that you have no rights!
+Indeed, the system account is `oc login -u system:admin` and it will NOT ask for a password, it will use the key in `~/.kube/config` (created during installation).


praveenkumar:
Hey @BrunoVernay Thanks for patch. We do let user know how to login as admin once minishift start finish as below why do you think user take that route (want to understand)?

```
       https://192.168.64.2:8443

   You are logged in as:
       User:     developer
       Password: developer

   To login as administrator:
       oc login -u system:admin
```

praveenkumar:
retest this please

hferentschik:
@BrunoVernay , thanks for the pull request. You are right, our docs do not mention anything about the default users. We just have this output as part of the bootstrapping. Good to get your feedback.

I don't think, however, that this is a case for troubleshooting. Instead we should probably add a paragraph or two to the usage guide - https://github.com/minishift/minishift/blob/master/docs/using.md. There we can explain about the two default users and what they do. WDYT?

I created https://github.com/minishift/minishift/issues/387 to track this. If you are up to it, feel free to give it a go.

Are you ok, if we close this pull request in the meantime?

BrunoVernay:
If everything goes well, I think the user will copy/paste the line "oc login -u system:admin" as-is. Not sure if he will bother reading the doc.

In case things goes bad, history shows that he could try to login as system/admin and he will succeed.  So he still won't read the doc, because the login is "successful" :-)

Only when other errors arise, will he go to the documentation. Even then, he might not head to the "login" section. 

That being said, I agree the documentation is concise and one should read it all easily. (But you guess that my point is that the "troubleshooting" would be more suited).    All in all it won't made such a difference, so **do for the best**. _I guarantee that I won't be upset, you are doing an amazing job already._

hferentschik:
{quote}
If everything goes well, I think the user will copy/paste the line "oc login -u system:admin" as-is. Not sure if he will bother reading the doc.
{quote}

Sure, in the best case one does that. Still the docs should be complete. Also I think it would be valuable to have a couple of notes regarding the differences in these accounts with potential links to the OpenShift docs.

{quote}
In case things goes bad, history shows that he could try to login as system/admin and he will succeed. So he still won't read the doc, because the login is "successful" :-)
{quote}

Right, so maybe we need an entry in troubleshooting as well for good measure.

{quote}
That being said, I agree the documentation is concise and one should read it all easily. (But you guess that my point is that the "troubleshooting" would be more suited). 
{quote}

Let's do both :-)

{quote}
All in all it won't made such a difference, so do for the best. I guarantee that I won't be upset, you are doing an amazing job already.
{quote}
Thanks :-)

PR #384, opened 2017-02-10
Issue #318 Incomplete handling of VM status for un-registration

praveenkumar:
This patch is mostly refactor registration/unregistration, now those capability is part of cluster pkg instead of cmd. Another change is we are now using `state.Running` as condition for register and unregister task so it will handle all different state of VM (not only stopped).

### DIFF
--- a/cmd/minikube/cmd/delete.go
+++ b/cmd/minikube/cmd/delete.go
@@ -21,11 +21,8 @@ import (
 	"os"
 
 	"github.com/docker/machine/libmachine"
-	"github.com/docker/machine/libmachine/drivers"
-	"github.com/docker/machine/libmachine/state"
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
-	"github.com/minishift/minishift/pkg/minishift/registration"
 	"github.com/spf13/cobra"
 )
 
@@ -38,19 +35,6 @@ var deleteCmd = &cobra.Command{
 		fmt.Println("Deleting the Minishift VM...")
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
-		host, err := api.Load(constants.MachineName)
-		if err != nil {
-			fmt.Println("Error occurred while deleting the VM: ", err)
-			os.Exit(1)
-		}
-
-		if !drivers.MachineInState(host.Driver, state.Stopped)() {
-			// Unregister Host VM
-			if err := registration.UnregisterHostVM(host, RegistrationParameters); err != nil {
-				fmt.Printf("Error unregistring the VM: %s", err)
-				os.Exit(1)
-			}
-		}
 
 		if err := cluster.DeleteHost(api); err != nil {
 			fmt.Println("Error deleting the VM: ", err)
--- a/cmd/minikube/cmd/root.go
+++ b/cmd/minikube/cmd/root.go
@@ -29,9 +29,9 @@ import (
 	"github.com/spf13/viper"
 
 	configCmd "github.com/minishift/minishift/cmd/minikube/cmd/config"
+	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/config"
 	"github.com/minishift/minishift/pkg/minikube/constants"
-	"github.com/minishift/minishift/pkg/minishift/registration"
 )
 
 var dirs = [...]string{
@@ -51,10 +51,6 @@ const (
 	showLibmachineLogs = "show-libmachine-logs"
 	username           = "username"
 	password           = "password"
-)
-
-var (
-	RegistrationParameters = new(registration.RegistrationParameters)
 )
 
 var viperWhiteList = []string{
@@ -153,8 +149,8 @@ func setupViper() {
 }
 
 func setRegistrationParameters() {
-	RegistrationParameters.Username = viper.GetString("username")
-	RegistrationParameters.Password = viper.GetString("password")
+	cluster.RegistrationParameters.Username = viper.GetString("username")
+	cluster.RegistrationParameters.Password = viper.GetString("password")
 }
 
 func ensureConfigFileExists(configPath string) {
--- a/cmd/minikube/cmd/stop.go
+++ b/cmd/minikube/cmd/stop.go
@@ -21,11 +21,8 @@ import (
 	"os"
 
 	"github.com/docker/machine/libmachine"
-	"github.com/docker/machine/libmachine/drivers"
-	"github.com/docker/machine/libmachine/state"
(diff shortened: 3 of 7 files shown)


cmd/minikube/cmd/root.go
```diff
@@ -53,9 +53,7 @@ const (
 	password           = "password"
 )
 
-var (
-	RegistrationParameters = new(registration.RegistrationParameters)
-)
+var ()
```
budhrg: Unwanted code

cmd/minikube/cmd/delete.go
```diff
@@ -38,19 +35,6 @@ var deleteCmd = &cobra.Command{
 		fmt.Println("Deleting the Minishift VM...")
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
-		host, err := api.Load(constants.MachineName)
-		if err != nil {
```
hferentschik: :+1:

cmd/minikube/cmd/root.go
```diff
@@ -153,8 +149,8 @@ func setupViper() {
 }
 
 func setRegistrationParameters() {
-	RegistrationParameters.Username = viper.GetString("username")
-	RegistrationParameters.Password = viper.GetString("password")
+	cluster.RegistrationParameters.Username = viper.GetString("username")
```
hferentschik: :+1:

cmd/minikube/cmd/stop.go
```diff
@@ -40,20 +37,6 @@ VM but does not delete any associated files. To start the cluster again, use the
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
 
-		host, err := api.Load(constants.MachineName)
-		if err != nil {
```
hferentschik: :+1:

pkg/minikube/cluster/cluster.go
```diff
@@ -101,6 +101,11 @@ func StartHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 	if err := h.ConfigureAuth(); err != nil {
 		return nil, fmt.Errorf("Error configuring authorization on host: %s", err)
 	}
+
+	if err := registration.RegisterHostVM(h, RegistrationParameters); err != nil {
```
hferentschik: :+1:

pkg/minikube/cluster/cluster.go
```diff
@@ -110,6 +115,11 @@ func StopHost(api libmachine.API) error {
 	if err != nil {
 		return err
 	}
+
```
hferentschik: :+1:

pkg/minikube/cluster/cluster.go
```diff
@@ -122,12 +132,27 @@ func DeleteHost(api libmachine.API) error {
...
```
hferentschik: :+1:

pkg/minikube/cluster/cluster.go
```diff
@@ -122,12 +132,27 @@ func DeleteHost(api libmachine.API) error {
...
 	m := util.MultiError{}
 	m.Collect(host.Driver.Remove())
 	m.Collect(api.Remove(constants.MachineName))
 	return m.ToError()
 }
 
+// Un-register host VM
+func unregister(host *host.Host) error {
+	if drivers.MachineInState(host.Driver, state.Running)() {
```
hferentschik: Hmm, I think here we are falling short a bit. What is the machine is not running and for some reason in one of the error states (for example after hibernation). This will silently go ahead deleting the VM w/o un-registrating it. 

Should we not in the case where it is not running, at least print a warning that it was not possible to un-register?

pkg/minishift/registration/utils.go
```diff
@@ -42,6 +42,7 @@ func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
 	}
 
 	if registrator != nil {
+		fmt.Println("Registering machine using subscription-manager")
```
hferentschik: yeah :-)

pkg/minikube/cluster/cluster.go
praveenkumar: > This will silently go ahead deleting the VM w/o un-registrating it.

No it will no un-register VM silently because before unregister it need to check if vm is support registration and execute ssh command `cat /etc/os-release` so if a vm is any other state than running it will display error about vm status and above command is not executable on vm.

praveenkumar:
> Looking good. Just wondering whether we should not at least log something when we cannot un-register, because the VM is not running.

@hferentschik if VM is not running (at least not in the stopped condition) then we can't even delete the VM because of that ssh commander execution.

pkg/minikube/cluster/cluster.go
hferentschik: I don't get it, so you want to rely on the code which checks whether un-registration is even required (based on the type of VM). AFAICT this code lives in _registration.UnregisterHostVM_ which inside the if statement checking whether the VM is running. 

> so if a vm is any other state than running it will display error about vm status and above command is not executable on vm.

If the VM is not running, the registrar check is not even occurring, right?

I guess that one could argue that the delete as whole will fail, due to _host.Driver.Remove()_ or _api.Remove(constants.MachineName)_ failing, but looking at the docs for _api.Remove_ it seems open how exactly the driver handles this.

Am I missing something here?

hferentschik: ahh, but without a running VM we cannot test whether the VM is registered, so we cannot know whether and what to report to the user. 

Really, we should not rely on the VM to determine whether it is registered. Instead we probably should write a config entry during start that the VM was registered and check for this entry during deletion. This was we don't have to rely on a running VM to even detect whether un-registration is necessary. WDYT?

praveenkumar: @hferentschik Agree, should we use config file which we create at the beginning for this purpose or create a another one?

hferentschik: > Agree,

Cool :-) Shall we create a new issue for this or do you want to continue on this pull request?

> should we use config file which we create at the beginning for this purpose or create a another one

I always thought about re-suing the existing one. Maybe we can start with that and see how it goes. If it creates problems we can always create a separate one. 

Basically I thought we read the config file as in _root.go_ (we already do as part of the Viper config) and keep the landed config in memory behind some Config interface. When we modify the config we flush. For this particulate case we would write a _isRegisteredVM_ flag. Maybe we put this into a nested _currentVM_ context. We will also need the current _oc_ to be used persistent in there. This is required for the _oc env_ command as well as for the registry exposure (issue #254).

hferentschik: Here is the follow up issue.

hferentschik:
Rebased and merged, thanks.

Issue #375, opened 2017-02-09
Add validation for minimal supported OpenShift version in start command 


hferentschik:
We should add some simple validation to the start command for the _openshift-version_ flag. We need to determine the minimum version required and validate that no earlier version is selected.

hferentschik:
@praveenkumar do you want to take care of this while you are on this?

hferentschik:
While on it, I would like to get rid of the _get-openshift-versions_ command (not the underlying functionality, just the command). There is no value in this for a minishift user in this form. I'd rather use this in conjunction with the idea of interactive selection of the OpenShift version to run, but one needs to apply some filters in this case. Printing all releases does not make much sense.

praveenkumar:
@hferentschik which version we should chose for validation (after 1.3.0 ? )

praveenkumar:
> . I'd rather use this in conjunction with the idea of interactive selection of the OpenShift version to run, but one needs to apply some filters in this case. Printing all releases does not make much sense.

Can we just show the version info which we support using minishift? I mean if we go with > v1.3.1 then show only those versions which are greater than 1.3.1 ?

hferentschik:
Merged via pull request #398. 

Created follow up issues #410 and #411

PR #398, opened 2017-02-14
Issue #375 Add validation for minimal supported OpenShift version in start command

praveenkumar:
Addresses issue #375

This patch will add a validation for supported openshift version (>v1.3.3) also same validation go for `get-openshift-versions`.

Todo
====

- Add test case for validation method
- Changes according to review.

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -123,6 +123,8 @@ func runStart(cmd *cobra.Command, args []string) {
 	libMachineClient := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer libMachineClient.Close()
 
+	validateOpenshiftVersion()
+
 	validateProxyArgs()
 
 	setDockerProxy()
@@ -333,6 +335,16 @@ func clusterUp(config *cluster.MachineConfig) {
 		// TODO glog is probably not right here. Need some sort of logging wrapper
 		glog.Errorln("Error starting the cluster: ", err)
 		os.Exit(1)
+	}
+}
+
+func validateOpenshiftVersion() {
+	if viper.IsSet(openshiftVersion) {
+		if !util.ValidateOpenshiftMinVersion(viper.GetString(openshiftVersion)) {
+			fmt.Printf("Minishift does not support Openshift version %s. "+
+				"You need to use a version >= 1.3.1\n", viper.GetString(openshiftVersion))
+			os.Exit(1)
+		}
 	}
 }
 
--- a/glide.yaml
+++ b/glide.yaml
@@ -39,6 +39,7 @@ import:
   version: 382f87b929b84ce13e9c8a375a4b217f224e6c65
 - package: github.com/xeipuuv/gojsonschema
 - package: github.com/blang/semver
+  version: 3.5.0
 - package: github.com/docker/go-units
 - package: github.com/pkg/errors
   version: ^0.8.0
--- a/pkg/minikube/constants/constants.go
+++ b/pkg/minikube/constants/constants.go
@@ -38,6 +38,9 @@ const MiniShiftEnvPrefix = "MINISHIFT"
 // MiniShiftHomeEnv is the environment variable used to change the Minishift home directory
 const MiniShiftHomeEnv = "MINISHIFT_HOME"
 
+const MinSupportedOpenshiftVersion = ">=1.3.1"
+const VersionPrefix = "v"
+
 const (
 	DefaultMemory   = 2048
 	DefaultCPUS     = 2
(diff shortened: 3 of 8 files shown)


glide.yaml
```diff
@@ -39,6 +39,7 @@ import:
   version: 382f87b929b84ce13e9c8a375a4b217f224e6c65
 - package: github.com/xeipuuv/gojsonschema
 - package: github.com/blang/semver
+  version: 3.5.0
```
hferentschik: Nice, you are setting fixed versions. I like :-)

praveenkumar: @hferentschik I put fix version for the package which have releases and tags but some of package which we consume doesn't have tags/release in place :(

hferentschik: Wild West :-(

pkg/version/version.go
```diff
@@ -45,10 +45,6 @@ func GetOpenShiftVersion() string {
 	return openshiftVersion
 }
 
-func GetOpenShiftSemverVersion() (semver.Version, error) {
-	return semver.Make(strings.TrimPrefix(GetOpenShiftVersion(), VersionPrefix))
-}
-
```
praveenkumar: I removed this method because this is not used anywhere in our codebase.

hferentschik: +1, In fact I once missed it. I think it was once used for the auto update which we for now disabled.

cmd/minikube/cmd/start.go
```diff
@@ -336,6 +339,16 @@ func clusterUp(config *cluster.MachineConfig) {
 	}
 }
 
+func validateOpenshiftVersion() {
```
hferentschik: :+1:

pkg/minikube/constants/constants.go
```diff
@@ -38,6 +38,8 @@ const MiniShiftEnvPrefix = "MINISHIFT"
 // MiniShiftHomeEnv is the environment variable used to change the Minishift home directory
 const MiniShiftHomeEnv = "MINISHIFT_HOME"
 
+const SupportedOpenshiftTag = "1.3.3"
```
hferentschik: MinSupportedOpenshiftTag, maybe? We support anything which comes above, right? Or you could add '>' here as well, then you would not need the Sprintf

pkg/minikube/openshiftversions/openshift_versions.go
```diff
@@ -42,7 +44,11 @@ func PrintOpenShiftVersions(output io.Writer) {
 	fmt.Fprint(output, "The following OpenShift versions are available: \n")
 
 	for _, version := range versions {
-		fmt.Fprintf(output, "\t- %s\n", *version.TagName)
+		v, _ := semver.Parse(*version.TagName)
+		rang, _ := semver.ParseRange(fmt.Sprintf(">%s", constants.SupportedOpenshiftTag))
```
hferentschik: _rang_? I guess just 'r' would be better. Or a different name altogether. I guess you drop the 'e' due to name conflict?

pkg/util/utils_test.go
```diff
@@ -76,6 +76,25 @@ func TestValidateProxyURI(t *testing.T) {
 	}
 }
 
+func TestValidateOpenshiftMinVersion(t *testing.T) {
+	verList := map[string]bool{
+		"v1.1.0":         false,
```
hferentschik: :+1:

pkg/minikube/constants/constants.go
```diff
@@ -38,6 +38,9 @@ const MiniShiftEnvPrefix = "MINISHIFT"
 // MiniShiftHomeEnv is the environment variable used to change the Minishift home directory
 const MiniShiftHomeEnv = "MINISHIFT_HOME"
 
+const MinSupportedOpenshiftTag = ">1.3.1"
```
hferentschik: why not '>='? We started with 1.3.1, right?

Right now I get:

    minishift start --openshift-version v1.3.1
    F0215 10:31:04.744867    4867 start.go:337] v1.3.1 is not supported with minishift

hferentschik: I think 'MinSupportedOpenshiftVersion' is better, tag might be confused with a git tag.

cmd/minikube/cmd/start.go
```diff
@@ -336,6 +338,14 @@ func clusterUp(config *cluster.MachineConfig) {
...
+	if viper.IsSet(openshiftVersion) {
+		if !util.ValidateOpenshiftMinVersion(viper.GetString(openshiftVersion)) {
+			glog.Exitf("%s is not supported with minishift", viper.GetString(openshiftVersion))
```
hferentschik: Let's use fmt.Println. For a user facing message we don't want to have all this log gibberish. I guess we could see how to configure glog, but we need a review of the whole logging approach anyways.

hferentschik: Also, let's be a not more verbose: "Minishift does not support Openshift in  version %s. You need to use a version >= 1.3.1.

hferentschik:
Merged, thanks

hferentschik:
I am fine with get-opensfhift-version now that it just returns the supported versions.

PR #399, opened 2017-02-14
Issue #355 Use persistent volume for host-volumes-dir

LalatenduMohanty:

### DIFF
--- a/cmd/minikube/cmd/start.go
+++ b/cmd/minikube/cmd/start.go
@@ -29,11 +29,11 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/cache"
+	"github.com/minishift/minishift/pkg/minishift/openshiftcluster"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
 	minishiftUtil "github.com/minishift/minishift/pkg/minishift/util"
 	"github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/version"
-	dockerhost "github.com/openshift/origin/pkg/bootstrap/docker/host"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
@@ -87,8 +87,9 @@ var startCmd = &cobra.Command{
 
 // Set default value for host data and config dir
 var (
-	hostConfigDirectory = "/var/lib/minishift/openshift.local.config"
-	hostDataDirectory   = "/var/lib/minishift/hostdata"
+	hostConfigDirectory  = "/var/lib/minishift/openshift.local.config"
+	hostDataDirectory    = "/var/lib/minishift/hostdata"
+	hostVolumesDirectory = "/var/lib/minishift/openshift.local.volumes"
 )
 
 // startFlagSet contains the minishift specific command line switches
@@ -171,6 +172,14 @@ func runStart(cmd *cobra.Command, args []string) {
 	envMap, err := cluster.GetHostDockerEnv(libMachineClient)
 	for k, v := range envMap {
 		os.Setenv(k, v)
+	}
+
+	//Create the host directories if not present
+	hostDirs := []string{viper.GetString(hostConfigDir), viper.GetString(hostDataDir), viper.GetString(hostVolumesDir)}
+	err = openshiftcluster.EnsureHostDirectoriesExist(libMachineClient, hostDirs)
+	if err != nil {
+		glog.Errorln("Error creating required host directories: ", err)
+		os.Exit(1)
 	}
 
 	clusterUp(&config)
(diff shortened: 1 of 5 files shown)


cmd/minikube/cmd/start.go
```diff
@@ -87,8 +86,9 @@ var startCmd = &cobra.Command{
 
 // Set default value for host data and config dir
 var (
-	hostConfigDirectory = "/var/lib/minishift/openshift.local.config"
-	hostDataDirectory   = "/var/lib/minishift/hostdata"
+	hostConfigDirectory  = "/var/lib/minishift/openshift.local.config"
+	hostDataDirectory    = "/var/lib/minishift/hostdata"
+	hostVolumesDirectory = "/var/lib/minishift/openshift.local.volumes"
```
hferentschik: So here we set the default, right? So we have knowledge where we things should be. Why can we not check the existence of this path and potentially create it. Also, this is a parameter the user can specify. If 'oc cluster up' does indeed not create the directory, if it is not the default, we would need the check anyways. We would need to retrieve the _hostVolumesDir_ flag value (which would be either the default we specify here or some user specified value) and make sure the directory exists. I don't see how this even can be handled in the ISO.

On that note, I would love to know why _cluster up_ would not create this directory if it is not the default. That in itself seems questionable.

cmd/minikube/cmd/start.go
```diff
@@ -303,6 +303,7 @@ func clusterUp(config *cluster.MachineConfig) {
 	// Set default value for host config and data
 	viper.Set(hostConfigDir, viper.GetString(hostConfigDir))
 	viper.Set(hostDataDir, viper.GetString(hostDataDir))
+	viper.Set(hostVolumesDir, viper.GetString(hostVolumesDir))
```
hferentschik: here we would have to add the validation. The VM is up and running. We can check and create the directory.

cmd/minikube/cmd/start.go
LalatenduMohanty: @hferentschik It is confirmed that `oc cluster up` does not create the directories. So I agree that we need to create the directories if do not exist.

cmd/minikube/cmd/start.go
LalatenduMohanty: Agree.

cmd/minikube/cmd/start.go
hferentschik: +1. Good that we got the bottom of this. Let's get this implemented asap. We might take issue #409 into consideration, but we can defer this as well.

LalatenduMohanty:
@gbraad @hferentschik The code is mostly done except one unit test in case you want to take a look.

cmd/minikube/cmd/start.go
LalatenduMohanty: @hferentschik I will #409 later as this PR is critical for next release and I do not want to delay it.

gbraad:
Currently running tests against both b2d and the CentOS image.

cmd/minikube/cmd/start.go
```diff
@@ -173,6 +173,15 @@ func runStart(cmd *cobra.Command, args []string) {
 		os.Setenv(k, v)
 	}
 
+	//Create the host directories if not present
+	hostDirs := []string{viper.GetString(hostConfigDir), viper.GetString(hostDataDir), viper.GetString(hostVolumesDir)}
+	err = cluster.EnsureHostDirectories(libMachineClient, hostDirs)
+	if err != nil {
+		// TODO glog is probably not right here. Need some sort of logging wrapper
```
hferentschik: Did you cut&paste this comment? I think we can skip this. We know that we have to revisit this.

cmd/minikube/cmd/start.go
```diff
@@ -173,6 +173,15 @@ func runStart(cmd *cobra.Command, args []string) {
 		os.Setenv(k, v)
 	}
 
+	//Create the host directories if not present
+	hostDirs := []string{viper.GetString(hostConfigDir), viper.GetString(hostDataDir), viper.GetString(hostVolumesDir)}
+	err = cluster.EnsureHostDirectories(libMachineClient, hostDirs)
```
hferentschik: _EnsureHostDirectoriesExist_ seems more telling to me.

hferentschik: This is more and more entangling the CLI command line parsing code with the _cluster up_ code. I would love to see issue #409 addressed at some stage.

pkg/minikube/cluster/cluster.go
```diff
@@ -618,3 +618,17 @@ func GetServiceURLs(api libmachine.API, namespace string, t *template.Template)
 
 	return serviceURLs, nil
 }
+
+//Create the host directories in the virtual machine if not present
+func EnsureHostDirectories(api libmachine.API, dirs []string) (error) {
```
hferentschik: I think we should move this somewhere else. _cluster.go_ is big enough as it stands. We could have a vmutils package of sorts, since this method for example is not specific per se to the cluster functionality. It just ensures that a set of directories exit.

Alternatively, we can create _oc_cluster.go_ somewhere under _pkg/minishift/oc/cluster_. This could be where we put other _oc cluster_ stuff as well (see issue #409).

cmd/minikube/cmd/start.go
LalatenduMohanty: @hferentschik Yes, I cut and pasted it. I thought it might make sense when we want to revisit these. But sure I can remove it.

pkg/minikube/cluster/cluster.go
gbraad: +1 agree. This is actually minishift 'specific' knowledge, and not so much cluster. It is something that needs to happen...

gbraad:
@budhrg you were able to reproduce the issue yesterday, could you run once with the centos image? And if possible, if time let's you, once with the b2d image.

cmd/minikube/cmd/start.go
gbraad: On a sidr note, we only need to create the host volume folder, but for completeness we do all.

I agree with the naming. When reviewing just now I had the same problem. It does not read well, as the intention is lost/missing.

gbraad: Note: the EnsureDirectories naming is similar to as used in OpenShift, but I had concerns with that in their code, as the intention was only clear after reading the implementation.

pkg/minikube/cluster/cluster.go
LalatenduMohanty: cluster.go is also a mess, I can see few functions which should not be part of it. I will try to fix some of it in this PR.

LalatenduMohanty: I think minishift/pkg/minishift/util/oc_cluster.go is the right location for this function.

hferentschik:
Merged thanks. Great detective everyone involved 👏

Issue #387, opened 2017-02-10
Document default user credentials and un-install instructions


hferentschik:
The documentation should mention the default users and their credentials. Also a word about privileges and which user to use for what would be good.

Should be added to the Usage guide - https://github.com/minishift/minishift/blob/master/docs/using.md

BrunoVernay:
I can write something if you want

budhrg:
@BrunoVernay We would love to review the PR :smile:

hferentschik:
Merged. See pull request #400 and #401

PR #400, opened 2017-02-14
Issue #387 Note about oc login related to issue

BrunoVernay:

Issue #378, opened 2017-02-09
Add virsh cleanup commands to troubleshooting guide


hferentschik:

hferentschik:
Merged via pull request #377

PR #401, opened 2017-02-14
Issue #387 Add Un-install instructions

BrunoVernay:


README.md
```diff
@@ -100,6 +100,12 @@ You can now install the latest beta version of minishift.
   $ brew cask install minishift-beta
 ```
 
+### Un-installing Minishift
+
+.1 Try to delete the Minishift VM: `minishift delete`
```
budhrg: This is not rendering as expected. See below
screenshot from 2017-02-15 00-33-48

README.md
```diff
@@ -100,6 +100,12 @@ You can now install the latest beta version of minishift.
...
+.1 Remove the folders: `~/.minishift` and `~/.kube`
+.1 Check that there are no remaining Minishift VM. Use KVM (virsh), VirtualBox or Xhyve depending on your driver.
```
LalatenduMohanty: I do not think we have a way to check if there is an existing VM in Xhyve. So we should add a note about Xhyve that user can not findout if the VM is present in Xhyve.

Preeticp:
There is a section on Minishift Delete in Using.md, (https://github.com/minishift/minishift/blob/master/docs/using.md#deleting-openshift) can we add this information there?

hferentschik:
I will take care of this as part of pull request #400 and issue #378

hferentschik:
@BrunoVernay, merged thanks. I added some more reference links as well. 

Thanks again for your help. Let us know if you find more :-)

BrunoVernay:
Yes I just saw, too late, that the changes had been already done.
There is just a typo: "conform" instead of "confirm"

hferentschik:
Ok, thanks

hferentschik:
Fixed :-)

Issue #249, opened 2017-01-04
Add cross-references to generic OpenShift content


thatdocslady:
Considering our approach of keeping our documentation minishift-specific, we should go through the openshift docs and cherry-pick content to link from our docs, in lieu of duplicating content that already exists.

hferentschik:
Merged via pull request  #404

PR #404, opened 2017-02-14
Issue #249 Adding links to generic openshift content in README

thatdocslady:
Fixes issue #249 

Added a few relevant links for further information on the OpenShift docs site, mainly for architecture and getting started with cluster administration.

Also reviewed using.md, and it seems that there are already some links to OpenShift docs so no need to add there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,6 +120,10 @@ You can now install the latest beta version of minishift.
 1. Last but not least, use your hypervisor management tool to confirm that there are no remaining
    artifacts related to the Minishift VM (eg use `virsh` in the KVM case).
 
+For more information about OpenShift cluster architecture,
+see [Architecture Overview](https://docs.openshift.org/latest/architecture/index.html) in the
+OpenShift documentation.
+
 <a name="quickstart"></a>
 ## Quickstart
 
@@ -163,6 +167,9 @@ with the KVM hypervisor driver.
   to use a different command to add `oc` to the _PATH_ environment variable.
   To verify the `oc` version, check the contents of the `~/.minishift/cache/oc` directory.
 
+For more information about interacting with OpenShift with the command-line interface and
+the Web console, see [Interacting with OpenShift](/docs/using.md#interacting-with-openshift)
+
 <a name="deploying-a-sample-application"></a>
 ### Deploying a sample application
 
@@ -193,6 +200,10 @@ To deploy the Node.js sample application from the command-line:
         $ minishift stop
         Stopping local OpenShift cluster...
         Stopping "minishift"...
+
+For more information about creating applications in OpenShift,
+see [Creating New Applications](https://docs.openshift.org/latest/dev_guide/application_lifecycle/new_app.html) in
+the OpenShift documentation.
 
 <a name="reusing-the-docker-daemon"></a>
 ### Reusing the Docker daemon


hferentschik:
Add to whitelist

README.md
```diff
@@ -5,6 +5,9 @@ cluster inside a VM. You can try out OpenShift or develop with it, day-to-day, o
 
 Minishift uses [libmachine](https://github.com/docker/machine/tree/master/libmachine) for
 provisioning VMs, and [OpenShift Origin](https://github.com/openshift/origin) for running the cluster.
+For general information about OpenShift cluster architecture,
```
hferentschik: Hmm, I don't think the leading paragraphs are a good place to add such a link.

hferentschik: Also why "general"? Is "more" not better? What does "general" mean  in this context anyways?

README.md
```diff
@@ -143,6 +146,10 @@ with the KVM hypervisor driver.
   to use a different command to add `oc` to the _PATH_ environment variable.
   To verify the `oc` version, check the contents of the `~/.minishift/cache/oc` directory.
 
+For general information about using the `oc` command and the command-line console,
+see [Get Started with CLI](https://docs.openshift.org/latest/cli_reference/get_started_cli.html) in
```
hferentschik: CLI == Command Line Interface - this does not mean anything yet. You need to specify which CLI. "Get Started with the oc CLI"

README.md
```diff
@@ -174,6 +181,10 @@ To deploy the Node.js sample application from the command-line:
         Stopping local OpenShift cluster...
         Stopping "minishift"...
 
+For general information about creating applications in OpenShift,
```
hferentschik: I would say for "more" information.

budhrg: +1 for `more`

README.md
Preeticp: For general information about using the oc command and the command-line console...> This should probably be - ... about using the oc command-line and the web console. 

...see Get Started with CLI in the OpenShift documentation > The link provided here mostly talks about downloading OC CLI in various Operating systems, which afaiu is exactly what Minishift does for you in the 2 steps of the Quickstart. Also, the link solely talks about OC CLI, there is no information about web console here, so we could either drop reference to the console totally or add another link for information about web console. 

My 2 cents: A more relevant link in this section would probably be 'https://docs.openshift.org/latest/cli_reference/basic_cli_operations.html#cli-reference-basic-cli-operations', which would help Developers explore OC further once Minishift launches an OC instance for them. I think that would be the logical next-step for the Dev-user.

README.md
Preeticp: I don't think the leading paragraphs are a good place to add such a link.
+1, maybe we could add this link towards the end of the Quickstart section.

hferentschik: >  maybe we could add this link towards the end of the Quickstart section.

sure.

README.md
hferentschik: Actually we have the link already now - https://github.com/minishift/minishift/blob/master/docs/using.md#openshift-client-binary-oc (to the external oc doc). So I think here we should link to our docs in using.md

hferentschik: > Also, the link solely talks about OC CLI, there is no information about web console here, so we could either drop reference to the console totally or add another link for information about web console.

Don't we have this as well - https://github.com/minishift/minishift/blob/master/docs/using.md#console

thatdocslady:
@hferentschik @Preeticp updated the PR with your feedback, moved some of the links around and changed the oc link to point to our usage docs. Hopefully this resolves the concerns :)

hferentschik:
Rebased and merged, thanks.

Issue #374, opened 2017-02-09
Add info about proxy configuration to docs


hferentschik:
Following up on issue #287

hferentschik:
We probably want something which explains how to enable it, what it actually means (in terms of Docker, OpenShift, shell, etc) and what the current limitations are (aka need Origin 1.5)

hferentschik:
Merged via pull request #408

PR #408, opened 2017-02-15
Issue #374 Add info about proxy configurations to docs

praveenkumar:


README.md
```diff
@@ -20,6 +20,7 @@ provisioning VMs, and [OpenShift Origin](https://github.com/openshift/origin) fo
   - [Deploying a sample application](#deploying-a-sample-application)
   - [Reusing the Docker daemon](#reusing-the-docker-daemon)
 - [Documentation](#documentation)
+- [Using Minishift with an HTTP/HTTPS Proxy](#using-minishift-with-an-proxy)
```
budhrg: `with a`

README.md
```diff
@@ -207,6 +208,26 @@ The following documentation is available:
 * [Release Notes](https://github.com/minishift/minishift/releases)
 * [Developing Minishift](./docs/developing.md)
 
+<a name="using-minishift-with-an-proxy"></a>
```
budhrg: Same here `#using-minishift-with-a-proxy)`

README.md
```diff
@@ -207,6 +208,26 @@ The following documentation is available:
...
+## Using Minishift with an HTTP/HTTPS Proxy
```
budhrg: same here `with a`.

README.md
```diff
@@ -207,6 +208,26 @@ The following documentation is available:
...
+To do this, pass the required flags during `minishift start`.
+
+For example:
+
+```shell
+$ minishift start --http-proxy http://YOURPROXY:PORT \
+                 --https-proxy https://YOURPROXY:PORT
+```
+
+ In an authenticated proxy environment, the proxy_user and proxy_password should be part of proxy URI.
```
budhrg: Good to enclose proxy_user and proxy_password in to back ticks.

Preeticp: URI>URL?

README.md
Preeticp: It might be better to move this section below Starting Minishift section, wdyt? The user will need to use these flags with Minishift start.

README.md
praveenkumar: Uniform Resource Identifier => URI

README.md
praveenkumar: I think better to have it a separate section because this is not everyone is going to use so putting it in this section will help only needy users.

README.md
```diff
@@ -207,6 +208,26 @@ The following documentation is available:
 * [Release Notes](https://github.com/minishift/minishift/releases)
 * [Developing Minishift](./docs/developing.md)
 
+<a name="using-minishift-with-a-proxy"></a>
```
hferentschik: I think this really should be in _using.md_ and not in the README. First we don't want to get the README to long and secondly we want this to be part of the generated docs later to be hosted on docs.openshift.org. As part of that we most likely won't include the README. That is specific to the GitHub repo.

README.md
```diff
@@ -207,6 +208,26 @@ The following documentation is available:
...
+                 --https-proxy https://YOURPROXY:PORT
+```
+
+ In an authenticated proxy environment, the `proxy_user` and `proxy_password` should be part of proxy URI.
+ 
+```shell
+ $ minishift start --http-proxy http://<proxy_username>:<proxy_password>@YOURPROXY:PORT \
+                  --https-proxy https://<proxy_username>:<proxy_password>YOURPROXY:PORT
+```
+
```
hferentschik: Looking good so far, but I was hoping for some additional context on what this does for the user. Aka, the various parts which gets configured for proxies.

Might be worth to have a look at https://github.com/openshift/origin/blob/master/docs/cluster_up_down.md#using-a-proxy and maybe even reference it.

README.md
LalatenduMohanty: +1 @hferentschik's comments. This PR should be part of generated/hosted docs.

budhrg: +1 too

README.md
hferentschik: Also, we probably want to add a note about the mimium version requirement for using proxies.

hferentschik:
Merged. I moved the proxy docs into the using document though.

PR #414, opened 2017-02-15
Issue #413 Update irc name in README

### LINKED ISSUE #413: Update irc name in README
Since we now have `minishift` irc, its time to update in README too.

budhrg:
Fix #413


hferentschik:
Good one :-) I thought once about it, but then forgot.

hferentschik:
Merged, thanks.

PR #416, opened 2017-02-16
Issue #415 Exporting pull request artifacts to a storage box

### LINKED ISSUE #415: Exporting pull request artifacts to a storage box
See https://github.com/minishift/minishift-ci-jobs/issues/1#issuecomment-280229034

budhrg:
Fix #415 

If the job failed in between then artifacts will not be stored, eg - https://ci.centos.org/job/minishift/152/console

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -24,7 +24,7 @@ set -e
 # that might interest this worker.
 if [ -e "jenkins-env" ]; then
   cat jenkins-env \
-    | grep -E "(JENKINS_URL|GIT_BRANCH|GIT_COMMIT|BUILD_NUMBER|ghprbSourceBranch|ghprbActualCommit|BUILD_URL|ghprbPullId|GH_TOKEN)=" \
+    | grep -E "(JENKINS_URL|GIT_BRANCH|GIT_COMMIT|BUILD_NUMBER|ghprbSourceBranch|ghprbActualCommit|BUILD_URL|ghprbPullId|GH_TOKEN|JOB_NAME|CICO_API_KEY)=" \
     | sed 's/^/export /g' \
     > ~/.jenkins-env
   source ~/.jenkins-env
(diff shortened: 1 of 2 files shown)


budhrg:
> Is this ready to go?

Yes, but need some documentation changes as you suggested.

> Where does it upload to, we need some documentation with this in the README on where to find the artifacts

Uploads to http://artifacts.ci.centos.org/minishift/minishift/{Job Build Number} . Documentation on its way. I am thinking to add a new section related to CI in [developing doc](https://github.com/minishift/minishift/blob/master/docs/developing.md) in between _Building minishift_ and _Releaseing Minishift_ sections. Let me know you have any other better place.

>  Does this also upload artifacts for pull request builds? That would seem over the top imo. I think only master builds should be uploaded.

I only upload the artifacts of pull request builds which pass successfully in CI. 
Eg : If you go here http://artifacts.ci.centos.org/minishift/minishift/, you will find artifacts of two builds `151 and `153` which got passed in CI and `152` build failed.

hferentschik:
> Documentation on its way. I am thinking to add a new section related to CI in developing doc in between Building minishift and Releaseing Minishift sections.

+1 Not only should there be a link to the upload directory, but also a link to the actual build job on Jenkins. After all one needs to match the build numbers to artifacts.

> Let me know you have any other better place.

Sounds good to me.

>  If you go here http://artifacts.ci.centos.org/minishift/minishift/, you will find artifacts of two builds 151 and 153which got passed in CI and152` build failed.

Sure, I get that. Failing builds are not uploading anything, but what's about pull request builds vs master builds?  But looking at the builds you reference, it seems to me that I can answer the question myself. These are pull request builds and not master builds. They should not be published.

That said, I think something is fishy here anyways. Where are the master builds on https://ci.centos.org/job/minishift/? I can only see pull request builds. That seems wrong. Also looking at the README, the CentOS build is marked as failing even though master is in a "good" state. I think CentOS CI is marked as failing, since the last pull request build (154) failed. This is wrong. The status should be against the latest master build. Look for example at https://ci.centos.org/job/vagrant-service-manager/ - see how there are master and pull request. One can differentiate between them in the build scripts. I think somewhere something is not setup properly.

budhrg:
@hferentschik 

CI now shows the link to artifacts as shown here  https://ci.centos.org/job/minishift/181/console .

Confirmed that artifacts are storing in http://artifacts.ci.centos.org/minishift/minishift/pr/416/ as per the format.

hferentschik:
Merged, thanks.

PR #433, opened 2017-02-20
Fix #135 Add a basic test scenario using godog

### LINKED ISSUE #135: Introduce acceptance test framework
My choice would be [godog](https://github.com/DATA-DOG/godog) which is a Cucumber port for Golang.

Another option could be [Ginko](http://onsi.github.io/ginkgo/), but imo it misses the clarity of Cucumber.

gbraad:
Fix #135


test/integration/features/basic.feature
```diff
@@ -0,0 +1,21 @@
+# 
+Feature: Bringup
+  In order to use MiniShift
+  As a user
+  I need to be able to bring up a test environment
+
+
```
budhrg: Extra line IMO

budhrg:
@gbraad I think its time to update this section too https://github.com/minishift/minishift/blob/master/docs/developing.md#integration-tests

budhrg:
Tests failing due to #434

test/integration/features/basic.feature
```diff
@@ -0,0 +1,21 @@
...
+  Scenario: Docker environment variables
+    Given a "developer"
+     When executing "minishift start --vm-driver=vbox --docker-env=FOO=BAR --docker-env=BAZ=BAT"
```
gbraad: currently failing due to #434, but perhaps better to use kvm, as centos-ci relies on this.

We can improve later to allowing several scenarios

test/integration/basic_test.go
```diff
@@ -0,0 +1,112 @@
...
+
+	s.BeforeScenario(func(interface{}) {
+		testDir := setUp()
+		os.RemoveAll(testDir)
+	})
+
+	s.AfterScenario(func(interface{}, error) {
+		m.runner.EnsureDeleted()
+	})
+
```
budhrg: Extra line here

test/integration/features/basic.feature
```diff
@@ -0,0 +1,21 @@
+# 
+Feature: Bringup
+  In order to use MiniShift
+  As a user
+  I need to be able to bring up a test environment
+
+
+  Scenario: Docker environment variables
+    Given a "developer"
+     When executing "minishift start --vm-driver=kvm --docker-env=FOO=BAR --docker-env=BAZ=BAT"
```
budhrg: `--vm-driver=kvm` No need to mention it as kvm is default driver

budhrg:
LGTM :+1:

test/integration/basic_test.go
```diff
@@ -0,0 +1,112 @@
...
+func (m *Minishift) shouldBeRunning() error {
+	if !m.runner.CheckStatus("Running") {
+		return fmt.Errorf("Expected minishift to be running")
+	}
+
+	return nil
+}
+
+func (m *Minishift) shouldHaveAValidIPAddress() error {
```
hferentschik: 👍 right, these methods are what I meant with building up a Minishift specific set of helper functions (step definitions) which can be reused by the various test descriptions (feature files)

test/integration/basic_test.go
```diff
@@ -0,0 +1,112 @@
...
+	cmdOutput := m.runner.RunCommand(command, true)
+	if !strings.Contains(cmdOutput, expectedStr) {
+		return fmt.Errorf("Output did not match. Expected: %s, Actual: %s", expectedStr, cmdOutput)
+	}
+
+	return nil
+}
+
+func FeatureContext(s *godog.Suite) {
+	runner := util.MinishiftRunner{
```
hferentschik: so this FeatureContext is global for all tests we are going to write, right?

hferentschik: Moving forward we can also split the methods in this file into multiple files grouping them in some way. I am just thinking that we start getting a lot of step definitions it makes sense to split them up. But I guess at least as long as we stay in the same package, this should not be a problem, right?

test/integration/basic_test.go
```diff
@@ -0,0 +1,112 @@
...
+		Args:       *args,
+		BinaryPath: *binaryPath}
+
+	m := &Minishift{runner: runner}
+
+	s.Step(`^a "([a-zA-Z\s]+)"$`, m.aPersona)
+	s.Step(`^should be running$`, m.shouldBeRunning)
```
hferentschik: so why are you so stringent with the regexp and force the match of the beginning and end of line? It makes the actual feature file hard to read imo. Why not '(should be running)'. That said, I think it should be 'Minishift should be running' ,since this explicitly maps to _minishift status_. Moving forward, I could imagine having something like 'process (.*) is running' which would verify whether a given process (say dockerd) is running.

test/integration/basic_test.go
```diff
@@ -0,0 +1,112 @@
...
+}
+
+func FeatureContext(s *godog.Suite) {
+	runner := util.MinishiftRunner{
+		Args:       *args,
+		BinaryPath: *binaryPath}
+
+	m := &Minishift{runner: runner}
+
+	s.Step(`^a "([a-zA-Z\s]+)"$`, m.aPersona)
```
hferentschik: How is that supposed to be used?

test/integration/features/basic.feature
```diff
@@ -0,0 +1,20 @@
...
+  Scenario: Docker environment variables
+    Given a "developer"
+     When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
+      And should be running
+     Then should have a valid IP address
+     Then executing "minishift ssh echo hello" should contain
+      """
+      hello
+      """
+     Then executing "minishift ssh cat /var/lib/boot2docker/profile" should contain
```
hferentschik: This really mixes condition and testable outcome. I'd like so see more something like this:

    Feature: Bringup
      In order to use MiniShift
      As a user
      I need to be able to bring up a test environment
    
      Scenario: Docker environment variables
        Given a "developer"
         When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
         Then Minishift should be running
         And  it should have a valid IP address
         
         When executing "minishift ssh echo hello"
         Then stdout from "minishift ssh echo hello" should contain
          """
          hello
          """
    
         When executing "minishift ssh cat /var/lib/boot2docker/profile" 
         Then stdout from "minishift ssh cat /var/lib/boot2docker/profile" should contain
          """
          export "FOO=BAR"
          export "BAZ=BAT"
          """    

In this case, the executing step would execute the command and store exit status, stdout and stderr output in a map or so. Then I can have as many assertions as I want after the execution. I can not only assert that something specific comes out stdout, but also that stderr was empty and exist status was '0', eg:

         When executing "minishift ssh cat /var/lib/boot2docker/profile" 
         Then stdout from "minishift ssh cat /var/lib/boot2docker/profile" should contain
          """
          export "FOO=BAR"
          export "BAZ=BAT"
          """    
          And stderr should be empty
          And exit status should be 0

In vagrant-service-manager this abstraction was provided by Aruba, but the basic capturing of output and exit code, should be easy enough to implement.

test/integration/flags.go
```diff
@@ -23,14 +23,8 @@ import (
 	"fmt"
 	"os"
```
hferentschik: +1, let's get rid of all this

test/integration/basic_test.go
gbraad: At the moment we are not using this, but a persona could be a defined user profile. Eg.

`John Appletree` uses a Macbook with 8G of memory. Therefore he will deploy a minishift instance with only 2G of memory and two vCPU.

'Bill Stevens' uses a Windows computer and always starts using a PowerShell instance, using Hyper-V

'Steve Linus'  uses Windows, but always resorts to open tools, and uses Cygwin as a terminal and  VirtualBox.

etc.

test/integration/basic_test.go
gbraad: consistency... but the status is a possible improvement. 

the `process (.*)` seems in the described scenario less logical, as minishift itself is not the process running, but the hypervisor monitor process. Also, dockerd lives inside the VM?

test/integration/features/basic.feature
gbraad: Personal preference as I do not prefer to keep state of a previous action, unless it is stored inside the application/context and there is a verifiable action from the applications interface.

And yes, it currently implements/expects the happy-flow.

test/integration/basic_test.go
hferentschik: I see, and given the persona some certain startup values are set or certain hardware prerequisites are verified? I guess I can see where this could be useful, not sure though whether we have any need for this in the immideate future.  Personally, I'd leave it out for now, but if you feel strongly about it.

test/integration/basic_test.go
hferentschik: maybe, still by being less stringent, I can write the Gherkin file more fluent. Right now there are quite some substantives missing - "And should be running" Who? One would like to say "it" or "Minishift". IMO it is common practice in Gherkin to match on relevant substrings, instead of matching one whole line.

test/integration/features/basic.feature
hferentschik: > Is this a change we need immediately?

I think so. The plan is to add integration tests now as part of code changes where possible. It makes sense to get the basic use case right. In many case we will have something of the form:

    When executing "minishift <foo>" 
    Then ...
    And ...

And I think in this cases where I execute a minishift command, we will also sometimes not just assert/verify stdout, stderr and exit status. We might have other checks as well, e.g. verifying that config.json has changed.

test/integration/basic_test.go
budhrg: Same here, let's not add this persona at the moment. We can add in future.

gbraad: Check. At this stage it is non-essential. But it is something we can keep on mind... Some of the features/issues we have are based around the different environment. For instance, starting from Cygwin is a no-go. If this is something that gets supported, it means we also need to create a test around this. A persona definition can help with this... and immediately conveys the idea when people also use this in issue reports or feature definitions. For now, no need...

test/integration/features/basic.feature
gbraad: Not sure what you mean with the verification of the content of a file, if it has been changed. This  has to be assigned a step in FeatureContext anyways, as there is not distinguishing between 'When', 'Then' or 'And' and no relation can be given to them.

The same goes for the described example. This means that the 'Then' step needs to register an index, based on the command given, because a successive step would need this to know which 'stderr' or 'stdout' is requested. I have no problem with this... but I can not enforce the usage of 'Then' and 'And', they have the same meaning in godog.

A way around this could be:

```
     When executing "minishift ssh cat /var/lib/boot2docker/profile" 
     Then stdout should contain
      """
      export "FOO=BAR"
      export "BAZ=BAT"
      """    
      And stderr should be empty
     When executing "minishift hostfolder mount Users"
     Then stdout should contain:
      """
      ... OK
      """
```
Which means that the index is set by the last executed command (or only the last command is kept on record). This way, no association needs to be kept. WDYT?

It would not be very beneficial to end up in the situation of doing;
```
     Then stdout from "minishift ssh cat /var/lib/boot2docker/profile" should contain
      """
      export "FOO=BAR"
      export "BAZ=BAT"
      """    
      And stderr from "minishift ssh cat /var/lib/boot2docker/profile" should be empty
      And exit status from "minishift ssh cat /var/lib/boot2docker/profile" should be 0
```
as this makes the Gherkin VERY stringent and inflexible.

test/integration/flags.go
gbraad: This was the idea of task #440

test/integration/features/basic.feature
hferentschik: > Not sure what you mean with the verification of the content of a file, if it has been changed. 

This was just an example of an additional test condition which one could have after executing a command. For example:

     When executing "minishift config set memory 2048" 
     Then stdout of "minishift config set memory 2048" should match
       """
         Memory set to 2048
       """
     And exit status of "minishift config set memory 2048" is 0
     And file '$MINISHIFT_HOME/config/config.json' should contain
     """
    \"memory\": 4096
    """

I am most interested in how this reads and how it communicates the actual test in terms of preconditions and following assertions. I get that the 'When', 'Then' part is not enforced in the step definitions.

> as there is not distinguishing between 'When', 'Then' or 'And' and no relation can be given to them.

Sure

> The same goes for the described example. This means that the 'Then' step needs to register an index, based on the command given, because a successive step would need this to know which 'stderr' or 'stdout' is requested. 

Right, so in the example the command itself is the key/index

>I have no problem with this... but I can not enforce the usage of 'Then' and 'And', they have the same meaning in good.

Sure. I don't think you can do that in Cucumber in general.

> Which means that the index is set by the last executed command (or only the last command is kept on record). This way, no association needs to be kept. WDYT?

Could work. It would break however, if your precondition consists of two commands (example might be contrived):

     When executing "minishift config set memory 2048" 
     And executing "minishift config unset memory" 
     And file '$MINISHIFT_HOME/config/config.json' should be empty

This case might not be so common though. 

> as this makes the Gherkin VERY stringent and inflexible.

Why? Because you repeat the command? It is basically just the index you have been talking about. If one wanted to work around that one could do something like:

     When executing "minishift config set memory 2048" as "set-command"
     Then exit status of "set-command" is 0

Effectively allowing to specify your own index. Not sure whether this is much better. 

I could go either way. Implicit index in a way that just the last command is kept, command itself is key/index or even the ability to provide key/index. WDYT?

gbraad: I would prefer going for the simpler approach where there is no index for the execution (and only the last command and status is kept). If we really need a scenario that needs to run two commands, we could improve on this...

The reasoning for me is that the Gherkin should be as minimal as possible, unless we need to perform those actions. It is easier to add something to the definition, than removing it later when we do not use it.

Feedback from QA/QE would be appreciated... else, we will have to make the decision.

hferentschik: > I would prefer going for the simpler approach where there is no index for the execution (and only the last command and status is kept). If we really need a scenario that needs to run two commands, we could improve on this...

Fair enough

gbraad:
Added `lastCommandOutput` to keep the output of the last executed command, for stdout, stderr and the exit code.

gbraad:
also changed the "minishift should be running" in an expression that checks state in a more re-usable way.

budhrg:
Also fix #440 440

centos_ci.sh
```diff
@@ -79,7 +79,7 @@ export PATH=$PATH:/tmp/glide/${GLIDE_OS_ARCH}
 # Test
 make clean test cross fmtcheck prerelease
 # Run integration test with 'kvm' driver
-VM_DRIVER=kvm make integration
+MINISHIFT_VM_DRIVER=kvm make integration
```
hferentschik: :+1:

docs/developing.md
```diff
@@ -192,7 +192,11 @@ Run `go test --help` to read more about test options.
 <a name="integration-tests"></a>
 #### Integration Tests
 
-\<WIP\> - see also GitHub issue [#135](https://github.com/minishift/minishift/issues/135)
+Integration are using [`godog`](https://github.com/DATA-DOG/godog), which uses Gherkin (Cucumber) to define testcases.
```
hferentschik: 👍

glide.yaml
```diff
@@ -56,6 +56,7 @@ import:
 - package: gopkg.in/cheggaaa/pb.v1
 - package: github.com/asaskevich/govalidator/
   version: 5
+- package: github.com/DATA-DOG/godog
```
hferentschik: Don't we have a version? At least a version range?

test/integration/basic_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	"github.com/DATA-DOG/godog"
+	"github.com/DATA-DOG/godog/gherkin"
+
+	"github.com/minishift/minishift/test/integration/util"
+)
+
+var lastCommandOutput CommandOutput
+
+type CommandOutput struct {
```

test/integration/features/basic.feature
```diff
@@ -0,0 +1,28 @@
+# 
+Feature: Bringup
+  In order to use MiniShift
+  As a user
+  I need to be able to bring up a test environment
+
+  Scenario: Basic test scenario
+    Given executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
+     Then minishift should be "Running"
+     Then minishift should have a valid IP address
```
hferentschik: _And_

test/integration/features/basic.feature
```diff
@@ -0,0 +1,28 @@
...
+      And stdout should contain
+      """
+      export "FOO=BAR"
+      export "BAZ=BAT"
+      """
+     When executing "minishift stop"
+     Then minishift should be "Stopped"
+     When executing "minishift delete"
+     Then minishift should be "Does Not Exist"
+
```
hferentschik: Missing new line (to keep Git happy)

test/integration/util/util.go
```diff
@@ -24,33 +24,49 @@ import (
...
-func (m *MinishiftRunner) RunCommand(command string, checkError bool) string {
+func (m *MinishiftRunner) RunCommand(command string) (stdOut string, stdErr string, exitCode int) {
 	commandArr := strings.Split(command, " ")
-	path, _ := filepath.Abs(m.BinaryPath)
+	path, _ := filepath.Abs(m.CommandPath)
 	cmd := exec.Command(path, commandArr...)
-	stdout, err := cmd.Output()
 
-	if checkError && err != nil {
+	var outbuf, errbuf bytes.Buffer
```
hferentschik: 👍

glide.yaml
```diff
@@ -56,6 +56,8 @@ import:
...
+  version: c3f5ce27930fff8efffaaf0acab7f24d5deecdbc
```
hferentschik: :-)

test/integration/basic_test.go
```diff
@@ -0,0 +1,167 @@
...
+	var givenArgs = flag.String("minishift-args", "", "Arguments to pass to minishift")
+	var givenPath = flag.String("binary", fmt.Sprintf("../../out/%s-amd64/minishift", runtime.GOOS), "Path to minishift binary")
+
+	runner := util.MinishiftRunner{
+		CommandArgs: *givenArgs,
+		CommandPath: *givenPath}
+
+	m := &Minishift{runner: runner}
+
+	s.Step(`minishift should be "([^"]*)"`, m.shouldHaveState)
```
hferentschik: Maybe "minishift state should be ..."

hferentschik: Technically I think it should be "Minishift" (capital 'M'). Same for the IP address case. I think we said we want to use _minishift_ for the cases where we actually refer to the binary itself. Opposed to _Minishift_ as the tool/program/cli.

hferentschik:
@gbraad where is the test report? Can we not spit out the HTML test report and put it into the build/out directory?

test/integration/setup.go
```diff
@@ -22,14 +22,10 @@ import (
...
-	"testing"
 )
 
-func setUp(t *testing.T) string {
-	testDir, err := ioutil.TempDir("", "minishift-integration-test-")
-	if err != nil {
-		t.Fatalf("Failed to create temp directory '%s'", testDir)
-	}
+func setUp() string {
+	testDir, _ := ioutil.TempDir("", "minishift-integration-test-")
```
hferentschik: FYI, in the long run I want to get rid off ioutil.TempDir. Instead I want to create the test directories under the build folder in the source checkout. Using the tmp dir approach makes it really hard to find test files when one needs to debug something. Also it is sometimes better to keep the testdir until an explicit clean, instead of getting it deleted after the test run. I guess we can address this in another issue though.

glide.yaml
praveenkumar: @gbraad I can see godog actually have release tag https://github.com/DATA-DOG/godog/releases can we use that instead of hash?

gbraad: @hferentschik @praveenkumar Did not see the releases tags on my clone. Will change this...

test/integration/setup.go
gbraad: agree

Makefile
```diff
@@ -88,7 +88,7 @@ test: vendor $(GOPATH)/src/$(ORG)
 
 .PHONY: integration
 integration: $(BUILD_DIR)/$(GOOS)-$(GOARCH)/minishift$(IS_EXE)
-	go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration
+	go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration -v
```
praveenkumar: Verbose 👍

glide.yaml
```diff
@@ -56,6 +56,8 @@ import:
 - package: gopkg.in/cheggaaa/pb.v1
 - package: github.com/asaskevich/govalidator/
   version: 5
+- package: github.com/DATA-DOG/godog
+  version: v0.6.2
```
praveenkumar: version as tag 👍

test/integration/util/util.go
```diff
@@ -91,27 +107,21 @@ func (m *MinishiftRunner) SetEnvFromEnvCmdOutput(dockerEnvVars string) error {
...
 }
 
-func NewOcRunner(t *testing.T) *OcRunner {
-	p, err := exec.LookPath("kubectl")
-	if err != nil {
-		t.Fatal("Couldn't find kubectl on path.")
-	}
-	return &OcRunner{BinaryPath: p, T: t}
+func NewOcRunner() *OcRunner {
+	p, _ := exec.LookPath("kubectl")
```
praveenkumar: Didn't get it why we are looking for `kubectl` path to get `oc` binary?

gbraad: This is is from a TODO @hferentschik added. I believe we need similar code to call `oc` later, so I left this unchanged for now.

gbraad: It is unused code for now... I changed it anyways as I saw another small change I could make.

test/integration/basic_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+func (m *Minishift) commandReturnShouldContain(commandField string, expected string) error {
+	return compareExpectedWithActualContains(expected, selectFieldFromLastOutput(commandField))
+}
+
+func (m *Minishift) commandReturnShouldContainContent(commandField string, expected *gherkin.DocString) error {
+	return compareExpectedWithActualContains(expected.Content, selectFieldFromLastOutput(commandField))
+}
+
+func (m *Minishift) commandReturnShouldEqual(commandField string, expected string) error {
```
gbraad: @hferentschik: as part of the statement...

test/integration/basic_test.go
```diff
@@ -0,0 +1,178 @@
...
+	return compareExpectedWithActualEquals(expected, selectFieldFromLastOutput(commandField))
+}
+
+func (m *Minishift) commandReturnShouldEqualContent(commandField string, expected *gherkin.DocString) error {
```
gbraad: @hferentschik: ... and as part of content.

glide.yaml
hferentschik: :+1:

test/integration/basic_test.go
hferentschik: cool

hferentschik:
Merged, thanks. A milestone :-)

budhrg:
Indeed :smile:  Thanks @gbraad

gbraad:
So we can finally start using this!

hferentschik:
> So we can finally start using this!

yeah :-) "But it would require an integration test" is no excuse anymore ;-)

PR #439, opened 2017-02-20
Issue #435 Enable CI when a push to master occurs

### LINKED ISSUE #435: Enable CI to get triggered when a push to master occurs
Currently, CI only get triggered when PR is sent.

We need to trigger it when PR get merged with master.

It should also store the generated artifacts as well.

See https://github.com/minishift/minishift/issues/415#issuecomment-280584915

budhrg:
Fix #435 . 

Previously not using github jenkins plugin. Enabled with this PR https://github.com/minishift/minishift-ci-jobs/pull/17

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -83,10 +83,19 @@ VM_DRIVER=kvm make integration
 
 # On reaching successfully at this point, upload artifacts
 PASS=$(echo $CICO_API_KEY | cut -d'-' -f1-2)
-mkdir -p $JOB_NAME/pr/$ghprbPullId
-cp -r out/* $JOB_NAME/pr/$ghprbPullId/
-# http://stackoverflow.com/a/22908437/1120530 using --relative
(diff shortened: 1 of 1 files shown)


hferentschik:
Merged, thanks.

hferentschik:
I merged this pull request since it looked ok and I was not sure how else I would be able to see whether it works, but it seems something is not working. Several pushes to master have occurred and I can neither see master build here https://ci.centos.org/job/minishift/, nor can I see any new artefacts in the archive server - http://artifacts.ci.centos.org/minishift/minishift/

budhrg:
> I merged this pull request since it looked ok and I was not sure how else I would be able to see whether it works, but it seems something is not working. Several pushes to master have occurred and I can neither see master build here https://ci.centos.org/job/minishift/, nor can I see any new artefacts in the archive server - http://artifacts.ci.centos.org/minishift/minishift/

@hferentschik , I noticed that I did a silly mistake of using different operator `-eq` in if condition during artifacts upload here https://github.com/minishift/minishift/blob/master/centos_ci.sh#L91 . 

Master Builds are running here https://ci.centos.org/job/minishift-build-master/.
Sent a new PR https://github.com/minishift/minishift/pull/442

Issue #434, opened 2017-02-20
Specifying virtualbox as --vm-driver does not work


gbraad:
Starting minishift with: `minishift start --vm-driver=virtualbox` will currently fail with compiling from master branch.

```
Starting local OpenShift cluster using 'virtualbox' hypervisor...
E0220 14:11:01.893091   26749 start.go:149] Error starting the VM: Error creating new host: Driver "virtualbox" not found. Do you have the plugin binary "docker-machine" accessible in your PATH?. Retrying.
E0220 14:11:01.894315   26749 start.go:149] Error starting the VM: Error creating new host: Driver "virtualbox" not found. Do you have the plugin binary "docker-machine" accessible in your PATH?. Retrying.
E0220 14:11:01.894404   26749 start.go:149] Error starting the VM: Error creating new host: Driver "virtualbox" not found. Do you have the plugin binary "docker-machine" accessible in your PATH?. Retrying.
E0220 14:11:01.894418   26749 start.go:155] Error starting the VM:  Error creating new host: Driver "virtualbox" not found. Do you have the plugin binary "docker-machine" accessible in your PATH?
Error creating new host: Driver "virtualbox" not found. Do you have the plugin binary "docker-machine" accessible in your PATH?
Error creating new host: Driver "virtualbox" not found. Do you have the plugin binary "docker-machine" accessible in your PATH?
```

Reproducible on Linux, Windows. Not on masOS.

budhrg:
This is little weird how minishift is behaving here with `--vm-driver=vbox` as supported drivers for windows are hyperv and  virtualbox as per this [src](https://github.com/minishift/minishift/blob/master/pkg/minikube/constants/constants_windows.go#L21).

I am getting error in windows for `minishift start --vm-driver=vbox` as:
```
$ minishift.exe start --vm-driver vbox
```

budhrg:
Linux behavior on compiling minishit with latest master HEAD ()
```
➜  minishift git:(master) ./out/linux-amd64/minishift start --vm-driver virtualbox
Starting local OpenShift cluster using 'virtualbox' hypervisor...
E0220 12:10:30.583162   16490 start.go:149] Error starting the VM: Error creating new host: Driver "virtualbox" not found. Do you have the plugin binary "docker-machine" accessible in your PATH?. Retrying.
[...]

PR #441, opened 2017-02-20
Issue #434 Resurrecting call to machine.StartDriver()

hferentschik:
Fix for issue #434. 

This should make the driver selection work again.

### DIFF
--- a/cmd/minishift/main.go
+++ b/cmd/minishift/main.go
@@ -18,8 +18,10 @@ package main
 
 import (
 	"github.com/minishift/minishift/cmd/minishift/cmd"
+	"github.com/minishift/minishift/pkg/minikube/machine"
 )
 
 func main() {
+	machine.StartDriver()
 	cmd.Execute()
 }


budhrg:
Added `gofmt` fix here https://github.com/minishift/minishift/pull/444.

IMO, this should be merged after #444 and #442 .

gbraad:
Code is not formatted correctly. The fix itself is important.

LalatenduMohanty:
@gbraad @budhrg I am confused. When I do `make fmt` I do not see any issues. What am I missing?

cmd/minishift/main.go
```diff
@@ -18,8 +18,10 @@ package main
 
 import (
 	"github.com/minishift/minishift/cmd/minishift/cmd"
+	"github.com/minishift/minishift/pkg/minikube/machine"
 )
 
 func main() {
+	machine.StartDriver()
```
LalatenduMohanty: I can not believe both of us missed it during review 🥇 :)

LalatenduMohanty:
@budhrg @gbraad , Got it it is `make fmtcheck`.

gbraad:
@LalatenduMohanty 
noticed the same yesterday. Not sure why we have two almost similar commands... `make fmt` print the names of the files modified, but it didn't seem to do anything. We also do not use this command inside the `Makefile`.

cmd/minishift/main.go
gbraad: well, the review itself was messy as it contained incomplete merge actions. However, the changeset that remained was quite large. Just wonder how this made things work on macOS.

budhrg:
@gbraad `make fmt` actually fixes the indentation issues in source fine as per Go guidelines. https://golang.org/cmd/gofmt/.

and `make fmtcheck` checks if any source file violates above guidelines.

So, it's recommanded in Golang world to always do `make fmt` or `gofmt -w <source dir>`.

cmd/minishift/main.go
hferentschik: >  review itself was messy as it contained incomplete merge actions

not sure what you mean?

> just wonder how this made things still work on macOS.

just to be clear, this method does by no means "start" anything as the method name makes you think. The interesting bit is:

    localbinary.CurrentBinaryIsDockerMachine = true


The if statement is pretty much useless and only comes into play in the test of this method. I was experimenting with this to better understand it and removed it w/i bringing it back. Never noticed, since at least for OS X, it does not make a difference. I have a theory mourned this, but it is not well formed yet.

PR #444, opened 2017-02-21
Issue #443 gofmtcheck fails in master

budhrg:
Fix #443 

* Also refactored the code

PR #446, opened 2017-02-21
Issue #445 Enable fmtcheck and prerelease checks in Circle CI

### LINKED ISSUE #445: Enable fmtcheck and prerelease checks in Circle CI
CircleCI is only checking unit test and cross compilation -  https://github.com/minishift/minishift/blob/master/circle.yml#L25.
We should enable it to also check `gofmt` and `prerelease`.

budhrg:
Fix #445

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -22,7 +22,7 @@ dependencies:
 ## Customize test commands
 test:
     override:
-        - make test cross:
+        - make test cross fmtcheck prerelease:
                 pwd:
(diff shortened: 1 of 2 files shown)


budhrg:
Should be merged after https://github.com/minishift/minishift/pull/444 and rebase.

hferentschik:
Rebased and merged. Good one, I was thinking several times that we need to align the targets we run on the various CI systems.

PR #456, opened 2017-02-21
Issue #435 Fix upload artifacts for master build and added docs

### LINKED ISSUE #435: Enable CI to get triggered when a push to master occurs

budhrg:
Fix #435


hferentschik:
Merged. Let's wait to see whether it works until we resolve the issue.

Issue #454, opened 2017-02-21
Upgrade to B2D ISO 1.0.1


hferentschik:

hferentschik:
Merged via pull request #455

PR #459, opened 2017-02-21
Issue #454 Adding the missing docs change

LalatenduMohanty:


hferentschik:
Longing for the day were we don't have the docs in checked in like this anymore and where we don't need this check anymore. It is questionable anyways, to update this on each pull request. This this docs are the only docs available atm, they really should match the latest release version, but we make it match master. 

The sooner we get the HTML docs published the better.

Issue #468, opened 2017-02-22
Registration code never called for non proxy use cases


LalatenduMohanty:
Because of the below code change the "registration code moved behind the if loop of proxy information .So it is only called when proxy is provided. 
[1] https://github.com/minishift/minishift/commit/60b8ddcc504113313d81bd6e32e4b66053a02530

LalatenduMohanty:
During testing I manually checked if the VM is registered only for proxy use cases. For non proxy use cases I just checked if "minishift start" is working fine and I missed this.

budhrg:
I just now feel how important it is to have negative test case too.

gbraad:
Please create a PR, so a test build will be created in CI.

hferentschik:
Merged via pull request #469

LalatenduMohanty:
@budhrg The unit test actually do not run the command on the VM,  So we need to discuss and plan it.

PR #469, opened 2017-02-22
Issue #468 Fixing the code to call registration with or without proxy config

LalatenduMohanty:


LalatenduMohanty:
Tested with proxy and without proxy and cross checked by login to the VM and checked out put of "sudo scubscription-manager status" and "yum repolist"

Issue #450, opened 2017-02-21
Restructuring and modifying Using.md


Preeticp:
We need to restructure and modify Using.md document to more accurately reflect Minishift's role as well as split the document at natural break points. The doc is already quite huge and the addition of PR #449 will add material to the split doc.

thatdocslady:
Excellent idea to track things here, I would recommend maybe waiting until we do the asciidoc conversion (or at least you can throw ideas in here and then I can get some head start while I do the conversion, to be followed by collaborating on it further). WDYT?

Preeticp:
@thatdocslady , when do you plan to start with the Ascidoc conversion? Sure let's collaborate and work on this.
This issue emerged out of discussions on Persistent configuration section with @hferentschik 

I see three tasks emerging from the discussion and did not intend to track any more for this issue, for now.  
1. Rewording the first section to capture the exact nature of the commands and what exactly they do, renaming and so on
2. A simple split of the doc and renaming the split doc. This is already worked out and not much to be done here.
3. A proof-read of the mounting custom shared folders. This arose only after #449 was merged. This could be separated and created as an independent issue/clubbed with this.

We can think of further splits or raise issues as and when more content comes in, once the asciidoc conversion is done.
So currently, we are talking about one more doc, let's work it out so that it's least troublesome.

hferentschik:
Why would this depend on the Asciidoc conversion. I assume that we first get the new tooling in place and then do the (automated) conversion. If this in in place by then no problem, right? Also, I would also think that the tooling can render Markdown and Asciidoc. After all the generated docs will stay in Markdown. So I would suggest we get tooling on place without changing anything in the docs yet. We can just render HTML from Markdown. Next we start the conversion to Asciidoc for the documents we want to convert. I'd really like to separate the tooling from actual content conversion.

thatdocslady:
Following discussion with @Preeticp , points 1+2 will be done before the asciidoc conversion, and point 3 isn't ready to document anyway so it will be deferred. 

@hferentschik to your question, a reminder that we already spoke about it when we set the tooling mission going, the manually-authored content needs to be converted to asciidoc anyway and stored in gh because of conditional text and content reuse with downstream docs. The automated conversion to asciidoc within the container will be done only to the command reference. I will wait until the tooling is in place to do so but it'll still need to be done afterwards.

Preeticp:
[...]

Issue #403, opened 2017-02-14
Enhance documentation with diagrams


BrunoVernay:
Some visual could help to grasp the beast.
Does it looks like a good idea?  
Does it belongs elsewhere and we only provide link? 

I started some simple  [diagram](https://www.planttext.com/?text=RL8zJyCm49tnhpYP1OOaweggg0As4e64b8d5E4vgDJb7zgL2eFBRkQIqRWE3Xxoll3jx5KXv6hfM2BBK6YWish2nDK42ZrOrNdL2eAiCFAcUN0zVGW2mnzPA4tH7-JaeIykPghMVYgp3DxnPbqNs2zwxZonDqri9Ai2kyhinhtPjPJp8RyZIxV1k9Fy6u2nIEwnjmzYUpynmwZ15TZZtsk0k0QTZa3UQ53Oy-NDly7MQVBaiYppdcAl8oXTNmJKNbqm_E2HbqNXM_HKjvX9krIykT6kUqQCX4y1vlJ61_BIgVvpRuLYMZzEGfddSHkmymeT1MPCaEVlW3lo0Z7JnJfaOTtwYfeyHZKa7Iko5M1ci-6dy0000) that could benefits some review

hferentschik:
> Some visual could help to grasp the beast.

+1

> Does it looks like a good idea?

Sure

> Does it belongs elsewhere and we only provide link?

For the rendered docs one would add a actual png/gif/jpg to the docs. Of course if this diagram is creates from some sort of text based tools, one would also check in the source for that. What are you using there? 

The one concern I have is that we end up with all sort of diagram languages. Moving forward the docs will most likely use Asciidoctor which supports various type of diagrams - http://asciidoctor.org/docs/asciidoctor-diagram/. So whatever we use should be supported there as well. I can see _plantuml_ in the asciidoctor supported  diagram tools. Is this what you were using?

budhrg:
This seems to be interesting. :+1:  for idea.

BrunoVernay:
Indeed, I used PlantUML as it
- seems easier to collaborate on text.
- focus on the content rather than colors and styles

About collaboration, you should be able to clic on the link (in the first post) to see the diagram and comment. I am sure the diagram is missing some containers.
There might be better ways to proceed, I am open.

Also I did not find recognized "icon" or shapes for Container and VM. So I used the UML stereotype:  <<...>>.

hferentschik:
[...]

Issue #389, opened 2017-02-10
Abilty to use persistent volumes 


hferentschik:
See also https://github.com/minishift/minishift/issues/207#issuecomment-274310292

Preeticp:
Adding this note to track required documentation. We will need to update documentation and add an example, on the lines of the example we had for the /data host directory earlier, in the Interacting with your local OpenShift>Persistent volumes section.

jorgemoralespou:
@hferentschik For this issue, it should be pretty simple to create the volumes in versions pre 1.5 using same approach as "oc cluster does", it's running a Job creating a sufficient number of persistent volumes.

https://github.com/openshift/origin/blob/master/pkg/bootstrap/docker/openshift/pvsetup.go

Although, oc versions pre-1.5 will not have the registry configured as persistent.

The goal here, being able to provide consistent behavior for different versions of "openshift" from the minishift side. Same could apply to proxy, so you provide the ability to users of minishift using "oc pre 1.4" with the possibility to configure proxies.

hferentschik:
Merged via pull request #754

Issue #342, opened 2017-01-30
Update service command to be based on app routes


budhrg:
`minishift service` command is not working ATM. See https://github.com/minishift/minishift/issues/79#issuecomment-275987353.

```
$ oc whoami
developer

$ oc status
In project demo on server https://192.168.42.243:8443
http://nodejs-ex-demo.192.168.42.243.xip.io to pod port 8080-tcp (svc/nodejs-ex)
  dc/nodejs-ex deploys istag/nodejs-ex:latest <-
    bc/nodejs-ex source builds https://github.com/openshift/nodejs-ex on openshift/nodejs:4 
    deployment #1 deployed 5 minutes ago - 1 pod
3 warnings identified, use 'oc status -v' to see details.

$ oc get pods
NAME                READY     STATUS      RESTARTS   AGE
nodejs-ex-1-btptw   1/1       Running     0          5m
nodejs-ex-1-build   0/1       Completed   0          8m

$ oc expose svc/nodejs-ex
route "nodejs-ex" exposed

$ oc get routes
NAME        HOST/PORT                              PATH      SERVICES    PORT       TERMINATION
nodejs-ex   nodejs-ex-demo.192.168.42.243.xip.io             nodejs-ex   8080-tcp   

$ oc get endpoints nodejs-ex  
NAME        ENDPOINTS         AGE
nodejs-ex   172.17.0.2:8080   9m

$ oc get svc
NAME        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
nodejs-ex   172.30.16.186   <none>        8080/TCP   8m

$ minishift service nodejs-ex -n demo 
Service demo/nodejs-ex does not have a node port. To assign a port automatically, the service type must be NodePort or LoadBalancer, but this service is of type ClusterIP.

$ minishift service nodejs-ex --url -n demo           
Service demo/nodejs-ex does not have a node port. To assign a port automatically, the service type must be NodePort or LoadBalancer, but this service is of type ClusterIP.
```

hferentschik:
I think this is partly due to how this is implemented. The service command is using the Kube API to to this, but we want to do this via oc. There is also  mismatch I think what this command is supposed to report, since I think it is actually working, just not what one might expect.

Bottom line, this command needs to switch to use _oc_. This will also allow us to get rid of the Kubernetes dependency then. See also issue #168.

budhrg:
This command seems to be working well but needs some documentation changes.

```
$ minishift service -h
[...]

PR #473, opened 2017-02-23
Issue #450 Restructured and modified using.md

Preeticp:
As part of this PR:
1. Split the Using.md doc, and created a new doc under docs folder called configuring-openshift.md as the doc was getting too big and to distinguish between Minshift's usage wrt the VM and wrt OpenShift.
2. Moved the following sections from using.md to the configuring-openshift.md: Updating OpenShift configuration; Persistent volumes; and Private container registries.
2. Renamed the first section and its subsections, slightly modified the content to clear the distinction between VM, oc, OpenShift, and minishift.
3. Reordered sections logically, keeping in find the future HTML output.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ You can now install the latest beta version of minishift.
         $ minishift delete
 
    This command deletes everything in the `MINISHIFT_HOME/.minishift/machines/minishift` directory.
-   Other cached data and the [persistent configuration](./docs/using.md#persistent-configuration) are not removed.
+   Other cached data and the [persistent configuration](./docs/managing-minishift.md#persistent-configuration) are not removed.
 
 1. To completely uninstall Minishift, delete everything in the `MINISHIFT_HOME` directory
    (default `~/.minishift`) and `~/.kube`:
@@ -124,8 +124,9 @@ You can now install the latest beta version of minishift.
 ## Quickstart
 
 This section contains a brief demo of Minishift and the provisioned OpenShift cluster.
-For details on the usage of Minishift refer to the [Using Minishift](/docs/using.md) guide.
-The interaction with OpenShift is via the command line tool _oc_ which is copied to your host.
+For details on the usage of Minishift refer to the [Managing Minishift](/docs/managing-minishift.md) topic.
+
+The interaction with OpenShift is via the command line tool _oc_ which is copied to your host. For more information on how Minishift can assist you in interacting with and configuring your local OpenShift instance refer to the [Interacting with OpenShift](/docs/interacting-with-minishift.md) topic.
 
 For more information about the OpenShift cluster architecture,
 see [Architecture Overview](https://docs.openshift.org/latest/architecture/index.html) in the
@@ -168,7 +169,7 @@ with the KVM hypervisor driver.
   To verify the `oc` version, check the contents of the `~/.minishift/cache/oc` directory.
 
 For more information about interacting with OpenShift with the command-line interface and
-the Web console, see [Interacting with OpenShift](./docs/using.md#interacting-with-openshift)
+the Web console, refer to the [Interacting with OpenShift](./docs/interacting-with-openshift.md) topic.
 
 <a name="deploying-a-sample-application"></a>
 ### Deploying a sample application
@@ -230,7 +231,8 @@ docker ps
 
 The following documentation is available:
 
-* [Using Minishift](./docs/using.md)
+* [Managing Minishift](./docs/managing-minishift.md)
+* [Interacting with OpenShift](./docs/interacting-with-openshift)
 * [Command reference](./docs/minishift.md)
 * [Troubleshooting](./docs/troubleshooting.md)
 * [Installing docker-machine drivers](./docs/docker-machine-drivers.md)
--- a/docs/interacting-with-openshift.md
+++ b/docs/interacting-with-openshift.md
@@ -1,0 +1,147 @@
+# Interacting with your local OpenShift
+
+Minishift creates a virtual machine (VM) and provisions a local, single-node OpenShift cluster within this VM. The following sections describe how Minishift can assist you in interacting with and configuring your local OpenShift instance. For details about managing the Minishift VM refer to the [Managing Minishift](./managing-minishift.md) section.
+
+<!-- MarkdownTOC -->
+
+- [Interacting with OpenShift](#interacting-with-openshift)
+  - [OpenShift client binary \(oc\)](#openshift-client-binary-oc)
+  - [Login](#login)
+  - [Console](#console)
+  - [Services](#services)
+- [Updating OpenShift configuration](#updating-openshift-configuration)
+- [Persistent volumes](#persistent-volumes)
+
+<!-- /MarkdownTOC -->
+
+<a name="interacting-with-openshift"></a>
+## Interacting with OpenShift
+
+<a name="openshift-client-binary-oc"></a>
+### OpenShift client binary (oc)
+
+The `minishift start` command creates an OpenShift instance using the
+[cluster up](https://github.com/openshift/origin/blob/master/docs/cluster_up_down.md) approach.
+
+For this purpose it copies the `oc` binary onto  your host. You find it under
+`~/.minishift/cache/oc/\<OpenShift version\>/oc`. You can add this binary to your `PATH` variable
+in order to use `oc`, for example:
+
+    $ export PATH=$PATH:~/.minishift/cache/oc/v1.4.1
+
+In future versions we will provide a command to assist in setting up the `PATH`.
+See github issue [#142](https://github.com/minishift/minishift/issues/142).
+
+For an introduction to `oc` usage, refer to the [Get Started with the CLI](https://docs.openshift.com/enterprise/3.2/cli_reference/get_started_cli.html)
+topic in the OpenShift documentation.
+
+<a name="login"></a>
+### Login
+
+Per default, _cluster up_ uses an [AllowAllPasswordIdentityProvider](https://docs.openshift.org/latest/install_config/configuring_authentication.html#AllowAllPasswordIdentityProvider)
+for authentication against the local cluster. This means any non-empty username and password can
+be used to login to the local cluster. The recommended username and password are
+developer/developer, since it also has a default project _myproject_ set up.
+
+To login as administrator, use the system account:
+
+```shell
+$ oc login -u system:admin
+```
+
+In this case [client certificates](https://docs.openshift.com/enterprise/3.2/architecture/additional_concepts/authentication.html#api-authentication)
+are used which are stored in `~/.kube/config`. _cluster up_ will install
+the appropriate certificates as part of the bootstrap.
+
+**Note:** If you type `oc login -u system -p admin`, you will get logged in, but not as an administrator,
+but rather as an unprivileged user with no particular rights.
+
+To view the currently available login contexts, run:
+
+```
+$ oc config view
+```
+
+<a name="console"></a>
+### Console
+
+To access the [OpenShift console](https://docs.openshift.org/latest/architecture/infrastructure_components/web_console.html),
+you can run this command in a shell after starting Minishift to get the URL address:
+
+```shell
+$ minishift console --url
+```
+Alternatively, after starting Minishift, you can use the command below to directly open the console in a browser:
+
+```shell
+$ minishift console
+```
+
+<a name="services"></a>
+### Services
+
+To access a service exposed with a node port, run this command in a shell after starting Minishift to get the address:
+
+```shell
+$ minishift service [-n NAMESPACE] [--url] NAME
+```
+
+<a name="updating-openshift-configuration"></a>
+## Updating OpenShift configuration
+
+  Once OpenShift is running, you can view and change the master and
+  node configuration of your OpenShift cluster.
+
+  You can view the current OpenShift master configuration (_master-config.yaml_) via:
+
+  ```shell
+  $ minishift openshift config view
+  ```
+
+  For displaying the node configuration, you can specify the `target` flag.
+  For more details about the `view` command refer to its [synopsis](./minishift_openshift_config_view.md).
+
+  Let's look at [Cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) (CORS) as an example for patching the OpenShift master configuration.
+  Per default, OpenShift will only allow cross origin resource requests from the IP of the
+  cluster as well as localhost. This is specified via the `corsAllowedOrigins` property in the
+  [master configuration](https://docs.openshift.com/enterprise/3.0/admin_guide/master_node_configuration.html#master-configuration-files) (_master-config.yaml_). To change this value and allow
+  cross origin requests from all domains, one can execute:
+
+  ```
+  $  minishift openshift config set --patch '{"corsAllowedOrigins": [".*"]}'
+  ```
+
+  Per default, the master configuration is targeted, but you can also path the node config
+  by specifying the `target` flag. For more details about the
+  `set` command refer to its [synopsis](./minishift_openshift_config_set.md).
+
+  **Note:** OpenShift will be restarted after applying the patch.
+
+<a name="persistent-volumes"></a>
+## Persistent volumes
+
+Minishift supports [PersistentVolumes](https://docs.openshift.org/latest/dev_guide/persistent_volumes.html)
+of type `hostPath`. These PersistentVolumes are mapped to a directory inside the Minishift VM.
+
+The MiniShift VM boots into a tmpfs, so most directories will not be persisted across reboots (for example, when you use `minishift stop`).
+However, MiniShift is configured to persist OpenShift specific configuration files and docker files stored under the following host directories respectively:
+
+* `/var/lib/minishift`
+* `/var/lib/docker`
+
+Here is an example PersistentVolume config to persist data in the `/var/lib/minishift` directory:
+
+```yaml
+apiVersion: v1
+kind: PersistentVolume
+metadata:
+  name: pv
+spec:
+  accessModes:
+    - ReadWriteOnce
+  capacity:
+    storage: 5Gi
+  hostPath:
+    path: /var/lib/minishift/pv
+```
+Efforts to let the user configure persistent-volumes are on, see github issue [#389](https://github.com/minishift/minishift/issues/389)
--- a/docs/managing-minishift.md
+++ b/docs/managing-minishift.md
@@ -1,90 +1,56 @@
-# Using Minishift
+# Managing Minishift
 
 The following sections describe different aspects of using Minishift and provide an
 overview of different components and services.
 
 <!-- MarkdownTOC -->
 
-- [Managing your OpenShift instance](#managing-your-openshift-instance)
-  - [Starting OpenShift](#starting-openshift)
-  - [Stopping OpenShift](#stopping-openshift)
-  - [Deleting OpenShift](#deleting-openshift)
-  - [Updating OpenShift configuration](#updating-openshift-configuration)
+- [Managing Minishift](#managing-minishift)
+  - [Minishift start](#minishift start)
+  - [Minishift stop](#minishift-stop)
+  - [Minishift delete](#minishift-delete)
 - [Minishift runtime options](#minishift-runtime-options)
   - [Flags](#flags)
   - [Environment variables](#environment-variables)
   - [Persistent configuration](#persistent-configuration)
     - [Setting persistent configuration values](#setting-persistent-configuration-values)
     - [Unsetting persistent configuration values](#unsetting-persistent-configuration-values)
-- [Interacting with OpenShift](#interacting-with-openshift)
-  - [OpenShift client binary \(oc\)](#openshift-client-binary-oc)
-  - [Login](#login)
-  - [Console](#console)
-  - [Services](#services)
 - [HTTP/HTTPS Proxies](#httphttps-proxies)
 - [Mounted host folders](#mounted-host-folders)
   - [Mounting custom shared folders](#mounting-custom-shared-folders)
 - [Networking](#networking)
-- [Persistent volumes](#persistent-volumes)
-- [Private container registries](#private-container-registries)
 
 <!-- /MarkdownTOC -->
 
-<a name="managing-your-openshift-instance"></a>
-## Managing your OpenShift instance
+<a name="managing-minishift"></a>
+## Managing Minishift
 
-This section contains information about basic virtual machine and OpenShift management operations.
+When you use Minishift, you interact with two distinct components, both of which are managed by Minishift:
+- the virtual machine (VM) created by Minishift and
+- the OpenShift instance provisioned by Minishift within the VM.
 
-<a name="starting-openshift"></a>
-### Starting OpenShift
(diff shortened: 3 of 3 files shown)


Preeticp:
@hferentschik, @gbraad, @lalatendum, I have a few questions:
1.  In the Managing Minishift Section (Minishift Start) I have tried to make the distinction between Minishift, VM, `oc`, and OpenShift clear. Is there any more information on this we can add to this or the other sub-sections? 
2.  Would it make sense to create a new doc for the 'Interacting with OpenShift' section too? It is on the lines of 'Configuring Openshift', hence the thought. At present it is a relatively small section, but could make sense if we intend to add more content to this.
3. Would it be more logical to move the 'Proxy' section below the 'Minishift runtime options' section. I am looking at this from the HTML output perspective, it seems logical to have 'Interacting with Openshift' Section and the 'Configuring OpenShift' sections one after the other.

docs/configuring-openshift.md
```diff
@@ -0,0 +1,138 @@
...
+
+  Per default, the master configuration is targeted, but you can also path the node config
+  by specifying the `target` flag. For more details about the
+  `set` command refer to its [synopsis](./minishift_openshift_config_set.md).
+
+  **Note:** OpenShift will be restarted after applying the patch.
+
+
+<a name="mounted-host-folders"></a>
+## Mounted host folders
```
hferentschik: Host folders is not really about OpenShift. This is a VM thing.

docs/configuring-openshift.md
```diff
@@ -0,0 +1,138 @@
...
+
+**Note:** If you mount the folder this way, you might run into issues when your password contains a `$` sign, as these
+are used by PowerShell as variables and get replaced. In that case, you can use `'` (single-quotes) instead and replace
+the value of `$env:computername` with the content of this variable.
+
+If your Windows account is tied to a Microsoft Account, you have to use the account as the email address you use for
+this. Eg. `jpillow@amigas.us`. The domain value, which contains the computername, is in that case is essential.
+
+<a name="networking"></a>
+## Networking
```
hferentschik: Not purely OpenShift specific either

docs/configuring-openshift.md
```diff
@@ -0,0 +1,138 @@
...
+
+The Minishift VM is exposed to the host system via a host-only IP address, that can be obtained
+with the `minishift ip` command.
+
+<a name="persistent-volumes"></a>
+## Persistent volumes
```
hferentschik: So right, persistent volumes would fall under OpenShift

docs/using.md
```diff
@@ -22,70 +21,35 @@ overview of different components and services.
...
+<a name="managing-minishift"></a>
+## Managing Minishift
 
-This section contains information about basic virtual machine and OpenShift management operations.
+This section contains information about the Minishift virtual machine and the management of your local OpenShift instance using Minishift.
 
-<a name="starting-openshift"></a>
-### Starting OpenShift
+<a name="minishift-start"></a>
+### Minishift start
```
hferentschik: 👍 Right, I think this makes more sense

docs/using.md
```diff
@@ -22,70 +21,35 @@ overview of different components and services.
...
+### Minishift stop
 
 The [`minishift stop`](./minishift_stop.md) command is used to stop your OpenShift instance.
 This command shuts down the Minishift virtual machine, but preserves the cluster state.
 Starting Minishift again will restore the cluster, allowing you to continue work from where you left-off.
 
-<a name="deleting-openshift"></a>
-### Deleting OpenShift
+<a name="minishift-delete"></a>
+### Minishift delete
```
hferentschik: 👍

Preeticp:
Thanks @hferentschik 
>Seems there was a bit too much moved into the configuring OpenShift document, but I guess generally the split makes sense. 
So if we move the mounted host folders and Networking back to Using.md as you suggest, it would then balance out I guess. 

>Either way, we need some introductory section in using.md, which actually describes this conceptual split between VM and OpenShift and that Minishift manages both.
>Last but not least, we need then at some stage also think about how and where we link into the new document. At the moment the content in there is orphaned.
Will work on the introduction. 
I have linked up the configuring OpenShift doc in the `minishift start` command section, but a link in the Introduction part itself would make it visible and upfront.

hferentschik:
> So if we move the mounted host folders and Networking back to Using.md as you suggest, it would then balance out I guess.

I guess 

> Will work on the introduction.

+1

As said, for me the direction this goes makes sense. Would like to know what others think on this one.

docs/configuring-openshift.md
```diff
@@ -0,0 +1,79 @@
+# Configuring OpenShift
+The following sections describe how Minishift can be used to configure OpenShift. For details about managing the Minishift VM refer to the [Using Minishift](./using.md) section.
+
```
LalatenduMohanty: IMO saying `how Minishift can be used to configure OpenShift` is wrong here. We can not configure any OpenShift using Minishift but the OpenShift instance started by Minishift. So the correct sentence would be `The following sections describe how the OpenShift instance started by Minishift can be configured`

docs/configuring-openshift.md
```diff
@@ -0,0 +1,79 @@
...
+<!-- MarkdownTOC -->
+
+- [Updating OpenShift configuration](#updating-openshift-configuration)
+- [Persistent volumes](#persistent-volumes)
+- [Private container registries](#private-container-registries)
+
```
LalatenduMohanty: When I checked the rendered pages I saw an extra empty line between `Persistent volumes` and `Private container registries` , not sure why. You might want to take a look.

docs/configuring-openshift.md
```diff
@@ -0,0 +1,79 @@
...
+```
+
+<a name="private-container-registries"></a>
+## Private container registries
+
+To access a private container registry, follow the steps on [this page](http://kubernetes.io/docs/user-guide/images/).
+
+We recommend you to use ImagePullSecrets, but if you would like to configure access on the
+Minishift VM you can place the `.dockercfg` in the `/home/docker` directory or the `config.json`
+in the `/home/docker/.docker` directory.
```
LalatenduMohanty: @hferentschik @praveenkumar @budhrg @gbraad Is this relevant? Has anyone tested this? Because I do not much information on this.

docs/using.md
```diff
@@ -5,86 +5,56 @@ overview of different components and services.
 
 <!-- MarkdownTOC -->
 
-- [Managing your OpenShift instance](#managing-your-openshift-instance)
-  - [Starting OpenShift](#starting-openshift)
-  - [Stopping OpenShift](#stopping-openshift)
-  - [Deleting OpenShift](#deleting-openshift)
-  - [Updating OpenShift configuration](#updating-openshift-configuration)
+- [Managing Minishift](#managing-minishift)
+  - [Minishift start](#Minishift start)
```
LalatenduMohanty: The redirection will not work as it should be #minishift-start.

docs/configuring-openshift.md
gbraad: @LalatenduMohanty you mean the persistent volume?  or the private registry? miss context

LalatenduMohanty:
The separation of docs looks logical to me. However we need to make sure that we are doing it right i.e. not missing anything.

docs/configuring-openshift.md
Preeticp: Ah, the importance of blank spaces..

docs/configuring-openshift.md
hferentschik: I never tested this. No idea whether it works. I'd be tempted to remove it, since you really should use a pull secret

praveenkumar: I also didn't test this if that works of not.

docs/configuring-openshift.md
```diff
@@ -0,0 +1,69 @@
+# Configuring OpenShift
+The following sections describe how Minishift can be used to configure your local OpenShift instance provisioned by the Minishift VM. For details about managing the Minishift VM refer to the [Using Minishift](./using.md) section.
```
hferentschik: "The following sections describe how Minishift can assist in the configuration of ..."

docs/configuring-openshift.md
```diff
@@ -0,0 +1,69 @@
...
+
+- [Updating OpenShift configuration](#updating-openshift-configuration)
+- [Persistent volumes](#persistent-volumes)
+
+<!-- /MarkdownTOC -->
+
+<a name="updating-openshift-configuration"></a>
+## Updating OpenShift configuration
+
+  Once you have [started](#starting-openshift) OpenShift, you can view and change the master and
```
hferentschik: that link cannot be right anymore after the split. Please make sure that all links still work.

hferentschik: "Once OpenShift is running" (you are not really "starting" OpenShift in our case)

docs/configuring-openshift.md
```diff
@@ -0,0 +1,69 @@
...
+<a name="persistent-volumes"></a>
+## Persistent volumes
+
+Minishift supports [PersistentVolumes](https://docs.openshift.org/latest/dev_guide/persistent_volumes.html)
+of type `hostPath`. These PersistentVolumes are mapped to a directory inside the Minishift VM.
+
+The MiniShift VM boots into a tmpfs, so most directories will not be persisted across reboots (for example, when you use `minishift stop`).
+However, MiniShift is configured to persist files stored under the following host directories:
+
+* `/data`
```
hferentschik: I don't think this is true atm. @praveenkumar WDYT?

docs/configuring-openshift.md
```diff
@@ -0,0 +1,69 @@
...
+* `/var/lib/minishift`
```
hferentschik: @praveenkumar '/var/lib/minishift' and '/var/lib/docker' is that correct.

docs/configuring-openshift.md
```diff
@@ -0,0 +1,69 @@
...
+metadata:
+  name: pv0001
+spec:
+  accessModes:
+    - ReadWriteOnce
+  capacity:
+    storage: 5Gi
+  hostPath:
+    path: /data/pv0001/
+```
```
hferentschik: We might want to add a reference to issue #389 and the fact that we want to improve the persistence volume setup.

Preeticp:
Thanks @LalatenduMohanty, based on your suggestion and everyone's input, it was decided to remove the Private registry section since it was not tested by anybody. I have also addressed your other comments.

docs/using.md
```diff
@@ -5,86 +5,56 @@ overview of different components and services.
...
-- [Persistent volumes](#persistent-volumes)
-- [Private container registries](#private-container-registries)
 
 <!-- /MarkdownTOC -->
 
-<a name="managing-your-openshift-instance"></a>
-## Managing your OpenShift instance
+<a name="managing-minishift"></a>
+## Managing Minishift
+When you use Minishift, you actually interact with:
```
hferentschik: "When you use Minishift, you interact with two distinct components, both of which are managed by Minishift:"

hferentschik: Could you please add a empty line between header and text. IMO that reads better.

docs/using.md
```diff
@@ -5,86 +5,56 @@ overview of different components and services.
...
+- the virtual machine(VM) created by Minishift and
+- the OpenShift instance provisioned by the VM.
```
hferentschik: here a diagram would be awesome. See also #403. I like the one referenced in the issue. Might needs some minor tweaks. Might be outside of the scope of this issue, even though we could add a single png for now.

docs/using.md
```diff
@@ -5,86 +5,56 @@ overview of different components and services.
...
+This section contains information about managing the Minishift VM. For details on managing your local OpenShift instance using Minishift refer to the [Interacting with OpenShift](#interacting-with-openshift) and the [Configuring OpenShift](./configuring-openshift.md) sections.
 
-<a name="starting-openshift"></a>
-### Starting OpenShift
+<a name="minishift-start"></a>
+### Minishift start
 
-The [`minishift start`](./minishift_start.md) command is used to start your OpenShift instance.
-This command creates and configures a virtual machine that runs a single-node OpenShift instance.
+The [`minishift start`](./minishift_start.md) command is used to start a local OpenShift instance.
```
hferentschik: first it creates the VM and then it _provisions_ OpenShift

docs/using.md
```diff
@@ -5,86 +5,56 @@ overview of different components and services.
...
+This command creates and configures the Minishift VM that runs a single-node OpenShift instance.
 
-<a name="stopping-openshift"></a>
-### Stopping OpenShift
+It also copies the `oc` binary to your host so that you can interact with OpenShift either through the `oc` command line tool or through the web console which can be accessed through the URL provided in the output of the `minishift start` command. For further information refer to the [Interacting with OpenShift](#interacting-with-openshift) and [Configuring OpenShift](./configuring-openshift.md) sections.
```
hferentschik: there is a specific command to get the URL, that should be used, instead of referring to some output of the start commands which will disappear. 

    $ minishift console --url

hferentschik: or 

    $ minishift console 

if you want to open in the browser directly

docs/using.md
```diff
@@ -5,86 +5,56 @@ overview of different components and services.
...
+
+<a name="minishift-stop"></a>
+### Minishift stop
 
 The [`minishift stop`](./minishift_stop.md) command is used to stop your OpenShift instance.
-This command shuts down the Minishift virtual machine, but preserves the cluster state.
-Starting Minishift again will restore the cluster, allowing you to continue work from where you left-off.
+This command shuts down the Minishift VM, but preserves the OpenShift cluster state.
+Starting Minishift again will restore the OpenShift cluster, allowing you to continue work from where you left-off.
```
hferentschik: Let's add a note here, that on re-start one needs to repeat the same parameters used in the original start command. This is the most common usage issue we have atm. We can also add a reference to issue #179

docs/using.md
```diff
@@ -323,40 +226,69 @@ authenticated users. The following procedure describes how to mount this folder.
...
-    storage: 5Gi
-  hostPath:
-    path: /data/pv0001/
+<a name="interacting-with-openshift"></a>
+## Interacting with OpenShift
+
+<a name="openshift-client-binary-oc"></a>
+### OpenShift client binary (oc)
+
+The `minishift start` command creates an OpenShift instance using the
```
hferentschik: This probably could go into the OpenShift document, right?

docs/using.md
```diff
@@ -323,40 +226,69 @@ authenticated users. The following procedure describes how to mount this folder.
...
+    $ export PATH=$PATH:~/.minishift/cache/oc/v1.4.1
+
+In future versions we will provide a command to assist in setting up the `PATH`.
+See Github issue [#142](https://github.com/minishift/minishift/issues/142).
+
+For an introduction to `oc` usage, refer to the [Get Started with the CLI](https://docs.openshift.com/enterprise/3.2/cli_reference/get_started_cli.html)
+section in the OpenShift documentation.
+
+<a name="login"></a>
+### Login
```
hferentschik: Same for this. OpenShift document?

docs/using.md
```diff
@@ -323,40 +226,69 @@ authenticated users. The following procedure describes how to mount this folder.
...
+run this command in a shell after starting Minishift to get the address:
+
+```shell
+$ minishift console
 ```
 
-<a name="private-container-registries"></a>
-## Private container registries
+<a name="services"></a>
+### Services
```
hferentschik: That's OpenShift, even though this is broken atm. See issue #342. Hopefully by the end of the sprint we can correct the docs.

docs/using.md
Preeticp: Totally agree @hferentschik and I suggested the same here in question 2: https://github.com/minishift/minishift/pull/473#issuecomment-281989882

docs/configuring-openshift.md
praveenkumar: No, we don't put files under `/data` now.

docs/configuring-openshift.md
praveenkumar: openshift specific configuration files are part of `/var/lib/minishift/` and docker stuff is under `/var/lib/docker/`

docs/configuring-openshift.md
LalatenduMohanty: @gbraad I meant private registry.

docs/configuring-openshift.md
hferentschik: thanks @praveenkumar, that's what I thought. @Preeticp, So this has to go.

docs/configuring-openshift.md
hferentschik: Ok, really how useful it is to put the persistent volume data there I am not sure. These are just symbolic links right? What is the path to the data disk otherwise? Maybe we can suggest something better here.

docs/using.md
hferentschik: 👍

docs/configuring-openshift.md
Preeticp: Have resolved this.

docs/configuring-openshift.md
```diff
@@ -0,0 +1,147 @@
+# Interacting with your local OpenShift
+
+The following sections describe how Minishift can assist you in interacting with and configuring your local OpenShift instance provisioned by the Minishift VM. For details about managing the Minishift VM refer to the [Using Minishift](./using.md) section.
```
hferentschik: "provisioned by" OpenShift is not 'provisioned by' the Minishift VM. It is either provision by Minishift

hferentschik: Should _using.md_ not be renamed as well, then?

docs/configuring-openshift.md
```diff
@@ -0,0 +1,147 @@
...
+Minishift supports [PersistentVolumes](https://docs.openshift.org/latest/dev_guide/persistent_volumes.html)
+of type `hostPath`. These PersistentVolumes are mapped to a directory inside the Minishift VM.
+
+The MiniShift VM boots into a tmpfs, so most directories will not be persisted across reboots (for example, when you use `minishift stop`).
+However, MiniShift is configured to persist OpenShift specific configuration files and docker files stored under the following host directories respectively:
+
+* `/var/lib/minishift`
+* `/var/lib/docker`
+
+Here is an example PersistentVolume config to persist data in the `/data` directory:
```
hferentschik: That is still wrong. /data is not going to work.

docs/configuring-openshift.md
hferentschik: the example and the paragraph above still mention /data

docs/using.md
```diff
@@ -5,86 +5,52 @@ overview of different components and services.
```
hferentschik: Should we not rename _using.md_ to something more meaningful then as well?

docs/using.md
```diff
@@ -5,86 +5,52 @@ overview of different components and services.
...
 <!-- /MarkdownTOC -->
 
-<a name="managing-your-openshift-instance"></a>
-## Managing your OpenShift instance
+<a name="managing-minishift"></a>
+## Managing Minishift
 
-This section contains information about basic virtual machine and OpenShift management operations.
+When you use Minishift, you interact with two distinct components, both of which are managed by Minishift:
+- the virtual machine(VM) created by Minishift and
```
hferentschik: space between machine and (VM)

docs/using.md
```diff
@@ -5,86 +5,52 @@ overview of different components and services.
...
+- the OpenShift instance provisioned by the VM.
```
hferentschik: Is it not provisioned "by" the VM. Maybe on or within the VM. Otherwise OpenShift is provisioned by Minishift

docs/using.md
```diff
@@ -5,86 +5,52 @@ overview of different components and services.
...
 
-<a name="starting-openshift"></a>
-### Starting OpenShift
+This section contains information about managing the Minishift VM. For details on using Minishift to manage your local OpenShift instance refer to the [Interacting with your local OpenShift](./configuring-openshift.md) section.
```
hferentschik: The following paragraphs, sections (more than one)

docs/using.md
```diff
@@ -5,86 +5,52 @@ overview of different components and services.
...
 
-The [`minishift start`](./minishift_start.md) command is used to start your OpenShift instance.
-This command creates and configures a virtual machine that runs a single-node OpenShift instance.
+<a name="minishift-start"></a>
+### Minishift start
 
-<a name="stopping-openshift"></a>
-### Stopping OpenShift
+The [`minishift start`](./minishift_start.md) command creates and configures the Minishift VM that provisions a local, single-node OpenShift instance.
```
hferentschik: Again, it is not the VM which provisions OpenShift. The VM is not an actor.

docs/configuring-openshift.md
Preeticp: So would replacing `/data` here with `/var/lib/minishift` and  below  path: /data/pv0001/ > path: /var/lib/minishift/pv0001/ be technically correct?

praveenkumar: @Preeticp this is something we need to test it out before put in the documents. I didn't test those stuff but if someone did and can give a go ahead then might be good to add it.

@LalatenduMohanty @gbraad  @budhrg  have you tested that config stuff?

gbraad:
No, this is why I asked earlier if Lalatendu mentioned these changes also.
Are they in the current testing plan? If not, let's create a task to review
this and add updated documentation separate from this task.

This way we can also line it up with other tasks, such as #480, and other
storage options. WDYT?

Preeticp:
@gbraad I agree, I can create an issue for the same or this is being tracked in #389, we could probably add a documentation and needs Investigation tags for the same? 
So as of now since we do not have validation for the example (post confirmation from @LalatenduMohanty and @budhrg) I will delete this and add a note about the same to the issue I create or in #389.

docs/configuring-openshift.md
LalatenduMohanty: Nope, I have only tested /var/lib/minishift.

hferentschik: this is a host path mount, no reason this should not work. The disconnect here is the use of /data (/data/pv0001/) which is not persistent. So provided you create for example _/var/lib/minishift/pv_ and use that directory it should all be good.

README.md
```diff
@@ -230,7 +231,8 @@ docker ps
 
 The following documentation is available:
 
-* [Using Minishift](./docs/using.md)
+* [Managing Minishift](./docs/managing-minishift.md)
+* [Interacting with OpenShift](./docs/interacting-with-openshift)
```
hferentschik: interacting-with-openshift.md

docs/interacting-with-openshift.md
```diff
@@ -0,0 +1,147 @@
...
+[cluster up](https://github.com/openshift/origin/blob/master/docs/cluster_up_down.md) approach.
+
+For this purpose it copies the `oc` binary onto  your host. You find it under
+`~/.minishift/cache/oc/\<OpenShift version\>/oc`. You can add this binary to your `PATH` variable
+in order to use `oc`, for example:
+
+    $ export PATH=$PATH:~/.minishift/cache/oc/v1.4.1
+
+In future versions we will provide a command to assist in setting up the `PATH`.
+See github issue [#142](https://github.com/minishift/minishift/issues/142).
```
hferentschik: It's GitHub not github

hferentschik:
Merged, thanks.

README.md
Preeticp: :( sorry, will be more careful next time

Issue #470, opened 2017-02-23
Edit and expand CONTRIBUTING.md


gbraad:
At the moment our `CONTRIBUTING.md` reflects our parent project.

thatdocslady:
Details:
- Remove CLA
- Expand contribution process info

hferentschik:
Merged via pull request #479

PR #479, opened 2017-02-23
Issue #470 edited and expanded contributing.md

thatdocslady:
Fixes issue #470 

This PR includes:

- More detailed and helpful contribution process description and contact info.
- Removed CLA.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,27 +2,23 @@
 
 ## Filing issues
 
-File issues using the standard Github issue tracker for the repo.
+File issues using the standard [Github issue tracker](https://github.com/minishift/minishift/issues) for the repository.
+Before you submit a new issue, we recommend that you:
 
-## How to become a contributor and submit your own code
+- Read the latest [release notes](https://github.com/minishift/minishift/releases) to check if the issue is known.
+- Search the list of issues to see if anyone already submitted a similar issue.
 
-### Contributor License Agreements
+## Contributing patches
 
-We'd love to accept your patches! Before we can take them, we have to jump a couple of legal hurdles.
+Thank you for your contributions! Please follow this process to submit a patch:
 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -2,27 +2,23 @@
...
 
-### Contributor License Agreements
+## Contributing patches
 
-We'd love to accept your patches! Before we can take them, we have to jump a couple of legal hurdles.
+Thank you for your contributions! Please follow this process to submit a patch:
 
-Please fill out either the individual or corporate Contributor License Agreement (CLA).
+1. Create an issue describing your proposed change to the repository.
+1. The repository owners will triage and respond to your issue promptly.
```
gbraad: Not a huge fan of the repeated index. This expects people to read it from the GitHub interface...

thatdocslady:
@gbraad this is our convention to use the 1-1-1-1 because this document is served whenever someone is about to contribute something and is available via live preview in github :)

CONTRIBUTING.md
LalatenduMohanty: -1 for the repeated index. We should use "Numbered Lists" formatting for markdown [1] . So that the document communicates the same for command line editors , notepads and github .
[1] https://en.support.wordpress.com/markdown-quick-reference/

thatdocslady:
@LalatenduMohanty @gbraad I was in this situation before and @hferentschik asked that I use the repeated index in all our docs. From my side I don't have a preference but we should decide on this together. Maybe we consider this document an exception, but as a general rule we try to let the markdown parser (and eventually when we switch to asciidoc, the asciidoctor parser) to interpolate the numbers in lists.

hferentschik:
I actually went ahead and merged this. Moving forward this will be convertd to asciidoc anyways, making this discussion obsolete. Also for consistency reasons I'd like to continue what we are doing in all other markdown files atm

hferentschik:
Merged, thanks.

Issue #472, opened 2017-02-23
Restore Kubernetes copyright notices


richardfontana:
This commit https://github.com/minishift/minishift/commit/6abc7d6b73aa903b8d542f311e72d528a7a27e32
and anything similar need to be reverted to the extent that any code from Minikube remains in a source file from which the Kubernetes copyright notice was removed. It would be okay for such files to have, for example:

> Copyright 2016 The Kubernetes Authors All rights reserved.
> Copyright 2016 Red Hat, Inc.

But you can't delete the earlier copyright notice.

hferentschik:
@LalatenduMohanty can you please hold off in this one?

hferentschik:
Merged via pull request #488

PR #488, opened 2017-02-27
Issue #472 Restore Kubernetes copyright notices

LalatenduMohanty:
This reverts commit 6abc7d6b73aa903b8d542f311e72d528a7a27e32.

### DIFF
--- a/hack/boilerplate/boilerplate.go.txt
+++ b/hack/boilerplate/boilerplate.go.txt
@@ -1,5 +1,5 @@
 /*
-Copyright (C) YEAR Red Hat, Inc.
+COPYRIGHTHOLDER.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
-   Copyright (C) 2016 Red Hat, Inc.
+   Copyright 2016 The Kubernetes Authors All rights reserved.
 
    Licensed under the Apache License, Version 2.0 (the "License");
    you may not use this file except in compliance with the License.
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-# Copyright (C) 2016 Red Hat, Inc.
+# Copyright 2016 The Kubernetes Authors All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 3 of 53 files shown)


hack/boilerplate/boilerplate.go.txt
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (C) YEAR Red Hat, Inc.
+Copyright YEAR The Kubernetes Authors All rights reserved.
```
hferentschik: -1 This is not going to work. Now the prerelease task is failing when it verified the headers. All files which don't get the reverted header still have "Copyright (C) 2016 Red Hat, Inc.". 

_boilerplate.py_ needs to change to handle both variations of the header. I would suggest to use a regular expressions in this case which allows the two variants "The Kubernetes Authors All rights reserved" and "Red Hat, Inc."

hferentschik: This applies to all boilerplate templates.

LalatenduMohanty: @hferentschik I have missed it somehow.  Will fix it.

hferentschik: > I have missed it somehow. Will fix it.

👍

LalatenduMohanty: @hferentschik The easiest way to skip the files with Kubernetes boilerplate .But it is not right in my opinion as we might get new files from minikube in future. The current implementations is very deeply tied up with one boilerplate for one type of file.  I am still figuring out what would be the right thing to do here.

hferentschik: > The easiest way to skip the files with Kubernetes boilerplate 

-1

> The current implementations is very deeply tied up with one boilerplate for one type of file

I don't get it. There are boilerplate headers for the various file types, yes, but how is this all different than say the YEAR? Why don't you put COPYRIGHTHOLDER as a token in the various templates and treat this token the same way as the YEAR? See https://github.com/minishift/minishift/blob/8b500a691c670f5ee4de93a11b157de24f761fae/hack/boilerplate/boilerplate.py#L148-L148

Am I missing something here?

LalatenduMohanty: @hferentschik Never thought of putting the copyrightholder as a token as I thought it wont be elegant to put the whole paragraph as a token. But lets try that.

hack/boilerplate/boilerplate.go.txt
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright (C) YEAR Red Hat, Inc.
+COPYRIGHTHOLDER.
```
hferentschik: I was first going to ask why you ended up to replace the whole line with this token and not just the copy right holder itself. But I get it now. Looking at _boilerplate.py_, I have to say I find this a strange way of doing in general. But let's be done with it for now.

hferentschik:
Merged, thanks.

Issue #487, opened 2017-02-27
Create minishift-data.json to persists data specific to minishift instance


budhrg:
We need `minishift-data.json` to persists few data specific to minishift instance like 'oc' version etc.

Currently, we need following data as:
```
{
  "ocVersion": "1.4.1",
  "ocPath": "/home/budhram/.minishift/cache/oc/v1.4.1/oc",
  "isRegistered": false 
}
``` 

`isRegistered` is updated when we register a VM, so that we can read that in stop/delete instead of having to ssh into the VM (which might be in an error state) and `ocVersion` and `ocPath` are needed to make sure we are using correct oc version on start/stop/delete for instance. 

This is a requirement for #342 and other issues.

@hferentschik Please add if I miss anything here.

hferentschik:
In code we need utility methods to read/write the config file and parse it into a types config struct of sorts.

gbraad:
@hferentschik @budhrg This is something I also need for the hostfolders.

budhrg:
As a first cut of it, we agree to have minimal config.json file with :
```
{
  "ocPath": "/home/budhram/.minishift/cache/oc/v1.4.1/oc"
}
```
As we go further, we can add more properties.

gbraad:
I think the use of 'MinishiftData' confuses as it has not much to do with the actual data for minishift (such as the volume files for an instance). It is better to call this something with 'instance' in the naming.

hferentschik:
> I think the use of 'MinishiftData' confuses as it has not much to do with the actual data for minishift (such as the volume files for an instance). It is better to call this something with 'instance' in the naming.

+1 especially once we can manage multiple instances

hferentschik:
Merged via pull request #521

PR #490, opened 2017-02-28
Fix #489 Handle subscription manager proxy from shell environment variable

### LINKED ISSUE #489: subscription-manager version call failing in proxy environment
Currently we are parsing proxy details using regex and then passing it to subscription manager. subscription manager do honor the shell proxy env and should consumed it instead we do extra work from parsing.

praveenkumar:
Subscription manager honors the shell proxy env and we shouldn't parse manually those stuff which might be error prone.


pkg/minishift/registration/redhat.go
```diff
@@ -55,26 +55,7 @@ func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) er
...
-						configCommand = configCommand +
-							fmt.Sprintf("--server.proxy_password %s ", param.ProxyPassword)
-					}
-				}
-				if _, err := registrator.SSHCommand(configCommand); err != nil {
-					return err
-				}
-			}
-			subscriptionCommand := fmt.Sprintf("sudo subscription-manager register --auto-attach "+
+			subscriptionCommand := fmt.Sprintf("sudo -E subscription-manager register --auto-attach "+
```
praveenkumar: "-E" is used to preserve the environment variable for the shell we need it here so our proxy env variable can be picked easily.

pkg/minishift/registration/redhat_test.go
```diff
@@ -27,14 +27,10 @@ import (
...
-		Password:        "foo",
-		ProxyUsername:   "foo",
-		ProxyPassword:   "foo",
-		ProxyServer:     "foo",
-		ProxyServerPort: "foo",
+		Username: "foo",
+		Password: "foo",
 	}
-	expectedCMDRegistration = fmt.Sprintf("sudo subscription-manager register --auto-attach --username %s --password %s ",
+	expectedCMDRegistration = fmt.Sprintf("sudo -E subscription-manager register --auto-attach --username %s --password %s ",
```
praveenkumar: "-E" is used to preserve the environment variable for the shell we need it here so our proxy env variable can be picked easily.

LalatenduMohanty:
@praveenkumar the only concern with the PR is I did not get any documentation about subscription manager honoring shell proxy env variable (at least in official documentation of RHEL) [1] [2]. 

May be the subscription manager is not aware of the proxy but it is the Linux shell which is doing the trick as once we set proxy env in shell all internet requests gets redirected to the proxy.

[1] https://access.redhat.com/solutions/57669
[2] https://access.redhat.com/documentation/en-US/Red_Hat_Subscription_Management/1/html/RHSM/rhsm-http-proxy.html

pkg/minishift/registration/redhat.go
hferentschik: Ok, but how was it working before? It was not then, right?

hferentschik:
> the only concern with the PR is I did not get any documentation about subscription manager honoring shell proxy env variable (at least in official documentation of RHEL) [1] [2].

I would also feel better, knowing that this is an officially supported way of doing this.

> May be the subscription manager is not aware of the proxy but it is the Linux shell which is doing the trick as once we set proxy env in shell all internet requests gets redirected to the proxy.

-1 the shell cannot do such a thing. The shell can pass its environment to a child process (subscription-manager), but it then cannot control its connections and whether or not it uses proxies.

I am more interested in getting an explanation how we initially thought it was working. 

Overall I am feeling reluctant to change this code once more. We already went through many iterations on this one. We don't have official confirmation that the approach via env variables works. I am not a big fan of the parsing of the proxy variable either, but  I am not sure whether it makes sense to change this at this point in time.

hferentschik:
Also using the proxy variables as passed to the registrar is more explicit. It becomes clear directly that proxy is used. If we remove this, there is some undocumented magic here, which might not be obvious to someone new to the code.

praveenkumar:
> I am more interested in getting an explanation how we initially thought it was working.

Initially we put our shell code in wrong place which was after registration and didn't able to pick those details. That's why it was not working.

>Overall I am feeling reluctant to change this code once more. We already went through many iterations on this one. 

I understand but manually parsing when it not required doesn't make sense, we might hit a case where our regex will fail to determine  username, password and proxy url, but if we are already asking user the complete string then this condition can be eliminated. 

> Also using the proxy variables as passed to the registrar is more explicit.

I agree but that is not clean, if we want to be more explicit then I think we should use username, password and proxy_url from user and not parse ourself.

hferentschik:
> I agree but that is not clean, if we want to be more explicit then I think we should use username, password and proxy_url from user and not parse ourself.

true true, but that conflicts how Docker and OpenShift takes proxy parameters. In this case we have to have much more parameters, since we have the ability to differentiate between HTTP and HTTPS proxy.

hferentschik:
> I understand but manually parsing when it not required doesn't make sense, we might hit a case where our regex will fail 

true, but we have some basic unit tests around this and if someone hits a bug we can fix it. 

> to determine username, password and proxy url, but if we are already asking user the complete string then this condition can be eliminated.

I agree, but it still needs to cover the OpenShift and Docker case as well.

pkg/minishift/registration/redhat.go
praveenkumar: @hferentschik right it was not working before.

hferentschik:
@praveenkumar let's also see what else @LalatenduMohanty has to say

praveenkumar:
> let's also see what else @LalatenduMohanty has to say

Agree will wait for his ack.

praveenkumar:
retest this please

LalatenduMohanty:
@praveenkumar @hferentschik I can check with subscription-manager developers about the support for proxy_url through environment variable and then we can decide on this PR. WDYT?

praveenkumar:
> I can check with subscription-manager developers about the support for proxy_url through environment variable and then we can decide on this PR. WDYT?

Sure, go ahead and check with them.

hferentschik:
> I can check with subscription-manager developers about the support for proxy_url through environment variable

Let's start there. Personally, looking at the pros and cons, I prefer the current approach. The con of parsing the full proxy url is out-weight by the otherwise clearer and more transparent code. Also the parsing of the proxy url is in a single methods with unit tests. If someone hits a problem there it is easy to fix.

pkg/minishift/registration/redhat.go
```diff
@@ -55,26 +55,7 @@ func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) er
 		return err
```
LalatenduMohanty: @praveenkumar if `subscription-manager version` command needs network access for this command to execute then we need to pass `-E` here.

LalatenduMohanty: @praveenkumar if subscription-manager version command needs network access for this command to execute then we need to pass -E here.

hferentschik:
Can we come to a conclusion here? What's the general sentiment here? -1 from me.

pkg/minishift/registration/redhat.go
praveenkumar: @LalatenduMohanty Agree, adding those flag for preserve env.

praveenkumar:
@hferentschik From them mail reply which @LalatenduMohanty  sent makes me positive about this PR since it's supported and we can get away with all manual parsing stuff.

LalatenduMohanty:
@praveenkumar @hferentschik I am not sure. The existing approach is more readable (code) for sure .But this PR will result in to less number of lines of code.  I am usually go for more readable code.

hferentschik:
Any update on this @praveenkumar. Did you confirm that it works?

BTW, did you not spend once some time to check what we could do to integration test this? Did you make any progress there?

pkg/minishift/registration/redhat_test.go
```diff
@@ -27,22 +27,18 @@ import (
 
 var (
 	param = &RegistrationParameters{
-		Username:        "foo",
-		Password:        "foo",
```
hferentschik: Actually, I don't think we should remove this. We should keep this in place. The whole idea is/was to allow other registrars. It was always the intention to pass proxy variables viaa RegistrationParameters. If we remove these parameters, then we are again making things Red Hat subscription manager specific. 

To stick with our initial intention, we should keep this (and everything related to it) and just not make use of it in redhat.go.

hferentschik:
I think we should keep the code as it is right now (in terms that it parses the proxy_host and puts it into the registration parameters), but if we want to just ignore these parameters in _redhat.go_ and rely on the env variables.

pkg/minishift/registration/redhat_test.go
praveenkumar: > Actually, I don't think we should remove this. We should keep this in place. The whole idea is/was to allow other registrars.

We are not removing username/password it's just when I deleted proxy param it shifted bit down line.

cmd/minishift/cmd/delete.go
```diff
@@ -40,8 +40,6 @@ func runDelete(cmd *cobra.Command, args []string) {
 	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer api.Close()
 
-	setSubcriptionManagerParameters()
-
```
praveenkumar: This we are removing because it's conflict with start.go variables and viper not able to make out what to do.

cmd/minishift/cmd/delete.go
```diff
@@ -50,11 +48,6 @@ func runDelete(cmd *cobra.Command, args []string) {
 }
 
 func init() {
-	deleteCmd.Flags().String(username, "", "Username for the virtual machine unregistration.")
-	deleteCmd.Flags().String(password, "", "Password for the virtual machine unregistration.")
-
-	deleteCmd.Flags().AddFlagSet(subscriptionManagerFlagSet)
-
```
praveenkumar: Removing this because viper need different flagset and variable for different command.

hferentschik:
> Removing this because viper need different flagset and variable for different command.

I see

cmd/minishift/cmd/delete.go
hferentschik: So are you saying we don't have the username/password flags on stop/delete at all anymore? So they won't get passed to un-register either?

praveenkumar: @hferentschik yes we don't have username/password flag for stop/delete any more but that doesn't effect our rhel registratrar once we have #487 implemented then take that details from there and use it for delete/stop. We can create a issue for same just to track.

hferentschik: > yes we don't have username/password flag for stop/delete any more

:-( This is really not how I thought it would turn out in the end. It is after all just tailored now for RHEL subscription.

> but that doesn't effect our reel regitratrar

Sure, I get that

> once we have #487 implemented then take that details from there and use it for delete/stop. We can create a issue for same just to track.

Ok, so there is a follow up plan, at least in terms of tracking it. Even though I like the idea of doing this via storing the credentials in the config file, it has the "issue" of storing the password plain text. 

BTW, surely it would be possible to have the same flag name on multiple sub-commands, right? This is the easier way out for us atm, but technically it should be possible. I cannot imagine that Viper has a constraint that all flags need to be unique per command.

hferentschik:
Rebased and merged, thanks.

PR #492, opened 2017-03-01
Issue #440 Use MINISHIFT_VM_DRIVER for integration tests

### LINKED ISSUE #440: Use MINISHIFT_VM_DRIVER for integration tests
The VM driver should be controlled by MINISHIFT_VM_DRIVER, instead of the VM_DRIVER in `flags.go`.

budhrg:
Fix #440

### DIFF
--- a/test/integration/flags.go
+++ b/test/integration/flags.go
@@ -33,7 +33,7 @@ func TestMain(m *testing.M) {
 
 var binaryPath = flag.String("binary", fmt.Sprintf("../../out/%s-amd64/minishift", runtime.GOOS), "path to minishift binary")
 
-var vmDriver = getenv("VM_DRIVER", "virtualbox") // Get the driver value from environment variable or use default
+var vmDriver = getenv("MINISHIFT_VM_DRIVER", "virtualbox") // Get the driver value from environment variable or use default
 
 var args = flag.String("minishift-args", fmt.Sprintf("--vm-driver %s", vmDriver), "Arguments to pass to minishift")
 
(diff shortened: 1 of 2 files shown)


test/integration/flags.go
```diff
@@ -33,7 +33,7 @@ func TestMain(m *testing.M) {
 
 var binaryPath = flag.String("binary", fmt.Sprintf("../../out/%s-amd64/minishift", runtime.GOOS), "path to minishift binary")
 
-var vmDriver = getenv("VM_DRIVER", "virtualbox") // Get the driver value from environment variable or use default
+var vmDriver = getenv("MINISHIFT_VM_DRIVER", "virtualbox") // Get the driver value from environment variable or use default
```
hferentschik: !? - You are changing nothing here, expect renaming the variable. That's not the intention here. If it were just about the name, we might as well skip it. 

The whole idea is that if you use 'MINISHIFT_VM_DRIVER', you don't have to use the '--vm-driver flag' at all, since Viper will take the setting from the environment https://github.com/minishift/minishift/blob/master/docs/using.md#environment-variables

So just making sure that the variable is set in the environment which starts the test will make it work.

hferentschik: Right, we don't need the _vmDriver_ var. That is what the whole issue is about. I have a feeling that we have some misunderstanding what we want to achieve here. Also this needs to be synced with ull request #433. I recommend we close this pull request and first get #433 merged and take it from there.

budhrg: Sure @hferentschik :+1:

budhrg:
Closing it as it will be synced after PR https://github.com/minishift/minishift/pull/433 get merged.

Issue #359, opened 2017-02-02
Introduce context "minishift"


rhuss:
It would be awesome if minishift would name its context "minishift" so that one can easily switch back to the minishift instance with

```
oc set-config use-context minishift
```

`minikube` does it that way.

praveenkumar:
@rhuss I checked with v1.4.1 origin and seems like oc binary doesn't have set-config option in this version, is it something which will come later version?

```
$ ./.minishift/cache/oc/v1.4.1/oc version
oc v1.4.1+3f9807a
kubernetes v1.4.0+776c994
features: Basic-Auth GSSAPI Kerberos SPNEGO

$ ./.minishift/cache/oc/v1.4.1/oc set-config
Error: unknown command "set-config" for "oc
```

rhuss:
@praveenkumar  Sorry, I meant

```
oc config use-context minishift
```

The context themselves are defined in `~/.kube/config` and you can use this command to switch between different K8s / OpenShift installations (this file is shared between K8s and OpenShift)

rhuss:
Minishift adds a context but it is named like `/192-168-64-3:8443/admin` which is not very intuitive and changing per new installation.

jorgemoralespou:
@praveenkumar Look here for a sample of how we do it:
https://github.com/openshift-evangelists/oc-cluster-wrapper/blob/master/oc-cluster#L285-L289

2017-02-03 11:59 GMT+01:00 Roland Huß <notifications@github.com>:

praveenkumar:
@jorgemoralespou @rhuss Thanks for quick note and code-snip.

praveenkumar:
I spent some time before this PR and tried to understand why we can't use same approach as kubernetes do for context - https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go#L175-L194  here you can see that before it set the context it have kubernetes up and running and then they create the config file. 

Since we are using oc cluser up approach `oc` itself create config file for us so we need to modify it with available options to get the context right (which I did as part of PR).

jorgemoralespou:
@praveenkumar @hferentschik I would request you to think about profiles (if you have already thought on how to implement them) as the context setting will be very relevant to when that happens.

hferentschik:
> I would request you to think about profiles

[...]

PR #499, opened 2017-03-02
[WIP] Issue #359 Introduce context minishift

praveenkumar:
This patch enable default context to "minishift"

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	units "github.com/docker/go-units"
 	"github.com/docker/machine/libmachine"
@@ -188,6 +189,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	clusterUp(&config)
+	setMinishiftContext(&config, host)
 }
 
 // Set Docker Proxy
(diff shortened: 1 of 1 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -188,6 +189,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	clusterUp(&config)
+	setMinishiftContext(&config, host)
```
praveenkumar: We do need to run this after cluster up because then only we have access to generated kube config file.

cmd/minishift/cmd/start.go
```diff
@@ -345,17 +347,7 @@ func initSubscriptionManagerFlags() {
...
-		OpenShiftVersion:  config.OpenShiftVersion,
-		MinishiftCacheDir: filepath.Join(constants.Minipath, "cache"),
-	}
-	err := oc.EnsureIsCached()
-	if err != nil {
-		glog.Errorln("Error starting the cluster: ", err)
-		atexit.Exit(1)
-	}
-
-	cmdName := filepath.Join(oc.GetCacheFilepath(), constants.OC_BINARY_NAME)
```
praveenkumar: Created a function so that consumed by context setting and cluster up both.

cmd/minishift/cmd/start.go
```diff
@@ -380,14 +372,56 @@ func clusterUp(config *cluster.MachineConfig) {
...
+
+	cmdName := filepath.Join(oc.GetCacheFilepath(), constants.OC_BINARY_NAME)
+	return cmdName
+}
+
+// set Minishift current-context as "minishift"
+// https://docs.openshift.com/enterprise/3.0/cli_reference/manage_cli_profiles.html (set-context section)
+func setMinishiftContext(config *cluster.MachineConfig, host *host.Host) {
+	cmdName := ocCommand(config)
+	ip, _ := host.Driver.GetIP()
```
praveenkumar: @budhrg  In this stage VM is already up and running and surely have IP (which used for provision)  if we required it in before provision step then I agree with you to handle this error.

praveenkumar:
Will close this now and open new one which use latest enabled for oc binary path.

Issue #465, opened 2017-02-22
Move OpenShift related commands into the 'openshift' context


hferentschik:
The introduction of the 'openshift' command (view, set, restart), it might make sense to move some other commands under there as well while we are still in beta state. I am thinking of

* 'service'
* ' get-openshift-versions' (renaming it to 'versions')
* 'logs' - we could keep two logs. One in root for the VM system logs and one for the OpenShift logs under 'openshift'

I am after something like this:
    
    $ minishift openshift -h
    Usage:
      minishift openshift SUBCOMMAND [flags]
      minishift openshift [command]
    
    Available Commands:
      config      Displays or patches OpenShift configuration.
      restart     Restarts the OpenShift cluster.
      versions  Displays Minishift compatible OpenShift versions
      service    Displays info about the specified service (route)
      logs         Displays the OpenShift logs

jorgemoralespou:
@hferentschik I think this might end being too verbose. What commands would not require openshift in the middle? just start/stop/delete? on the other hand, I keep looking into alignment with upstream, so minikube users has the easiest entry path into openshift.

hferentschik:
>  I think this might end being too verbose. 

right, verbosity is an issue here. On the other hand one gains clarity and consistency. Especially since we most likely add an `oc` context as well. logs is for sure discussable, but at least versions and service should go into openshift. versions is specific to Minishift anyways. service also differs in the sense that we need to change the implementation of this command.

> What commands would not require openshift in the middle? just start/stop/delete

config, console, start, stop, delete, ip, docker-env, ssh, version

Anything which affects more the VM and Docker. Also as mentioned above oc related things, like the discussed 'oc env'. 'console' might arguably go under 'openshift' as well.

>  I keep looking into alignment with upstream, so minikube users has the easiest entry path into openshift

Sure, but alignment is only possible to a certain degree. After all our main goal is to make the usage of OpenShift easier. That partly conflicts with Minikube which has a different goal. Though similar, I would always expect there to be a difference in the concrete commands offered. In fact creating this openshift context could even add clarity. Commands which live outside this context can be considered shared between Qinikube and Minishift. They affect the basic life cycle commands for the VM + OpenShift/Kubernetes. The openshift context then contains commands specific to OpenShift. I actually see this as an advantage.

jorgemoralespou:
As long as there is no commands in minikube that we move to the openshift context, you have a +1 from me.

hferentschik:
Fair enough :-)

[...]

PR #500, opened 2017-03-02
Issue #465 Move OpenShift related commands into the 'openshift' context

praveenkumar:
We will push one by one all the required commands to openshift context once we have clear picture about logs and service.

### DIFF
--- a/docs/interacting-with-openshift.md
+++ b/docs/interacting-with-openshift.md
@@ -9,6 +9,7 @@ Minishift creates a virtual machine (VM) and provisions a local, single-node Ope
   - [Login](#login)
   - [Console](#console)
   - [Services](#services)
+  - [Logs](#logs)
 - [Updating OpenShift configuration](#updating-openshift-configuration)
 - [Persistent volumes](#persistent-volumes)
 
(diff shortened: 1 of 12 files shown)


praveenkumar:
> We need to adjust the help text for OpenShiftConfigCmd in openshift.go.

Done, if you think we should have a different help string let me know.

> One could consider moving the console command. That really does not exist in Kube, but belongs to OpenShift. WDYT?

Ah, minikube changed it to `dashboard` now should we also change or leave it as it is, will create a issue as per your understanding.

> Did you check whether the docs refer to any of the moved commands?

I rebased to master today and put changes to doc side which recently re-structured by @Preeticp

docs/interacting-with-openshift.md
```diff
@@ -83,7 +84,16 @@ $ minishift console
...
 ```shell
-$ minishift service [-n NAMESPACE] [--url] NAME
+$ minishift openshift service [-n NAMESPACE] [--url] NAME
+```
+
+<a name="logs"></a>
+### Logs
+
+To access running OpenShift container logs, run this command in a shell after starting Minishift to get the logs:
+
```
Preeticp: To access running OpenShift container logs, run this command in a shell after starting Minishift to get the logs: > To access logs for a running OpenShift container, use this command in a shell after starting Minishift:

hferentschik:
Rebased and merged. Thanks.

Issue #502, opened 2017-03-02
Travis CI builds keep hitting GitHub rate limit for unit tests


hferentschik:
Despite having set _MINISHIFT_GITHUB_API_TOKEN_ as an environment variable in the Travis CI settings,  builds still regularly hit the GitHub API rate limit during unit tests.

It seems the environment variable is not set/available for the process running the unit tests.

The reason might be, that Travis CI only sets the variable for the main build process. Make in turn starts a sub process for each line of a recipe. It does in this case not pass on all environment variables, but only the one set in the Makefile (https://www.gnu.org/software/make/manual/make.html#Environment). For that reason the process running the unit tests would not have _MINISHIFT_GITHUB_API_TOKEN_ set and hence uses the GitHub API unauthenticated.

hferentschik:
Actually all wrong ;-) Sub process do get the env variable set as well. In fact all is working fine on master. 

The issue is that pull requests from forked repositories don't get the environment variable applied out of security reasons - https://docs.travis-ci.com/user/pull-requests#Pull-Requests-and-Security-Restrictions. This means pull request builds won't have the GitHub token variable injected. 

The issue will arise if there are many pull requests being build in a short timeframe. Then some of them can hit the rate limit.

If we wanted to work around this, we would need to put something into our code which skips the affected unit tests when building Travis CI pull requests. Not sure whether we want to go down this route.

budhrg:
I also see following in their security reason page - 

> Travis CI makes encrypted variables and data available only to pull requests coming from the same repository. These are considered trustworthy, as only members with write access to the repository can send them.

I think they should at least allow Pull requests from forked repositories that belongs to members with write access to the repository.

Can we query them about it?

hferentschik:
Merged via pull request #505

PR #505, opened 2017-03-02
Issue #502 Skipping test accessing GitHub API if no token is defined

hferentschik:
- Also changing to verbose output for unit tests to make sure that skipping test becomes obvious

If we were to address the re-occuring Travis failures, this would be probably how we would have to do it. Not convinced this is the best strategy. Pros and cons really. Thoughts?


Makefile
```diff
@@ -84,7 +84,7 @@ clean:
 
 .PHONY: test
 test: vendor $(GOPATH)/src/$(ORG)
-	@go test $(shell $(PACKAGES))
+	@go test -v $(shell $(PACKAGES))
```
praveenkumar: :+1:

pkg/minikube/update/update_test.go
```diff
@@ -1,183 +0,0 @@
-/*
```
praveenkumar: We are removing this because all test code is actually commented out and we are not using update functionality right now.

praveenkumar:
> If we were to address the re-occuring Travis failures, this would be probably how we would have to do it. Not convinced this is the best strategy. Pros and cons really. Thoughts?

I think this is best way of handing it if we are using travis to run linux specific test.

pkg/minikube/openshiftversions/openshift_versions_test.go
```diff
@@ -153,3 +156,10 @@ func TestPrintOpenShiftVersions(t *testing.T) {
 			2, outputBuffer.String()) //TODO(aprindle) change the 2
 	}
 }
+
+func EnsureGitHubApiAccessTokenSet(t *testing.T) {
+	if githubutils.GetGitHubApiToken() == "" {
+		t.Skip("Skipping GitHub API based test, because no access token is defined in the environment.\n " +
+			"To run this test check https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/ and set for example MINISHIFT_GITHUB_API_TOKEN (see github.go).")
+	}
```
LalatenduMohanty: Good Idea.

pkg/minikube/update/update_test.go
LalatenduMohanty: @hferentschik May be we should keep the file for now and then when we decide what we want to do with update it will be automatically taken care of.

pkg/minikube/openshiftversions/openshift_versions_test.go
budhrg: :+1:   Also, it is not affecting other CI's like centos

pkg/minikube/update/update_test.go
budhrg: I am in favor of keeping dead code so, +1 removing it.

Makefile
budhrg: :+1: , I had in my mind from long time.

budhrg:
> I think this is best way of handing it if we are using travis to run linux specific test.

@praveenkumar @hferentschik , We have CentOS CI too which is linux platform too.

pkg/minikube/update/update_test.go
praveenkumar: Yes, having unused code doesn't make sense and we can add again once we have functionality with us.

hferentschik: No sense to keep this. In fact it was even disabled at the time of fork. If we go to write tests we most likely start from scratch anyways.

hferentschik:
> I think this is best way of handing it if we are using travis to run linux specific test.

I think it is a good compromise. I enabled the verbose flag for the test output, so the skip message is right in your face. I expect all of us to run with a token anyways locally. 

This also won't affect other CI systems where the token is set for all pull requests. We will catch potential problems there. The only time we get false positives is when someone touches code around the GitHub API usage. Potentially this could break some unit tests, but would not be caught on the Travis CI pull request build. However, as said, in this case we should capture the issue on one of the other CI systems.

hferentschik:
I take it we agree to merge this?

praveenkumar:
> I take it we agree to merge this?

yes please :+1:

Issue #501, opened 2017-03-02
Allow Minishift to be easily configured to use `nip.io` instead of `xip.io`


jstrachan:
I find `xip.io` to be pretty flaky really - it often goes dark on me for a few minutes. I've found `nip.io` to be much more stable. 

Whats the simplest way to get minishift to use it for route postfixes?

jstrachan:
we maybe wanna make nip.io the default anyways?

hferentschik:
> Whats the simplest way to get minishift to use it for route postfixes?

    $ minishift openshift config set --patch '{"routingConfig": {"subdomain": "192.168.99.101.nip.io"}}'

The IP differs. That's the IP of your VM. You can get it via `minishift ip`. 

There is also a `--routing-suffix` flag as part of the `minishift start` command, which would set the specified prefix as part of the _cluster up_. However, since IP are dynamic, you would not know beforehand what IP you get. This relates to issue #41 

> we maybe wanna make nip.io the default anyways?

Do you run some actual tests that nip.io is more stable?

jorgemoralespou:
FYI,
oc-cluster has moved from xip.io to nip.io in master code. Not sure what
version will bring it, but I doubt it will be 1.5.

cc/ @csrwng

hferentschik:
> oc-cluster has moved from xip.io to nip.io in master code

Interesting, I guess they did some tests then. 

Not sure whether we should put something in the Minishift right now or whether we just wait that oc-cluster uses nip.io

jorgemoralespou:
A correction, the PR is not yet merged.

https://github.com/openshift/origin/pull/13023

I would think that we should want to set nip.io as default if no
--routing-suffix has been set, so would make the transition to this
easier/smoother for users.

https://github.com/openshift-evangelists/oc-cluster-wrapper/blob/master/oc-cluster#L51

hferentschik:
> I would think that we should want to set nip.io as default if no --routing-suffix has been set, so would make the transition to this easier/smoother for users.

Fair enough. Easy enough change to make on our side.

jstrachan:
@hferentschik there's the catch 22 as you don't know the IP address until you run `minishift start`. Is there a way to pass in the `route-suffix-suffix`? ;) so that `route-suffix` gets set to use the dynamically created IP + the `route-suffix-suffix` as part of the regular `minishift start` CLI?

LalatenduMohanty:
[...]

PR #506, opened 2017-03-02
Issue #501 Making nip.io the default routing suffix

hferentschik:
Fix for issue #501.


hferentschik:
@LalatenduMohanty Added some documentation as well now.

cmd/minishift/cmd/start_test.go
```diff
@@ -169,12 +173,13 @@ func TestStartClusterUpWithOverrideHostDataDirFlag(t *testing.T) {
 	defer viper.Reset()
 
 	viper.Set("host-data-dir", "/var/tmp/foo")
-	clusterUp(&testMachineConfig)
+	clusterUp(&testMachineConfig, "192.168.99.42")
```
LalatenduMohanty: We should be using the `testIp` instead of the hardcoded Ip. looks like a typo.

cmd/minishift/cmd/start_test.go
```diff
@@ -187,12 +192,13 @@ func TestStartClusterUpWithOverrideHostVolumesDirFlag(t *testing.T) {
 	defer viper.Reset()
 
 	viper.Set("host-volumes-dir", "/var/tmp/foo")
-	clusterUp(&testMachineConfig)
+	clusterUp(&testMachineConfig, "192.168.99.42")
```
LalatenduMohanty: same as above.

cmd/minishift/cmd/start_test.go
```diff
@@ -206,13 +212,14 @@ func TestStartClusterUpWithFlag(t *testing.T) {
 
 	viper.Set("public-hostname", "foobar")
 	viper.Set("skip-registry-check", "true")
-	clusterUp(&testMachineConfig)
+	clusterUp(&testMachineConfig, "192.168.99.42")
```
LalatenduMohanty: same as above

cmd/minishift/cmd/start_test.go
LalatenduMohanty: There are similar use of the IP string and not the `testIp` variable in below test cases also.

hferentschik: fixed, thanks

docs/interacting-with-openshift.md
```diff
@@ -99,33 +99,48 @@ $ minishift logs
...
-  ```
-  $  minishift openshift config set --patch '{"corsAllowedOrigins": [".*"]}'
-  ```
+Per default, the master configuration is targeted, but you can also patch the node configuration
+by specifying the `target` flag. For more details about the
+`set` command refer to its [synopsis](./minishift_openshift_config_set.md).
+
+A second use case for the `openshift config` command is the ability to change OpenShift's routing suffix.
+If you use a static routing suffix, you can just specify the `routing-suffix` flag as part of the
+[`start`](./minishift_start.md) command. However, per default Minishift uses a dynamic routing prefix
```
Preeticp: Since by default Minishift uses dynamic routing prefix, maybe we can skip the information about static routing suffix on line 129.

docs/interacting-with-openshift.md
```diff
@@ -99,33 +99,48 @@ $ minishift logs
...
+based on [nip.io](http://nip.io/). In this case the VM's IP is part of the routing suffix, for example
+_192.168.99.103.nip.io_. There is an alternative to nip.io called [xip.io](http://xip.io/). It is
+based on the same principles. In case you are experiencing issues with nip.io, you can switch to
+xip.io using the `openshift config` command:
+
```
Preeticp: Maybe we could move this sentence after the command on line 137: 'There is an alternative to nip.io called [xip.io](http://xip.io/)...the `openshift config` command' currently the flow is broken for the nip.io example.

docs/interacting-with-openshift.md
hferentschik: This is important, right? This is explaining the difference and in case of a static routing suffix you actually don't need to patch the config.

docs/interacting-with-openshift.md
hferentschik: Why is the flow broken? You are switching from nip.io to xip.io. Somewhere you need to explain what xip.io is, right?

docs/interacting-with-openshift.md
Preeticp: Sure the information may be important but based on the above lines it felt that for Minishift we do not encounter a static routing suffix situation, so as such is the information relevant?

docs/interacting-with-openshift.md
Preeticp: Sorry, I was expecting a command for nip.io on the lines of this: +$ minishift openshift config set --patch '{"routingConfig": {"subdomain": "192.168.99.103.xip.io"}}' on line 132.
Misread this, please ignore.

Preeticp: Also in this issue we are talking about making nip.io default, so I would expect a command for nip.io and not xip.io, no?

docs/interacting-with-openshift.md
Preeticp: Ok @hferentschik got the context, LGTM

docs/interacting-with-openshift.md
```diff
@@ -99,33 +99,48 @@ $ minishift logs
...
+```
+$  minishift openshift config set --patch '{"corsAllowedOrigins": [".*"]}'
+```
 
-  ```
-  $  minishift openshift config set --patch '{"corsAllowedOrigins": [".*"]}'
-  ```
+Per default, the master configuration is targeted, but you can also patch the node configuration
+by specifying the `target` flag. For more details about the
+`set` command refer to its [synopsis](./minishift_openshift_config_set.md).
```
thatdocslady: Same comment as previous occurance (change to "For details about.., see the <link> command reference."

docs/interacting-with-openshift.md
```diff
@@ -99,33 +99,48 @@ $ minishift logs
...
+
+A second use case for the `openshift config` command is the ability to change OpenShift's routing suffix.
```
thatdocslady: "Another use-case example.... is changing the OpenShift routing suffix."

docs/interacting-with-openshift.md
```diff
@@ -99,33 +99,48 @@ $ minishift logs
...
+If you use a static routing suffix, you can just specify the `routing-suffix` flag as part of the
```
thatdocslady: * remove "just"
* as part > as a part

docs/interacting-with-openshift.md
thatdocslady: per default > by default

docs/interacting-with-openshift.md
```diff
@@ -99,33 +99,48 @@ $ minishift logs
...
+[`start`](./minishift_start.md) command. However, per default Minishift uses a dynamic routing prefix
+based on [nip.io](http://nip.io/). In this case the VM's IP is part of the routing suffix, for example
```
thatdocslady: * "In this case, the VM IP is a part of..."
* Full stop and new sentence before "For example"

docs/interacting-with-openshift.md
```diff
@@ -99,33 +99,48 @@ $ minishift logs
...
+
+```
+$ minishift openshift config set --patch '{"routingConfig": {"subdomain": "192.168.99.103.xip.io"}}'
+```
 
-  Per default, the master configuration is targeted, but you can also path the node config
-  by specifying the `target` flag. For more details about the
-  `set` command refer to its [synopsis](./minishift_openshift_config_set.md).
+You need to replace the IP in the above command with the IP of your VM, which you
+can obtain via the [`ip`](./minishift_ip.md) command.
```
thatdocslady: via > by running the (don't use "via" please)

docs/interacting-with-openshift.md
```diff
@@ -99,33 +99,48 @@ $ minishift logs
...
 
-  **Note:** OpenShift will be restarted after applying the patch.
+**Note:** OpenShift will be transparently restarted after applying the patch.
```
thatdocslady: "After you apply the patch, OpenShift will transparently restart."

PR #519, opened 2017-03-04
[WIP] Issue #487 Introduce minishift-data.json to persist oc path

### LINKED ISSUE #487: Create minishift-data.json to persists data specific to minishift instance

budhrg:
Fix #487 

@hferentschik , @gbraad , @praveenkumar, @LalatenduMohanty  Opening PR for initial review. 

Working on its tests now.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -30,6 +30,7 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/cache"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
+	"github.com/minishift/minishift/pkg/minishift/minishiftdata"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
 	minishiftUtil "github.com/minishift/minishift/pkg/minishift/util"
 	"github.com/minishift/minishift/pkg/util"
@@ -80,6 +81,7 @@ var (
 	registryMirror   []string
 	openShiftEnv     []string
 	shellProxyEnv    string
+	machinesDirPath  string
 )
 
 // startCmd represents the start command
@@ -125,6 +127,8 @@ func SetRunner(newRunner util.Runner) {
 func runStart(cmd *cobra.Command, args []string) {
 	libMachineClient := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer libMachineClient.Close()
+
+	machinesDirPath = libMachineClient.GetMachinesDir()
 
 	validateOpenshiftVersion()
 
@@ -179,7 +183,18 @@ func runStart(cmd *cobra.Command, args []string) {
 		atexit.Exit(1)
 	}
 
-	clusterUp(&config)
+	oc, err := clusterUp(&config)
+	if err != nil {
+		glog.Errorln("Error starting the cluster: ", err)
+		atexit.Exit(1)
+	}
+
+	// Update minishift-data.json for oc path
+	minishiftData, err := minishiftdata.Create(machinesDirPath)
+	minishiftData.OcPath = filepath.Join(oc.GetCacheFilepath(), constants.OC_BINARY_NAME)
+	if err = minishiftdata.WriteMinishiftData(minishiftData); err != nil {
+		glog.Warningf("Error updating oc path in minishift-data.json file: ", err)
+	}
 }
 
 // Set Docker Proxy
@@ -302,13 +317,12 @@ func initSubscriptionManagerFlags() {
 }
 
 // clusterUp downloads and installs the oc binary in order to run 'cluster up'
-func clusterUp(config *cluster.MachineConfig) {
-	oc := cache.Oc{
+func clusterUp(config *cluster.MachineConfig) (*cache.Oc, error) {
+	oc := &cache.Oc{
 		OpenShiftVersion:  config.OpenShiftVersion,
 		MinishiftCacheDir: filepath.Join(constants.Minipath, "cache"),
 	}
-	err := oc.EnsureIsCached()
-	if err != nil {
+	if err := oc.EnsureIsCached(); err != nil {
 		glog.Errorln("Error starting the cluster: ", err)
 		atexit.Exit(1)
 	}
@@ -338,12 +352,13 @@ func clusterUp(config *cluster.MachineConfig) {
 		}
 	})
 
-	err = runner.Run(cmdName, cmdArgs...)
+	err := runner.Run(cmdName, cmdArgs...)
 	if err != nil {
 		// TODO glog is probably not right here. Need some sort of logging wrapper
 		glog.Errorln("Error starting the cluster: ", err)
 		atexit.Exit(1)
 	}
+	return oc, nil
 }
 
 func validateOpenshiftVersion() {
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -38,6 +38,7 @@ import (
 	"github.com/docker/machine/libmachine/state"
 	"github.com/golang/glog"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/minishiftdata"
 	"github.com/minishift/minishift/pkg/minishift/registration"
 	minishiftUtil "github.com/minishift/minishift/pkg/minishift/util"
 	"github.com/minishift/minishift/pkg/util"
@@ -218,7 +219,7 @@ func createVirtualboxHost(config MachineConfig) drivers.Driver {
 func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 	fmt.Println(fmt.Sprintf("Downloading ISO '%s'", m.MinikubeISO))
 
-	// store the miniube-iso inside the .minikube dir
+	// store the iso inside the .minishift/cache/iso dir
 	response, err := http.Get(m.MinikubeISO)
 	if err != nil {
 		return err
@@ -253,8 +254,6 @@ func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 }
 
 func (m *MachineConfig) ShouldCacheMinikubeISO() bool {
-	// store the miniube-iso inside the .minikube dir
-
 	urlObj, err := url.Parse(m.MinikubeISO)
 	if err != nil {
 		return false
@@ -338,6 +337,12 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 
 	if err := api.Save(h); err != nil {
 		return nil, fmt.Errorf("Error attempting to save store: %s", err)
+	}
+
+	// Create minishift-data.json
+	_, err = minishiftdata.Create(api.GetMachinesDir())
+	if err != nil {
+		return nil, fmt.Errorf("Error creating minishift-data.json: %s", err)
 	}
 
 	if config.ShellProxyEnv != "" {
--- a/pkg/minishift/minishiftdata/minishiftdata.go
+++ b/pkg/minishift/minishiftdata/minishiftdata.go
@@ -1,0 +1,76 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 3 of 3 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -179,7 +183,18 @@ func runStart(cmd *cobra.Command, args []string) {
...
 
-	clusterUp(&config)
+	oc, err := clusterUp(&config)
+	if err != nil {
+		glog.Errorln("Error starting the cluster: ", err)
+		atexit.Exit(1)
+	}
+
+	// Update minishift-data.json for oc path
+	minishiftData, err := minishiftdata.Create(machinesDirPath)
```
hferentschik: We want to probably create the initial file in root.go, together with the other data files like _config.json_.

cmd/minishift/cmd/start.go
```diff
@@ -302,13 +317,12 @@ func initSubscriptionManagerFlags() {
 }
 
 // clusterUp downloads and installs the oc binary in order to run 'cluster up'
-func clusterUp(config *cluster.MachineConfig) {
-	oc := cache.Oc{
+func clusterUp(config *cluster.MachineConfig) (*cache.Oc, error) {
```
hferentschik: -1, why would _clusterUp_ return _oc_? It would be much better to create _Oc_ outside this method and pass it into _clusterUp_.

pkg/minikube/cluster/cluster.go
```diff
@@ -218,7 +219,7 @@ func createVirtualboxHost(config MachineConfig) drivers.Driver {
 func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 	fmt.Println(fmt.Sprintf("Downloading ISO '%s'", m.MinikubeISO))
 
-	// store the miniube-iso inside the .minikube dir
+	// store the iso inside the .minishift/cache/iso dir
```
hferentschik: actually it is $MINISHIFT_HOME

pkg/minikube/cluster/cluster.go
```diff
@@ -340,6 +339,12 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 		return nil, fmt.Errorf("Error attempting to save store: %s", err)
 	}
 
+	// Create minishift-data.json
+	_, err = minishiftdata.Create(api.GetMachinesDir())
```
hferentschik: !? Create again? You had a create in start.go as well, right - https://github.com/minishift/minishift/pull/519/files#r104312417

pkg/minishift/minishiftdata/minishiftdata.go
```diff
@@ -0,0 +1,76 @@
...
+	"os"
+	"path/filepath"
+)
+
+var (
+	minishiftDataPath string
+)
+
+type MinishiftData struct {
+	OcPath string `json:"OcPath"`
```
hferentschik: What's \`json:"OcPath"\` good for?

pkg/minishift/minishiftdata/minishiftdata.go
```diff
@@ -0,0 +1,76 @@
...
+			return nil, err
+		}
+
+		return data, nil
+	}
+
+	return ReadMinishiftData()
+}
+
+func WriteMinishiftData(data *MinishiftData) error {
```
hferentschik: Why would I pass the struct to write. _Write_ should be a method of the struct itself. It knows which data to persist, right?

pkg/minishift/minishiftdata/minishiftdata.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	if err = ioutil.WriteFile(minishiftDataPath, jsonData, 0644); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func ReadMinishiftData() (*MinishiftData, error) {
```
hferentschik: This should probably also be a function of the _MinishiftData_ struct itself. We have a tendency here for no encapsulation at all. Golang might not be fully fledged OO language, but one can encapsulate data and methods nevertheless.

cmd/minishift/cmd/start.go
hferentschik: I guess the benefit here is that you can get the _ machinesDirPath_. On the other hand we know this.

pkg/minishift/minishiftdata/minishiftdata.go
hferentschik: It is a public method, but to me it is not really clear how this would be used. It depends on _minishiftDataPath_ being set, right? What ensures that this is the case?

pkg/minikube/cluster/cluster.go
hferentschik: And why here?

cmd/minishift/cmd/start.go
hferentschik: How would one get hold of _MinishiftData_ at another point where we don't know 'machinesDirPath'?

pkg/minishift/minishiftdata/minishiftdata.go
```diff
@@ -0,0 +1,76 @@
...
+	OcPath string `json:"OcPath"`
+}
+
+// Create minishift-data.json with empty data or return with data if exists
+func Create(machinesDirPath string) (*MinishiftData, error) {
+	data := &MinishiftData{
+		OcPath: "",
+	}
+
+	minishiftDataPath = filepath.Join(machinesDirPath, constants.MachineName, "minishift-data.json")
```
hferentschik: I think it is better to pass the full directory, or maybe even the full path to the file. Using _constants.MachineName_  is exactly what we need to avoid. This is what we need to change when trying to support multiple machines. Everywhere we we just use _constants.MachineName_ we are creating us a problem for later.

gbraad: @hferentschik related, but not for this change, how about the configuration of hostfolders? Are they based on a host/machine-id/machine-name or globally... ? I would see this defined on the machine level... WDYT?

gbraad:
I agree with @hferentschik on a global object (singleton accessor or whatever) that can be used for configuration reading/writing of any kind. The current implementation would have to rely on viper.Get() which will result in conflicts when config information is/needs to be machine specific

pkg/minishift/minishiftdata/minishiftdata.go
```diff
@@ -0,0 +1,76 @@
...
+var (
+	minishiftDataPath string
+)
+
+type MinishiftData struct {
+	OcPath string `json:"OcPath"`
+}
+
+// Create minishift-data.json with empty data or return with data if exists
+func Create(machinesDirPath string) (*MinishiftData, error) {
```
gbraad: I think the naming is confusing;
  * Create an object?
  * Create the file? 

The function name and the package path itself do not do a good job at conveyong the intention if the functionality.

`minishiftdata` sounds like all the datafiles that are handled by Minishift... or is it metadata? WDYT?

cmd/minishift/cmd/start.go
```diff
@@ -179,7 +183,18 @@ func runStart(cmd *cobra.Command, args []string) {
...
+	oc, err := clusterUp(&config)
+	if err != nil {
+		glog.Errorln("Error starting the cluster: ", err)
+		atexit.Exit(1)
+	}
+
+	// Update minishift-data.json for oc path
+	minishiftData, err := minishiftdata.Create(machinesDirPath)
+	minishiftData.OcPath = filepath.Join(oc.GetCacheFilepath(), constants.OC_BINARY_NAME)
+	if err = minishiftdata.WriteMinishiftData(minishiftData); err != nil {
```
gbraad: This does not encapsulate data and logic... makes it hard to actually use the functionality.

pkg/minikube/cluster/cluster.go
```diff
@@ -340,6 +339,12 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("Error creating minishift-data.json: %s", err)
```
gbraad: Would like to see Create handle the actual error message creation. This should only handle the error and/or pass on the error as-is.

pkg/minishift/minishiftdata/minishiftdata.go
```diff
@@ -0,0 +1,76 @@
...
+	data := &MinishiftData{
+		OcPath: "",
+	}
+
+	minishiftDataPath = filepath.Join(machinesDirPath, constants.MachineName, "minishift-data.json")
+	if _, err := os.Stat(minishiftDataPath); os.IsNotExist(err) {
+		if err := WriteMinishiftData(data); err != nil {
+			return nil, err
```
gbraad: As mentioned above, this would need to return and error with "Creation of %s failed: %s", filename, err.Error() or similar.

pkg/minishift/minishiftdata/minishiftdata.go
```diff
@@ -0,0 +1,76 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package minishiftdata
+
+import (
+	"encoding/json"
```
gbraad: glad we are not using Viper here... 👍

pkg/minishift/minishiftdata/minishiftdata.go
gbraad: Isn't this *InstanceData* or ... I do not like the use of Data here, as it confuses to what the actual use it.

pkg/minishift/minishiftdata/minishiftdata.go
gbraad: I see this pattern used often in other pieces of Go code,. It would be acceptable for me as long as this functionality is not exposed as a public function.

cmd/minishift/cmd/start.go
hferentschik: root.go actually already creates the machines directory. So imo, the best place to create the instance is there, making sure it gets initialised in the persistent pre-run.

cmd/minishift/cmd/start.go
budhrg: @hferentschik , yes I am thinking of that too. `clusterUp` will just prepare the command and run it.

pkg/minishift/minishiftdata/minishiftdata.go
praveenkumar: @gbraad I agree, but I think what @hferentschik suggesting is better way to do it and we should make it struch method instead of function.

pkg/minishift/minishiftdata/minishiftdata.go
budhrg: @hferentschik  
> Everywhere we we just use constants.MachineName we are creating us a problem for later.

Agree :+1:

pkg/minishift/minishiftdata/minishiftdata.go
budhrg: Yes, forgot to remove `json:"OcPath"` once decided to go with `OcPath` for json key.

budhrg:
Thanks @hferentschik @gbraad @praveenkumar for initial feedback. 

Will work on it today.

budhrg:
I am closing this PR since its purpose of initial feedback is served. Thanks to all reviewers.

cmd/minishift/cmd/start.go
budhrg: > We want to probably create the initial file in root.go, together with the other data files like config.json.

@hferentschik No, we don't get `machinesDirPath` here since for that it has to be instance of libmachineClient which happens only in start.go 

> root.go actually already creates the machines directory. So imo, the best place to create the instance is there, making sure it gets initialised in the persistent pre-run.

Reason as above.

budhrg: > I guess the benefit here is that you can get the _ machinesDirPath_. On the other hand we know this.

Or @hferentschik as part of this comment, you want to build it as :
```
machinesDirPath := filepath.Join(constants.Minipath, "machines")   
``` 
?

PR #525, opened 2017-03-07
Issue #523 doc review for updating openshift config

### LINKED ISSUE #523: Review OpenShift configuration documentation
After merging PR #506 new doc review comments are discovered, I will make the changes and submit a new PR.

thatdocslady:
Fixes #523 

This PR includes doc review that didn't make into PR #506 before it was merged.

### DIFF
--- a/docs/interacting-with-openshift.md
+++ b/docs/interacting-with-openshift.md
@@ -11,6 +11,8 @@ Minishift creates a virtual machine (VM) and provisions a local, single-node Ope
   - [Services](#services)
   - [Logs](#logs)
 - [Updating OpenShift configuration](#updating-openshift-configuration)
+  - [Example: Configuring cross-origin resource sharing](#configuring-cross-origin-resource-sharing)
+  - [Example: Changing the OpenShift routing suffix](#changing-openshift-routing-suffix)
 - [Persistent volumes](#persistent-volumes)
 
 <!-- /MarkdownTOC -->
@@ -99,48 +101,58 @@ $ minishift logs
 <a name="updating-openshift-configuration"></a>
 ## Updating OpenShift configuration
 
-Once OpenShift is running, you can view and change the master and
-node configuration of your OpenShift cluster.
+While OpenShift is running, you can view and change the master or the node configuration of your cluster.
 
-You can view the current OpenShift master configuration (_master-config.yaml_) via:
+To view the current OpenShift master configuration (_master-config.yaml_), run the following command:
 
 ```shell
 $ minishift openshift config view
 ```
 
-For displaying the node configuration, you can specify the `target` flag.
-For more details about the `view` command refer to its [synopsis](./minishift_openshift_config_view.md).
+To show the node configuration instead of the master configuration, specify the `target` flag.
 
-Let's look at [Cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) (CORS) as an example for patching the OpenShift master configuration.
-Per default, OpenShift will only allow cross origin resource requests from the IP of the
-cluster as well as localhost. This is specified via the `corsAllowedOrigins` property in the
-[master configuration](https://docs.openshift.com/enterprise/3.0/admin_guide/master_node_configuration.html#master-configuration-files) (_master-config.yaml_). To change this value and allow
-cross origin requests from all domains, one can execute:
+For details about the `view` command, see the [synopsis](./minishift_openshift_config_view.md) command reference.
+
+**Note:** After you update the OpenShift configuration, OpenShift will transparently restart.
+
+<a name="configuring-cross-origin-resource-sharing"></a>
+### Example: Configuring cross-origin resource sharing
+
+In this example, you configure [cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) (CORS)
+by updating the OpenShift master configuration to allow additional IP addresses to request resources.
+
+By default, OpenShift only allows cross-origin resource requests from the IP address of the
+cluster or from localhost. This setting is stored in the `corsAllowedOrigins` property of the
+[master configuration](https://docs.openshift.com/enterprise/3.0/admin_guide/master_node_configuration.html#master-configuration-files) (_master-config.yaml_).
+
+To change the property value and allow cross-origin requests from all domains,
+run the following command:
 
 ```
 $  minishift openshift config set --patch '{"corsAllowedOrigins": [".*"]}'
 ```
 
-Per default, the master configuration is targeted, but you can also patch the node configuration
-by specifying the `target` flag. For more details about the
-`set` command refer to its [synopsis](./minishift_openshift_config_set.md).
+<a name="changing-openshift-routing-suffix"></a>
+### Example: Changing the OpenShift routing suffix
 
-A second use case for the `openshift config` command is the ability to change OpenShift's routing suffix.
-If you use a static routing suffix, you can just specify the `routing-suffix` flag as part of the
-[`start`](./minishift_start.md) command. However, per default Minishift uses a dynamic routing prefix
-based on [nip.io](http://nip.io/). In this case the VM's IP is part of the routing suffix, for example
-_192.168.99.103.nip.io_. There is an alternative to nip.io called [xip.io](http://xip.io/). It is
-based on the same principles. In case you are experiencing issues with nip.io, you can switch to
-xip.io using the `openshift config` command:
+In this example, you change the OpenShift routing suffix in the master configuration.
+
+If you use a static routing suffix, you can set the `routing-suffix` flag as part of the
+[`start`](./minishift_start.md) command. By default, Minishift uses a dynamic routing prefix
+based on [nip.io](http://nip.io/), in which the IP address of the VM is a part of the routing suffix,
+for example _192.168.99.103.nip.io_.
+
+If you experience issues with `nip.io`, you can use [xip.io](http://xip.io/), which is
+based on the same principles.
+
+To set the routing suffix to `xip.io`, run the following command:
 
 ```
-$ minishift openshift config set --patch '{"routingConfig": {"subdomain": "192.168.99.103.xip.io"}}'
+$ minishift openshift config set --patch '{"routingConfig": {"subdomain": "<IP-ADDRESS>.xip.io"}}'
 ```
 
-You need to replace the IP in the above command with the IP of your VM, which you
-can obtain via the [`ip`](./minishift_ip.md) command.
-
-**Note:** OpenShift will be transparently restarted after applying the patch.
+Make sure to replace the IP address in the above example with the IP address of your Minishift VM.
+You can retrieve the IP address by running the [`ip`](./minishift_ip.md) command.
 
 <a name="persistent-volumes"></a>
 ## Persistent volumes


docs/interacting-with-openshift.md
```diff
@@ -99,48 +99,58 @@ $ minishift logs
...
-cluster as well as localhost. This is specified via the `corsAllowedOrigins` property in the
-[master configuration](https://docs.openshift.com/enterprise/3.0/admin_guide/master_node_configuration.html#master-configuration-files) (_master-config.yaml_). To change this value and allow
-cross origin requests from all domains, one can execute:
+For details about the `view` command, see the [synopsis](./minishift_openshift_config_view.md) command reference.
+
+### Use-case examples
+
+#### Configuring cross-origin resource sharing
+
+In this use-case example, you update the OpenShift master configuration to change the resource IP
```
hferentschik: In this use-case example, you update the OpenShift master configuration to change the IP addresses from which cross-origin requests are allowed (see [cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) (CORS)).

"change the resource IP" is imo not correct.

docs/interacting-with-openshift.md
```diff
@@ -99,48 +99,58 @@ $ minishift logs
...
+address in [cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) (CORS).
+
+By default, OpenShift only allows cross-origin resource requests from the IP address of the
+cluster or from localhost. You specify the IP address in the `corsAllowedOrigins` property of the
```
hferentschik: You specify the allowed addresses in the `corsAllowedOrigins` property

docs/interacting-with-openshift.md
hferentschik: I think one can drop "example"

docs/interacting-with-openshift.md
```diff
@@ -99,48 +99,58 @@ $ minishift logs
...
 ```
 $  minishift openshift config set --patch '{"corsAllowedOrigins": [".*"]}'
 ```
 
-Per default, the master configuration is targeted, but you can also patch the node configuration
-by specifying the `target` flag. For more details about the
-`set` command refer to its [synopsis](./minishift_openshift_config_set.md).
+#### Changing the OpenShift routing prefix
+
+In this use-case example, you change the OpenShift routing suffix in the master configuration.
```
hferentschik: I think one can drop "example"

docs/interacting-with-openshift.md
```diff
@@ -99,48 +99,58 @@ $ minishift logs
...
+
+If you use a static routing suffix, you can set the `routing-suffix` flag as part of the
+[`start`](./minishift_start.md) command. However, by default Minishift uses a dynamic routing prefix
+based on [nip.io](http://nip.io/).
+
+When the IP address of the VM is apart of the routing suffix, for example
```
hferentschik: "When the IP address of the VM is apart of the routing suffix" - what does that even mean? This makes no sense.

hferentschik: The causal chain is also broken here. "When the IP address of the VM is apart of the routing suffix" => "you can use xip.io instead of nip.io". That't not correct. xip.io and nip.io are equivalent solutions.

docs/interacting-with-openshift.md
```diff
@@ -99,48 +99,58 @@ $ minishift logs
...
+To set the routing suffix to `xip.io`, run the following command:
 
 ```
-$ minishift openshift config set --patch '{"routingConfig": {"subdomain": "192.168.99.103.xip.io"}}'
+$ minishift openshift config set --patch '{"routingConfig": {"subdomain": "<IP-ADDRESS>.xip.io"}}'
 ```
 
-You need to replace the IP in the above command with the IP of your VM, which you
-can obtain via the [`ip`](./minishift_ip.md) command.
+Make sure to replace the IP address in the above example with the IP address of your Minishift VM.
```
hferentschik: You are using IP_ADDRESS in the example command, but IP below.

docs/interacting-with-openshift.md
```diff
@@ -99,48 +99,58 @@ $ minishift logs
...
+is based on the same principles and can be used if you experience issues with `nip.io`.
 
-A second use case for the `openshift config` command is the ability to change OpenShift's routing suffix.
-If you use a static routing suffix, you can just specify the `routing-suffix` flag as part of the
-[`start`](./minishift_start.md) command. However, per default Minishift uses a dynamic routing prefix
-based on [nip.io](http://nip.io/). In this case the VM's IP is part of the routing suffix, for example
-_192.168.99.103.nip.io_. There is an alternative to nip.io called [xip.io](http://xip.io/). It is
-based on the same principles. In case you are experiencing issues with nip.io, you can switch to
-xip.io using the `openshift config` command:
+To set the routing suffix to `xip.io`, run the following command:
```
hferentschik: -1, you are not setting the routing prefix to xip.io, the routing prefix is 'ip + xip.io'

docs/interacting-with-openshift.md
```diff
@@ -99,48 +99,58 @@ $ minishift logs
...
+$ minishift openshift config set --patch '{"routingConfig": {"subdomain": "<IP-ADDRESS>.xip.io"}}'
 ```
 
-You need to replace the IP in the above command with the IP of your VM, which you
-can obtain via the [`ip`](./minishift_ip.md) command.
+Make sure to replace the IP address in the above example with the IP address of your Minishift VM.
+You can retrieve the IP address by running the [`ip`](./minishift_ip.md) command.
 
-**Note:** OpenShift will be transparently restarted after applying the patch.
+After you update the configuration, OpenShift will transparently restart.
```
hferentschik: Why did you remove "Note"? This was a general not which applies to the patch command. It really should be moved up in this case.

docs/interacting-with-openshift.md
```diff
@@ -99,48 +99,58 @@ $ minishift logs
 <a name="updating-openshift-configuration"></a>
 ## Updating OpenShift configuration
 
-Once OpenShift is running, you can view and change the master and
-node configuration of your OpenShift cluster.
+While OpenShift is running, you can view and change the master or the node configuration of your OpenShift cluster.
```
budhrg: Since, we mentioned `OpenShift` at the beginning of sentence, I think we can omit during the sentence like here `configuration of your OpenShift cluster`.
It can be `configuration of your cluster`.

docs/interacting-with-openshift.md
budhrg: Space needed in `apart` :smile:

docs/interacting-with-openshift.md
thatdocslady: This is what the text afterwards is telling us. If it's not correct, please provide what is the correct purpose and I'll change it.

thatdocslady: "use-case" is an adjective and can't live alone, it needs to be qualified. Either say "use-case scenario" or "use-case example", but not just "use-case".

docs/interacting-with-openshift.md
thatdocslady: Ok so this means that the original text wasn't correct, I'll fix it so that it all makes sense now.

docs/interacting-with-openshift.md
thatdocslady: See previous comment

docs/interacting-with-openshift.md
thatdocslady: Please look at the original text and not just at the updated text, you will see that I'm trying to do an English grammar proofread on the text without changing the meaning. If the original text is not accurate then we need to look at it from that angle, but it's hard to implement your feedback like this.

docs/interacting-with-openshift.md
thatdocslady: I removed the note because it's not a "by the way" thing, but expected behavior as a part of the process. You're right though that it should be listed before the examples, because it's a global thing.

docs/interacting-with-openshift.md
hferentschik: I gave you the correct sentence: "In this use-case example, you update the OpenShift master configuration to change the IP addresses from which cross-origin requests are allowed (see cross-origin resource sharing (CORS))."

hferentschik: I never seen use-case used as an adjective. IMO the most common use is as noun - http://www.yourdictionary.com/use-case

docs/interacting-with-openshift.md
hferentschik: > Please look at the original text and not just at the updated text, you will see that I'm trying to do an English grammar proofread on the text without changing the meaning

So where in the original text do I say that the IP address is _apart_ from the routing prefix? In fact I am saying quite the opposite - "In this case the VM's IP is part of the routing suffix". The IP _is part_ of the routing prefix. This is what makes it dynamic, since we are assigning IPs dynamically, so you don't know the exact routing prefix until provisioning is complete. Hence the routing prefix is dynamic or at least has an dynamic part, opposed to something like 'foo.bar.com'

hferentschik: > Space needed in apart 

right, so maybe this is a typo 'a part', but that makes quite a difference here.

hferentschik: Also the logic is still not correct. From "IP part of the routing prefix" does not follow that one can use [xip|nip].io. xip.ip is just a plain substitute/alternative to nip.io.

docs/interacting-with-openshift.md
```diff
@@ -99,48 +99,58 @@ $ minishift logs
...
+To show the node configuration instead of the master configuration, specify the `target` flag.
 
-Let's look at [Cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) (CORS) as an example for patching the OpenShift master configuration.
-Per default, OpenShift will only allow cross origin resource requests from the IP of the
-cluster as well as localhost. This is specified via the `corsAllowedOrigins` property in the
-[master configuration](https://docs.openshift.com/enterprise/3.0/admin_guide/master_node_configuration.html#master-configuration-files) (_master-config.yaml_). To change this value and allow
-cross origin requests from all domains, one can execute:
+For details about the `view` command, see the [synopsis](./minishift_openshift_config_view.md) command reference.
+
+### Use-case examples
```
hferentschik: I think this additional level can be skipped, especially when looking at the resulting table of contents, which needs by the way updating anyways.

docs/interacting-with-openshift.md
thatdocslady: If it's hyphenated then it's an adjective, and in RH docs we generally don't use it non-hyphenated. I'm not going to split hairs with you on this, so I will remove it entirely and just keep "example".

docs/interacting-with-openshift.md
thatdocslady: I only use IP because that's the name of the command (a name that I don't agree with but I have no control over). The best practice is to always spell out "IP address", especially in cases like this one where we're writing about suffixes and prefixes and routing and network elements.

docs/interacting-with-openshift.md
```diff
@@ -99,48 +101,58 @@ $ minishift logs
...
-[master configuration](https://docs.openshift.com/enterprise/3.0/admin_guide/master_node_configuration.html#master-configuration-files) (_master-config.yaml_). To change this value and allow
-cross origin requests from all domains, one can execute:
+For details about the `view` command, see the [synopsis](./minishift_openshift_config_view.md) command reference.
+
+**Note:** After you update the OpenShift configuration, OpenShift will transparently restart.
+
+<a name="configuring-cross-origin-resource-sharing"></a>
+### Example: Configuring cross-origin resource sharing
+
+In this example, you configure [cross-origin resource sharing](https://en.wikipedia.org/wiki/Cross-origin_resource_sharing) (CORS)
```
hferentschik: 👍

docs/interacting-with-openshift.md
```diff
@@ -99,48 +101,58 @@ $ minishift logs
...
-A second use case for the `openshift config` command is the ability to change OpenShift's routing suffix.
-If you use a static routing suffix, you can just specify the `routing-suffix` flag as part of the
-[`start`](./minishift_start.md) command. However, per default Minishift uses a dynamic routing prefix
-based on [nip.io](http://nip.io/). In this case the VM's IP is part of the routing suffix, for example
-_192.168.99.103.nip.io_. There is an alternative to nip.io called [xip.io](http://xip.io/). It is
-based on the same principles. In case you are experiencing issues with nip.io, you can switch to
-xip.io using the `openshift config` command:
+In this example, you change the OpenShift routing suffix in the master configuration.
+
+If you use a static routing suffix, you can set the `routing-suffix` flag as part of the
```
hferentschik: 👍 thanks, now its correct ;-)

hferentschik:
Merged, thanks.

Issue #480, opened 2017-02-24
Auto mounting functionality for shared folders


gbraad:
1. ~~Allow configuration to auto mount shares~~
2. ~~Prevent remounting mounted shares~~
3. ~~Check if host is reachable~~
4. ~~Command to add configuration by name~~
5. ~~Command to remove configuration by name~~
6. ~~Command to mount share by name~~
7. ~~Command to unmount share by name~~
8. ~~Auto mount option for `Users` share~~

gbraad:
Development can be tracked in the branch: https://github.com/gbraad/minishift/tree/mount-shares

  * https://github.com/gbraad/minishift/commit/cf05002d591b980c149e52224568990f475c7fae

Configuration needed:

```
{
    "iso-url": "file://C:/Users/gbraad/.minishift/cache/iso/minishift-cifs.iso",
    "vm-driver": "hyperv",
    "shares": [
        {"name": "Users", "uncpath": "//RHGBWIN10/Users", "username": "gbraad@redhat.com", "password": "********", "domain": "RHGBWIN10"},
        {"name": "Disks", "uncpath": "//10.0.21.33/disks", "username": "root", "password": "********", "domain": ""}
    ]
}
```

gbraad:
@hferentschik @LalatenduMohanty I would not really like to introduce a new verb as command, but I think this is needed? ...

```
$ minishift mount [params]
$ minishift unmount [name]
```
This feels weird...

So, what about?
```
$ minishift share mount [name]
$ minishift share unmount [name]
```

Or place them inside the `config` command?

And how to structure them for edit and save? WDYT?

hferentschik:
>  I would not really like to introduce a new verb as command

Sure. How do you want to use that? Once Minishift is up you do mount/umount? I guess I am trying to figure out how this is tied into the current behavior of say VirtualBox and that it transparently mounts your home directory (same for xhyve with the appropriate flags). 

Will host folder mounts on Windows with Samba shares work differently (in terms off when and how they are established) than with VirtualBox and guest additions? Or moving forward with sshfs?

@gbraad maybe you can outline how current host folder mount behavior is going to be unified with what you are doing now and this new command. 

Generally, I like the idea. I like the idea of having the ability to potentially umount a directory and mount a different one.  What about something like

    $ minishift hostfolder [view|mount|umount]

[...]

Issue #422, opened 2017-02-17
Can't connect to Windows filesharing when Microsoft Account is used


gbraad:
When the user has switched to a Microsoft Account, the user is unable to login to filesharing using the `DESKTOP-..\username`.

gbraad:
Related to #316

The pattern `MicrosoftAccount\gbraad@redhat.com` needs to be used for login

gbraad:
Using a MicrosoftAccount will fail to mount using `mount -t cifs` with:
```
mount error(13): Permission denied
```

gbraad:
A user when using a Microsoft Account needs to specify the username WITHOUT the MicrosoftAccount prefix, but specifying the domain part as the local computername.

eg.
```
$ mount -t cifs //10.0.21.126/users /mnt/sda1/users/ -o credentials=/tmp/credentials
```

`/tmp/credentials`
```
username=gbraad@redhat.com
password=[microsoft account password]
domain=DESKTOP-4GCR9CL
```

gbraad:
Documented in #449

Issue #316, opened 2017-01-23
Implement mounted host folder support using Windows shares


hferentschik:
Following up on the investigation host folder mounts on Windows + Hyper-V (issue #293)

gbraad:
Hopefully this can all be automated. On the host you can create a shared folder using PowerShell:

```
Set-Variable MINISHIFTHOME $HOME\.minishift
Set-Variale MINISHIFTSHARED $MINISHIFTHOME\Shared

if (!(Test-Path $MINISHIFTSHARED )) { 
   New-Item MINISHIFTSHARED -type Directory 
}

New-SMBShare –Name "MinishitShare" –Path MINISHIFTSHARED `
   –ContinuouslyAvailable $True `
   –FullAccess mygroup\administrators  `
   -ChangeAccess "mygroup\authenticated users" `
   -ReadAccess "mygroup\authenticated users"
```

and then inside the VM this share can be accessed and mounted. Note that you might have to set to EVERYONE, but this is not very safe.

hferentschik:
@gbraad, nice. So, you are saying this is completely password free? Anyone can mount this Samba share and read/write to it? I think when I did some manual tests, I mounted the share using the users credentials? Obviously password free would be easier for us, question is whether this is not a bit too open? What does Docker for Windows do?

Also, what's about Windows 7? How do you do it there w/o PowerShell. My understanding is that most/many Windows developers are still on this OS version and we need to cater for them as well.

gbraad:
@hferentschik Windows7 has no Hyper-V option. It is only available from 2012, 8.x and up. For Windows7 you have to use VirtualBox.

The password is actually an issue. You can use "Authenticated Users", but this needs to be an existing user. We can also create a user for use with PowerShell. How will we deal with the password? Generate one with a random value... and then remove the user on `minishift delete`?

There is another issue with the share, how long will it exist? `new-smbshare` has an option: `-ContinuouslyAvailable<Boolean>` [TechNet][technet continuouslyavailable]

[technet continuouslyavailable]: https://technet.microsoft.com/en-us/library/jj635722.aspx

hferentschik:
> Windows7 has no Hyper-V option. 

Of course, I forgot. Windows 7 is VirtualBox only. However, given that we don't know whether we can use guest additions, the share options is a viable solution there as well, right? 

> The password is actually an issue. You can use "Authenticated Users", but this needs to be an existing user. We can also create a user for use with PowerShell. How will we deal with the password? 

[...]

PR #526, opened 2017-03-08
Issue #480 Mount CIFS shares

gbraad:
Related to #480 

~~Sharing for feedback/sync related to sshfs and config
I do read the instance config, but do not write it yet.
I have a type variable, but do not enforce it just yet in code.
This also means, the mount code for cifs is generally named and not moved into a separate file just yet.~~

~~Still a lot of duplication to be removed as a refactor, so do not look at newlines/whitelines, missing boilerplate, docs, but functionality...~~

```
    "HostFolders": [
        {
            "Name": "Users",
            "Type": "cifs",
            "Options": {
                "password": "********",
                "domain": "RHGBWIN10",
                "uncpath": "//RHGBWIN10/Users",
                "username": "gbraad@redhat.com"
            }
        },
        {
            "Name": "unreachable",
            "Type": "cifs",
            "Options": {
                "domain": "",
                "password": "********",
                "uncpath": "//unreachable/disks",
                "username": "root"
            }
        },
        {
            "Name": "Disks",
            "Type": "cifs",
            "Options": {
                "domain": "",
                "password": "********",
                "uncpath": "//10.0.21.33/disks",
                "username": "root"
            }
        }
    ],
```

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -100,6 +100,14 @@ var settings []Setting = []Setting{
 		callbacks:   []setFn{RequiresRestartMsg},
 	},
 	{
+		name: "hostfolders-automount",
+		set:  SetBool,
+	},
+	{
+		name: "hostfolders-mountpath",
+		set:  SetString,
+	},
+	{
 		name: config.WantUpdateNotification,
 		set:  SetBool,
 	},
--- a/cmd/minishift/cmd/hostfolder/hostfolder.go
+++ b/cmd/minishift/cmd/hostfolder/hostfolder.go
@@ -1,0 +1,159 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package hostfolder
+
+import (
+	"fmt"
+	"github.com/golang/glog"
+	"os"
+
+	"github.com/docker/machine/libmachine"
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	hostfolderActions "github.com/minishift/minishift/pkg/minishift/hostfolder"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+)
+
+var (
+	mountAll bool
+)
+
+// HostfolderCmd represents the resource command for host folders
+var HostfolderCmd = &cobra.Command{
+	Use:   "hostfolder SUBCOMMAND [flags]",
+	Short: "Manage and control host folders for use by the OpenShift cluster.",
+	Long:  `Manage and control host folders for use by the OpenShift cluster.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		cmd.Help()
+	},
+}
+
+var hostfolderMountCmd = &cobra.Command{
+	Use:   "mount HOSTFOLDER_NAME",
+	Short: "Mount a host folder to the running cluster",
+	Long:  `Mount a host folder to the running cluster`,
+	Run: func(cmd *cobra.Command, args []string) {
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		defer api.Close()
+		host, _ := api.Load(constants.MachineName)
+
+		var err error = nil
+		if mountAll {
+			err = hostfolderActions.MountHostfolders(host.Driver)
+		} else {
+			if len(args) < 1 {
+				fmt.Fprintln(os.Stderr, "usage: minishift hostfolder mount [HOSTFOLDER_NAME|--all]")
+				atexit.Exit(1)
+			}
+			err = hostfolderActions.Mount(host.Driver, args[0])
+		}
+
+		if err != nil {
+			glog.Errorln(err)
+			atexit.Exit(1)
+		}
+
+	},
+}
+
+// TODO: refactor; lot of repetition with previous command
+var hostfolderUmountCmd = &cobra.Command{
+	Use:   "umount HOSTFOLDER_NAME",
+	Short: "Unmount a host folder from the running cluster",
+	Long:  `Unmount a host folder from the running cluster`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) < 1 {
+			fmt.Fprintln(os.Stderr, "usage: minishift hostfolder umount HOSTFOLDER_NAME")
+			atexit.Exit(1)
+		}
+
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		defer api.Close()
+		host, _ := api.Load(constants.MachineName)
+
+		err := hostfolderActions.Umount(host.Driver, args[0])
+		if err != nil {
+			glog.Errorln(err)
+			atexit.Exit(1)
+		}
+	},
+}
+
+var hostfolderListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "List an overview of defined host folders",
+	Long:  `List an overview of defined host folders that can be mounted to a running cluster`,
+	Run: func(cmd *cobra.Command, args []string) {
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		defer api.Close()
+		host, _ := api.Load(constants.MachineName)
+
+		hostfolderActions.List(host.Driver)
+	},
+}
+
+var hostfolderAddCmd = &cobra.Command{
+	Use:   "add HOSTFOLDER_NAME",
+	Short: "Add a host folder definition",
+	Long:  `Add a host folder definition that can be mounted to a running cluster`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) < 1 {
+			fmt.Fprintln(os.Stderr, "usage: minishift hostfolder add HOSTFOLDER_NAME")
+			atexit.Exit(1)
+		}
+
+		hostfolderActions.Add(args[0])
+	},
+}
+
+var hostfolderSetupUsersCmd = &cobra.Command{
+	Use:   "setup-users",
+	Short: "Setup credentials for the Users share on a Windows host",
+	Long:  `Setup credentials for the Users share on a Windows host`,
+	Run: func(cmd *cobra.Command, args []string) {
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		defer api.Close()
+
+		hostfolderActions.SetupUsers()
+	},
+}
+
+var hostfolderRemoveCmd = &cobra.Command{
+	Use:   "remove HOSTFOLDER_NAME",
+	Short: "Remove a host folder definition",
+	Long:  `Remove a host folder definition`,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) < 1 {
+			fmt.Fprintln(os.Stderr, "usage: minishift hostfolder remove HOSTFOLDER_NAME")
+			atexit.Exit(1)
+		}
+
+		hostfolderActions.Remove(args[0])
+	},
+}
+
+func init() {
+	// parent set in root.go (RootCmd)
+	//HostfolderCmd.PersistentFlags().Bool(hostfolderActions.AutoMountHostfolders, false, "Auto-mount hostfolders on startup")
+	HostfolderCmd.AddCommand(hostfolderMountCmd)
+	hostfolderMountCmd.Flags().BoolVarP(&mountAll, "all", "a", false, "Mount all defined host folders to the cluster instance.")
+	HostfolderCmd.AddCommand(hostfolderUmountCmd)
+	HostfolderCmd.AddCommand(hostfolderListCmd)
+	HostfolderCmd.AddCommand(hostfolderAddCmd)
+	HostfolderCmd.AddCommand(hostfolderRemoveCmd)
+	HostfolderCmd.AddCommand(hostfolderSetupUsersCmd)
+}
--- a/cmd/minishift/cmd/ip.go
+++ b/cmd/minishift/cmd/ip.go
@@ -18,10 +18,12 @@ package cmd
 
 import (
 	"fmt"
+	"net"
 
 	"github.com/docker/machine/libmachine"
 	"github.com/golang/glog"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/util"
 
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
@@ -49,6 +51,36 @@ var ipCmd = &cobra.Command{
 	},
 }
 
+var hostIpCmd = &cobra.Command{
+	Use:   "hostip",
+	Short: "Gets the IP addresses of the host machine.",
+	Long:  `Gets the IP addresses of the host machine and prints it to standard output.`,
+	Run: func(cmd *cobra.Command, args []string) {
+
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		defer api.Close()
+		host, err := api.Load(constants.MachineName)
+		if err != nil {
+			glog.Errorln("Error getting IP: ", err)
+			atexit.Exit(1)
+		}
+
+		// TODO: too much logic
+		instanceip, _ := host.Driver.GetIP()
+
+		for _, hostaddr := range util.HostIPs() {
+
+			if util.NetworkContains(hostaddr, instanceip) {
+				hostip, _, _ := net.ParseCIDR(hostaddr)
+				if util.IsIPReachable(host.Driver, hostip.String(), false) {
+					println(hostip.String())
+				}
+			}
+		}
+	},
+}
+
 func init() {
 	RootCmd.AddCommand(ipCmd)
+	RootCmd.AddCommand(hostIpCmd)
 }
--- a/pkg/minishift/config/config.go
+++ b/pkg/minishift/config/config.go
@@ -27,12 +27,14 @@ var Config *InstanceConfig
 type InstanceConfig struct {
 	FilePath string `json:"-"`
 	OcPath   string
+
+	HostFolders []HostFolder
 }
 
 // Create new object with data if file exists or
 // Create json file and return object if doesn't exists
 func NewInstanceConfig(path string) (*InstanceConfig, error) {
-	cfg := new(InstanceConfig)
+	cfg := &InstanceConfig{HostFolders: []HostFolder{}}
 	cfg.FilePath = path
 
 	// Check json file existence
--- a/pkg/minishift/config/hostfolder.go
+++ b/pkg/minishift/config/hostfolder.go
@@ -1,0 +1,52 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package config
+
+import (
+	"fmt"
+
+	"github.com/spf13/viper"
+)
+
+type HostFolder struct {
+	Name    string
+	Type    string
+	Options map[string]string
+}
+
+const (
+	HostfoldersDefaultPath  = "/mnt/sda1"
+	HostfoldersMountPathKey = "hostfolders-mountpath"
+)
+
+func GetHostfoldersMountPath(name string) string {
+	overrideMountPath := viper.GetString(HostfoldersMountPathKey)
+	if len(overrideMountPath) > 0 {
+		return fmt.Sprintf("%s/%s", overrideMountPath, name)
+	}
+
+	return fmt.Sprintf("%s/%s", HostfoldersDefaultPath, name)
+}
+
+func (hf *HostFolder) Mountpoint() string {
+	overrideMountPoint := hf.Options["mountpoint"]
+	if len(overrideMountPoint) > 0 {
+		return overrideMountPoint
+	}
+
+	return GetHostfoldersMountPath(hf.Name)
+}
--- a/pkg/minishift/hostfolder/hostfolder.go
+++ b/pkg/minishift/hostfolder/hostfolder.go
@@ -1,0 +1,423 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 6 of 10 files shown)


hferentschik:
Looking at the example config, I assume you are defining three different host folders. I am wondering where the target path is in the config. _Name_ is just an identifier, right? And _uncpath_ is the share to mount. Where is the target path on the VM? Or do you not see this as part of this config? Other than that, the config is pretty much what I expected after our recent discussions. 

Looking at the rest of the code now ...

cmd/minishift/cmd/hostfolder/hostfolder.go
```diff
@@ -0,0 +1,107 @@
...
+	},
+}
+
+var hostfolderMountCmd = &cobra.Command{
+	Use:   "mount HOSTFOLDER_NAME",
+	Short: "",
+	Long:  ``,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) < 1 {
+			fmt.Fprintln(os.Stderr, "usage: minishift hostfolder mount HOSTFOLDER_NAME")
```
budhrg: I think `usage` should be `Usage`

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -129,6 +131,12 @@ func (provisioner *MinishiftProvisioner) Provision(swarmOptions swarm.Options, a
 		return err
 	}
 
+	if hostfolder.IsAutoMount() {
```
hferentschik: 👍

pkg/minishift/util/hostip.go
```diff
@@ -0,0 +1,86 @@
...
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/docker/machine/libmachine/drivers"
+)
+
+// IsIPReachable returns true is IP address is reachable from the virtual instance
+func IsIPReachable(driver drivers.Driver, ip string, printOutput bool) bool {
```
hferentschik: Really, what you want is just a general SSHCommander, right? You don't need to pass the whole driver, right?

hferentschik: Just using the SSHCommander might make it easier to test. Also you just pass what you need. And one can switch where the commander is coming from

gbraad:
@hferentschik since the Name is unique, I also mount the share at a fixed location inside the VM. The path is constructed as `/mnt/sda1/` + sharename

pkg/minishift/hostfolder/hostfolder.go
```diff
@@ -0,0 +1,245 @@
...
+		println("No host folders defined")
+		return
+	}
+
+	share := getHostfolderByName(sharename)
+	if share != nil {
+
+		// TODO: refactor -> duplication
+		sharename := share.Name
+		mountpoint := fmt.Sprintf("/mnt/sda1/%s", sharename)
```
gbraad: this would prevent possible issues with access and SELinux issues... wdyt?

cmd/minishift/cmd/ip.go
```diff
@@ -49,6 +50,27 @@ var ipCmd = &cobra.Command{
 	},
 }
 
+var hostIpCmd = &cobra.Command{
```
gbraad: right...

cmd/minishift/cmd/hostfolder/hostfolder.go
```diff
@@ -0,0 +1,107 @@
...
+
+	"github.com/docker/machine/libmachine"
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	hostfolderActions "github.com/minishift/minishift/pkg/minishift/hostfolder"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+)
+
+// HostfolderCmd represents the resource command for host folders
+var HostfolderCmd = &cobra.Command{
```
gbraad: view would only show a simple overview of Name, and UncPath for cifs, but not password. Just enough for someone to do `rm` or `add` of the entry

cmd/minishift/cmd/config/config.go
```diff
@@ -100,6 +100,10 @@ var settings []Setting = []Setting{
 		callbacks:   []setFn{RequiresRestartMsg},
 	},
 	{
+		name: "hostfolders-automount",
+		set:  SetBool,
```
gbraad: About this, I noticed an inconsistency of these options... dashed and cased... is that something we did?

pkg/minishift/util/hostip.go
gbraad: If I know how to get it there at the right place... the driver seemed the easier way. Will check this later, as I used the SSHCommander before in an earlier test when some of this was still in the provisioner.

budhrg:
Noticed following on starting VM with this minishift build:

```
# .../machines/minishift/minishift.json 
{
	"OcPath": "/home/budhram/.minishift/cache/oc/v1.4.1/oc",
	"HostFolders": null
}
```
`null` should be `[]`.

gbraad:
@budhrg this is likely because of how encoding deals with an initialized array of struct. Will see what can be done

gbraad:
@praveenkumar @budhrg @hferentschik please provide feedback on how-to-test and functionality. Want to get this ready for integration ASAP.

gbraad:
@budhrg to prevent the encoding of empty (unintialized) arrays/nil to `null` in json, you need to do the following:

```
	cfg := &InstanceConfig{HostFolders: []HostFolder{}}
```

instead of
```
	cfg := new(InstanceConfig)
```
[ref](https://github.com/minishift/minishift/blob/master/pkg/minishift/config/config.go#L35)

cmd/minishift/cmd/hostfolder/hostfolder.go
hferentschik: right, and as said, it should report whether it is mounted or not

cmd/minishift/cmd/hostfolder/hostfolder.go
```diff
@@ -0,0 +1,164 @@
...
+		err := hostfolderActions.Mount(host.Driver, args[0])
+		if err != nil {
+			glog.Errorln(err)
+			atexit.Exit(1)
+		}
+	},
+}
+
+var hostfolderMountAllCmd = &cobra.Command{
+	Use:   "mount-all",
```
hferentschik: What's about a 'mount -a' instead of an explicit 'mount-all'?

pkg/minishift/config/hostfolder.go
```diff
@@ -0,0 +1,31 @@
...
+)
+
+type HostFolder struct {
+	Name    string
+	Type    string
+	Options map[string]string
+}
+
+func (hf *HostFolder) Mountpoint() string {
+	return fmt.Sprintf("/mnt/sda1/%s", hf.Name)
```
hferentschik: I might be missing something, but how do I get a my host folder C:\Users\Hardy mounted under /c/Users/Hardy?

hferentschik:
@gbraad running out of time today :-(. I adde some comments, but I really need to run this on a Windows machine to get a full picture.

cmd/minishift/cmd/hostfolder/hostfolder.go
budhrg: I think mount-all gives clarity. `mount-a` doesn't feel good to me at least. No sure about others.

pkg/minishift/config/hostfolder.go
budhrg: Not sure it is possible or not. I think we are assuming here to mount folders into some already fix target location like `/mnt/sda1` here. cc @gbraad

gbraad:
> how do I get a my host folder C:\Users\Hardy mounted under /c/Users/Hardy?

First of all, we do not know where the hostfolder is located on your local machine. This could be `d:\` or even nested on `z:\storagebackend\Users`, because all we know is the UNC Path `//[ip address]/Sharename`.

Second, the username `hardy` is also not known to me, the UNC path does not reveal this. When mounted, the user's documents folder is inside the share, and only his folder is. The username as mentioned is unknown. I can not assume from the username what the name of the user folder is... again this can be different. Especially in the case of a Microsoft Account, the username is `gbraad@redhat.com` but the folder on the disk can be `gbraad`, `g.braad` or even `gbr2431`.

The reason for not placing it in something like `/c/` is also something related to the use of a general CIFS share. In https://youtu.be/fJFvg69iEjw?t=224 you can see that I added a remote share, and this will be mounted by the name in the same path. This is something I can expect to be the same for all hostfolders in this case, instead of something `/c/Users/` and `/?/disks/`.

The actual path of the mountpoint is available as a function on the structure, and can therefore as be modified based on a config option. It seemed to me logical to go for consistency, instead of allowing the user to just randomly specify a mountpoint within the VM. An improvement is to use `/mnt/data/` instead of using `/mnt/sda1`, but this would break the consistency between the current ISO images we have. Both have as `/mnt/sda1/`

cmd/minishift/cmd/hostfolder/hostfolder.go
gbraad: as mentioned @budhrg, `mount -a` do not express intention well enough to me. The same we might want as `umount-all`. Or we could drop these commands. I added them to also be able to test the automount without having to restart all the time.

pkg/minishift/config/hostfolder.go
gbraad: See comment below. Several things are at play here...

hferentschik:
> First of all, we do not know where the hostfolder is located on your local machine. This could be d:\ or even nested on z:\storagebackend\Users, because all we know is the UNC Path //[ip address]/Sharename.

Sure, I am not so concerned about 'C:\Users\Hardy',  I am fine that the mount source is a generic UNC path. Even though for the common cases, like Windows 10 default setup, it would be 'C:\Users\Hardy'. We talked about once in the beginning of the issue, that we want to match the default behaviour of docker-machine. So whether this default behaviour is happening automatically or whether I manually set it up, is a different discussion, but it should be possible to do. So in the case of docker-machine and Windows, the default path is /c/Users/Hardy. On Windows this might not be so useful, but in case you want to use the Samba share on Linux or OS X it becomes important. There you want to have something like /home/hardy. This allows you do to volume mounts while using the tab completion. This is very nice. Provided the directory you want to mount is within your home directory, the paths will be identical. 

> Second, the username hardy is also not known to me, the UNC path does not reveal this. When mounted, the user's documents folder is inside the share, and only his folder is. The username as mentioned is unknown. I can not assume from the username what the name of the user folder is... again this can be different. Especially in the case of a Microsoft Account, the username is gbraad@redhat.com but the folder on the disk can be gbraad, g.braad or even gbr2431.

See above. I am most concerned about not being able to specify where in the VM I can mount the share to. I am happy to have a default for the mount point, but it should also be configurable.

> The actual path of the mountpoint is available as a function on the structure, and can therefore as be modified based on a config option

I guess this is what I am after. How? So how do I get a given CIFS share to a directory of my liking in the VM?

>  It seemed to me logical to go for consistency, instead of allowing the user to just randomly specify a mountpoint within the VM

I think we need the flexibility. What the default is can be discussed, but we need to flexibility to specify the mountpoint and also the code which will ensure that the mount point exists and has the right permissions. CIFS and SSHFS is supposed to be an alternative to guest additions or the NFS based approached in case of xhyve. 

Also on Linux and OS X the default behaviour of docker-machine is not random at all, but makes a lot of sense.

pkg/minishift/util/hostip.go
hferentschik: 👍

cmd/minishift/cmd/config/config.go
hferentschik: Not sure which ones you are referring to precisely, but afaicr this is all inherited

cmd/minishift/cmd/hostfolder/hostfolder.go
hferentschik: > mount-a doesn't feel good to me at least. No sure about others.

Why not? It aligns with the actual _mount_ command on OS X and Linux.

> mount -a does not express intention well enough to me

Why not? I am trying to understand. IMO this leads us into command proliferation.

pkg/minishift/config/hostfolder.go
hferentschik: I don't understand why the mountpoint cannot be made configurable? What becomes the default is one question, but why restricting the user like this. At least on OS X and Linux there is a very good reason to use /Users/john or /home/john. So if this is not the default, I definitely want to configure it this way. Also, our starting point was to mimic current docker-machine behaviour.

hferentschik:
@gbraad FYI, you will need to rebase this pull request. The command synopses files don't need to be checked in anymore and the docs you added needs to go into the respective adoc file now. Should be straight forward to figure out.

hferentschik:
Ok, so I tried this on my Windows 10 machine which uses a registered Microsoft account as login type. I create a host folder via `hostfolder add` which results in the following config:

    {
    	"OcPath": "",
    	"HostFolders": [
    		{
    			"Name": "foo",
    			"Type": "cifs",
    			"Options": {
    				"domain": "",
    				"password": "*****",
    				"uncpath": "\\\\192.168.99.1\\hardy",
    				"username": "foo@acme.com"
    			}
    		}
    	]
    }

The username is my Microsoft account email and the password my Microsoft account password. When I then try to mount I just get:

    $ .\minishift.exe hostfolder mount foo
    Mounting 'foo': '\\192.168.99.1\hardy' ... FAIL

Any ideas? I checked that the host is reachable from with the VM as well.

hferentschik:
Regarding auto-mount. You still have to define the host folder config (basically before you run _start_) and then it gets automatically mounted, right? This is not auto-mount in the sense that some sort of default share gets transparently mounted? 

I am mainly asking, since this needs to be integrated into DevStudio. How would this look like? Eclipse tooling basically expects that it can just call _start_. So I guess this makes it the responsibility of the Windows installer to set this up (asking for UNC path and credentials as part of the install). I guess we need to loop in @dgolovin and discuss this as part of the Kabal.

hferentschik:
@gbraad btw, there is also the CDK to consider. The aim is to match CDK 2.x functionality and there the vagrant-sshfs plugin mounted per default as per docker-machine.

gbraad:
@hferentschik please read: https://github.com/minishift/minishift/blob/master/docs/source/using/managing-minishift.adoc#cifs on how to use a Microsoft bound account. Tl;dr you are missing the domain part of the login. If it does not work, please reopen: https://github.com/minishift/minishift/issues/422#issuecomment-281009269

gbraad:
> The aim is to match CDK 2.x functionality and there the vagrant-sshfs plugin mounted per default 

Not following what you mean with this. Technically possible, but not part of this PR... or is it? At the moment the sshfs functionality is still in POC-phase

gbraad:
> this makes it the responsibility of the Windows installer 

Yes, because I can not extract login credentials. This needs to be provided by the user and also needs to be able to be reset when the user has changed them. This need to change can be triggered for instance by a domain policy, so this is not uncommon. Note: docker-machine provides a 'Reset credentials' option, as mentioned in https://github.com/minishift/minishift/issues/316#issuecomment-275075150

hferentschik:
> please read: https://github.com/minishift/minishift/blob/master/docs/source/using/managing-minishift.adoc#cifs on how to use a Microsoft bound account. 

I did. What I took from there was that I need to use my email as per Microsoft account

> If your Windows account is linked to a Microsoft account, you must use the full Microsoft account email address to authenticate, for example jpillow@amigas.us

> Tl;dr you are missing the domain part of the login

This is the part I don't get. According to the docs, it says:

> This ensures that the domain value that contains the computer name is provided.

I read this in a way that, if I use the email and password for my Microsoft account, then some domain value is "provided". TBH, I don't really know how to take this. I get prompted for a domain, but what it it? And should it not be provided?

gbraad:
> "\\\\192.168.99.1\\hardy"

Is this share available? Please use the command `minishift hostfolder setup-users` to store the credentials and entry to the Users share.

For more information, please look at: https://youtu.be/fJFvg69iEjw?t=25

hferentschik:
> Yes, because I can not extract login credentials. This needs to be provided by the user and also needs to be able to be reset when the user has changed them. This need to change can be triggered for instance by a domain policy, so this is not uncommon. Note: docker-machine provides a 'Reset credentials' option, as mentioned in #316 (comment)

No problem. Makes perfect sense to me. I am just saying that this is something we need to sync with @dgolovin and the DevStudio team, since it requires changes in the installer. From that perspective, the SSHFS solution is better, since it can be done password free.

gbraad:
> From that perspective, the SSHFS solution is better, since it can be done password free.

Totally agree, however... People coming from a Windows-world might prefer their own tools and techstack...  it is not uncommon to work on a shared drive in a Windows environment.

hferentschik:
>> The aim is to match CDK 2.x functionality and there the vagrant-sshfs plugin mounted per default

> Not following what you mean with this. Technically possible, but not part of this PR... or is it? At the moment the sshfs functionality is still in POC-phase

Sorry, might have phrased this wrong. I did not say that SSHFS should be part of this pull request. As you say, this is still work in progress. This was a comment regarding the need for making the mount point at least configurable. With CDK 2.x we followed the approach of mounting the users home directory under '/home/john', 'Users/johen', resp '/c/Users/john', depending on OS. So bottom line, we need the mount point configurable.

gbraad:
> This ensures that the domain value that contains the computer name is provided.

This has not been properly reviewed... This should have been "Please ensure"

hferentschik:
> From that perspective, the SSHFS solution is better, since it can be done password free.

>> Totally agree, however... People coming from a Windows-world might prefer their own tools and techstack... it is not uncommon to work on a shared drive in a Windows environment.

Sure, sure, I am fine using CIFS shares on Windows. I think this stuff is great. What I am saying is:

* mountpoint needs to be configurable 
* default mountpoint needs to be discussed 
* if we make this the default host folder solution for Windows we need to sync with the installer, so that it can setup the necessary configuration as part of the install
* concrete commands needs to be discussed, for example I still prefer 'mount --all' over another command like 'mount-all'. I would like to know how others think abbot this one.

gbraad:
> I would like to know how others think abbot this one.

Still waiting for feedback... have requested this several times.

gbraad:
> default mountpoint needs to be discussed

Default mountpoint can be a given value; and allow optional override from config

gbraad:
> mountpoint needs to be configurable

Per `hostfolder add` ? Possible... on input the default can be shown and allow a non-empty value to override?

gbraad:
> "\\192.168.99.1\hardy"

Samba does not accept `\\\\`, will implement a rewrite to an internal representation using '//' instead. This would solve the problem with slashes being wrong...

hferentschik:
> Per hostfolder add ?

Yes

> on input the default can be shown and allow a non-empty value to override?

+1

hferentschik:
> Samba does not accept \\\\, will implement a rewrite to an internal representation using '//' instead. This would solve the problem with slashes being wrong...

👍 Sounds good. If we can catch this that would be great. It is really a classic after all.

gbraad:
>mount-all

This is a command we want in the first place? separate from the discussion if it should be `mount-all`, `mount -a` or `mount --all`... If not, the solution is much simpler. Mounting of all the host folders happens anyways on start of minishift. This command could be helpful when automount on startup is disabled.

> automount

Should the `hostfolders-automount` maybe be `hostfolders-automount- on-start` ?

hferentschik:
@praveenkumar, @LalatenduMohanty what's your take on some of the questions/issues discussed in this pull request? https://github.com/minishift/minishift/pull/526#issuecomment-286515525 is summarising the bigger questions we have atm.

gbraad:
> concrete commands needs to be discussed

`minishift hostfolder [mount|umount|add|remove|view]` exist in this change.

Additionally we have a `mount-all`, but I actually did not discuss this originally as I am not sure if we should include it. It is more explicit in meaning, but it can also be hidden away inside the `mount` command.

cmd/minishift/cmd/hostfolder/hostfolder.go
gbraad: This was not implemented because of the roundtripping that is involved. On Windows it took me 40 seconds or more the get the state of just 3 host folders. See #529...

pkg/minishift/config/hostfolder.go
gbraad: > mimic current docker-machine behaviour

Mimic which behaviour? Docker machine does not support this functionality... Docker machine shares a drive: https://rominirani.com/docker-on-windows-mounting-host-directories-d96f3f056a2c

gbraad:
> > on input the default can be shown and allow a non-empty value to override?
> +1

Would the default (initial) mountpoint be configured on global level with Viper, or on instance level?
The same question applies to `hostfolders-automount`.

cmd/minishift/cmd/config/config.go
LalatenduMohanty: What about just keeping the name as `hostfolders`.

cmd/minishift/cmd/hostfolder/hostfolder.go
gbraad: Actually, I have a solution... a single `mount` call and grep over this for each of the mounts... this way the mount is not verified on the host.

As mentioned, I think it is better to call the command `list`.

hferentschik:
> Would the default (initial) mountpoint be configured on global level with Viper, 

hmm, I don't think so. Is it not more that this would be part of the interactive configuration where the default value would be displayed. That said, is there an non-interactive mode where I can create the config for a host folder in a single call without user interaction? Is this something you are planning to do or would you assume that the user run this case edits the JSON config directly (which might be good enough for a first cut)?

pkg/minishift/config/config.go
```diff
@@ -27,12 +27,14 @@ var Config *InstanceConfig
...
 	OcPath   string
+
+	HostFolders []HostFolder
 }
 
 // Create new object with data if file exists or
 // Create json file and return object if doesn't exists
 func NewInstanceConfig(path string) (*InstanceConfig, error) {
-	cfg := new(InstanceConfig)
+	cfg := &InstanceConfig{HostFolders: []HostFolder{}}
```
hferentschik: I am wondering whether it is the right thing to add this here? There is also the config.json in ~/.minishift/config which contains the other configuration values like cpus, memory etc. I am wondering whether this config should not go there (we might need to add some helper methods to read/write to this config similar to what we have here).

The reason I am asking is, that this config is specific to the instance. It also gets deleted with the instance. So if you do a `minishift delete` you need to re-create the settings. Settings in ~/.minishift/config/config.json on the other hand persist across VM deletion which I think is more what a user would like to do. WDYT?

hferentschik:
@gbraad IMO with all these comments, the pull request gets a bit unwieldy. What's about you close this pull request and create a new one, once you rebased and applied some of the changes we discussed. 

This way we get a bit of a clean slate for further reviews. WDYT?

pkg/minishift/config/config.go
gbraad: hostfolders are in my opinion instance specific... don't they? I originally had them in the global config, but this felt weird. I might have a specific set of shares for an instance...

budhrg: Yes, IMO it is instance specific. making it part of ~/.minishift/config/config.json will hinder the multiple instance story.

I might able to mount different folder to different instances. `many-many relationship` here.

budhrg:
+1 for Close this PR and start new with rebase

gbraad:
CentOS-CI: Connection to 172.19.3.110 closed. during the compilation...

gbraad:
Added list with mounted state:
```
Name    Mountpoint        Remote path           Mounted
disks   /mnt/sda1/disks   //10.0.21.33/Disks    Y
ds      /home/user/Mine   //10.0.21.33/Disks    Y
dsd     /mnt/sda1/dsd     dsa                   N
```

and custom mountpoints

hferentschik:
> Added list with mounted state:

Nice

> and custom mountpoints

+1, will take a look asap.

gbraad:
@hferentschik #581 is the new PR. Busy with some of the tests, but a small issue with this now... so not pushed any more changes.

cmd/minishift/cmd/config/config.go
gbraad: @LalatenduMohanty actually we are dealing with a resource, which is generally pluralized. I actually wanted to do the same for the actual command...

PR #527, opened 2017-03-08
Issue #509 Add developer User to sudoers

### LINKED ISSUE #509: Give the developer sudoers role
Recently @hferentschik pointed a nice article about how we can implement same linux philosophy of having sudo to perform admin tasks and I think this will helpful for us and for users.

- https://bierkowski.com/openshift-cli-morsels-sudo-and-user-impersonation/

praveenkumar:
Fix #509

### DIFF
--- a/pkg/minishift/openshift/oc.go
+++ b/pkg/minishift/openshift/oc.go
@@ -1,0 +1,44 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openshift
+
+import (
+	instanceState "github.com/minishift/minishift/pkg/minishift/config"
+	"github.com/minishift/minishift/pkg/util"
+)
+
+// runner executes commands on the host
+var runner util.Runner = &util.RealRunner{}
+
+// Add developer user to cluster sudoers
+func AddSudoersRoleForUser(user string) error {
+	cmdName := instanceState.Config.OcPath
+	cmdArgs := []string{"login", "-u", "system:admin"}
+	if _, err := runner.Output(cmdName, cmdArgs...); err != nil {
+		return err
+	}
+	// https://docs.openshift.org/latest/architecture/additional_concepts/authentication.html#authentication-impersonation
+	cmdArgs = []string{"adm", "policy", "add-cluster-role-to-user", "sudoer", user}
+	if _, err := runner.Output(cmdName, cmdArgs...); err != nil {
+		return err
+	}
+	cmdArgs = []string{"login", "-u", user}
+	if _, err := runner.Output(cmdName, cmdArgs...); err != nil {
+		return err
+	}
+	return nil
+}
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -66,6 +66,22 @@ func (m *Minishift) shouldHaveAValidIPAddress() error {
 	return nil
 }
 
+func (m *Minishift) executingOcCommand(command string) error {
+	ocRunner := m.runner.GetOcRunner()
+	if ocRunner == nil {
+		return fmt.Errorf("Minishift is not Running")
+	}
+	cmdOut, cmdErr, cmdExit := ocRunner.RunCommand(command)
+	lastCommandOutput = CommandOutput{
+		command,
+		cmdOut,
+		cmdErr,
+		cmdExit,
+	}
+
+	return nil
+}
+
 func (m *Minishift) executingCommand(command string) error {
 	// TODO: there must be smarter way to destruct
 	cmdOut, cmdErr, cmdExit := m.runner.RunCommand(command)
@@ -146,6 +162,7 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`Minishift should have state "([^"]*)"`, m.shouldHaveState)
 	s.Step(`Minishift should have a valid IP address`, m.shouldHaveAValidIPAddress)
 	s.Step(`executing "minishift ([^"]*)"`, m.executingCommand)
+	s.Step(`executing "oc ([^"]*)`, m.executingOcCommand)
 	s.Step(`([^"]*) should contain ([^"]*)`, m.commandReturnShouldContain)
 	s.Step(`([^"]*) should contain`, m.commandReturnShouldContainContent)
 	s.Step(`([^"]*) should equal ([^"]*)`, m.commandReturnShouldEqual)
@@ -154,7 +171,7 @@ func FeatureContext(s *godog.Suite) {
 
 	s.BeforeScenario(func(interface{}) {
 		testDir := setUp()
-		os.RemoveAll(testDir)
+		fmt.Println("Running Integration test in : ", testDir)
 	})
 
 	s.AfterScenario(func(interface{}, error) {
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -8,6 +8,17 @@ Feature: Basic
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
 
+     # Check if developer have sudo permission then oc should output as below with other
+     # cluster roles
+     # sudoer   /sudoer   developer
+     When executing "oc --as system:admin get clusterrolebindings"
+     Then stderr should be empty
+     And  exitcode should equal 0
+     And  stdout should contain
+     """
+     sudoer
+     """
+
      When executing "minishift ssh echo hello"
      Then stderr should be empty
       And exitcode should equal 0
--- a/test/integration/util/util.go
+++ b/test/integration/util/util.go
@@ -27,6 +27,9 @@ import (
 
 	"bytes"
 	"syscall"
+
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	instanceState "github.com/minishift/minishift/pkg/minishift/config"
 )
 
 type MinishiftRunner struct {
@@ -34,9 +37,13 @@ type MinishiftRunner struct {
 	CommandArgs string
 }
 
-func (m *MinishiftRunner) RunCommand(command string) (stdOut string, stdErr string, exitCode int) {
+type OcRunner struct {
+	CommandPath string
+}
+
+func runCommand(command string, commandPath string) (stdOut string, stdErr string, exitCode int) {
 	commandArr := strings.Split(command, " ")
-	path, _ := filepath.Abs(m.CommandPath)
+	path, _ := filepath.Abs(commandPath)
 	cmd := exec.Command(path, commandArr...)
 
 	var outbuf, errbuf bytes.Buffer
@@ -65,6 +72,11 @@ func (m *MinishiftRunner) RunCommand(command string) (stdOut string, stdErr stri
 	return
 }
 
+func (m *MinishiftRunner) RunCommand(command string) (stdOut string, stdErr string, exitCode int) {
+	stdOut, stdErr, exitCode = runCommand(command, m.CommandPath)
+	return
+}
+
 func (m *MinishiftRunner) Start() {
 	m.RunCommand(fmt.Sprintf("start %s", m.CommandArgs))
 }
@@ -78,6 +90,13 @@ func (m *MinishiftRunner) EnsureRunning() {
 
 func (m *MinishiftRunner) IsRunning() bool {
 	return m.GetStatus() == "Running"
+}
+
+func (m *MinishiftRunner) GetOcRunner() *OcRunner {
+	if m.IsRunning() {
+		return NewOcRunner()
+	}
+	return nil
 }
 
 func (m *MinishiftRunner) EnsureDeleted() {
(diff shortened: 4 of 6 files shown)


pkg/minishift/openshift/oc.go
```diff
@@ -0,0 +1,44 @@
...
+
+package openshift
+
+import (
+	instanceState "github.com/minishift/minishift/pkg/minishift/config"
+	"github.com/minishift/minishift/pkg/util"
+)
+
+// runner executes commands on the host
+var runner util.Runner = &util.RealRunner{}
```
budhrg: `util.Runner` is it required? 

I think datatype will be inferred here.

test/integration/util/util.go
```diff
@@ -120,7 +122,7 @@ type OcRunner struct {
 }
 
 func NewOcRunner() *OcRunner {
-	p, _ := exec.LookPath("oc")
```
praveenkumar: This will not work because as per doc LookPath only able to search binary if that is part of $PATH https://golang.org/pkg/os/exec/#example_LookPath

test/integration/util/util.go
```diff
@@ -120,7 +122,7 @@ type OcRunner struct {
...
+	p := path.Join(os.Getenv(constants.MiniShiftHomeEnv), "cache", "oc", "v1.4.1", "oc")
```
budhrg: :+1:

pkg/minishift/openshift/oc.go
```diff
@@ -0,0 +1,44 @@
...
+	cmdArgs = []string{"adm", "policy", "add-cluster-role-to-user", "sudoer", "developer"}
+	if _, err := runner.Output(cmdName, cmdArgs...); err != nil {
+		return err
+	}
+	cmdArgs = []string{"login", "-u", "developer"}
+	if _, err := runner.Output(cmdName, cmdArgs...); err != nil {
+		return err
+	}
+	return nil
+}
```
budhrg: go fmt will fail here.

pkg/minishift/openshift/oc.go
praveenkumar: @budhrg without runner we can't execute the command to system (VM)

budhrg: Sorry, should have commented clearly.

I mean to say this `var runner = &util.RealRunner{}` , We can omit datatype during variable definition.

test/integration/util/util.go
praveenkumar: This is still hard code the oc version so I think will use `version.GetOpenShiftVersion()`

pkg/minishift/openshift/oc.go
praveenkumar: @budhrg right but this will make little bit clear where actually it's coming from but yes we can omit.

test/integration/util/util.go
praveenkumar: Looks like `version.GetOpenShiftVersion()` not going to work because it will going to get `v0.0.0-unset` as version.

budhrg: Can we at least extract oc version into some constants ? 

We eventually might need to extract common test constants to some file

praveenkumar: @budhrg I think best way to handle it during setup add the `testDir` to $PATH and then use `LookPath` but let's discuss it.

budhrg: @praveenkumar , 
I think we can cache the common dirs like `oc` dir, as 

```
OcDir := path.Join(os.Getenv(constants.MiniShiftHomeEnv), "cache", "oc")
ISODir := path.Join(os.Getenv(constants.MiniShiftHomeEnv), "cache", "iso")
```

etc.

Doesn't feel good to add testDir to $PATH. This also might not work in Windows but unsure.

budhrg: Not exactly what I thought. What if I had to use this path in any other code? 

I think let's not do this as part of it. We can refactor it later or discuss the proper solution.

test/integration/features/basic.feature
```diff
@@ -7,6 +7,7 @@ Feature: Basic
     Given executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
+      And Developer should have sudoers access
```
gbraad:
```gherkin
When executing "openshift fdfsfd"  # or similar
Then developer should have sudo access
```

You intertwine the cases of action and behaviour

test/integration/basic_test.go
```diff
@@ -66,6 +66,16 @@ func (m *Minishift) shouldHaveAValidIPAddress() error {
 	return nil
 }
 
+func shouldHaveSudoerAccess() error {
+	ocRunner := util.NewOcRunner()
+	cmdArgs := []string{"--as", "system:admin", "get", "node"}
+	cmdOut, _ := ocRunner.RunCommand(cmdArgs)
+	if !strings.Contains(string(cmdOut), "NAME") {
+		return fmt.Errorf("Not able to set developers as sudoers: %s", string(cmdOut))
```
gbraad: maybe I am not following your intention here. likely due to the fact of the error message.

Does it set the access level, or perform an action as a certain user?

budhrg:
@praveenkumar , Could we have commit message in this format `Issue #<issue no> message`? I mean `#` is missing.

test/integration/features/basic.feature
```diff
@@ -28,4 +36,4 @@ Feature: Basic
      When executing "minishift stop"
      Then Minishift should have state "Stopped"
      When executing "minishift delete"
-     Then Minishift should have state "Does Not Exist"
+     Then Minishift should have state "Does Not Exist"
```
gbraad: newline or not? I get confused what people want...

test/integration/features/basic.feature
```diff
@@ -8,6 +8,14 @@ Feature: Basic
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
```
gbraad: It is probably better to actually capture this in a new function that describes more in detail what happens... and why this output. but for the sake of the PR, it might be an idea to just add a comment above:

```gherkin
  # Check if 
  When executing "oc ..."
```

test/integration/features/basic.feature
praveenkumar: @gbraad It's not the new line it just that I put my features bit before it so that's comes it like that. I don't like this way so showing stuff from github (it's confusing) :(

test/integration/features/basic.feature
praveenkumar: @gbraad I agree, I added some comment so that we are clear what we want to achieve here.

test/integration/features/basic.feature
praveenkumar: @gbraad nvm I got what you wanted to say. I added newline to EOF.

gbraad: Check...

but we can get the IP of the VM. this is `minishift ip`, and we can actually store this if needed... we will probably need this more often. But let's leave this as a possible improvement.

We could collect a list for this in a new issue.

test/integration/features/basic.feature
```diff
@@ -8,6 +8,19 @@ Feature: Basic
...
 
+     # Check if developer have sudo permission then oc should output as below
+     # Since we can't determine the vm ip before had so just make sure `Name`
+     # should be part of output string
+     # NAME           STATUS    AGE
+     # 192.168.64.2   Ready     1m
+     When executing "oc --as system:admin get node"
```
hferentschik: This is really a indirect test. It makes several assumptions which are not clear, for example that a non sudoer user cannot view this. I think we should add a couple of things here. The direct test imo is `oc get cluster role sudoer` needs to return sudoer. 

If we want we can than also keep this test, but would also add the opposite. `When executing "oc  get node" where the output is "User "developer" cannot list all nodes in the cluster" and the exit status 1

test/integration/util/util.go
hferentschik: I don't get it. Why are you not using '~/.minishift/machines/minishift.json'? That's why we introduced it, right? The LookupPath was a crutch until we could do it. 

So as a minimum the oc path should be read from the config file. 

There are a couple of other issues here which need addressing. Preferably sooner than later. 

* The integration tests are interfering atm with the default Minishift directory. As part of the integration test setup we should create a known directory within our source tree (e.g. out/integration-test or something). Easiest would be to create the directory in make and then set this directory as value for MINISHIFT_HOME when calling the integration test. Problem is that in this case one needs to drive the test via Make. 
* The second problem is constants.MachineName, we need to take this value from an env variable as well, otherwise even though we can separate the home directories, we still cannot create multiple machines since the names are clashing. It needs to be tested whether the name can easily change. There might be some hard coded values which will cause trouble.

Bottom line, as a bare minimum we need to adjust the path. I'd love to see the other two issues addressed as well, but let's discuss.

test/integration/util/util.go
```diff
@@ -34,9 +36,9 @@ type MinishiftRunner struct {
 	CommandArgs string
 }
 
-func (m *MinishiftRunner) RunCommand(command string) (stdOut string, stdErr string, exitCode int) {
+func runCommand(command string, commandPath string) (stdOut string, stdErr string, exitCode int) {
```
hferentschik: hmm, I'd probably wrap this into a struct similar to _MinishiftRunner_, just _OcRunner_ and then I let me give _MinishiftRunner_ an instance of _OcRunner_.

test/integration/util/util.go
```diff
@@ -120,7 +127,7 @@ type OcRunner struct {
 }
 
 func NewOcRunner() *OcRunner {
```
hferentschik: Maybe this constructor should actually take the path. This way it becomes more flexible. One can determine the right path from the outside and create a Runner with any odd oc version, not just the one provided by minishift.json

pkg/minishift/openshift/oc.go
```diff
@@ -0,0 +1,44 @@
...
+import (
+	instanceState "github.com/minishift/minishift/pkg/minishift/config"
+	"github.com/minishift/minishift/pkg/util"
+)
+
+// runner executes commands on the host
+var runner util.Runner = &util.RealRunner{}
+
+// Add developer user to cluster sudoers
+func AddSudoersRoleForDeveloperUser() error {
```
hferentschik: Looking into the future something like _AddSudoersRoleForUser(user string)_ might be better. Once we add the ability to create users, we probably want to add sudoer roles fro them as well.

test/integration/util/util.go
hferentschik: This was a crutch until we would be able to tell which oc version to use. We can now.

hferentschik:
Ahh, and this really need to be mentioned in the docs somewhere. Probably somewhere were we talk about the various logins.

budhrg:
Hi @Preeticp , I think we don't add labels to PR so I have removed it. Added doc label to related issue. Not sure about needs-investigation label.

Preeticp:
@budhrg thanks for moving the tags to the issue.
Afaiu, 'needs-investigation' tag is for tasks which are still being investigated and worked on by the developers, and I figured work is still happening on this PR. If dev work on this PR is complete, then we do not need the 'needs-investigation' tag.

budhrg:
> Afaiu, 'needs-investigation' tag is for tasks which are still being investigated and worked on by the developers

@Preeticp , actually `needs-investigation` is to be added to issues which developers think that they can't able to complete at the moment probably due to incomplete knowledge and think if done bit of investigation then can able to complete the task. Example, https://github.com/minishift/minishift/issues/529 where we found slow communication when using Minishift + Hyper-V. Now, at least for me I don't know much the internals of hyper-v hypervisor, and need to investigate why this behavior particularly for Hyper-V.

Good to know you are also taking part in labelling :+1:

hferentschik:
Actually, I think we are talking about just a couple of sentences here and maybe a link to some relevant information. @praveenkumar, can you not add this so that we can merge this asap.

praveenkumar:
> can you not add this so that we can merge this asap.

@hferentschik I hit with some of different issue during the change, will able to push changes as per your suggestion soon

test/integration/basic_test.go
```diff
@@ -154,7 +171,7 @@ func FeatureContext(s *godog.Suite) {
 
 	s.BeforeScenario(func(interface{}) {
 		testDir := setUp()
-		os.RemoveAll(testDir)
```
praveenkumar: This is no use in our case and shouldn't be part of BeforeScenario.

test/integration/basic_test.go
```diff
@@ -154,7 +171,7 @@ func FeatureContext(s *godog.Suite) {
...
+		fmt.Println("Running Integration test in : ", testDir)
```
praveenkumar: This will help to find where the actual test is running to findout which directory it's in.

hferentschik: +1, eventually it would be nice to have this as some configurable log output, but let's have it as message to std out for now

test/integration/features/basic.feature
```diff
@@ -8,6 +8,17 @@ Feature: Basic
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
 
+     # Check if developer have sudo permission then oc should output as below with other
+     # cluster roles
+     # sudoer   /sudoer   developer
+     When executing "oc --as system:admin get clusterrolebindings"
```
hferentschik: 👍 awesome

test/integration/util/util.go
```diff
@@ -80,6 +92,13 @@ func (m *MinishiftRunner) IsRunning() bool {
 	return m.GetStatus() == "Running"
 }
 
+func (m *MinishiftRunner) GetOcRunner() *OcRunner {
+	if m.IsRunning() {
```
hferentschik: :-) Right, that's what I meant

test/integration/util/util.go
```diff
@@ -115,12 +134,10 @@ func (m *MinishiftRunner) CheckStatus(desired string) bool {
 	return m.GetStatus() == desired
 }
 
-type OcRunner struct {
-	CommandPath string
-}
-
 func NewOcRunner() *OcRunner {
-	p, _ := exec.LookPath("oc")
+	jsonDataPath := filepath.Join(os.Getenv(constants.MiniShiftHomeEnv), "machines", constants.MachineName+".json")
```
hferentschik: 👍

hferentschik:
Merged, thanks. Very nice indeed.

Issue #139, opened 2016-11-04
Create build target to build HTML documentation


hferentschik:
At release time the documentation should be rendered into HTML. This allows to have versioned documentation which can be deployed on some web server (and included into the dist zip).

gbraad:
Which tools are we using to generate the documentation? using pandoc (or jekyll) allows us to create a template and generate HTML from there based on markdown. Are we still switching?

hferentschik:
Right now we have no tooling as part of Minihshift. Downstream, there are several tools at play, I think including pandoc. Also the preferred format is actually  asciidoc.

The problem with the downstream toolchain adds additional prerequisites to the build setup which might be even hard to meet on say Windows. So the idea is to provide a Docker container with all the required tooling and build/render the docs in the container, while still editing on the host using your preferred editor. Also one can then use Minishift itself to run this container :-)

gbraad:
@hferentschik a container is ideal; https://github.com/gbraad/docugen/blob/master/Dockerfile I used pandoc and jekyll (and also did PDF generation)... but I would strongly suggest not to do PDF (using xelatex) as it increases the container size dramatically. Also, pandoc in most distros is way too old... so an image with the latest is preferred.

hferentschik:
> a container is ideal; 

lol, we seem to have same experience. We did this when switching to awestruct for the in.relation.to website - https://github.com/hibernate/in.relation.to/tree/staging/docker

> https://github.com/gbraad/docugen/blob/master/Dockerfile I used pandoc and jekyll

I see. Does this allow for live rendering of the site? Basically you make a change and just reload (or even better use live-reload) the page in the browser. 

@rkratky, maybe this setup can help you as well.

gbraad:
> live rendering

`jekyll` can run in a server-watch mode, and actually gitbook AFAIR also, but that is kinda nice for technical documentation: https://github.com/gbraad/dockerfile-gitbook-server.

hferentschik:
whichever way works. 

Once we have the Dockerfile with the dependencies to build and render the docs we can integrate everything into the Makefile and add some convenience targets.

hferentschik:
@thatdocslady on that note, we will need the appropriate styles from docs.openshift.org so that we can render the right HTML. Could you please follow up on these?

LalatenduMohanty:
Closing the issue as #530 is merged.

Issue #140, opened 2016-11-04
Switch documentation to Asciidoc


hferentschik:
Relates to issue #139. Using Asciidoc(tor) gives us more possibilities to build a complete documentation set with  various output formats (for example HTML or PDF).

Given that Asciidoc also renders directly when viewed on GitHub, we don't loose anything there.

hferentschik:
As a reference - https://github.com/spf13/hugo/issues/470
There seems to be a way to generate Asciidoc as well from the Cobra source files. See also gen_help_test.go in the root directory of the repo.

aheusingfeld:
There is an ongoing discussion in the asciidoctor project https://github.com/asciidoctor/asciidoctor.org/issues/462 - AFAIK the java tooling is still the best way to convert markdown to asciidoc at the moment, though Dan started a converter in https://github.com/opendevise/kramdown-asciidoc

hferentschik:
@aheusingfeld thanks for the info. Interesting. A shame though to add something like Java or Ruby into the mix. I guess there is nothing intrinsically wrong with Markdown. Provided the site generator we are going to use can cope with it, there is really no need to switch to Asciidoc(tor) for now.

aheusingfeld:
Well, we're still just talking about the one-time effort to convert the marksown files into asciidoc files, right?

To afterwards generate output from the asciidoc sources, asciidoctor is definitely the best option. What I personally like about asciidoc is the increased flexibility i.e. I very often include parts of the sourcecode into my asccidoc documentation files. This ensures that samples in the docs are always uptodate.

BTW: What are your requirements and which output formats would you want to target ATM and midterm?

hferentschik:
> Well, we're still just talking about the one-time effort to convert the marksown files into asciidoc files, right?

Yes and no. There is a bunch of markdown files which we only need to convert once. However, there are also some generated Markdown files (https://github.com/minishift/minishift/blob/master/gen_help_text.go) which Cobra generates from the CLI sources. This generator afaiu just generates manpage or markdown format. So if we decide to keep this generated docs (which is nice, since this ensures that the docs are in sync with the actual sources), we would need to re-generate each time we build the the Cobra docs.

> To afterwards generate output from the asciidoc sources, asciidoctor is definitely the best option.

Sure, no questions asked.

> What I personally like about asciidoc is the increased flexibility i.e. I very often include parts of the sourcecode into my asccidoc documentation files. 

Same here, Asciidoc is ways more powerful

[...]

PR #530, opened 2017-03-09
Issue #139 Restructuring of docs and creating build target for generating the docs

hferentschik:
This addresses issue #139 and follows up on pull request #520.

The pull request is complete, but cannot be merged yet. This pull request addresses the technical aspects of generating the docs and eventually publishing them to docs.openshift.org. The non generated documentation needs to be converted from Markdown to Asciidoc, links must be adjusted, etc

What this pull request does is to introduce Middleman as site building tool. The actual build happens in a Docker container which can also be build from the sources. Provided one has a running Docker container and your shell is setup to use it, just run:

    $ make gen_docs

To edit the documentation and have automatic rebuild of the site, use:

    $ make serve_docs

What needs to be done next:

* Discussion around final layout (do we want to keep the use vs reference split)?
* Adoc conversion
* README refactoring
* Integration into docs.openshift.org build
* Potentially some styling to match the OpenShift L&F instead of the plain Middleman style

I suggest we push this onto a feature branch on the main repo. We can then complete the outstanding tasks on the feature branch and merge back to master once we ready to do a deploy onto docs.openshift.org. Drawback of course is, that one needs to manually sync ongoing docs changes from master.


hferentschik:
@rkratky, have a look. It's come a long way from your #520 pull request, but I think it covers all our goals/requirements we started off with. I abandoned ascii_binder. AFAICT it does not work with a mixed repo, where the docs are just a part of something bigger.

hferentschik:
@thatdocslady, @rkratky we need to have a discussion on how we want to take this to the finish line.

budhrg:
> I suggest we push this onto a feature branch on the main repo. We can then complete the outstanding tasks on the feature branch and merge back to master once we ready to do a deploy onto docs.openshift.org. Drawback of course is, that one needs to manually sync ongoing docs changes from master.

:+1:

thatdocslady:
@hferentschik since I'm not familiar with the inner workings of dockerfiles, I can only provide feedback for the docs you added in the developer document, which isn't a deal-breaker at this point and might change once we do further testing and fine-tune the setup. I recommend for the sake of moving forward with the implementation to create a separate issue for doc review and go forward with this (assuming that everyone's testing is ok). 

Also, I love Middleman so I'm happy. We use it for a lot of static sites in OSAS and AsciiBinder isn't necessary for our local staging if we accept vanilla asciidoc preview. We will provide OpenShift with adoc files anyway, so I'm happy with whatever local preview tooling that works within the container.

hferentschik:
> since I'm not familiar with the inner workings of dockerfiles, 

well, you don't need to review the Dockerfile per se, but I would be very interested to know whether you can get the docs build working. After all, you will need it on an ongoing basis

> I can only provide feedback for the docs you added in the developer document

I think we can pretty much ignore any content changes to the documents under _docs/source/use_ until we have a plan on how we do the final migration. I just added the content to developing.md to make sure that people know how to use this.

> which isn't a deal-breaker at this point and might change once we do further testing and fine-tune the setup. 

Exactly. So here would be my first question. As per @rkratky proposal, there was a split between between _reference_ and _use_. The former contained the generated command synopses whereas the latter contained our manually maintained documentation. I can see some value in this split, but it is technically really not required. In fact we have everything together right now. The reason I am asking is, because right now links from the usage documentation to the  synopses are all broken. I would like to adjust the links (preferably as part of this pull request), but first want to confirm that we want to take this route. @rkratky, any specific reason you did the split?

> I recommend for the sake of moving forward with the implementation to create a separate issue for doc review and go forward with this (assuming that everyone's testing is ok).

So right, if we agree this is the way forward, we would create follow up issues. This would include:

* Conversion to adoc for the usage docs
* Creation of docs entry point
* Refactoring of README
* Integration into docs.openshift.org build/setup

@thatdocslady any suggestions how to deal with current docs vs the changes in this pull request until we are ready to merge this into master? A shared feature branch? Do you think we can for the time being manually sync changes to documentation on master to such a branch? An other approach could be to take the current docs and put them for example to GitHub Pages and refer to them in the README. Once this is done we can merge this as it is and everyone can work on the same documents. Once we have the new docs hosted on doc.openshift.org we can then change the pointer. WDYT?

> Also, I love Middleman so I'm happy.

+1 I like it as well. I used Awestruck a bit, but I think it has several glitches and not such a big user base. Middleman worked out of the box and gives us all we need. 

> AsciiBinder isn't necessary for our local staging if we accept vanilla asciidoc preview.

Even w/o, it seems to me that AsciiBinder is for one particular use case and does not fit anything else.

> We will provide OpenShift with adoc files anyway, so I'm happy with whatever local preview tooling that works within the container.

We need make sure this is true. Now we are actually at a stage where we can try some things in terms of integration. We should hook up with them and try this out. Maybe they have a test instance where we can try  to integrate.

thatdocslady:
@hferentschik ok this makes sense. I will ping you on IRC if I run into trouble with testing the doc build.

The split between usage and reference is not because of any technical requirements, but it's a best practice for documentation (industry and RH), content chunking and modular doc libraries are important for information breakdown, so the plan was always to have Welcome>Getting Started>Using>Developing>Reference. Each of these "parts" would be a TOC entry and under each would be the relevant documents. I hope this makes sense.

AFA issue to open, I see we have issues for:

- Make build target for HTML #139 
- Switch to adoc (not just the usage, but all manually authored docs) #140 

I will create separate issues for:
- Refactoring the README
- OpenShift sync

I will send an email to OpenShift letting them know this is coming and how they usually pull in adoc files from other repos. They already have something like this in their Origin repo: https://bluejeans.com/mariel

Regarding the branch management, I think a shared feature branch would make it much easier for me (and others) to test the changes (vs. your fork for example), and if we can get the PR fixed up and merged in a few days (up to 1 week) I don't foresee many doc issues that would need to be manually synced. We can send an email to the list to ask the engineers to submit docs as separate PRs and not merge them if they have stuff coming in the next week or so. WDYT?

hferentschik:
Here is what I would like to do instead of working on a feature branch. I pushed the current docs we have (effectively the *.md files to Minishift's GitHub Pages). See

* https://minishift.io/docs/managing-minishift.html
* https://minishift.io/docs/minishift.html

Really the behaviour should be identical as if viewed as part of the repo itself (please try and verify).

As part of this pull request, I'll update the links in the README as well to point to the links in minishift.io. This allows us then to merge this pull request to master and all work together on creating the new site. Once we are all happy and the new site is up and running on docs.openshift.org we just the links in the README.

I will also update the layout to have the following directories: "getting-started, "using" , "developing", and "command-ref"

budhrg:
> As part of this pull request, I'll update the links in the README as well to point to the links in minishift.io. This allows us then to merge this pull request to master and all work together on creating the new site. Once we are all happy and the new site is up and running on docs.openshift.org we just the links in the README.

Sounds good to me.

> I will also update the layout to have the following directories: "getting-started, "using" , "developing", and "command-ref"

+1, I believe we are putting 'troubleshooting' doc inside developing. 

Any plan for roadmap to show as separate page ?

hferentschik:
> +1, I believe we are putting 'troubleshooting' doc inside developing.

Not sure, my initial thinking was "using"

> Any plan for roadmap to show as separate page ?

ROADMAP, I would expect to stay in the GitHub repo together with README and CONTRIBUTING, etc. Again, open for discussion. README and CONTRIBUTING need to stay, ROADMAP could be moved. I think, if so, I'd add it to the root of the site together with the yet to be added landing page.

Preeticp:
+1 for putting troubleshooting in 'using' since we are troubleshooting the usage of Minishift.

My 2 cents, Roadmap would probably let potential community members know what we are aiming at and where they could contribute, so makes sense to move it.

thatdocslady:
Ok so here's what I'm thinking for file structure: After we extract non-project (or "user-facing") info from the README, it will remain in the root GH repo. The contributing doc and the roadmap doc should also stay there because they are considered "project" docs and not "user-facing" (same as license, for example). We can link from the minishift docs (in some "additional resources" section wherever we put it) and from the README, but generally those types of docs should be in github because they are not product-specific.

thatdocslady:
And +100 on the github pages solution if it means we can merge to master and then move forward with content-specific items.

LalatenduMohanty:
> As part of this pull request, I'll update the links in the README as well to point to the links in minishift.io. This allows us then to merge this pull request to master and all work together on creating the new site. Once we are all happy and the new site is up and running on docs.openshift.org we just the links in the README.

@hferentschik I am not if I understood correctly.  Will there is any content left of minishift.io after the generated html pages move to docs.minishift.org? or it will again just a redirect as it was before?

hferentschik:
>  Will there is any content left of minishift.io after the generated html pages move

Up to us. But most likely we remove it

> or it will again just a redirect as it was before?

It still does redirects. That has not changed. It's just that a sub directory hosts the docs. These URLs are not re-directed.

hferentschik:
@LalatenduMohanty @budhrg @thatdocslady @Preeticp - I pushed an update which creates a site with the discussed directory layout. I also updated CONTRIBUTING.md and README.md to point to https://minishift.io/docs. So imo we could actually go ahead and merge this. 

Navigation of the docs on https://minishift.io/docs is suboptimal of course, since there is no nav bar, nor enough inter connecting links. But that's really the same right now. One thing worth doing might be to add a index page under https://minishift.io/docs. I think this might help a bit and people sometimes get to the root of things, by removing parts of the URL. In our case accessing /docs would then redirect to openshift.org.

Depending on how long the conversion takes and when we have a hosted site on docs.openshift.org, we can discuss whether we are going to push updates to minishift.io/docs. 

Thoughts?

@thatdocslady, @Preeticp - are you guys able to build the site now?

hferentschik:
@budhrg did you encounter any issues getting it to run?

hferentschik:
FYI, if you really want awesome integration, run the Middleman server directly on the host. Then you can see changes appearing in the browser as you go along. Awesome. When using the container approach, it works as well, but with a quite long delay, since change detection is base on polling in this case.

thatdocslady:
@hferentschik I'm still getting the stupid missing docker error but if everyone else is able to test it then let's merge and I'll try to solve my snowflake problems later.

hferentschik:
> I'm still getting the stupid missing docker error 

Did you run:

in the shell from which you run the build? 

> but if everyone else is able to test

Let's see the overall feedback on this

rkratky:
@hferentschik, ad the split between Reference and Use docs: My only reasoning was that it seemed logical. The manually created (User) docs describe installation, first steps, etc. The automatically generated (Reference) docs are basically manual pages. However, I don't have a strong opinion on the subject -- this really is up to Mikey.

hferentschik:
> ad the split between Reference and Use docs: My only reasoning was that it seemed logical. The manually created (User) docs describe installation, first steps, etc. The automatically generated (Reference) docs are basically manual pages. However, I don't have a strong opinion on the subject -- this really is up to Mikey.

+1 Got you. We kept the split and even introduced more directories. Now it is  "getting-started, "using" , "developing", and "command-ref"

Preeticp:
> @thatdocslady, @Preeticp - are you guys able to build the site now?
@hferentschik finally was able to build the site with the help of @budhrg, tricky places were the Go dependencies, getting the path set and having the Docker daemon working.
At first glance:
1. Looks good and neat, yellow background I guess would change once it is in the final destination.
2. Commands and their outputs have got numbered (Managing Minishift>Persistent configuration)
3. Table renders well, but no lines demarcating boundaries.
4. Navigation wise, Adding breadcrumbs, an Index page along with the relevant portions of the Readme would be the next steps.
All these can be fixed through the style sheet I guess.

hferentschik:
> Looks good and neat, yellow background I guess would change once it is in the final destination.

There are two aspects to it. In its final destination the docs.openshift.org style applies. The generated HTML is completely free of
any styling. As a pointed out somewhere earlier, it would be nice to get the docs.openshift.org styling into our sire as well, since it
gives a better feel how things will look at in the end. However, this is not a must.

> Commands and their outputs have got numbered (Managing Minishift>Persistent configuration)

That's a style things

> Table renders well, but no lines demarcating boundaries.

Style

> Navigation wise, Adding breadcrumbs, an Index page along with the relevant portions of the Readme would be the next steps.

Here we need the landing page and if we get the docs.openshift.org styles, I take it that we get their navigation and bread crumbs as well.

rkratky:
@hferentschik, the PR works nicely -- just tried it. Apologies for stalling on this for so long -- I should've let you go ahead with it in the first place. It would've been a lot faster (not to mention more elegant). I still think AB deserved a try, but this is just fine.

Minor nitpick: could we use a Fedora/CentOS-based image as the basis? You know, dogfooding...

hferentschik:
> the PR works nicely -- just tried it.

Awesome, thanks for trying

> Apologies for stalling on this for so long -- I should've let you go ahead with it in the first place.

No problem. We did not have time for it anyways.

> I still think AB deserved a try, but this is just fine.

Well, I am trying to understand how this even would work. I actually did try, but AsciiBinder requiring a "special" git repo seems to make it only suitable for the cases where the repo is a pure docs repo. Copying files into a tmp directory, to then create a git repo is just not working, not even speaking of trying live reload to work. The moment you copy the sources to a temp directory, you would need to edit these sources and not the sources in the actual repo anymore. I am happy to have a discussion with some AsciiBinder expert on how this is would work in the case where the docs are a sub-folder of a wider project.

> Minor nitpick: could we use a Fedora/CentOS-based image as the basis? You know, dogfooding...

Fedora is notoriously bad when it comes to Ruby. I rather have a base image is designed to work with Ruby.

hferentschik:
@LalatenduMohanty please go ahead and merge if you have time.

docs/Dockerfile
```diff
@@ -0,0 +1,30 @@
+# Copyright 2016 The Kubernetes Authors All rights reserved.
```
LalatenduMohanty: I think the license header is worng here isn't it? It should be Red Hat license header.

prerelease.sh
```diff
@@ -39,19 +39,3 @@ if [[ ! -z "${files}" ]]; then
...
-else
-  make gendocs
-  files=$(git diff)
-  if [[ $files ]]; then
-    echo "Help text is out of date: $files \n Please run 'make gendocs'\n and make sure that those doc changes are committed"
-    exit 1
-  else
-    echo "Help text is up to date"
-  fi
-fi
```
LalatenduMohanty: @hferentschik Are we removing this check because we have the docs in adoc?

LalatenduMohanty: Did we move the check to some other place? I considered this check was a good thing.

LalatenduMohanty:
@hferentschik PR LGTM except I have couple of questions. Also `make gen_docs` worked for me.

prerelease.sh
hferentschik: >  Are we removing this check because we have the docs in doc?

No

> Did we move the check to some other place? 

No

> I considered this check was a good thing.

In a scenario where you build the docs and deploy it on a server, this check does not make any sense. The docs will be build from scratch, so the files are per definition in sync.

hferentschik:
Thanks for merging :-)

PR #539, opened 2017-03-13
Issue #531 Integration test to verify MACHINE_NAME.json

### LINKED ISSUE #531: Add integration test to verify MACHINE_NAME.json
Since integration setup in place, add integration test for MACHINE_NAME.json. 

Followup of PR https://github.com/minishift/minishift/pull/521.

budhrg:
Fix #531

### DIFF
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -31,7 +31,9 @@ import (
 	"github.com/DATA-DOG/godog"
 	"github.com/DATA-DOG/godog/gherkin"
 
+	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/test/integration/util"
+	"path/filepath"
 )
 
 var lastCommandOutput CommandOutput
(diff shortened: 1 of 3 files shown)


test/integration/util/util.go
```diff
@@ -41,6 +42,10 @@ type OcRunner struct {
 	CommandPath string
 }
 
+const (
+	OpenShiftVersion = "v1.4.1" // Should match OPENSHIFT_VERSION in Makefile
```
hferentschik: -1, this creates dependencies which will be hard to track and lead to failing tests as we use different versions. It also prevents us from running the test against other versions of OpenShift.

test/integration/features/basic.feature
```diff
@@ -7,6 +7,7 @@ Feature: Basic
     Given executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
+      And Minishift should have instance config file created
```
hferentschik: I think if we want this, we should wait for pull request #540. BTW, you do much more in _shouldHaveInstanceConfig_ than verifying that the config file exist. You also verify its content. This is not apparent here.

test/integration/basic_test.go
```diff
@@ -66,6 +68,21 @@ func (m *Minishift) shouldHaveAValidIPAddress() error {
 	return nil
 }
 
+func (m *Minishift) shouldHaveInstanceConfig() error {
```
hferentschik: I am not sure whether his really should be a function of Minishift.

hferentschik:
I am going to close this pull request in this form. It feels wrong to have this here, especially since in our harness itself we already rely on the config file existing - https://github.com/minishift/minishift/blob/master/test/integration/util/util.go#L137

test/integration/features/basic.feature
budhrg: @hferentschik How about two test case here ? One verify instance config file creation and one verify its content. I am thinking about these two lines:
```
And Minishift should have instance config file created
And Minishift config file should have appropriate content
```

test/integration/util/util.go
budhrg: Then, we can have regex pattern checking to verify the content of json file for oc path as

`MACHINE_HOME_DIR/cache/oc/v(some regex pattern)/oc`

test/integration/basic_test.go
budhrg: We can extract it to public util function. Also, it can't be function of MinishiftRunner

budhrg:
@hferentschik 
> I am going to close this pull request in this form. It feels wrong to have this here, especially since in our harness itself we already rely on the config file existing

Does it mean, we don't need any test to verify MACHINE_NAME.json ?

hferentschik:
> Does it mean, we don't need any test to verify MACHINE_NAME.json ?

I don't think so

Issue #532, opened 2017-03-10
Split scenario in basic.feature into several smaller scenarios


agajdosi:
basic.feature currently uses one big scenario which covers whole process of using Minishift - it goes from starting minishift, checking ssh functionality, checking docker environmental variables, stopping to deleting minishift. Splitting it into smaller scenarios gives us the possibility to finish other scenarios when one of them fails, currently all steps are skipped when something fails. But more importantly we will be able to tag scenarios, so only part of basic.feature can be run e.g. tag `@fast` for quick developer tests, `@full` for automated or QE use.

Smaller scenarios are currently pre-indicated by blank lines between groups of scenario steps.

hferentschik:
Merged via pull request #540

PR #540, opened 2017-03-13
Issue #532 Splitting huge scenario into several shorter scenarios

agajdosi:
Big main scenario split into few smaller scenarios respecting blank lines.

### DIFF
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -159,7 +159,7 @@ func FeatureContext(s *godog.Suite) {
 
 	m := &Minishift{runner: runner}
 
-	s.Step(`Minishift should have state "([^"]*)"`, m.shouldHaveState)
+	s.Step(`Minishift (?:has|should have) state "([^"]*)"`, m.shouldHaveState)
 	s.Step(`Minishift should have a valid IP address`, m.shouldHaveAValidIPAddress)
 	s.Step(`executing "minishift ([^"]*)"`, m.executingCommand)
 	s.Step(`executing "oc ([^"]*)`, m.executingOcCommand)
@@ -169,15 +169,14 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`([^"]*) should equal`, m.commandReturnShouldEqualContent)
 	s.Step(`([^"]*) should be empty`, m.commandReturnShouldBeEmpty)
 
-	s.BeforeScenario(func(interface{}) {
+	s.BeforeSuite(func() {
 		testDir := setUp()
 		fmt.Println("Running Integration test in : ", testDir)
 	})
 
-	s.AfterScenario(func(interface{}, error) {
+	s.AfterSuite(func() {
 		m.runner.EnsureDeleted()
 	})
-
 }
 
 func TestMain(m *testing.M) {
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -3,14 +3,16 @@ Feature: Basic
   As a user
   I need to be able to bring up a test environment
 
-  Scenario: Basic test scenario
-    Given executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
+  Scenario: Starting Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
 
-     # Check if developer have sudo permission then oc should output as below with other
-     # cluster roles
-     # sudoer   /sudoer   developer
+  Scenario: OpenShift developer account has sudo permissions
+     Check if developer have sudo permission then oc should output as below with other
+     cluster roles
+     sudoer   /sudoer   developer
      When executing "oc --as system:admin get clusterrolebindings"
      Then stderr should be empty
      And  exitcode should equal 0
@@ -19,6 +21,8 @@ Feature: Basic
      sudoer
      """
 
+  Scenario: User able to do ssh into machine via Minishift
+    Given Minishift has state "Running"
      When executing "minishift ssh echo hello"
      Then stderr should be empty
       And exitcode should equal 0
@@ -27,6 +31,8 @@ Feature: Basic
       hello
       """
 
+  Scenario: User is able to set custom Docker specific environment variables
+    Given Minishift has state "Running"
      When executing "minishift ssh cat /var/lib/boot2docker/profile"
      Then stderr should be empty
       And exitcode should equal 0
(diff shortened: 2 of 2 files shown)


test/integration/features/basic.feature
```diff
@@ -3,14 +3,16 @@ Feature: Basic
...
+  Scenario: Starting Minishift
+    Given Minishift should have state "Does Not Exist"
+     When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
 
-     # Check if developer have sudo permission then oc should output as below with other
-     # cluster roles
-     # sudoer   /sudoer   developer
+  Scenario: OpenShift developer account has sudo permissions
```
budhrg: :+1:

budhrg:
:+1:  Like the idea of scenario based tests cases

budhrg:
add to whitelist

test/integration/basic_test.go
```diff
@@ -169,12 +169,12 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`([^"]*) should equal`, m.commandReturnShouldEqualContent)
 	s.Step(`([^"]*) should be empty`, m.commandReturnShouldBeEmpty)
 
-	s.BeforeScenario(func(interface{}) {
+	s.BeforeSuite(func() {
```
hferentschik: So what does this mean in terms of having multiple feature files? When is this run? 

@gbraad How is this going to work when we start having multiple feature files? Correct me from wrong, but we can have multiple test files in each of which we define _func FeatureContext(s *godog.Suite)_. As long as the file is annotated with _'// +build integration'_ it will add a new suite? So running the setup and teardown before/after suite would be ok for now. Would this be correct? 

@agajdosi Could you investigate this please?

If this is the case, should we not extract all these step definitions (or at least the ones which we want to share) and put them into a separate file _common_steps.go_ with a function _addCommonSteps(s *godog.Suite)_?

test/integration/features/basic.feature
```diff
@@ -3,14 +3,16 @@ Feature: Basic
   As a user
   I need to be able to bring up a test environment
 
-  Scenario: Basic test scenario
-    Given executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
+  Scenario: Starting Minishift
+    Given Minishift should have state "Does Not Exist"
```
hferentschik: We need a new step definition here or adjust the existing one to match also on "is in state". 
"Given" is a precondition which should be "Minishift [has|is in] state "Does Not Exist". "should have" makes no sense in this context.

test/integration/features/basic.feature
```diff
@@ -3,14 +3,16 @@ Feature: Basic
   As a user
   I need to be able to bring up a test environment
 
-  Scenario: Basic test scenario
-    Given executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
+  Scenario: Starting Minishift
```
hferentschik: 👍 for the way you go about scenarios. I like this.

test/integration/features/basic.feature
```diff
@@ -19,6 +21,8 @@ Feature: Basic
      sudoer
      """
 
+  Scenario: Minishift is able to do ssh into machine
```
hferentschik: " user is able to do ssh into machine" (Minishift is sshing)

test/integration/features/basic.feature
```diff
@@ -27,6 +31,8 @@ Feature: Basic
       hello
       """
 
+  Scenario: Minishift --docker-env works properly
```
hferentschik: "A user can set custom Docker specific environment variables"

hferentschik: BTW, this is a bit unfortunate, since this scenario depends on the fact how Minshift was started earlier. If it were not so expensive to start/stop Minishift, this really should go into its own feature/suite. 

Moving forward, I could see that we have a docker.feature, were we test various other things around Docker, eg pushing to the exposed OpenShift registry. In this case I think it makes sense to move this scenario into the Docker feature.

test/integration/features/basic.feature
```diff
@@ -35,8 +41,12 @@ Feature: Basic
       export "FOO=BAR"
       export "BAZ=BAT"
       """
-
+  Scenario: Stopping Minishift
+    Given Minishift should have state "Running"
```
hferentschik: As in all "Given" clauses, it should be "Minishift [has|is in] state "Running""

hferentschik:
@agajdosi btw, you need to run `make fmt` to make sure the code complies with the coding style. Also please prefix the commit message with 'Issue #xyz'

test/integration/features/basic.feature
agajdosi: agree to that, actually I had it there but then changed it back. I think that probably all "shoulds" could be removed.

test/integration/basic_test.go
agajdosi: @hferentschik This is run at the start and at the end of feature, not sure with the rest for now.
I agree to your suggestion with common_steps.go, this will save us from repeating.

test/integration/features/basic.feature
agajdosi: @hferentschik One another thing is that in future we will probably have more scenarios on minishift start. For example stopped -> running, running -> running. These are different test cases, so "minishift has state" is needed.

test/integration/features/basic.feature
agajdosi: @hferentschik I agree best placement for this test will be in docker.feature in future and that proper scenario for this begins with minishift start and then check if env is there. Until then it can be here I think even if it does not fit into feature well. Your naming 

> "A user can set custom Docker specific environment variables"

seems fine to me.

gbraad:
Tried to add comments last night, but Github refused this... and today the same. So instead:


Why do you remove this?
```
-		fmt.Println("Running Integration test in : ", testDir)
```

gbraad:
@hferentschik About the first question. I have used the beforeSuite in early tests, but ran into issues at that moment with the test and moved to using the beforeScenario. The issue I was experiencing could be related to the fact of still transitioning from partially using Testing to godog only. This has to be tried again... especially when several feature files exist and they all mention a beforeSuite, what will happen? This should be extracted into a general main.go which has the Testing runner and the beforeSuite definition.

gbraad:
> " user is able to do ssh into machine" (Minishift is sshing)

Agree, it is not Minishift which is doing the action... but the user.

test/integration/features/basic.feature
budhrg: Yes, we had done similar thing in ADB for service-manager tests cases - https://github.com/projectatomic/vagrant-service-manager/tree/master/features

agajdosi:
@gbraad Thank you for this catch, it is mistake from my side.

test/integration/features/basic.feature
agajdosi: @hferentschik found out my reply disappeared somehow. I am wondering if we also want to remove should from for example this:

     Then stderr should be empty
      And exitcode should equal 0
      And stdout should contain

hferentschik:
Rebased and merged, thanks.

PR #543, opened 2017-03-13
Issue #359 Add context as minishift

praveenkumar:
This patch will include context as `minishift` by default.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -192,6 +192,10 @@ func runStart(cmd *cobra.Command, args []string) {
 		glog.Errorln("Error adding developer user to sudoers", err)
 		atexit.Exit(1)
 	}
+	if err := openshift.AddContextForProfile(constants.MachineName, ip, "developer", "myproject"); err != nil {
+		glog.Errorln("Error adding OpenShift Context", err)
+		atexit.Exit(1)
+	}
 }
 
 // Set Docker Proxy
--- a/pkg/minishift/openshift/oc.go
+++ b/pkg/minishift/openshift/oc.go
@@ -17,6 +17,10 @@ limitations under the License.
 package openshift
 
 import (
+	"fmt"
+	"strings"
+
+	"github.com/minishift/minishift/pkg/minikube/constants"
 	instanceState "github.com/minishift/minishift/pkg/minishift/config"
 	"github.com/minishift/minishift/pkg/util"
 )
@@ -37,6 +41,28 @@ func AddSudoersRoleForUser(user string) error {
 		return err
 	}
 	cmdArgs = []string{"login", "-u", user}
+	if _, err := runner.Output(cmdName, cmdArgs...); err != nil {
+		return err
+	}
+	return nil
+}
+
+// Add Current Profile Context
+func AddContextForProfile(profile string, ip string, username string, namespace string) error {
+	cmdName := instanceState.Config.OcPath
+	ip = strings.Replace(ip, ".", "-", -1)
+	cmdArgs := []string{"config", "set-context", profile,
+		fmt.Sprintf("--cluster=%s:%d", ip, constants.APIServerPort),
+		fmt.Sprintf("--user=%s/%s:%d", username, ip, constants.APIServerPort),
+		fmt.Sprintf("--namespace=%s", namespace),
+	}
+
+	if _, err := runner.Output(cmdName, cmdArgs...); err != nil {
+		return err
+	}
+
+	cmdArgs = []string{"config", "use-context", profile}
+
 	if _, err := runner.Output(cmdName, cmdArgs...); err != nil {
 		return err
 	}
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -15,11 +15,50 @@ Feature: Basic
      sudoer   /sudoer   developer
      When executing "oc --as system:admin get clusterrolebindings"
      Then stderr should be empty
-     And  exitcode should equal 0
-     And  stdout should contain
+      And exitcode should equal 0
+      And stdout should contain
      """
      sudoer
      """
+
+  Scenario: Current Context should be minishift
+    Check the context is set as per profile
+    In our case profile is minishift
+    When executing "oc config current-context"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+    """
+    minishift
+    """
+
+  Scenario: Create a dummy context
+    When executing "oc config set-context dummy"
+    Then stderr should be empty
+     And exitcode should equal 0
+
+  Scenario: Switch to dummy context
+    When executing "oc config use-context dummy"
+    Then stderr should be empty
+     And exitcode should equal 0
+
+  Scenario: Check project info for dummy context
+    When executing "oc project -q"
+    Then exitcode should equal 1
+
+  Scenario: Switch context to minishift
+    When executing "oc config use-context minishift"
+    Then stderr should be empty
+     And exitcode should equal 0
+
+  Scenario: Check project info for minishift context
+    When executing "oc project -q"
+    Then stderr should be empty
+     And exitcode should equal 0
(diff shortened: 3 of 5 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -192,6 +192,10 @@ func runStart(cmd *cobra.Command, args []string) {
 		glog.Errorln("Error adding developer user to sudoers", err)
 		atexit.Exit(1)
 	}
+	if err := openshift.AddContextForProfile(constants.MachineName, ip); err != nil {
```
praveenkumar: This will set context as profile name instead hard coded `minishift` once we will have `MachineName` as env then user can get different context for different profiles.

pkg/minishift/openshift/oc.go
```diff
@@ -42,3 +46,24 @@ func AddSudoersRoleForUser(user string) error {
...
 	return nil
 }
+
+// Add Current Profile Context
+func AddContextForProfile(profile string, ip string) error {
+	cmdName := instanceState.Config.OcPath
+	ip = strings.Replace(ip, ".", "-", -1)
+	cmdArgs := []string{"config", "set-context", profile,
+		fmt.Sprintf("--cluster=%s:%d", ip, constants.APIServerPort),
+		fmt.Sprintf("--user=developer/%s:%d", ip, constants.APIServerPort),
```
praveenkumar: Here `developer` is hard coded but not the profile and each profile is going to have a user developer by default.

cmd/minishift/cmd/start.go
```diff
@@ -192,6 +192,10 @@ func runStart(cmd *cobra.Command, args []string) {
...
+		glog.Errorln("Error adding OpenShift Context", err)
```
budhrg: I think we can improve this error message to include profile name.
```
glog.Errorf("Error adding %s Context: %s", constants.MachineName, err)
```

pkg/minishift/openshift/oc.go
```diff
@@ -42,3 +46,24 @@ func AddSudoersRoleForUser(user string) error {
...
+		"--namespace=myproject"}
```
budhrg: Why namespace is hardcorded here? We can have different namespace right?

pkg/minishift/openshift/oc.go
budhrg: Not matching with name of function since few things are hardcoded.
I would suggest to pass, `username` and `namespace` to this method and use those values. This will help to make this method general to use. 

Then, this method can be called as `AddContextForProfile("minishift", "some_ip", "user_name", "namespace")

WDYT ?

cmd/minishift/cmd/start.go

docs/interacting-with-openshift.md
```diff
@@ -36,6 +36,14 @@ in order to use `oc`, for example:
 In future versions we will provide a command to assist in setting up the `PATH`.
 See GitHub issue [#142](https://github.com/minishift/minishift/issues/142).
 
+The minishift start command creates a `oc context` called `minishift`. This context contains the configuration to
```
hferentschik: 👍 Nice

pkg/minishift/openshift/oc.go
hferentschik: I am wondering the same thing. Is the _namespace_ really required? Can it not be omitted?

hferentschik: Would it make sense to run first:

    $ oc config set-cluster minishift --server=https://192.168.99.105:8443

hferentschik: For what it's worth, the following works for me and does not rely on a namespace:

    $ oc config set-context hardy --cluster=192-168-99-106:8443 --user=developer/192-168-99-106:8443

Also, looking at `oc config view`, it seems the --cluster is referring to the cluster name, which if you look at the config is:

     - cluster:
        certificate-authority-data: REDACTED
        server: https://192.168.99.106:8443
      name: 192-168-99-106:8443

So the name is uses actually hyphens not an actual IPs with dots. This is also what one should refer to in the various commands (unless of course one also set the cluster name as mentioned in the previous comment). 

My vote would be to create a 'minishift' cluster alias as well and then use this name to create the context. Also the actual minishift context does imo not need a namespace.

test/integration/features/basic.feature
```diff
@@ -19,6 +19,16 @@ Feature: Basic
      sudoer
      """
 
+     # check the context is set as per profile
+     # In our case profile is minishift
```
hferentschik: You might want to switch context as well (we can use one of the preexisting context we know does exist) and then back to _minishift_ and then run an actual command.

hferentschik:
@praveenkumar regarding docs, you will need to rebase on master and apply your doc changes to the new doc file.

pkg/minishift/openshift/oc.go
praveenkumar: > My vote would be to create a 'minishift' cluster alias as well and then use this name to create the context.

I would go with profile alias and then use same alias to create context.

praveenkumar:
>  regarding docs, you will need to rebase on master and apply your doc changes to the new doc file.

@hferentschik  done.

hferentschik:
@praveenkumar, did you see the comments on pull request #558? Especially this one https://github.com/minishift/minishift/pull/558#issuecomment-287703534? I think we need to do a bit more here when it comes to our usage of _oc_.

cmd/minishift/cmd/start.go
```diff
@@ -192,6 +192,10 @@ func runStart(cmd *cobra.Command, args []string) {
 		glog.Errorln("Error adding developer user to sudoers", err)
 		atexit.Exit(1)
 	}
+	if err := openshift.AddContextForProfile(constants.MachineName, ip, "developer"); err != nil {
```

hferentschik: Nice also that is is calling a method outside the command structure.

pkg/minishift/openshift/oc.go
```diff
@@ -42,3 +46,23 @@ func AddSudoersRoleForUser(user string) error {
 	}
 	return nil
 }
+
+// Add Current Profile Context
+func AddContextForProfile(profile string, ip string, username string) error {
+	cmdName := instanceState.Config.OcPath
+	ip = strings.Replace(ip, ".", "-", -1)
```
hferentschik: :+1:, right now the _--cluster_ parameter will target the right cluster nickname

test/integration/features/basic.feature
```diff
@@ -21,6 +21,17 @@ Feature: Basic
      sudoer
      """
 
+  Scenario: Current Context should be minishift
```
hferentschik: Nice. The only thing I am wondering about is whether we should extend on the testing. Right now you only check whether initially the context is _minishift_.

We could extend this to test the whole flow. After verifying that the current context is _minishift_ and maybe that the current project (oc project -q) is myproject, we write a new scenario "User can switch context". 

In this scenario we create a dummy context `oc config set-context fubar ` and switch to it `oc config use-context fubar`. Now a `oc project -q` should return an error.

Then in the final scenario "User can switch back to minishift context", we do a `oc config use-context minishift` followed by `oc project myproject` which should now work again, right? WDYT?

praveenkumar: @hferentschik Agree, let me extend it. was thinking something different which didn't make sense but this does.

test/integration/features/basic.feature
```diff
@@ -15,12 +15,51 @@ Feature: Basic
...
+    In our case profile is minishift
+    When executing "oc config current-context"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+    """
+    minishift
+    """
+
+  Scenario: Create a dummy context
```
hferentschik: Hmm, not each step is a scenario per se.

hferentschik:
Rebased and merged, thanks. I re-wrote the Cucumber test a bit. Not each command execution is a scenario.

PR #558, opened 2017-03-16
[WIP] Issue #342 Fix service command

budhrg:
- Use oc binary to fetch route and open it in browser
- Remove k8s dependency in service and service list files

### Execution log
```
$ minishift openshift service nationalparks-katacoda
Opening the service /nationalparks-katacoda in the default browser...

$ minishift openshift service nationalparks-katacoda -n myproject
Opening the service myproject/nationalparks-katacoda in the default browser...

$ minishift openshift service nationalparks-katacoda -n foo      
Namespace foo doesn't exits
```

Early feedback/review appreciated.

cc @hferentschik @LalatenduMohanty

### DIFF
--- a/cmd/minishift/cmd/openshift/service.go
+++ b/cmd/minishift/cmd/openshift/service.go
@@ -19,39 +19,25 @@ package openshift
 import (
 	"fmt"
 	"os"
-	"strings"
-	"text/template"
 
-	"github.com/docker/machine/libmachine"
 	"github.com/pkg/browser"
 	"github.com/spf13/cobra"
 
-	"github.com/minishift/minishift/pkg/minikube/cluster"
-	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/openshift"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 )
 
 var (
-	namespace          string
-	serviceURLMode     bool
-	serviceURLFormat   string
-	serviceURLTemplate *template.Template
-	https              bool
+	namespace string
+	urlMode   bool
+	https     bool
 )
 
 // serviceCmd represents the service command
 var serviceCmd = &cobra.Command{
 	Use:   "service [flags] SERVICE",
-	Short: "Prints the URL for the specified service to the console.",
-	Long:  `Prints the URL for the specified service to the console.`,
-	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		t, err := template.New("serviceURL").Parse(serviceURLFormat)
-		if err != nil {
-			fmt.Fprintln(os.Stderr, "The URL format specified in the --format option is not valid: \n\n", err)
-			atexit.Exit(1)
-		}
-		serviceURLTemplate = t
-	},
+	Short: "Open the URL for the specified service in the browser.",
+	Long:  `Open the URL for the specified service in the browser.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(args) == 0 || len(args) > 1 {
 			fmt.Fprintln(os.Stderr, "You must specify the name of the service.")
@@ -60,22 +46,13 @@ var serviceCmd = &cobra.Command{
 
 		service := args[0]
 
-		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
-		defer api.Close()
-
-		url, err := cluster.GetServiceURL(api, namespace, service, serviceURLTemplate)
+		url, err := openshift.GetServiceURL(service, namespace, https)
 		if err != nil {
 			fmt.Fprintln(os.Stderr, err)
-			if _, ok := err.(cluster.MissingNodePortError); !ok {
-				fmt.Fprintln(os.Stderr, "Verify that Minishift is running and that the correct namespace is specified in the -n option.")
-			}
 			atexit.Exit(1)
 		}
 
-		if https {
-			url = strings.Replace(url, "http", "https", 1)
-		}
-		if serviceURLMode {
+		if urlMode {
 			fmt.Fprintln(os.Stdout, url)
 		} else {
 			fmt.Fprintln(os.Stdout, "Opening the service "+namespace+"/"+service+" in the default browser...")
@@ -85,9 +62,8 @@ var serviceCmd = &cobra.Command{
 }
 
 func init() {
-	serviceCmd.Flags().StringVarP(&namespace, "namespace", "n", "default", "The namespace of the service.")
-	serviceCmd.Flags().BoolVar(&serviceURLMode, "url", false, "Access the service in the command-line console instead of the default browser.")
-	serviceCmd.PersistentFlags().StringVar(&serviceURLFormat, "format", "http://{{.IP}}:{{.Port}}", "The URL format of the service.")
+	serviceCmd.Flags().StringVarP(&namespace, "namespace", "n", "", "The namespace of the service.")
+	serviceCmd.Flags().BoolVar(&urlMode, "url", false, "Access the service in the command-line console instead of the default browser.")
 	serviceCmd.Flags().BoolVar(&https, "https", false, "Access the service with HTTPS instead of HTTP.")
 	OpenShiftConfigCmd.AddCommand(serviceCmd)
 }
--- a/cmd/minishift/cmd/openshift/service_list.go
+++ b/cmd/minishift/cmd/openshift/service_list.go
@@ -18,16 +18,8 @@ package openshift
 
 import (
 	"fmt"
-	"os"
 
-	"github.com/docker/machine/libmachine"
-	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
-	"k8s.io/kubernetes/pkg/api/v1"
-
-	"github.com/minishift/minishift/pkg/minikube/cluster"
-	"github.com/minishift/minishift/pkg/minikube/constants"
-	"github.com/minishift/minishift/pkg/util/os/atexit"
 )
 
 var serviceListNamespace string
@@ -38,30 +30,12 @@ var serviceListCmd = &cobra.Command{
 	Short: "Gets the URLs of the services in your local cluster.",
 	Long:  `Gets the URLs of the services in your local cluster.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
-		defer api.Close()
-		serviceURLs, err := cluster.GetServiceURLs(api, serviceListNamespace, serviceURLTemplate)
-		if err != nil {
-			fmt.Fprintln(os.Stderr, err)
-			fmt.Fprintln(os.Stderr, "Check that Minishift is running and that the correct namespace is specified in the -n option if it is required.")
-			atexit.Exit(1)
-		}
-
-		var data [][]string
-		for _, serviceURL := range serviceURLs {
-			data = append(data, []string{serviceURL.Namespace, serviceURL.Name, serviceURL.URL})
-		}
-
-		table := tablewriter.NewWriter(os.Stdout)
-		table.SetHeader([]string{"Namsepace", "Name", "URL"})
-		table.SetBorders(tablewriter.Border{Left: true, Top: true, Right: true, Bottom: true})
-		table.SetCenterSeparator("|")
-		table.AppendBulk(data) // Add Bulk Data
-		table.Render()
+		// TODO
+		fmt.Println("In Service List")
 	},
 }
 
 func init() {
-	serviceListCmd.Flags().StringVarP(&serviceListNamespace, "namespace", "n", v1.NamespaceAll, "The namespace of the services.")
+	serviceListCmd.Flags().StringVarP(&serviceListNamespace, "namespace", "n", "", "The namespace of the services.")
 	serviceCmd.AddCommand(serviceListCmd)
 }
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -17,7 +17,6 @@ limitations under the License.
 package cluster
 
 import (
-	"bytes"
 	"encoding/json"
 	"flag"
 	"fmt"
@@ -27,7 +26,6 @@ import (
 	"os"
 	"path/filepath"
 	"strings"
-	"text/template"
 	"time"
 
 	"github.com/docker/machine/drivers/virtualbox"
@@ -42,9 +40,6 @@ import (
 	minishiftUtil "github.com/minishift/minishift/pkg/minishift/util"
 	"github.com/minishift/minishift/pkg/util"
 	pb "gopkg.in/cheggaaa/pb.v1"
-	kubeapi "k8s.io/kubernetes/pkg/api"
-	"k8s.io/kubernetes/pkg/client/unversioned"
-	"k8s.io/kubernetes/pkg/client/unversioned/clientcmd"
 )
 
 var (
@@ -54,10 +49,6 @@ var (
 
 const (
 	fileScheme = "file"
-
-	serviceAPIAnnotationsPrefix = "api.service.kubernetes.io/"
-	serviceAPIAnnotationScheme  = serviceAPIAnnotationsPrefix + "scheme"
-	serviceAPIAnnotationPath    = serviceAPIAnnotationsPrefix + "path"
 )
 
 //This init function is used to set the logtostderr variable to false so that INFO level log info does not clutter the CLI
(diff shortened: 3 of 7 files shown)


cmd/minishift/cmd/openshift/service.go
```diff
@@ -19,39 +19,27 @@ package openshift
...
 )
 
 var (
-	namespace          string
-	serviceURLMode     bool
-	serviceURLFormat   string
-	serviceURLTemplate *template.Template
-	https              bool
+	namespace      string
+	serviceURLMode bool
```
gbraad: URLMode a boolean? means something can be toggled... do not see the naming

cmd/minishift/cmd/openshift/service.go
```diff
@@ -63,31 +51,23 @@ var serviceCmd = &cobra.Command{
...
 
-		if https {
-			url = strings.Replace(url, "http", "https", 1)
-		}
 		if serviceURLMode {
 			fmt.Fprintln(os.Stdout, url)
 		} else {
 			fmt.Fprintln(os.Stdout, "Opening the service "+namespace+"/"+service+" in the default browser...")
-			browser.OpenURL(url)
+			browser.OpenURL("http://" + url)
```
gbraad: http hardcoded? routes can be defined as secured...
https://docs.openshift.com/enterprise/3.0/dev_guide/routes.html

cmd/minishift/cmd/openshift/service.go
```diff
@@ -63,31 +51,23 @@ var serviceCmd = &cobra.Command{
...
 		}
 	},
 }
 
 func init() {
-	serviceCmd.Flags().StringVarP(&namespace, "namespace", "n", "default", "The namespace of the service.")
+	serviceCmd.Flags().StringVarP(&namespace, "namespace", "n", "", "The namespace of the service.")
 	serviceCmd.Flags().BoolVar(&serviceURLMode, "url", false, "Access the service in the command-line console instead of the default browser.")
-	serviceCmd.PersistentFlags().StringVar(&serviceURLFormat, "format", "http://{{.IP}}:{{.Port}}", "The URL format of the service.")
-	serviceCmd.Flags().BoolVar(&https, "https", false, "Access the service with HTTPS instead of HTTP.")
```
gbraad: I think we want this...

cmd/minishift/cmd/openshift/service.go
gbraad: I would rather call this more meaningful: openServiceURLInBrowser  (InConsole)

pkg/minishift/openshift/oc.go
```diff
@@ -42,3 +47,38 @@ func AddSudoersRoleForUser(user string) error {
 	}
 	return nil
 }
+
+// Get the route for service
+func GetServiceURL(service, namespace string) (string, error) {
+	cmdName := instanceState.Config.OcPath
+	cmdArgText := fmt.Sprintf("get route/%s", service)
+	if namespace != "" && !isProjectExists(namespace) {
```
gbraad: No need to check also the namespace here...
likely bail out when namespace == ""

or is an empty namespace something we can query?

pkg/minishift/openshift/oc.go
```diff
@@ -42,3 +47,38 @@ func AddSudoersRoleForUser(user string) error {
...
+		return "", errors.New(fmt.Sprintf("Namespace %s doesn't exits", namespace))
+	}
+	if namespace != "" {
```
gbraad: seems not... so empty is bad...

return/fail earlier

pkg/minishift/openshift/oc.go
```diff
@@ -42,3 +47,38 @@ func AddSudoersRoleForUser(user string) error {
...
+
+	var route routeApi.Route
+	json.Unmarshal(cmdOut, &route)
+
+	return route.Spec.Host, nil
+}
+
+// Check whether project exists or not
+func isProjectExists(project string) bool {
+	cmdName := instanceState.Config.OcPath
```
gbraad: return false when project == ""

gbraad:
BTW, the dependency is gone? but it is still needed in `glide.yaml`?

budhrg:
@gbraad  

> BTW, the dependency is gone? but it is still needed in glide.yaml?

There are other places where k8s is used like here https://github.com/minishift/minishift/blob/master/pkg/minikube/kubeconfig/config.go#L26-L28 .

pkg/minishift/openshift/oc.go
budhrg: @gbraad here empty namespace is also a valid case.

Eg

minishift openshift service <name>     will use current namespace (which is default to empty string).
However, if user want to open app from different namespace/project then he need to specify namespace then this `if` statement check that and add `-n namespace` to command.

cmd/minishift/cmd/openshift/service.go
budhrg: @gbraad `urlMode`  might be good fit here.

See comment https://github.com/minishift/minishift/issues/342#issuecomment-286039538 .

```
$ minishift openshift service nationalparks-katacoda -n myproject --url
http://nationalparks-katacoda-myproject.192.168.99.100.nip.io/

```

cmd/minishift/cmd/openshift/service.go
budhrg: We agreed here https://github.com/minishift/minishift/issues/342#issuecomment-286039062 to drop these flags since these doesn't make sense at the moment to our use case.

cmd/minishift/cmd/openshift/service.go
gbraad: I meant the naming inside the code is meaningless (it is not clear from the name what the intent/usage is). using the `--url` flag forces `--urlmode`... which means, the display result in console?

http://www.itiseezee.com/?p=83

budhrg: Actually, this naming thing is already there for [`minishift console`](https://minishift.io/docs/interacting-with-openshift.html#console) 
If we had used `dashboard` instead of `console` for command then we could have used `console` across minishift to refer printing in shell/console. 

Also, if we are changing the name here anyway, then we need to do across the code for consistency.

WDYT @gbraad @LalatenduMohanty @hferentschik @praveenkumar ?

```
$ minishift console   # Open Openshift UI

$ minishift console --url   #  Print url in console
```

budhrg:
I am assuming, for integration testing, I need to deploy application application first, need to wait for some time (bit tricky here), expose route and verify the service command behavior.

Any thoughts @hferentschik ?

I am thinking on these lines as
```
  Scenario: Deploying an application
    Verify if developer can able to deploy same application from github
    url https://github.com/openshift/nodejs-ex.
      When executing "oc new-app https://github.com/openshift/nodejs-ex"
      And I wait for <> time                           # need to check this condition
      And executes "oc expose svc/nodejs-ex"
      And executes "minishift openshift service nodejs-ex --url"
      Then stdout should match "http://nodejs-ex-myproject.<REGEX FOR IP>.nip.io"   
```

hferentschik:
> I am assuming, for integration testing, I need to deploy application application first, need to wait for some time (bit tricky here), expose route and verify the service command behaviour.

-1 for waiting. As we discussed, this is never a good strategy. If you want to wait until the build completes the easiest is to tail the build (`oc logs -f bc/nodejs-ex`). This command will return once the build is complete at which time you know the app is ready. 

There is a little gotcha here (which is a feature request in Origin as well). When you tail the build config, directly after you triggered the build and the s2i image is not there, the command will return directly. Really, one would like the command to wait, from the moment the build gets created. Anyways, so there might be a wait needed (for now), but it should not be a wait for the overall app to be available and hopefully this would also be only temporary.

That said, if you are just after checking that the service url is correct, there is not need to wait for the build to complete. You can expose the service directly after you issues the app creation command. You don't have to wait for the build to complete. Also, for getting the service url, the service does not need to be deployed yet. Different, if you want to also hit the app to make sure it works. Then you need to wait for the app to be deployed.

hferentschik:
$ minishift openshift service nationalparks-katacoda
    Opening the service /nationalparks-katacoda in the default browser...

Why would this not show a namespace? It should be the same output as the following command:

    $ minishift openshift service nationalparks-katacoda -n myproject
    Opening the service myproject/nationalparks-katacoda in the default browser...

provided the current namespace is _myproject_. As discussed earlier, it is probably best to quickly query the current namespace (_oc project -q_) and then drive the rest of the code identical to the case where the namespace is specified explicitly. This makes the ux more consistent.

cmd/minishift/cmd/openshift/service_list.go
```diff
@@ -38,30 +30,12 @@ var serviceListCmd = &cobra.Command{
...
-		}
-
-		table := tablewriter.NewWriter(os.Stdout)
-		table.SetHeader([]string{"Namsepace", "Name", "URL"})
-		table.SetBorders(tablewriter.Border{Left: true, Top: true, Right: true, Bottom: true})
-		table.SetCenterSeparator("|")
-		table.AppendBulk(data) // Add Bulk Data
-		table.Render()
+		// TODO
+		fmt.Println("In Service List")
```
hferentschik: :-)

pkg/minikube/cluster/cluster.go
```diff
@@ -462,165 +453,8 @@ type ipPort struct {
 	Port int
 }
 
-func GetServiceURL(api libmachine.API, namespace, service string, t *template.Template) (string, error) {
```
hferentschik: 👍 for deleting everything we don't need and ties to the the Kube API

pkg/minishift/openshift/oc.go
```diff
@@ -17,8 +17,13 @@ limitations under the License.
 package openshift
 
 import (
+	"encoding/json"
+	"errors"
+	"fmt"
 	instanceState "github.com/minishift/minishift/pkg/minishift/config"
 	"github.com/minishift/minishift/pkg/util"
+	routeApi "github.com/openshift/origin/pkg/route/api"
```
hferentschik: Do we really need to import this. We are pretty much free of dependencies to the OpenShift and probably could remove it.

pkg/minishift/openshift/oc.go
```diff
@@ -42,3 +47,47 @@ func AddSudoersRoleForUser(user string) error {
...
+
+	// json filter
+	cmdArgText += " -o json"
+	cmdOut, err := runner.Output(cmdName, strings.Split(cmdArgText, " ")...)
+	if err != nil {
+		return "", err
+	}
+
+	var route routeApi.Route
+	json.Unmarshal(cmdOut, &route)
```
hferentschik: I am not sure whether extracting the host is worth this OpenShift import. I'd either extract it from the default stdout output, or if json, then I'd unmarshal into a map or create my own struct to parse this single value.

test/integration/basic_test.go
```diff
@@ -161,13 +173,14 @@ func FeatureContext(s *godog.Suite) {
 
 	s.Step(`Minishift (?:has|should have) state "([^"]*)"`, m.shouldHaveState)
 	s.Step(`Minishift should have a valid IP address`, m.shouldHaveAValidIPAddress)
-	s.Step(`executing "minishift ([^"]*)"`, m.executingCommand)
-	s.Step(`executing "oc ([^"]*)`, m.executingOcCommand)
+	s.Step(`execut[es|ing]+ "minishift ([^"]*)"`, m.executingCommand)
+	s.Step(`execut[es|ing]+ "oc ([^"]*)`, m.executingOcCommand)
```
hferentschik: 'executes'?

test/integration/features/basic.feature
```diff
@@ -5,49 +5,57 @@ Feature: Basic
...
+    And stdout should contain
+    """
+    export "FOO=BAR"
+    export "BAZ=BAT"
+    """
+
+  Scenario: User is able to get url on running service command
+    Given Minishift has state "Running"
+    When executing "oc new-app https://github.com/openshift/nodejs-ex"
+    And executes "oc expose svc/nodejs-ex"
```
hferentschik: I don't think this is grammatically correct. How come you want to say now 'executes' instead of 'executing'?

test/integration/features/basic.feature
```diff
@@ -5,49 +5,57 @@ Feature: Basic
...
+    And executes "minishift openshift service nodejs-ex --url"
+    And stdout should match /nodejs-ex-myproject.(.*).nip.io/
```
hferentschik: Here you would have your 'Then'

test/integration/features/basic.feature
gbraad: not right-aligned anymore?

test/integration/basic_test.go
gbraad: It should be present continuous tense (+ing)... and keep consistency (as AND is part of the same tense)

test/integration/basic_test.go
```diff
@@ -161,13 +173,14 @@ func FeatureContext(s *godog.Suite) {
...
 	s.Step(`([^"]*) should contain ([^"]*)`, m.commandReturnShouldContain)
 	s.Step(`([^"]*) should contain`, m.commandReturnShouldContainContent)
 	s.Step(`([^"]*) should equal ([^"]*)`, m.commandReturnShouldEqual)
 	s.Step(`([^"]*) should equal`, m.commandReturnShouldEqualContent)
 	s.Step(`([^"]*) should be empty`, m.commandReturnShouldBeEmpty)
+	s.Step(`([^"]*) should match /([^"]*)/`, m.commandReturnShouldMatch)
```
gbraad: 'should match'  and 'should equal' are similar in meaning... so,why the need for different implementations?

test/integration/basic_test.go
```diff
@@ -106,14 +106,22 @@ func compareExpectedWithActualContains(expected string, actual string) error {
...
 		return fmt.Errorf("Output did not match. Expected: %s, Actual: %s", expected, actual)
 	}
 
 	return nil
 }
 
+func findExpectedMatchInActual(expectedMatch, actual string) error {
+	r, _ := regexp.Compile(expectedMatch)
+	if !r.MatchString(actual) {
+		return fmt.Errorf("Actual output %s doesn't contain match %s", actual, expectedMatch)
```
gbraad: 'ExpectedMatchInActual' => 'match in' means the same as 'contain'

and then 'match' and 'contain match' are not the same thing... ??? is it not just contain???

test/integration/basic_test.go
gbraad: For consistency, what about the Content function... ?

test/integration/features/basic.feature
```diff
@@ -5,49 +5,57 @@ Feature: Basic
...
-      And exitcode should equal 0
-      And stdout should contain
-      """
-      export "FOO=BAR"
-      export "BAZ=BAT"
-      """
+    When executing "minishift ssh cat /var/lib/boot2docker/profile"
+    Then stderr should be empty
+    And exitcode should equal 0
+    And stdout should contain
```
gbraad: Then...

Please, keep right-aligned for readability!

```
    When executing "minishift ssh cat /var/lib/boot2docker/profile"
      Then stderr should be empy
        And exitcode should equal 0
        And stdout should contain
```

test/integration/features/basic.feature
```diff
@@ -5,49 +5,57 @@ Feature: Basic
...
   Scenario: User is able to do ssh into machine via Minishift
     Given Minishift has state "Running"
-     When executing "minishift ssh echo hello"
-     Then stderr should be empty
-      And exitcode should equal 0
-      And stdout should contain
-      """
-      hello
-      """
+    When executing "minishift ssh echo hello"
```
gbraad: Right-aligned

test/integration/features/basic.feature
```diff
@@ -5,49 +5,57 @@ Feature: Basic
...
-     cluster roles
-     sudoer   /sudoer   developer
-     When executing "oc --as system:admin get clusterrolebindings"
-     Then stderr should be empty
-     And  exitcode should equal 0
-     And  stdout should contain
-     """
-     sudoer
-     """
+    Check if developer have sudo permission then oc should output as below with other
```
gbraad: Right-algined

test/integration/features/basic.feature
```diff
@@ -5,49 +5,57 @@ Feature: Basic
...
+    cluster roles
+    sudoer   /sudoer   developer
```
gbraad: Wat did you change here?

test/integration/features/basic.feature
```diff
@@ -5,49 +5,57 @@ Feature: Basic
 
   Scenario: Starting Minishift
     Given Minishift has state "Does Not Exist"
-     When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
-     Then Minishift should have state "Running"
+      When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
```
gbraad: Out of alignment???

cmd/minishift/cmd/openshift/service.go
```diff
@@ -19,39 +19,25 @@ package openshift
...
 
 var (
-	namespace          string
-	serviceURLMode     bool
-	serviceURLFormat   string
-	serviceURLTemplate *template.Template
-	https              bool
+	namespace string
+	urlMode   bool
+	https     bool
```
gbraad: +1 :-)

pkg/minishift/openshift/oc.go
```diff
@@ -42,3 +47,47 @@ func AddSudoersRoleForUser(user string) error {
 	}
 	return nil
 }
+
+// Get the route for service
+func GetServiceURL(service, namespace string, https bool) (string, error) {
+	urlScheme := "http://"
+	if https {
```
gbraad: prefer this simpler approach than the original string.replace 👍

test/integration/basic_test.go
hferentschik: :+1:

test/integration/basic_test.go
hferentschik: :+1:

hferentschik:
@praveenkumar, @budhrg - there is another thing, you guys need to think through and address in the context you are working. The sudoers change gives us the ability to run oc as admin w/o changing the current context in the Kube config. And issue #359 will give us some added convenience when working with login context. However, that does not stop the oauth token to expire or the user to log out manually. In this case running _oc_ commands will fail. What is our strategy there? There are several options:

- Seemingly simplest is to transparently login the developer user again. The problem with this approach is that once we introduce the ability to create users and have proper passwords, this is not going to work anymore
- Check which user is currently logged in and if it is the 'developer' user, do a transparent login, otherwise ask the user to re-login via the _oc_ CLI
- When no user is logged or the oauth token is expired, we always ask the user to re-login.

To me the latter is the most consistent approach, since it does not make any assumptions (a user could change the authentication scheme for and give the 'developer' user a proper password, in which case option #1 and #2 will fail. Interested to hear your thoughts around this.

Either way, the handling of running an _oc_ command which fails due to no active login context should be handled by our helper methods running _oc_ commands.

BTW, what is the current time the auth token is valid?

hferentschik:
Thinking even more about it, there is another case we need to consider. What if a user uses _oc_ to log in to another OpenShift cluster and while he is connected to this cluster runs a minishift commands, which in turn calls out to _oc_. In this case we query the wrong cluster. This ties into issue #359 I think. When running an _oc_ command we might have to first check what the current context is and if it is not the minishift context, we need to either temporarily switch to the minishift context or we need to abort the command with a warning. 

Alternatively we maybe have to investigate again, whether we have somehow the ability to run _oc_ commands against Minishift's OpenShift cluster w/o changing the current login context in the users' Kube config.

hferentschik:
To guarantee that we do not interfere with the user, we would need to copy the relevant config data (for system:admin and the minishift cluster) to the Minishift home directory. Then we would run each 'oc' command with the '--config' option. This would guarantee Minishift's _oc_ usage will always be independent of the user's usage.

hferentschik:
So, here is what I think should happen:

- `minishift start` is executed and starts the bootstrap
- VM is created
- _cluster up_ provisions OpenShift
- we create a _minishift_ context (as well as a _minishift_cluster_ alias for the cluster) - see issue 359 and pr #543 
- we copy (or re-create) the needed entries for using the Minishift OpenShift cluster as system:admin from _~/.kube/config_ to _MINISHIFT_HOME/machines/<machine-name\>/_
- all _oc_ commands executed from within Minishift use this internal configuration file (_--config_ flag)
- sudoer permissions are created and any other customisations run 

The impact for this particular pull request is, that _oc_ commands will be executed as _system:admin_ which means, the default namespace will be _default_ and one needs to specify '-namespace myproject' explicitly.

Thoughts?

budhrg:
> we copy (or re-create) the needed entries for using the Minishift OpenShift cluster as system:admin from ~/.kube/config to MINISHIFT_HOME/machines/<machine-name>/

@hferentschik , I am guessing `MINISHIFT_HOME/machines/<machine-name>` is docker-machine managed directory?

Are you saying this config? `MINISHIFT_HOME/machines/<machine-name>.json`?

hferentschik:
> Are you saying this config? MINISHIFT_HOME/machines/<machine-name>.json?

We need a separate directory and config file for that

test/integration/features/basic.feature
```diff
@@ -5,49 +5,57 @@ Feature: Basic
...
+    And executes "oc expose svc/nodejs-ex"
+    And executes "minishift openshift service nodejs-ex --url"
+    And stdout should match /nodejs-ex-myproject.(.*).nip.io/
 
   Scenario: Stopping Minishift
     Given Minishift has state "Running"
-     When executing "minishift stop"
-     Then Minishift should have state "Stopped"
+    When executing "minishift stop"
+    Then Minishift should have state "Stopped"
```
gbraad: ditto...

test/integration/features/basic.feature
```diff
@@ -5,49 +5,57 @@ Feature: Basic
...
 
   Scenario: Deleting Minishift
     Given Minishift has state "Stopped"
-     When executing "minishift delete"
-     Then Minishift should have state "Does Not Exist"
+    When executing "minishift delete"
```
gbraad: ditto...

test/integration/features/basic.feature
gbraad:
```
     Then stdout should match /nodejs-ex-myproject.(.*).nip.io/
```

budhrg:
Thanks @gbraad , @hferentschik for reviews. Closing it since will open new PR due to some pre-requisite as mentioned in this comment https://github.com/minishift/minishift/pull/558#issuecomment-287703534

test/integration/basic_test.go
budhrg: @gbraad ,

I think `should match` is needed to match some expression which I think should not be implemented from `should equal` as both have different meanings/implementaion.

gbraad: It is mostly about the wording... This comments was given before doing the whole reviews was concluded.

the meaning of "match" and "equal" are too close that from the phrasing alone it is not easy to distinguish the actual use-case.

test/integration/basic_test.go
budhrg: @gbraad , @hferentschik , `match` is totally different from `contain`.

`match` do regex match whereas `contain` check if it is part of string or not.
`contain` doesn't able to perform regex match.

https://play.golang.org/p/g2dY9piGQK

If we are changing the implementation of `compareExpectedWithActualContains` then IMO we are creating confusion by having two behavior handle by it.

Issue #571, opened 2017-03-20
Failing builds due to unresolvable glide.sh


gbraad:
While building on CI the following errors occur:

CentOS-CI: Connection to 172.19.3.110 closed. during the compilation (when glide needs to be installed)
Travis-CI: The command "GLIDE_TAG=$(curl -s https://glide.sh/version)" failed and exited

The domain http://glide.sh is not available: https://github.com/Masterminds/glide/issues/784

This can ofcourse always occur, but maybe it is better to rely on only larger operated services during a build process/install procedure.

Since we are using the releases, why not pin the GLIDE_TAG? This will result in a more predictable outcome anyways.

Analysis:
  * centos-ci: tagged version using releases
  * travis-ci: tagged version using releases
  * appveyor: go get
  * circl-ci: go get

I believe we should have a preferred way for all platforms, and preferably pinned to a version.

Note:
On Fedora I have been using the packaged versions anyways, Golang is 1.7.5 and Glide is also recent enough. In this case: 
```
$ dnf install -y golang glide
```
Will suffice to perform a build. However, packaged versions for CentOS are available after enabling epel-release, but they are slightly older (but in my opinion workable enough). Anyways, if the version is pinned to a specified version, or to the packaged version, the outcome is more predictable.

praveenkumar:
I would say we should use same approach for glide installation (go get), whatever CI we are using, this way we are sure that our all CI have same version because if we use rpm/deb package for centos/travis CI then might be we don't get what master have and hit inconsistent results.

gbraad:
agree. I also believe the go get method seems to be the better approach.

hferentschik:
> I would say we should use same approach for glide installation (go get), whatever CI we are using

+1

hferentschik:
Merged via pull request #572

gbraad:
will prepare another PR for the `go get` method of installation

hferentschik:
>will prepare another PR for the go get method of installation

The problems with 'go get'  is that it will always fetch HEAD. This is actually not optimal, since we are exposing us of getting non reproducible builds. Setting a fixed version of Glide is imo the better approach.  

I would like to align Glide setup approaches across our CI setups, but preferably in a way we've done here - with a explicitly set Glide version.

Windows still needs some special treatment though, since there we are running into a problems where we exceed the maximum path length.

gbraad:
[...]

PR #572, opened 2017-03-20
Issue #571 pin glide version for CI

gbraad:


praveenkumar:
If we can have same way to get tag version for other 2 CI (appveyor and circleci) that would be nice I am sure we can have for circleci because for osx we have wget but not sure about appveyor otherwise I would say let have `go get` for every CI just sake of consistency.

gbraad:
`go get` can work, but also the tagged version we could. for appveyor we just need to implement this logic with powershell. I would say to accept this solution for the moment and then work on an improvement.

hferentschik:
Merged, thanks

PR #576, opened 2017-03-20
Issue #562 - Adding BINPATH argument to 'make integration'

### LINKED ISSUE #562: Add flag or make new target to avoid compiling binary while running integration test
Right now `make integration` compiles the binary and use it to verify integration test.

One might need to test the local binary or download from URL for integration tests.

This IMO can be achieved by two ways:
- create flag like URL=`some url` to use url, like `make integration URL=<url>`
- create flag like BINARY=`path to binary`, `make integration BINARY=<path to binary>`
- new target altogether

Any thoughts?

agajdosi:
To be able to select custom binary on which tests will be started there is BINPATH argument for make created. To run make integration on already existing binary, run "make integration BINPATH=<path-to-binary>". In this case new build will not be triggered. When no BINPATH is selected, integration will build new binary and run on it as it is currently implemented.

Fixes #562.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,7 @@ REPOPATH ?= $(ORG)/minishift
 ifeq ($(GOOS),windows)
 	IS_EXE := .exe
 endif
+MINISHIFT_BINARY ?= $(GOPATH)/src/$(REPOPATH)/out/$(GOOS)-$(GOARCH)/minishift$(IS_EXE)
 
 LDFLAGS := -X $(REPOPATH)/pkg/version.version=$(VERSION) \
 	-X $(REPOPATH)/pkg/version.isoVersion=$(ISO_VERSION) \
@@ -103,8 +104,8 @@ test: vendor $(GOPATH)/src/$(ORG)
 	@go test -v $(shell $(PACKAGES))
 
 .PHONY: integration
-integration: $(BUILD_DIR)/$(GOOS)-$(GOARCH)/minishift$(IS_EXE)
-	go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration -v
+integration: $(MINISHIFT_BINARY)
+	go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration -v -args --binary $(MINISHIFT_BINARY)
 
 .PHONY: fmt
 fmt:
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -23,7 +23,6 @@ import (
 	"fmt"
 	"net"
 	"os"
-	"runtime"
 	"strconv"
 	"strings"
 	"testing"
@@ -151,7 +150,8 @@ func (m *Minishift) commandReturnShouldBeEmpty(commandField string) error {
 
 func FeatureContext(s *godog.Suite) {
 	var givenArgs = flag.String("minishift-args", "", "Arguments to pass to minishift")
-	var givenPath = flag.String("binary", fmt.Sprintf("../../out/%s-amd64/minishift", runtime.GOOS), "Path to minishift binary")
+	var givenPath = flag.String("binary", "", "Path to minishift binary")
+	flag.Parse()
 
 	runner := util.MinishiftRunner{
 		CommandArgs: *givenArgs,
@@ -171,7 +171,8 @@ func FeatureContext(s *godog.Suite) {
 
 	s.BeforeSuite(func() {
 		testDir := setUp()
-		fmt.Println("Running Integration test in : ", testDir)
+		fmt.Println("Running Integration test in: ", testDir)
+		fmt.Println("using binary: ", *givenPath)
 	})
 
 	s.AfterSuite(func() {


Makefile
```diff
@@ -103,8 +103,13 @@ test: vendor $(GOPATH)/src/$(ORG)
 	@go test -v $(shell $(PACKAGES))
 
 .PHONY: integration
+ifdef BINPATH
```
hferentschik: I think this can be solved more elegantly by introducing a variable 'MINISHIFT_INTEGRSTION_TEST_BINARY' (be clear with the naming).

We then can define at the top of the Makefile after the respective use variables are defined:

    MINISHIFT_INTEGRSTION_TEST_BINARY ?= $(BUILD_DIR)/$(GOOS)-$(GOARCH)/minishift$(IS_EXE)

Then instead of an if statement in the integration target we have:

    go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration -v -args --binary $(MINISHIFT_INTEGRSTION_TEST_BINARY)

The '?=' in the variable declaration will make sure that the default is used if the varible is not defined on the command line, otherwise the command line value is used.

test/integration/basic_test.go
```diff
@@ -152,6 +152,7 @@ func (m *Minishift) commandReturnShouldBeEmpty(commandField string) error {
 func FeatureContext(s *godog.Suite) {
 	var givenArgs = flag.String("minishift-args", "", "Arguments to pass to minishift")
 	var givenPath = flag.String("binary", fmt.Sprintf("../../out/%s-amd64/minishift", runtime.GOOS), "Path to minishift binary")
```
hferentschik: Should we maybe remove the default here, making this a required argument. This wrong to use _../../out/%s-amd64/minishift_ at this point. WDYT?

test/integration/basic_test.go
```diff
@@ -171,7 +172,9 @@ func FeatureContext(s *godog.Suite) {
 
 	s.BeforeSuite(func() {
 		testDir := setUp()
-		fmt.Println("Running Integration test in : ", testDir)
+		fmt.Println("Running Integration test in: ", testDir)
```
hferentschik: We will need a proper logging framework at some stage. For now it should be ok.

Makefile
amitkrout: 👍

agajdosi: @hferentschik Thanks for tips, I really like your idea. I am just not sure with naming of variable - there is conflict between clarity and friendliness I think. For me `make integration MINISHIFT_INTEGRSTION_TEST_BINARY=~/minishift/minishift` feels unfriendly, also when we add tags support then it will be even longer. WDYT for example about `TEST_BINARY`?

test/integration/basic_test.go
agajdosi: @hferentschik Agree with you, when I implement your suggestion above this step is not required.

Makefile
hferentschik: Sure, let's shorten the name. What's about MINISHIFT_BINARY?

test/integration/basic_test.go

Makefile
agajdosi: @hferentschik Works for me. Going on it now.

test/integration/basic_test.go
agajdosi: @hferentschik just realized that make file and basic_test.go are in different folders, so there are troubles with using relative path. basic_test.go needs to have `../../out/linux-amd64/minishift` and make file needs `./out/linux-amd64/minishift`. Working solution for me is to define the path as absolute path by prepending $GOPATH: `MINISHIFT_BINARY ?= $(GOPATH)/src/$(REPOPATH)/out/$(GOOS)-$(GOARCH)/minishift$(IS_EXE)`

Makefile
```diff
@@ -24,6 +24,7 @@ REPOPATH ?= $(ORG)/minishift
 ifeq ($(GOOS),windows)
 	IS_EXE := .exe
 endif
+MINISHIFT_BINARY ?= $(GOPATH)/src/$(REPOPATH)/out/$(GOOS)-$(GOARCH)/minishift$(IS_EXE)
```
hferentschik: This is not a valid dependency. 

    make: *** No rule to make target `/Users/hardy/go/src/github.com/minishift/minishift/out/darwin-amd64/minishift', needed by `integration'.  Stop.

I guess we should use `$(GOPATH)/bin/minishift$(IS_EXE)`

hferentschik:
Merged with a change to the default value of _MINISHIFT_BINARY_. Thanks

Makefile
agajdosi: @hferentschik original path is: `$(BUILD_DIR)/$(GOOS)-$(GOARCH)/minishift$(IS_EXE)`, which is equal on linux to `./out/linux-amd64/minishift` and it starts from make file folder. When I tried to make it absolute path, I have added `$(GOPATH)/src/$(REPOPATH)` = `/home/agajdosi/golang/src/github.com/minishift/minishift`.

Issue #578, opened 2017-03-20
Add possibility to pass tags to godog via make integration target


agajdosi:
As test suite is going to grow there is need to organize tests in multiple feature files and to be able to select just some of them. Gherkin language implements tags for this case. We can have `@basic` tag on whole basic feature and nowhere else, which will secure that with basic tag only those few tests will run.
```
@basic
Feature: Basic
  In order to use Minishift
  As a user 
...
```
For some cases we can also use tags directly on selected scenarios, so for example `@openshift` tag can run 2 scenarios from basic.feature, 4 scenarios from another feature file etc.
```
 @openshift
 Scenario: OpenShift developer account has sudo permissions
     Check if developer have sudo permission then oc should output as below with other
     cluster roles
     sudoer   /sudoer   developer
```
Also tags could be useful when writing downstream or platform specific scenarios.

**Suggestion:**
- tags can be specified with TEST_TAGS argument: `make integration TEST_TAGS="openshift proxy docker"`
- implementation should accept multiple tags
- when nothing is specified for TEST_TAGS argument, tests will run with native `@basic` tag, meaning fast smoke testing will be executed

hferentschik:
Merged via pull request #584

PR #584, opened 2017-03-21
Issue #578 - Add tag support for godog tests into make integration

agajdosi:
To run integration tests on only specified features and scenarios, one can run e.g.: `make integration TEST_TAGS="openshift docker proxy"`.
When TEST_TAGS is not specified, then tests run with native tag: `basic`, which is bind to smoke test in basic.feature file. This means that when we add more time consuming feature files , one will be still able to run quick check with just `make integration`. Or run smoke test and feature close to specific issue he is dealing with. For example `make integration TEST_TAGS="smoke proxy"` when one is dev dealing with proxy stuff.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,7 @@ ifeq ($(GOOS),windows)
 	IS_EXE := .exe
 endif
 MINISHIFT_BINARY ?= $(GOPATH)/bin/minishift$(IS_EXE)
+GODOG_OPTS ?= ""
 PACKAGES := go list ./... | grep -v /vendor
 SOURCE_DIRS = cmd pkg test
 
@@ -143,7 +144,7 @@ test: vendor $(ADDON_ASSET_FILE)
 
 .PHONY: integration
 integration: $(MINISHIFT_BINARY)
-	go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration -v -args --binary $(MINISHIFT_BINARY)
+	go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration -v -args --binary $(MINISHIFT_BINARY) $(GODOG_OPTS)
 
 .PHONY: fmt
 fmt:
--- a/docs/source/developing/developing.adoc
+++ b/docs/source/developing/developing.adoc
@@ -209,11 +209,39 @@ For more information about test options, run the `go test --help` command and re
 ==== Integration Tests
 
 Integration tests utilize https://github.com/DATA-DOG/godog[`godog`], which uses Gherkin (Cucumber)
-to define test cases. The test cases are defined in `test/integration/*.feature`, and can be run with:
+to define test cases. The test cases are defined in `test/integration/*.feature`. To execute them on binary created by `make build` you 
+can simply run:
 
 ----
 $ make integration
 ----
+
+To run integration tests on minishift binary in different location you can use `MINISHIFT_BINARY` argument:
+
+----
+$ make integration MINISHIFT_BINARY=<path-to-custom-binary>
+----
+
+Additional properties for Godog runner can be specified with `GODOG_OPTS` argument. Following options are available:
+
+- `tags`: when used only scenarios and features containing at least one of selected tags will be executed.
+- `paths`: defines paths to different feature files or folders containing feature files. This can be used to run feature files
+outside of `test/integration/features` folder.
+- `format`:  you can change format of Godog's output, for example you can use `pretty` format instead of native `progress`.
+- `stop-on-failure`: set to true to stop integration tests on failure.
+- `no-colors`: set to true to disable ansi colors of Godog's output.
+- `definitions`: set to true to print all available step definitions.
+
+For example to run integration tests on two specific feature files, using only `@basic` and `@openshift` tags and without ansi colors, 
+following command can be used:
+
+----
+$ make integration GODOG_OPTS="-paths ~/tests/custom.feature,~/my.feature -tags basic,openshift -no-colors true"
+----
+
+Please note that when multiple values are used for options in GODOG_OPTS, then those have to be separated with comma without a whitespace.
+While `-tags basic,openshift` will be parsed properly by make, from `-tags basic, openshift` only `@basic` tag will be used.  
+
 
 [[format-source]]
 === Formatting the source
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -34,6 +34,7 @@ import (
 )
 
 var lastCommandOutput CommandOutput
+var givenArgs, givenPath string
 
 type CommandOutput struct {
 	Command  string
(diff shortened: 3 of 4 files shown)


Makefile
```diff
@@ -117,7 +118,7 @@ test: vendor
 
 .PHONY: integration
 integration: $(MINISHIFT_BINARY)
-	go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration -v -args --binary $(MINISHIFT_BINARY)
+	go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration -v -args --binary $(MINISHIFT_BINARY) --tags "$(TEST_TAGS)"
```
hferentschik: does that not force me to always specify a tag? Should there not be a way run all tests w/o specifying a tag? Trying to understand your thinking here.

test/integration/basic_test.go
```diff
@@ -181,11 +179,19 @@ func FeatureContext(s *godog.Suite) {
 }
 
 func TestMain(m *testing.M) {
+	flag.StringVar(&givenArgs, "minishift-args", "", "Arguments to pass to minishift")
+	flag.StringVar(&givenPath, "binary", "", "Path to minishift binary")
+	flag.StringVar(&givenTags, "tags", "", "Tags of tests to pass to godog")
+	flag.Parse()
+
+	givenTags = strings.Replace(givenTags, " ", ", ", -1)
```
hferentschik: Again, should this not be a bit more defensive? What if no tag is given? Will it run all or nothing?

test/integration/features/basic.feature
```diff
@@ -8,7 +9,7 @@ Feature: Basic
      When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT"
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
-
+
```
praveenkumar: This added only because of a whitespace so might be good to remove that space.

Makefile
```diff
@@ -25,6 +25,7 @@ ifeq ($(GOOS),windows)
 	IS_EXE := .exe
 endif
 MINISHIFT_BINARY ?= $(GOPATH)/bin/minishift$(IS_EXE)
+TEST_TAGS ?= basic
```
praveenkumar: @hferentschik so looks like if user doesn't define a tag during run of integration then it only consider to run basic tests but I do anticipate it should be running all the test instead of basic when no tags are defined.

test/integration/features/basic.feature
agajdosi: @praveenkumar yes, thank you for a catch

Makefile
agajdosi: @praveenkumar @hferentschik My idea here was to make it easy to use - having basic test as a smoke test which you can run while on coffee break etc. And to not force dev to type a lot: `make integration TEST_TAGS=basic`, I put basic test as native option there. However I agree with you that this is not so obvious setup. Let me know about this and I will update this PR to run all tests on default.

agajdosi:
@hferentschik Good point, I am going to add documentation on both new features of `make integration` in next update of this branch.

test/integration/basic_test.go
agajdosi: @hferentschik If no tags are given `make integration TEST_TAGS=""` then all tests are executed. Currently only problem I have found is this case: `make integration TEST_TAGS="basic openshift "` with extra space or comma at the end it fails.

hferentschik: A couple of things, I think this should be guarded with an if statement, if no tags are given the 'tag' argument should not even be set in the good.Options.

Secondly, why do we not pass the tags comma separated in the first place?

Makefile
hferentschik: So here is an idea, why don't we make this variable "CUCUMBER_OPTS" and make it per default the empty string. Then we pass the whole _CUCUMBER_OPTS_ to the _go test_ command. This means per default not tags are applied.

If you then want to run with tags, you do:

    $ make integration CUCUMBER_OPTS="--tags basic,smoke,whatever"

This might be also more generic in case we want to pass parameters.

agajdosi: @hferentschik working on it now. Only I would change CUCUMBER_OPTS to GODOG_OPTS as except tags and feature paths, all other options are relevant only for godog, for example StopOnFailure, NoColors etc.

docs/source/developing/developing.adoc
```diff
@@ -209,12 +209,40 @@ For more information about test options, run the `go test --help` command and re
...
+following command can be used:
+
+----
+$ make integration GODOG_OPTS="-paths ~/tests/custom.feature,~/my.feature -tags basic,openshift -no-colors true"
+----
+
+Please note that when multiple values are used for options in GODOG_OPTS, then those have to be separated with comma without a whitespace.
+While `-tags basic,openshift` will be parsed properly by make, from `-tags basic, openshift` only `@basic` tag will be used.  
+
+
```
agajdosi: Hi @Preeticp, I have tried to add documentation for new features I have added to `make integration` as proposed by @hferentschik. Can I ask you for a review of this? I have no experience with writing proper documentation so there are some mistakes I guess... Thankful for any suggestion. Just right now I have realized I have one extra line here.

hferentschik:
@agajdosi so you updated the pull request now, right? I guess we can take [WIP] off again.

agajdosi:
@hferentschik Yes, PR is updated. I have added docs into developing-integration as you suggested and created GODOG_OPTS argument in make, which supports most options available in godog Options. I dropped concurrency for which test suite is currently not prepared I guess.

Makefile
```diff
@@ -28,6 +28,7 @@ ifeq ($(GOOS),windows)
 	IS_EXE := .exe
 endif
 MINISHIFT_BINARY ?= $(GOPATH)/bin/minishift$(IS_EXE)
+GODOG_OPTS ?= ""
```
hferentschik: :+1:

docs/source/developing/developing.adoc
hferentschik: no worries. Looking good

test/integration/basic_test.go
```diff
@@ -181,11 +178,25 @@ func FeatureContext(s *godog.Suite) {
 }
 
 func TestMain(m *testing.M) {
+	flag.StringVar(&givenArgs, "minishift-args", "", "Arguments to pass to minishift")
```
hferentschik: Cool, now we can handle all potential flags. Personally I'd put the command line parsing into a separate function. But this will work.

docs/source/developing/developing.adoc
Preeticp: Sorry @agajdosi, just saw this, will review this.

hferentschik:
Merged, thanks.

@agajdosi, awesome work. Loving it. This is what I had in mind. I in particular like 'GODOG_OPTS="-format pretty"'.  I am tempted to make 'pretty' the default, but at the very least I would think it should be used on the CI servers, since one gets a better insight what's happening.

Also funny to see how much the setup "costs" and how fast the actual tests are. The bottle neck is the setup.

Issue #534, opened 2017-03-12
Review asciidoc documentation


hferentschik:
This covers the non generated markdown documentation

thatdocslady:
This is a duplicate of issue #140, please close one of the issues

hferentschik:
Pul request #530 already does a initial conversion to asciidoc. But it needs to be reviewed correct styling applied.

hferentschik:
Merged via pull request #591

Issue #533, opened 2017-03-12
Convert README to asciidoc and remove install/quickstart/docker-daemon info


hferentschik:
- Convert to Asciidoc
- Extract all content which is better suited for docs site

thatdocslady:
Following discussion on PR #621 it was decided to split the work on the README as follows:

I created a new issue #623  to address copying (not move) the "Installation", "Quickstart", and "Reusing the Docker daemon" sections from the README to new topics and place them in their final destinations as asciidoc files in the appropriate sub-directories. 

This is an interim step on the way to getting our docs published, since we cannot remove the information from the readme yet, but we also need to have a complete doc set to test in the coming days and to publish ASAP.

After the docs are published, this issue will address converting the README to asciidoc and removing the duplicate sections, replacing them with a general high-level getting started section and appropriate links to the documentation.

How does this sound? @LalatenduMohanty @Preeticp

LalatenduMohanty:
@thatdocslady +1. We need to change the original description of this issue as per the new plan.

Preeticp:
Sounds good, +1 @thatdocslady

hferentschik:
Merged via pull request #684

PR #591, opened 2017-03-22
Issue 534: Reviewed and edited asciidoc files

thatdocslady:
Fixes issue #534 

This PR handles the following tasks:
- Formatting and structuring of the newly-converted asciidoc against the [OpenShift docs guidelines](https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc)
- Adding index.adoc files to the root and sub-directories (removed the index.html.erb file because the index.adoc file replaced it, aligned with [OpenShift repo structure(https://github.com/openshift/openshift-docs) )
- Light proofread for writing styles and consistent voice

Out of scope:
- README content (will be handled in issue #533)
- Topic metadata and link/xref comparison (sent questions to OpenShift)


hferentschik:
@thatdocslady this pull request does not apply w/o merge conflicts

hferentschik:
> Adding index.adoc files to the root and sub-directories (removed the index.html.erb file because the index.adoc file replaced it,

-1 this is our entry point into the docs for now. We need it. The landing page is _overview.adoc_

hferentschik:
> The command ref docs are missing required topic metadata,

this is auto-generated. They will be added during the conversion from markdown to asciidoc

hferentschik:
> There are a bunch of .keep files, what is their purpose

git ignored empty directories. These files ensure that a specific directory stays part of the repo. For example the command-ref directory.

docs/source/developing/index.adoc
```diff
@@ -0,0 +1,14 @@
...
+:toc: macro
+:toc-title:
+:toclevels: 1
+
+toc::[]
+
+This section contains information about how to extend, build, and release Minishift, and
+how to contribute to various aspects of the project.
+
+- link:../developing/developing.html[Developing Minishift]
```
hferentschik: I don't get it. You are own the developing directory. The link is just './developing.html'. You are effectively going up one directory and then down again.

docs/source/developing/writing-docs.adoc
```diff
@@ -0,0 +1,48 @@
...
+----
+$ make serve_docs
+----
+
+The latter will start the https://middlemanapp.com[Middleman] server on
+port 4567. You can access the rendered documentation by browsing to
+http://<IP of Docker daemon>:4567.
+
+=== Deploying Minishift Documentation
+
```
hferentschik: The deploying the docs is here, but writing the docs in a separate doc?

docs/source/index.adoc
```diff
@@ -0,0 +1,31 @@
+[[minishift-index]]
+= Minishift
```
hferentschik: -1, the landing page is in overview.adoc. By creating this file you are destroying the entry point into the site for Middleman? How do you preview the site?

thatdocslady:
> @hferentschik : -1 this is our entry point into the docs for now. We need it. The landing page is overview.adoc

I was able to serve the docs without the index.html.erb file, Middleman simply took the index.adoc and served it instead. I wouldn't delete this file if it was needed to serve the docs. It's Middleman-specific and if we can serve the docs locally with the same file that we would use to publish then it puts us ahead of the game.

docs/source/index.adoc
```diff
@@ -0,0 +1,31 @@
...
+[[welcome-minishift]]
+== Welcome to Minishift
+
+Minishift is a tool that helps you run OpenShift locally by running a single-node OpenShift
+cluster inside a VM. You can try out OpenShift or develop with it, day-to-day, on your local host.
+
+This documentation library contains the following sections to help you get started with, use, and
+extend Minishift:
+
+- link:../getting-started/index.html[Getting started with Minishift]
```
hferentschik: Are these links working?

docs/source/index.html.erb
```diff
@@ -1,66 +0,0 @@
----
-title: Welcome to Middleman
```
hferentschik: -1 we need the era file. Please update the pull request accordingly

hferentschik:
> hardy, ok I think I will go with option 1 and do another pass on all the links using link format (not xref), and test everything locally.

+1 Let's make links work locally using 'link' for now. This also works in docs.openshift.org btw. To properly review the changes and also to continue working on the docs we need to keep this working. Whether and how we later change to xref needs to be discussed. 

> Just to clarify, I am not removing the index.erb file but it will be superceded by index.adoc

please _do_ remove index.erb. index.adoc and index.erb render into the same HTML document. So one needs to make a decision. I initially thought that we would keep index.erb for our local entry point and use overview.adoc for docs.openshift.org, however, as you said, a single index.adoc is probably better.

You will need to change this line - https://github.com/minishift/minishift/blob/master/docs/source/_topic_map.yml#L7 - though. this needs to be 'index'. You then can remove overview.adoc.

Note, that in index.adoc you should include the variables file and make sure you render somewhere the version of Minishift. It is normal that for the local site, the version won't display correctly and that there is a warning. It will work on docs.openshift.org though. This is due to a bug in AsciiBinder - https://github.com/redhataccess/ascii_binder/issues/60

The essential thing really is that we don't break the links for now. Does this make sense?

thatdocslady:
@hferentschik ok I understand all that now :) the site generates locally from index.adoc with the variable warning, all the links work, and all links use the format link instead of xref.

hferentschik:
> ok I understand all that now :) 

cool. I guess we talked past each other there for a while. By the way, I really like a single index file. When I was preparing the Makefile to create this tarball for integration into docs.openshift.org, I tried that (at least I thought so) and seen some errors. Hence I went down this path with overview vs index. 

> the site generates locally from index.adoc with the variable warning

Right, this is really unfortunate. But as said, this is a AscciiBinder bug and since there is no conditional include, we need to decide for which site it is going to work for now. Really best option for us would be to fix AsciiBinder, since everything else is just a hack/workaround something which should just work. 

>  all the links work

+1

> and all links use the format link instead of ref.

Right. This is in the end what we need to discuss with the docs.openshift.org team. I am confused with this xref format as per https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#internal-cross-references
AFAIK the way to do inter document linking across documents in adoc is via <<document-b.adoc#section-b,Section B>>. This would be my preferred choice, since it is also agnostic of the output. But even with xref, I don't understand why one would need to do this relative to the root of the topic map. This breaks portability of the documentation, since now the file depends on its location. For me this is the same issue as with the include. Another AscciiBinder bug?

thatdocslady:
@hferentschik yeah I don't know why this requirement exists for openshift but I see it with the link format as well (as in if I don't include the breadcrumb up and down it breaks the link).

By the way I'm still reviewing and fixing links in some topics so please hold on further link feedback.

hferentschik:
> By the way I'm still reviewing and fixing links in some topics so please hold on further link feedback.

np

thatdocslady:
@hferentschik finished fixing and testing the links, and sent an email to openshift with the open questions. I believe this PR is ready now for the official review and merge, and maybe we can fix whatever open issues come up in separate PRs?

hferentschik:
@thatdocslady have you tried the <<document-b.adoc#section-b,Section B>> syntax for cross referencing. This is the proper syntax and I should give us what we are after.

thatdocslady:
@hferentschik I used the {outfilesuffix} variable to avoid the whole adoc/html confusion, and we don't really have #section links that point to external file (i.e. the only cases we have these are where the links go to a different section of the same file). At any rate, the current link format parses correctly and matches what the openshift guidelines request, not sure why you suggest to change it.

hferentschik:
> I used the {outfilesuffix} variable to avoid the whole adoc/html confusion

that's good. I already using it in the generated files as well

> and we don't really have #section links that point to external file (i.e. the only cases we have these are where the links go to a different section of the same file) 

the <<>> is not only for inter-document linking, but also for cross referencing within the same document. This is Asciidoctor though. Not sure whether the idea is to stick with pure asciidoc syntax.

> At any rate, the current link format parses correctly and matches what the openshift guidelines request, not sure why you suggest to change it.

Because using <<>> manages also the creation of the link text. Atm you need to manually do that. Asciidoctor can create this for you, no need to keeping this in sync manually.

hferentschik:
Merged, thanks. Definitely a better entry into the site now. I just added an additional temporary note.png to avoid the broken image links. Eventually I'd really would like to get at least some basic docs.openshift.org L&F going.

Issue #551, opened 2017-03-15
Potentially missing docs about docker client prerequisite


thatdocslady:
While trying to reuse the docker daemon (for make gen_docs) I ran into this error:

https://paste.fedoraproject.org/paste/Tr6MGw6aAET24o5iwltrOl5M1UNdIGYhyRLivL9gydE=

It turns out that the docker client is needed in order to reuse the docker daemon from minishift. But our docs right now don't specify this prerequisite:

https://github.com/minishift/minishift#reusing-the-docker-daemon

@gbraad pointed out that this prerequisite might be an assumption that we make (as in we assume that people will already have it), but I think it's better to have it there than to risk people getting stuck (especially since this is required for *any* docker daemon reuse, and not just for doc builds).

hferentschik:
>  pointed out that this prerequisite might be an assumption that we make

IMO, it is very unlikely that anyone working in this space would not have the docker client. Also, for anyone working with containers it is clear that if there is a daemon and a client. So yes, there are assumptions here, but personally I don't think we need to be more explicit here. Where do you stop?

LalatenduMohanty:
@hferentschik @thatdocslady I think we should add this to documentation as we expect the user to get the docker client and minishift does not take care of downloading and putting the docker client in to the path.

thatdocslady:
Exactly, we specifically say that our developers can reuse the docker daemon so it makes it seem like they don't have to do anything, but similar to the hypervisor requirements and extra steps, for first-time users or people who want to try minishift this step is important and shouldn't be assumed.

hferentschik:
Two things : "say that our developers can reuse the docker daemon":

* "our developers" - exactly, imo we are targeting developers. Minishift is not a consumer tool, it is for developers. I think one can expect a minimal understanding of containers, Docker, etc from developers starting to think about container orchestration  
* "reuse the docker daemon" - I think for anyone working with containers it is clear what this means. Docker being a client server architecture, if one can re-use the daemon, that does not make the client go away

>  first-time users or people who want to try minishift this step is important

not at all. For them it is most likely irrelevant. This has really no tie to what Minishift otherwise offers.

thatdocslady:
Well if you feel so strongly about this then I won't fight about it, but let's see if we get any ricochets from not stating this once we start having users who need to use this.

gbraad:
> Where do you stop?

[...]

PR #595, opened 2017-03-23
Issue #551: Added instructions on docker client prerequisites

thatdocslady:
Fixes issue #551 

Related to issue #552 but not sure if can count as a "fix".

This PR addes prerequisites instructions on how to set up your local machine to reuse the docker daemon from Minishift.

NOTE: I based this mainly on the discussion that we had in issue #552 but I'm not 120% sure about some of the specific details, so would appreciate corrections where needed. I couldn't test this again because my machine is already set up so it's working already :-D

### DIFF
--- a/README.md
+++ b/README.md
@@ -218,18 +218,26 @@ When running OpenShift in a single VM, it is recommended to reuse the Docker dae
 for pure Docker use-cases as well.
 By using the same docker daemon as Minishift, you can speed up your local experiments.
 
-To be able to work with the docker daemon on your Mac or GNU/Linux host use the
-[`docker-env command`](https://minishift.io/docs/minishift_docker-env.html) in your shell:
+1. Make sure that you have the Docker client binary installed on your machine. For information about
+specific binary installations for your operating system, see the [Docker installation page](https://docs.docker.com/engine/installation/).
+
+1. Start Minishift with the [`minishift start`](https://minishift.io/docs/minishift_start.html) command.
+
+1. Use the [`minishift docker-env`](https://minishift.io/docs/minishift_docker-env.html) command
+to export the environment variables that are required to reuse the daemon.
 
 ```
 eval $(minishift docker-env)
 ```
 
 You should now be able to use _docker_ on the command line of your host, talking to the docker daemon
-inside the Minishift VM:
+inside the Minishift VM. To test the connection, run the following command:
+
 ```
 docker ps
 ```
+
+If successful, the shell will print a list of running containers.
 
 <a name="documentation"></a>
 ## Documentation


README.md
```diff
@@ -218,19 +218,45 @@ When running OpenShift in a single VM, it is recommended to reuse the Docker dae
 for pure Docker use-cases as well.
 By using the same docker daemon as Minishift, you can speed up your local experiments.
 
-To be able to work with the docker daemon on your Mac or GNU/Linux host use the
-[`docker-env command`](https://minishift.io/docs/minishift_docker-env.html) in your shell:
+1. Make sure that you have `docker` and `docker-client` installed on your machine.
```
hferentschik: -1 to this change in this form. We are targeting all three operating systems - Linux, OS X and Windows. This documentation is Linux specific and in fact even specific to a certain distribution (Fedora I believe). Package names are different on various distributions and also how you control services. 

For Windows and OS X this instructions don't apply and in fact these are the two main platforms where one most likely want to use Minishift's Docker daemon. On Linux is makes handly sense imo, since there I rather run Docker natively. Different story on OS X and Windows where I need a VM to run Docker. In this case, it makes sense to "save" running an additional instance and use Minishift's Docker daemon instead. 

I think it is just not feasible to document how one has to install Docker. There are two many variations here. If anything we can add a note, that one needs the docker client.

README.md
```diff
@@ -218,19 +218,45 @@ When running OpenShift in a single VM, it is recommended to reuse the Docker dae
...
+
+1. Enable and start the `docker` and `docker-containerd` services.
+
+```
+$ systemctl start docker && systemctl enable docker
```
hferentschik: This does not make sense. You are effectively starting the Docker daemon here, right? This is exactly what we not want. You are describing how to install and run the Docker daemon on Fedora, but the whole point is that you want to reuse the Docker daemon of the Minishift VM.

README.md
LalatenduMohanty: @thatdocslady As @hferentschik mentioned, OS X and Windows is our primary target for this usecase. Linux is secondary as User can run native docker. I think we can refer users to https://docs.docker.com/engine/installation/binaries/ for it. For Linux we just need to make sure docker is installed. If we want to use docker client we do not need to start the docker daemon. I think we do not need to put all the details but enough that user can figure it out.

README.md
LalatenduMohanty: As I mentioned in my other comment, `docker` daemon need not be started to use installed docker as a client binary.

thatdocslady:
Ok @LalatenduMohanty @hferentschik I understand, thanks for the feedback. I removed the detailed instructions, and just let the first step to make sure that the client is installed. The rest of the instructions are the same as the section had before my edits, just broken down into steps (instead of a paragraph, the way they were listed before).

README.md
hferentschik: " Make sure that you have `docker` and `docker-client` installed on your machine."

"docker" and "docker-client" are rpm package names. They don't exist on other OS types and they are also not consistently names across Linux distributions. What we can write is that you need the "Docker client binary" installed on your system. How to do this varies a lot. Check out https://docs.docker.com/engine/installation/ to see how varied it can be.

thatdocslady:
@hferentschik ok I updated the instructions and added the link as an extra nice thing.

README.md
```diff
@@ -218,19 +218,29 @@ When running OpenShift in a single VM, it is recommended to reuse the Docker dae
 for pure Docker use-cases as well.
 By using the same docker daemon as Minishift, you can speed up your local experiments.
 
-To be able to work with the docker daemon on your Mac or GNU/Linux host use the
-[`docker-env command`](https://minishift.io/docs/minishift_docker-env.html) in your shell:
+1. Make sure that you have the Docker client binary installed on your machine. For information about
+specific binary installations for your operating system, see the [Docker installation page](https://docs.docker.com/engine/installation/).
+
+1. Start Minishift with the `minishift start` command.
```
hferentschik: We could add the link to the start command here as well as we do for the _docker-env_ command.

README.md
```diff
@@ -218,19 +218,29 @@ When running OpenShift in a single VM, it is recommended to reuse the Docker dae
...
+
+1. Run the [`minishift docker-env`](https://minishift.io/docs/minishift_docker-env.html) command
```
hferentschik: -1, this command does nothing. It does not export any variables. In fact you cannot export variables into the parent shell. You just need the `eval $(minishift docker-env)` call. That will evaluate the output of `minishift docker-env` in the context of my current shell.

README.md
```diff
@@ -218,19 +218,29 @@ When running OpenShift in a single VM, it is recommended to reuse the Docker dae
...
+to export the environment variables that are required to reuse the daemon.
+
+1.  Configure your shell to use the exported environment variables.
```
hferentschik: This is the command which exports the environment variables. It is not using any previously exported variables.

thatdocslady:
Updated again @hferentschik

hferentschik:
Merged, thanks.

Issue #570, opened 2017-03-20
Quote passwords when passing them to subscription manager


dustymabe:
Passwords that have special characters in them need to be quoted on the command line (if you run it through a shell which will interpret such characters). We either need to quote the password with `''` or we need to not run the command through a shell, which may be preferable.

Either way, once a fix is proposed, please try to run through the process with some special characters defined (`%$^`, etc) and make sure you can log in.

dustymabe:
FYI - if this is not already fixed (maybe I missed it somewhere), then it should probably be fixed before next CDK release.

hferentschik:
@dustymabe, just to better understand the problem here. You are using the latest CDK 3 beta and try to register it by passing your subscription credentials during the start command. You are having a password with special characters which you escape accordingly on the command line when executing the start command. However, this will still fail to start the CDK, since we are internally not passing the password quoted to the subscription manager? Is this correct? 

Maybe you could provide an example command and the error you are seeing?

BTW, since this seems to be CDK issue, it should actually be tracked in the CDK JIRA issue tracker - https://issues.jboss.org/projects/CDK. Any change you can open an issue there?

dustymabe:
> @dustymabe, just to better understand the problem here. You are using the latest CDK 3 beta and try to register it by passing your subscription credentials during the start command. You are having a password with special characters which you escape accordingly on the command line when executing the start command. However, this will still fail to start the CDK, since we are internally not passing the password quoted to the subscription manager? Is this correct?

that is correct

> Maybe you could provide an example command and the error you are seeing?

i run minishift start with my user/pass in env vars and it fails

```
[cloud-user@cloudhost ~]$ minishift start
Starting local OpenShift cluster using 'kvm' hypervisor...
Registering machine using subscription-manager
E0319 20:21:38.506799   18203 start.go:150] Error starting the VM: Error registering the VM: ssh command error:
command : sudo subscription-manager register --auto-attach --username yourusername --password pa$$word 
err     : exit status 70
output  : Invalid username or password. To create a login, please visit https://www.redhat.com/wapps/ugc/register.html
Registering to: subscription.rhsm.redhat.com:443/subscription
. Retrying.
```

> BTW, since this seems to be CDK issue, it should actually be tracked in the CDK JIRA issue tracker - https://issues.jboss.org/projects/CDK. Any change you can open an issue there?

[...]

PR #616, opened 2017-03-28
Issue #570 Add steps to handle shell special variables

praveenkumar:
Around spending around a day time that I came up with for handing the shell specific characters and tested in all 3 platforms (windows/Linux/OSx).

### DIFF
--- a/docs/source/using/troubleshooting.adoc
+++ b/docs/source/using/troubleshooting.adoc
@@ -113,6 +113,15 @@ cause passwords to fail.
 Workaround: When creating and entering passwords, wrap the string with
 single quotes in the following format: `<password>`
 
+NOTE: The characters `", $, ` and \` are still interpreted by the shell,
+even when you wrap the string with single quotes. If your password contain
+these special characters, make sure you escape them using the
+`backslash (\)` character.
+
+----
+Pass$$wd" => 'Pass\$\$d\"'
+----
+
 [[troubleshooting-snapshots]]
 == Snapshots with virsh
 


docs/source/using/troubleshooting.adoc
```diff
@@ -113,6 +113,14 @@ cause passwords to fail.
 Workaround: When creating and entering passwords, wrap the string with
 single quotes in the following format: `<password>`
 
+Note: However, ", $, \`, and \ are still interpreted by the shell,
```
thatdocslady: - Note format should be NOTE: (asciidoc)
- Remove "However", and change to "The characters" (as a prefix for the list)
- Wrap the symbols with ``

docs/source/using/troubleshooting.adoc
```diff
@@ -113,6 +113,14 @@ cause passwords to fail.
...
+even when you wrapped with quotes. If your password contain those special
```
thatdocslady: change to "even when you wrap the string with single quotes"

docs/source/using/troubleshooting.adoc
```diff
@@ -113,6 +113,14 @@ cause passwords to fail.
...
+characters then make sure you escape them using `backslash (\)` character.
```
thatdocslady: - remove "then", replace with comma (,)
- "using the backslash"

docs/source/using/troubleshooting.adoc
```diff
@@ -113,6 +113,14 @@ cause passwords to fail.
...
+
+----
+Pass$$wd" => 'Pass\$\$d\"'
```
budhrg: +1 for adding example

budhrg:
LGTM 👍

docs/source/using/troubleshooting.adoc
Preeticp: change to 'If your password contains these special...'

docs/source/using/troubleshooting.adoc
```diff
@@ -113,6 +113,15 @@ cause passwords to fail.
 Workaround: When creating and entering passwords, wrap the string with
 single quotes in the following format: `<password>`
 
+NOTE: The characters `", $, ` and \` are still interpreted by the shell,
```
hferentschik: -1 I don't think this is correct. Single quotes protects everything except other single quotes. So to use the example 'Pass\$\$d\' will literally print the backslashes and all. If you put a string into single quotes, the only thing which is problematic is the single quote itself, since it will in this case end the single quoted string. You cannot either escape the single quote within a single quoted string. You rather have to concatenate two strings. 

Escaping works in double quoted strings which is not what we describe here.

See also http://linux.101hacks.com/bash-scripting/quotes-inside-shell-script/

It might be that 'Pass\$\$d\"' actually works, but this is because of a bug. When you specify this string on the shell it will be passed as as _Pass\$\$d\"_ to Minishift. In this we pass this escaped string on to registration where we don't escape via single quotes. This means the special characters get escaped and it seemingly works. 

The statement "The characters `", $, ` and \` are still interpreted by the shell even when you wrap the string with single quotes" is, however,  wrong.

@praveen Have you tried passing a password with a space? 'hello world' for example? The proper solution is to put the password in single quotes when calling he ssh command. You also need to parse the password for potential single quotes and do something like 'my'"'"'password'

praveenkumar: @hferentschik Looks like I was not clear enough with the details. Let me try to explain with the logs trace and experiment which I did to find out why we need another level of escaping.

1. Start minishift with `'Password$$'` parameter.

```
$ ./minishift start --username username --password 'Password$$' --show-libmachine-logs --v=5
...
Using SSH private key: /Users/prkumar/.minishift/machines/minishift/id_rsa (-rw-------)
&{[-F /dev/null -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=none docker@192.168.64.2 -o IdentitiesOnly=yes -i /Users/prkumar/.minishift/machines/minishift/id_rsa -p 22] /usr/bin/ssh <nil>}
About to run SSH command:
sudo -E subscription-manager register --auto-attach --username username --password Password$$ 
SSH cmd err, output: exit status 70: Registering to: subscription.rhsm.redhat.com:443/subscription
Invalid username or password. To create a login, please visit https://www.redhat.com/wapps/ugc/register.html
```

Above you can see the way SSH command detect it and try to run it will see without single quoted password parameter which interpreted differently in bash, same as you also pointed out in the comment.

2. Since now we have the command and ssh key already configured to box so I just killed the minishift process and VM is still running to do some checks on ssh side.

On the host let's try to run same ssh command with different escaping methods.

```
$ /usr/bin/ssh -F /dev/null -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=none docker@192.168.64.2 -o IdentitiesOnly=yes -i /Users/prkumar/.minishift/machines/minishift/id_rsa -p 22 sudo -E subscription-manager register --auto-attach --username username --password 'Password$$'
subscription.rhsm.redhat.com:443/subscription
Invalid username or password. To create a login, please visit https://www.redhat.com/wapps/ugc/register.html

$ /usr/bin/ssh -F /dev/null -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=none docker@192.168.64.2 -o IdentitiesOnly=yes -i /Users/prkumar/.minishift/machines/minishift/id_rsa -p 22 sudo -E subscription-manager register --auto-attach --username username --password /''Password$$'/'
Registering to: subscription.rhsm.redhat.com:443/subscription
The system has been registered with ID: 981f2562-3a0f-44a7-ba40-a71ab3866b73

$ /usr/bin/ssh -F /dev/null -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=none docker@192.168.64.2 -o IdentitiesOnly=yes -i /Users/prkumar/.minishift/machines/minishift/id_rsa -p 22 sudo -E subscription-manager register --auto-attach --username username --password Password\$\$
Registering to: subscription.rhsm.redhat.com:443/subscription
The system has been registered with ID: 981f2562-3a0f-44a7-ba40-a71ab3866b73
```

Above you can see last 2 methods works for a special character, for 2nd method it will only work with `$` special character but 3rd method will work for each special shell character as long as we provide escape. 

> the statement "The characters ", $, and ` are still interpreted by the shell even when you wrap the string with single quotes" is, however, wrong.

I agree and I really want some good wording around it why we need it.

praveenkumar: > Have you tried passing a password with a space? 'hello world' for example? The proper solution is to put the password in single quotes when calling he ssh command.

This is something we can't set as per access.redhat.com portal if you try to put space or any below character then you can't set your RHN password. But I understand your intention here, also http://teaching.idallen.com/cst8207/14w/notes/445_quotes_for_remote.html#quoting-variables-in-remote-command-lines helped me to understand the ssh behavior.

```
<extract from access.redhat.com>
Please correct the following errors:

Password must be ASCII and cannot contain the following special characters (") (<) (>) (space)
```

hferentschik: > I agree and I really want some good wording around it why we need it.

Ok, as long as we agree on this. However, it will always be hard to explain, unless we don't fix the underlying problem. For the user, using the single quotes on the console is all which should be required. He does the right thing. 

So ssh does not use the single quote syntax to escape the password and it also does not handle spaces. Fair enough. In this case we have to do whatever ssh expect and that is escaping these meta characters. Why offloading this to the user and creating confusion. We have the password string, right? All we need is a single function which takes the password and escapes it to ssh standards. Am I missing something here?

praveenkumar: > We have the password string, right? All we need is a single function which takes the password and escapes it to ssh standards.

@hferentschik That make sense and I think mostly we just need to convert that string to a ssh-acceptable so end user doesn't have issue. might be simple regex or replace works, have to check.

praveenkumar:
Closing this PR because we want to handle it programmatically.

Issue #620, opened 2017-03-29
Fix documentation links in README file


thatdocslady:
After we have a final target for the published documentation, we need to remove the existing list in https://github.com/minishift/minishift#documentation and replace it with a link to the doc library locations.

thatdocslady:
Closing this issue because it'll be handled as a part of #533 after docs publication

PR #621, opened 2017-03-29
[WIP] Issue 533: Refactored README file

thatdocslady:
Fixes issue #533 

This PR handles the following changes:

- Converted to asciidoc
- Extracted installation steps, quickstart, and Docker daemon reuse info to separate topics
- Converted all links to relative links in new topics 
- Light proofread of content
- Updated topic map

Out of scope:

- Refactor links in README is not possible yet until we have final doc target (created issue #620 )


hferentschik:
@thatdocslady, did we not say that we want to keep the README untouched until we have the updates site hosted somewhere? If we were to merge this right now, we would have "lost" some documentation, since the extracted docs won't be accessible anywhere right now.

Can you not exclude the changes to the README and keep them until we are ready at which stage we can just add them?

thatdocslady:
@hferentschik this is exactly what I asked you the other day on IRC and you said not to wait. I'm sorry but you keep confusing me... 

Anyway, since OpenShift isn't responding yet to the request for staging/testing, I recommend that we merge and then push the update to GitHub Pages so that we have the docs available in our public instance.

hferentschik:
>  this is exactly what I asked you the other day on IRC and you said not to wait. I'm sorry but you keep confusing me...

the changes are fine, we just need to wait with modifying the README. 

> I recommend that we merge and then push the update to GitHub Pages so that we have the docs available in our public instance.

That is not so easy. The current site is based on the markdown pages. We will need to see how we go about making this work with the new site. This is what I wanted to avoid. I was hoping that we can get this over with within one sprint, so that we don't have to use GitHub pages anymore. An update of the GitHub pages would be happening with the net release. It is a released version we are documenting, not in flight changes. Users won't have a binary with the latest changes. They are using the latest released version. This is what should be documented.

thatdocslady:
I don't know what to do anymore about this :( I guess the GitHub Pages site is already outdated after we did the asciidoc conversion, and when you say "the changes are fine" I don't know which changes you mean because this task for refactoring the README is all about converting to asciidoc and extracting the content to separate topics. 

There was no point in converting the readme to asciidoc without extracting the content, so when you told me to go ahead I understood this is what you meant when you said "to prepare for the deployment", because now we have an updated topic map and all the topics in their proper place for the tarball generation.

What do you recommend we do now? I don't want to lose all the work I did on this, and it's something that we do need to have ready before we test the docs on the openshift site. I already pinged openshift devs on Monday (Tuesday?), maybe you can help with reaching out to them so that we can move forward?

LalatenduMohanty:
@hferentschik @thatdocslady  What about keeping the README as markdown and other changes in the PR as it is. I think that will work with our current tooling and then we convert it to asciidoc when it is right .

thatdocslady:
@LalatenduMohanty the README format is actually not relevant here because it's only accessible from the GitHub live-rendering, so it'll be rendered from either markdown or asciidoc. What @hferentschik is referring to is the extracted content, that will now live in separate topics.

I was thinking about it and maybe I can change back the links from the readme to point to the asciidoc files in GitHub for installation and quickstart (since most people land still in GitHub first) and use the follow-up issue #620 to also address these? That way the extracted content isn't "hidden" and you can still navigate to it even without GitHub pages.

LalatenduMohanty:
@thatdocslady Closing this PR as you are planning to send a new PR for #623

PR #626, opened 2017-03-31
Issue #535 CentOS CI changes for docs artifacts uploads

### LINKED ISSUE #535: Create Jenkins job to build docs

budhrg:
Fix #535

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ clean_docs: build_docs_container
 serve_docs: synopsis_docs build_docs_container
 	cd docs && docker run -e OPENSHIFT_VERSION=$(OPENSHIFT_VERSION) -e MINISHIFT_VERSION=$(MINISHIFT_VERSION) -p 35729:35729 -p 4567:4567 -tiv $(shell pwd)/docs:/home/docs:Z minishift/docs serve[--watcher-force-polling]
 
-$(DOCS_SYNOPISIS_DIR)/*.md: vendor
+$(DOCS_SYNOPISIS_DIR)/*.md: vendor $(ADDON_ASSET_FILE)
 	@# https://github.com/golang/go/issues/15038#issuecomment-207631885 ( CGO_ENABLED=0 )
 	DOCS_SYNOPISIS_DIR=$(DOCS_SYNOPISIS_DIR) CGO_ENABLED=0 go run -ldflags="$(LDFLAGS)" -tags gendocs gen_help_text.go
 
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -22,79 +22,176 @@ set -e
 
 # Source environment variables of the jenkins slave
 # that might interest this worker.
-if [ -e "jenkins-env" ]; then
-  cat jenkins-env \
-    | grep -E "(JENKINS_URL|GIT_BRANCH|GIT_COMMIT|BUILD_NUMBER|ghprbSourceBranch|ghprbActualCommit|BUILD_URL|ghprbPullId|GH_TOKEN|CICO_API_KEY)=" \
-    | sed 's/^/export /g' \
-    > ~/.jenkins-env
-  source ~/.jenkins-env
-fi
+function load_jenkins_vars() {
+  if [ -e "jenkins-env" ]; then
+    cat jenkins-env \
+      | grep -E "(JENKINS_URL|GIT_BRANCH|GIT_COMMIT|BUILD_NUMBER|ghprbSourceBranch|ghprbActualCommit|BUILD_URL|ghprbPullId|GH_TOKEN|CICO_API_KEY|JOB_NAME)=" \
+      | sed 's/^/export /g' \
+      > ~/.jenkins-env
+    source ~/.jenkins-env
+  fi
 
-# We need to disable selinux for now, XXX
-/usr/sbin/setenforce 0
+  echo 'CICO: Jenkins ENVs loaded'
+}
 
-# Get all the deps in
-yum -y install \
-  gcc \
-  make \
-  git \
-  curl \
-  kvm \
-  qemu-kvm \
-  libvirt
+function install_core_deps() {
+  # We need to disable selinux for now, XXX
+  /usr/sbin/setenforce 0
 
-# Start Libvirt
-sudo systemctl start libvirtd
+  # Get all the deps in
+  yum -y install gcc \
+                 make \
+                 git \
+                 curl
 
-# Setup 'kvm' driver
-curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.7.0/docker-machine-driver-kvm > /usr/local/bin/docker-machine-driver-kvm && \
-chmod +x /usr/local/bin/docker-machine-driver-kvm
+  echo 'CICO: Core dependencies installed'
+}
 
-# Install Go 1.7
-curl -LO https://storage.googleapis.com/golang/go1.7.linux-amd64.tar.gz
-tar -xf go1.7.linux-amd64.tar.gz
-mv go /usr/local
+function install_kvm_virt() {
+  sudo yum -y install kvm \
+                      qemu-kvm \
+                      libvirt
+  # Start Libvirt
+  sudo systemctl start libvirtd
+  echo 'CICO: KVM hypervisor installed and started'
 
-# Setup GOROOT
-export GOROOT=/usr/local/go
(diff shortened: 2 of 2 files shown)


hferentschik:
@budhrg any idea why the image build failed?

hferentschik:
retest this please

hferentschik:
@budhrg Well done. I like this now much better. Once we figure out why it fails, we can merge.

hferentschik:
@budhrg 

> useradd: UID 0 is not unique

Seems the UID is '0' for the build slave. :-(

centos_ci.sh
```diff
@@ -22,79 +22,155 @@ set -e
...
+PASS=$(echo $CICO_API_KEY | cut -d'-' -f1-2)
+
+echo "JOB_NAME: $JOB_NAME"
+
+# Intentionally using wrong condition to test docs artifacts uploads
+if [[ "$JOB_NAME" != "minishift-docs" ]]; then
+  install_docker;
+  prepare;
+  cd $GOPATH/src/github.com/minishift/minishift
+  make gen_adoc_tar
```
hferentschik: @budhrg can you try:

    make gen_adoc_tar UID=1000

This should overwrite the 0. Question is whether it will work when the files are created.

budhrg: ok

budhrg:
@hferentschik 

> Seems the UID is '0' for the build slave. :-(

They run as root

centos_ci.sh
```diff
@@ -22,79 +22,155 @@ set -e
...
+    cp -r out/* minishift/pr/$ghprbPullId/
+    RSYNC_PASSWORD=$1 rsync -a --relative minishift/pr/$ghprbPullId/ minishift@artifacts.ci.centos.org::minishift/
+    echo "Find Artifacts here http://artifacts.ci.centos.org/minishift/minishift/pr/$ghprbPullId ."
+  fi
+}
+
+function docs_artifacts_upload() {
+  set +x
+
+  version=$(cat docs/build/variables.adoc | cut -d' ' -f2)
```
hferentschik: nice

budhrg: @hferentschik Also check Makefile changes too

i need to remove clean from gen_adoc_tar ,  to pass permission issue I got https://ci.centos.org/job/minishift-pr/514/

budhrg: Hoping this build passes https://ci.centos.org/job/minishift-pr/515/

centos_ci.sh
```diff
@@ -22,79 +22,155 @@ set -e
...
+  mkdir -p minishift/docs/$(version)
+  cp docs/build/minishift-adocs.tar minishift/docs/$(version)/
+  ln -s $(pwd)/minishift/docs/$(version) minishift/docs/latest
```
hferentschik: 👍 that's what I meant. See whether this works, once the build goes all the way through

budhrg: Sure

centos_ci.sh
budhrg: Seems to be not working https://ci.centos.org/job/minishift-pr/515/console :(

budhrg: Hit by 
```
00:04:32.031 rake aborted!
00:04:32.031 Errno::EACCES: Permission denied @ dir_s_mkdir - build
00:04:32.032 /home/docs/Rakefile:20:in `block in <top (required)>'
00:04:32.032 Tasks: TOP => adoc_tar => init
00:04:32.032 (See full trace by running task with --trace)
00:04:32.061 make: *** [gen_adoc_tar] Error 1
```

Makefile
```diff
@@ -41,6 +41,10 @@ else
 	UID = $(shell id -u)
 endif
 
+# Set some default if UID is 0 or user is root
+ifeq ($(UID),0)
```
hferentschik: This is most likely not a good idea

Makefile
```diff
@@ -79,7 +83,7 @@ build_docs_container:
 
 .PHONY: gen_adoc_tar
 gen_adoc_tar: synopsis_docs build_docs_container
-	cd docs && docker run -e MINISHIFT_VERSION=$(MINISHIFT_VERSION) -tiv $(shell pwd)/docs:/home/docs:Z minishift/docs clean adoc_tar
+	cd docs && docker run -e MINISHIFT_VERSION=$(MINISHIFT_VERSION) -tiv $(shell pwd)/docs:/home/docs:Z minishift/docs adoc_tar
```
hferentschik: we will need the clean. If not for the CI build, but for dev it is important.

centos_ci.sh
hferentschik: Well, the image build works, but then of course the permissions are screwed up. One way to this this would be to actually create a user on the slave side as well. Same as we do in the Dockerfile. And then we switch to this user. I think that might be a good idea.

budhrg: hmm, so @hferentschik you are suggesting to this for only docs operation right?

budhrg: That seems to bit of work.

For this, now we need to install all core deps, docker with password with sudoer role.
Any idea how we can do it?

hferentschik: > you are suggesting to this for only docs operation right?

not really, nothing wrong with doing this regardless.

hferentschik: > now we need to install all core dips

not sure what you mean? What other core deps would you need. You don't really need anything else on top of what you otherwise would need, right? You create the user and his home directory and take it from there. Similar to what the Dockerfile does really. You might have to add  add the user to the docker group as well to be able to run docker without sudo, but that should be it.

budhrg:
retest this please

budhrg:
retest this please

centos_ci.sh
```diff
@@ -22,79 +22,155 @@ set -e
...
+}
+
+# Execution begins here
+load_jenkins_vars;
+
+# Intentionally using wrong condition to test docs artifacts uploads
+if [[ "$JOB_NAME" != "minishift-docs" ]]; then
+  prepare_docs_user;
+  # Run everything beyond this point as 'docs' user
+  runuser -l docs -c install_core_deps;
```
hferentschik: interesting, hope this is going to work

hferentschik: Looking at the output it shows that 'runuser' gets its own environment. This is not just continuing in the context of this script which makes sense since a new process is created. So this process needs to call for example the _centos_ci.sh_ script. What you need to do is to check which user id the script is running under. If it is the root user, you call the script again (with the right parameters) using runuser. If the user is is non 0, you know the script is executed as non root user and you can continue.

hferentschik: What's the state on this one? I was thinking about this a bit more, if possible I'd like to run all builds as a new/created user. This way we can extend the master build as well. On a push to master we build the docs tarball as well and then use it in a checkout of docs.openshift.org to build the full docs site which we then can stage on the archive server.

Makefile
```diff
@@ -110,7 +110,7 @@ clean_docs: build_docs_container
 serve_docs: synopsis_docs build_docs_container
 	cd docs && docker run -e OPENSHIFT_VERSION=$(OPENSHIFT_VERSION) -e MINISHIFT_VERSION=$(MINISHIFT_VERSION) -p 35729:35729 -p 4567:4567 -tiv $(shell pwd)/docs:/home/docs:Z minishift/docs serve[--watcher-force-polling]
 
-$(DOCS_SYNOPISIS_DIR)/*.md: vendor
+$(DOCS_SYNOPISIS_DIR)/*.md: vendor $(ADDON_ASSET_FILE)
```
hferentschik: That might be required. Did you get compilation errors when starting from a clean build, right?

hferentschik:
@budhrg where do we stand on this atm? Do we have an ETA?

Makefile
budhrg: Yes @hferentschik

budhrg:
@hferentschik , IMO this PR is ready to merge now.

Now, we can upload the docs tar to artifacts server with link http://artifacts.ci.centos.org/minishift/minishift/docs/latest-alpha/.

Successful job link https://ci.centos.org/job/minishift-pr/644/console.

This PR only have the code to build docs tar and upload tar into specific location for docs.openshift.org use
which will be trigger manually by specific 'minishift-docs' job URL as we discussed.

Created https://github.com/minishift/minishift/issues/716 to address the hosting of docs based on master build.

Issue #511, opened 2017-03-03
Add a version command as openshift subcommand


praveenkumar:
Currently user doesn't know which version of OpenShift they are using/running, adding version command will help in that sense.

hferentschik:
If we add the version command _minishift openshift version_ reporting the version of the currently running OpenShift cluster, we could add the _list_versions_ command as subcommand: _minishift openshift version list_. This would align with _minishift openshift service_ and _minishift openshift service list_.

praveenkumar:
We might to look into to get tags from the docker hub for the released so that in downstream we can just change the url to internal registry and everything works.

gbraad:
#515

hferentschik:
Merged via pull request #628

PR #628, opened 2017-03-31
Issue #511 Add a version command as openshift subcommand

praveenkumar:
This patch will enable what we wanted to achieve with `version` and `version list`.

```
17:48 $ ./minishift start --openshift-version v1.5.0-alpha-0
Minishift does not support Openshift version v1.5.0-alpha-0 You need to use a version >=v1.5.0-rc.0

17:52 $ ./minishift openshift version
openshift v1.5.0-rc.0+49a4a7a
kubernetes v1.5.2+43a9be4
etcd 3.1.0

17:54 $ ./minishift openshift version list
	- v3.6.0-alpha.0
	- v1.5.0-rc.0
```

### DIFF
--- a/cmd/minishift/cmd/openshift/version.go
+++ b/cmd/minishift/cmd/openshift/version.go
@@ -1,0 +1,61 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openshift
+
+import (
+	"fmt"
+	"github.com/docker/machine/libmachine"
+	"github.com/docker/machine/libmachine/provision"
+	"github.com/golang/glog"
+	"github.com/minishift/minishift/pkg/minikube/cluster"
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/docker"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+// version command represent current running openshift version and available one.
+var versionCmd = &cobra.Command{
+	Use:   "version [command] [flags]",
+	Short: "Print the current running openshift version to stdout",
+	Long:  `Print the current running openshift version to stdout`,
+	Run:   runVersion,
+}
+
+func runVersion(cmd *cobra.Command, args []string) {
+	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+	defer api.Close()
+
+	host, err := cluster.CheckIfApiExistsAndLoad(api)
+	if err != nil {
+		fmt.Println(nonExistentMachineError)
+		atexit.Exit(1)
+	}
+	sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
+	dockerCommander := docker.NewVmDockerCommander(sshCommander)
+	version, err := dockerCommander.Exec(" ", "origin", "openshift", "version")
+	if err != nil {
+		glog.Errorln("Error restarting OpenShift cluster: ", err)
+		atexit.Exit(1)
+	}
+	fmt.Fprintln(os.Stdout, version)
+}
+
+func init() {
+	OpenShiftConfigCmd.AddCommand(versionCmd)
+}
--- a/pkg/minishift/openshift/version/openshift_versions.go
+++ b/pkg/minishift/openshift/version/openshift_versions.go
@@ -17,48 +17,113 @@ limitations under the License.
 package version
 
 import (
+	"encoding/json"
 	"fmt"
+	"github.com/minishift/minishift/pkg/util"
 	"io"
-
-	"github.com/google/go-github/github"
-	"github.com/minishift/minishift/pkg/minishift/util"
-	githubutil "github.com/minishift/minishift/pkg/util/github"
-	"github.com/pkg/errors"
+	"net/http"
+	"sort"
+	"strings"
+	"time"
 )
 
-const githubOwner = "openshift"
-const githubRepo = "origin"
-
-func PrintOpenShiftVersionsFromGitHub(output io.Writer) {
-	PrintOpenShiftVersions(output)
+type ImageTags struct {
+	Count    int         `json:"count"`
+	Next     string      `json:"next"`
+	Previous interface{} `json:"previous"`
+	Results  []Results   `json:"results"`
 }
 
-func PrintOpenShiftVersions(output io.Writer) {
-	versions, err := getVersions()
+type Results struct {
+	Name        string      `json:"name"`
+	FullSize    int         `json:"full_size"`
+	Images      []ImageInfo `json:"images"`
+	ID          int         `json:"id"`
+	Repository  int         `json:"repository"`
+	Creator     int         `json:"creator"`
+	LastUpdater int         `json:"last_updater"`
+	LastUpdated time.Time   `json:"last_updated"`
+	ImageID     interface{} `json:"image_id"`
+	V2          bool        `json:"v2"`
+}
+
+type ImageInfo struct {
+	Size         int         `json:"size"`
+	Architecture string      `json:"architecture"`
+	Variant      interface{} `json:"variant"`
+	Features     interface{} `json:"features"`
+	Os           interface{} `json:"os"`
+	OsVersion    interface{} `json:"os_version"`
+	OsFeatures   interface{} `json:"os_features"`
+}
+
+func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) {
+	resp, err := getResponseBody("https://registry.access.redhat.com/v1/repositories/openshift3/ose/tags")
 	if err != nil {
-		fmt.Println(err.Error())
+		fmt.Println("Error Occured", err)
 		return
 	}
-	fmt.Fprint(output, "The following OpenShift versions are available: \n")
-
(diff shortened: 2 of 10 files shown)


pkg/minikube/openshiftversions/openshift_versions.go
```diff
@@ -17,48 +17,119 @@ limitations under the License.
 package openshiftversions
 
 import (
+	"encoding/json"
 	"fmt"
 	"io"
-
```
praveenkumar: I removed github stuff completely and using docker registry for upstream and access.registry.redhat.com for downstream.

pkg/minikube/openshiftversions/openshift_versions_test.go
```diff
@@ -17,149 +17,67 @@ limitations under the License.
 package openshiftversions
 
 import (
-	"bytes"
-	"encoding/json"
-	"fmt"
-	"net/http"
-	"net/http/httptest"
-	"net/url"
+	"io/ioutil"
```
praveenkumar: Same with the tests, I removed all github related test cases and mostly those were skipping because rate limit issue and here it's pure expected and actual values now.

pkg/util/utils.go
```diff
@@ -74,9 +74,14 @@ func ValidateProxyURI(uri string) bool {
 	return govalidator.IsURL(uri)
 }
 
-func ValidateOpenshiftMinVersion(ver string) bool {
+func ValidateOpenshiftMinVersion(ver string, minVersion string) bool {
```
praveenkumar: I added another parameter to make sure unit test should work.

budhrg:
@praveenkumar ,

The below line:
```
Minishift does not support Openshift version v1.5.0-alpha-0 You need to use a version >=v1.5.0-rc.0
```
need full stop before `You` and at the end.

cmd/minishift/cmd/openshift/version.go
```diff
@@ -0,0 +1,61 @@
+/*
+Copyright (C) 2016 Red Hat, Inc.
```
budhrg: 2017

hferentschik:
@praveenkumar - there are several merge conflicts after merging the addon changes. I think we need to start with rebasing this pull request

cmd/minishift/cmd/openshift/version.go
```diff
@@ -0,0 +1,61 @@
...
+	defer api.Close()
+
+	host, err := cluster.CheckIfApiExistsAndLoad(api)
+	if err != nil {
+		fmt.Println(nonExistentMachineError)
+		atexit.Exit(1)
+	}
+	sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
+	dockerCommander := docker.NewVmDockerCommander(sshCommander)
+	version, err := dockerCommander.Exec(" ", "origin", "openshift", "version")
```
hferentschik: Maybe we can add a _Version_ command to _openshift.go_?

pkg/minikube/openshiftversions/openshift_versions.go
```diff
@@ -17,48 +17,125 @@ limitations under the License.
...
 	if err != nil {
 		return nil, err
 	}
-	defer resp.Body.Close()
+	return resp, nil
+}
 
-	if len(releases) == 0 {
-		return nil, errors.New("There are no OpenShift versions available.")
+func versionOrdinal(version string) string {
```
hferentschik: Are you ordering the releases here? Purely string based? Should we not use semantic ordering and the semantic versioning library we already have for that - github.com/blang/semver?

pkg/minikube/openshiftversions/openshift_versions.go
LalatenduMohanty: I think this is a good idea.

pkg/minikube/openshiftversions/openshift_versions.go
hferentschik: And if we don't do semantic version sorting, why don't you just use _sort.Strings_?

LalatenduMohanty: Also I think we should add a unit test for this function.

praveenkumar: @hferentschik our downstream release doesn't follow Semantic version (http://semver.org/) that's the reason I didn't use it.

pkg/minikube/openshiftversions/openshift_versions.go
```diff
@@ -17,48 +17,125 @@ limitations under the License.
...
+	Variant      interface{} `json:"variant"`
+	Features     interface{} `json:"features"`
+	Os           interface{} `json:"os"`
+	OsVersion    interface{} `json:"os_version"`
+	OsFeatures   interface{} `json:"os_features"`
 }
 
-func PrintOpenShiftVersions(output io.Writer) {
-	versions, err := getVersions()
+func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) {
```
hferentschik: So where is the difference in  _PrintDownStreamVersions_ vs _PrintUpStreamVersions_ apart from the URL they access? Is this not really the same method, just different URL in which case we can just make the URL a method parameter? Probably a good idea to externalise the URL and put them for example into the constant file

pkg/minikube/openshiftversions/openshift_versions.go
```diff
@@ -17,48 +17,125 @@ limitations under the License.
...
-	githubutil "github.com/minishift/minishift/pkg/util/github"
-	"github.com/pkg/errors"
+	"net/http"
+	"strings"
+	"time"
 )
 
-const githubOwner = "openshift"
-const githubRepo = "origin"
+type ImageTags struct {
```
hferentschik: So why all these structs, if you just want the version? One does not have to unmarshal into a struct. You can just have a map and retrieve the data you are after.

pkg/minikube/openshiftversions/openshift_versions.go
```diff
@@ -17,48 +17,125 @@ limitations under the License.
...
 	}
-	fmt.Fprint(output, "The following OpenShift versions are available: \n")
-
-	for _, version := range versions {
-		if util.ValidateOpenshiftMinVersion(*version.TagName) {
-			fmt.Fprintf(output, "\t- %s\n", *version.TagName)
+	defer resp.Body.Close()
+	decoder := json.NewDecoder(resp.Body)
+	var data map[string]string
+	err = decoder.Decode(&data)
```
hferentschik: I might be missing something, but you define all these structs, but read into a  map[string]string?

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -17,48 +17,136 @@ limitations under the License.
...
-		fmt.Println(err.Error())
+		fmt.Println("Error Occured", err)
 		return
 	}
-	fmt.Fprint(output, "The following OpenShift versions are available: \n")
-
-	for _, version := range versions {
-		if util.ValidateOpenshiftMinVersion(*version.TagName) {
-			fmt.Fprintf(output, "\t- %s\n", *version.TagName)
+	defer resp.Body.Close()
```
hferentschik: I think the problem is that this function does too much. It is almost identical to the upstream version, apart from parsing the versions. 

So why don't you extract a function which returns a slice of version strings and pass this slice to a method which can be shared between the upstream and downstream print method.

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -17,48 +17,136 @@ limitations under the License.
...
+func PrintUpStreamVersions(output io.Writer, minSupportedVersion string) {
+	resp, err := getResponseBody("https://registry.hub.docker.com/v2/repositories/openshift/origin/tags/")
+	if err != nil {
+		fmt.Fprintf(output, "Error Occured %s", err)
+		return
+	}
+	defer resp.Body.Close()
+	decoder := json.NewDecoder(resp.Body)
+	var data ImageTags
+	err = decoder.Decode(&data)
```
hferentschik: This is really just a 'var data map[string]interface{}'
You then retrieve the "results" value from the map which us a slice over which you can iterate. Each element should be a map with a name key.

hferentschik: Not saying this needs to be changed, but it would be for sure less code.

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -17,48 +17,136 @@ limitations under the License.
...
 	if err != nil {
 		return nil, err
 	}
-	defer resp.Body.Close()
+	return resp, nil
+}
 
-	if len(releases) == 0 {
-		return nil, errors.New("There are no OpenShift versions available.")
+func versionOrdinal(version string) string {
```
hferentschik: Let's replace this sort method with something simpler. Either sort.Strings or a custom sort.Sort mile based on semver.

pkg/minishift/openshift/version/openshift_versions.go
hferentschik: what's about 'latest'? The response contains a 'latest' tag? Do we want this in the list of versions or should be handle it separately.

praveenkumar: @hferentschik I just don't want to have latest in the version string so I just ignore when `latest` string comes up.

pkg/minishift/openshift/version/openshift_versions.go
LalatenduMohanty: Please description of the function.

LalatenduMohanty: Also we should add a unit test for this function.

hferentschik:
Merged, thanks. Needed to run _make fmt_ though first.

pkg/util/utils.go
```diff
@@ -106,3 +106,36 @@ func (m MultiError) ToError() error {
 	}
 	return fmt.Errorf(strings.Join(errStrings, "\n"))
 }
+
+func VersionOrdinal(version string) string {
+	// ISO/IEC 14651:2011
+	// https://www.iso.org/standard/57976.html
+	// This method is applicable for 255 characters string
+	// to determine their collating order in a sorted list.
+	// It create a collating sorted list and return the string.
```
LalatenduMohanty: I think we should ass comments before the function, so that it would be visible in https://godoc.org/github.com/minishift/minishift

PR #646, opened 2017-04-03
[WIP] Issue #254 Adding document for using OpenShift docker registry

LalatenduMohanty:

### DIFF
--- a/docs/source/using/accessing-the-openshift-docker-registry.adoc
+++ b/docs/source/using/accessing-the-openshift-docker-registry.adoc
@@ -1,0 +1,51 @@
+= Accessing the OpenShift Docker Registry
+:icons:
+:toc: macro
+:toc-title:
+:toclevels: 1
+
+[[openshift-docker-registry-intro]]
+== OpenShift Docker Registry
+
+OpenShift provides an integrated docker registry which can be used for containerized application development as well. 
+
+This helps to speed up the local development as images present in registry can be used for applications running in OpenShift. Refer to the link:..accessing-the-openshift-docker-registry{outfilesuffix}#staging-app-with-openshift-registry[Staging your application with the OpenShift registry] to know more.
+
+[[login-to-openshift-docker-registry]]
+=== Login to the docker registry
+
+. Start minishift with the link:../command-ref/minishift_start{outfilesuffix}[`minishift start`] command.
+
+. Add the oc binary to the PATH environment variable.
+
+ $ export PATH=$PATH:~/.minishift/cache/oc/{openshift-version}
+
+. Export the environment variables that are required link:../using/reusing-docker-daemon{outfilesuffix}[to reuse the Docker daemon]
+
+ $ eval $(minishift docker-env)
+
+. Login to the docker registry
+
+ $ docker login -u developer -p `oc whoami -t` 172.30.1.1:5000 
+
+[[staging-app-with-openshift-registry]]
+=== Staging your application with the OpenShift registry
+
+The following example shows how to deploy an OpenShift application directly from a locally built docker image.
+
+In this example OpenShift project "myproject" is used as "myproject" is automatically created and available for use after `minishift start`.
+
+. Build the Docker image as usual.
+
+. Tag the image against the OpenShift registry with the registry.
+
+ $ docker tag my-app 172.30.1.1:5000/myproject/my-app
+
+. Push the image to the registry to create an image stream with the same name as the application.
+
+ $ docker push 172.30.1.1:5000/myproject/my-app
+
+. Create an application from the image stream and expose the image.
+
+ $ oc new-app --image-stream=my-app --name=my-app
+ $ oc expose service my-app --hostname=my-app.minishift
--- a/docs/source/_topic_map.yml
+++ b/docs/source/_topic_map.yml
@@ -29,6 +29,8 @@ Topics:
         File: reusing-docker-daemon
       - Name: Troubleshooting Minishift
         File: troubleshooting
+      - Name: Accessing the OpenShift docker registry
+        File: accessing-the-openshift-docker-registry
   - Name: Developing Minishift
     Dir: developing
     Topics:
--- a/docs/source/developing/writing-docs.adoc
+++ b/docs/source/developing/writing-docs.adoc
@@ -17,6 +17,9 @@ documentation is a mix of generated
 https://en.wikipedia.org/wiki/Markdown[Markdown] files and manually
 maintained https://en.wikipedia.org/wiki/AsciiDoc[Asciidoc] files.
 
+If you are adding a new file to the doumentation please update 
+the `_topic_map.yml`  present in the root of `docs/source` sub-folder.
+
 Per default the documentation is built in a Docker container. This way
 you avoid having to install all the required dependencies on your
 development machine. All you need is a running Docker daemon. In case
--- a/docs/source/using/index.adoc
+++ b/docs/source/using/index.adoc
@@ -13,4 +13,5 @@ OpenShift cluster. It also helps troubleshooting common issues.
 - link:../using/managing-minishift{outfilesuffix}[Managing Minishift]
 - link:../using/interacting-with-openshift{outfilesuffix}[Interacting with OpenShift]
 - link:../using/reusing-docker-daemon{outfilesuffix}[Reusing the Docker daemon]
+- link:../using/accessing-the-openshift-docker-registry{outfilesuffix}[Accessing the OpenShift Docker Registry]
 - link:../using/troubleshooting{outfilesuffix}[Troubleshooting Minishift]


docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,25 @@
+= Accessing the OpenShift Docker Registry
+:icons:
+:toc: macro
+:toc-title:
+:toclevels: 1
+
+OpenShift provides an integrated docker registry which can be used as a local registry.
+
+This helps to speed up the local development as the images present in the registry can be used as imagestream for running an application in the OpenShift
```
budhrg: Need a full stop at the end of this end. One line is need saying "To achieve that, perform following steps (or similar lines)".

hferentschik: "as the images present in the registry can be used as imagestream for running an application in the OpenShift" - what do you mean with that? 

Also "in the OpenShift" is not clear to me.

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+. Start minishift with the link:../command-ref/minishift_start{outfilesuffix}[`minishift start`] command.
+
+. Add the oc binary to the PATH environment variable.
+
+ $ export PATH=$PATH:~/.minishift/cache/oc/v1.5.0-rc.0
```
hferentschik: don't hard code the oc version. use a variable instead, see interacting-with-openshift.adoc in the OpenShift client library section.

hferentschik: The whole process is already described in re-using the docker daemon. I guess you can refer to that.

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,25 @@
...
+
+. Export the environment variables that are required to reuse the daemon.
+
+ $ eval $(minishift docker-env)
+
+. Login to the docker registry
+
+ $ docker login -u developer -p `oc whoami -t` 172.30.1.1:5000 
+
```
hferentschik: No example of how to get do a push to this registry?

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: I mean we can use the docker images in the application running in OpenShift. Something like this https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/staging.adoc

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: Added a link to point users to upstream document of docker registry as I did not want to rewrite the same stuff here.

hferentschik: I think a full blown example would be nice here, especially since the document becomes very short otherwise. Maybe we should in this case just add a short section to managing minishift? Personally I think there would be enough here to harbour a separate document.

docs/source/using/accessing-the-openshift-docker-registry.adoc
hferentschik: > I mean we can use the docker images in the application running in OpenShift.

Why don't you write this then?

hferentschik: I mean why don't you replicate the example? IMO it is a good example and would clarify what you mean.

budhrg:
retest this please

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: Make sense. I have added as an example. It is similar to what ADB docs hvae.

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
+= Accessing the OpenShift Docker Registry
+:icons:
+:toc: macro
+:toc-title:
+:toclevels: 1
+
+OpenShift provides an integrated docker registry which can be used as a local registry.
```
hferentschik: "which can be used as a local registry" - why 'local'? What does this mean really?

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+This helps to speed up the local development as images present in registry can be used for applications running in OpenShift.
```
hferentschik: How? Without a pointer it or example it is really hard to know what you mean.

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
+= Accessing the OpenShift Docker Registry
+:icons:
+:toc: macro
+:toc-title:
+:toclevels: 1
+
```
hferentschik: we should try to have a second level heading (==) here. The pages render better then. I know some other files don't have it either yet, but the aim should be to have it.

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
...
+This helps to speed up the local development as images present in registry can be used for applications running in OpenShift.
+
+== Login to the docker registry 
+
+. Start minishift with the link:../command-ref/minishift_start{outfilesuffix}[`minishift start`] command.
+
+. Add the oc binary to the PATH environment variable.
+
+ $ export PATH=$PATH:~/.minishift/cache/oc/{openshift-version}
```
hferentschik: Does the variable substitution work w/o including _variables.adoc_?

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+. Export the environment variables that are required link:../using/reusing-docker-daemon{outfilesuffix}[to reuse the Docker daemon]
+
+ $ eval $(minishift docker-env)
```
hferentschik: for everything up to here, why don't we just reference reusing-docker-daemon. This is exactly the same thing.

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+. Login to the docker registry
+
+ $ docker login -u developer -p `oc whoami -t` 172.30.1.1:5000 
+
+== Staging your application with the OpenShift registry
+
+This scenario is intended to illustrate how to test an application Docker image to a local instance of OpenShift platform that runs in Minishift.
```
hferentschik: "The following example shows how to deploy an OpenShift application directly from a locally built Docker image"

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+. Build the Docker image as usual.
+
+. Tag the image against the OpenShift registry with the hostname.
```
hferentschik: "with the hostname" - that's not correct. We are not using a hostname here. We are using the _fixed_ internal OpenShift registry IP of 172.30.1.1.

It should in fact be mentioned somewhere, that usually the internal IP of the OpenShift registry is dynamic, but that _cluster up_ uses a fixed IP to allow for this use case.

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+. Export the environment variables that are required link:../using/reusing-docker-daemon{outfilesuffix}[to reuse the Docker daemon]
+
+ $ eval $(minishift docker-env)
+
+. Login to the docker registry
+
+ $ docker login -u developer -p `oc whoami -t` 172.30.1.1:5000 
+
+== Staging your application with the OpenShift registry
```
budhrg: 👍 for this one.

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+This scenario is intended to illustrate how to test an application Docker image to a local instance of OpenShift platform that runs in Minishift.
+
+. Build the Docker image as usual.
+
+. Tag the image against the OpenShift registry with the hostname.
+
+ $ docker tag my-app 172.30.1.1:5000/myproject/my-app
```
budhrg: Is `myproject` is hardcoded here and is the default project we have ?

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: Sure. Will do that.

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: Not sure how to reword this. This supposed to communicate the user that the docker registry provided by OpenShift through Minishift can used as a docker registry which is local to the workstation i.e. user has a local docker registry to push docker images.

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: The next section in the same doc explains that. Should I add pointer to the next section?

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: ahh, you mean this version should have proper value when the html is generated?

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: I have referred  reusing-docker-daemon here. Also have the command as we do not the user to go to another page just for one command.

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: myproject is the default project in Minishift.

docs/source/using/accessing-the-openshift-docker-registry.adoc
hferentschik: +1

docs/source/using/accessing-the-openshift-docker-registry.adoc
hferentschik: yes, it will

docs/source/using/accessing-the-openshift-docker-registry.adoc
hferentschik: 👍

docs/source/using/accessing-the-openshift-docker-registry.adoc
hferentschik: it might make sense to add a description here, explaining where _myproject_ is coming from.

docs/source/using/accessing-the-openshift-docker-registry.adoc
hferentschik: " which can be used for development as well" (maybe)

docs/source/using/accessing-the-openshift-docker-registry.adoc
hferentschik: It is more than just one command. All three commands are the same. So by just referencing by saying: "Provided you console is configured to use the Minishift Docker daemon (see bla) we can skip to repeat and maintain the identical steps at two places. If a user knows what "reusing the Minishfit Docker daemon means", all is good, otherwise it makes sense imo top start there.

Preeticp: we can skip to repeat and maintain the identical steps at two places.
>+1 you could combine the first 2 steps and link it to https://github.com/minishift/minishift#starting-minishift.
3rd step can stay as is with the link. In both the steps we do not need to add the commands, as otherwise we would need to maintain them in both places in case of any future changes.

docs/source/using/accessing-the-openshift-docker-registry.adoc
Preeticp: Link to docker registry yet to be added.

docs/source/using/accessing-the-openshift-docker-registry.adoc
Preeticp: +1, should we add 'using Minishift' to the end of that line?

docs/source/using/accessing-the-openshift-docker-registry.adoc
Preeticp: We could add this information as a note here.

docs/source/using/accessing-the-openshift-docker-registry.adoc
```diff
@@ -0,0 +1,44 @@
...
+
+. Push the image to the registry to create an image stream with the same name as the application.
+
+ $ docker push 172.30.1.1:5000/myproject/my-app
+
+. Create an application from the image stream and expose the image.
+
+ $ oc new-app --image-stream=my-app --name=my-app
+ $ oc expose service my-app --hostname=my-app.minishift
```
Preeticp: Please add a line here summarizing what these steps did or what the user must/can do once he completes these steps. Maybe something like, 'You can now deploy the OpenShift application using the locally built Docker image.

budhrg:
retest this please

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: @hferentschik I am not sure about if we need to add variables.adoc

budhrg: @LalatenduMohanty , if you need to replace `{openshift-version}` to actual version, then yes it is needed.

budhrg:
@LalatenduMohanty , you also need to update [_topic_map.yml](https://github.com/minishift/minishift/blob/master/docs/source/_topic_map.yml#L31) for the relevant changes.

LalatenduMohanty:
> you also need to update _topic_map.yml for the relevant changes.

@budhrg thanks. @hferentschik @Preeticp @thatdocslady  This is not listed in https://github.com/minishift/minishift/blob/master/docs/source/developing/writing-docs.adoc. Should I update it?

Preeticp:
Yes please @LalatenduMohanty

LalatenduMohanty:
Closing this PR and will send a fresh PR after addressing the comments.

docs/source/using/accessing-the-openshift-docker-registry.adoc
LalatenduMohanty: @Preeticp IMO this is not required as the last step deploy the application and there is no extra step required to deploy the application.

hferentschik:
+1

docs/source/using/accessing-the-openshift-docker-registry.adoc
Preeticp: Ok

Issue #619, opened 2017-03-29
Convert CONTRIBUTING.md and ROADMAP.md to asciidoc


thatdocslady:
For consistency across all our docs, these files should be converted to asciidoc and adjusted to include links to new/updated topics.

hferentschik:
Merged via pull request #663

Issue #317, opened 2017-01-23
Implement host folder mounts using sshfs


hferentschik:
To have a consistent hypervisor independent solution for host folder mounts, we could implement a vagrant-sshfs equivalent for Minishift. This would come down to:

* [x] Prepare our ISOs with sshfs
* [ ] Use Golang's sftp server functionality to create a sftp daemon process on the host
  * [ ] As with vagrant-sshfs, the tricky part is to be able to properly spawn daemon processes on on all OS types. For Windows in particular it means we will have to make Win32 API calls. 
  * [ ] Expose commands within the CLI to start/stop the sftp server
* [ ] Write the glue code which wires up the sftp server on host with sshfs in the VM

praveenkumar:
+ SELinux context should be set accordingly for mounted folders.

gbraad:
https://github.com/docker/machine/pull/4018

gbraad:
On PR #1528 the question is raised how to solve the non-interactive adding of a hostfolder using in a cross-type manner.

Probably for this, we can not rely on the viper/cobra to read the flags, but need to parse passed options ourselves (as done with the `--profile` option).

The question is how to rescribe the mandatory and optional fields that are passed as options to the HostFolder type. I would suggest to add a little bit of metadata for this in the HostFolders.go file and describe them by `type` specifier.

LalatenduMohanty:
@hferentschik looks like libmachine added some support for SSHFS [1] . Have you checked it out. 

`Added new docker-machine mount command for mounting machine directories over SSHFS`
[1] https://github.com/docker/machine/releases/tag/v0.13.0

gbraad:
This implementation has been discussed before (even with the Minikube guys), but it seems it got finally released in 0.13.0. Good, as #1685 pulls this in as dependency (https://github.com/minishift/minishift/pull/1685/commits/119a9104d5f3d261940fa770d34938b39838f21c) so nothing prevents us from using this implementation, or being compatible with. This has to be decided on short notice.

gbraad:
Please consider the comment: https://github.com/minishift/minishift/issues/1024#issuecomment-351619634 Is the sftp-server binary for OpenSSH-Win32 still considered? Be aware, that the developer mode, or enabling the new #1836 functionality claims ports on the host we might need for the SSHFS functionality.

praveenkumar:
@hferentschik you are still owner for this task :)

praveenkumar:
[...]

PR #663, opened 2017-04-04
Issue 619: Convert contributing and roadmap docs to asciidoc

thatdocslady:
Fixes issue #619 

Converted the files to asciidoc and adjusted some light proofreading and formatting.

Out of scope: Linking to these files from elsewhere (only one reference was found and it's in the README which will already be handled as a part of #533 ).


ROADMAP.adoc
```diff
@@ -0,0 +1,77 @@
...
+The following sections give a high level overview of the envisioned
+features and how they fit into upcoming releases. Of course this is
+subject to change and actual version names might change as well, because
+http://semver.org/[semantic versioning] is used to determine the final
+version name of a given release.
+
+[[minishift-100]]
+=== Minishift 1.0.0
+
+- Ability to select the OpenShift version of the cluster [line-through]#link:https://github.com/minishift/minishift/issues/316[#141]#
```
hferentschik: You lost some formatting here. In Markdown completed features used strikethrough. Now there is no indication anymore what is completed. If not strikethrough, we might now use checkbox '- [ ] ' syntax which is quite nice as well. All expect "Consistent host folder mounting https://github.com/minishift/minishift/issues/316[#316], https://github.com/minishift/minishift/issues/317[#317]" should be marked completed.

ROADMAP.adoc
```diff
@@ -0,0 +1,77 @@
...
+- Consistent host folder mounting https://github.com/minishift/minishift/issues/316[#316], https://github.com/minishift/minishift/issues/317[#317]
+- Exposure of the OpenShift registry https://github.com/minishift/minishift/issues/254[#254]
+- Ability to configure/patch OpenShift's master and node configuration [line-through]#https://github.com/minishift/minishift/issues/276[#276]#
+- Ability to customize cluster (for example, adding templates and imagestreams) https://github.com/minishift/minishift/issues/177[#177]
+
+[[minishift-110]]
+=== Minishift 1.1.0
+
+- Add 'minishift login' command https://github.com/minishift/minishift/issues/349[#349]
+- Add 'minishift' context https://github.com/minishift/minishift/issues/359[#359]
```
hferentschik: This should be moved to the 1.0 section and marked as completed.

ROADMAP.adoc
thatdocslady: Oh damn. Well the [line-through] tag is the asciidoc way of marking strikethrough, and I tested this locally fine, but now after submitting the PR I see that it doesn't live-render in GitHub. Need to think about how to mitigate, maybe we can just write something like [DONE] next to features that were completed instead?

thatdocslady:
@hferentschik the checkbox formatting works so I'll go ahead and apply it instead of the strikethrough

hferentschik:
> the checkbox formatting works so I'll go ahead and apply it instead of the strikethrough

works for me

hferentschik:
Merged, thanks

PR #672, opened 2017-04-05
Issue #570 Handle shell special charcter in password string

praveenkumar:
This patch will handle shell special character for RHN password string.

Addresses issue #570

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -18,11 +18,12 @@ package util
 
 import (
 	"fmt"
-	"github.com/asaskevich/govalidator"
 	"io"
 	"os"
 	"strings"
 	"time"
+
+	"github.com/asaskevich/govalidator"
 )
 
 // Until endlessly loops the provided function until a message is received on the done channel.
@@ -105,6 +106,16 @@ func (m MultiError) ToError() error {
 		errStrings = append(errStrings, err.Error())
 	}
 	return fmt.Errorf(strings.Join(errStrings, "\n"))
+}
+
+// EscapeStringForSSHUse will parse string and replace
+// shell special characters with escape
+func EscapeStringForSSHUse(s string) string {
+	r := strings.NewReplacer("$", "\\$",
+		`"`, `\"`,
+		"`", "\\`",
+		`\`, `\\`)
+	return r.Replace(s)
 }
 
 func VersionOrdinal(version string) string {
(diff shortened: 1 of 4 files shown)


pkg/minishift/util/password_parse.go
```diff
@@ -0,0 +1,33 @@
...
+	"strings"
+)
+
+// ParseRegistrationPassword will parse password and replace
+// shell special characters with escape
+func ParseRegistrationPassword(password string) string {
+	r := strings.NewReplacer("$", "\\$",
+		`"`, `\"`,
+		"`", "\\`",
+		`\`, `\\`)
```
praveenkumar: Right now I added what I found as shell special character which might be interpreter differently. If you have anything in mind, I will add it to this list.

hferentschik: IMO this is not a registration password specific method, but a generic "EscapeStringForSSHUse" method. In which case I would move this into the general _util_ package, outside of _minishift_. This can probably used at other places as well. WDYT?

pkg/minishift/util/password_parse_test.go
```diff
@@ -0,0 +1,30 @@
...
+limitations under the License.
+*/
+
+package util
+
+import (
+	"testing"
+)
+
+func TestParseRegistrationPassword(t *testing.T) {
```
hferentschik: I'd add for completeness some positive tests as well to show that strings with no special characters stay untouched.

pkg/minishift/util/password_parse_test.go
```diff
@@ -0,0 +1,30 @@
+/*
+Copyright (C) 2016 Red Hat, Inc.
```
budhrg: 2017 here

pkg/minishift/util/password_parse.go
```diff
@@ -0,0 +1,33 @@
...
+package util
+
+import (
+	"strings"
+)
+
+// ParseRegistrationPassword will parse password and replace
+// shell special characters with escape
+func ParseRegistrationPassword(password string) string {
+	r := strings.NewReplacer("$", "\\$",
```
budhrg: 👍

pkg/util/utils.go
```diff
@@ -18,11 +18,12 @@ package util
```
budhrg: 👍  for adding it as part of here.

hferentschik:
Merged, thanks

PR #678, opened 2017-04-05
Issue #677 Fix regression on 'service list' implementation

### LINKED ISSUE #677: Regression on "service list"
It seems the code which I added to make `minishift openshift service list` work for showing all routes available to user is not working.

I see the changes I added as part of my [PR here](https://github.com/budhrg/minishift/blob/d91a72f506efc62f4e70c45a7039fd6312c3134c/cmd/minishift/cmd/openshift/service_list.go).

This seems to be lost in rebase.

budhrg:
Fix #677

### DIFF
--- a/cmd/minishift/cmd/openshift/service_list.go
+++ b/cmd/minishift/cmd/openshift/service_list.go
@@ -55,6 +55,6 @@ var serviceListCmd = &cobra.Command{
 }
 
 func init() {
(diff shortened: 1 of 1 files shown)


hferentschik:
-1, this was actually intentional. We can discuss later.

budhrg:
> This was actually intentional to achieve consistency. I think there would be another way, but there is more to it than just adding using an empty string here. Let's discuss this first.

Sure @hferentschik

PR #684, opened 2017-04-06
Issue 533 README cleanup

thatdocslady:
Fixes issue #533 

The PR converts the README to adoc, removes duplicate install/quickstart/docker information, and fixes the documentation link.

NOTE: I guess because it's been a while since I touched this work, somehow more commits managed to sneak into this while I rebased. My Git-foo isn't helpful here so would appreciate help in trying to fix this.


README.adoc
```diff
@@ -0,0 +1,83 @@
...
+OpenShift. Make sure that the hypervisor of your choice is installed and
+enabled on your system before you install Minishift.
+
+For detailed installation instructions for Minishift and the required dependencies, see
+the https://docs.openshift.org/latest/minishift/getting-started/index.html[Getting started] documentation.
+
+[[documentation]]
+== Documentation
+
+Minishift documentation is published as a part of the OpenShift Origin documentation librarry.
```
hferentschik: "librarry", one 'r' should be enough ;-)

README.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+- https://github.com/minishift/minishift/releases[Release Notes]
+- link:./ROADMAP.adoc[Roadmap]
+
+[[limitations]]
+== Limitations
+
+The following features are not supported in Minishift.
+
+- Features that require a Cloud Provider, such as:
```
hferentschik: This is not really right. PersistentVolumes for example does not require a cloud provider. (I know this is who it was).

To be honest, I am tempted to say, let's remove this from here. If anything we can find some space in the backwaters of the documentation. It feels odd to have such a short README and having nothing else to do, then talk about limitations. If anything we should add a "Features" section where we highlight what we can do.

hferentschik:
Merged, thanks. I dropped the unwanted commits and as discussed removed the limitations sections.

I like how short and to the point the README is now. Good stuff.

README.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+To download the latest binary and review release notes, see
+the https://github.com/minishift/minishift/releases[Minishift releases] page.
+
+Minishift requires a hypervisor to start the virtual machine containing
+OpenShift. Make sure that the hypervisor of your choice is installed and
+enabled on your system before you install Minishift.
+
+For detailed installation instructions for Minishift and the required dependencies, see
+the https://docs.openshift.org/latest/minishift/getting-started/index.html[Getting started] documentation.
```
Preeticp: Suggest to make this paragraph the first one in this section.
Starting the section with - To seems a bit abrupt.

README.adoc
```diff
@@ -0,0 +1,83 @@
...
+
+[[documentation]]
+== Documentation
+
+Minishift documentation is published as a part of the OpenShift Origin documentation librarry.
+Check out the latest official Minishift documentation for information about getting started,
```
Preeticp: 'latest official' :) not sure those words are needed here?
Docs in GitHub are not unofficial, nor would they be stale, in fact these would be more current than the docs on the site right?

README.adoc
Preeticp: +1 Limitations should be moved to the backwaters and features added

README.adoc
thatdocslady: -1 I think that this order is ok because it gives you the starting point, some prerequisite notice, and only then sends you to the instructions.

README.adoc
thatdocslady: the docs in github will be the latest unofficial after GA, we'd branch out the release and the published docs will only be updated every time we GA

README.adoc
Preeticp: If that is so I would suggest you to reword the sentence starting from To download

Issue #649, opened 2017-04-03
Avoid the use of 'glog.Errorln' in hostfolder related commands


hferentschik:
$ minishift hostfolder list
    E0331 21:38:02.699866   86851 list.go:38] Error:  Host does not exist: “minishift"

or

    $ minishift hostfolder mount
    E0331 21:39:02.418972   86885 mount.go:44] Error:  Host does not exist: “minishift"

This really relates to issue #169, but short of finding a complete solution to this issue or getting log to print "nicer" error message, I think we should in this cases avoid the use of glog and in particular glog.Errorln, since it really does not make for a nice user facing message.

hferentschik:
Resolved via pull request #736

Issue #169, opened 2016-11-29
Develop a logging facade 


hferentschik:
Logging seems to be a bit all over the place in Minishift. There are ftm.Print statements as well we usages of glog. On top of this used libraries like for example libmachine have their own logging wrapper. 

Obviously the target for ftm vs glog is different, but if nothing else we should document a logging strategy. Personally I think it should even be possible to somehow have some unifying/configurable wrapper.

hferentschik:
If nothing else we should document when and how to log in the dev guide. From a users perspective it would make sense on document how to enable the various logging capabilities.

budhrg:
Some design philosophy on logging by William Kennedy https://www.goinggo.net/2017/05/design-philosophy-on-logging.html

budhrg:
Found one package https://github.com/sirupsen/logrus . Seems to be maintained often.

hferentschik:
This issue is not about introducing a new logging dependency (at least not if we don't have to). It is about unifying how we do logging, how the log format looks like, how we can configure  it and how we maybe can introduce logging categories.

anjannath:
Based on @hferentschik [comment.](https://github.com/minishift/minishift/issues/169#issuecomment-311740188) we need to decide the following.

-  [ ] Logging format (Structure, what information in logs).
-  [ ] Logging levels / categories.
-  [ ] Logging pkg to use. (logrus, glog, go-logging, zap, etc..)

anjannath:
We (@budhrg and /me) looked into the code and found out that, _minishift_ does not use that much of logging, the logging is mostly spit out by _libmachine_. Since there is not much logging in _minishift_, i think its best to understand what is expected out of this. Are we going to handle everything on the stdout using a logging framework? that means the msgs from preflight checks are also handled by the logging framework.

As for frameworks, _libmachine_ uses its own logging [pkg](https://github.com/docker/machine/tree/master/libmachine/log). In minishift we are using **glog**, which allows us to have different levels of logging using the `-v` flag. If we use the log pkg from _libmachine_ then our logs will come together with _libmachine_ logs and we won't have levels anymore. We also need to decide what kind of information belongs to which level, AFAIK right now there is no criteria as to which msg belong to which level of verbosity.

/cc @minishift/minishift-dev

gbraad:
Well, as in the original description, many logs are provided as `fmt.Print`, which can be wrapped as a logging level, either info, error, etc. this has to be reviewed and decided upon.

[...]

PR #685, opened 2017-04-06
Issue #649 Avoid using glog.Errorln for user facing commands

praveenkumar:
This patch remove glog.Errorln from user facing commands so user don't get something like `E0331 21:38:02.699866   86851 list.go:38]` when hit a error.

Addresses issue #649

### DIFF
--- a/cmd/minishift/cmd/addon/addon_enable_test.go
+++ b/cmd/minishift/cmd/addon/addon_enable_test.go
@@ -19,20 +19,21 @@ package addon
 import (
 	"bytes"
 	"fmt"
-	"github.com/minishift/minishift/pkg/testing/cli"
-	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"io"
 	"os"
 	"path/filepath"
 	"testing"
+
+	"github.com/minishift/minishift/pkg/testing/cli"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
 )
 
 func Test_addon_name_must_be_specified_for_enable_command(t *testing.T) {
 	tmpMinishiftHomeDir := cli.SetupTmpMinishiftHome(t)
-	origStdout, stdOutWriter, stdOutReader := cli.CaptureStdOut(t)
-	defer cli.TearDown(tmpMinishiftHomeDir, origStdout)
+	origStdout, origStderr, streamWriter, streamReader := cli.CaptureStreamOut(t, 1)
+	defer cli.TearDown(tmpMinishiftHomeDir, origStdout, origStderr)
 
-	atexit.RegisterExitHandler(cli.CreateExitHandlerFunc(t, stdOutWriter, stdOutReader, 1, emptyEnableError))
+	atexit.RegisterExitHandler(cli.CreateExitHandlerFunc(t, streamWriter, streamReader, 1, emptyEnableError))
 
 	runEnableAddon(nil, nil)
 }
@@ -41,15 +42,15 @@ func Test_unkown_name_for_enable_command_returns_error(t *testing.T) {
 	tmpMinishiftHomeDir := cli.SetupTmpMinishiftHome(t)
 	os.Mkdir(filepath.Join(tmpMinishiftHomeDir, "addons"), 0777)
 
-	origStdout, stdOutWriter, stdOutReader := cli.CaptureStdOut(t)
-	defer cli.TearDown(tmpMinishiftHomeDir, origStdout)
+	origStdout, origStderr, streamWriter, streamReader := cli.CaptureStreamOut(t, 0)
+	defer cli.TearDown(tmpMinishiftHomeDir, origStdout, origStderr)
 
 	testAddOnName := "foo"
 	runEnableAddon(nil, []string{testAddOnName})
 
-	stdOutWriter.Close()
+	streamWriter.Close()
 	var buffer bytes.Buffer
-	io.Copy(&buffer, stdOutReader)
+	io.Copy(&buffer, streamReader)
 
 	expectedOut := fmt.Sprintf(noAddOnToEnableMessage+"\n", testAddOnName)
 	if expectedOut != buffer.String() {
--- a/cmd/minishift/cmd/addon/addons_list.go
+++ b/cmd/minishift/cmd/addon/addons_list.go
@@ -20,13 +20,14 @@ import (
 	"os"
 	"text/template"
 
-	"github.com/golang/glog"
+	"fmt"
+	"io"
+	"sort"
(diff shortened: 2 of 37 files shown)


cmd/minishift/cmd/hostfolder/mount.go
```diff
@@ -62,7 +63,7 @@ var hostfolderMountCmd = &cobra.Command{
 		}
 
 		if err != nil {
-			glog.Errorln(err)
+			fmt.Println("Error to mount host folder: ", err)
```
hferentschik: You are missing something here. See a few lines above

hferentschik: Line 45.

hferentschik:
Could you also please adjust the commit message?

pkg/minikube/tests/ssh_mock.go
```diff
@@ -104,7 +105,7 @@ func (s *SSHServer) Start() (int, error) {
 					//Note: string(req.Payload) adds additional characters to start of input, execRequest used to solve this issue
 					var cmd execRequest
 					if err := ssh.Unmarshal(req.Payload, &cmd); err != nil {
-						glog.Errorln("Unmarshall encountered error: %s", err)
+						fmt.Println("Unmarshall encountered error: ", err)
```
hferentschik: not sure whether we need to change it here, but I guess it does not matter

cmd/minishift/cmd/addon/addons_list.go
```diff
@@ -63,13 +64,13 @@ func init() {
 	var err error
 	defaultListTemplate, err = template.New("list").Parse(defaultAddonListFormat)
 	if err != nil {
-		glog.Errorln("Error creating list template:", err)
+		fmt.Println("Error creating list template: ", err)
```
hferentschik: Ok, let's look at this once more. This is over and over the same pattern. Log somethings and then call _Exit_. What's about making the code better? If we either

* make atExit a variadic function, aka _Exit(exitCode int, messages ...string)
* our by adding _atexit.ExitWithMessage(exitCode int, message string)

We encapsulate the actual logging and how we do it in one single spot. If the then later, for example as part of #169 want to use a logger, we only have to change this at one single spot, instead of touching 20 files again.

The second option is most probably the better, since in the first we would need to decide what to do with multiple message and what it actually means.

praveenkumar: @hferentschik you just stole my idea :), that something I had in mind, was going through the changes.

hferentschik: :-)

pkg/util/os/atexit/atexit.go
```diff
@@ -29,6 +30,13 @@ func Exit(code int) {
 	os.Exit(code)
 }
 
+// ExitWithMessage exit and print the error message.
```
praveenkumar: @hferentschik Hope this is what you wanted, I didn't put everything because let me clear about the requirement.

praveenkumar: I will adjust the unit test also accordingly.

hferentschik: Right, an additional Exit method. Since Go does not have overloading we need to provide a new name or use a variadic method.

praveenkumar: @hferentschik Agree, I went for new name.

hferentschik:
@praveenkumar, is this finished now? Just wondering due to the title still having [WIP]?

praveenkumar:
@hferentschik Now it finished, I had to check the unit tests and required some other changes. Have a look now.

pkg/util/os/atexit/atexit.go
```diff
@@ -29,6 +31,12 @@ func Exit(code int) {
...
+func ExitWithMessage(code int, msg string) {
```
LalatenduMohanty: 👍

budhrg: Nice, we would have really required this at some point of time.

pkg/util/os/atexit/atexit.go
```diff
@@ -29,6 +31,12 @@ func Exit(code int) {
...
+	fmt.Println(msg)
```
hferentschik: I suggest one more change, namely that we change the output steam depending on the exit code. If the exit code is '0' we log to stdout and if it is not equal '0' we print to stderr. We can use _fmt.Fprintln_ for that, selecting the right stream.

pkg/util/os/atexit/atexit.go
```diff
@@ -29,6 +31,16 @@ func Exit(code int) {
 	os.Exit(code)
 }
 
+// ExitWithMessage exit and print the error message.
+func ExitWithMessage(code int, msg string) {
+	if code == 0 {
+		fmt.Fprintln(os.Stdout, msg)
```
hferentschik: 👍 thanks

praveenkumar:
Looks like test again started to fail :(

hferentschik:
> Looks like test again started to fail :(

It fails, since there are broken test.

praveenkumar:
@hferentschik please verify also check the test cases.

cmd/minishift/cmd/addon/addon_enable_test.go
```diff
@@ -19,20 +19,21 @@ package addon
...
+
+	"github.com/minishift/minishift/pkg/testing/cli"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
 )
 
 func Test_addon_name_must_be_specified_for_enable_command(t *testing.T) {
 	tmpMinishiftHomeDir := cli.SetupTmpMinishiftHome(t)
-	origStdout, stdOutWriter, stdOutReader := cli.CaptureStdOut(t)
-	defer cli.TearDown(tmpMinishiftHomeDir, origStdout)
+	origStdout, origStderr, streamWriter, streamReader := cli.CaptureStreamOut(t, 1)
```
hferentschik: this feels just wrong. First one has so many return values and second you need to pass the exit code here which somehow decides which stream to capture. This is confusing at best. I think I have an idea to fix that. We need to encapsulate this into a struct. This will also make the teardown easier. 

I can take it from here if you like unless you want to give it a go.

hferentschik:
I think I found what we are looking for ...

praveenkumar:
> I think I found what we are looking for ...

👍 
Please go ahead, I did spend lot of time today to understand streams part :)

hferentschik:
Superseded by pull request #736 which amend this pull request.

PR #679, opened 2017-04-05
Issue #254 Adding document for using OpenShift docker registry

LalatenduMohanty:
Signed-off-by: Lalatendu Mohanty <lmohanty@redhat.com>

Issue #671, opened 2017-04-05
Create addon for persistence volume creation


hferentschik:
This might be temporary and be replaced/changed when we get persistence volumes as part of _cluster up_

jorgemoralespou:
@hferentschik look into how oc-cluster sets up PV. It's as easy as running a job. Or you can explore how I do it in the oc-cluster which don't require jobs.

hferentschik:
+1

LalatenduMohanty:
So seems like we can just point `--host-pv-dir = /var/lib/minishift/openshift.local.pv` for `oc cluster up` and it will create the directories which can be used for PV. Refer [1]

[1] https://github.com/openshift/origin/pull/12456

jorgemoralespou:
@LalatenduMohanty still I think an addon can provide benefit as this are the default 100 PVs that get created with "oc cluster".
There might be cases where you want to create the PVs in the host folder you share with the VM, or with a different type of options. I would not discard still adding some user experience, as if I could, I would just not create those 100 PVs by default, which takes some time to process.

csrwng:
@jorgemoralespou the pv provisioning happens entirely in the background, but you can start using them as soon as the first pv is created. What's the downside?

jorgemoralespou:
@csrwng as I mentioned before, this creates 100 directories in my host for each cluster I start, which although in the background is something it don't make me happy. I would rather prefer to control the amount or have less (max 20). Also, I can't modify any of the options for these (e.g. annotate with a service class, prebound volume,...) so I need to have the ability to set these. I still need a command that let's me be more expressive with the volumes I create.

praveenkumar:
@hferentschik Can we close this issue now?

hferentschik:
Let's close it for now. We have been talking about being able to skip the 'cluster up' persistence volume creation and doing it more explicitly in an add-on. However, as long as 'cluster up' dies create the persistent volumes regardless, we should be using them. We can create a new issue in case we want to look into this later again.

PR #704, opened 2017-04-07
Issue #254 Expose openshift docker registry

praveenkumar:
This patch is update expose registry addon to make sure it works with exposed route.

Addresses issue #254

### DIFF
--- a/addons/expose-registry/expose-registry.addon
+++ b/addons/expose-registry/expose-registry.addon
@@ -2,22 +2,14 @@
 # Description: Exposes the OpenShift internal registry as route
 # URL: https://docs.openshift.org/latest/install_config/registry/securing_and_exposing_registry.html
 
-oc expose service docker-registry -n default
-openshift admin ca create-server-cert --signer-cert=/var/lib/origin/openshift.local.config/master/ca.crt --signer-key=/var/lib/origin/openshift.local.config/master/ca.key --signer-serial=/var/lib/origin/openshift.local.config/master/ca.serial.txt --hostnames='docker-registry-default.#{routing-suffix},docker-registry.default.svc.cluster.local,172.30.1.1' --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key
-openshift cli secrets new registry-secret /etc/secrets/registry.crt /etc/secrets/registry.key
-openshift cli secrets link registry registry-secret
-openshift cli secrets link default  registry-secret
-openshift cli volume dc/docker-registry --add --type=secret --secret-name=registry-secret -m /etc/secrets
-openshift cli set env dc/docker-registry REGISTRY_HTTP_TLS_CERTIFICATE=/etc/secrets/registry.crt REGISTRY_HTTP_TLS_KEY=/etc/secrets/registry.key
+openshift admin ca create-server-cert --signer-cert=/var/lib/origin/openshift.local.config/master/ca.crt --signer-key=/var/lib/origin/openshift.local.config/master/ca.key --signer-serial=/var/lib/origin/openshift.local.config/master/ca.serial.txt --hostnames='docker-registry-default.#{routing-suffix},docker-registry-default.#{routing-suffix}:443,docker-registry.default.svc.cluster.local,172.30.1.1' --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key
+openshift cli create route edge docker-registry --service=docker-registry --ca-cert=/var/lib/origin/openshift.local.config/master/ca.crt --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key --hostname='docker-registry-default.#{routing-suffix}'
 
-openshift cli patch dc/docker-registry -p '{"spec": {"template": {"spec": {"containers":[{"name":"registry","livenessProbe":  {"httpGet": {"scheme":"HTTPS"}}}]}}}}'
-openshift cli patch dc/docker-registry -p '{"spec": {"template": {"spec": {"containers":[{"name":"registry","readinessProbe":  {"httpGet": {"scheme":"HTTPS"}}}]}}}}'
+ssh sudo mkdir -p /etc/docker/certs.d/docker-registry-default.#{routing-suffix}:443
+ssh sudo chown docker /etc/docker/certs.d/docker-registry-default.#{routing-suffix}:443
 
 ssh sudo mkdir -p /etc/docker/certs.d/docker-registry-default.#{routing-suffix}
 ssh sudo chown docker /etc/docker/certs.d/docker-registry-default.#{routing-suffix}
 
-ssh sudo mkdir -p /etc/docker/certs.d/172.30.1.1:5000
-ssh sudo chown docker /etc/docker/certs.d/172.30.1.1:5000
-
+docker cp origin:/var/lib/origin/openshift.local.config/master/ca.crt /etc/docker/certs.d/docker-registry-default.#{routing-suffix}:443/ca.crt
 docker cp origin:/var/lib/origin/openshift.local.config/master/ca.crt /etc/docker/certs.d/docker-registry-default.#{routing-suffix}/ca.crt
-docker cp origin:/var/lib/origin/openshift.local.config/master/ca.crt /etc/docker/certs.d/172.30.1.1:5000/ca.crt


praveenkumar:
Let me know if we need to add it as part of default addons, please test it thoroughly.

hferentschik:
So @praveenkumar, @LalatenduMohanty - what is the error? For me it looks like we are terribly close here. First of all I am able to log in, secondly I seem to be able to push, but I get an error of the form:

    c51fcb726795: Pushing [==================================================>] 254.5 MB
    738aef80313f: Pushing [==================================================>] 200.4 MB/200.4 MB
    34e7b85d83e4: Pushing [==================================================>] 191.9 MB
   unauthorized: authentication required

It also seems that despite of the error message the push seems to complete sometimes (judging by the progress bars).

hferentschik:
Here is another output:

     c0306a6c25d7: Pushing [==================================================>] 1.536 kB
     09b9eec43a71: Pushing [==================================================>] 4.096 kB
     66e6729bcc7f: Pushing [==================================================>] 4.608 kB
     c9031116ff3d: Preparing
     fe6747d73cad: Pushing  2.56 kB
     1edbd65151e2: Waiting
     6a15a9547a32: Waiting
     a43e94a16e03: Waiting
     88c9c552d8d8: Waiting
     b9eda2e140e8: Waiting
     98821eacb28b: Waiting
     596280599f68: Waiting
     5d6cbe0dbcf9: Waiting
     unauthorized: authentication required     

It seems to happen after pushing a few layers. @csrwng Any ideas what could be going on here?

csrwng:
@hferentschik I am getting the same error when using minishift, but not when using latest origin directly on my Mac with 'cluster up'. It seems that it's likely a registry issue. I will look into it further today.

hferentschik:
> It seems that it's likely a registry issue. I will look into it further today.

awesome thanks. Much appreciated.

csrwng:
So it seems that the problem is that in some cases the registry is using the hostname without a port, whereas in other cases it is using the port (:443). 

I was able to get it to work by using the port explicitly. After creating the registry route, I ssh'd into the minishift vm and copied the ca.crt:

```
mkdir -p /etc/docker/certs.d/docker-registry-default.192.168.64.2.nip.io:443
cp /var/lib/minishift/openshift.local.config/master/ca.crt /etc/docker/certs.d/docker-registry-default.192.168.64.2.nip.io:443f
```

Then logged in to docker using the port explicitly:

```
docker login -u developer -p $(oc whoami -t) docker-registry-default.192.168.64.2.nip.io:443
```

And tagged and pushed an image using the 443 port explicitly:
```
docker tag nginx docker-registry-default.192.168.64.2.nip.io:443/myproject/nginx:latest
docker push docker-registry-default.192.168.64.2.nip.io:443/myproject/nginx:latest
```

csrwng:
I also ran into the issue that after restarting the minishift vm that I couldn't push anything to the registry. This is because the registry is using host-dir volume for its storage. The default location for it is in `/var/lib/origin/openshift.local.pv` which is not persisted in the minishift vm. When running cluster up for minishift, you'll need to specify `--host-pv-dir=/var/lib/minishift/openshift.local.pv` so that you get it in a directory that's persisted.

hferentschik:
@csrwng, thanks a lot. We will adjust the add-on and look at the persistence issue. We might be onto something here.

LalatenduMohanty:
@hferentschik @praveenkumar related https://github.com/minishift/minishift/issues/671#issuecomment-293044571

hferentschik:
@csrwng 
> So it seems that the problem is that in some cases the registry is using the hostname without a port, whereas in other cases it is using the port (:443)

Any idea why? Do you think this is intended behaviour?

csrwng:
@hferentschik It looks like this is a red hat docker bug. It should not be the intended behavior
https://github.com/openshift/origin/issues/13646#issuecomment-293029660

hferentschik:
> It looks like this is a red hat docker bug. It should not be the intended behavior

Ok. It felt really odd.

budhrg:
Same authentication problem for me too
```
$ docker push docker-registry-default.192.168.42.232.nip.io/myproject/busybox
The push refers to a repository [docker-registry-default.192.168.42.232.nip.io/myproject/busybox]
c0de73ac9968: Pushing [==================================================>] 1.143 MB
unauthorized: authentication required
```

csrwng:
@budhrg try adding the port to the registry when you tag an image. See https://github.com/minishift/minishift/pull/704#issuecomment-293037352

praveenkumar:
@csrwng So I updated the addon as you suggested to explicit port and still face auth issue when try to push to registry. Something this addon missing?

```
$ docker login -u developer -p qco3IFwvAZBDTGZPwklSSm_hITGXxVe8Z8CXyeuQkQ8 docker-registry-default.192.168.42.23.nip.io:443
Login Succeeded
$ docker tag busybox docker-registry-default.192.168.42.23.nip.io:443/myproject/busybox:latest
$ docker push docker-registry-default.192.168.42.23.nip.io:443/myproject/busybox:latest
The push refers to a repository [docker-registry-default.192.168.42.23.nip.io:443/myproject/busybox]
c0de73ac9968: Preparing 
unauthorized: authentication required
```

csrwng:
@praveenkumar What do you see in the registry logs?

hferentschik:
@csrwng got it to work now as well :-)

hferentschik:
So here is what I did (based on this pull request), using the latest B2D ISO (Boot2Docker version 1.12.3, build HEAD : 7fc7575)

     $ minishift start
     $ docker login -u developer -p $(oc whoami -t) docker-registry-default.$(minishift ip).nip.io:443
     $ docker pull ghost
     $ docker tag ghost docker-registry-default.$(minishift ip).nip.io:443/myproject/ghost:latest
     $ docker push docker-registry-default.$(minishift ip).nip.io:443/myproject/ghost:latest
     $ oc new-app --image-stream=ghost --name=ghost
     $ oc expose service ghost
     $ minishift openshift service ghost -n my project

This will in the end open the Ghost app in my default browser. Note, that this also demonstrate that anyuid is working, since otherwise the Ghost image would not start. Also, check out how nicely the various _oc_, _docker_ and _minishift_ commands interact which each other. Pretty neat imo.    

So let's find out why @praveenkumar and @budhrg don't seem to get this to work with the B2D ISO. Also the next step is to take care of the persistence volume part for the registry, so that the images will survive. 

@csrwng, thinking about what you are saying regarding the registry. This is a general "problem" right? This is a pretty big issue in terms of restarting Minishift in general. So if I understand this correctly, this also means that all build images of the user are gone. So rollback etc would not work. I guess a running app itself would come back up, right?

budhrg:
@hferentschik @praveenkumar  Working for me now

```
$ minishift start                                              
 19.96 MB / 19.96 MB [============================================================================================================================] 100.00% 0s
..........

   To login as administrator:
       oc login -u system:admin

Applying addon registry:
.....
default.192.168.42.19.nip.io/ca.crt
  [539.868758ms]

$ minishift ssh
                        ##         .
                  ## ## ##        ==
               ## ## ## ## ##    ===
           /"""""""""""""""""\___/ ===
      ~~~ {~~ ~~~~ ~~~ ~~~~ ~~~ ~ /  ===- ~~~
           \______ o           __/
             \    \         __/
              \____\_______/
 _                 _   ____     _            _
| |__   ___   ___ | |_|___ \ __| | ___   ___| | _____ _ __
| '_ \ / _ \ / _ \| __| __) / _` |/ _ \ / __| |/ / _ \ '__|
| |_) | (_) | (_) | |_ / __/ (_| | (_) | (__|   <  __/ |
|_.__/ \___/ \___/ \__|_____\__,_|\___/ \___|_|\_\___|_|
Boot2Docker version 1.12.3, build HEAD : 7fc7575 - Thu Oct 27 17:23:17 UTC 2016
Docker version 1.12.3, build 6b644ec

docker@minishift:~$ exit

$ docker login -u developer -p $(oc whoami -t) docker-registry-default.$(minishift ip).nip.io:443
Login Succeeded

$ docker pull busybox
Using default tag: latest
latest: Pulling from library/busybox
7520415ce762: Pull complete 
Digest: sha256:32f093055929dbc23dec4d03e09dfe971f5973a9ca5cf059cbfb644c206aa83f
Status: Downloaded newer image for busybox:latest

$ docker tag busybox docker-registry-default.$(minishift ip).nip.io:443/myproject/busybox:latest

$ docker push docker-registry-default.$(minishift ip).nip.io:443/myproject/busybox:latest
The push refers to a repository [docker-registry-default.192.168.42.19.nip.io:443/myproject/busybox]
c0de73ac9968: Pushed 
latest: digest: sha256:efe08f2a8a18ff5ec89e0f89f904d20ee540bc18f4d8e622086122937cb02b94 size: 2100

$ oc new-app --image-stream=busybox --name=busybox
--> Found image d7f9d17 (4 weeks old) in image stream "myproject/busybox" under tag "latest" for "busybox"

    * This image will be deployed in deployment config "busybox"
    * The image does not expose any ports - if you want to load balance or send traffic to this component
      you will need to create a service with 'expose dc/busybox --port=[port]' later
    * WARNING: Image "myproject/busybox:latest" runs as the 'root' user which may not be permitted by your cluster administrator

--> Creating resources ...
    deploymentconfig "busybox" created
--> Success
    Run 'oc status' to view your app.
```

csrwng:
> This is a general "problem" right? This is a pretty big issue in terms of restarting Minishift in general.

Yes it means that not only the registry data is not persisted but that data in persistent volumes (at least those created by default by cluster up) is not persisted. Luckily it's an easy fix... like I said, all you need to do is add `--host-pv-dir` to the `oc cluster up` invocation. Apps and other things do come back up because those are persisted in etcd data which you are putting in the `/var/lib/minishift` directory already.

hferentschik:
> Apps and other things do come back up because those are persisted in etcd data which you are putting in the /var/lib/minishift directory already.

got you thanks. We will make sure to set _--host-pv-dir_ as part of this. 

That said, @csrwng, is there a chance that one can make the amount of persistence volumes created configurable? Is there an issue for that or should we create one.

hferentschik:
Apart from fixing the add-on and setting the right 'cluster up' flags, I am not wondering which minishift/oc command would run to discover the right hostname:port combination to use for login into the registry. 

Our initial thinking was to do something like `minishift openshift service docker-registry --url -n default` (something along these lines). This actually returns an URL, which is already sub-optimal. However, if we are going to need to add the port as well, we need something else here as well, otherwise we are ending up hard-coding something again. 

So bottom line, can we find a way to avoid having to add the port (preferred) or can we run some oc command to get the required format for logging into the registry via the Docker CLI.

Thoughts?

budhrg:
> Our initial thinking was to do something like minishift openshift service docker-registry --url -n default (something along these lines). This actually returns an URL, which is already sub-optimal. However, if we are going to need to add the port as well, we need something else here as well, otherwise we are ending up hard-coding something again.

Yes, openshift service command will give url but w/o port. 

> So bottom line, can we find a way to avoid having to add the port (preferred) or can we run some oc command to get the required format for logging into the registry via the Docker CLI.

Yes need bit of investigation here.

LalatenduMohanty:
Ok working for me on boot2docker. Finally we got it working !

+1 to take care of --host-pv-dir  in this PR.  Regarding the how to put the print the port, I am not sure at this meoment and need to dig more. 

```
$ docker push docker-registry-default.$(minishift ip).nip.io:443/myproject/ghost:latest
The push refers to a repository [docker-registry-default.192.168.64.7.nip.io:443/myproject/ghost]
c0306a6c25d7: Pushed 
09b9eec43a71: Pushed 
66e6729bcc7f: Pushed 
c9031116ff3d: Pushed 
fe6747d73cad: Pushed 
1edbd65151e2: Pushed 
6a15a9547a32: Pushed 
a43e94a16e03: Pushed 
88c9c552d8d8: Pushed 
b9eda2e140e8: Pushed 
98821eacb28b: Pushed 
596280599f68: Pushed 
5d6cbe0dbcf9: Pushed 
latest: digest: sha256:d811df4394c753a0f7db99b94f3c5d37c5db56dcfa2c0d690ba89c4567dc1cf3 size: 17426
```

hferentschik:
Btw, we will need docs as well here.

addons/expose-registry/expose-registry.addon
```diff
@@ -2,22 +2,14 @@
 # Description: Exposes the OpenShift internal registry as route
```
hferentschik: We probably want to think about a better name and description here as well

hferentschik:
I have really a hard time seeing how we make this discoverable. ATM, I don't see a way to determine the port.

If it where just the port, one could use _oc_:

    oc --as system:admin get  -o jsonpath={.spec.host} route/docker-registry -n default

But afaict the 443 needed to make this work comes out of thin air. In the end consumers would need to hard code again.

Then we are almost better of with the fixed IP 172.30.1.1. It ain't pretty, but one could do something like this:

    oc --as system:admin get -o "jsonpath={.spec.clusterIP}:{.spec.ports[*].port}" service/docker-registry -n default

At least this is explainable.

csrwng:
> That said, @csrwng, is there a chance that one can make the amount of persistence volumes created configurable? Is there an issue for that or should we create one.

@hferentschik please create one

Also, on the port discussion... the router, as installed by default, will always expose port 80 and 443. So if you're exposing a secure route the only port it could be is 443. This is unless you explicitly configure different ports for the router. And if that were the case, you can always inspect the router dc to see what port is configured for secure routes.

hferentschik:
So after much contemplating, I think we don't get around adding a new command, something like _minishift openshift registry_ which returns the right details for the user to login and use the OpenShift registry. 

This way we can also leave it open what we expose, we can go either way. The internal IP or the route.

jorgemoralespou:
I would recommend the route as it's more aligned to usage of registries.
Plus, routes are meant for external to internal communication. We should
eat or own dog food and show the best practices if we can.

praveenkumar:
> Btw, we will need docs as well here.

@hferentschik Doc part is taken care of #679 and we will update there once we have this PR in.

hferentschik:
I's prefer to make the docs part of this pull request and adjust whatever makes sense from  #679.

hferentschik:
> I would recommend the route as it's more aligned to usage of registries.

Sure, we can expose the route if it is configured, otherwise we revert to the internal IP. This is in relation to issue #758.

LalatenduMohanty:
@hferentschik @praveenkumar I am not sure how we can take PR #679 and do changes  (some git magic maybe) in this PR. We can still merge #679 and then modify it in this PR and I think that would be clean.

hferentschik:
In the end the docs need to change anyways in light of #758. It will be around the new command and less explicit about the fixed internal IP.

hferentschik:
I created a new pull request (#766) which is based on this, but also makes use of the 'echo' command. I also removed the URL meta info link and changes the name. Please have a look.

Issue #711, opened 2017-04-07
Align Middleman layout for doc site with docs.openshift.org


hferentschik:

budhrg:
Closed via https://github.com/minishift/minishift/pull/712

PR #712, opened 2017-04-07
Issue #711 Syncing Middleman layout with docs.openshift.org to have a…

hferentschik:
… L&F as in the final docs

Addresses issue #711.

Getting local docs development looking closer to the final product.


hferentschik:
@budhrg something for you ;-)

hferentschik:
To be clear, this is not 100%. It is a quick ans dirty, but it is for sure better than the old layout. Content looks as in docs.openshift.org.

docs/source/layouts/layout.erb
```diff
@@ -1,19 +1,309 @@
...
+    s.parentNode.insertBefore(gcse, s);
+  })();
+</script>
+<gcse:search></gcse:search>
+
+        </div>
+        <ul class="nav nav-sidebar">
+    <li class="nav-header">
+      <a class="" href="javascript:void(0);" data-toggle="collapse" data-target="#topicGroup0">
+        <span id="tgSpan0" class="fa fa-angle-right"></span>About
```
budhrg: Let's remove about as the link to `../../welcome/index.html` is broken.

screenshot from 2017-04-09 21-52-35

budhrg:
@hferentschik Also, find that the section is highlighted wrong in the drop down list see below.

link

> To be clear, this is not 100%

This is what you mean by not 100% ?

docs/source/layouts/layout.erb
hferentschik: I kept it intentionally, since I wanted more than a single node to expand in the navigation bar. But I don't feel strongly about this, so I removed it.

hferentschik:
> Also, find that the section is highlighted wrong in the drop down list see below.

As I said, this is not 100%. The navigation is not dynamic (yet). That would be the last missing piece of the puzzle. To make this happen though we would have to process the topic map. 

Right now the navigation is static. I removed the highlight, but this also means that nothing will be highlighted now as you click.

budhrg:
Merged and thanks! 👍

Issue #522, opened 2017-03-07
Include bashcompletion for Minishift


LalatenduMohanty:
At this point we have to type the complete command for minishift, but we should find out a way to have bash completion for the commands.
[1] http://freecode.com/projects/bashcompletion

praveenkumar:
@LalatenduMohanty how it feasible for windows binary? I think right now what we have as part of viper is good enough which suggest command if you put some wrong one.

LalatenduMohanty:
@praveenkumar May be not much for Windows CMD and Powershell. Not sure if it will be applicable to the new Windows bash shell. However it will be still useful to OS X and Linux users.

jorgemoralespou:
And should be easy to implement, like "oc completion bash|zsh" provides
output that can be directly source as bash_completion. Also "oc-cluster"
does it.

gbraad:
*Windows?*
This is not possible for Windows WSL, as the application would likely not be able to communicate with the hypervisor (it is a picoprocess). And for PowerShell? Hmmmm... haven't investigate this, as normally the cmdlets can be inspected/introspected by the runtime, but for a regular binary this is not possible.

LalatenduMohanty:
Looks like cobra provides an fuction for generating bash completion. https://godoc.org/github.com/spf13/cobra#Command.GenBashCompletion. 

Minikube already using it.

praveenkumar:
@LalatenduMohanty @hferentschik we need a milestone for it if we want to implement it.

boeboe:
+1 

also looking forward for this functionality ;)

gbraad:
See: https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/completion.go

anjannath:
+1

This is a must needed feature, looking forward to having it soon.

praveenkumar:
resolved by #1887

PR #714, opened 2017-04-08
[WIP] Issue #522 Add bash completion to Minishift

LalatenduMohanty:
Modified the source code from Minikube for Minishift.

### DIFF
--- a/cmd/minishift/cmd/completion.go
+++ b/cmd/minishift/cmd/completion.go
@@ -1,0 +1,105 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+const longDescription = `
+	Outputs minishift shell completion for the given shell (bash)
+
+	This depends on the bash-completion binary.  Example installation instructions:
+	OS X:
+		$ brew install bash-completion
+		$ source $(brew --prefix)/etc/bash_completion
+		$ minishift completion bash > ~/.minishift-completion
+		$ source ~/.minishift-completion
+
+	Fedora:
+		$ sudo dnf install bash-completion
+		$ minishift completion bash > ~/.minishift-completion
+		$ source ~/.minishift-completion
+
+	Ubuntu:
+		$ apt-get install bash-completion
+		$ source /etc/bash-completion
+		$ source <(minishift completion bash)
+
+	Additionally, you may want to output completion to a file and source in your .bashrc
+`
+
+const boilerPlate = `
+# Copyright 2016 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


cmd/minishift/cmd/completion.go
```diff
@@ -0,0 +1,105 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
```
hferentschik: 2017? And 'Red Hat\'? Or you want to keep the original since you copied from Minikube?

cmd/minishift/cmd/completion.go
```diff
@@ -0,0 +1,105 @@
...
+	Ubuntu:
+		$ apt-get install bash-completion
+		$ source /etc/bash-completion
+		$ source <(minishift completion bash)
+
+	Additionally, you may want to output completion to a file and source in your .bashrc
+`
+
+const boilerPlate = `
+# Copyright 2016 The Kubernetes Authors.
```
hferentschik: I argue there is no need to add a licence header to this completion file at all, or? And if we do, I'd say we use ours.

cmd/minishift/cmd/completion.go
```diff
@@ -0,0 +1,105 @@
...
+var completionCmd = &cobra.Command{
+	Use:   "completion SHELL",
+	Short: "Outputs minishift shell completion for the given shell (bash)",
+	Long:  longDescription,
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) != 1 {
+			fmt.Println("Usage: minishift completion SHELL")
+			atexit.Exit(1)
+		}
+		if args[0] != "bash" {
```
hferentschik: So only bash works, but we still force the user to specify a shell type? Is this to make it future proof? I'd argue that we treat no argument the same as 'bash' for now.

cmd/minishift/cmd/completion.go
```diff
@@ -0,0 +1,105 @@
...
+
+const longDescription = `
+	Outputs minishift shell completion for the given shell (bash)
+
+	This depends on the bash-completion binary.  Example installation instructions:
+	OS X:
+		$ brew install bash-completion
+		$ source $(brew --prefix)/etc/bash_completion
+		$ minishift completion bash > ~/.minishift-completion
+		$ source ~/.minishift-completion
```
hferentschik: When I try this and then do 'minishift start <tab>' I get:

    minishift start -bash: __ltrim_colon_completions: command not found

Might be that I don't have the latest bash_completion installed, not sure?

hferentschik: Also, running the command, creates such a long output that I cannot even scroll back to the beginning. I literally did not know what to do.  I wonder whether one could better this a bit, by not dumping to the console like this.

hferentschik:
See also https://github.com/kubernetes/kubernetes/issues/27876. There seem to be issues with this.

hferentschik:
Seems to be bash version dependent :-(

cmd/minishift/cmd/completion.go
LalatenduMohanty: yes, thats the reason.

budhrg:
Also, not working for me too https://gist.github.com/budhrg/1c07774e87cd23856a4563e1ff01276e

hferentschik:
IMO this is ways to fragile. 

From a usage point of view, I'd also think that the command itself should create the completion file in the current directory and the output to screen is a slight modified version of what's in the help right now. Effectively it would say: "Created file bla in current directory. <Followed of an adjusted version of the help text".

cmd/minishift/cmd/completion.go
LalatenduMohanty: Yes, thats the idea I think. Shells are different than each other.

cmd/minishift/cmd/completion.go
LalatenduMohanty: @hferentschik I think we can add our boilerplate.

cmd/minishift/cmd/completion.go
LalatenduMohanty: This should be fixable. We can create the file and put it in current working directory and giving relevant instructions to users what they can do with it.

LalatenduMohanty:
@hferentschik @budhrg I had tested this PR on OS X Sierra and Fedora 25 and it worked for me like a charm.  Let me look in to the issues you guys found and I will let you know my suggestion.  Personally I like bash-completions, so want to make it work.

LalatenduMohanty:
Closing the PR now. Will resend after improving this.

Issue #701, opened 2017-04-06
Minishift cleanup steps


ALRubinger:
1.0.0.rc.1 has some issues with our work in appdev.openshift.io.  We're needing to downgrade local envs to use Beta5.

```
[alr@tubescreamer minishift]$ rm -rf ~/.minishift/
[alr@tubescreamer minishift]$ minishift start --memory=4096
Starting local OpenShift cluster using 'kvm' hypervisor...
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso'
 40.00 MB / 40.00 MB [=================] 100.00% 0s
E0406 13:56:18.369253   27603 start.go:160] Error starting the VM: Error creating the VM. Error creating machine: Error in driver during machine creation: [Code-9] [Domain-20] operation failed: domain 'minishift' already exists with uuid 9016d6b8-dedc-465d-8d5b-9a9beb3aeb1e. Retrying.
E0406 13:59:24.599334   27603 start.go:160] Error starting the VM: Error configuring authorization on host: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (60) exceeded. Retrying.
```

Help?

jorgemoralespou:
rm -rf is not a good solution, although I've seen it suggested many times, if you have the VM already created, so only should be run after __minishift delete__

I would recommend you deleting the VM from KVM, although I don't know how to do that.

Maybe @budhrg or @gbraad or @praveenkumar

LalatenduMohanty:
@ALRubinger  as @jorgemoralespou mentioned the issue seems to be with your setup and `$ minishift delete` is the ideal way to do it. However if that does not apply and you want completely delete everything and start from fresh here are the steps

*  Check if the VM is still running `$ sudo virsh list --all`
* If it is running, stop the minishift VM i.e. `$ sudo virsh destroy minishift`
* Delete the VM `$ sudo virsh undefine minishift`
*  Delete the .minishift/machines dir i.e. $ rm -rf ~/.minishift/machines. Or as an extreme step `$rm -rf ~/.minishift`

ALRubinger:
@jorgemoralespou @LalatenduMohanty You guys fixed me up!  If you want to point me to a docs place I can contribute that (or where it already exists) I'll be happy to pay it forward.

jorgemoralespou:
That should be @LalatenduMohanty and possibly @thatdocslady

[...]

PR #717, opened 2017-04-09
Issue #701 Updated instructions for KVM wrt deleting existing VMs…

Preeticp:
…also changed TOC levels for better visibility of issues within the TOC itself.

@hferentschik, out of the 7 issues in this doc, 5 were specifically to do with the drivers, hence rather than do away with identifying issues with the drivers, I have marked out those issues specific to drivers and those that are generic, this is in line with CDK docs too.

To improve visibility of the individual issues, I have changed the TOC to include level 2 which in OpenShift docs should render similar to this, afaics: https://docs.openshift.org/latest/dev_guide/application_lifecycle/promoting_applications.html.

@LalatenduMohanty pointed out that the Uninstalling section should be a separate document from the Installing one, currently it is not exactly visible and hence easily missed out by users. This section already had the instructions provided by @LalatenduMohanty, I have merely added a link to the KVM section. Let me know your thoughts on separating the Uninstall section into a separate doc.

### DIFF
--- a/docs/source/using/troubleshooting.adoc
+++ b/docs/source/using/troubleshooting.adoc
@@ -3,7 +3,7 @@
 :icons:
 :toc: macro
 :toc-title:
-:toclevels: 1
+:toclevels: 2
 
 toc::[]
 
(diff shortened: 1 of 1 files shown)


docs/source/getting-started/installing.adoc
```diff
@@ -128,4 +128,4 @@ $ rm -rf ~/.kube
 
 .  With your hypervisor management tool, confirm that there are no
 remaining artifacts related to the Minishift VM. For example, if you use
-KVM, you need to run the `virsh` command.
+KVM, you need to run the `virsh` command to avoid the issue link:../using/troubleshooting{outfilesuffix}#domain-minishift-already-exists[Error starting the VM: ...operation failed: domain 'minishift' already exists...].
```
LalatenduMohanty: @Preeticp I do not think this is required. 914172
Users who want to know uninstall  steps for Minishift should not get redirected to  troubleshooting issue `Error starting the VM: ...operation failed: domain 'minishift' already exists`.

LalatenduMohanty:
Apart from the single comment, rest LGTM.

docs/source/getting-started/installing.adoc
budhrg: Ya agree

Preeticp:
@LalatenduMohanty, removed the link.

LalatenduMohanty:
@Preeticp #722

docs/source/using/troubleshooting.adoc
```diff
@@ -13,11 +13,11 @@ toc::[]
 This section contains solutions to common problems that you might
 encounter while using Minishift.
 
-[[troubleshooting-kvm]]
-== KVM driver
+[[troubleshooting-drivers]]
+== Troubleshooting problems with drivers
```
hferentschik: I'd argue that one first has general issues and then driver issues. WDYT?

Also "Troubleshooting" seems redundant for this heading and make is imo unnecessarily long.

docs/source/using/troubleshooting.adoc
```diff
@@ -51,21 +51,39 @@ $ systemctl enable virtlogd
 ----
 
 [[domain-minishift-already-exists]]
-=== Error starting the VM: ... operation failed: domain 'minishift' already exists ...
+=== KVM: `Error starting the VM: ...operation failed: domain 'minishift' already exists...`
```
hferentschik: This title is too long. We should just add the essential part of the error message, e.g. 

KVM:  'minishift' already exists...

docs/source/using/troubleshooting.adoc
```diff
@@ -51,21 +51,39 @@ $ systemctl enable virtlogd
...
++
+----
+ $ rm -rf ~/.minishift/machines
+----
+
+In case all of this fails, you may want to link:../getting-started/installing{outfilesuffix}#uninstall-instructions[uninstall Minishift] and do a fresh install of Minishift.
 
 [[create-vmnet-interface-permission]]
-=== Error: could not create vmnet interface, permission denied or no entitlement
+=== xhyve: `Error: could not create vmnet interface, permission denied or no entitlement`
```
hferentschik: Also too long, maybe "xhyve: could not create vmnet interface"

docs/source/using/troubleshooting.adoc
```diff
@@ -91,23 +109,21 @@ system.
...
 
-[[troubleshooting-authentication]]
-== Users and authentication
+
+[[troubleshooting-general-problems]]
+== Troubleshooting general problems
 
 [[special-characters-passwords]]
-=== Some special characters cause passwords to fail
+=== Users and authentication: Some special characters cause passwords to fail
```
hferentschik: I think 'Users and authentication:' is not required. "special characters cause passwords to fail" should do.

docs/source/using/troubleshooting.adoc
```diff
@@ -116,11 +132,8 @@ cause passwords to fail.
 Workaround: When creating and entering passwords, wrap the string with
 single quotes in the following format: `<password>`
 
-[[troubleshooting-snapshots]]
-== Snapshots with virsh
-
 [[minishift-delete-fails-undefine-snapshots]]
-=== `minishift delete` fails to undefine snapshots of running instances made using virsh on KVM/libvirt
+=== Snapshots with virsh: `minishift delete` fails to undefine snapshots of running instances made using virsh on KVM/libvirt
```
hferentschik: "Undefining virsh snapshots fail"

docs/source/using/troubleshooting.adoc
```diff
@@ -51,21 +51,39 @@ $ systemctl enable virtlogd
...
++
+----
 $ sudo virsh undefine minishift
 ----
 
-[[troubleshooting-xhyve]]
-== xhyve driver
+. Delete the `.minishift/machines` directory using:
++
+----
```
hferentschik: I'd start to consistently mark the type of the source examples, in this case we should add 

'[source,sh]'

(this applies to all commands in this page)

docs/source/using/troubleshooting.adoc
Preeticp: I thought of putting general issues before the Driver issues similar to CDK docs, but then retained the existing order considering that drivers are prerequisites and an issue with them will prevent further use of Minishift. Also, the Driver issues seem like showstoppers and there are 5 of them, whereas we only have 1 general issue out of which only 1 is most likely to affect users often. If you think reversing the order is more useful, I will change it.

Preeticp:
@hferentschik rebasing did not help me get the L&F of the final site, maybe I am missing something. One suggestion: do you think it would be better to make the issues like 'KVM: Error creating new host: dial tcp: missing address' as heading level 4 and increase the TOC level to 3?
Despite the changes, the titles still seem long.

hferentschik:
>  rebasing did not help me get the L&F of the final site, 

You also need to get the latest changes from master and then rebase your pull request on top of this. Did you do this? Are you still not able to see the doc changes in their final form? I really think you need to see it.

docs/source/using/troubleshooting.adoc
```diff
@@ -78,6 +108,7 @@ Reset the `minishift`-specific IP database, make sure to remove the
 `minishift` entry section from the `dhcpd_leases` file, and reboot your
 system.
 
+[source,sh]
```
hferentschik: this is not shell source, we can remove it from there

hferentschik:
Rebased and merged, thanks

Issue #708, opened 2017-04-07
Clarify warning message around setting config options


dustymabe:
```
[student@workstation ~]$ minishift config set memory 10240
To apply the changes, you must delete the current VM with `minishift delete` and start a new VM with `minishift start`.
[student@workstation ~]$ minishift config set cpus 4
To apply the changes, you must delete the current VM with `minishift delete` and start a new VM with `minishift start`.
```

I get these messages even if there is no "current VM". We should either not print these messages out if there is no current VM or we should re-word so that is doesn't imply that there is a current VM.

hferentschik:
> I get these messages even if there is no "current VM". We should either not print these messages out if there is no current VM or we should re-word so that is doesn't imply that there is a current VM.

I think there is even another issue for that.

LalatenduMohanty:
This is a duplicate of https://github.com/minishift/minishift/issues/471

hferentschik:
Merged via pull request #731

Issue #471, opened 2017-02-23
Issues with minishift config command


LalatenduMohanty:
I observed following issues with `minishift config` command. 

1. Irrespective of the VM state i.e. even if machine is deleted, `minishift config set memory <>`  returns the same output/message. 

e.g. 
```
$ minishift config  set memory 2000
To apply the changes, you must delete the current VM with `minishift delete` and start a new VM with `minishift start`.
```
2. There is no mention of unit of configurable properties mentioned in the usage.  For example if I want to set 2GB RAM ,  there is no mention what should be the value of the RAM i.e. 2GB , 2048 MB or just 2048. 

3.  I think usage of `minishift config set --help` should have the configurable properties names etc. But we can discuss this.

```
$ minishift config
Modifies Minishift configuration properties. Some of the configuration properties are equivalent
to the options that you set when you run the minishift start command.

Configurable properties (enter as SUBCOMMAND): 

 * vm-driver
 * v
 * cpus
 * disk-size
 * host-only-cidr
 * memory
 * show-libmachine-logs
 * log_dir
 * openshift-version
 * iso-url
 * WantUpdateNotification
 * ReminderWaitPeriodInHours

Usage:
  minishift config SUBCOMMAND [flags]
  minishift config [command]

Available Commands:
  get         Gets the value of a configuration property from the Minishift configuration file.
  set         Sets the value of a configuration property in the Minishift configuration file.
  unset       Clears the value of a configuration property in the Minishift configuration file.
  view        Display the properties and values of the Minishift configuration file.

Global Flags:
      --alsologtostderr                  log to standard error as well as files
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory
      --logtostderr                      log to standard error instead of files
      --password string                  Password for the virtual machine registration.
      --show-libmachine-logs             Show logs from libmachine.
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
      --username string                  User name for the virtual machine registration.
  -v, --v Level                          log level for V logs
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging

Use "minishift config [command] --help" for more information about a command.

$ minishift config  set --help
Sets the value of one or more configuration properties in the Minishift configuration file.
These values can be overwritten by flags or environment variables at runtime.

Usage:
  minishift config set PROPERTY_NAME PROPERTY_VALUE [flags]

[...]

PR #731, opened 2017-04-11
Issue #708 Rewording around warning message for config set

praveenkumar:
This patch is only for rewording around config set method.

### DIFF
--- a/cmd/minishift/cmd/config/validations.go
+++ b/cmd/minishift/cmd/config/validations.go
@@ -24,7 +24,9 @@ import (
 	"strconv"
 
 	units "github.com/docker/go-units"
+	"github.com/docker/machine/libmachine"
 
+	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 )
 
@@ -38,7 +40,19 @@ func IsValidDriver(string, driver string) error {
 }
 
 func RequiresRestartMsg(string, string) error {
-	fmt.Fprintln(os.Stdout, "To apply the changes, you must delete the current VM with `minishift delete` and start a new VM with `minishift start`.")
+	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+	defer api.Close()
+
+	_, err := cluster.CheckIfApiExistsAndLoad(api)
+	if err != nil {
+		fmt.Fprintln(os.Stdout, "No Minishift VM exists. Settings will be applied on next `minishift start`")
+	} else {
+		fmt.Fprintln(os.Stdout, "Settings are only applied when a new Minishift instance is created. "+
+			"You currently have an existing VM. "+
+			"To apply the changes, "+
+			"you must delete the current VM with `minishift delete` "+
+			"and start a new VM with `minishift start`.")
+	}
 	return nil
 }
 


cmd/minishift/cmd/config/validations.go
```diff
@@ -38,7 +38,9 @@ func IsValidDriver(string, driver string) error {
 }
 
 func RequiresRestartMsg(string, string) error {
-	fmt.Fprintln(os.Stdout, "To apply the changes, you must delete the current VM with `minishift delete` and start a new VM with `minishift start`.")
+	fmt.Fprintln(os.Stdout, "To apply the changes, "+
+		"you must delete the current VM with `minishift delete` if VM exist "+
```
LalatenduMohanty: s/exist/exists/

hferentschik: I think we can do better than this. We literally should check whether the machine exists. If so we write print:

"No Minishift VM defined. Settings will be applied on next 'minishift start'"

If the machine does exist:

"Settings are only applied when a new Minishift instance is created. You currently have an existing VM. To apply the changes, you must delete the current VM with `minishift delete` and start a new VM with `minishift start`."

LalatenduMohanty: @hferentschik +1.

hferentschik: There are several ways to go about finding out whether the VM exits.

- Without modifying the current callback signature
  - You can just do what we do in other commands as well. Directly create the api `libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))`
  - We can put something into root.go. In _persistentPreRun_ we determine once for all commands whether the machine exists and set a variable accordingly. This variable can be queried by the other commands. AS a side effect we can remove some VM existence checks and we also get this into one single place. 
- with changing the callback signature
   - You can add additional parameters to the callback method, for example the name and of the VM. This way we stop the proliferation of using the constants file.

Either way, there are several ways to do that.

LalatenduMohanty: I had raised the issue as part of https://github.com/minishift/minishift/issues/471

thatdocslady: So is it safe to review this message now, or will we be changing the command itself to first check if the vm exists?

cmd/minishift/cmd/config/validations.go
```diff
@@ -38,7 +40,19 @@ func IsValidDriver(string, driver string) error {
...
+	defer api.Close()
+
+	_, err := cluster.CheckIfApiExistsAndLoad(api)
+	if err != nil {
+		fmt.Fprintln(os.Stdout, "No Minishift VM exists. Settings will be applied on next `minishift start`")
+	} else {
+		fmt.Fprintln(os.Stdout, "Settings are only applied when a new Minishift instance is created. "+
+			"You currently have an existing VM. "+
+			"To apply the changes, "+
+			"you must delete the current VM with `minishift delete` if VM exist "+
```
hferentschik: no "if VM exists" - we know it does, right?

praveenkumar:
> You can just do what we do in other commands as well. Directly create the api libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))

@hferentschik this is what I chose for this issue right now.
    
> We can put something into root.go. In persistentPreRun we determine once for all commands whether the machine exists and set a variable accordingly. This variable can be queried by the other commands. AS a side effect we can remove some VM existence checks and we also get this into one single place.

@hferentschik As you said we might have to remove some VM existance checks from different command so I think we should do it after our release.

cmd/minishift/cmd/config/validations.go
praveenkumar: @hferentschik Right my mistake. Updated.

hferentschik:
Merged. I changed the message and included the name of the setting being set as well.

PR #732, opened 2017-04-11
Issue #693 edited doc strings for hostfolder command

### LINKED ISSUE #693: Edit the docs strings for host folder commands
Need to review and edit the doc strings for the minishift hostfolder commands. This issue is related to issue #655 but they will be handled separately.

thatdocslady:
Fixes #693 

This PR provides proofreading of the doc strings for the hostfolder command and sub-commands.

NOTE: Unlike other commands (i.e. addon, config), the actual file names for the sub-commands don't include the parent command in the file name (i.e. "hostfolder_add.go" vs. "add.go"). Not sure if this is a problem but it isn't consistent.

cc @gbraad @hferentschik  @LalatenduMohanty

### DIFF
--- a/cmd/minishift/cmd/hostfolder/add.go
+++ b/cmd/minishift/cmd/hostfolder/add.go
@@ -34,8 +34,8 @@ var (
 
 var hostfolderAddCmd = &cobra.Command{
 	Use:   "add HOSTFOLDER_NAME",
-	Short: "Add a host folder definition",
-	Long:  `Add a host folder definition that can be mounted to a running cluster`,
+	Short: "Adds a host folder definition.",
+	Long:  `Adds a host folder definition. The defined host folder can be mounted to a running OpenShift cluster.`,
 	Run: func(cmd *cobra.Command, args []string) {
 
 		var err error = nil
@@ -44,7 +44,7 @@ var hostfolderAddCmd = &cobra.Command{
 			err = hostfolderActions.SetupUsers(true)
 		} else {
 			if len(args) < 1 {
-				fmt.Fprintln(os.Stderr, "usage: minishift hostfolder add HOSTFOLDER_NAME")
+				fmt.Fprintln(os.Stderr, "Usage: minishift hostfolder add HOSTFOLDER_NAME")
 				atexit.Exit(1)
 			}
 			err = hostfolderActions.Add(args[0], !instanceOnly)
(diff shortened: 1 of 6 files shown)


gbraad:
This command is the same with `mount` and `umount` on a system. This is
purposely typed this way.

cmd/minishift/cmd/hostfolder/add.go
```diff
@@ -59,10 +59,10 @@ var hostfolderAddCmd = &cobra.Command{
 
 func init() {
 	HostfolderCmd.AddCommand(hostfolderAddCmd)
-	hostfolderAddCmd.Flags().BoolVarP(&instanceOnly, "instance-only", "", false, "Define host folder only for the current cluster instance.")
+	hostfolderAddCmd.Flags().BoolVarP(&instanceOnly, "instance-only", "", false, "Defines the host folder only for the current cluster.")
 
 	// Windows-only
 	if runtime.GOOS == "windows" {
-		hostfolderAddCmd.Flags().BoolVarP(&usersShare, "users-share", "", false, "Define host folder for the Users share on a Windows host.")
+		hostfolderAddCmd.Flags().BoolVarP(&usersShare, "users-share", "", false, "Defines the shared `Users` folder as the host folder on a Windows host.")
```
hferentschik: why the backticks? Did you try this on Windows?

hferentschik:
> There is a typo in the actual syntax of the unmount sub-command (it's spelled "umount" in the code and it should be "unmount"). Is it way too late to fix that?

it is supposed to be 'umount'

cmd/minishift/cmd/hostfolder/add.go
thatdocslady: The single quotes are for the parsed docs, not for being copied/pasted along with the folder name. This way when we generate the command reference the folder name will show up in the correct formatting.

gbraad: It should not cause issues during execution, but the caution is in place.

cmd/minishift/cmd/hostfolder/add.go
```diff
@@ -59,10 +59,10 @@ var hostfolderAddCmd = &cobra.Command{
 
 func init() {
 	HostfolderCmd.AddCommand(hostfolderAddCmd)
-	hostfolderAddCmd.Flags().BoolVarP(&instanceOnly, "instance-only", "", false, "Define host folder only for the current cluster instance.")
+	hostfolderAddCmd.Flags().BoolVarP(&instanceOnly, "instance-only", "", false, "Defines the host folder only for the current cluster.")
```
gbraad: "current OpenShift cluster"

cmd/minishift/cmd/hostfolder/add.go
hferentschik: > It should not cause issues during execution, but the caution is in place.

Exactly my thinking as well.

hferentschik: > The single quotes are for the parsed docs, not for being copied/pasted along with the folder name.

Not sure what you are saying here, but my guess is that you are adding it, because in the docs the backticks then have a special meaning. If this is the case, a big -1.

This help messages are primarily for the console. It just happens to be that we also generate markdown/asciidoc from it. The primary output we are targeting it the console.

For me as a console user I would wonder what the backticks are about. They are not common practice and they are potentially in the way, for example when I want to cut&paste the message.

So let's please remove them.

thatdocslady:
@gbraad @hferentschik fixed the strings per your feedback, I get your point about the umount thing. Can you please address my question about the file names? Is this something that we are ok with?

hferentschik:
> Can you please address my question about the file names? Is this something that we are ok with?

file names are not a problem

gbraad:
@hferentschik Actually... based on the `umount` command, I haven't implemented a `-a`/`--all` option for this. We might want this too?

hferentschik:
Merged, thanks

PR #733, opened 2017-04-11
Issue #692 edited doc strings for addons command

### LINKED ISSUE #692: Edit doc strings for add-ons commands
Following work on the add-ons documentation, it was decided that it would be better practice to handle the docs and the doc strings separately. This issue covers only the doc strings, and issue #664 will cover the doc topics.

thatdocslady:
Fixes #692 

This PR includes proofreading for doc strings of the addons command and sub-commands.

NOTES:

1. Intentionally left out the *_test.go files since they will not be user-facing texts.

2. In the addons_list sub-command there is a mention of "template" but only in the error message, and it's not controlled or described anywhere. I'm wondering if we should write something about it before the users get hit with an error?

### DIFF
--- a/cmd/minishift/cmd/addon/addon_enable.go
+++ b/cmd/minishift/cmd/addon/addon_enable.go
@@ -27,21 +27,21 @@ const (
 	priorityFlag   = "priority"
 	addOnConfigKey = "addons"
 
-	emptyEnableError       = "An addon name needs to be specified. Use `minishift addons list` to view installed addons."
-	noAddOnToEnableMessage = "No addon with name %s installed"
+	emptyEnableError       = "You must specify an add-on name. Use `minishift addons list` to view installed add-ons."
+	noAddOnToEnableMessage = "No add-on with the name %s is installed."
 )
 
 var priority int
 
 var addonsEnableCmd = &cobra.Command{
 	Use:   "enable ADDON_NAME",
-	Short: "Enables the specified addon.",
-	Long:  "Enables the specified addon to be run after cluster creation.",
+	Short: "Enables the specified add-on.",
+	Long:  "Enables the specified add-on and applies the add-on the next time a cluster is created.",
 	Run:   runEnableAddon,
 }
 
 func init() {
-	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of this addon during addon execution.")
+	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of the add-on when it is applied.")
 	AddonsCmd.AddCommand(addonsEnableCmd)
 }
 
@@ -65,7 +65,7 @@ func runEnableAddon(cmd *cobra.Command, args []string) {
 func enableAddon(addOnManager *manager.AddOnManager, addonName string, priority int) {
 	addOnConfig, err := addOnManager.Enable(addonName, priority)
 	if err != nil {
-		fmt.Println(fmt.Sprintf("Unable to enable plugin %s: %s", addonName, err.Error()))
+		fmt.Println(fmt.Sprintf("Unable to enable add-on %s: %s", addonName, err.Error()))
 		atexit.Exit(1)
 	}
 
--- a/cmd/minishift/cmd/addon/addons_list.go
+++ b/cmd/minishift/cmd/addon/addons_list.go
@@ -51,13 +51,13 @@ type DisplayAddOn struct {
 
 var addonsListCmd = &cobra.Command{
 	Use:   "list",
-	Short: "Lists all installed Minishift addons",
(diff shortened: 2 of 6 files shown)


cmd/minishift/cmd/addon/addon_enable.go
```diff
@@ -27,21 +27,21 @@ const (
...
-	Short: "Enables the specified addon.",
-	Long:  "Enables the specified addon to be run after cluster creation.",
+	Short: "Enables the specified add-on.",
+	Long:  "Enables the specified add-on and instructs Minishift to run the add-on the next time a cluster is created.",
 	Run:   runEnableAddon,
 }
 
 func init() {
-	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of this addon during addon execution.")
+	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of the add-on at runtime.")
```
hferentschik: -1, this is not the priority at runtime. "Add-on application" time maybe.

thatdocslady: @hferentschik -1, "application" is ambiguous. I guess we should be more explicit what does "execution" mean, because from what it looks like here you launch/run the add-ons when you create a cluster and then the add-ons continues to run as long as the cluster is running. @budhrg commented that we need examples here, so maybe that'll help? Let's discuss.

hferentschik: > 1, "application" is ambiguous.

I don't think 'application' is ambiguous in this case. It is when the add-ons get applied. So something like this works for me as well: "The priority of the add-on when it gets applied"

> because from what it looks like here you launch/run the add-ons when you create a cluster and then the add-ons continues to run as long as the cluster is running. 

Add-ons get applied once and they enable/provision a certain feature. This feature is available at runtime. What we are talking about here is the priority/order in which the add-ons are getting applied. 

So, lets go with "The priority of the add-on when it gets applied"

thatdocslady: Sounds good, when it's in verb form it's not ambiguous :)

thatdocslady:
@hferentschik implemented feedback. Can you please see my question about the templates? I'm guessing these are the JSON files, but we don't actually mention them anywhere in the command reference strings. Should we?

cmd/minishift/cmd/addon/util.go
```diff
@@ -38,7 +38,7 @@ func GetAddOnManager() *manager.AddOnManager {
 	addOnConfigs := getAddOnConfiguration()
 	m, err := manager.NewAddOnManager(constants.MakeMiniPath("addons"), addOnConfigs)
 	if err != nil {
-		glog.Errorln("Unable to initialise addon manager.", err)
+		glog.Errorln("Unable to start the add-on manager.", err)
```
hferentschik: -1, we are not starting anything. "Initialise" is the right word.

cmd/minishift/cmd/addon/util.go
```diff
@@ -48,7 +48,7 @@ func GetAddOnManager() *manager.AddOnManager {
 func GetExecutionContext(ip string, routingSuffix string, ocPath string, kubeConfigPath string, sshCommander provision.SSHCommander) *command.ExecutionContext {
 	context, err := command.NewExecutionContext(ocPath, kubeConfigPath, sshCommander)
 	if err != nil {
-		glog.Errorln("Unable to initialise execution context.", err)
+		glog.Errorln("Unable to start the execution context.", err)
```
hferentschik: Same as above. We are initialising the execution context, aka preparing it to be used. We are not starting anything.

cmd/minishift/cmd/addon/addons_list.go
```diff
@@ -51,13 +51,13 @@ type DisplayAddOn struct {
...
-	Short: "Lists all installed Minishift addons",
-	Long:  "Lists all installed Minishift addons as well as there current status (enabled/disabled)",
+	Short: "Lists all installed Minishift add-ons.",
+	Long:  "Lists all installed Minishift add-ons and their current status, such as enabled/disabled.",
 	Run:   runListCommand,
 }
 
 func init() {
-	addonsListCmd.Flags().BoolVar(&verbose, "verbose", false, "A more verbose format of the output including the addon description.")
+	addonsListCmd.Flags().BoolVar(&verbose, "verbose", false, "Runs the command with a more verbose format of the output that includes the add-on description.")
```
hferentschik: "Runs the command ..." -> Prints the add-on list with a more verbose format of the output that includes the add-on description.

thatdocslady:
@hferentschik updated.

hferentschik:
Merged, thanks

Issue #664, opened 2017-04-04
Edit the add-ons docs


thatdocslady:
Following the merge of the add-ons functionality, the docs still need some proofreading:

- Propose to change "Addons" to "Add-ons" to follow grammatical conventions.
- Propose to move the whole section into its own topic, currently it's a part of "managing minishift" and it's making the page run really long. IMO it deserves its own topic.
- Some minor grammar/formatting fixes

thatdocslady:
@LalatenduMohanty @hferentschik @Preeticp WDYT?

hferentschik:
> Propose to change "Addons" to "Add-ons" to follow grammatical conventions.

+1

> Propose to move the whole section into its own topic, currently it's a part of "managing minishift" and it's making the page run really long. IMO it deserves its own topic.

+1, in fact we should move host folder out as well, especially since it gets updated now as well. The rest can stay I guess.

We might want to think about a better way of "Managing" for the remaining content in the chapter. It's ok, but somehow also not quite hard. I cannot put a finger on a proper title. Any ideas?

> Some minor grammar/formatting fixes

Sure.

Added this issue to the next milestone.

thatdocslady:
@hferentschik 

> We might want to think about a better way of "Managing" for the remaining content in the chapter. It's ok, but somehow also not quite hard. I cannot put a finger on a proper title. Any ideas?

Looking at the different sections in this topic, I agree that it seems a little bit like a "catch-all" bucket for now. I have some ideas, let me know what you think:

1. Remove the minishift start/stop/delete sections. TBH they basically repeat what's in the command reference, and the quickstart also has the start command in more details. Alternatively we can move those sections to the quickstart section itself, above the example procedure, and label them as "common commands". It makes more sense to have them there IMO.

2. Assuming that we are extracting the host folders and add-ons content into their own topics, we're basically left with a bunch of configuration sections, so maybe we can rename the topic to "Configuring Minishift", and then it's more accurately reflective of the content.

3. If we proceed with no.2 here, would it make sense to keep the proxy information here? Or should we extract it to some separate topic and call it "Networking" or something? Are we planning to document other aspects of networking in the future? Could be more useful for the modular docs approach.

thatdocslady:
@hferentschik in addition to the doc updates, I see that we also need to proofread the doc strings in the actual addon commands. Should I open a separate issue for that or can I submit it all in a single PR?

thatdocslady:
[...]

PR #735, opened 2017-04-11
Issue #664: Extracted and edited add-ons docs

thatdocslady:
Fixes issue #664 

This PR handles the add-ons docs:

1. Extracted the content from the "Managing Minishift" topic into a separate topic.
2. Light proofread and formatting fixes.

NOTE: I noticed that the terminology for applying/executing an add-on isn't consisten (for example, if you see PR #733 you will notice that the term used there is "execute" and in this topic we say "apply"). We should decide on a verb and stick with it. If it helps, "execute" implies that you start something and it keeps running until you stop it, and "apply" means a one-off action that leaves a permanent result that you don't need to touch again. Maybe we can even borrow the start/stop verbs from systemctl (especially since we already use enable/disable?


hferentschik:
>  If it helps, "execute" implies that you start something and it keeps running until you stop it

I tend to disagree

thatdocslady:
@hferentschik updated per feedback, you're right I forgot to update the topic map and now the index is also updated.

hferentschik:
> updated per feedback, you're right I forgot to update the topic map and now the index is also updated.

and _layout.erb_ on _layouts_. If you don't want to do it, I can care of it.

docs/source/_topic_map.yml
```diff
@@ -25,6 +25,8 @@ Topics:
         File: managing-minishift
       - Name: Interacting with Openshift
         File: interacting-with-openshift
+      - Name: Add-ons
```
hferentschik: This should go above "Interacting with OpenShift". In the order of things, it comes before interacting with OpenShift. That said, it should also be "OpenShift"

docs/source/using/addons.adoc
```diff
@@ -0,0 +1,220 @@
...
+----
+$ minishift addons list --verbose=true
+----
+
+This command prints a list of installed add-ons. You should at least see the _anyuid_ addon listed.
+This is an important add-on that allows you to run images that do not use a pre-allocated UID.
+By default, this is not allowed in OpensShift.
+
+[[section-enabling-disabling-addons]]
+== Enabling and disabling addons
```
hferentschik: add-ons

docs/source/using/addons.adoc
```diff
@@ -0,0 +1,220 @@
...
+
+Add-ons are enabled with the link:../command-ref/minishift_addons_enable{outfilesuffix}[`minishift addons enable`]
+command and disabled with the link:../command-ref/minishift_addons_disable{outfilesuffix}[`minishift addons disable`] command.
+The following examples show how to enable and disable the _anyuid_ add-on.
+
+[[example-enable-anyuid]]
+.Example: Enabling the anyuid addon
```
hferentschik: add-on

docs/source/using/addons.adoc
```diff
@@ -0,0 +1,220 @@
...
+----
+
+This command extracts the default add-ons to the add-on installation directory
+(`$MINISHIFT_HOME/addons`). To view the list of installed add-ons, you can then run:
+
+----
+$ minishift addons list --verbose=true
+----
+
+This command prints a list of installed add-ons. You should at least see the _anyuid_ addon listed.
```
hferentschik: add-on

thatdocslady:
@hferentschik updated.

docs/source/using/addons.adoc
```diff
@@ -0,0 +1,221 @@
...
+useful in cases where you know that a command such as `oc` might take a few seconds before a certain
+resource can be queried.
+
+NOTE: Trying to add a command that is not listed will cause an error when the add-on gets parsed.
+
+[[addon-variable-interpolation]]
+=== Variable interpolation
+
+Minishift also allows the use of variables as part of the supported commands.
+Variables have the from _#{<variable-name>}_. <<example-addon-variable>>
```
hferentschik: This is another broken reference now

hferentschik:
Merged, making an additional commit to fix the broken links.

PR #736, opened 2017-04-11
Issue #649 Introducing Tee struct and cmd helper functions

hferentschik:
Based on pull request #685.
Addresses issue #649.

### DIFF
--- a/pkg/testing/cli/tee.go
+++ b/pkg/testing/cli/tee.go
@@ -1,0 +1,113 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cli
+
+import (
+	"bytes"
+	"fmt"
+	"io"
+	"os"
+	"sync"
+)
+
+type Tee struct {
+	waitGroup    *sync.WaitGroup
+	OrigStdout   *os.File
+	OrigStderr   *os.File
+	Stdout       *os.File
+	Stderr       *os.File
+	StdoutBuffer *bytes.Buffer
+	StderrBuffer *bytes.Buffer
+	silent       bool
+}
+
+// NewTee splits os.Stdout and os.Stderr using a pipe. While the tee is not closed data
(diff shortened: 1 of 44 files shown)


hferentschik:
@praveenkumar please have a look at this. It is based on your changes, but changes the way the tests work with stdout and stderr streams. It introduces a Tee struct for that.

I also created some helper function for interacting with the VM. The idea is the same as with collecting the _Exit_ call in one place. Reduce code duplication and have a single point where a specific message is defined.

praveenkumar:
@hferentschik Why it show single test case for addon when we have around 4-5 test cases.

```
$ go test -v github.com/minishift/minishift/cmd/minishift/cmd/addon/
=== RUN   Test_addon_name_must_be_specified_for_disable_command
An addon name needs to be specified. Use `minishift addons list` to view installed addons.
ok  	github.com/minishift/minishift/cmd/minishift/cmd/addon	0.006s
```

pkg/testing/cli/tee.go
```diff
@@ -0,0 +1,110 @@
...
+	var stderrBuffer bytes.Buffer
+	go f(stderrRead, origStderr, &stderrBuffer)
+	return &Tee{waitGroup, origStdout, origStderr, stdoutWrite, stderrWrite, &stdoutBuffer, &stderrBuffer}, nil
+}
+
+func (t *Tee) Close() {
+	t.Stdout.Close()
+	t.Stderr.Close()
+
+	t.waitGroup.Wait()
```
praveenkumar: I think we need to reorder it because `Wait()` will make sure all the waitgroups are executed and it's counter become zero. So if we close `t.Stdout` before then might be some operation which happening in the go-routine might fail. We don't observe it currently becasue our stdout/stderr stream string are very small.

```
t.waitGroup.Wait()

t.Stdout.Close()
t.Stderr.Close()
```

hferentschik: fair enough

hferentschik: Actually not so sure. What if an io operations gets stuck. Closing the stream will make progress one way or another. If it is stuck and we continue to wait, nothing is going to happen.

hferentschik:
> Why it show single test case for addon when we have around 4-5 test cases.

bugger, good catch. There is a bug. Let me push an update later.

hferentschik:
@praveenkumar, I pushed an update. There was a bug in the exit handler framework, in a sense that if a test file had multiple tests and several used the exit handler framework, only the first test would run. I rectified that and now all tests are run :-) Also needed to fix a couple of tests which did not pass. 

Awesome catch!

hferentschik:
BTW, I silenced the stdout and stderr to keep the test results more readable, but you can verify by looking at the tests being run. Or one can within a test, enable the test output for debugging purposes while working on the test.

pkg/testing/cli/tee.go
praveenkumar: Let's have it like that, we can discuss that later but right now for our use-case it's going to work for sure.

hferentschik:
thanks for merging

Issue #305, opened 2017-01-17
Allow to pass through custom driver options as part of 'minishift start'


hferentschik:
Some drivers support additional arguments to for example configure/enable host folder mounts or to select a switch (Hyper-V). AFAICT, there is no mechanism in Minishift atm which allows for passing custom arguments to the driver.

hferentschik:
This has to be in some form generic, since we cannot add all the various options to the Minishift CLI. 

One thing I am considering is to drive this purely through persistent configuration and _config.json_. Given that we are targeting advanced use cases this might be a good trade-off. The idea would be to have a hypervisor specific section in the config, eg _hyper-v_ with key/value pairs of custom options. At startup we would then check whether for the selected driver there are any custom options configured. If so we would apply them.

hferentschik:
Actually we can already pass through options. The way the drivers handle this is by taking parameters via environment variables. See also https://github.com/zchee/docker-machine-driver-xhyve/blob/master/xhyve/xhyve.go and https://github.com/docker/machine/blob/master/drivers/hyperv/hyperv.go. 

Not sure whether this is enough or whether we should put something around this. At the very least it needs documenting.

gbraad:
@hferentschik have you been able to verify the use of the environment variables? I tried just now, and it seems they are ignored.

hferentschik:
See https://github.com/minishift/minishift/issues/590#issuecomment-293397120

hferentschik:
Using the algorithm described in issue #590, we should be able to pass any of the environment variables a driver x supports through to the actual driver. See for example https://docs.docker.com/machine/drivers/hyper-v/ for the Hyper-V options.

Obviously there is an overlap here with the options one can set via the command line or the config file. A priority order needs to be determined. If a user sets a driver specific environment variable, when does it get applied. Do we first apply the environment variables and then the settings from the command line or vice versa. 

We can also consider whether we want the user to prefix the variables with 'MINISHIFT_' This would avoid potential conflicts when users set these variables for the use with docker-machine as well. Also it might be more aligned with the current use of Minishift env variables. 

[...]

PR #738, opened 2017-04-11
[WIP] Issue #590 Allow hyperv virtual switch as start option and env variable

### LINKED ISSUE #590: Allow selection of virtual switch for use with Hyper-V
Related to: #589

budhrg:
Fix #590 

Tested with :
- `MINISHIFT_HYPERV_VIRTUAL_SWITCH`  environment variable
- `--hyperv-virtual-switch` option in start command
- `minishift config set hyperv-virtual-switch switch_name`

Above three are working as expected.

Working on tests !

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -109,6 +109,10 @@ var settings []Setting = []Setting{
 		set:  SetString,
 	},
 	{
+		name: "hyperv-virtual-switch",
+		set:  SetString,
+	},
+	{
 		name: config.WantUpdateNotification,
 		set:  SetBool,
 	},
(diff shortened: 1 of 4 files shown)


hferentschik:
I don't get it? What happened to the generic approach? I thought we agreed that we try that or at least find out why it is not working. It seems wrong to add a driver specific option this way and it does not scale at all. 

If we merge this, how are we going about, if we find a generic solution to the problem? Adding is easy, taking something away once you exposed it as an API not so much.

hferentschik:
Unless someone gives me an explanation on why a generic approach is not possible, this is a no go.

hferentschik:
I think we can close this pull request. We are looking first at implementing a generic solution as part of #305

budhrg:
> I think we can close this pull request. We are looking first at implementing a generic solution as part of #305

👍

Issue #696, opened 2017-04-06
Need admin user with cluster-admin role


burrsutter:
./minishift version
    Minishift version: 1.0.0-beta.5
    CDK Version: 3.0.0-beta.3

Need to login into the web UI/console

 https://screencast.com/t/YxEIldeXNa

jorgemoralespou:
@burrsutter 

With minishift 1.0.0.rc.1 which has addons and sudo for developer, it's easy to do it yourself. There is already an issue to create an admin user by default, although this mechanism is not secure at all, neither current login.

Just do:

````
oc adm policy add-cluster-role-to-user cluster-admin admin --as=system:admin
````
This creates gives cluster-role to a user named admin. Since minishift uses anypasswd to authenticate, just login with a user named admin (with any password)

**Comment by LalatenduMohanty** [comment]
Here is the steps for addons. It will be available in the next version of CDK 3.0 

```
$ minishift addons install --defaults
$ minishift addons enable cluster-admin
$ minishift start
```

**Comment by LalatenduMohanty** [comment]
@burrsutter or you can just run this after `minishift start` as @jorgemoralespou mentioned.
```
$ oc adm policy  --as system:admin add-cluster-role-to-user cluster-admin developer
```
 The addons is the suggested solution from Minishift, please give it a try and let us know if thsi works for you.

hferentschik:
@burrsutter, you will never be able to login as _system:admin_. _system:admin_ is a special account which only allows login via certificates (which as it happens to be Minishift copies into your _~/.kube/config_). That's the reasons why you can use the _system:admin_ account from the console using _oc_. If you grep _~/.kube/config_ for system.admin and the IP of the current Minishfit VM, you most likely find the entry which makes the login possible and you will also see that it is certificate based. 

So the question is what you really want. You want to log into the console and see the _default_ namespace? Is this what you are after? Then just give the 'developer' user the cluster-admin role as described above. Either via running the _oc adm_ command directly or if you want to have this as a permanent thing, via enabling the _cluster_admin_ addon. Once enabled, the developer user will get the _cluster_admin_ role each time you create a Minishift instance.

jorgemoralespou:
@hferentschik @LalatenduMohanty I would argue that making developer a cluster-admin is a good practice, hence I supported @jstrachan request to have a dedicated user being cluster admin (which we valuated as a possible addon).

[...]

PR #741, opened 2017-04-12
Issue #696 Added admin user with cluster-admin role

praveenkumar:
This patch will fix cluster-admin addon to provide cluster-admin access to admin user.


addons/cluster-admin/cluster-admin.addon
```diff
@@ -1,4 +1,5 @@
 # Name: cluster-admin
 # Description: Adds the cluster-admin role to the developer user.
 
-oc adm policy add-role-to-user cluster-admin developer
+oc create user admin --full-name=admin
```
hferentschik: that was easy :-) Good stuff.

addons/cluster-admin/cluster-admin.addon
```diff
@@ -1,4 +1,5 @@
 # Name: cluster-admin
 # Description: Adds the cluster-admin role to the developer user.
```
hferentschik: The description needs to be changed

hferentschik:
So the idea is to modify our existing default cluster-admin plugin? No new/additional add-on? 

We might want to at least change the name in the addon file to create-cluster-admin. Maybe even rename file and directory name. Thoughts?

@budhrg @LalatenduMohanty ^^

addons/cluster-admin/cluster-admin.addon
```diff
@@ -1,4 +1,5 @@
 # Name: cluster-admin
-# Description: Adds the cluster-admin role to the developer user.
+# Description: Create admin user and add cluster-admin role to it.
 
-oc adm policy add-role-to-user cluster-admin developer
+oc create user admin --full-name=admin
```
LalatenduMohanty: I think we should go for an user with name "cluster-admin"

addons/cluster-admin/cluster-admin.addon
```diff
@@ -1,4 +1,5 @@
 # Name: cluster-admin
```
LalatenduMohanty: Also as discussed with @hferentschik @budhrg  @praveenkumar  The addon name should be "create-cluster-addon"

hferentschik: I assume you mean "create-cluster-admin" ;-)

praveenkumar:
@hferentschik @LalatenduMohanty finally I am endup doing change from directory level as well because when default addons install it go to that directory which named as `addon` name before install https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/addon/addon_install.go#L84

addons/create-cluster-admin/create-cluster-admin.addon
```diff
@@ -0,0 +1,5 @@
+# Name: create-cluster-admin
+# Description: Create admin user and add cluster-admin role to it.
+
+oc create user admin --full-name=admin
```
LalatenduMohanty: @praveenkumar I suggest the user name should be "cluster-admin". I think thats what others agreed also @hferentschik @budhrg

praveenkumar: @LalatenduMohanty No we agreed about only the addon name not about `user` and why you think user should be `cluster-admin` ?

hferentschik: I prefer 'admin', mainly it is easier to type and it aligns with what we had before.

jorgemoralespou:
I think this should be add-cluster-role-to-user since what you're assigning is a cluster wide role.

In case of doubt, better to ask @liggitt.

Jordan,
To create an admin user as cluster admin what command should be?

````
oc adm policy add-role-to-user cluster-admin admin
````

or 

````
oc adm policy add-cluster-role-to-user cluster-admin
````

**Comment by liggitt** [comment]
add-cluster-role-to-user

**Comment by LalatenduMohanty** [comment]
@praveenkumar @hferentschik I suggested `cluster-admin` because It might confuse user with the `system:admin` which comes from `cluster up` output.  We need to educate users that they can not login as system:admin but we have created an alternative for them.  WDYT ?

@jorgemoralespou suggestion?

```
 To login as administrator:
       oc login -u system:admin
```

File: addons/create-cluster-admin/create-cluster-admin.addon [hunk 6]
```diff
@@ -0,0 +1,5 @@
...
+oc adm policy add-cluster-role-to-user cluster-admin admin
```
hferentschik: btw, there is a new line missing here

praveenkumar: @hferentschik will add that.

praveenkumar:
> I suggested cluster-admin but not just admin because otherwise it might confuse user with the system:admin which comes from cluster up output. 

I am not sure how creating cluster-admin as user going to help that. Will do final update on this PR once we all agree what actually we need. (I don't have any strong feeling about username also but I just not convince that changing that might help).

hferentschik:
Ok, so there are several aspects here and I am not sure whether we are all on the same level. 
So what we want to do is to create a new user _admin_ (or _cluster-admin_ as @LalatenduMohanty prefers) who has the cluster admin role. The main use case is to be able to login as this user via the web console and getting access to the otherwise hidden namespaces of OpenShift.

Looking at the actual commands, afaict we are doing the right thing. What we are discussing right now is the metadata of the add-on:

    # Name: create-cluster-admin
    # Description: Create admin user and add cluster-admin role to it.

I think the description is ok, right? Maybe we can be a bit more precise? "Creating a user 'admin' and assigning the cluster-admin role to him"

 Leaves the 'Name', which is what a user sees when he does things like _minishift addons [list|enbable|disable]_. At the moment it is _create-cluster-admin_. I can see how this is ambiguous, on the other hand, things like _create-admin-user-with-cluster-role_ might be a bit unwieldy to mange. As said, this are the names a user needs to work with from the CLI. So  _create-admin-user-with-cluster-role_ is probably a more accurate name, but I am not sure whether it is a good choice. Alternatively, just _admin-user_. It does not explain in the title what it does, but it is less ambiguous and it is the description which adds additional context.

So right now my preferred options are:
* _create-admin-user-with-cluster-role_
* _admin-user_

Can we align on one of these names?

LalatenduMohanty:
> So right now my preferred options are:
> 
>     create-admin-user-with-cluster-role
>     admin-user

As discussed  we will keep the addon name as `admin-user` and the user name as `admin`.

PR #743, opened 2017-04-12
Issue #740: Extracted and edited Mount host folder section…

### LINKED ISSUE #740: Extract Mount Host folder section from Managing Minishift
As per #734, we intend to create a separate Page for Mount host folders. Creating this issue to only handle extraction of this content to prevent blocking action on #734 as well as to ensure that fresh content required for this section as per #655, goes directly into this extracted section

Preeticp:
…from Managing Minishift doc.
This PR fixes #740. 
Extracted Mount host folders section from Managing Minishift doc.
Did a light proof read of the content.

### DIFF
--- a/docs/source/using/managing-minishift.adoc
+++ b/docs/source/using/managing-minishift.adoc
@@ -127,6 +127,7 @@ The easiest way to change a persistent configuration option is with
 the link:../command-ref/minishift_config_set{outfilesuffix}[`config set`] sub-command. For example:
 
 [listing.console]
+
 ----
 # Set default memory 4096 MB
 $ minishift config set memory 4096
@@ -136,6 +137,7 @@ To view all persistent configuration values, you can use the
 link:../command-ref/minishift_config_view{outfilesuffix}[`view`] sub-command:
 
 [listing.console]
+
 ----
 $ minishift config view
 - memory: 4096
@@ -145,6 +147,7 @@ Alternatively, you can display a single value with the
 link:../command-ref/minishift_config_get{outfilesuffix}[`get`] sub-command:
 
 [listing.console]
+
 ----
 $ minishift config get memory
 4096
@@ -157,11 +160,12 @@ To remove a persistent configuration option, you can use the
 link:../command-ref/minishift_config_unset{outfilesuffix}[`unset`] sub-command. For example:
 
 [listing.console]
+
 ----
 $ minishift config unset memory
 ----
 
-[[ersistent-volumes]]
+[[persistent-volumes]]
 == Persistent Volumes
 
 As part of the <<interacting-with-openshift#openshift-client-binary,cluster up>> provisioning
@@ -184,6 +188,7 @@ flags during `minishift start`.
 For example:
 
 [listing.console]
+
 ----
 $ minishift start --http-proxy http://YOURPROXY:PORT --https-proxy https://YOURPROXY:PORT
 ----
@@ -192,9 +197,10 @@ In an authenticated proxy environment, the `proxy_user` and
 `proxy_password` must be a part of proxy URI.
 
 [listing.console]
+
 ----
  $ minishift start --http-proxy http://<proxy_username>:<proxy_password>@YOURPROXY:PORT \
-                   --https-proxy https://<proxy_username>:<proxy_password>YOURPROXY:PORT
+                   --https-proxy https://<proxy_username>:<proxy_password>@YOURPROXY:PORT
 ----
 
 You can also use the `--no-proxy` flag to specify a comma-separated list of hosts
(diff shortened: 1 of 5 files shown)


docs/source/using/mounting-host-folders.adoc
```diff
@@ -0,0 +1,122 @@
...
+|Virtualbox |Windows |C://Users |/c/Users
+|VMWare Fusion |OSX |/Users |/Users
+|Xhyve |OSX |/Users |/Users
+|========================================
+
+NOTE: Host folder sharing is not implemented in the KVM and Hyper-V drivers. You can
+link:#mounting-custom-shared-folders[mount a CIFS-based shared folder] inside the VM instead.
+
+[[mounting-custom-shared-folders]]
+== Mounting custom shared folders
```
thatdocslady: Looking at the section as a standalone file, I don't think we need this title. The section below it can be a part of the overview, and actually I would probably raise the CIFS and SSHFS headings up one level so that they can appear in the navigation.

docs/source/using/mounting-host-folders.adoc
```diff
@@ -0,0 +1,122 @@
...
+.  Find the local IP address from the same network segment as your
+Minishift instance.
++
+----
+$ Get-NetIPAddress | Format-Table`
+----
+
+.  Create a mountpoint and mount the shared folder.
++
+[source,sh]
```
thatdocslady: we don't need this [source, sh], I actually went and removed as many of those as I could find, it's a conversion leftover, sorry if I missed this!

docs/source/using/mounting-host-folders.adoc
```diff
@@ -0,0 +1,122 @@
...
+----
+$ minishift ssh "sudo mkdir -p /Users"
+$ minishift ssh "sudo mount -t cifs //[machine-ip]/Users /Users -o username=[username],password=[password],domain=$env:computername
+----
++
+If no error appears, the mount succeeded.
+
+.  Verify the share mount.
++
+[source,sh]
```
thatdocslady: same comment, please remove (also other instances, I won't repeat the comment)

docs/source/using/mounting-host-folders.adoc
Preeticp: @hferentschik, as per this comment: https://github.com/minishift/minishift/pull/717#discussion_r110606862 we were supposed to be adding these. Can we discuss and decide on why we need these tags and how they help us?

hferentschik: > we don't need this [source, sh], I actually went and removed as many of those as I could find, it's a conversion leftover, sorry if I missed this!

-1, imo the opposite is true. We should add them were they are missing. This is the information the syntax highlighter needs to decide on the form of syntax highlighting. Why would you want to remove this? 

> Can we discuss and decide on why we need these tags and how they help us?

I am surprised that we need to discuss this, but sure:
* http://www.methods.co.nz/asciidoc/source-highlight-filter.html
* http://asciidoctor.org/docs/user-manual/#source-code-blocks

We are using embedded (source) code here. On the HTML side source rendering occurs with one of a whole list of source code highlighters (at least it can be) - CoreRay, prettify, etc. This highlighters are quite advanced and can depending on the type of source - sh, java, php, etc - apply syntax highlighting as expected per language type. However, to do this you need to provide some meta information. In this case [source,sh]. Or in case we are having a JSON example we should use [source,json]. 

Even if the highlighter does not get used, adding this information makes sense, since it adds useful meta information. 

So -1 for removing this. On the contrary, we should try adding them. This is actually what I am doing. 

On that note, longer free standing code examples and table should always also get a anchor, so that they become reference-able. Personally I like to use prefixes for this 'example-<bla>', 'table-<bla>'. If you reference these type of things from somewhere in the document it becomes clear what you are referencing even w/o looking at the reference target, e.g. <<table-addon-command-list>> makes clear that I am referencing a table even without looking at the actual target. Effectively you are creating a namespace over your anchors.

docs/source/using/mounting-host-folders.adoc
Preeticp: +1

docs/source/using/mounting-host-folders.adoc
Preeticp: Thanks for the explanation @hferentschik will retain and add these tags where missing.

hferentschik: on second thought and after doing some research and trying to align with the existing docs.openshift.org, I suggest we use:

    [listing.console] 
Or 
   
    [listing,role="console"]

[source,sh] is a bit controversial. Really this is not shell script as such, but console command [source,console] could work, but as it turns out, docs.openshift.org does not differentiate in this case and just adding [source] will apply some styles which differ from the default listing format used. 

[listing.console] will give us the best of both worlds for now I think. As part of docs.openshift.org the output between the Minishift docs and the rest of the docs won't differ, but in our asciidoc we properly mark the type of listing/source and in HTML there will be a 'console' class added, which might come in handy at some stage.

So, I'd say we use [listing.console] for now.

hferentschik: Btw, I'd also like to discuss switching to 'sentence-per-line'. WDYT?

LalatenduMohanty: @Preeticp @hferentschik @thatdocslady Regarding code highlighting e.g. `[source, bash], [source, yaml]` I see them used in openshift-docs. As per AsciiDoctor documentation source highlighters supported.  However it is not mentioned in the guideline  https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#code-blocks. 

Also about `[listing.console] ` , it is not present in in the existing openshift-docs or in the doc_guidelines. So I am not sure if this plays well with AsciiDoctor and the tool chain to convert downstream docs. I am expecting some tool chain is used to convert these docs to downstream docs. 

Can we please discuss and come to some guideline for Minishift docs so that  in future it is relatively clear when someone sends a documentation PR.

Preeticp:
Regarding code highlighting e.g. [source, bash], [source, yaml] I see them used in openshift-docs.
> +1

Also about [listing.console] , it is not present in the existing openshift-docs or in the doc_guidelines. 
> True. Here are a few links from Asciidoctor user manual wrt [listing]: 
http://asciidoctor.org/docs/user-manual/#listing-blocks and 
http://asciidoctor.org/docs/user-manual/#built-in-blocks-summary

Can we please discuss and come to some guideline for Minishift docs so that in future it is relatively clear when someone sends a documentation PR.
>+1, it would help us understand better as well as get everyone on the same page.

docs/source/using/mounting-host-folders.adoc
hferentschik: > So I am not sure if this plays well with AsciiDoctor and the tool chain to convert downstream docs. 

This is AsciiDoctor syntax why would it not play nice with AsciiDoctor? This just makes the type 'listing' explicit and adds a 'console' role to it. If you don't add 'listing' it is assumed.

thatdocslady: The issue isn't with AsciiDoctor, it's with the toolchains that we use downstream (and also upstream via AsciiBinder), and we simply don't know how those tools will respond to tags that are not defined in the guidelines. Granted, the guidelines might not be inclusive of all the supported tags, but if we want to use tags that are not listed we should consult the OpenShift writers and verify that these tags won't break anything throughout all the toolchains that we use. 

@LalatenduMohanty there is no conversion planned for downstream docs, this is the whole point of the convergence initiative, to be able to reuse content both in upstream and downstream, which is one of the reasons that following conventions and communicating whenever we want to establish/extend/modify the conventions is critical.

Bottom line - I personally don't see the value in adding extra tags to code blocks, but if we want to implement something like this we should discuss it first internally, then with openshift and possibly also the content services tooling team, and document the conventions so that we don't have to remember who decided what when and why.

Preeticp: Folks I have opened a new issue #768  for discussing Usage of tags. I have copied all relevant comments in the issue. The reasons, we have a release on us and this PR is blocking work on other PRs.
So currently I will not touch existing Tags, we will undertake that for all docs once we decide on that. I will change the name as suggested by @hferentschik and push my changes. Hope this is alright by all.

thatdocslady: +1, I agree that the tags can be fixed/changed/decided independently of this PR and it's better to push what we agreed on and then take a pass and fix globally later.

LalatenduMohanty: I would suggest to fix this PR with https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#code-blocks and then discuss new ideas in #768.

docs/source/using/managing-minishift.adoc
```diff
@@ -196,116 +196,6 @@ Use the `openshift-version` option to request a specific version of OpenShift. Y
 all Minishift-compatible OpenShift versions with
 the link:../command-ref/minishift_openshift_list-versions{outfilesuffix}[`minishift openshift list-versions`] command.
```
Preeticp: added @

budhrg: :+1:

Preeticp:
@hferentschik, I have changed the name to Host Folder, can you please merge this PR so that other issues are not blocked because of this.
 As stated in: https://github.com/minishift/minishift/pull/743#discussion_r111952216, I have created issue #768 to discuss the tags we want to use for docs.

hferentschik:
Rebased and merged

PR #744, opened 2017-04-12
Issue #742 Downloading message for 'oc' binary

### LINKED ISSUE #742: 'minishift start' doesn't tell what the 2nd progress bar means
Hi,

It looks like we are missing something here : 
```
$ minishift start    
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso'
 40.00 MB / 40.00 MB [============================================================================================================================================================] 100.00% 0s
*************************
 19.96 MB / 19.96 MB [============================================================================================================================================================] 100.00% 0s
*************************
...
```

The progress bar looks  like downloading `oc` binary here. But there is no message like we have for `Downloading ISO ...`

budhrg:
Fix #742 

Now, the message is 

```
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso'
 40.00 MB / 40.00 MB [============================================================================================================================================================] 100.00% 0s
Downloading OpenShift client binary oc with version 'v1.5.0-rc.0'
 19.96 MB / 19.96 MB [============================================================================================================================================================] 100.00% 0s
```

The URL here being used to download oc seems to be https://github-cloud.s3.amazonaws.com/releases/22442668/bb24051a . 

Not sure should I use this or message above is fine. I went with the later one for better readability.

### DIFF
--- a/pkg/util/github/github.go
+++ b/pkg/util/github/github.go
@@ -125,7 +125,7 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
 		return errors.Wrap(err, fmt.Sprintf("Cannot download OpenShift release asset %d", assetID))
 	}
 	if len(url) > 0 {
-		glog.V(2).Infof("Downloading %s %s\n", binaryType.String(), *release.TagName)
+		fmt.Println(fmt.Sprintf("Downloading OpenShift binary '%s' in version '%s'", binaryType.String(), *release.TagName))
 		httpResp, err := http.Get(url)
 		if err != nil {
 			return errors.Wrap(err, "Cannot download OpenShift release asset.")


pkg/util/github/github.go
```diff
@@ -125,7 +125,7 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
 		return errors.Wrap(err, fmt.Sprintf("Cannot download OpenShift release asset %d", assetID))
 	}
 	if len(url) > 0 {
-		glog.V(2).Infof("Downloading %s %s\n", binaryType.String(), *release.TagName)
+		fmt.Println(fmt.Sprintf("Downloading OpenShift client binary %s with version '%s'", binaryType.String(), *release.TagName))
```
hferentschik: The message becomes wrong here. _DownloadOpenShiftReleaseBinary_ can be used to download _openshift_ or _oc_. By now we only use it for _oc_ apparently, but still the method does allow to download _openshift_, in which case the message becomes wrong. So, if we want to change this, it would be more along the lines:

"Downloading binary %s  in version %s\n", binaryType.String(), *release.TagName)"
or
"Downloading OpenShift binary  %s  in version %s\n", binaryType.String(), *release.TagName)"

Also, I think it is "you are downloading something _in_ a specific version" not _with_.

LalatenduMohanty: @hferentschik So there are two types of binary i.e. OpenShift server or client. I do not think we will ever going to download the OpenShift server binary in this function unless we are moving away from "cluster up". Which I think not going to happen in foreseeable future. So saying `OpenShift client binary` should be fine IMO.

budhrg: I am tend to agree with @hferentschik here to make use of general approach "Downloading OpenShift binary ...." to align with code.

Now result is : 
```
$ minishift start 
Downloading OpenShift binary 'oc' in version 'v1.5.0-rc.0'
 1.94 MB / 19.96 MB [==============>-------------------------------------------------------------------------------------------------------------------------------------------]   9.73% 1m10s
```

LalatenduMohanty: @budhrg Sure. Thats better I think. I had missed the ` *release.TagName` in @hferentschik suggestion.

pkg/util/github/github.go
```diff
@@ -125,7 +125,7 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
 		return errors.Wrap(err, fmt.Sprintf("Cannot download OpenShift release asset %d", assetID))
 	}
 	if len(url) > 0 {
-		glog.V(2).Infof("Downloading %s %s\n", binaryType.String(), *release.TagName)
+		fmt.Println(fmt.Sprintf("Downloading OpenShift binary '%s' in version '%s'", binaryType.String(), *release.TagName))
```
LalatenduMohanty: Nitpicking. Do we need the the `in` in the sentence i.e. `Downloading OpenShift binary 'oc' in version 'v1.5.0-rc.0' ` VS `Downloading OpenShift binary 'oc' version 'v1.5.0-rc.0'`

hferentschik: I don't care

budhrg:
Merged!

PR #746, opened 2017-04-12
[WIP] Issue #590 Allow selection of virtual switch for use with Hyper-V

### LINKED ISSUE #590: Allow selection of virtual switch for use with Hyper-V

budhrg:
Fix #590 

@hferentschik Sending PR for initial review.

Found that the environment options are not there for KVM and Xhyve.

```
PS C:\gowork\src\github.com\minishift\minishift> $env:HYPERV_VIRTUAL_SWITCH="minishift_switch"
PS C:\gowork\src\github.com\minishift\minishift>
PS C:\gowork\src\github.com\minishift\minishift>
PS C:\gowork\src\github.com\minishift\minishift> $env:HYPERV_VIRTUAL_SWITCH
minishift_switch
PS C:\gowork\src\github.com\minishift\minishift>
PS C:\gowork\src\github.com\minishift\minishift>
PS C:\gowork\src\github.com\minishift\minishift> minishift.exe start -v 5 --show-libmachine-logs
Starting local OpenShift cluster using 'hyperv' hypervisor...
Found binary path at C:\gowork\bin\minishift.exe
Launching plugin server for driver hyperv
Plugin server listening at address 127.0.0.1:50000
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(minishift) Calling .GetMachineName
(minishift) Calling .DriverName
Creating CA: C:\Users\budhram\.minishift\certs\ca.pem
Creating client certificate: C:\Users\budhram\.minishift\certs\cert.pem
Running pre-create checks...
(minishift) Calling .PreCreateCheck
(minishift) DBG | [executing ==>] : C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive @(Get-Command Get-VM).ModuleName
(minishift) DBG | [stdout =====>] : Hyper-V
(minishift) DBG |
(minishift) DBG | [stderr =====>] :
(minishift) DBG | [executing ==>] : C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive @([Security.Principal.WindowsPrincipal][Security.Principal.
WindowsIdentity]::GetCurrent()).IsInRole("Hyper-V Administrators")
(minishift) DBG | [stdout =====>] : True
(minishift) DBG |
(minishift) DBG | [stderr =====>] :
(minishift) DBG | [executing ==>] : C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive (Get-VMSwitch).Name
(minishift) DBG | [stdout =====>] : DockerNAT
(minishift) DBG | minishift_switch
(minishift) DBG |
(minishift) DBG | [stderr =====>] :
(minishift) Calling .GetConfigRaw
Creating machine...
(minishift) Calling .Create
(minishift) Downloading C:\Users\budhram\.minishift\cache\boot2docker.iso from file://C:/Users/budhram/.minishift/cache/iso/minishift-b2d.iso...
(minishift) Creating SSH key...
(minishift) Creating VM...
(minishift) DBG | [executing ==>] : C:\WINDOWS\System32\WindowsPowerShell\v1.0\powershell.exe -NoProfile -NonInteractive (Get-VMSwitch).Name
(minishift) DBG | [stdout =====>] : DockerNAT
(minishift) DBG | minishift_switch
*****************************************************
(minishift) Using switch "minishift_switch"
*****************************************************
(minishift) DBG |
.....
``` 

Check now, driver is picking the right switch above.

Also working on tests.

### DIFF
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -197,13 +197,18 @@ func engineOptions(config MachineConfig) *engine.Options {
 	return &o
 }
 
-func createVirtualboxHost(config MachineConfig) drivers.Driver {
+func createVirtualboxHost(config MachineConfig) (drivers.Driver, error) {
 	d := virtualbox.NewDriver(constants.MachineName, constants.Minipath)
 	d.Boot2DockerURL = config.GetISOFileURI()
 	d.Memory = config.Memory
 	d.CPU = config.CPUs
 	d.DiskSize = int(config.DiskSize)
-	return d
+
+	if err := setDriverOptionsFromEnvironment(d); err != nil {
+		return nil, err
+	}
+
+	return d, nil
 }
 
 func (m *MachineConfig) CacheMinikubeISOFromURL() error {
@@ -283,6 +288,7 @@ func (m *MachineConfig) IsMinikubeISOCached() bool {
 
 func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 	var driver interface{}
+	var err error
 
 	if config.ShouldCacheMinikubeISO() {
 		if err := config.CacheMinikubeISOFromURL(); err != nil {
@@ -292,17 +298,21 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 
 	switch config.VMDriver {
 	case "virtualbox":
-		driver = createVirtualboxHost(config)
+		driver, err = createVirtualboxHost(config)
 	case "vmwarefusion":
-		driver = createVMwareFusionHost(config)
+		driver, err = createVMwareFusionHost(config)
 	case "kvm":
-		driver = createKVMHost(config)
+		driver, err = createKVMHost(config)
 	case "xhyve":
-		driver = createXhyveHost(config)
+		driver, err = createXhyveHost(config)
 	case "hyperv":
-		driver = createHypervHost(config)
+		driver, err = createHypervHost(config)
 	default:
 		glog.Exitf("Unsupported driver: %s\n", config.VMDriver)
+	}
+
+	if err != nil {
+		return nil, err
 	}
 
 	data, err := json.Marshal(driver)
--- a/pkg/minikube/cluster/cluster_darwin.go
+++ b/pkg/minikube/cluster/cluster_darwin.go
@@ -22,7 +22,7 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/constants"
 )
 
-func createVMwareFusionHost(config MachineConfig) drivers.Driver {
+func createVMwareFusionHost(config MachineConfig) (drivers.Driver, error) {
 	d := vmwarefusion.NewDriver(constants.MachineName, constants.Minipath).(*vmwarefusion.Driver)
 	d.Boot2DockerURL = config.GetISOFileURI()
 	d.Memory = config.Memory
@@ -31,7 +31,12 @@ func createVMwareFusionHost(config MachineConfig) drivers.Driver {
 	// TODO(philips): push these defaults upstream to fixup this driver
 	d.SSHPort = 22
 	d.ISO = d.ResolveStorePath("boot2docker.iso")
-	return d
+
+	if err := setDriverOptionsFromEnvironment(d); err != nil {
+		return nil, err
+	}
+
+	return d, nil
 }
 
 type xhyveDriver struct {
@@ -51,8 +56,8 @@ type xhyveDriver struct {
 	Virtio9pFolder string
 }
 
-func createXhyveHost(config MachineConfig) *xhyveDriver {
-	return &xhyveDriver{
+func createXhyveHost(config MachineConfig) (*xhyveDriver, error) {
+	d := &xhyveDriver{
 		BaseDriver: &drivers.BaseDriver{
 			MachineName: constants.MachineName,
 			StorePath:   constants.Minipath,
@@ -65,4 +70,6 @@ func createXhyveHost(config MachineConfig) *xhyveDriver {
 		Virtio9pFolder: "/Users",
 		UUID:           "F4BB3F79-AB4E-4708-95CA-E32FBFCDEFD3",
 	}
+
+	return d, nil
 }
--- a/pkg/minikube/cluster/cluster_linux.go
+++ b/pkg/minikube/cluster/cluster_linux.go
@@ -39,8 +39,8 @@ type kvmDriver struct {
 	IOMode         string
 }
 
-func createKVMHost(config MachineConfig) *kvmDriver {
-	return &kvmDriver{
+func createKVMHost(config MachineConfig) (*kvmDriver, error) {
+	d := &kvmDriver{
 		BaseDriver: &drivers.BaseDriver{
 			MachineName: constants.MachineName,
 			StorePath:   constants.Minipath,
@@ -56,4 +56,6 @@ func createKVMHost(config MachineConfig) *kvmDriver {
 		CacheMode:      "default",
 		IOMode:         "threads",
 	}
+
+	return d, nil
 }
--- a/pkg/minikube/cluster/cluster_non_darwin_panic.go
+++ b/pkg/minikube/cluster/cluster_non_darwin_panic.go
@@ -20,10 +20,10 @@ package cluster
 
 import "github.com/docker/machine/libmachine/drivers"
 
-func createVMwareFusionHost(config MachineConfig) drivers.Driver {
+func createVMwareFusionHost(config MachineConfig) (drivers.Driver, error) {
 	panic("vmwarefusion is not supported")
 }
 
-func createXhyveHost(config MachineConfig) drivers.Driver {
+func createXhyveHost(config MachineConfig) (drivers.Driver, error) {
 	panic("xhyve is not supported")
 }
--- a/pkg/minikube/cluster/cluster_non_linux_panic.go
+++ b/pkg/minikube/cluster/cluster_non_linux_panic.go
@@ -20,6 +20,6 @@ package cluster
 
 import "github.com/docker/machine/libmachine/drivers"
 
-func createKVMHost(config MachineConfig) drivers.Driver {
+func createKVMHost(config MachineConfig) (drivers.Driver, error) {
 	panic("kvm is not supported")
 }
--- a/pkg/minikube/cluster/cluster_non_windows_panic.go
+++ b/pkg/minikube/cluster/cluster_non_windows_panic.go
@@ -20,6 +20,6 @@ package cluster
 
 import "github.com/docker/machine/libmachine/drivers"
 
-func createHypervHost(config MachineConfig) drivers.Driver {
+func createHypervHost(config MachineConfig) (drivers.Driver, error) {
 	panic("hyperv is not supported")
 }
--- a/pkg/minikube/cluster/cluster_windows.go
+++ b/pkg/minikube/cluster/cluster_windows.go
@@ -22,12 +22,17 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/constants"
 )
 
-func createHypervHost(config MachineConfig) drivers.Driver {
+func createHypervHost(config MachineConfig) (drivers.Driver, error) {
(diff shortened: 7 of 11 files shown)


hferentschik:
> Found that the environment options are not there for KVM and Xhyve.  

I am confused, so what's:

* https://github.com/zchee/docker-machine-driver-xhyve/blob/master/xhyve/xhyve.go
* https://github.com/dhiltgen/docker-machine-kvm/blob/master/kvm.go

KVM seems to be indeed short of env options, but it is how it is. xhyve looks complete though.

pkg/minishift/util/set_env_driver_options.go
```diff
@@ -0,0 +1,30 @@
...
+	supportedFlags := d.GetCreateFlags()
+
+	checkFlags := &drivers.CheckDriverOptions{
+		FlagsValues: map[string]interface{}{},
+		CreateFlags: supportedFlags,
+	}
+
+	for _, flag := range supportedFlags {
+		flagName := flag.String()
+		flagEnv := strings.Replace(strings.ToUpper(flagName), "-", "_", -1)
```
hferentschik: -1, why do you  do any conversion here. This might or might not work. There is no enforced standard on how the env variable looks like. _GetCreateFlags_ returns a slice of _[]mcnflag.Flag_. Each instance of mcnflag.Flag has a public member _EnvVar_. So you don't need to convert anything. You literally iterate over the slice, check whether the flag has a value for _EnvVar_ (it might not have, see KVM driver) and use this flag name to call os.Getenv.

Otherwise this is the right approach.

pkg/minikube/cluster/cluster.go
```diff
@@ -199,6 +200,10 @@ func engineOptions(config MachineConfig) *engine.Options {
 
 func createVirtualboxHost(config MachineConfig) drivers.Driver {
 	d := virtualbox.NewDriver(constants.MachineName, constants.Minipath)
+	if err := minishiftUtil.SetEnvDriverOptions(d); err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error in setting config from flags: %s", err.Error()))
```
hferentschik: Not sure whether it is a good idea to use atexit on this level. I would think atexit is something used from the cmd part of Minishift. Here on the pkg level we should stick to the Go way of doing things and return an error. By calling Exit you terminate the program here and now. What if some cleanup is needed higher up. So we need to change the signature here and return an error.

pkg/minikube/cluster/cluster_darwin.go
```diff
@@ -17,13 +17,20 @@ limitations under the License.
...
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/util"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
 )
 
 func createVMwareFusionHost(config MachineConfig) drivers.Driver {
 	d := vmwarefusion.NewDriver(constants.MachineName, constants.Minipath).(*vmwarefusion.Driver)
+	if err := util.SetEnvDriverOptions(d); err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error in setting config from flags: %s", err.Error()))
```
hferentschik: same as above

hferentschik:
> Check now, driver is picking the right switch above.

👍

hferentschik:
Should this not be a fix for #305 though?

hferentschik:
@budhrg, did you see my comment here - https://github.com/minishift/minishift/issues/305#issuecomment-293522953. Have you thought about whether we should use a prefix? There are pro and cons for this.  I am atm somewhere 50/50.

@LalatenduMohanty, @praveenkumar, @gbraad  - any thoughts on this. Should we accept environment provided settings as per driver documentation or should we enforce that the user uses these variables with a _MINISHIFT_ prefix? 

The pros for prefix:
* It is aligned with how Viper works https://docs.openshift.org/latest/minishift/using/managing-minishift.html#environment-variables
* It creates a namespace. In case a user is also using docker-machine, the env variables if for example added to the profile can be separated 

The cons:
* Less intuitive. In the docs we most likely are referring to the driver docs. There the plain variables will be listed. If we require a MINISHIFT prefix, the user needs one more piece if information to get it right

Alternatively we could look for both. W/ and w/o prefix.

pkg/minishift/util/set_env_driver_options.go
```diff
@@ -0,0 +1,30 @@
+package util
+
+import (
+	"github.com/docker/machine/libmachine/drivers"
+	"os"
+	"strings"
+)
+
+func SetEnvDriverOptions(d drivers.Driver) error {
```
hferentschik: _SetDriverOptionsFromEnvironment_, _ConfigureDriverFromEnvironment_?

hferentschik:
> Also working on tests.

+1, From a testing perspective, you would most likely be better off, not passing the Driver into the function, but just the  []mcnflag.Flag (making the call to CheckDriverOptions outside the helper function). This will most likely make your testing easier. You should be able to easily tests then that you algorithm picks up environment variables as specified in a flag.

budhrg:
> https://github.com/zchee/docker-machine-driver-xhyve/blob/master/xhyve/xhyve.go
https://github.com/dhiltgen/docker-machine-kvm/blob/master/kvm.go
KVM seems to be indeed short of env options, but it is how it is. xhyve looks complete though.

Missed to look into those separate drivers. Thanks.

budhrg:
> Should we accept environment provided settings as per driver documentation or should we enforce that the user uses these variables with a MINISHIFT prefix?

So, this PR https://github.com/minishift/minishift/pull/738/files was for supporting `HYPERV_VIRTUAL_SWITCH`  with _MINISHIFT_ prefix. 

If we are going to support for other options then I believe we need to update those files right? 

Also, as part fo this comment https://github.com/minishift/minishift/issues/305#issuecomment-293522953 you mentioned about generic option `--driver-opts`.

At the moment, I see lots of ideas popping out. In all the approaches, there are a duplication of driver options. We need to decide
- which strategy to choose
- if multiple then which priority
 
Let's discuss @hferentschik @LalatenduMohanty @praveenkumar

gbraad:
>
> there are a duplication of driver options.
>

duplication leads to conflicting behaviour (unless prioirty is described)
and worst of all, confusion and possible inconsistency. it is better to
have a single approach. the PR #738 solves a problem as a short-term
solution

hferentschik:
> So, this PR https://github.com/minishift/minishift/pull/738/files was for supporting HYPERV_VIRTUAL_SWITCH with MINISHIFT prefix.

Sure, but what's your point? The code in this pull request is a generic solution for issue #305, right? It does not only solve #590, right? So all I'm saying is that the issue used for this work is #305. 

> If we are going to support for other options then I believe we need to update those files right?

Which files? I don't understand what you mean.

> Also, as part fo this comment #305 (comment) you mentioned about generic option --driver-opts.

Right, this is something we can do as well. I'd prefer that over adding explicit driver specific options. This can be paired with allowing setting of these options in _config.json_. However, as I mentioned, I think we should do this in a follow up issue. How exactly a CLI solution looks like might need some more discussion.

> We need to decide
> which strategy to choose

Not sure what you mean

> if multiple then which priority

Correct, as pointed out in the issue, we need to decide on the intended behaviour.

pkg/minishift/util/set_driver_options_from_env.go
```diff
@@ -0,0 +1,46 @@
...
+
+func SetDriverOptionsFromEnvironment(d drivers.Driver) error {
+	supportedFlags := d.GetCreateFlags()
+
+	checkFlags := &drivers.CheckDriverOptions{
+		FlagsValues: map[string]interface{}{},
+		CreateFlags: supportedFlags,
+	}
+
+	for _, flag := range supportedFlags {
```
budhrg: @hferentschik  Using this approach instead of manually creating the environment variable. Please check.

pkg/minishift/util/set_driver_options_from_env.go
```diff
@@ -0,0 +1,46 @@
...
+		switch reflect.TypeOf(flag).String() {
```
hferentschik: why do you switch on the string representation of the type. That's fragile, you really should switch on the type - https://golang.org/doc/effective_go.html#type_switch

But even more importantly, why the switch at all? AFAICT, the current code does not require any type switch atm.

LalatenduMohanty:
> any thoughts on this. Should we accept environment provided settings as per driver documentation or should we enforce that the user uses these variables with a MINISHIFT prefix?

+1 for prefix. It would allow users to use multiple of tools based on libmachine isn't? e.g. Minikube and Minishift together with specific env variables for Minishift.

pkg/minishift/util/set_driver_options_from_env.go
budhrg: @hferentschik , the underline type for `flag` can be `mcnflag.StringFlag`, `mcnflag.IntFlag` etc. or multiple.

as `flag` is a interface type here.

hferentschik: yeah, but you are switching on a string, the string representation of of the type, instead of the type, aka StringFlag itself.

budhrg: @hferentschik  Fixed

pkg/minikube/cluster/cluster.go
budhrg: @hferentschik , Fixed

budhrg:
Hi @hferentschik , 

I did try to use reflection here to set the fields of driver for kvm, check below
```
func createKVMHost(config MachineConfig) (drivers.Driver, error) {
  api := libmachine.NewClient("", "")
  host, _ := api.NewHost("kvm", []byte("{}"))

  driver := host.Driver

  fmt.Println("Type of driver: ", reflect.TypeOf(driver))

  // Using  http://stackoverflow.com/a/6396678/1120530
  s := reflect.Indirect(reflect.ValueOf(driver))

  fmt.Println("Kind of s : ", s.Kind())
  
  if s.Kind() == reflect.Struct {
    f := s.FieldByName("Memory")
    
    if f.IsValid() {          #### f is invalid here since there is no Field like Memory in rpc.RPCClientDriver
      if f.CanSet() {
        if f.Kind() == reflect.Int {
          x := int64(config.Memory)
          if !f.OverflowInt(x) {
            f.SetInt(x)
          }
        }
      }
    }
  }

  fmt.Println("driver memory : ", s.FieldByName("Memory").String())

  fmt.Println("Type of driver: ", reflect.TypeOf(driver))

  if err := setDriverOptionsFromEnvironment(driver); err != nil {
    return nil, err
  }

  return driver, nil
}
```

It seems the `host.Driver` return the driver of type `rpc.RPCClientDriver` which doesn't have any field like `Memory`, or else

```
...
(default) Calling .SetConfigFromFlags
E0413 23:40:41.612434     482 start.go:168] Error starting the VM: gob: type not registered for interface: drivers.CheckDriverOptions. Retrying.
Found binary path at /usr/local/bin/docker-machine-driver-kvm
Launching plugin server for driver kvm

Type of driver:  *rpcdriver.RPCClientDriver
Kind of s :  struct
driver memory :  <invalid Value>
Type of driver:  *rpcdriver.RPCClientDriver

Plugin server listening at address 127.0.0.1:39403
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(default) Calling .GetMachineName
(default) Calling .DriverName
(default) Calling .GetCreateFlags

Type of driver:  *rpcdriver.RPCClientDriver
Kind of s :  struct
driver memory :  <invalid Value>
Type of driver:  *rpcdriver.RPCClientDriver

(default) Calling .SetConfigFromFlags
E0413 23:40:41.642119     482 start.go:168] Error starting the VM: gob: type not registered for interface: drivers.CheckDriverOptions. Retrying.
...
```

Let me know if I am doing anything wrong here.

budhrg:
Closing the PR now. Sending new one.

PR #754, opened 2017-04-14
Issue #389 Enable ability to use persistent volumes using host-pv-dir

praveenkumar:
Addresses issue #389 

This patch will enable to use of `host-pv-dir` which solves persistent dir issue.

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -44,6 +44,15 @@ Feature: Basic
     minishift
     """
 
+  Scenario: Persistent volumes should be listed when minishift start
+    When executing "oc get pv --as system:admin -o=name"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     persistentvolume/pv0001
+     """
+
   Scenario: User is able to do ssh into Minishift VM
     Given Minishift has state "Running"
      When executing "minishift ssh echo hello"
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -72,6 +72,7 @@ const (
 	serverLogLevel    = "server-loglevel"
 	openshiftEnv      = "openshift-env"
 	metrics           = "metrics"
+	hostPvDir         = "host-pv-dir"
 
 	// Setting proxy
 	httpProxy   = "http-proxy"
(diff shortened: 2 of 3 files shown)


hferentschik:
Can we maybe add an integration test, showing that the persistence volumes are set up? We can query for their existence and maybe run a persistence volume claim?

test/integration/features/basic.feature
```diff
@@ -44,6 +44,15 @@ Feature: Basic
...
     """
 
+  Scenario: Persistent volumes should be listed when minishift start
+    When executing "oc get pv --as system:admin -o=name"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     persistentvolume/pv0001
+     """
```
LalatenduMohanty: This is tricky as the actual output will contain 100 PVs . @praveenkumar  I guess there should be something in Gherkin to cross check long sequential outputs. Also if we can deploy a simple application which uses PV that would be great.

praveenkumar: @LalatenduMohanty That's a very basic test, deployment a app using a PV might not work for this moment because app deployment still issue for use due to that waiting time.

LalatenduMohanty: @praveenkumar fair enough.

budhrg:
LGTM too :+1:

test/integration/features/basic.feature
hferentschik: > for this moment because app deployment still issue for use due to that waiting time.

I think we said we have a solution for that now. _rollout watch_. So writing tests using app deployments should not be an issue more. Biggest "issue" is most likely added test execution time.

However, I don't think you need to deploy an app to make a persistence volume claim. You can claim is directly via the API.

praveenkumar: > However, I don't think you need to deploy an app to make a persistence volume claim. You can claim is directly via the API.

Yes I hope you are talking about https://docs.openshift.com/enterprise/3.1/dev_guide/persistent_volumes.html#requesting-storage I did try it but problem is this should be in a file or part of stdin. I was able to create a file in setup method but oc binary is not able to pick it up. We can discuss it today.

hferentschik:
Merged. I added a short paragraph to the docs as well.

LalatenduMohanty:
@hferentschik Thanks for adding the documentation, but we should have done the review for that. Also there is a typo in it https://github.com/minishift/minishift/commit/7d799be7d19439cf83b453a613f8287c7979b552#commitcomment-21796197

PR #759, opened 2017-04-17
Issue #320 Add --follow option to 'minishift logs' command

### LINKED ISSUE #320: Add --follow option to `minishift logs` subcommand
Currently we can't follow up logs of `origin` container but only able to dump it to `stdout`. Docker provide a way to follow running container logs using `--follow` and it would be good UX if we can have it for minishift.

budhrg:
Fix #320


pkg/minikube/cluster/cluster.go
```diff
@@ -371,12 +373,30 @@ func GetHostDockerEnv(api libmachine.API) (map[string]string, error) {
...
+		if err != nil {
+			return "", err
+		}
+
+		err = c.Shell(cmdToRun)
+		if err != nil {
+			return "", errors.Wrap(err, "Error creating ssh client")
+		}
+
+		return "", err
```
praveenkumar: Shouldn't this `return "", nil`, I understand if there is no err then err will be `nil` but here it would be good to put it explicit.

pkg/minikube/cluster/cluster.go
```diff
@@ -371,12 +373,30 @@ func GetHostDockerEnv(api libmachine.API) (map[string]string, error) {
...
+	}
+
+	s, err := host.RunSSHCommand(cmdToRun)
 	if err != nil {
 		return "", nil
```
praveenkumar: I am sure this is not part of the review but why we are returning `nil` when a err occur shouldn't it be `return "", err` ?

budhrg: Fixed @praveenkumar

budhrg:
retest this please

pkg/minikube/cluster/cluster.go
```diff
@@ -39,11 +39,13 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/registration"
 	minishiftUtil "github.com/minishift/minishift/pkg/minishift/util"
 	"github.com/minishift/minishift/pkg/util"
+	"github.com/pkg/errors"
```
LalatenduMohanty: Not sure why in some cases the standard errors package is added like this. Happened to me too.

pkg/minikube/cluster/cluster.go
```diff
@@ -371,16 +373,34 @@ func GetHostDockerEnv(api libmachine.API) (map[string]string, error) {
...
+
 		return "", nil
 	}
-	return s, err
+
+	s, err := host.RunSSHCommand(cmdToRun)
+	if err != nil {
+		return "", err
+	}
+	return s, nil
```
LalatenduMohanty: Couple of suggestions. 

* I think `host.RunSSHCommand(cmdToRun)` should be in an else statement as it is mutually exclusive to `if follow` code block. I do not see  `s, err := host.RunSSHCommand(cmdToRun)` is getting executed if user has specified -f. 
* Also we can write the code without `cmdToRun` variable.

budhrg: Removed `cmdToRun` variable, ignoring 1st one as it is a standard pattern.

pkg/minikube/cluster/cluster_test.go
```diff
@@ -50,3 +53,52 @@ func TestLocalBoot2DockerURL(t *testing.T) {
...
+			cmd := logsCmd
+			if test.follow {
+				cmd = logsCmdFollow
+			}
+			if _, err = GetHostLogs(api, test.follow); err != nil {
+				t.Errorf("Error getting logs of the running OpenShift cluster: %s", err)
+			}
+			if _, ok := s.Commands[cmd]; !ok {
+				t.Errorf("Expected command %s to run but did not.", cmd)
+			}
```
LalatenduMohanty: @budhrg Any specific reason we are running the commands though the SSHServer i.e. `s.Commands[cmd]`. As you are already calling `GetHostLogs`?

budhrg: @LalatenduMohanty , We are not running any command here. We are just checking the expected command is present in the commands list in ssh server. `GetHostLogs` run the command which suppose to sent over ssh. 

Similar approach what we are doing with registration https://github.com/minishift/minishift/blob/master/pkg/minishift/registration/redhat_test.go#L112.

pkg/minikube/cluster/cluster.go
budhrg: @LalatenduMohanty I saw that this package give nice functions top of standard `errors` package. 

Also, it seems to get installed as part of other dependent package.

pkg/minikube/cluster/cluster.go
```diff
@@ -371,16 +373,32 @@ func GetHostDockerEnv(api libmachine.API) (map[string]string, error) {
...
 // GetHostLogs gets the openshift logs of the host VM.
-func GetHostLogs(api libmachine.API) (string, error) {
+// If follow is specified, it will tail the logs
+func GetHostLogs(api libmachine.API, follow bool) (string, error) {
 	host, err := CheckIfApiExistsAndLoad(api)
 	if err != nil {
 		return "", err
 	}
+
+	if follow {
```
hferentschik: How is all this in any form implementing a steaming of the logs? AFAICT, this command will just collect the output and return it in the end. Am I missing something here. I don't think that in its current form the we have the right helpers and infrastructure to implement streaming. 

I am also a bit surprised why we are trying to implement this atm. Neither it is part of the sprint afaik, nor is it easy to implement.

budhrg: @hferentschik ,  Well out of curiosity during weekend I was just going through 'logs' command and found `docker logs origin -f` is the underline command we need. So, then the main crux was how to streaming it as you mention. 

Then, I looked into [minikube code](https://github.com/kubernetes/minikube/blob/master/pkg/minikube/cluster/cluster.go#L383-L393) which seems to be pretty easy to modify and adapt here too.

budhrg:
@hferentschik , I think it would be nice to have this PR too in the release.
It's complete functionally along with tests. 

Not in a hurry though!

hferentschik:
>  I think it would be nice to have this PR too in the release.

I'd like to look at this closely first

pkg/minikube/cluster/cluster.go
```diff
@@ -371,16 +373,32 @@ func GetHostDockerEnv(api libmachine.API) (map[string]string, error) {
 }
 
 // GetHostLogs gets the openshift logs of the host VM.
-func GetHostLogs(api libmachine.API) (string, error) {
+// If follow is specified, it will tail the logs
+func GetHostLogs(api libmachine.API, follow bool) (string, error) {
```
hferentschik: I know you have not added this method, but what is it doing here in _cluster.go_? And why is it called _GetHostLogs_? Shouldn't this really be in _pkg/minishift/openshift_?

pkg/minikube/cluster/cluster.go
hferentschik: This really should be a if/else or two methods. It's a bit tricky to see the two returns and effective two different code paths

pkg/minikube/cluster/cluster_test.go
```diff
@@ -50,3 +53,52 @@ func TestLocalBoot2DockerURL(t *testing.T) {
 		t.Fatalf("Expected URL : %s", localISOUrl)
 	}
 }
+
+func TestHostGetLogs(t *testing.T) {
+	api := tests.NewMockAPI()
```
hferentschik: Can you run me through this test and explain to me how this is a test for the tailing functionality? Or what you are testing at all here?

At best you are testing that the log command w/ and w/o the -f option is received by the ssh server. But what does that prove really? Am I missing something here?

pkg/minikube/cluster/cluster_test.go
hferentschik: > Similar approach what we are doing with registration

Sure, but in registration it is a different ball game. We are testing that different commands are received depending on other conditions. In the case of the logs command, there is no question really. As I said above, at best this proves that the -f flag is passed along. Not sure whether this helps anything. This is not a test for the actual follow functionality.

hferentschik: > Any specific reason we are running the commands though the SSHServer i.e. s.Commands[cmd]. As you are already calling GetHostLogs?

@LalatenduMohanty This is a mock ssh server. You use it, to verify that a certain command is going over the ssh connection to the VM. The idea is that you execute the actual production code and then verify against the mock server whether the expected commands got received in a given order.

budhrg: > As I said above, at best this proves that the -f flag is passed along. Not sure whether this helps anything. This is not a test for the actual follow functionality

For that case, we need to write integration test then to verify logs are being shown as part of -f

pkg/minikube/cluster/cluster.go
budhrg: Yes, it is inherited from Minikube fork, I was wondering the same. will update the name and also move it to openshift domain

pkg/minikube/cluster/cluster_test.go
budhrg: Yes @hferentschik , was only trying to verify the log commands to be received by ssh server. 

To really test logs are working, we need to put test in integration test.

hferentschik: I don't really think this test serves any purpose here. It's a lot of code barely verifying anything. I suggest we remove this test.

pkg/minikube/cluster/cluster.go
hferentschik: > will update the name and also move it to openshift domain

+1

pkg/minikube/cluster/cluster_test.go
budhrg: ok @hferentschik

hferentschik:
@budhrg are you going to update this pull request?

hferentschik:
@budhrg ok, thanks

Issue #729, opened 2017-04-10
OcRunner  needs to handle spaces in quoted arguments


NickLaMuro:
Given the following addon:

**./tmp/minishift_addons/foo/foo.adon**
```bash
# Name: foo
# Description: create foo project

oc adm new-project foo --description='Bar Baz'
```

You would get the following error when attempting to boot minishift with that addon enabled:

```console
$ minishift addons install ./tmp/minishift_addons/foo
$ minishift addons enable foo
$ minishift start
...
Applying addon foo:
-- adm new-project foo --description='Bar Baz'
error: you must specify one argument: project name
See 'oc adm new-project -h' for help and examples.
E0407 17:09:06.185670   20557 start.go:232] Error executing addon commands Error executing command oc adm new-project foo --description='Bar Baz'.
```

Bug Location
-------------
I think the problem is within the `strings.Split()` that happens in to locations in the `pkg/minishift/oc/oc.go` file:

- https://github.com/minishift/minishift/blob/cc4c2a2/pkg/minishift/oc/oc.go#L56
- https://github.com/minishift/minishift/blob/cc4c2a2/pkg/minishift/oc/oc.go#L65

The argument of `oc adm new-project foo --description='Bar Baz'` would be split into something like this:

```go
["oc", "adm", "new-project", "foo", "--description='Bar", "Baz'"]
```

Which, when passed to `os/exec`'s `Run()` function, would wrap all args in quotations and attempt to run it.  When `oc` parses the command, it sees both `"foo"` and `"Baz'"` as inputs for the project name, and returns that error accordingly (`"--description='Bar"` would also probably be considered malformed, but hard to say based on the output alone).

Possible solution
-----------------
I have put together a lib that _**might**_ be a better alternative to the `strings.Split()` here:

https://github.com/NickLaMuro/go_cmd_arg_parser

As the README in the project suggests, I am neither an experienced programmer in `go`, nor the original contributor to that function (shamelessly stolen from stackoverflow; again see README), so would advise at the very least vetting this code heavily before considering implementing it, if not re-writing it yourself.

That said, it also handles the corner case of passing in escaped quotes into a string, which is something I expect a user might do.

hferentschik:
@NickLaMuro, thanks for the bug report and this very thorough digging. You are right, this is an issue with the _oc_ runner helper. So in fact, this will not only affect add-ons, but also any other place in the code where we use the runner. We need to handle this better indeed. We will look at your suggested approach.

[...]

Issue #221, opened 2016-12-23
Unable to use cached boot2docker.iso in Windows


budhrg:
Hi,

Unable to use cached boot2docker.iso in Windows 10 for both HyperV and VirtualBox  5.0.30 drivers.
And, for HyperV, it is directory communicating to github url for ISO url. 

I need to start with `--iso-url` option as `minishift.exe start --vm-driver virtualbox --iso-url file://C:/boot2docker.iso`.

### VirtualBox
```
PS D:\> ls C:\Users\budhram\.minishift\cache\iso\
    Directory: C:\Users\budhram\.minishift\cache\iso
Mode                LastWriteTime         Length Name
----                -------------         ------ ----
-a----       12/29/2016   7:37 PM       37748736 boot2docker.iso

PS D:\> .\minishift.exe version
minishift version: 1.0.0-beta.1

PS D:\> .\minishift.exe start --vm-driver virtualbox -v 5 --show-libmachine-logs
E1230 13:54:23.197297    5848 update.go:76] Get https://api.github.com/repos/jimmidyson/minishift/releases/latest: dial tcp: lookup api.github.com: getaddrinfow: No such host is known.
Starting local OpenShift cluster...
Found binary path at D:\minishift.exe
Launching plugin server for driver virtualbox
Plugin server listening at address 127.0.0.1:49698
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(minishift) Calling .GetMachineName
(minishift) Calling .DriverName
Running pre-create checks...
(minishift) Calling .PreCreateCheck
(minishift) DBG | COMMAND: C:\Program Files\Oracle\VirtualBox\VBoxManage.exe --version
(minishift) DBG | STDOUT:
(minishift) DBG | {
(minishift) DBG | 5.0.30r112061
(minishift) DBG | }
(minishift) DBG | STDERR:
(minishift) DBG | {
(minishift) DBG | }
(minishift) DBG | Hyper-V is not installed.
(minishift) DBG | %!(EXTRA *exec.Error=exec: "vmms.exe": executable file not found in %PATH%)COMMAND: wmic cpu get VirtualizationFirmwareEnabled
(minishift) DBG | STDOUT:
(minishift) DBG | {
(minishift) DBG | VirtualizationFirmwareEnabled
(minishift) DBG | TRUE
(minishift) DBG |
(minishift) DBG | }
[...]

PR #764, opened 2017-04-18
Issue #729 Adding a helper function to split command string

hferentschik:
Addresses issue #729 

Pretty much as suggested in issue. Added some more comments and additional unit test.

### DIFF
--- a/pkg/util/cmd/split_args.go
+++ b/pkg/util/cmd/split_args.go
@@ -1,0 +1,77 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"strings"
+	"unicode"
+)
+
+// This is a '\'
+const escapeRune = rune(92)
+
+// SplitCmdString takes a single line command string and splits it into individual arguments to be used with
+// exec.Command. It does so by honouring quotes within the original command
+func SplitCmdString(cmd string) []string {
+	lastQuote := rune(0)
+	consecutiveEscapeCount := 0
+	splitWithQuotes := func(c rune) bool {
+		switch {
+		// don't split if the character matches the "last matched quote"
+		case c == lastQuote:
+			if consecutiveEscapeCount%2 == 0 {
+				lastQuote = rune(0)
+			}
+			consecutiveEscapeCount = 0
+			return false
+		// found an escape character, don't split
+		case c == escapeRune:
+			consecutiveEscapeCount += 1
+			return false
+		// don't split if we are still in a quoted string
+		case lastQuote != rune(0):
+			consecutiveEscapeCount = 0
+			return false
+		// don't split if the character is a quote type rune, and set the lastQuote value to that
+		case unicode.In(c, unicode.Quotation_Mark):
+			lastQuote = c
+			consecutiveEscapeCount = 0
+			return false
+		// split the string if it is a space and none of the other conditionals are satisfied
+		default:
+			consecutiveEscapeCount = 0
+			return unicode.IsSpace(c)
+		}
+	}
+	result := strings.FieldsFunc(cmd, splitWithQuotes)
+	return trimOuterQuotesFromArgs(result)
+}
+
+// Remove outer quotation marks from string
+func trimOuterQuotesFromArgs(args []string) []string {
+	var result []string
+	for i := range args {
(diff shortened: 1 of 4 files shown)


pkg/util/cmd/split_args.go
```diff
@@ -0,0 +1,77 @@
...
+	splitWithQuotes := func(c rune) bool {
+		switch {
+		// don't split if the character matches the "last matched quote"
+		case c == lastQuote:
+			if consecutiveEscapeCount%2 == 0 {
+				lastQuote = rune(0)
+			}
+			consecutiveEscapeCount = 0
+			return false
+		// found an escapce character, don't split
```
NickLaMuro: Typo:  "escape"

pkg/util/cmd/split_args.go
```diff
@@ -0,0 +1,77 @@
...
+		// don't split if we are still in a quoted string
+		case lastQuote != rune(0):
+			consecutiveEscapeCount = 0
+			return false
+		// don't split if the character is a quote type rune, and set the lastQuote value to that
+		case unicode.In(c, unicode.Quotation_Mark):
+			lastQuote = c
+			consecutiveEscapeCount = 0
+			return false
+		// split the string if it is a space
```
NickLaMuro: Not sure if it is worth noting in this comment, but maybe include something along the lines of "... and none of the other conditionals are satisfied".   That said, these are all very concise comments, so maybe leave it as is if that was the goal (I myself tend to get a little lengthy... character flaw...).

pkg/util/cmd/split_args_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if args[3] != "01/01/1900" {
+		t.Errorf("Expected the second arg to be '01/01/1900', but was '%s'", args[2])
+	}
+
+	if args[4] != "+%s" {
+		t.Errorf("Expected the second arg to be '+%%s', but was '%s'", args[2])
+	}
+}
+
+func TestProperlyHandlesNestedSingleQuotesWithinDoubleQuotes(t *testing.T) {
```
budhrg: :+1:

pkg/util/cmd/split_args_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if args[3] != "01/01/1900" {
+		t.Errorf("Expected the second arg to be '01/01/1900', but was '%s'", args[2])
+	}
+
+	if args[4] != "+%s" {
+		t.Errorf("Expected the second arg to be '+%%s', but was '%s'", args[2])
+	}
+}
+
+func TestProperlyHandlesNestedSingleQuotesInNestedDoubleQuotes(t *testing.T) {
```
budhrg: :+1:

pkg/util/cmd/split_args_test.go
```diff
@@ -0,0 +1,175 @@
...
+	if args[3] != "01/01/1900" {
+		t.Errorf("Expected the second arg to be '01/01/1900', but was '%s'", args[2])
+	}
+
+	if args[4] != "+%s" {
+		t.Errorf("Expected the second arg to be '+%%s', but was '%s'", args[2])
+	}
+}
+
+func TestProperlyHandlesEscapedDoubleQuotesWithinDoubleQuotes(t *testing.T) {
```
budhrg: :+1:

pkg/util/cmd/split_args_test.go
```diff
@@ -0,0 +1,175 @@
...
+func TestReturnsAnArrayOfStrings(t *testing.T) {
+	args := SplitCmdString(`date 1423`)
+
+	if args[0] != "date" {
+		t.Errorf("Expected the first arg to be 'date', but was '%s'", args[0])
+	}
+
+	if args[1] != "1423" {
+		t.Errorf("Expected the second arg to be '1423', but was '%s'", args[1])
+	}
```
praveenkumar: Not a blocker but we can do this like and can be applied for each test case.

```
expectedArgsList := []string{"date", "1423"}
for i, v := range expectedArgsList {
		if args[i] != v {
			t.Errorf("Expected the %d arg to be %s, but Got '%s'", (i+1), v, args[i])
		}
	}
```

pkg/util/cmd/split_args_test.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+func TestProperlyHandlesWhitespaceWithinSingleQuotes(t *testing.T) {
+	args := SplitCmdString(`date -f '%a %b %d %T %Z %Y' "01/01/1900" '+%s'`)
```
budhrg: If we use approach suggested here https://github.com/minishift/minishift/pull/764/files/14f24adb92d3ed3f95fc78c4c2f0dbee7d5b7d2f#r112129096, then I believe, this test case would be much smaller.

budhrg:
@hferentschik , here the reference to issue id is wrong it should be https://github.com/minishift/minishift/issues/729

hferentschik:
> here the reference to issue id is wrong it should be #221

thanks, will fix

pkg/util/cmd/split_args.go
hferentschik: thanks :-)

pkg/util/cmd/split_args.go
hferentschik: added it to be explicit

Issue #749, opened 2017-04-13
Introduce an addon command "echo"


LalatenduMohanty:
It will be helpful for printing meaningful messages to the console when an addon executes.

hferentschik:
_echo_ or _print_?

LalatenduMohanty:
@hferentschik If we want to keep it simple and stupid I would suggest echo as it will just echo the string from the addon file.

LalatenduMohanty:
@jorgemoralespou Any suggestion?

budhrg:
Closed via https://github.com/minishift/minishift/issues/766.

PR #766, opened 2017-04-18
Issue #254, #749 Adding docs and test to echo command and using it as part of the registry addon

hferentschik:
Addresses issue #254 and #749


hferentschik:
@praveenkumar, @LalatenduMohanty - please have a look. This is a combination of your pull requests, adjusting bits and pieces as I tested the echo functionality as part of the 'registry-route' add-on

addons/registry-route/registry-route.addon
```diff
@@ -0,0 +1,18 @@
...
+ssh sudo mkdir -p /etc/docker/certs.d/docker-registry-default.#{routing-suffix}:443
+ssh sudo chown docker /etc/docker/certs.d/docker-registry-default.#{routing-suffix}:443
+
+ssh sudo mkdir -p /etc/docker/certs.d/docker-registry-default.#{routing-suffix}
+ssh sudo chown docker /etc/docker/certs.d/docker-registry-default.#{routing-suffix}
+
+docker cp origin:/var/lib/origin/openshift.local.config/master/ca.crt /etc/docker/certs.d/docker-registry-default.#{routing-suffix}:443/ca.crt
+docker cp origin:/var/lib/origin/openshift.local.config/master/ca.crt /etc/docker/certs.d/docker-registry-default.#{routing-suffix}/ca.crt
+
+echo Add-on '#{addon-name}' created docker-registry route. You can now login to the OpenShift registry via:
```
LalatenduMohanty: Suggestion  /`You can now login to the OpenShift registry via:`/Please run following commands to login to the OpenShift docker registry/

LalatenduMohanty: Also when an user has multiple addon this is not very obvious for the user . e.g. 
```
......
Applying addon registry-route:........
  Add-on 'registry-route' created docker-registry route. You can now login to the OpenShift registry via:
  eval $(minishift docker-env)
  docker login -u developer -p `oc whoami -t` docker-registry-default.192.168.64.7.nip.io:443

Applying addon test-echo:
  Hello World

Applying addon admin-user:.user "admin" created
.cluster role "cluster-admin" added: "admin"

```

docs/source/using/addons.adoc
```diff
@@ -71,6 +71,11 @@ running as well. This is similar to running `eval $(minishift docker-env)` on yo
 then executing any `docker` command. See also
 link:../command-ref/minishift_docker-env{outfilesuffix}[`minishift docker-env`].
 
+echo::
+If the add-on command starts with _echo_, the arguments following the _echo_ command are printed
+to the console. This can be used to provide additional feedback during add-on execution or to print
+a user hint at the end of the execution.
```
LalatenduMohanty: I think we should just keep `This can be used to provide additional feedback during add-on execution` and remove `or to print a user hint at the end of the execution` as this is not required explicitly IMO.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -182,6 +185,8 @@ func (m *AddOnManager) applyAddOn(addOn addon.AddOn, context *command.ExecutionC
 			return err
 		}
 	}
+	fmt.Println()
+	fmt.Println()
```
LalatenduMohanty: May be we can use single line  fmt.Printf("\n\n").

addons/registry-route/registry-route.addon
```diff
@@ -0,0 +1,18 @@
...
+echo eval $(minishift docker-env)
+echo docker login -u developer -p `oc whoami -t` docker-registry-default.#{routing-suffix}:443
```
LalatenduMohanty: Should we put `$ ` in front of the commands to make it obvious?

addons/registry-route/registry-route.addon
hferentschik: > Also when an user has multiple addon this is not very obvious for the user 

What is not obvious?

hferentschik: > Suggestion /You can now login to the OpenShift registry via:/Please run following commands to login to the OpenShift docker registry/

Sure

docs/source/using/addons.adoc
hferentschik: ok

pkg/minishift/addon/manager/addon_manager.go
hferentschik: I guess you mean 'fmt.Print("\n\n")'. Sure

addons/registry-route/registry-route.addon
hferentschik: +1, I actually was contemplating this myself.

hferentschik:
@LalatenduMohanty applied your suggestions and fixed another bug along the way. Please have another look

hferentschik:
retest this please

LalatenduMohanty:
@hferentschik I do not think the user can not install the registry-route addon just using the minishift binary i.e. without the source for the addon. As per the code we are only calling `bindata.RestoreAssets` for the default addons.  So raised an issue for this #771

hferentschik:
>  I do not think the user can not install the registry-route addon just using the minishift binary

That's correct. That's what we discussed, right? For now we are not really exposing this add-on. I thought we agreed on this.

Issue #758, opened 2017-04-16
Add new command 'openshift registry'


hferentschik:
We need to add a command which provides the host+port information for login and use of the OpenShift registry.

    $ minishift openshift registry
    172.30.1.1:5000

We can also add a _login_ flag or subcommand:

    $ minishift openshift registry --login
   
This print and executes the docker login command. This way he user does not have to retrieve the login token. We can handle this transparently. Obviously we need to check whether the _docker_ binary is available and if not return an appropriate errors message.

LalatenduMohanty:
@hferentschik couple of things.  If we are adding a sub command then it can not depend on the registry addon which going to expose the  route. So in this case we need to do the required for exposing the registry in the command itself. 

Also I am not sure how to retrieve the registry information i.e. `172.30.1.1:5000` other than hard coding it.

hferentschik:
>  couple of things. If we are adding a sub command then it can not depend on the registry addon which going to expose the route. 

What do you mean with "depend".  The command does not have to know whether the add-on ran or not. We can dynamically check whether the route is exposed or not. If so it can print the route host+port, otherwise the internal host and port. We can fully abstract the user from all the issues around the routes. See also  https://github.com/minishift/minishift/pull/704#issuecomment-293925484.

> Also I am not sure how to retrieve the registry information i.e. 172.30.1.1:5000 other than hard coding it.

https://github.com/minishift/minishift/pull/704#issuecomment-293925484

     $ oc --as system:admin get -o "jsonpath={.spec.clusterIP}:{.spec.ports[*].port}" service/docker-registry -n default

We know that the docker-registry service exists. Using _oc_ we can query it and extract what we want. We can also use _oc_ to check whether the route exist. If so, we can use it's details.

LalatenduMohanty:
> #704 (comment)
> 
>  $ oc --as system:admin get -o "jsonpath={.spec.clusterIP}:{.spec.ports[*].port}" service/docker-registry -n default

Right, this will be helpful.

hferentschik:
Merged via pull request #767

PR #767, opened 2017-04-18
Issue #758 Adding minishift openshift registry

LalatenduMohanty:

Addresses issue #758

### DIFF
--- a/cmd/minishift/cmd/openshift/registry.go
+++ b/cmd/minishift/cmd/openshift/registry.go
@@ -1,0 +1,61 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openshift
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/docker/machine/libmachine"
+	"github.com/minishift/minishift/cmd/minishift/cmd/util"
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/openshift"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+)
+
+var registryCmd = &cobra.Command{
+	Use:   "registry [flags]",
+	Short: "Prints host IP and port of the OpenShift registry.",
+	Long:  `Prints host IP and port of the OpenShift docker registry in the format 'host IP:port'.`,
+	Run: func(cmd *cobra.Command, args []string) {
+
+		//Check if Minishift VM is running
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		defer api.Close()
+
+		util.ExitIfUndefined(api, constants.MachineName)
+
+		host, err := api.Load(constants.MachineName)
+		if err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		}
+
+		util.ExitIfNotRunning(host.Driver)
+
+		registryInfo, err := openshift.GetDockerRegistryInfo()
+		if err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		}
+
+		fmt.Fprintln(os.Stdout, registryInfo)
+	},
+}
+
+func init() {
+	OpenShiftConfigCmd.AddCommand(registryCmd)
+}
--- a/pkg/minishift/openshift/registry.go
+++ b/pkg/minishift/openshift/registry.go
@@ -1,0 +1,40 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openshift
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	instanceState "github.com/minishift/minishift/pkg/minishift/config"
+)
+
+func GetDockerRegistryInfo() (string, error) {
+	namespace := "default"
+	service := "service/docker-registry"
+	cmdArgText := fmt.Sprintf("get -o jsonpath={.spec.clusterIP}:{.spec.ports[*].port} %s -n %s --config=%s", service, namespace, systemKubeConfigPath)
+	tokens := strings.Split(cmdArgText, " ")
+	cmdName := instanceState.InstanceConfig.OcPath
+	cmdOut, err := runner.Output(cmdName, tokens...)
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("No information found for '%s'", service))
+	}
+
+	registryInfo := byteArrayToString(cmdOut)
+	return registryInfo, nil
+}
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -74,6 +74,16 @@ Feature: Basic
       export "BAZ=BAT"
       """
 
+  Scenario: Checking OpenShift registry
+    Given Minishift has state "Running"
+     When executing "minishift openshift registry"
+     Then stderr should be empty
+      And exitcode should equal 0
+      And stdout should contain
+      """
+      172.30.1.1:5000
+      """
+
   Scenario: Stopping Minishift
     Given Minishift has state "Running"
      When executing "minishift stop"


budhrg:
retest this please

cmd/minishift/cmd/openshift/registry.go
```diff
@@ -0,0 +1,44 @@
...
+
+var registryCmd = &cobra.Command{
+	Use:   "registry [flags]",
+	Short: "Prints the docker registry endpoint information.",
+	Long:  `Prints the openshift docker registry service endpoint information.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		registryInfo, err := openshift.GetDockerRegistryInfo()
+		if err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		} else {
```
praveenkumar: No need of `else` block here because if `err` is not nil then process get terminated.

pkg/minishift/openshift/registry.go
```diff
@@ -0,0 +1,46 @@
...
+	"fmt"
+	"path/filepath"
+	"strings"
+
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	instanceState "github.com/minishift/minishift/pkg/minishift/config"
+)
+
+func init() {
+	systemKubeConfigPath = filepath.Join(constants.Minipath, "machines", constants.MachineName+"_kubeconfig")
```
praveenkumar: -1 Why this is part of `init` method? this Can be a simple variable if needed I think.

budhrg:
retest this please

budhrg:
@LalatenduMohanty Some integration tests are failing, might wanna have a look.

pkg/minishift/openshift/registry.go
hferentschik: +1, we should stay away from using _init_ as much as possible

cmd/minishift/cmd/openshift/registry.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+
+	"github.com/minishift/minishift/pkg/minishift/openshift"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+)
+
+var registryCmd = &cobra.Command{
+	Use:   "registry [flags]",
+	Short: "Prints the docker registry endpoint information.",
```
hferentschik: "docker registry endpoint"? that does not sound right.
 What's about: "Prints host and IP of the OpenShift registry" (short)
"Prints host and IP of the OpenShift registry in the format 'host:port'" (long)

cmd/minishift/cmd/openshift/registry.go
hferentschik: +1

pkg/minishift/openshift/registry.go
LalatenduMohanty: @hferentschik @praveenkumar Why? I thought init functions are nice ways to make sure that some variables/statements get evaluated at the end of rest of the variable initializations. May be I am missing something here.

cmd/minishift/cmd/openshift/registry.go
```diff
@@ -0,0 +1,44 @@
...
+	Long:  `Prints the openshift docker registry service endpoint information.`,
+	Run: func(cmd *cobra.Command, args []string) {
+		registryInfo, err := openshift.GetDockerRegistryInfo()
```
hferentschik: You should add a VM running state check here. If the VM and hence OpenShift is not running, or not even defined, there is no point to execute an _oc_ command. The error will only confuse. 

We already have several commands which check whether host exists at all and whether it is running or not.

LalatenduMohanty:
I will update https://github.com/minishift/minishift/pull/679 after this is merged. 

Regarding tests I am not sure what kind of tests I can add. suggestions?

LalatenduMohanty:
I will update https://github.com/minishift/minishift/pull/679 after this is merged. 

Regarding tests,  I am not sure what kind of tests I can add as part of this PR. My be an integration test, suggestions? @budhrg @hferentschik @praveenkumar

LalatenduMohanty:
@budhrg  `make integration` works for me in the local system. 
```
$ make integration
go install -pkgdir=/Users/lmohanty/code/gocode/src/github.com/minishift/minishift/out/bindata -ldflags="-X github.com/minishift/minishift/pkg/version.version=1.0.0-rc.1 -X github.com/minishift/minishift/pkg/version.isoVersion=v1.0.2 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v1.5.0-rc.0" ./cmd/minishift
go test -timeout 3600s github.com/minishift/minishift/test/integration --tags=integration -v -args --binary /Users/lmohanty/code/gocode/bin/minishift ""
Running Integration test in: /var/folders/81/q3z841ds2yb7xblc91xww10m0000gn/T/minishift-integration-test-054001306
using binary: /Users/lmohanty/code/gocode/bin/minishift
......................................... 41


9 scenarios (9 passed)
41 steps (41 passed)
5m31.964870701s
PASS
ok  	github.com/minishift/minishift/test/integration	331.979s
```

hferentschik:
I think this is a timing issue. The persistent volumes might not be created in time. We need to investigate. We might need to implement the discussed retry step now. We cannot keep backing out otherwise valuable tests.

LalatenduMohanty:
Right, the PV test is failing and it is a timing issue. If you check the PR the step 49 is the integration test for PV. 

```
20:41:18 --- Failed steps:
20:41:18 
20:41:18     Then  stderr should be empty # features/basic.feature:49
20:41:18       Error: Output did not match. Expected: , Actual: No resources found.
20:41:18 
```

pkg/minishift/openshift/registry.go
praveenkumar: @LalatenduMohanty As per https://golang.org/doc/effective_go.html#init `Use of init functions is to verify or repair correctness of the program state before real execution begins.` In current code segment if I directly use `systemKubeConfigPath = filepath.Join(constants.Minipath, "machines", constants.MachineName+"_kubeconfig")` doesn't that get evaluated? I don't see anything here we are correcting as part of init method.

LalatenduMohanty:
Raised #774  for tracking the integration test timing issue.

praveenkumar:
> May be an integration test, suggestions? @budhrg @hferentschik @praveenkumar

Yes create a integration test where you execute `minishift openshift registry` and it should provide required output.

pkg/minishift/openshift/registry.go
LalatenduMohanty: @praveenkumar yes, I had similar understanding. The idea is to make sure systemKubeConfigPath contains the required value even if it is evaluated something else in the same file. But at this point the registry.go is a small file and does not contain much. So I am fine with making it as a variable.

LalatenduMohanty: @praveenkumar However I do not see any downside to this.

praveenkumar: @LalatenduMohanty Usually using init method without any concrete usecase is highly discouraged in golang world afaik so if you don't have any strong reason to use then better to avoid.

LalatenduMohanty: @praveenkumar I can not use variable instead of init method as  it is already initialized as part of service.go https://github.com/minishift/minishift/blob/master/pkg/minishift/openshift/service.go#L35-L37. So either I use it or I can reuse the `systemKubeConfigPath` without doing anything. WDYT?

LalatenduMohanty:
> Yes create a integration test where you execute minishift openshift registry and it should provide required output.

Yes, it is already done.

pkg/minishift/openshift/registry.go
praveenkumar: @LalatenduMohanty Since both are part of openshift package then better to go with using `systemKubeConfigPath` instead again creating same variable for in this file. Also if you want to remove init method then you also need to change the variable name otherwise it will conflict with name which is already in package.  Below changes works without any init method.

```
var systemKubeConfig = filepath.Join(constants.Minipath, "machines", constants.MachineName+"_kubeconfig")
cmdArgText := fmt.Sprintf("get -o jsonpath={.spec.clusterIP}:{.spec.ports[*].port} %s -n %s --config=%s", service, namespace, systemKubeConfig) 
```

cmd/minishift/cmd/openshift/registry.go
```diff
@@ -0,0 +1,58 @@
...
+	Long:  `Prints host and IP of the OpenShift docker registry in the format 'host:port'.`,
+	Run: func(cmd *cobra.Command, args []string) {
+
+		//Check if Minishift VM is running
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		defer api.Close()
+
+		_, err := cluster.CheckIfApiExistsAndLoad(api)
+		if err != nil {
+			fmt.Fprintln(os.Stdout, "No Minishift instance exists. Minishift should be in runnning state for this command to work")
```
hferentschik: -1, you should use ExitWithMessage

hferentschik: also why are you using fmt.Fprintln here?

hferentschik: This is just the existence check. You also need the running check

hferentschik: check _mount.go_:

		defer api.Close()

		util.ExitIfUndefined(api, constants.MachineName)

		host, err := api.Load(constants.MachineName)
		if err != nil {
		}

		util.ExitIfNotRunning(host.Driver)

cmd/minishift/cmd/openshift/registry.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/minishift/minishift/cmd/minishift/cmd/util"
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/openshift"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+)
+
+var registryCmd = &cobra.Command{
+	Use:   "registry [flags]",
+	Short: "Prints host and IP of the OpenShift registry.",
```
hferentschik: host + port

pkg/minishift/openshift/registry.go
```diff
@@ -0,0 +1,40 @@
...
+	instanceState "github.com/minishift/minishift/pkg/minishift/config"
+)
+
+func GetDockerRegistryInfo() (string, error) {
+	namespace := "default"
+	service := "service/docker-registry"
+	cmdArgText := fmt.Sprintf("get -o jsonpath={.spec.clusterIP}:{.spec.ports[*].port} %s -n %s --config=%s", service, namespace, systemKubeConfigPath)
+	tokens := strings.Split(cmdArgText, " ")
+	cmdName := instanceState.InstanceConfig.OcPath
+	cmdOut, err := runner.Output(cmdName, tokens...)
```
hferentschik: Actually, the command is not quite right. We have an abstraction to run _oc_ commands and it would be safer to run with _as_

LalatenduMohanty: Do you mean https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/start.go#L217 ?

hferentschik:
Merged, thanks.

Issue #204, opened 2016-12-20
Automatic version upgrade is broken


cmoulliard:
OC : MacOS El Capitan - 10.11.6

```
minishift start
Starting local OpenShift cluster...
Downloading ISO
 36.00 MB / 36.00 MB [========================================================================================================================================================================================] 100.00% 0s
E1220 12:44:16.250063    2462 start.go:130] Error starting host: Error creating. Error creating machine: Error in driver during machine creation: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (60) exceeded. Retrying.
```

astefanutti:
Maybe a duplicate of #110 and #160.

hferentschik:
@cmoulliard which hypervisor are you trying to target? VirtualBox or xhyve? Can you try with an explicit --vm-driver option

     $ minishift start --vm-driver [virtualbox|xhyve]

Also, if you are using xhyve, make sure to get the latest release. 

Last but not least, make sure to call `minishift delete` prior to starting or even more aggressively delete the _~/.minishift_ directory.

hferentschik:
@cmoulliard btw, the default on OS X is the xhyve hypervisor. Is this what you want to use?

cmoulliard:
yes xhyve hypervisor.

astefanutti:
@cmoulliard I faced the same issue and reinstalled / upgraded xhyve and docker-machine-driver-xhyve and it worked for me.

cmoulliard:
I experiment a new problem now

```
minishift start
There is a newer version of minishift available. Do you want to
automatically update from v0.9.0 to v1.0.0-beta.1 now? [y/N] y
Updating to version 1.0.0-beta.1
Downloading updated binary checksum to validate updated binary
E1221 09:07:44.140066    9558 update.go:155] Cannot download checksum: received 404
```

jimmidyson:
See https://github.com/minishift/minishift/blob/6b83d570d5f3c4b5f1680dc436f236c018146344/pkg/minikube/update/update.go#L48

LalatenduMohanty:
@cmoulliard for workaround you can download the latest binary from https://github.com/minishift/minishift/releases/tag/v1.0.0-beta.1 and use it and it will work.

LalatenduMohanty:
The update process from 0.9.0 to 1.0.0-beta1 is broken as @cmoulliard  @jimmidyson mentioned.

`
[...]

Issue #72, opened 2016-09-09
Add update command


davsclaus:
It would be good with an update command as not many users would know that its the status command that can tell if there is a newer version and do an auto update.

jimmidyson:
BTW all commands apart from version trigger an update check.

davidaparicio:
```
There is a newer version of minishift available. Do you want to
automatically update from v0.8.0 to v0.9.0 now? [y/N] Y
Skipping autoupdate
```
It asks to me, only one time..
Why "Y" isn't working? Is it only "y" (lowercase)?

LalatenduMohanty:
Yes, it seems it is only checking for a small `y` 

```
                if confirm == "y" {
                        fmt.Printf("Updating to version %s\n", latestVersion)
                        updateBinary(latestVersion, downloadBinary, updateLinkPrefix, downloadLinkFormat)
                        return
                }
```

jimmidyson:
@davidaparicio @LalatenduMohanty Fixed in #127

hferentschik:
I kind of like the idea of an explicit _upate-vm_ command. There is the auto-update, but at least imo it gets into the way every now and then and people also report that they error/warning messages from exceeding the rate limit towards the GitHub API. I think I'd prefer a dedicate command to check the VM version and potentially update it. 

Also, given there is not much feature wise which changes in the VM over time, I wonder whether the automatic checks are not a bit of an overkill anyways.

hferentschik:
Merged via pull request #930

PR #769, opened 2017-04-18
Issue #72 Add update command

dharmit:
`minishift update` prompts user to update minishift if a newer version
is available

### DIFF
--- a/cmd/minishift/cmd/update.go
+++ b/cmd/minishift/cmd/update.go
@@ -1,0 +1,40 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
@dharmit Does this PR work for you? I think it should not as you need to also fix https://github.com/minishift/minishift/issues/204#issuecomment-268503443  to make it work.

hferentschik:
@dharmit, thanks for the pull request. Have you actually tested it. Adding the update command is one part, but the release format has changed as well. The binaries were released directly before, but now they are in zip/tar bundles. AFAIR _MaybeUpdateFromGithub_ downloads from the release page which will be the bundle, not the extracted binary. 

To make this work, one has to download the bundle, unpack it and then do the binary update.

hferentschik:
add to whitelist

dharmit:
@LalatenduMohanty @hferentschik Thanks for the explanation. It helps a lot. I'll close this PR for time being and work on the next part which is downloading the tar/zip and unpacking it. Will create a new PR once I've done that.

Issue #653, opened 2017-04-04
Update EAP addon to have correct EAP templates


budhrg:
Hi,

I am just wondering why EAP templates doesn't have templates with no database for general templates. 
Are we are only providing persistent related templates?

When I do `oc get templates -n openshift`, I am getting following: 
```
oc get templates -n openshift
NAME                                DESCRIPTION                                                                        PARAMETERS        OBJECTS
cakephp-mysql-persistent            An example CakePHP application with a MySQL database. For more information ab...   20 (4 blank)      9
dancer-mysql-persistent             An example Dancer application with a MySQL database. For more information abo...   17 (5 blank)      9
django-psql-persistent              An example Django application with a PostgreSQL database. For more informatio...   18 (5 blank)      9
jenkins-ephemeral                   Jenkins service, without persistent storage....                                    6 (all set)       6
jenkins-persistent                  Jenkins service, with persistent storage....                                       7 (all set)       7
jenkins-pipeline-example            This example showcases the new Jenkins Pipeline integration in OpenShift,...       16 (4 blank)      8
logging-deployer-account-template   Template for creating the deployer account and roles needed for the aggregate...   0 (all set)       9
logging-deployer-template           Template for running the aggregated logging deployer in a pod. Requires empow...   35 (18 blank)     1
mariadb-persistent                  MariaDB database service, with persistent storage. For more information about...   8 (3 generated)   4
mongodb-persistent                  MongoDB database service, with persistent storage. For more information about...   9 (3 generated)   4
mysql-persistent                    MySQL database service, with persistent storage. For more information about u...   9 (3 generated)   4
nodejs-mongo-persistent             An example Node.js application with a MongoDB database. For more information...    17 (4 blank)      9
postgresql-persistent               PostgreSQL database service, with persistent storage. For more information ab...   8 (2 generated)   4
rails-pgsql-persistent              An example Rails application with a PostgreSQL database. For more information...   21 (4 blank)      9
```

It seems that CDK 2* had templates with no persistent, : 
```
### CDK 2.4
oc get templates -n openshift
NAME                                 DESCRIPTION                                                                        PARAMETERS      OBJECTS
cakephp-example                      An example CakePHP application with no database. For more information about u...   13 (4 blank)    6
cakephp-mysql-example                An example CakePHP application with a MySQL database. For more information ab...   19 (4 blank)    8
eap64-basic-s2i                      Application template for EAP 6 applications built using S2I.                       13 (3 blank)    5
[...]

PR #772, opened 2017-04-19
Issue #653 Add EAP7 templates and imagestream

praveenkumar:
Templates and ImageStream are taken from https://github.com/jboss-openshift/application-templates/tree/ose-v1.3.5


LalatenduMohanty:
I can see the EAP7 in the web console under Java.

Issue #775, opened 2017-04-19
Add the Fuse Integration Services 2.0 templates


praveenkumar:
Templates are available https://github.com/jboss-fuse/application-templates/tree/GA/quickstarts

hferentschik:
Merged via pull request #776

PR #776, opened 2017-04-19
Issue #775 Add Fuse templates and imagestream

praveenkumar:
@LalatenduMohanty Please test it out and also ask fuse folks to test it out.


hferentschik:
Merged, thanks. As discussed, I did not verify whether this templates work or not. Let's point the Fuse guys to it and see whether we get some feedback.

PR #773, opened 2017-04-19
Issue #305 Dynamic driver configuration

hferentschik:
Fix for issue #305 and #686 (by the fact that the code got refactored and also applies now the host-only-cidr) 

Also addresses  #774 with secondary commit.

Issue #356, opened 2017-02-01
Even with administrator privileges Hyper-V complains about insufficient privileges


thesteve0:
I cam across the error above when using Hyper-V. I have since fixed it so I can not show a screen shot of it not working. 

Basically it comes down to the fact that being administrator on the machine is not enough permission to use the Hyper-V commands from a console or non GUI interface (basically any automation). The fix is to add the current user to the Hyper-V group. 

Here is a command you can use to run it
https://blogs.msdn.microsoft.com/virtual_pc_guy/2016/05/30/adding-yourself-to-the-hyper-v-administrators-group-with-powershell/
https://blogs.msdn.microsoft.com/virtual_pc_guy/2010/09/28/creating-a-hyper-v-administrators-local-group-through-powershell/

But they can also go into the GUI and do this. Steps:
1) Windows key and then Computer Management
2) Local Users And Groups -> Groups
3) Double click on the Hyper-V admin group
4) Add their account to that group. 
The user then has to log out and log in before the changes will take affect. 

There also seems to be problems with a missing Hyper-V Admin group in windows 10 before the "Winter Update" patch (which basically re-installed windows). 

I am not sure if this affects other versions of windows and I only have windows 10 pro.

thesteve0:
@hferentschik I am not sure about labels but this is not minor. Without this I couldn't get Minishift to install AT ALL on windows. It was only after doing this that I could actually even get minishift to get past the stage where is first tries to make the VM.

budhrg:
@thesteve0 , I don't know but I always need to start PowerShell or CMD as "Run as Administrator".

I hope after gone through above instructions I needn't to run it that way.

hferentschik:
@thesteve0, can we clarify what we mean here? You want to execute minishift in non elevated mode? And your workaround allows you to do that? Or is your workaround needed to make even running Minishift in elevated mode possible?

Personally I always start Minishift from an elevated Powershell terminal, similar to @budhrg.  That has always been my understanding based on the docs for the Hyper-V driver - https://docs.docker.com/machine/drivers/hyper-v/

I also never worked with Hyper-V group so far. Not sure how much this is a bug in Minishift or how much we should and can "automate". AFAIU you need to run these commands in elevated mode, so you would have to run Minishift at least once in this mode, right?

thesteve0:
[...]

PR #779, opened 2017-04-19
Issue #590 Note to add creation of hyperv switch during start in Windows

### LINKED ISSUE #590: Allow selection of virtual switch for use with Hyper-V

budhrg:
Fix #590 for documentation part as actual implementation is part of #773 .

Need to be merged after #773

### DIFF
--- a/docs/source/getting-started/installing.adoc
+++ b/docs/source/getting-started/installing.adoc
@@ -45,6 +45,12 @@ https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/e
 https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/quick_start/walkthrough_virtual_switch[add a Virtual Switch]
 using the Hyper-V Manager. Make sure that you pair the virtual switch
 with a __network card (wired or wireless) that is connected to the network__.
+Minishift tries per default to use the first virtual switch it detects.
+This can lead to conflicts where this switch is used by another application.
+For example, in the case when https://docs.docker.com/docker-for-windows/install[Docker For Windows] is installed on the host.
+In this case the switch _DockerNAT_ is used by Docker for Windows and another switch needs to be created for Minishift.
+For the switch that is created for Minishift to work properly, you need to set the environment variable _HYPERV_VIRTUAL_SWITCH_ to the name of the switch.
+You can find more info in link:../using/managing-minishift{outfilesuffix}[Managing Minishift] page under _Driver specific environment variables_ section.
 
 - https://www.virtualbox.org/wiki/Downloads[VirtualBox]
 +


docs/source/getting-started/installing.adoc
```diff
@@ -45,6 +45,10 @@ https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/e
 https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/quick_start/walkthrough_virtual_switch[add a Virtual Switch]
 using the Hyper-V Manager. Make sure that you pair the virtual switch
 with a __network card (wired or wireless) that is connected to the network__.
+In case, you have any existing Virtual Switch like _DockerNAT_ which comes when installing https://docs.docker.com/docker-for-windows/install[docker for Windows],
```
hferentschik: Minishift tries per default to use the first virtual switch it detects. This can lead to conflicts where this switch is used by another application. This is for example the case when Docker For Windows is installed on the host. In this case the switch _DockerNAT_ is used by Docker for Windows and another switch needs to be created for Minishift. In this case you need to set the environment variable _HYPERV_VIRTUAL_SWITCH_ to the name of this switch.

Something like the above.  Note, it is "Docker for Windows" and "virtual switch". Watch capitalisation.

docs/source/getting-started/installing.adoc
```diff
@@ -45,6 +45,10 @@ https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/e
...
+then you need to add the name of newly created virtual switch into environment variable
+`HYPERV_VIRTUAL_SWITCH` as `$env:HYPERV_VIRTUAL_SWITCH=name_of_switch`.
+You can find more info in link:../using/managing-minishift{outfilesuffix}[Managing Minishift] page under _Driver specific environment variables_ section.
```
hferentschik: the link should go directly to the appropriate section. I guess that won't work, until the other pull request is merged.

budhrg: Yes @hferentschik , this PR should be merged after yours. Written in the description of the PR too

docs/source/getting-started/installing.adoc
budhrg: Updated the PR with your suggestion and more details

docs/source/getting-started/installing.adoc
```diff
@@ -45,6 +45,12 @@ https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/e
 https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/quick_start/walkthrough_virtual_switch[add a Virtual Switch]
 using the Hyper-V Manager. Make sure that you pair the virtual switch
 with a __network card (wired or wireless) that is connected to the network__.
+Minishift tries per default to use the first virtual switch it detects.
+This can lead to conflicts where this switch is used by another application.
```
LalatenduMohanty: Does this mean one switch is used for one VM? AFAIK Minishift would work fine as long as the virtual switch is an external type switch.

docs/source/getting-started/installing.adoc
```diff
@@ -45,6 +45,12 @@ https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/e
...
+For example, in the case when https://docs.docker.com/docker-for-windows/install[Docker For Windows] is installed on the host.
+In this case the switch _DockerNAT_ is used by Docker for Windows and another switch needs to be created for Minishift.
+For the switch that is created for Minishift to work properly, you need to set the environment variable _HYPERV_VIRTUAL_SWITCH_ to the name of the switch.
```
LalatenduMohanty: I think we are not communicating properly here. In case of docker NAT switch we need another switch as Minishift needs an external switch.

hferentschik: What's your suggestion for improvement?

Preeticp: Minishift by default, tries to use the first virtual switch it detects. This can lead to conflicts when this switch is used by another application. For example, if Docker For Windows is installed on the host, then the switch DockerNAT is used by Docker for Windows and another switch needs to be created for Minishift. To enable the switch created for Minishift to work properly, you need to set the environment variable HYPERV_VIRTUAL_SWITCH to the name of the switch. For more information see the environment variables section.(link to managing minishift environment variable section)

'set the environment variable HYPERV_VIRTUAL_SWITCH to the name of the switch' this part is not very clear, how about giving an example here.

Also, I suggest we create a separate section for Hyperv in Drivers doc. I am working on #356 and can add this information along with it to the drivers doc. I think this note is too long for this section, we can add a short note here and link it to the driver's doc WDYT?

LalatenduMohanty: IMO we do not need the docker NAT example here. We should just write that `If you have multiple virtual switches , please set the environment variable  HYPERV_VIRTUAL_SWITCH to the name of right external type switch for Minishift`.

LalatenduMohanty: May be we should point users to a proper link for how to set environment variable in Windows.

hferentschik: come on

LalatenduMohanty: @Preeticp IMO this part of documentation should not be part of drivers doc but it should part of installation doc.

hferentschik: this will as always depend on the type of shell you are using

hferentschik: @LalatenduMohanty, ok let's go for a shorter note and @Preeticp can add more in a dedicated section somewhere. That we somewhat the initial plan anyways

hferentschik:
Merged.

thomasmoree:
Hi, I'm Thomas, an [Nursing Dissertation Help](https://mydissertations.co.uk/nursing-dissertation-writing.php) professional committed to your academic success. Join hands with me, via My Dissertations, for exceptional assignments tailored to your needs.

Issue #691, opened 2017-04-06
Clean up documentation set


thatdocslady:
Now that we have the docs published (YAY!) there are a few things that need to be handled:

- Global pass on internal relative links to make sure they work and fix where needed
- Check for old GitHub links and replace them with openshift.org links or relative links as needed
- Sanity check on formatting and asciidoc behavior (i.e. making sure code blocks, tables, etc look ok)
- Verify that each index.adoc topic contains the correct links in the correct order (found already an instance where it's outdated)
- Verify that the topic map reflects the updated doc set
- AOB

There are a few issues that need to be resolved before this is done, i.e. #664 and #655, so putting needs-investigation on this to prevent confusion (hopefully).

thatdocslady:
TODO: Adjust the navigation so that the getting started section goes Installation > Quickstart > Docker Machine drivers

thatdocslady:
TODO: make sure "docker" is lowercase

hferentschik:
Merged via pull request #1048

PR #790, opened 2017-04-20
Issue #734 refactor Managing Minishift topic

### LINKED ISSUE #734: Refactor "Managing Minishift" topic
Following discussiong in #664, we should consider how to restructure the Managing Minishift topic to have easier flow and more accurate purpose.

My recommendations:

1. Delete or move the minishift start/stop/delete sections. TBH they basically repeat what's in the command reference, and the quickstart also has the start command in more details. Alternatively we can move those sections to the quickstart section itself, above the example procedure, and label them as "common commands". It makes more sense to have them there IMO.

2. Assuming that we are extracting the host folders and add-ons content into their own topics, we're basically left with a bunch of configuration sections, so maybe we can rename the topic to "Configuring Minishift", and then it's more accurately reflective of the content.

3. If we proceed with no.2 here, would it make sense to keep the proxy information here? Or should we extract it to some separate topic and call it "Networking" or something? Are we planning to document other aspects of networking in the future? Could be more useful for the modular docs approach.

Original comment in https://github.com/minishift/minishift/issues/664#issuecomment-291885952

thatdocslady:
Fixes #734 

This PR addresses several needs for the Managing Minishift topic:

1. Move the basic commands (start/stop/delete/ip) to the Quickstart topic
2. Extract proxies info to a separate topic
3. Minor proofing

NOTE: there's some conflict in the file when I tried to rebase (last rebase was yesterday), I don't know who touched the file but I can't rebase gracefully, would appreciate help on this.

### DIFF
--- a/docs/source/_topic_map.yml
+++ b/docs/source/_topic_map.yml
@@ -10,7 +10,7 @@ Topics:
     Topics:
       - Name: Overview
         File: index
-      - Name: Installing Docker Machine drivers
+      - Name: Installing Docker Machine Drivers
         File: docker-machine-drivers
       - Name: Installing Minishift
         File: installing
@@ -23,17 +23,19 @@ Topics:
         File: index
       - Name: Managing Minishift
         File: managing-minishift
+      - Name: HTTP/HTTPS Proxies
+        File: proxies
       - Name: Add-ons
         File: addons
       - Name: Host Folders
         File: mounting-host-folders
       - Name: Interacting with OpenShift
         File: interacting-with-openshift
-      - Name: Minishift Docker daemon
+      - Name: Minishift Docker Daemon
         File: docker-daemon
       - Name: Troubleshooting Minishift
         File: troubleshooting
-      - Name: Accessing the OpenShift docker registry
+      - Name: Accessing the OpenShift Docker Registry
         File: accessing-the-openshift-docker-registry
   - Name: Developing Minishift
     Dir: developing
@@ -44,7 +46,7 @@ Topics:
         File: developing
       - Name: Minishift CI
         File: ci
-      - Name: Writing documentation
+      - Name: Writing Documentation
         File: writing-docs
       - Name: Releasing Minishift
         File: releasing
--- a/docs/source/layouts/layout.erb
+++ b/docs/source/layouts/layout.erb
@@ -96,15 +96,17 @@
 
                   <li><a class="" href="/using/managing-minishift.html">Managing Minishift</a></li>
 
-                  <li><a class="" href="/using/addons.html">Minishift Add-ons</a></li>
+                  <li><a class="" href="/using/proxies.html">HTTP/HTTPS Proxies</a></li>
+
+                  <li><a class="" href="/using/addons.html">Add-ons</a></li>
 
                   <li><a class="" href="/using/mounting-host-folders.html">Host Folders</a></li>
 
                   <li><a class="" href="/using/interacting-with-openshift.html">Interacting with OpenShift</a></li>
 
-                  <li><a class="" href="/using/docker-daemon.html">Minishift Docker daemon</a></li>
+                  <li><a class="" href="/using/docker-daemon.html">Minishift Docker Daemon</a></li>
 
-                  <li><a class="" href="/using/accessing-the-openshift-docker-registry.html">Accessing the OpenShift Registry</a></li>
+                  <li><a class="" href="/using/accessing-the-openshift-docker-registry.html">Accessing the OpenShift Docker Registry</a></li>
 
                   <li><a class="" href="/using/troubleshooting.html">Troubleshooting Minishift</a></li>
               </ul>
--- a/docs/source/using/index.adoc
+++ b/docs/source/using/index.adoc
@@ -6,6 +6,7 @@ This section describes how to manage your Minishift VM and interact with your
 OpenShift cluster. It also helps troubleshooting common issues.
 
 - link:../using/managing-minishift{outfilesuffix}[Managing Minishift]
+- link:../using/proxies{outfilesuffix}[HTTP/HTTPS Proxies]
 - link:../using/addons{outfilesuffix}[Minishift Add-ons]
 - link:../using/mounting-host-folders{outfilesuffix}[Host Folders]
 - link:../using/interacting-with-openshift{outfilesuffix}[Interacting with OpenShift]
--- a/docs/source/using/managing-minishift.adoc
+++ b/docs/source/using/managing-minishift.adoc
@@ -19,40 +19,6 @@ The following sections contain information about managing the Minishift VM.
 For details about using Minishift to manage your local OpenShift cluster,
 see the link:../using/interacting-with-openshift{outfilesuffix}[Interacting with OpenShift] section.
 
-[[minishift-life-cycle]]
-== Minishift life-cycle
-
-[[minishift-start-intro]]
-=== Minishift start
-
-The link:../command-ref/minishift_start{outfilesuffix}[`minishift start`] command creates and
-configures the Minishift VM and provisions a local, single-node
-OpenShift instance within the VM.
-
-The command also copies the `oc` binary to your host so that you can interact
-with OpenShift through the `oc` command line tool or through the Web console,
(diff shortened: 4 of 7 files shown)


hferentschik:
> there's some conflict in the file when I tried to rebase (last rebase was yesterday), I don't know who touched the file but I can't rebase gracefully, would appreciate help on this.

    $ git checkout master
    $ git pull
    $ git checkout issue-734
    $ git rebase master
    # this will halt at some stage telling you that there are merge conflicts
    # resolve them
    $ git add .
    $ git rebase --continue
    # build and inspect site, especially around the sections which had merge conflicts

docs/source/layouts/layout.erb
```diff
@@ -96,15 +96,17 @@
 
                   <li><a class="" href="/using/managing-minishift.html">Managing Minishift</a></li>
 
-                  <li><a class="" href="/using/addons.html">Minishift Add-ons</a></li>
+                  <li><a class="" href="/using/addons.html">HTTP/HTTPS proxies</a></li>
```
hferentschik: This cannot be right. The link is to addons, but the title HTTP/HTTPS?

docs/source/using/index.adoc
```diff
@@ -6,6 +6,7 @@ This section describes how to manage your Minishift VM and interact with your
 OpenShift cluster. It also helps troubleshooting common issues.
 
 - link:../using/managing-minishift{outfilesuffix}[Managing Minishift]
+- link:../using/proxies{outfilesuffix}[HTTP/HTTPS Proxies]
```
hferentschik: I am not sure whether I am in favour to break all sections out. I think having the basic life cycles methods and things like HTTP proxy in one file makes sense. we start to loose any sort of grouping here and just add any odd topic to the highest level.

docs/source/using/managing-minishift.adoc
```diff
@@ -19,37 +19,6 @@ The following sections contain information about managing the Minishift VM.
 For details about using Minishift to manage your local OpenShift cluster,
 see the link:../using/interacting-with-openshift{outfilesuffix}[Interacting with OpenShift] section.
 
-[[minishift-start-intro]]
```
hferentschik: -1 for moving this to quick start, the basic life cycle methods and runtimes options are closely related. I am not sure what we try to achieve here?

thatdocslady:
@hferentschik thanks for the commands, I tried running them and I see that someone added a "minishift lifecycle" heading without any content. Any idea what this is about?

docs/source/variables.adoc
```diff
@@ -1,2 +1,2 @@
-:minishift-version: unset
-:openshift-version: unset
+:minishift-version: 1.0.0-rc.1
```
hferentschik: -1, variables.adoc should not even be checked in. If it is, then it should be removed. Might have been accidentally added at some stage. 

It is in .gitignore, so I don't know how it ended up there.

docs/source/layouts/layout.erb
thatdocslady: @hferentschik oops sorry, copy-paste error. That's weird, my local staging didn't show this as a problem...

docs/source/_topic_map.yml
```diff
@@ -5,7 +5,7 @@ Distros: openshift-*
 Topics:
   - Name: Overview
     File: index
-  - Name: Getting Started
+  - Name: Getting started
```
hferentschik: I am confused. Why are we using a different standard now then the rest of docs.openshift.org. I thought we are aligning with them. AFAICT they are using capitalisation.

hferentschik:
>  tried running them and I see that someone added a "minishift lifecycle" heading without any content. Any idea what this is about?

It is the heading under which start/stop/etc fall.

thatdocslady:
@hferentschik ok so I will remove it, because a part of the task was to move the commands away from this topic anyway.

docs/source/using/managing-minishift.adoc
```diff
@@ -104,7 +73,7 @@ places all runtime state into `~/.minishift`. This environment variable is
 currently experimental and semantics might change in future releases.
 
 [[persistent-configuration]]
-=== Persistent configuration
+== Persistent configuration
```
hferentschik: persistent options are part of runtime options. the runtime section starts with the four ways to control the runtime behaviour. Then we have three sub headings, one for each of these ways (the 4th is just default value). So I don't understand why this section should move up one level.

thatdocslady: I was trying to avoid deeper nesting and this way the persistent options and volumes show up in the topic TOC. If you don't think this is useful I can move them back down one level.

hferentschik: Or one just increases the TOC level?

hferentschik:
> ok so I will remove it, because a part of the task was to move the commands away from this topic anyway.

I have not idea where this would be coming from? I mean removing the commands. For now I think they are fine where they are now.

docs/source/variables.adoc
thatdocslady: I didn't touch the variables.adoc file so maybe this is something from the rebase merge? not sure how to fix this.

docs/source/_topic_map.yml
thatdocslady: it's not consistent there either but you're right, now I see that for the most part they capitalize so I'll revert

docs/source/using/index.adoc
LalatenduMohanty: I had initially suggested proxy as a separate doc, but now looking at the changes now, it  does not look good. I think we should move back the proxy piece back to the original doc.

LalatenduMohanty: I had initially suggested proxy as a separate doc, but now looking at the changes, it  does not look good. I think we should move back the proxy piece back to the original doc.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: +1 to move persistent options  below runtime options.

docs/source/variables.adoc
hferentschik: Sure, as I said, it most likely got added accidentally prior it being part of .gitignore. I did not notice until now. You basically did a build, the files got updated and you committed them. 

What I am saying is, that the file should be removed from the repo now.

docs/source/using/index.adoc
hferentschik: "Managing Minishift" has currently the theme of the basic life cycle methods start/stop/delete plus the sections which are relevant to this major life cycles. Proxies an runtime options are really directly related to this are relevant in particular for start. Really add-ons fitted quite well there as well, but might have been a big too long. However, the current content seems to me coherent as is. 

I don't get why it would move under quickstart. A quickstart should be a short demo of the happy path for the Minshift usage with pointers into more detailed sections. Moving  start/stop/delete into quickstart adds ways to much detailed doc into it. And what's the structure here? "Basic Minshift commands" (start/stop/delete) and then "Starting Minishift", How does this fit together. I don't expect the quickstart to start with such a detailed documentation of start/stop/delete. And "Networking" gets completely misplaced in this case.

hferentschik:
I suggest we close this pull request. I feel this is not moving the docs into the right direction. Neither in terms of the "Managing Minishift" content (or what would be left of it), nor in terms of what becomes of the Quickstart. 

Let's close this pull request and imo there is right now nothing to be done there.

thatdocslady:
@hferentschik @LalatenduMohanty ok I understand that the initial instructions in the issue changed so yes we can close this PR, and I can handle general proofreading for the topic as a part of #691

docs/source/using/index.adoc
LalatenduMohanty: > I don't get why it would move under quickstart.

@hferentschik Sorry I do not understand your comment. Not sure if you are commenting my comment. I suggested httpproxy stuff should be moved back to its original doc i.e. to managing minishift doc.

LalatenduMohanty: @hferentschik ok got it. You comment is only for minishift/start/stop/delete commands. Never mind.

LalatenduMohanty:
@thatdocslady closing it.

Issue #548, opened 2017-03-14
Expand basic.feature to cover elementary usage of Minishift


agajdosi:
Currently basic.feature does not cover much of Minishift use and needs an expansion. After short call on last Friday and further discussion with @amitkrout on this topic we propose following for basic.feature:

- As we are at the start with integration tests, we should focus on covering developers' use first, meaning we should prioritize time over case coverage.
- It would be good to cover elementary usage of Minishift in basic.feature, ideally following simple user story,
- which can be further expanded to cover more cases if those are not time consuming (no extra Minishift start, no extra creations of OpenShift applications).
- This will make basic.feature a relatively fast test for frequent use of developers, which will assure them that Minishift is passing smoke test and at least basic sanity on given platform.

Story we have in mind is: one wants to try OpenShift, for this case she choose Minishift, she should be able to start Minishift, she should be able to do ssh into machine, be able to deploy example application on OpenShift via oc command, able to stop and delete Minishift. These steps are time consuming but are needed. Further we can add less consuming cases to cover more. As result of this we are currently proposing following new scenarios aka tests:

- check that OpenShift console is accessible, that OpenShift returns at /healthz and /healthz/ready "ok",
- check that oc is on its place,
- make sure an example application can be deployed on OpenShift - this is the only new feature which is really time expensive, could take from 3 to 5 mins (will be covered in #750)
- delete application is fine,
- check that `minishift config set / get / view / unset` works - we would not be able to test real functionality of these changes as this would cost one extra start, but we can at least check that config is w/r-able, that commands works well,
- `minishift openshift config set / view` - probably no need to test `minishift openshift restart` as this functionality is part of `set`,
- check that external network is accessible from both host and guest,
- would be nice to check that `minishift status` is actually working.

WDYT? Is there anything which should be added?

hferentschik:
> As we are at the start with integration tests, we should focus on covering developers' use first, meaning we should prioritize time over case coverage.

+1

> check that OpenShift console is accessible, that OpenShift returns at /healthz and /healthz/ready "ok",

Here we probably want another common helper which allows us to probe an URL and expect a given HTTP status and potentially response content

> check that oc is on its place

+1

> make sure an example application can be deployed on OpenShift - this is the only new feature which is really time expensive, could take from 3 to 5 mins

+1, we need to decide on how we want to handle determining on when the app is ready

[...]

PR #791, opened 2017-04-20
Issue #548 Expanding tests to cover simple usage story

agajdosi:
Added test of OpenShift's http endpoints and web console
Added test of config command and its subcommands: set, get, unset, view
Added test of console and ip commands

Addresses issue #548

### DIFF
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -19,9 +19,14 @@ limitations under the License.
 package integration
 
 import (
+	"crypto/tls"
+	"encoding/json"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"net"
+	"net/http"
+	"net/url"
 	"os"
 	"strconv"
 	"strings"
@@ -36,7 +41,7 @@ import (
 )
 
 var lastCommandOutput CommandOutput
-var givenArgs, givenPath string
+var givenArgs, givenPath, testDir string
 
 type CommandOutput struct {
 	Command  string
@@ -124,13 +129,60 @@ func (m *Minishift) executingCommand(command string) error {
 	return nil
 }
 
-func (m *Minishift) executingMinishiftCommandSucceeds(command string) error {
+func (m *Minishift) executingMinishiftCommandSucceeds(command, expectedResult string) error {
 	err := m.executingCommand(command)
 	if err != nil {
 		return err
 	}
-	if lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr) != 0 {
-		return fmt.Errorf("Command did not execute successfully. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
+	success := (lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr) != 0)
+	switch expectedResult {
+	case "succeeds":
+		if success {
+			return fmt.Errorf("Command did not execute successfully. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
+		}
+	case "fails":
+		if success {
+			return fmt.Errorf("Command executed successfully, however was expected to fail. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
+		}
+	default:
+		return fmt.Errorf("Expected result: %s not recognized, please use: 'fails' or 'succeeds'", expectedResult)
+	}
+	return nil
+}
+
+func (m *Minishift) commandReturnEquals(commandField, command, expected string) error {
+	m.executingCommand(command)
+	return compareExpectedWithActualEquals(expected+"\n", selectFieldFromLastOutput(commandField))
+}
+
+func (m *Minishift) commandReturnContains(commandField, command, expected string) error {
+	m.executingCommand(command)
+	return compareExpectedWithActualContains(expected, selectFieldFromLastOutput(commandField))
+}
+
+//  To get values of nested keys, use following dot formating in Scenarios: key.nestedKey
+//  If an array is expected, use following formating: [value1 value2 value3].
+func (m *Minishift) configContains(configPath, expectingResult, expectedKeyPath, expectedValue string) error {
+	data, err := ioutil.ReadFile(testDir + "/" + configPath)
+	if err != nil {
+		return fmt.Errorf("Cannot read config file")
+	}
+	var values map[string]interface{}
+	json.Unmarshal(data, &values)
+	actualValue := ""
+	keyPath := strings.Split(expectedKeyPath, ".")
+	for _, element := range keyPath {
+		switch value := values[element].(type) {
+		case map[string]interface{}:
+			values = value
+		case []interface{}, nil, string, float64:
+			actualValue = fmt.Sprintf("%v", value)
+		default:
+			return fmt.Errorf("Unexpected type in JSON config, not supported by testsuite")
+		}
+	}
+	if (expectingResult == "contains") && (actualValue != expectedValue) {
+		return fmt.Errorf("For key '%s' config contains unexpected value '%s'", expectedKeyPath, actualValue)
 	}
 
 	return nil
@@ -189,6 +241,70 @@ func (m *Minishift) commandReturnShouldBeEmpty(commandField string) error {
 	return compareExpectedWithActualEquals("", selectFieldFromLastOutput(commandField))
 }
 
+func (m *Minishift) shouldBeInValidFormat(commandField, format string) error {
+	result := selectFieldFromLastOutput(commandField)
+	result = strings.TrimRight(result, "\n")
+	switch format {
+	case "URL":
+		_, err := url.ParseRequestURI(result)
+		if err != nil {
+			return fmt.Errorf("Command did not returned URL in valid format: %s", result)
+		}
+	case "IP":
+		if net.ParseIP(result) == nil {
+			return fmt.Errorf("%s of previous command is not a valid IP address: %s", commandField, result)
+		}
+	default:
+		return fmt.Errorf("Format %s not implemented.", format)
+	}
+	return nil
+}
+
+func (m *Minishift) getHTTPResponse(partOfResponse, url, urlSuffix, assertion, expected string) error {
+	switch url {
+	case "OpenShift":
+		url = m.getOpenShiftUrl() + urlSuffix
+	}
+	transport := &http.Transport{
+		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
+	}
+	client := &http.Client{Transport: transport}
+	response, err := client.Get(url)
+	if err != nil {
+		return fmt.Errorf("Server returned error on url: %s", url)
+	}
+	defer response.Body.Close()
+	var result string
+	switch partOfResponse {
+	case "body":
+		html, _ := ioutil.ReadAll(response.Body)
+		result = string(html[:])
+	case "status code":
+		result = fmt.Sprintf("%d", response.StatusCode)
+	default:
+		return fmt.Errorf("%s not implemented", partOfResponse)
+	}
+
+	switch assertion {
+	case "contains":
+		if !strings.Contains(result, expected) {
+			return fmt.Errorf("%s of reponse from %s does not contain expected string. Expected: %s, Actual: %s", partOfResponse, url, expected, result)
+		}
+	case "is equal to":
+		if result != expected {
+			return fmt.Errorf("%s of response from %s is not equal to expected string. Expected: %s, Actual: %s", partOfResponse, url, expected, result)
+		}
+	default:
+		return fmt.Errorf("Assertion type: %s is not implemented", assertion)
+	}
+	return nil
+}
+
+func (m *Minishift) getOpenShiftUrl() string {
+	cmdOut, _, _ := m.runner.RunCommand("console --url")
+	return strings.TrimRight(cmdOut, "\n")
+}
+
 func FeatureContext(s *godog.Suite) {
 	runner := util.MinishiftRunner{
 		CommandArgs: givenArgs,
@@ -199,8 +315,10 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`Minishift (?:has|should have) state "([^"]*)"`, m.shouldHaveState)
 	s.Step(`Minishift should have a valid IP address`, m.shouldHaveAValidIPAddress)
 	s.Step(`executing "minishift ([^"]*)"`, m.executingCommand)
-	s.Step(`executing "minishift ([^"]*)" succeeds$`, m.executingMinishiftCommandSucceeds)
+	s.Step(`executing "minishift ([^"]*)" (.*)$`, m.executingMinishiftCommandSucceeds)
 	s.Step(`executing "oc ([^"]*)" retrying (\d+) times with wait period of (\d+) seconds$`, m.executingRetryingTimesWithWaitPeriodOfSeconds)
+	s.Step(`([^"]*) of command "minishift ([^"]*)" is equal to "([^"]*)"`, m.commandReturnEquals)
+	s.Step(`([^"]*) of command "minishift ([^"]*)" contains "([^"]*)"`, m.commandReturnContains)
 	s.Step(`executing "oc ([^"]*)`, m.executingOcCommand)
 	s.Step(`executing "oc ([^"]*)" succeeds$`, m.executingOcCommandSucceeds)
 	s.Step(`([^"]*) should contain ([^"]*)`, m.commandReturnShouldContain)
@@ -208,9 +326,13 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`([^"]*) should equal ([^"]*)`, m.commandReturnShouldEqual)
 	s.Step(`([^"]*) should equal`, m.commandReturnShouldEqualContent)
 	s.Step(`([^"]*) should be empty`, m.commandReturnShouldBeEmpty)
+	s.Step(`([^"]*) should be valid ([^"]*)`, m.shouldBeInValidFormat)
+	s.Step(`(body|status code) of HTTP request to "([^"]*)" (?:|at "([^"]*)" )(contains|is equal to) "([^"]*)"`, m.getHTTPResponse)
+	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value "(.*)"`, m.configContains)
+	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)"(.*)`, m.configContains)
 
 	s.BeforeSuite(func() {
-		testDir := setUp()
+		testDir = setUp()
 		fmt.Println("Running Integration test in:", testDir)
 		fmt.Println("using binary:", givenPath)
 	})
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -10,16 +10,63 @@ Feature: Basic
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
 
+  Scenario: OpenShift is ready after startup
+    After startup of Minishift OpenShift instance should respond correctly on its html endpoints
+    and OpenShift web console should be accessible.
+    Given Minishift has state "Running"
+     Then status code of HTTP request to "OpenShift" at "/healthz" is equal to "200"
+      And body of HTTP request to "OpenShift" at "/healthz" contains "ok"
+      And status code of HTTP request to "OpenShift" at "/healthz/ready" is equal to "200"
+      And body of HTTP request to "OpenShift" at "/healthz/ready" contains "ok"
+      And status code of HTTP request to "OpenShift" at "/console" is equal to "200"
+      And body of HTTP request to "OpenShift" at "/console" contains "<title>OpenShift Web Console</title>"
+
+  Scenario Outline: User can set, get, view and unset values in configuration file
+    User is able to setup persistent configuration of Minishift using "config" command
+    and its subcommands, changing values stored in "config/config.json".
+    Given Minishift has state "Running"
+     When executing "minishift config set <property> <value>" succeeds
+     Then JSON config file "config/config.json" contains key "<property>" with value "<value>"
+      And stdout of command "minishift config get <property>" is equal to "<value>"
+      And stdout of command "minishift config view --format {{.ConfigKey}}:{{.ConfigValue}}" contains "<property>:<value>"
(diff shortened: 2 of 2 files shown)


LalatenduMohanty:
retest this please

hferentschik:
Add to whitelist

test/integration/features/basic.feature
```diff
@@ -10,6 +10,61 @@ Feature: Basic
...
+      export "FOO=BAR"
+      export "BAZ=BAT"
+      """
+
+  Scenario Outline: User can set, get, view and unset values in configuration file
+    User is able to setup persistent configuration of Minishift using "config" command
+    and its subcommands.
+    Given Minishift has state "Running"
+     When executing "minishift config set <property> <value>"
+     Then JSON config file contains key "<property>" with value "<value>"
```
hferentschik: I suggest to make this more generic, something along the lines:

    Then config file 'config/config.json' contains <regexp>

"config file" is a trigger to tell us that we should look for a file in $MINISHIFT_HOME. We then explicitly specify the file we expect to be changed. opposed to saying JSON config. The latter is ambiguous, since we have more than one JSON configuration file. We really want to have a generic method to check that an actual change happened in a given (config) file. Similar reasoning for suggesting the use of 'regex'. It is more generic and can easily be used to match key value pairs as well.

hferentschik: BTW, I recommend you have a close look at Aruba - https://github.com/cucumber/aruba - which is a Ruby gem for testing CLI tools via Cucumber. I think they have some awesome steps there and my vision has always been to mimic a big part of this.

hferentschik: We need to find a balance here. To generic steps might be awkward to use, but to small scoped steps means one needs to write actual step implementations for each single thing one wants to test.

test/integration/features/basic.feature
```diff
@@ -10,6 +10,61 @@ Feature: Basic
...
+      And stdout of command "minishift config get <property>" is equal to "<value>"
+      And stdout of command "minishift config view --format {{.ConfigKey}}:{{.ConfigValue}}" contains "<property>:<value>"
+     When executing "minishift config unset <property>"
+     Then stdout of command "minishift config get <property>" is equal to "<nil>"
+      And JSON config file does not contain key "<property>"
+      
+  Examples: Config values to work with
```
hferentschik: +1, nice using a Scenario Outline here

hferentschik: For what it's worth, these config commands don't even need a running VM. Not sure whether it would make sense to somehow group these tests somewhere together.

test/integration/basic_test.go
```diff
@@ -177,15 +260,22 @@ func FeatureContext(s *godog.Suite) {
...
 	s.Step(`executing "oc ([^"]*)`, m.executingOcCommand)
 	s.Step(`([^"]*) should contain ([^"]*)`, m.commandReturnShouldContain)
 	s.Step(`([^"]*) should contain`, m.commandReturnShouldContainContent)
 	s.Step(`([^"]*) should equal ([^"]*)`, m.commandReturnShouldEqual)
 	s.Step(`([^"]*) should equal`, m.commandReturnShouldEqualContent)
 	s.Step(`([^"]*) should be empty`, m.commandReturnShouldBeEmpty)
+	s.Step(`response.?([^"]*)? at "([^"]*)" contains "([^"]*)"`, m.getUrlResponse)
+	s.Step(`JSON config file (contains|does not contain) key "(.*)" with value "(.*)"`, m.configContains)
+	s.Step(`JSON config file (contains|does not contain) key "(.*)"(.*)`, m.configContains)
+	s.Step(`stdout should be valid address`, m.shouldBeValidAddress)
```
hferentschik: "address" being IP (address)? address is very ambitious. What's wrong using the already existing "should contain" step, specifying a regular expression for an IP. If not, and you want to add a specific step, then I think it should be "(stdout|stderr) should be a valid IP address". I think this is clearer for a reader of the test. Also you might want to be able to select the source (stdout vs stderr).

test/integration/basic_test.go
```diff
@@ -177,15 +260,22 @@ func FeatureContext(s *godog.Suite) {
...
 	s.Step(`executing "oc ([^"]*)" retrying (\d+) times with wait period of (\d+) seconds$`, m.executingRetryingTimesWithWaitPeriodOfSeconds)
+	s.Step(`([^"]*) of command "minishift ([^"]*)" is equal to "([^"]*)"`, m.commandReturnEquals)
+	s.Step(`([^"]*) of command "minishift ([^"]*)" contains "([^"]*)"`, m.commandReturnContains)
 	s.Step(`executing "oc ([^"]*)`, m.executingOcCommand)
 	s.Step(`([^"]*) should contain ([^"]*)`, m.commandReturnShouldContain)
 	s.Step(`([^"]*) should contain`, m.commandReturnShouldContainContent)
 	s.Step(`([^"]*) should equal ([^"]*)`, m.commandReturnShouldEqual)
 	s.Step(`([^"]*) should equal`, m.commandReturnShouldEqualContent)
 	s.Step(`([^"]*) should be empty`, m.commandReturnShouldBeEmpty)
+	s.Step(`response.?([^"]*)? at "([^"]*)" contains "([^"]*)"`, m.getUrlResponse)
```
hferentschik: "response" is a bit vague.  I also find the second capture group a bit odd. What else is there expect "OpenShift". How is this going to be used moving forward for non OpenShift URLs?

hferentschik: Also one probably wants to distinguish between HTTP return code and HTTP response body.

test/integration/features/basic.feature
```diff
@@ -10,6 +10,61 @@ Feature: Basic
...
+    Given Minishift has state "Running"
+     When executing "minishift console --url"
+     Then stdout should be valid address
+      And "minishift ip" should return same ip
+
+  Scenario: OpenShift is ready after startup
+    After startup of Minishift OpenShift instance should respond correctly on its html endpoints
+    and OpenShift web console should be accessible.
+    Given Minishift has state "Running"
+     Then response of OpenShift at "/healthz" contains "ok"
```
hferentschik: "Then body of HTTP request to OpenShift at "/healthz" contains "ok"

or we could make it even more generic and introduce some variables which get interpolated. Then we can have a generic step like:

    "Then body of HTTP request "https://#{vm_ip}:8443/healthz" contains "ok", 

but also 

    "Then body of HTTP request to "https://myapp.#{routing-suffix}/foo" contains "Welcome to me app"

What I am suggesting here is to provide a syntax as well as a predefined set of known variables which can be used for this and maybe other commands. Variables which then in the step definition implementation get replaced with the actual dynamic values. Without this type of mechanism we will have a lot of problems writing tests around things which depend on for example the routing suffix or the IP.

test/integration/features/basic.feature
```diff
@@ -10,6 +10,61 @@ Feature: Basic
...
+    | memory    | 2222  |
+    | cpus      | 3     |
+
+  Scenario: Commands console and ip returns valid ip
+    User is able to get ip of Minishift VM with command "ip" and url adress of OpenShift instance
+    by running command "console" with flag "--url". Both commands should contain same IP.
+    Given Minishift has state "Running"
+     When executing "minishift console --url"
+     Then stdout should be valid address
+      And "minishift ip" should return same ip
```
hferentschik: This one is odd. What's "same IP"?

    And "minishift ip" should match IP of VM

or we use the variable idea I mention below:

    When executing "minishift ip"
    Then stdout should match #{vm-ip}

test/integration/features/basic.feature
agajdosi: @hferentschik Good idea, I felt a little bit lost here, your suggestion of `Then config file 'config/config.json' contains` seems good to me. With regex I am wondering if that is not going to be too technical, want to keep step definitions really human/manager/everybody readable as cucumber suggests. Will try how that looks and feels and will let you know.

test/integration/features/basic.feature
agajdosi: @hferentschik After we finish basic scenario, then I would like to focus more on various commands, meaning that config command should be tested in specialized scenario in different situations - when VM is not existing, is running, stopped. Also there is need to actually check that changes in config files affects the machine after start.

test/integration/features/basic.feature
agajdosi: @hferentschik Tried to make it at least a little bit generic, which caused some lines to sound odd. Suggestion for variables as #{var} is nice, I was also wondering about this, but I did not find a solution how to do that. Will take a look on that again.

test/integration/basic_test.go
agajdosi: @hferentschik thought this could be used for example with "of node-js application".

test/integration/basic_test.go
agajdosi: @hferentschik yes, it means valid IP address. Reason why I am not using should contain is that I was not able to implement a way how to introduce variables like you suggest: #{var}. If I will find solution how to add them, then I agree that should contain is better to use.

test/integration/features/basic.feature
hferentschik: > With regex I am wondering if that is not going to be too technical

But contains already uses a regexp under the hood right? Does "contain" not imply regexp. In many cases you probably just specify a specific word or phrase, but this could also be a regexp, right? Say you want to match a URL where some part is dynamic (aka the IP). How do you do that?

test/integration/features/basic.feature
hferentschik: >  then I would like to focus more on various commands, meaning that config command should be tested in specialized scenario in different situations

+1

> Also there is need to actually check that changes in config files affects the machine after start.

+1

test/integration/features/basic.feature
hferentschik: > I was also wondering about this, but I did not find a solution how to do that. Will take a look on that again.

Ok. The idea is to populate some sort of context map  when Minishift starts and when you process a URL as part of the step definition for HTTP, you would grep for anything #{...} and check whether there is a value for this in the context. Effectively this is what for example addons do with their interpolation context. 

Only drawback with this is, that is is a bit of magic and one needs to know that we support this. I guess one can mitigate this, but printing out the context during Minishfift start.

test/integration/basic_test.go
hferentschik: +1

test/integration/basic_test.go
hferentschik: ok

test/integration/features/basic.feature
agajdosi: @hferentschik I am fine with having regexs in basic_test.go file, but having it as part of feature file seems to me like moving code into feature file making it less understandable.

I agree that we probably need a step definition which reads a file and checks if it contains string or regexp, however this does not check that JSON syntax is fine and that the file is valid JSON. Currently I am using json.Unmarshall function from encoding/json package to check for key and value, which I think is more secure.

test/integration/features/basic.feature
agajdosi: @hferentschik So I did some research today - it would be ideal to have this replacement of ${variable} for actual value accessible in every step definition. My idea here is to use nested steps which were introduced 5 days ago in godog. For example like this:

Step: 
`Then body of http request on "${server_ip}/health" is "ok"`
Matching:
```
s.Step(`step:"minishift ip" should return same ip`, m.shouldReturnSameIp)
s.Step(`(.*)`, injectVar)
```
injectVar step definition:
```
func injectVar(stepName string) golang.Steps {
   [supplementing ${server_ip} for actual IP of server in stepName string]
   return golang.Steps{
     fmt.Sprintf(`step:%s`, stepName),
   }
}
```
Doing it like this every step is going to trigger injectVar step implementation, which will inject the variable value and then will call actual step implementation. I am having problems with fact that godog.Steps seems to be undefined whatever I try to do, but theoretically is this a way to go?

Second option is to implement the replacement of ${var} only specified steps.

Did not found any other option yet...

agajdosi: @hferentschik Found out this to be doable, we just need to update godog package from 0.6.2 to 0.7.1 in glide.yaml.

I had a [conversation](https://github.com/DATA-DOG/godog/issues/81) about this with l3pp4rd, contributor of godog:

> But it looks quite misleading to make such dynamic implementations in step level. I do not consider this as a good practice.

hferentschik: Interesting with nested steps. Need to better understand how this works. I am not quite sure how this would work in our case. Could this nested step approach not conflict in some cases?

> Second option is to implement the replacement of ${var} only specified steps.

Right, that was my thinking initially.

> Found out this to be doable, we just need to update godog package from 0.6.2 to 0.7.1 in glide.yaml.

An update is probably good regardless. There might be other fixes we get via this.

> I had a conversation about this with l3pp4rd, contributor of godog:
>> But it looks quite misleading to make such dynamic implementations in step level. I do not consider this as a good practice.

I agree, I also don't consider this good practice. As I mentioned before it introduces some sort of magic. The problem with say the HTTP step is, that it is not only the path which matters for us. There is also the app part which comes in front of the routing suffix, http://\<app\>.\<project\>.\<routing-suffix\>/...

Or take 'minishift ip' case. How do I handle this. I guess one can create dedicated steps for all these cases, a bit what you had:

     Then "minishift ip" should match the Minishift VM IP

So for all cases where we would need some dynamic part, we would need to have a dedicated step which under the hood do the checks. You still need to keep information like IP, routing suffix, etc in some sort of map so.

hferentschik: It is a bit of a catch 22 really, the dynamic variables is probably not best practice, but having all these explicit step definitions seems not really so nice either. What's about 'minishift console'? We could not use the execute step, since we would need a dedicated:

    'minishift console --url' returns console URL

hferentschik: > The steps should represent the things your users are exposed to - in other words the public interface of your application.

IP and routing suffix are public in our case. I guess our problems was not totally clear in https://github.com/DATA-DOG/godog/issues/81

test/integration/features/basic.feature
```diff
@@ -10,6 +10,42 @@ Feature: Basic
      Then Minishift should have state "Running"
       And Minishift should have a valid IP address
 
+  Scenario: OpenShift is ready after startup
+    After startup of Minishift OpenShift instance should respond correctly on its html endpoints
+    and OpenShift web console should be accessible.
+    Given Minishift has state "Running"
+     Then body of HTTP request to "OpenShift" at "/healthz" contains "ok"
```
hferentschik: Very nice, I like this a lot. I also like how this handles the dynamic part. So, if I get this right

    Then body of HTTP request to "OpenShift" at "/healthz" contains "ok"

Will check the actual OpenShift URL + whatever is specified under "at", but I could also do:

    Then body of HTTP request to "http://foo.bar.com" contains "ok"

Is this correct.

hferentschik: Technically though, I think we should be looking at the returned status code:

    Then response code of HTTP request to "OpenShift" at "/healthz" equals "200"

See also https://docs.openshift.org/latest/dev_guide/application_health.html. AFAIU there is no need to return anything in the body. It is the return code which matters. I recommend to add an appropriate  step and update the test.

test/integration/features/basic.feature
```diff
@@ -10,6 +10,42 @@ Feature: Basic
...
+      And body of HTTP request to "OpenShift" at "/console" contains "<title>OpenShift Web Console</title>"
+
+  Scenario Outline: User can set, get, view and unset values in configuration file
+    User is able to setup persistent configuration of Minishift using "config" command
+    and its subcommands, changing values stored in "config/config.json".
+    #To get values of nested keys, use following dot formating: key.nestedKey.
+    #If an array is expected, use following formating: [value1 value2 value3].
+    Given Minishift has state "Running"
+     When executing "minishift config set <property> <value>" succeeds
+     Then JSON config file "config/config.json" contains key "<property>" with value "<value>"
```
hferentschik: Awesome. Now one sees directly which config file, whereas the fact that is is relative to $MINISHIFT_HOME is hidden. Very nice test.

test/integration/basic_test.go
```diff
@@ -132,6 +137,41 @@ func (m *Minishift) executingMinishiftCommandSucceeds(command string) error {
 	if lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr) != 0 {
 		return fmt.Errorf("Command did not execute successfully. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
 	}
+	return nil
+}
+
+func (m *Minishift) commandReturnEquals(commandField, command, expected string) error {
```
hferentschik: +1

test/integration/basic_test.go
```diff
@@ -132,6 +137,41 @@ func (m *Minishift) executingMinishiftCommandSucceeds(command string) error {
...
+	m.executingCommand(command)
+	return compareExpectedWithActualEquals(expected+"\n", selectFieldFromLastOutput(commandField))
+}
+
+func (m *Minishift) commandReturnContains(commandField, command, expected string) error {
```
hferentschik: +1

test/integration/basic_test.go
```diff
@@ -132,6 +137,41 @@ func (m *Minishift) executingMinishiftCommandSucceeds(command string) error {
...
+func (m *Minishift) configContains(configPath, expectingResult, expectedKeyPath, expectedValue string) error {
+	data, err := ioutil.ReadFile(testDir + "/" + configPath)
+	if err != nil {
+		return fmt.Errorf("Cannot read config file")
+	}
+	var values map[string]interface{}
+	json.Unmarshal(data, &values)
+	actualValue := ""
+	keyPath := strings.Split(expectedKeyPath, ".")
+	for _, element := range keyPath {
```
hferentschik: So here you are taking height for nested properties, right? Something like 'addons.anyuid.Name', right? Does it work. For all I can tell you are not using anything nested in the tests yet. Just wondering. Looks ok on first sight.

test/integration/basic_test.go
```diff
@@ -132,6 +137,41 @@ func (m *Minishift) executingMinishiftCommandSucceeds(command string) error {
...
+		switch value := values[element].(type) {
+		case map[string]interface{}:
+			values = value
+		case []interface{}, nil, string, float64:
+			actualValue = fmt.Sprintf("%v", value)
```
hferentschik: Wau, that's nifty.

test/integration/features/basic.feature
```diff
@@ -10,6 +10,42 @@ Feature: Basic
...
+    User is able to get ip of Minishift VM with command "ip" and url adress of OpenShift instance
+    by running command "console" with flag "--url". Both commands should contain same IP.
+    Given Minishift has state "Running"
+     When executing "minishift console --url" succeeds
+     Then stdout should be valid URL
+      And "minishift ip" and "minishift console --url" should return same IP address
+
   Scenario: OpenShift developer account has sudo permissions
      The 'developer' user should be configured with the sudoer role after starting Minishift
      When executing "oc --as system:admin get clusterrolebindings" succeeds
```
hferentschik: I guess here we should laos add 

    Then executing "oc --as system:admin get clusterrolebindings" succeeds

Also, it would be nice to have the counter example

    Then executing "oc get clusterrolebindings" fails

test/integration/features/basic.feature
```diff
@@ -10,6 +10,42 @@ Feature: Basic
...
+    | memory    | 2222  |
+    | cpus      | 3     |
+
+  Scenario: Commands console and ip returns valid ip
+    User is able to get ip of Minishift VM with command "ip" and url adress of OpenShift instance
+    by running command "console" with flag "--url". Both commands should contain same IP.
+    Given Minishift has state "Running"
+     When executing "minishift console --url" succeeds
+     Then stdout should be valid URL
+      And "minishift ip" and "minishift console --url" should return same IP address
```
hferentschik: I find this misleading "return same IP address", 'minishift console --url' does not return an IP, it returns an URL. Also it seems arbitrary to compare these two in a step. Here is what I would suggest:

    When executing "minishift console --url" succeeds
    Then stdout should be valid URL
    And stdout should be the OpenShift Console URL 
    When executing  "minishift ip" succeeds 
    Then stdout should be valid IP
    And stdout should be the Minishift VM IP

Not using variables, I think that would be the most concise test. 

Also have you seen https://github.com/DATA-DOG/godog/issues/81#issuecomment-300853730

test/integration/basic_test.go
agajdosi: Yes, this is to be prepared for nested values. For now nested values are not used, but I tested this on OpenShift's config so it should be probably ready for scenarios for `minishift openshift config xxx`. Wanted to make this really reusable.

test/integration/features/basic.feature
agajdosi: @hferentschik Yes, that is correct, it accepts both dynamic and explicit addresses.

> Technically though, I think we should be looking at the returned status code

To be honest, I am not sure here. Spent some time on going through documentation and I have not found an information about these endpoints - does they return only "ok" response, or do they have more responses - eg. "not ready"?

However I am going to add check for response code into implementation, so we can later switch from one to second just by altering steps in the .feature file.

test/integration/features/basic.feature
agajdosi: @hferentschik Good catch, given steps are missing in some places, adding them.

test/integration/features/basic.feature
agajdosi: @hferentschik Looking on it now.

> You can create a placeholder value map. And whenever a scenario produces dynamic value, you add a placeholder for this value, like suite.placeholders["VM_IP"] = parsedIP.

This is golden. Also your suggestion for steps looks nice, totally agree it is better than `is same as...`. So right now I am heading towards:

```
When executing "minishift console --url" succeeds
Then stdout should be valid URL
And stdout should be the OPENSHIFT_CONSOLE_URL 
When executing  "minishift ip" succeeds 
Then stdout should be valid IP
And stdout should be the MINISHIFT_VM_IP
```

agajdosi: @hferentschik Steps suggested by you seems fine, I am just wondering how to get the IP of VM by not using `minishift ip` command. Idea was to extract it from `minishift console --url` and validate them. But it does not make much more sense. 

So maybe steps:

```
And stdout should be the Minishift VM IP
And stdout should be the OpenShift Console URL
```
could be skipped for now. And later when we have a way how to get these values from another souce, we can add them and also add DYNAMIC_VALUE substitution. WDYT?

hferentschik:
Merged, thanks. I added another commit which refactored the existing code. There was a misunderstanding I think around the use of the Minishift struct.

agajdosi:
@hferentschik Sure, thanks.

Issue #750, opened 2017-04-13
Add integration test to create and deploy of example application


agajdosi:
As discussed in #548 we need to test basic functionality and usage of OpenShift as part of quick smoke test which basic.feature is going to be. 

Simple plan for this issue is to make sure with integration tests that an example application can be deployed on OpenShift. However it will be good to further expand this plan by more scenarios - check that application can be deleted, etc.

amitkrout:
Considering this as a first call for s2i ruby-ex example application deployment, the basic scenario steps would be like 
```
1. oc login
2. Create a new namespace 
3. deploy the app
4. oc logs for the deployed app
5. Delete namespace
6. logout
```
However here the problem is with the readiness of the deployment environment. So sleep is added before each scenario steps to handle the readiness. IMO this is not the efficient way to handle this. There should be a better way to handle readiness.

For initial implementation considering sleep as a solution, then adding sleep time amount will be a challenge. For example : Just after ```minishift start```, running ```oc login``` command throws error and after sometime the same command ```oc login``` will work as expected. 

Now the question is how much sleep time we should put before each scenario step to make whole deployment ideal wrt sleep time. 

This scenario i have executed successfully.
```
sleep for 120sec
oc login
sleep for 10sec
Create a new namespace 
sleep for 10sec
deploy the app
sleep for 30sec
oc logs for the deployed app
sleep for 5sec
Delete namespace
sleep for 5sec
logout
```
Please share your thoughts

agajdosi:
@amitkrout 
> For example : Just after minishift start, running oc login command throws error and after sometime the same command oc login will work as expected.

Then we should add the OpenShift part right before "Stopping minishift" scenario, so previous scenarios will make some time for oc, but still it won't go under 100sec I guess. Other sleeps seem to be quite short.

> IMO this is not the efficient way to handle this. There should be a better way to handle readiness.

Would be nice to come up with some try-retry(-timeout) concept, but right now I am not sure how to design it well so we won't repeat ourself in each step definition. Only idea I have right now is to create a step definition with match `try for x seconds: "(.*)"`, which will then call a function tryRetry. This function will accept arg1=matched string, arg2=timeout. In this function we will then handle the work around matched string which will later call a normal function of step definition. But still we will repeat regex match on 2 places - in func FeatureContext(s *godog.Suite) and then in func tryRetry()...

amitkrout:
[...]

Issue #641, opened 2017-04-03
Add new add-on command to wait for build to complete


hferentschik:
We should add a new addon command which allows for a build to complete and a service be accessible. It is not enough to just tail the build, instead one need to determine when a pod switches into the running state or in case of readiness checks defined as part of the application until the readiness checks pass.

hferentschik:
See also https://github.com/openshift/jenkins-client-plugin

jorgemoralespou:
@hferentschik 
I would have a waitForService and a waitForDeployment

waitForService can wait until an endpoint is > 0 (there is an actual endpoint that can be used)
waitForDeployment waits for a pod of that particular deployment to be Running

hferentschik:
> waitForService and a waitForDeployment

sounds good. I like the names of the commands. Maybe we can unify a bit and have 'wait --forService' and 'wait --forDeployment'. If we then rename 'sleep' to 'wait', we have some sort of wait related group of commands. WDYT?

What would be the parameter for these new commands? The OpenShift service/deployment config names?

> waitForService can wait until an endpoint is > 0 

What do you mean exactly? How do you determine whether the endpoint "is > 0". Are you thinking of extracting  the readiness probe URL and using that? Or are you talking about a OpenShift API we should use. I kind of like the idea to extract and monitor the readiness probe where possible

> waitForDeployment waits for a pod of that particular deployment to be Running

+1

I think this is the most important new feature from the set of changes we discussed. This will open up new possibilities and enable doing proper deployments of applications. A lot of the other issues we discussed are more of cosmetic nature. I definitely want this for the next version.

LalatenduMohanty:
> I like the names of the commands. Maybe we can unify a bit and have 'wait --forService' and 'wait --forDeployment'. If we then rename 'sleep' to 'wait', we have some sort of wait related group of commands. WDYT?

+1 sounds good to me.

jorgemoralespou:
> waitForService can wait until an endpoint is > 0
When there is an endpoint you get a value for this expression. This can be queried to be different from null.

````
oc get endpoints/test -o jsonpath='{.subsets[*].addresses[*].ip}'
````

> waitForDeployment waits for a pod of that particular deployment to be Running

For this, you can monitor the pod status for a specific deploymentconfig 

Not working, but should be something similar.
````
oc get pods -o jsontemplate=".items[?(@.metadata.labels.deploymentconfig=='test')]].status.phase=='Running'"
````

Asking for help.

[...]

PR #796, opened 2017-04-21
Issue #750 Expand basic.feature to test creation and deployment of example app

amitkrout:
Addresses issue #750

Basic scenario steps added for ruby-ex

1. oc login
2. Create a new namespace 
3. deploy the app
4. oc logs for the deployed app
5. oc expose
5. Delete namespace
6. logout

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -84,6 +84,71 @@ Feature: Basic
       172.30.1.1:5000
       """
 
+  # User can deploy the example Ruby application ruby-ex
+  Scenario: User can login to the server
+    When executing "oc login --username=developer --password=developer"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Login successful
+     """
+  
+  Scenario: User can create new namespace ruby for application ruby-ex
+    When executing "oc new-project ruby"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Now using project "ruby"
+     """
+  
+  Scenario: User can deploy application ruby-ex to namespace ruby 
+    When executing "oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-ex.git"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Success
+     """
+  
+  Scenario: User will Wait for deployment endpoint status of application ruby-ex
+    When executing "oc rollout status deploymentconfig ruby-ex --watch"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     "ruby-ex-1" successfully rolled out
+     """
+  
+  Scenario: User can create route for ruby-ex to make it visiable outside of the cluster
+    When executing "oc expose svc/ruby-ex"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     exposed
+     """
+  
+  Scenario: User can delete namespace ruby
+    When executing "oc delete project ruby"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     "ruby" deleted
+     """
+  
+  Scenario: User can log out the session
+    When executing "oc logout"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Logged "developer" out
+     """
+  # End of Ruby application ruby-ex deployment
+  
   Scenario: Stopping Minishift
     Given Minishift has state "Running"
      When executing "minishift stop"


test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
       172.30.1.1:5000
       """
 
+  Scenario: Deploy ruby-ex s2i image
+    Given Minishift has state "Running"
+     When executing "oc login --username=developer --password=developer" retrying 24 times with wait period of 5 seconds
```
hferentschik: Retrying 24 times with a wait time of 5 seconds? Why would you do that? If Minishift is running, the login needs to work directly. Why do you want to retry here? Did you see any problems? 

Retry is for cases where there is an expectation that the command can fail. This is not the case here.

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     Then stderr should be empty
+      And exitcode should equal 0
+      And stdout should contain
+      """
+      Login successful
+      """
+    Then executing "oc new-project rubys2i" retrying 3 times with wait period of 5 seconds
```
hferentschik: Again, why the retry. The creation of the app should work always. If not, there is an issue. Effectively you are putting into code/test here, that it is ok for the 'oc new-project rubys2i' command to fail 2 times before it eventually will work. This is not the case

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Now using project
+     """
+    Then executing "oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-ex.git" retrying 3 times with wait period of 5 seconds
```
hferentschik: Same as above. No retry.

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+      And stdout should contain
+      """
+      Login successful
+      """
+    Then executing "oc new-project rubys2i" retrying 3 times with wait period of 5 seconds
+     And stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Now using project
```
hferentschik: You should match the project name. What if it still says "Now using project my-project". You are missing the most important part of the assertion.

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     """
+    Then executing "oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-ex.git" retrying 3 times with wait period of 5 seconds
+     And stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Success
+     """
+    Then executing "oc logs -f bc/ruby-ex" retrying 18 times with wait period of 5 seconds
```
hferentschik: -1, so here we have a point where we indeed have to do something. The problem is that _oc logs_ will return until the builder image is downloaded. So we used to indeed have to have some sort of loop here. However, there is a better way of doing this. What you want is to wait for the app to be deployed. There is a command for that _oc rollout status deploymentconfig NAME --watch_. See also issue #641

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Push successful
+     """
+    Then executing "oc expose svc/ruby-ex" retrying 12 times with wait period of 5 seconds
```
hferentschik: -1, no retry

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     exposed
+     """
+    Then executing "oc delete project rubys2i" retrying 3 times with wait period of 5 seconds
+     And stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     deleted
+     """
+    Then executing "oc logout" retrying 3 times with wait period of 5 seconds
```
hferentschik: -1

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     Push successful
+     """
+    Then executing "oc expose svc/ruby-ex" retrying 12 times with wait period of 5 seconds
+     And stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     exposed
+     """
+    Then executing "oc delete project rubys2i" retrying 3 times with wait period of 5 seconds
```
hferentschik: -1

hferentschik:
add to whitelist

test/integration/features/basic.feature
amitkrout: Agree, it should not be the case. This step executed successfully in the sequence without retry.

test/integration/features/basic.feature
amitkrout: This conception is inherited from 2.* downstream releases [https://github.com/projectatomic/adb-tests/blob/master/cdk-testsuite/utils/openshiftUtils.py](url). In 2.* after ```vagranrt up``` we had timing issue  with immediate ```oc login``` which was resolved as per the above implementation. I believe this is not the same case here. In the above step i removed the retry and it went fine without error/exception.

test/integration/features/basic.feature
amitkrout: Looking to this

amitkrout: This is a better solution of waiting for deployment endpoint status rather than retrying on ``` oc logs -f bc/ruby-ex``` multiple times. 
```
$ oc rollout status deploymentconfig ruby-ex --watch
Deployment config "ruby-ex" waiting on image update
Waiting for latest deployment config spec to be observed by the controller loop...
Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
Waiting for rollout to finish: 0 of 1 updated replicas are available...
Waiting for latest deployment config spec to be observed by the controller loop...
Waiting for latest deployment config spec to be observed by the controller loop...
replication controller "ruby-ex-1" successfully rolled out
```

test/integration/features/basic.feature
amitkrout: removed retry

test/integration/features/basic.feature
amitkrout: removed retry

test/integration/features/basic.feature
amitkrout: removed retry

test/integration/features/basic.feature
amitkrout: Added. Now the match tring would be ```Now using project "rubys2i"```

test/integration/features/basic.feature
amitkrout: retry removed

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+  Scenario: Deploy ruby-ex s2i image
+    Given Minishift has state "Running"
+     When executing "oc login --username=developer --password=developer"
+     Then stderr should be empty
+      And exitcode should equal 0
+      And stdout should contain
+      """
+      Login successful
+      """
+    Then executing "oc new-project rubys2i"
```
hferentschik: "When"

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And stderr should be empty
```
hferentschik: Then

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Now using project "rubys2i"
+     """
+    Then executing "oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-ex.git"
```
hferentschik: When

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And stderr should be empty
```
hferentschik: Then

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Success
+     """
+    Then executing "oc rollout status deploymentconfig ruby-ex --watch"
```
hferentschik: When

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And stderr should be empty
```
hferentschik: Then

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     "ruby-ex-1" successfully rolled out
+     """
+    Then executing "oc expose svc/ruby-ex"
```
hferentschik: When

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And stderr should be empty
```
hferentschik: Then

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     exposed
+     """
+    Then executing "oc delete project rubys2i"
```
hferentschik: When

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
...
+     And stderr should be empty
```
hferentschik: Then

test/integration/features/basic.feature
```diff
@@ -84,6 +84,58 @@ Feature: Basic
       172.30.1.1:5000
       """
 
+  Scenario: Deploy ruby-ex s2i image
```
hferentschik: You are not deploying "ruby-ex s2i image", you are deploying the ruby-ex application. s2i is the process of converting source to image. That's only a small part of what's going on.

It's more like: "A user can deploy the example Ruby application ruby-ex"

hferentschik: That said, I think we can spit this ling scenario up into several smaller ones:
* User can login to OpenShift
* User can create new project
* User can deploy the example Ruby application ruby-ex
* User can delete project

Having multiple scenarios will also help with the "When", "Then", "And" logic

hferentschik:
retest this please

test/integration/features/basic.feature
amitkrout: > You are not deploying "ruby-ex s2i image", you are deploying the ruby-ex application. s2i is the process of converting source to image. That's only a small part of what's going on.

I had a misconception on s2i. Thanks for the correction. For readability propose we can add comment line in the feature file as 
```
# A user can deploy the example Ruby application ruby-ex
    User can login to OpenShift
    User can create new project
    User can deploy the example Ruby application ruby-ex
    User can delete project
# End of Ruby application ruby-ex deployment
```

> 

That said, I think we can spit this ling scenario up into several smaller ones:

    User can login to OpenShift
    User can create new project
    User can deploy the example Ruby application ruby-ex
    User can delete project

Having multiple scenarios will also help with the "When", "Then", "And" logic

Multiple scenario also helps to avoid resource problem for more application deployment. For example considering the single big scenario if i am deploying 3 application sequentially and 1st application failed to deploy, may be there is a chance of failure due to less resource available to the 2nd app and so on. Because the failed app will not be removed as it skips the step. By spiting the the bigger scenario we can overcome to this problem

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
```diff
@@ -84,6 +84,71 @@ Feature: Basic
       172.30.1.1:5000
       """
 
+  # User can deploy the example Ruby application ruby-ex
+  Scenario: User can login to the server
```
hferentschik: 👍

test/integration/features/basic.feature
```diff
@@ -84,6 +84,71 @@ Feature: Basic
...
+    When executing "oc login --username=developer --password=developer"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Login successful
+     """
+  
+  Scenario: User can create new namespace ruby for application ruby-ex
```
hferentschik: :-)

test/integration/features/basic.feature
```diff
@@ -84,6 +84,71 @@ Feature: Basic
...
+    When executing "oc new-project ruby"
+    Then stderr should be empty
+     And exitcode should equal 0
+     And stdout should contain
+     """
+     Now using project "ruby"
+     """
+  
+  Scenario: User can deploy application ruby-ex to namespace ruby
```
hferentschik: :-)

hferentschik:
Merged, thanks.

PR #797, opened 2017-04-21
Issue #356 Insufficient administrator privileges to run Hyper-V

Preeticp:
Addresses issue #356

This PR adds a note for Hyper-V users to add themselves to the Hyper-V Administrator group.
Also, corrected a couple of minor typos in the Installation Guide.

### DIFF
--- a/docs/source/getting-started/installing.adoc
+++ b/docs/source/getting-started/installing.adoc
@@ -48,14 +48,14 @@ https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/quick_start/wa
 using the Hyper-V Manager.
 - Make sure that you pair the virtual switch
 with a __network card (wired or wireless) that is connected to the network__.
-- If you have multiple virtual switches , set the environment variable _HYPERV_VIRTUAL_SWITCH_ to the name of external virtual switch you want to use for Minishift. See also the section about setting
-link:../using/managing-minishift{outfilesuffix}#driver-specific-environment-variables[driver specific enviropnment variables].
+- If you have multiple virtual switches , set the environment variable `HYPERV_VIRTUAL_SWITCH` to the name of the external virtual switch you want to use for Minishift. For further information see
+link:../using/managing-minishift{outfilesuffix}#driver-specific-environment-variables[setting driver specific environment variables].
 ====
 
 - https://www.virtualbox.org/wiki/Downloads[VirtualBox]
 +
 NOTE: It is recommended to use `Virtualbox 5.1.12` or later on Windows to avoid the issue
-link:../using/troubleshooting{outfilesuffix}#machine-doesnt-exist[Error: getting state for host: machine does not exist].
+link:../using/troubleshooting{outfilesuffix}#machine-doesnt-exist[Error: machine does not exist].
 
 If you encounter issues related to the hypervisor, see
 the link:../using/troubleshooting{outfilesuffix}[Troubleshooting] section.
--- a/docs/source/using/troubleshooting.adoc
+++ b/docs/source/using/troubleshooting.adoc
@@ -30,7 +30,6 @@ If you use `virsh` on KVM/libvirt to create snapshots in your development
 workflow, and then use `minishift delete` to delete the snapshots along with
 the VM, you might encounter the following error:
 
-[source,sh]
 ----
 $ minishift delete
 Deleting the Minishift VM...
@@ -45,7 +44,6 @@ Workaround: To delete the snapshots you need to perform the following steps.
 .  Delete the definitions.
 +
 
-[source,sh]
 ----
 $ sudo virsh snapshot-delete --metadata minishift <snapshot-name>
 ----
@@ -53,7 +51,6 @@ $ sudo virsh snapshot-delete --metadata minishift <snapshot-name>
 .  Undefine the Minishift domain.
 +
 
-[source,sh]
 ----
 $ sudo virsh undefine minishift
 ----
@@ -67,7 +64,6 @@ You can now run `minishft delete` to delete the VM and restart Minishift.
 In case the above steps do not resolve the issue, you can also
 use the following command to delete the snapshots:
 
-[source,sh]
 ----
 $ rm -rf ~/.minishift/machines
 ----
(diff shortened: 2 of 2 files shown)


docs/source/getting-started/installing.adoc
```diff
@@ -48,14 +48,14 @@ https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/quick_start/wa
 using the Hyper-V Manager.
 - Make sure that you pair the virtual switch
 with a __network card (wired or wireless) that is connected to the network__.
-- If you have multiple virtual switches , set the environment variable _HYPERV_VIRTUAL_SWITCH_ to the name of external virtual switch you want to use for Minishift. See also the section about setting
-link:../using/managing-minishift{outfilesuffix}#driver-specific-environment-variables[driver specific enviropnment variables].
+- If you have multiple virtual switches , set the environment variable _HYPERV_VIRTUAL_SWITCH_ to the name of the external virtual switch you want to use for Minishift. For further information see
```
hferentschik: :+1

docs/source/using/troubleshooting.adoc
```diff
@@ -194,3 +194,29 @@ flag in the `minishift start` command. Alternatively, the problem might be
 an outdated version of VirtualBox.
 
 To avoid this issue, it is recommended to use VirtualBox 5.1.12 or later.
+
+[[insufficient-privileges]]
+== Hyper-V: Administrator privileges insufficient to run Hyper-V
+
```
hferentschik: I think there are some pieces missing here. In its core the problem is that even if you have Admin privileges, you need to run the shell in elevated mode to start Minishift. This is what @thesteve0 expected. You open your normal power shell and just type 'minshift start'. However, this does not work. In fact whatever the error is in this case, is what should be added to this trouble shooting entry.

So once the user hits the error, he has to choices. Either open a new shell in elevated mode or he does what we describe below and add himself to the HyperV admin group. If so, he can then open a user shell (non elevated) and start Minishift.

So "Hyper-V: Administrator privileges insufficient to run Hyper-V", is actually standalone misleading. As admin you can for sure start Hyper-V. The issue is trying to start Minishift in a normal user mode shell. I'd start with checking what the error message is. This is probably the title for the trouble shooting entry. Then I describe the two choices one has to fix it. Run shell in elevated mode or add myself to the Hyper-V admin group.

@Preeticp Does this make sense

Right now the title is confusing and there are logical gaps in why you can/should add yourself to the Hyper-V admin group.

docs/source/using/troubleshooting.adoc
```diff
@@ -194,3 +194,29 @@ flag in the `minishift start` command. Alternatively, the problem might be
...
+. Log off and log in for the change to take effect.
+
+You can also use the GUI to add yourself to the Hyper-V Administrators group as follows:
+
+. Click the *Start* button and choose *Computer Management*.
+. In the *Computer Management* window, select *Local Users And Groups* and then double click on *Groups*.
+. Double click on the *Hyper-V Administrators* group, the *Hyper-V Administrators Properties* dialog box is displayed.
+. Add your account to the Hyper-V Administrators group and log off and log in for the change to take effect.
+
+Now you can run the Hyper-V commands from a console or a shell as a normal user.
```
hferentschik: "from a console or a shell as a normal user" - so what's the difference between console and shell? Effectively that's the same thing, right?

docs/source/using/troubleshooting.adoc
Preeticp: Makes sense @hferentschik, I have got hold of a Windows machine for now, will verify this. I think @budhram is also trying it out.
I will add the missing pieces.

docs/source/using/troubleshooting.adoc
```diff
@@ -194,3 +194,29 @@ flag in the `minishift start` command. Alternatively, the problem might be
...
 To avoid this issue, it is recommended to use VirtualBox 5.1.12 or later.
+
+[[insufficient-privileges]]
+== Hyper-V: Administrator privileges insufficient to run Hyper-V
+
+If you run Minishift with Hyper-V on Windows, it is recommended that you add yourself to the Hyper-V Administrators group. You do not need to run as an Administrator in your shell.
+
+For example, if use PowerShell you can add yourself to the Hyper-V Administrators group as follows:
+
+. As an administrator run the following command:
```
budhrg: Will verify and add comment.

docs/source/using/troubleshooting.adoc
hferentschik: Awesome. I am fine with how we explain to add oneself to the Hyper-V admin group. It is really that the reasoning why this is required is missing. Effectively this is similar to "sudo" on Linux.

docs/source/getting-started/installing.adoc
LalatenduMohanty: 👍

docs/source/getting-started/installing.adoc
```diff
@@ -48,14 +48,14 @@ https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/quick_start/wa
...
+link:../using/managing-minishift{outfilesuffix}#driver-specific-environment-variables[setting driver specific environment variables].
 ====
 
 - https://www.virtualbox.org/wiki/Downloads[VirtualBox]
 +
 NOTE: It is recommended to use `Virtualbox 5.1.12` or later on Windows to avoid the issue
-link:../using/troubleshooting{outfilesuffix}#machine-doesnt-exist[Error: getting state for host: machine does not exist].
+link:../using/troubleshooting{outfilesuffix}#machine-doesnt-exist[Error: machine does not exist].
```
LalatenduMohanty: 👍

docs/source/using/troubleshooting.adoc
LalatenduMohanty: > So "Hyper-V: Administrator privileges insufficient to run Hyper-V", is actually standalone misleading.

+1.

Preeticp:
I'd start with checking what the error message is. This is probably the title for the trouble shooting entry.
> @hferentschik Checked and verified the error and updated the doc accordingly. Also added the few missing pieces.

budhrg:
Verified that the instructions are working and also able to do minishift start and deploy application without running PowerShell in elevated mode but having user in HyperV administrator group.

hferentschik:
+1, now we document the root cause. ;-)

Merged, thanks

Issue #721, opened 2017-04-10
Use topic map to dynamically generate navigation in Middleman site


hferentschik:
Following up with issue #711, the output of the Middleman generated site mimics now the output of the AsciiBinder build of docs.openshift.org. The missing piece is to incorporate the topic map and generate the links for the left side navigation dynamically. 

For this we need to add some Middleman helper generating the navigation bar dynamically from the topic map.

budhrg:
Closed via https://github.com/minishift/minishift/pull/810

PR #810, opened 2017-04-23
Issue #721 Using topic map to generate navigation

hferentschik:
Addresses issue #721


docs/helpers/custom_helpers.rb
```diff
@@ -0,0 +1,48 @@
...
+  	topic_map["Topics"].each_with_index { |topic, index|
+  		if topic["Dir"].nil?
+  			next "<li><a class=\"\" href=\"/#{topic["File"]}.html\">&nbsp;#{topic["Name"]}</a></li>"
+  		end
+
+  		expanded = ""
+  		subnav = ""
+  		entries = ""
+  		unless topic["Topics"].nil?
+            topic["Topics"].each { |subtopic|
```
budhrg: could we use `do ... end` format here instead of `{...}`? less intuitive from Ruby point of view

budhrg:
`make clean_docs && gen_docs` is giving me error on opening the url in browser.
screenshot from 2017-04-24 11-32-51

Looks like there is no `build/_topic.yml` file.

docs/helpers/custom_helpers.rb
```diff
@@ -0,0 +1,48 @@
+require 'yaml'
+
+module CustomHelpers
+
+  def create_navigation
+  	navigation = ""
+  	topic_map = YAML.load_file('build/_topic_map.yml')
```
budhrg: should be `source/_topic_map.yml`

docs/helpers/custom_helpers.rb
```diff
@@ -0,0 +1,48 @@
...
+  	topic_map["Topics"].each_with_index { |topic, index|
```
budhrg: could we use do ... end format here instead of {...}? less intuitive from Ruby point of view

docs/helpers/custom_helpers.rb
budhrg: Indentation seems to be messed up here. Unable to read the code properly.

docs/source/layouts/layout.erb
```diff
@@ -72,139 +72,7 @@
...
-                  <li><a class="" href="/command-ref/minishift_start.html">minishift start</a></li>
-
-                  <li><a class="" href="/command-ref/minishift_status.html">minishift status</a></li>
-
-                  <li><a class="" href="/command-ref/minishift_stop.html">minishift stop</a></li>
-
-                  <li><a class="" href="/command-ref/minishift_version.html">minishift version</a></li>
-              </ul>
-            </li>
+          <%= create_navigation %>
```
budhrg: `navigation_menu` ?

docs/helpers/custom_helpers.rb
```diff
@@ -0,0 +1,48 @@
...
+  		if topic["Dir"].nil?
```
budhrg: Le'ts extract `topics["Dir"]` and similar entities into some variable. Looks like used in multiple places

docs/helpers/custom_helpers.rb
hferentschik: ok

hferentschik:
> Looks like there is no build/_topic.yml file.

It's _build/_topic_map.yml_ and there is one. It's generated.

docs/helpers/custom_helpers.rb
hferentschik: -1, that won't have the command ref links. They are getting added as part of the build and the file gets generated under _build/_topic_map.yml_.

docs/helpers/custom_helpers.rb
hferentschik: ok

budhrg:
> It's _build/topic_map.yml and there is one. It's generated.

@hferentschik  Getting remove during `make gen_docs` step, See below:

```
....
      create  build/command-ref/minishift_version.html
      create  build/command-ref/minishift_hostfolder_umount.html
      remove  build/_topic_map.yml
```

budhrg:
Hi @hferentschik , 

The PR looks great with all layout stuff (after fixing topic.yml locally), Highlight in navigation works fine except the pointer 

1493035944_ic_play_arrow_48px 

which is functionationing wrong after clicking link inside it like `Quickstart`.
See below
wrong_pointer

docs/helpers/custom_helpers.rb
hferentschik: ok

docs/helpers/custom_helpers.rb
hferentschik: I reformatted the file.

hferentschik:
> Getting remove during make gen_docs step, See below:

true, true - that's an issue when generating the site. I just tested the _serve_ target. It's a middleman thing. Need to think ...

hferentschik:
> Highlight in navigation works fine except the pointer

I did not look into the required styles there. I guess we can ...

hferentschik:
> Getting remove during make gen_docs step, See below:

Fixed, as well as the caret.

docs/source/layouts/layout.erb
hferentschik: just for you ;-)

budhrg:
LGTM :+1:

PR #813, opened 2017-04-24
[WIP] Issue #72 Add update command

dharmit:
Addresses issue #72

`minishift update` prompts user to update minishift if a newer version
is available.


pkg/minikube/update/update.go
```diff
@@ -64,18 +67,20 @@ func MaybeUpdateFromGithub(output io.Writer) {
 
 func MaybeUpdate(output io.Writer, githubOwner, githubRepo, binaryName, lastUpdatePath string, localVersion semver.Version) {
 
-	downloadBinary := binaryName + "-" + runtime.GOOS + "-" + runtime.GOARCH
+	latestVersion, err := getLatestVersionFromGitHub(githubOwner, githubRepo)
+	if err != nil {
+		glog.Errorln(err)
```
LalatenduMohanty: We are now using atexit.ExitWithMessage method instead of glog.Errorln().

pkg/minikube/update/update.go
```diff
@@ -90,11 +95,13 @@ automatically update from %s%s to %s%s now? [y/N] `,
 
 		if strings.ToLower(confirm) == "y" {
```
LalatenduMohanty: It should accept both `y` and `Y`.

pkg/minikube/update/update.go
```diff
@@ -90,11 +95,13 @@ automatically update from %s%s to %s%s now? [y/N] `,
...
 			fmt.Printf("Updating to version %s\n", latestVersion)
-			updateBinary(latestVersion, downloadBinary, updateLinkPrefix, downloadLinkFormat)
+			updateBinary(latestVersion, downloadTarball, updateLinkPrefix, downloadLinkFormat)
 			return
 		}
 
-		fmt.Println("Skipping auto-update.")
+		fmt.Println("Skipping update.")
+	} else {
+		fmt.Println("Already using latest version.")
```
LalatenduMohanty: I think we should mention the version too. so that user will know which is the latest version here.

LalatenduMohanty: s/Already/You are already/

pkg/minikube/update/update.go
```diff
@@ -175,41 +182,77 @@ func updateBinary(v semver.Version, downloadBinary, updateLinkPrefix, downloadLi
...
+
+	tarballChecksum, err := checksumFor(crypto.SHA256, tarballBytes)
+
+	if !bytes.Equal(tarballChecksum, checksum) {
+		fmt.Println(fmt.Errorf("Updated file has wrong checksum. Expected: %x, got: %x", tarballChecksum, checksum))
+		os.Exit(1)
+	}
+
+	// Write tar file to disk
+	err = ioutil.WriteFile("/tmp/minishift.tgz", tarballBytes, 0644)
```
LalatenduMohanty: Please use something similar to `tempDir, err := ioutil.TempDir("", "minishift")` to create the temp directory and then use it for the purpose.

pkg/minikube/update/update.go
hferentschik: :+1

pkg/minikube/update/update.go
```diff
@@ -64,18 +67,20 @@ func MaybeUpdateFromGithub(output io.Writer) {
...
 func MaybeUpdate(output io.Writer, githubOwner, githubRepo, binaryName, lastUpdatePath string, localVersion semver.Version) {
 
-	downloadBinary := binaryName + "-" + runtime.GOOS + "-" + runtime.GOARCH
+	latestVersion, err := getLatestVersionFromGitHub(githubOwner, githubRepo)
+	if err != nil {
+		glog.Errorln(err)
+		return
+	}
+
+	downloadTarball := fmt.Sprintf("%s-%s-%s-%s.tgz", binaryName, latestVersion, runtime.GOOS, runtime.GOARCH)
```
hferentschik: We are not only downloading tarballs. Windows archive format is zip

pkg/minikube/update/update.go
```diff
@@ -90,11 +95,13 @@ automatically update from %s%s to %s%s now? [y/N] `,
 
 		if strings.ToLower(confirm) == "y" {
 			fmt.Printf("Updating to version %s\n", latestVersion)
-			updateBinary(latestVersion, downloadBinary, updateLinkPrefix, downloadLinkFormat)
+			updateBinary(latestVersion, downloadTarball, updateLinkPrefix, downloadLinkFormat)
```
hferentschik: -1 to variable name. We are dealing with archives in general. We have tar as well as zip format.

pkg/minikube/update/update.go
hferentschik: +1

pkg/minikube/update/update.go
hferentschik: -1 how is this working for Windows? There are several things wrong here. The use of the _/tmp_ directory as well as hard coding the name. You will need to look at what you are actually downloading here.

pkg/minikube/update/update.go
```diff
@@ -236,3 +279,18 @@ func downloadChecksum(v semver.Version, downloadBinary, downloadLinkFormat strin
...
+func checksumFor(h crypto.Hash, payload []byte) ([]byte, error) {
+	if !h.Available() {
+		return nil, errors.New("requested hash function not available")
+	}
+	hash := h.New()
+	hash.Write(payload) // guaranteed not to error
+	return hash.Sum([]byte{}), nil
+}
+
+func removeFiles() {
```
hferentschik: -1, if you use a tmp directory you can clean up via a defer

pkg/minikube/update/update.go
hferentschik: In fact, I think we should use the tmp directory of the MINISHIFT_HOME to do all this, e.g. _ ~/.minishift/tmp/_

pkg/minikube/update/update.go
praveenkumar: I think that what's `ToLower` will do.

budhrg:
Some few unit tests too.

hferentschik:
I think we need to discuss also, how this is supposed to work. There are several gaps here.

dharmit:
@hferentschik @budhrg I've added code to cover the scenario of a Windows system. However, not able to test it since I don't have a Windows system around. 

Would like to cover any possible gaps that are left uncovered in this PR.

dharmit:
> how this is supposed to work

Based on what I have added through this PR and what I've understood by looking at the code:

- when a user does `minishift update`, we check for latest version on GitHub. 
- If a newer version of minishift is available, we download the archive (tarball for Linux & Mac, zip for Windows). The way we force minishift to think that it's running on older version is by modifying `Makefile`.
- Then we compare the checksum of downloaded archive with the checksum we downloaded before downloading the archive itself.
- If the checksums match, we extract the content of the archive to a tempdir which we delete with help of `defer` call. (thank you @hferentschik)
- We then apply the update which replaces the minishift binary on disk (`~/go/bin/minishift` on Linux dev environment) with the binary from archive.
- Earlier we used to run `minishift update` after downloading the new binary. But that failed because the new archive (`1.0.0-rc.2` at the time of opening this PR) didn't have `update` switch. As a result `minishift update` would error out with `no such command` message. Me and @LalatenduMohanty debugged this but couldn't find any other possible explanation. So I modified it to print the version instead i.e. `minishift version`.

> Some few unit tests too.

Sure. I'd need some help on that however.

hferentschik:
I'll add some comments on #72

pkg/minikube/update/update.go
```diff
@@ -175,27 +190,54 @@ func updateBinary(v semver.Version, downloadBinary, updateLinkPrefix, downloadLi
...
+
+	archiveChecksum, err := checksumFor(crypto.SHA256, archiveBytes)
+
+	if !bytes.Equal(archiveChecksum, checksum) {
+		fmt.Println(fmt.Errorf("Updated file has wrong checksum. Expected: %x, got: %x", archiveChecksum, checksum))
+		os.Exit(1)
+	}
+
+	// Create a temporary directory to store archive contents
+	dir, err := ioutil.TempDir("", "minishift")
```
hferentschik: We should use the Minishift tmp dir.

pkg/minikube/update/update.go
```diff
@@ -175,27 +190,54 @@ func updateBinary(v semver.Version, downloadBinary, updateLinkPrefix, downloadLi
...
+	if err != nil {
+		atexit.ExitWithMessage(1, "Could not create a temporary directory to store archive contents")
```
hferentschik: We should not be using atexit anywhere outside of the commands. If there is an error, it should be returned.

hferentschik:
> Some few unit tests too.

+1 I'd like to see this working in a unit test as well.

pkg/minikube/update/update.go
budhrg: `confirm == 'Y' || strings.ToLower(confirm)` should be there.

pkg/minikube/update/update.go
```diff
@@ -162,40 +180,73 @@ func updateBinary(v semver.Version, downloadBinary, updateLinkPrefix, downloadLi
...
+
+	dir, err := ioutil.TempDir(constants.Minipath, "tmp/download")
+	if err != nil {
+		return err
+		// atexit.ExitWithMessage(1, "Could not create a temporary directory to store archive contents")
+	}
+
+	defer os.RemoveAll(dir) // clean up
+
+	binary = extractAndReplace(runtime.GOOS, dir, archiveBytes)
```
LalatenduMohanty: I am not sure why we need to pass `archiveBytes` instead of `updatedArchive`. cc @hferentschik

pkg/minikube/update/update.go
```diff
@@ -236,3 +288,51 @@ func downloadChecksum(v semver.Version, downloadBinary, downloadLinkFormat strin
...
+}
+
+func extractAndReplace(goos, dir string, archiveBytes []byte) string {
+	downloadedArchive := filepath.Join(dir, "minishift.tgz")
+	ungzip := filepath.Join(dir, "minishift.gzip")
+	extract := filepath.Join(dir, "minishift-extract")
+	binary := filepath.Join(extract, "minishift")
+
+	// Write archive file to disk
+	err := ioutil.WriteFile(downloadedArchive, archiveBytes, 0644)
```
LalatenduMohanty: Something is wrong here. `archiveBytes` has []byte for .zip and .tgz, isn't it? But we are writing only minishift.tgz.

pkg/minikube/update/update.go
```diff
@@ -162,40 +180,73 @@ func updateBinary(v semver.Version, downloadBinary, updateLinkPrefix, downloadLi
...
+	archiveChecksum, err := checksumFor(crypto.SHA256, archiveBytes)
+
+	if !bytes.Equal(archiveChecksum, checksum) {
+		fmt.Println(fmt.Errorf("Updated file has wrong checksum. Expected: %x, got: %x", archiveChecksum, checksum))
+		os.Exit(1)
+	}
+
+	// Create a temporary directory inside minishift directory to store archive contents
+
+	dir, err := ioutil.TempDir(constants.Minipath, "tmp/download")
```
hferentschik: Actually, I change my mind. If we determine the tmp directory so late in the game, it is better to use just ioutil.TempDir without _constants.Minipath,_. Using this constant so deep in the code is bad. If we want to use the Minishift tmp dir we need to pass the tmp dir as a parameter.

pkg/minikube/update/update.go
hferentschik: It does not make much sense to me either. I recommend to have a look at _https://github.com/minishift/minishift/blob/8a552763f5ea21f9c9b6dea4cb874ede9e735896/pkg/util/github/github.go#L91-L91_. There we also deal with downloading various archive types and unpacking them.

hferentschik: Why _ extractAndReplace_, what does get replaced there?

pkg/minikube/update/update.go
hferentschik: It not wrong, just fishy. The code just writes the file to _minishift.tgz_ regardless. It then checks the OS and if on Windows just unzips it, regardless of the extension. 

So why not download and safe the archive as is, with the proper filename. Then you decide whether to use tar or zip based on the filename. 

Also, as has as I can tell, this method is supposed returns the path to the extracted minishift executable - binary := filepath.Join(extract, "minishift"). However, on Windows the binary is _minishift.exe_. I have not tried, but would be very surprised if this works. One needs to handle the download and naming explicitly.

pkg/minikube/constants/constants_darwin.go
```diff
@@ -26,3 +26,4 @@ var SupportedVMDrivers = [...]string{
 
 const DefaultVMDriver = "xhyve"
 const OC_BINARY_NAME = "oc"
+const BINARY_NAME = "minishift"
```
budhrg: @dharmit , please remove this and other related changes as it is not being used now.

pkg/minikube/update/update.go
dharmit: > I am not sure why we need to pass `archiveBytes` instead of `updatedArchive`

My understanding is that we need to work on `[] byte` for writing and unpacking the binary from tar/zip. So instead of having to do `archiveBytes, err := ioutil.ReadAll(updatedArchive)` again, I passed it to the function. Is that a bad idea or is their a better way to do it?

> Why _ extractAndReplace_, what does get replaced there?

Agree with you. Renaming it to `extractBinary` since it's returning a binary. Does that sound OK?

pkg/minikube/update/update.go
dharmit: Right. Quite a few glitches there! How about I check the OS first and perform operations based on that? If OS is Windows, save to a file with `.zip` extension else `.tgz` extension? Then extract data out of it and return the path to binary (with `exe` extension on Windows). Modifying the code for your feedback.

cmd/minishift/cmd/update.go
```diff
@@ -0,0 +1,45 @@
...
+
+func runUpdate(cmd *cobra.Command, args []string) {
+	err := update.MaybeUpdateFromGithub(os.Stdout)
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Update failed: %s", err))
+	}
+}
+
+func init() {
+	RootCmd.AddCommand(updateCmd)
```
dharmit: Sounds great. How about taking it as a separate issue/PR outside this one? I'll send in a new one once this gets merged.

Or do you think that as an absolute must for this PR to move further?

hferentschik:
Could you squash your commits, effectively we are just dealing with updates to single file.

cmd/minishift/cmd/update.go
hferentschik: As you like.

pkg/minikube/update/update.go
```diff
@@ -49,64 +50,86 @@ const (
...
 
-var (
-	lastUpdateCheckFilePath = constants.MakeMiniPath("last_update_check")
-)
-
-func MaybeUpdateFromGithub(output io.Writer) {
+func MaybeUpdateFromGithub(output io.Writer) error {
 	localVersion, err := version.GetSemverVersion()
 	if err != nil {
 		glog.Errorln(err)
```
hferentschik: let's remove the  log and be consistent. once you log once you don't. Let's just return the error and let the caller deal with it

pkg/minikube/update/update.go
```diff
@@ -49,64 +50,86 @@ const (
...
+		return err
 	}
-	MaybeUpdate(output, githubOwner, githubRepo, githubRepo, lastUpdateCheckFilePath, localVersion)
+
+	latestVersion, err := getLatestVersionFromGitHub(githubOwner, githubRepo)
+	if err != nil {
+		return err
+	}
+
+	if checkLatestVersion(localVersion, latestVersion) {
```
hferentschik: _isNewerVersion_ or something like this. _checkLatestVersion_ does not really make clear what it does

pkg/minikube/update/update.go
```diff
@@ -49,64 +50,86 @@ const (
...
-		return
+	if runtime.GOOS == "windows" {
+		extName = "zip"
+	} else {
+		extName = "tgz"
 	}
-	latestVersion, err := getLatestVersionFromGitHub(githubOwner, githubRepo)
+
+	archiveName := fmt.Sprintf("minishift-%s-%s-%s.%s", latestVersion, runtime.GOOS, runtime.GOARCH, extName)
+	// updateLinkPrefix := "https://github.com/" + githubOwner + "/" + githubRepo + "/releases/tag/" + version.VersionPrefix
```
hferentschik: I guess this needs to be clean up, right

pkg/minikube/update/update.go
```diff
@@ -49,64 +50,86 @@ const (
...
+	downloadLinkFormat := "https://github.com/" + githubOwner + "/" + githubRepo + "/releases/download/v%s/%s"
+
+	fmt.Printf("Updating to version %s\n", latestVersion)
+	// err := updateBinary(latestVersion, archiveName, updateLinkPrefix, downloadLinkFormat)
```
hferentschik: same as above

pkg/minikube/update/update.go
```diff
@@ -149,71 +164,66 @@ func getTimeFromFileIfExists(path string) time.Time {
...
-		glog.Errorf("Cannot download checksum: %s", err)
-		os.Exit(1)
-	}
-
-	currentBinary, err := osext.Executable()
-	if err != nil {
-		glog.Errorf("Cannot find current binary to execute: %s", err)
-		os.Exit(1)
-	}
+func downloadAndSaveArchive(v semver.Version, archiveName, downloadLinkFormat string) (string, string, error) {
```
hferentschik: would be nice with some godocs telling you what this actually returns

pkg/minikube/update/update.go
```diff
@@ -149,71 +164,66 @@ func getTimeFromFileIfExists(path string) time.Time {
...
+	} else {
+		// Store archive in tgz file
+		downloadedArchive = filepath.Join(dir, "minishift.tgz")
+
+		if err := writeArchiveToDisk(downloadedArchive, archiveBytes); err != nil {
+			return "", "", err
 		}
-		os.Exit(1)
 	}
+	return downloadedArchive, dir, nil
```
hferentschik: why do you need to return the full path to the archive + the directory. The latter is implicit, right? Even if you later on need it again, you can easily get it from _downloadedArchive_, right?

pkg/minikube/update/update.go
```diff
@@ -49,64 +50,86 @@ const (
...
-		var confirm string
-		fmt.Scanln(&confirm)
+	defer os.RemoveAll(dir)
 
-		if strings.ToLower(confirm) == "y" {
-			fmt.Printf("Updating to version %s\n", latestVersion)
-			updateBinary(latestVersion, downloadBinary, updateLinkPrefix, downloadLinkFormat)
-			return
-		}
+	if err := downloadAndVerifyChecksum(downloadedArchive, downloadLinkFormat, archiveName, latestVersion); err != nil {
```
hferentschik: This really should be part of _downloadAndSaveArchive_. If the checksum does not match, downloadAndSaveArchive_ should just throw an error. Verifying the checksum is integral part of downloadAndSaveArchive

pkg/minikube/update/update.go
```diff
@@ -49,64 +50,86 @@ const (
...
+	if err != nil {
+		glog.Errorf("Cannot find current binary to execute: %s", err)
+		os.Exit(1)
 	}
-}
 
-func shouldCheckURLVersion(filePath string) bool {
-	if !viper.GetBool(config.WantUpdateNotification) {
-		return false
+	err = extractAndUpdateBinary(currentBinary, downloadedArchive, dir)
```
hferentschik: This still puts extraction and update together. Why not just _extractBinary_ which in turn returns the path to the extracted binary. Then you pass this path to a _updateBinary_

hferentschik: currentBinary is an unused parameter.

pkg/minikube/update/update.go
```diff
@@ -228,11 +238,107 @@ func downloadChecksum(v semver.Version, downloadBinary, downloadLinkFormat strin
...
+		binary = filepath.Join(extract, "minishift")
+	}
+
+	binaryFile, err := os.Open(binary)
+	if err != nil {
+		return err
 	}
 
-	return hex.DecodeString(strings.TrimSpace(string(b)))
+	// binaryBytes, err := ioutil.ReadAll(binaryFile)
```
hferentschik: I guess this needs to be clean up as well

pkg/minikube/update/update.go
```diff
@@ -228,11 +238,107 @@ func downloadChecksum(v semver.Version, downloadBinary, downloadLinkFormat strin
...
+
+	// if err != nil {
+	// 	return err
+	// }
+
+	// if err := ioutil.WriteFile(currentBinary, binaryBytes, 0755); err != nil {
+	// 	return err
+	// }
+
+	err = update.Apply(binaryFile, update.Options{
```
hferentschik: this part to the end should be really in a separate method.

pkg/minikube/update/update.go
dharmit: Returning the full path so that it can be used in other functions that are going to extract the binary or replace existing binary with updated on.

Agree with `dir` part. Can get rid of it.

pkg/minikube/update/update.go
```diff
@@ -49,66 +50,100 @@ const (
...
+	args := []string{currentBinary, "version"}
+	err = syscall.Exec(currentBinary, args, env)
+	if err != nil {
+		glog.Errorf("Failed to execute updated binary %s: %s", currentBinary, err)
+		os.Exit(1)
 	}
-	lastUpdateTime := getTimeFromFileIfExists(filePath)
-	if time.Since(lastUpdateTime).Hours() < viper.GetFloat64(config.ReminderWaitPeriodInHours) {
-		return false
+	return nil
```
hferentschik: Is this temporary code?

pkg/minikube/update/update.go
```diff
@@ -49,66 +50,100 @@ const (
...
+	if err != nil {
```
hferentschik: This can never happen, given the err check from above

pkg/minikube/update/update.go
```diff
@@ -49,66 +50,100 @@ const (
...
+
+	// Extract and replace the downloaded archive in place of current binary
+
+	binary, err := extractBinary(downloadedArchive)
 	if err != nil {
-		glog.Errorln(err)
-		return
+		return err
 	}
-	if localVersion.Compare(latestVersion) < 0 {
```
hferentschik: This got completely dropped now. I think we this interactive step. We should get this back. I'd put that in the update command and expose from update.go

* CurrentVersion()
* LatestVersion()
* IsNewer(current, latest)
* Update (basically current MaybeUpdate)

This way you can implement the actual update logic in the update command. Determine whether there is a newer version, if not, display that Minishift is up to date. ATM, there is no feedback whatsoever, if your binary is up to date. 

If there is a newer binary, you ask for confirmation of the update. If update is confirmed, you make the call for updating the binary.

pkg/minikube/update/update.go
```diff
@@ -129,91 +164,68 @@ func getLatestVersionFromGitHub(githubOwner, githubRepo string) (semver.Version,
...
-	err = syscall.Exec(currentBinary, args, env)
-	if err != nil {
-		glog.Errorf("Failed to execute updated binary %s: %s", currentBinary, err)
-		os.Exit(1)
-	}
-}
-
-func updateBinaryFile(url string, checksum []byte) {
-	fmt.Println("Downloading updated binary")
+	fmt.Println("Downloading updated archive")
```
hferentschik: "Downloading release" or  fmt.Sprintf("Downloading %s, url)

pkg/minikube/update/update.go
```diff
@@ -129,91 +164,68 @@ func getLatestVersionFromGitHub(githubOwner, githubRepo string) (semver.Version,
...
-		os.Exit(1)
 	}
-}
 
-func downloadChecksum(v semver.Version, downloadBinary, downloadLinkFormat string) ([]byte, error) {
-	fmt.Println("Downloading updated binary checksum to validate updated binary")
-	u := fmt.Sprintf(downloadLinkFormat, v, downloadBinary+".sha256")
+	// Download and verify checksum
+
+	fmt.Println("Downloading updated archive checksum to validate updated archive")
```
hferentschik: I don't think this message is relevant for the user

pkg/minikube/update/update.go
dharmit: @hferentschik i think I missed removing it during the refactoring.

pkg/minikube/update/update.go
```diff
@@ -129,91 +164,68 @@ func getLatestVersionFromGitHub(githubOwner, githubRepo string) (semver.Version,
...
+	dir, err := ioutil.TempDir("", "download")
+	if err != nil {
+		return "", errors.New("Could not create a temporary directory to store archive contents")
+	}
+
+	var downloadedArchive string
+
+	if runtime.GOOS == "windows" {
+		// Store archive in zip file on Windows
+		downloadedArchive = filepath.Join(dir, "minishift.zip")
```
hferentschik: You have the true achieve name in the url, right? In _archiveName_ resp. _url_. Why not getting it from there? This seems to be fragile and makes assumptions which might change.

pkg/minikube/update/update.go
```diff
@@ -129,91 +164,68 @@ func getLatestVersionFromGitHub(githubOwner, githubRepo string) (semver.Version,
...
-	if err != nil {
-		glog.Errorf("Cannot download checksum: %s", err)
-		os.Exit(1)
-	}
+// downloadAndVerifyArchive downloads the archive of latest version of minishift from GitHub, saves it to a temporary location,
+// and verifies the checksum of downloaded archive with checksum available on GitHub.
+// It returns a string containing path to the downloaded archive.
+// It returns an error if any of the steps failed.
+func downloadAndVerifyArchive(latestVersion semver.Version, archiveName, downloadLinkFormat string) (string, error) {
+	url := fmt.Sprintf(downloadLinkFormat, latestVersion, archiveName)
```
hferentschik: why don't you not just pass the url into this method? This seems to be all that is really relevant, right?

pkg/minikube/update/update.go
```diff
@@ -129,91 +164,68 @@ func getLatestVersionFromGitHub(githubOwner, githubRepo string) (semver.Version,
...
-}
-
-func updateBinaryFile(url string, checksum []byte) {
-	fmt.Println("Downloading updated binary")
+	fmt.Println("Downloading updated archive")
 	httpResp, err := http.Get(url)
 	if err != nil {
-		glog.Errorf("Cannot download binary: %s", err)
-		os.Exit(1)
+		return "", errors.New("Cannot download archive")
```
hferentschik: use the explicit URL, more context about the failure

pkg/minikube/update/update.go
```diff
@@ -228,11 +240,108 @@ func downloadChecksum(v semver.Version, downloadBinary, downloadLinkFormat strin
...
+		return "", errors.New(fmt.Sprintf("received %d", checksumResp.StatusCode))
 	}
 
-	return hex.DecodeString(strings.TrimSpace(string(b)))
+	downloadedChecksum, err := hex.DecodeString(strings.TrimSpace(string(b)))
+	if err != nil {
+		return "", err
+	}
+
+	archiveBytes, err = ioutil.ReadFile(downloadedArchive)
```
hferentschik: You already have the archive bytes, why re-read them?

gbraad:
I believe the current issue with:

```powershell
PS C:\work\src\github.com\minishift\minishift> C:\work\bin\minishift.exe update
Updating to version 1.0.1
Downloading updated archive
 3.27 MB / 3.43 MB [=======================================================================================================================================================>-------]  95.29% 0sD
ownloading updated archive checksum to validate updated archive
 65 B / 65 B [=====================================================================================================================================================================] 100.00% 0s

E0515 21:17:22.983876   11760 update.go:135] Failed to execute updated binary C:\work\bin\minishift.exe: not supported by windows
```

might be related to kicking the executable out from underneath the running process and replace it with a new executable. you are then trying to invoke it from the same execution context. I guess this can lead to a disallowed state. After the/at the time of the run, you can actually find a `.minishift.exe.old` and a `minishift.exe` which can both be executed from cmd.

budhrg:
> E0515 21:17:22.983876   11760 update.go:135] Failed to execute updated binary C:\work\bin\minishift.exe: not supported by windows

I am having same error. 

> you are then trying to invoke it from the same execution context. I guess this can lead to a disallowed state.

It seems so.

budhrg:
@dharmit will look tomorrow sure

hferentschik:
> might be related to kicking the executable out from underneath the running process

+1, I am surprised it works so far, trying to call this executable which you just updated from within the same process it questionable. Looking at the go-update example, it does actually terminate after the update. 

I know this was code which already existed, but I am not sure whether this was ever tested on Windows. I think we might be ok with removing it altogether and relying on the error reporting of go-update.

budhrg:
Update from my side,

Currently going through the comments in go-update issue https://github.com/inconshreveable/go-update/issues/5.

Doing more investigation on https://github.com/rcrowley/goagain/ as mentioned in this comment https://github.com/inconshreveable/go-update/issues/5#issuecomment-69524926.

https://github.com/inconshreveable/go-update hasn't been updated since a year.

My plan is to check the code mentioned https://github.com/inconshreveable/go-update/issues/5#issuecomment-69524926 or do investigating in https://github.com/rcrowley/goagain/ to see if it works in our usecase.

dharmit:
> I know this was code which already existed, but I am not sure whether this was ever tested on Windows. I think we might be ok with removing it altogether and relying on the error reporting of go-update. Calling version on the new binary is nothing which is per se required as part of the update.

+1

I had kept the piece of code that executes something based off the newly downloaded binary because the old code did that. I'm wondering if it adds any value to execute `minishift version` after the update? Earlier, it was calling `minishift update` but that failed since we were downloading a new binary from GitHub which didn't have the `update` command.

pkg/minikube/update/update.go
dharmit: Yes, we can use the last part of the url. That also helps use remove the `runtime.GOOS` check, I guess.

```go
urlSplit := strings.Split(url, "/")

downloadedArchive = filepath.Join(dir, urlSplit[len(urlSplit)-1])
if err := writeArchiveToDisk(downloadedArchive, archiveBytes); err != nil {
    return "", err
}
```

cmd/minishift/cmd/update.go
```diff
@@ -0,0 +1,69 @@
...
+		atexit.ExitWithMessage(1, fmt.Sprintf("Update failed: %s", err))
+	}
+
+	latestVersion, err := update.LatestVersion()
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Update failed: %s", err))
+	}
+
+	if update.IsNewerVersion(localVersion, latestVersion) {
+		fmt.Printf("A newer version of minishift is available.\nDo you want to update from %s to %s now? [y/N]: ", localVersion, latestVersion)
```
budhrg: Should be `[y/n]`.

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,336 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = updateBinary(binary)
+	if err != nil {
+		return err
+	}
+
+	// env := os.Environ()
```
budhrg: Are we really not verifying now?

cmd/minishift/cmd/update.go
dharmit: That indicates that if a user will hit return without entering a "y" or "N", default action is to not update the binary. That's how `dnf update` and `apt-get update` work as well. `apt` defaults to "Y" and `dnf` defaults to "N", IIRC.

pkg/minishift/update/update.go
dharmit: We're not verifying it by the way of executing `minishift version` because that doesn't work well on Windows. If you have suggestion(s) around some other way of doing it, I'm open to doing it.

budhrg: Ack @dharmit , look reasonable to me to with it.

budhrg:
Closing this PR as it is superseded by the PR https://github.com/minishift/minishift/pull/930.

Thanks @dharmit  for the `minishift update` implementation.

PR #818, opened 2017-04-25
Issue #814 major.minor.patch version format for jteeuwen/go-bindata

### LINKED ISSUE #814: Set explicit version for jteeuwen/go-bindata
I'm parsing glide.yaml in order to retrieve the url to github tarballs, but there is no "3.0" release of 
https://github.com/jteeuwen/go-bindata/releases - please use e.g. the latest "3.0.7"

This is the line in glide.yaml
https://github.com/minishift/minishift/blob/fd33efc0d76e3f21f170ea113c42f42135c9a374/glide.yaml#L47

budhrg:
Fix #814

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -25,6 +25,7 @@ import:
   - libmachine/state
   - libmachine/swarm
 - package: github.com/google/go-github
+  version: 30a21ee1a3839fb4a408efe331f226b73faac379
   subpackages:
   - github
 - package: github.com/inconshreveable/go-update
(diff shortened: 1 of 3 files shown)


glide.yaml
```diff
@@ -25,6 +25,7 @@ import:
   - libmachine/state
   - libmachine/swarm
 - package: github.com/google/go-github
+  version: 30a21ee1a3839fb4a408efe331f226b73faac379
```
budhrg: The latest commit of `google/go-github` is not working due to no of arguments change in few functions.

praveenkumar:
I would say we should hold this PR till 1.0 release because it have Makefile changes and I am bit spectacle about taking those in downstream.

hferentschik:
> I would say we should hold this PR till 1.0 release because it have Makefile changes and I am bit spectacle about taking those in downstream.

+1 There is no need at this stage to fiddle with versions. There are plenty of unrelated updates to the lock file.

I am going to close this issue and we can look at it again in due time.

Issue #782, opened 2017-04-20
eval $(minishift docker-env) not working onWindows with cmd or PowerShell


budhrg:
The command `eval $(minishift docker-env)` as mentioned in [reusing docker daemon](https://docs.openshift.org/latest/minishift/using/docker-daemon.html) is not working due to `eval` not found in Windows.
```
PS C:\> eval $(minishift docker-env)
eval : The term 'eval' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify
that the path is correct and try again.
At line:1 char:1
+ eval $(minishift docker-env)
+ ~~~~
    + CategoryInfo          : ObjectNotFound: (eval:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException
```

We need to come up with strategy for Windows users here.

#### NOTE: MingW Terminal  is working fine.

hferentschik:
This is if at all a documentation issue. `minishift docker-env` tells you the right syntax when executing the command. There is logic for determining the OS and shell type.

So really the user should run `minishift docker-env` and the cut&paste the appropriate output.

budhrg:
Resolved via https://github.com/minishift/minishift/pull/821

PR #821, opened 2017-04-25
Issue #782 Improving docs to configure console to reuse Docker daemon

hferentschik:
Addresses issue #782


hferentschik:
@budhrg here we go. This avoids all OS specifics.

budhrg:
LGTM :+1: 

I was thinking on the same lines :smile:

budhrg:
Merged and thanks

Issue #142, opened 2016-11-04
Add 'oc-env' command which display information to eval the oc binary to PATH


hferentschik:
See issue #141 

> We can also offer something similar to docker-env, say oc-env which will output something which is evaluable in the shell and updates the PATH in order to execute oc

LalatenduMohanty:
+1, this will be super useful for OC binary users.

budhrg:
Also, along with adding cached oc binary into PATH, display other information can be useful
```
# You can access the OpenShift console on: https://10.1.2.2:8443/console
# To use OpenShift CLI, run: oc login https://10.1.2.2:8443
export OPENSHIFT_URL=https://10.1.2.2:8443
export OPENSHIFT_WEB_CONSOLE=https://10.1.2.2:8443/console
export DOCKER_REGISTRY=hub.openshift.centos7-adb.10.1.2.2.xip.io
``` 

Taken from [ADB doc](https://github.com/projectatomic/adb-atomic-developer-bundle/blob/master/docs/using.adoc#configure-your-environment)

hferentschik:
Actually we might want to create a _oc_ subcommand instead, allowing for potentially other _oc_ related commands. See also issue #349

budhrg:
Related to #465

budhrg:
@hferentschik @praveenkumar @LalatenduMohanty 

So here we need to display OS specific instructions to be evaluated in shell.

I am bit concern how we handle Windows specific instructions. Eval is not available as per this https://github.com/minishift/minishift/issues/782 so should we just display the instructions and tell user to run it ?

hferentschik:
I don't understand why the problem is. I might be missing something. But 'docker-env' is already doing all the Os and shell specific stuff - https://github.com/minishift/minishift/blob/7974f5edc9697c6b878794aa9f98d9d0fe668588/cmd/minishift/cmd/env.go#L66-L66. So 'oc-env' would do the equivalent, and yes, the command to change the PATH needs to be different depending on OS and shell. 

From a ux point of view, we do the same as with 'docker-env'. The user runs 'minishift oc-env'. This will determine OS and shell and print the right command for the user to configure his shell with the modified PATH. The user then cut&paste this command into the shell.

Where exactly is the problem?

hferentschik:
[...]

PR #835, opened 2017-04-27
Issue #142 Introduce oc-env command

hferentschik:
Addresses issue #142

### DIFF
--- a/cmd/minishift/cmd/oc_env_test.go
+++ b/cmd/minishift/cmd/oc_env_test.go
@@ -1,0 +1,34 @@
+// +build !windows
+
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"testing"
+)
+
+func Test_unix_oc_path(t *testing.T) {
+	shellConfig, err := getOcShellConfig("/Users/john/.minishift/cache/oc/v1.5.0/oc", "")
+	if err != nil {
+		t.Fatalf("Unexepcted error: %s", err)
+	}
+
+	if shellConfig.OcPath != "/Users/john/.minishift/cache/oc/v1.5.0" {
+		t.Fatalf("Unexepcted oc path: %s", shellConfig.OcPath)
+	}
+}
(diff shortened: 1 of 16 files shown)


gbraad:
I noticed side of the boilerplate show a copyright of 2016. I will still merge as this is not a breaking issue...

gbraad:
Not merging: test for ocpath fails

hferentschik:
> Not merging: test for ocpath fails

Right, on Windows this one cannot run. Give me a minute.

hferentschik:
> I noticed side of the boilerplate show a copyright of 2016. I will still merge as this is not a breaking issue...

Let me fix that as well then

gbraad:
centos CI fails the scenario? Can't go into more details now.

hferentschik:
> centos CI fails the scenario? Can't go into more details now.

That's on and off at the moment. We thought we already addressed it by retrying the command. I can increase the retry count and we can see whether we get less failures. We went from failing every time, to every third time or so. However, we should take a closer look at this test again, since I am a bit suspicious why a retry and wait is required.

hferentschik:
retest this please

hferentschik:
@gbraad I actually disabled the failing integration test scenario and created https://github.com/minishift/minishift/issues/837 to follow up on this. It fails too often. Either the implementation for retrying has a bug or there is something else going on which is not addressed by the retry.

cmd/minishift/cmd/oc_env_test.go
```diff
@@ -0,0 +1,34 @@
+// +build !windows
```
budhrg: Looks like this test is not passing Windows. What is the issue there?

cmd/minishift/cmd/oc_env_test_windows.go
```diff
@@ -0,0 +1,32 @@
+/*
```
budhrg: :+1:  Works

cmd/minishift/cmd/oc_env_test.go
hferentschik: It cannot pass on windows. It need to be excluded

gbraad:
OK. LGTM.
The test has passed by several times I noticed while glancing at IRC during my leave. Needs a real fix. Agree to disable.

hferentschik:
@gbraad green light :-)

docs/source/using/interacting-with-openshift.adoc
```diff
@@ -24,14 +24,15 @@ For details about managing the Minishift VM, see the link:../using/managing-mini
 The `minishift start` command creates an OpenShift instance using the
 https://github.com/openshift/origin/blob/master/docs/cluster_up_down.md[cluster up] approach.
 
-For this purpose it copies the `oc` binary onto your host. You will find it under
-`~/.minishift/cache/oc/{openshift-version}/oc` provided you
-use the default version of OpenShift. You can add this
-binary to your `PATH` variable in order to use `oc`, for example:
+For this purpose it copies the `oc` binary onto your host. You will find it under `~/.minishift/cache/oc/{openshift-version}/oc`, provided you use Minishift's default version of OpenShift.
+You can add this binary to your `PATH` using link:../command-ref/minishift_oc-env{outfilesuffix}[`minishift oc-env`] which displays the command you need to type into your shell.
+The output of `oc_env` will differ depending on OS and shell type.
```
budhrg: here it should be `oc-env`   not `oc_env`.

gbraad:
Thanks

Issue #618, opened 2017-03-29
Update the "Writing and Publishing documentation" topic


thatdocslady:
The following information is missing from this topic:

- Authoring guidelines: Should include basic topic structure, links to OpenShift docs guidelines, etc
- Deploying docs: How to trigger job to stage and tarball the docs after local testing, what happens on the OpenShift side, etc

thatdocslady:
TODO: Include instruction on updating topic map in case of adding new topics.

hferentschik:
Merged via pull request #1030.

PR #836, opened 2017-04-27
[do-not-merge] Issue #618 Update "writing and publishing docs" topic

thatdocslady:
Fixes issue #618 

This PR aims to formalize the documentation guidelines and conventions. I also proofread the topic itself to reflect the proposed conventions so you can see what a "clean" topic looks like.

TL;DR: Follow OpenShift doc guidelines whenever possible, exceptions noted.

I wrote some specific examples of common conventions that people might have questions about, but the basic guideline is to keep it simple and not overkill with tags, because we need the docs to be parsed through multiple toolchains upstream and downstream.

NOTE: We are currently working in RH CCS (content services) to unify the AsciiDoc conventions from several products into a single source-of-truth. This will impact Minishift but it will take some time and it will be partially based on OpenShift conventions so future changes will probably be minor.

Please review and hopefully we can reach an agreement, and after we finalize this I will go ahead and handle #691 so that the doc set follows the conventions.

### DIFF
--- a/docs/source/developing/writing-docs.adoc
+++ b/docs/source/developing/writing-docs.adoc
@@ -10,21 +10,77 @@ toc::[]
 [[writing-docs-overview]]
 == Overview
 
-This section contains information about contributing to the Minishift documentation.
-
-[[section-building-minishift-docs]]
-== Building Minishift documentation
-
 Minishift documentation is located in the `docs` sub-folder.
 The documentation is a mix of generated https://en.wikipedia.org/wiki/Markdown[Markdown] files and manually maintained https://en.wikipedia.org/wiki/AsciiDoc[Asciidoc] files.
 
-If you are adding a new file to the doumentation please update the `_topic_map.yml`  in the root of `docs/source` sub-folder.
+[[contribute-to-docs]]
+== Contributing to the documentation
 
-Per default the documentation is built in a Docker container. This way
+Minishift is an open-source project and we welcome contributions. Similar to code
+contributions, if you want to add or edit Minishift documentation, you can create an
+issue in our link:https://github.com/minishift/minishift/issues[Github issue tracker] and
+submit a pull request with the changes.
+
+[[docs-conventions-guidelines]]
+== Conventions and guidelines
+
+Minishift documentation is authored in the AsciiDoc markup format. Since Minishift is closely related to the OpenShift Origin
+project, we follow the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc[OpenShift Documentation Guidelines]
+for tagging, formatting, and structure wherever possible.
+
+The documentation source structure follows a modular format, where each file is called
+a _topic_ and each topic is stored under a sub-folder based on the relevant category, such as
+getting started, using Minishift, or command reference.
+
+TIP: You can check out the link:https://github.com/minishift/minishift/blob/master/docs/source/developing/writing-docs.adoc[raw file]
+of this topic or other existing Minishift topics to see examples of how the information is structured and which tags to use.
+
+[[common-conventions]]
+=== Commonly-used conventions
+
+- Each file must contain a topic metadata header with a top-level heading followed by required AsciiDoc variable declarations.
+The only exception to this is the command reference topics, which are automatically generated and do not require this header.
+You can view the raw file of this topic to see an example. Also, see
+the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#topic-metadata[topic metadata]
+section in the OpenShift documentation guidelines for more information.
+
+- Whenever you link to internal documentation topics, you should use relative links that can
+be resolved during the publication workflow instead of hard-coding URLs. See the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#internal-cross-references[internal cross-references]
+section of the OpenShift documentation guidelines for more information and examples.
+
+- Each section heading must be preceded by a unique anchor ID, to help make sure that
+cross-references resolve to the correct section. The anchor ID text should match the title
+as much as possible and it does not need to include the prefix _section_ or _topic_. See the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#unique-ids[unique IDs]
+section of the OpenShift documentation guidelines for more information and examples.
+
+[[adding-new-topic]]
+=== Adding a new topic
+
+If you are adding a new topic, or file, to the documentation library, you must also
+update the navigation in the following ways:
+
+- Add an entry to the relevant *_index.adoc_* file for that topic. For example, if you
+add a new topic in the *_docs/source/using/_* directory, you must update the *_docs/source/using/index.adoc_*
+file in the same directory.
+
+- Add an entry to the *__topic_map.yml_* in the *_docs/source_* sub-folder. The entry must
+contain the topic title and the file name.
+
+Make sure to add the new topic in the same order in all of the locations.
+
+[[building-docs]]
+== Building the documentation
+
+Minishift documentation is located in the *_docs_* sub-folder. The
+documentation is a mix of generated
+https://en.wikipedia.org/wiki/Markdown[Markdown] files and manually
+maintained https://en.wikipedia.org/wiki/AsciiDoc[Asciidoc] files.
+
+By default, the documentation is built in a Docker container. This way
 you avoid having to install all the required dependencies on your
 development machine. All you need is a running Docker daemon. In case
-you don't have one, use Minishift itself. For further information refer to
-link:../command-ref/minishift_docker-env{outfilesuffix}[`minishift docker-env`].
+you don't have one, use Minishift itself. For more information, see the
+link:../command-ref/minishift_docker-env{outfilesuffix}[`minishift docker-env`] command.
 
 To build the Docker image, run:
 
@@ -32,7 +88,7 @@ To build the Docker image, run:
 $ make build_docs_container
 ----
 
-To generate the documentation into the directory `docs/build`, run:
+To generate the documentation into the directory *_docs/build_*, run:
 
 ----
 $ make gen_docs
@@ -66,14 +122,25 @@ To build and serve the documentation for editing, run:
 $ make serve_docs
 ----
 
-The latter will start the https://middlemanapp.com[Middleman] server on port 4567.
-You can access the rendered documentation by browsing to http://<IP of Docker daemon>:4567.
+The `make serve_docs` command starts the link:https://middlemanapp.com[Middleman] server on
+port 4567. You can access the rendered documentation by browsing to _http://<IP-of-Docker-daemon>:4567_.
 
-[[section-deploying-minishift-docs]]
-== Deploying Minishift documentation
+If you encounter issues with the local staging of the documentation, you can run the following
+command to delete the *_build_* sub-folder and the variables file:
 
-The Minishift documentation is deployed on _docs.openshift.org_.
-To integrate with this site we deliver a tarball containing the Minishift asciidoc files as well as some link:http://www.asciibinder.org/[AsciiBinder] meta data.
+----
+$ make clean_docs
+----
+
+[[deploying-docs]]
+== Deploying the documentation
+
+The Minishift documentation is deployed on _docs.openshift.org_ under
+link:https://docs.openshift.org/latest/minishift[https://docs.openshift.org/latest/minishift].
+
+To integrate with the _docs.openshift.org_ we deliver a tarball containing
+the Minishift AsciiDoc files as well as some link:http://www.asciibinder.org/[AsciiBinder]
+meta data.
 
 This tarball can be built using:
 
@@ -81,24 +148,22 @@ This tarball can be built using:
 $ make gen_adoc_tar
 ----
 
-Once the build completes there will be a _minishift-adoc.tar_ in the _docs/build_ directory
-of your checkout.
+After the build completes there will be a *_minishift-adoc.tar_* file in the *_docs/build_* directory
+of your local repository.
 
-How this tarball is integrated into the docs.openshift.org site build is described in
-<<section-building-openshift-docs>>.
-
-[[section-building-openshift-docs]]
+[[building-openshift-docs]]
 === Building OpenShift documentation
 
-To start with you need a checkout of the link:https://github.com/openshift/openshift-docs.git[openshift-docs]
-GitHub repository. You need all the tooling to build openshift-docs. Refer to
-link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/tools_and_setup.adoc[Install and set up the tools and software] which is part of the openshift-docs repository.
+Before you start, you need to check out the link:https://github.com/openshift/openshift-docs.git[openshift-docs]
+GitHub repository. You need all the tooling to build openshift-docs. See the
+link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/tools_and_setup.adoc[Install and set up the tools and software] section,
+which is part of the _openshift-docs_ repository.
 
-Once you can successfully run `rake build` in your openshift-docs checkout, you can integrate the
+After you successfully run `rake build` in your local repository, you can integrate the
 Minishift documentation.
 
-Start with building the documentation tarball as described in <<section-deploying-minishift-docs>>,
-then follow these steps (assuming you are in the directory of your openshift-docs checkout):
+After you build the Minishift documentation tarball, follow these steps from the directory of
+your local openshift-docs repository:
 
 ----
 $ mkdir minishift
@@ -110,4 +175,7 @@ $ cd ..
 $ rake build
 ----
 
-If the build completes successfully, the site is available under _preview/openshift-origin/latest/welcome/index.html_.
+If the build completes successfully, the site is available under *_preview/openshift-origin/latest/welcome/index.html_*.
+
+OpenShift runs a nightly build in which they update the documentation. After you update the documentation,
+you must wait for the nightly build that runs at midnight EST daily to view the documentation on _docs.openshift.org_.
--- a/cmd/minishift/cmd/addon/addon_install.go
+++ b/cmd/minishift/cmd/addon/addon_install.go
@@ -21,7 +21,7 @@ import (
 
 	"strings"
 
-	"github.com/minishift/minishift/out/bindata"
+	"github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
 )
@@ -36,10 +36,9 @@ const (
 )
 
 var (
-	defaultAssets = []string{"anyuid", "admin-user", "xpaas"}
-	force         bool
-	enable        bool
-	defaults      bool
+	force    bool
+	enable   bool
+	defaults bool
 )
 
 var addonsInstallCmd = &cobra.Command{
@@ -59,8 +58,8 @@ func init() {
 func runInstallAddon(cmd *cobra.Command, args []string) {
 	addOnManager := GetAddOnManager()
 	if defaults {
-		unpackAddons(addOnManager.BaseDir())
-		fmt.Println(fmt.Sprintf("Default add-ons %s installed", strings.Join(defaultAssets, ", ")))
+		util.UnpackAddons(addOnManager.BaseDir())
+		fmt.Println(fmt.Sprintf("Default add-ons %s installed", strings.Join(util.DefaultAssets, ", ")))
 		return
 	}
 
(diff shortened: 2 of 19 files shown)


docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,149 @@ toc::[]
...
 $ make gen_adoc_tar
 ----
 
-Once the build completes there will be a _minishift-adoc.tar_ in the _docs/build_ directory
-of your checkout.
-
-How this tarball is integrated into the docs.openshift.org site build is described in
-<<section-building-openshift-docs>>.
+After the build completes there will be a *_minishift-adoc.tar_* file in the *_docs/build_* directory
+of your local repository.
```
thatdocslady: I'm not sure what your comment suggests here, I removed the internal xref because it points to exactly the next section and therefore unnecessary.

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,149 @@ toc::[]
...
+
+To update the navigation in the local staging, you must also add an entry to the *_layouts.erb_*
+file, located in the *_docs/source/layouts_* sub-folder.
+
+
+[[building-docs]]
+== Building the documentation
+
+Minishift documentation is located in the *_docs_* sub-folder. The
 documentation is a mix of generated
```
thatdocslady: the entire doc set with all the dependencies is located in the /docs/ directory, and there are relevant helper files in that root

Preeticp: Ok

docs/source/developing/writing-docs.adoc
Preeticp: I was talking about adding this introductory line to Building OpenShift Documentation section using the deleted sentence : "The following section describes the integration of the ......site

thatdocslady: I don't think this is necessary, but now that I look at the section again it does seem to be missing a "why", as in what is the use-case for building the openshift docs (when would you want to do it). I'll investigate.

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,156 @@ toc::[]
...
+how information is structured and which tags to use.
+
+[[common-conventions]]
+=== Commonly-used conventions
+
+- Each file must contain a mandatory header section with metadata that is required for publication.
+The only exception to this is the command reference topics, which are automatically generated and
+do not require this header.
+
+- Whenever we link to internal documentation topics, you should use relative links that can
```
hferentschik: One task of the doc review should be to review all internal links

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,156 @@ toc::[]
...
+be resolved during the publication workflow instead of hard-coding URLs.
+
+- Relative links must include breadcrumbs from the *_docs/source_* sub-folder to the topic,
```
hferentschik: Not sure what you are saying here. How are links relate to the breadcrumb?

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,156 @@ toc::[]
...
+to make sure that the generated URL is correct, even if you are linking from the same sub-folder.
+
+- Because code blocks must be parsed through multiple toolchains, you should avoid complex
```
hferentschik: What exactly is not supported? So we have basically two main use cases atm. Console listings and YAML/JSON config. The latter should imo be marked with [source,yaml|json]. Code formatting for these is in place and also used within the existing docs.openshift.org. 

Regarding console listings, I am in favour with marking them with [listing.console]. This makes things more explicit in the source and has not effect in the rendered docs. 

> Because code blocks must be parsed through multiple toolchains, you should avoid complex

what are the multiple toolchains and which concrete asciidoctor syntax should be avoided?

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,156 @@ toc::[]
...
+or specialized source type tagging to prevent doc builds breaking. Instead, use simple four
+dashes (*----*) before and after the code block.
+
+- Each section heading must be preceded by a unique anchor ID, to help make sure that
+cross-references resolve to the correct section. The anchor ID text should match the title
+as much as possible and it does not need to include the prefix _section_ or _topic_.
```
hferentschik: IMO it helps with a prefix, for example it makes more explicit in source whether you are linking to a section, example or table. So imo +1 for a prefix

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,156 @@ toc::[]
...
+
+[[known-issues-docs]]
+=== Known issues
+
+- Use only the *link:* tag for any internal, relative, or external links. The *xref:* tag
```
hferentschik: -1, internal links should use the <<>> syntax.

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,156 @@ toc::[]
...
+might cause issues with the local staging.
+
+- Currently we use the variables file to generate version numbers only. In the future we plan to expand
```
hferentschik: Why is this a "known issue"? Imo this is not relevant here and the first item can be part of the link discussion a bit higher up

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,156 @@ toc::[]
...
+update the navigation in the following ways:
+
+- Add an entry to the relevant *_index.adoc_* file for that topic. For example, if you
+add a new topic in the *_docs/source/using/_* directory, you must update the *_docs/source/using/index.adoc_*
+file in the same directory.
+
+- Add an entry to the *__topic_map.yml_* in the *_docs/source_* sub-folder. The entry must
+contain the topic title and the file name.
+
+- Add an entry to the *_layouts.erb_* file located in the *_docs/source/layouts_* sub-folder.
```
hferentschik: That step is obsolete

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,156 @@ toc::[]
...
+=== Adding a new topic
+
+If you are adding a new topic, or file, to the documentation library, you must also
+update the navigation in the following ways:
+
+- Add an entry to the relevant *_index.adoc_* file for that topic. For example, if you
+add a new topic in the *_docs/source/using/_* directory, you must update the *_docs/source/using/index.adoc_*
+file in the same directory.
+
+- Add an entry to the *__topic_map.yml_* in the *_docs/source_* sub-folder. The entry must
```
hferentschik: order in topic map and index.adoc should be in the same order

docs/source/developing/writing-docs.adoc
```diff
@@ -101,4 +172,7 @@ $ cd ..
 $ rake build
 ----
 
-If the build completes successfully, the site is available under _preview/openshift-origin/latest/welcome/index.html_.
+If the build completes successfully, the site is available under *_preview/openshift-origin/latest/welcome/index.html_*.
+
+OpenShift runs a nightly build in which they update the documentation. After you check-in the updated tarball,
```
hferentschik: "After you check-in the updated tarball,"? We are not checking in a tarball. I guess you are referring to http://artifacts.ci.centos.org/minishift/minishift/docs/latest/. This tarball is created on request by a CI job.

hferentschik: https://github.com/minishift/minishift/blame/master/docs/source/developing/releasing.adoc#L44

hferentschik: So the workflow is that the tarball gets created on request and then gets consumed with 24 h by doc.openshift.org

hferentschik:
I also suggest to adopt and include a "one sentence per line" approach. This makes reviews much easier and avoids unnecessary overflows.

thatdocslady:
@hferentschik I updated the PR with some of your suggestions, thanks for the feedback!

A few discussion points:

1. According to the OpenShift conventions, you need to use xref for internal links and link for external links. However, if you remember when we talked to them a while ago we mentioned that xref doesn't work for us so they said to use link, but the conventions don't mention the double angle-brackets <<>> and we shouldn't use something that isn't defined in the conventions. 

2. Related to point 1, I would actually avoid linking inside the same topic, because we use modular documentation so the topics are short enough, and if you click a link in a web page that goes to the same topic it might not work properly (i.e. if the topic is right after the link, the page won't actually skip and then it can confuse people.

3. The section/topic prefix is not necessary and isn't defined in the conventions, and also in general it's better not to use this in case we need to move sections around. For example, if a section becomes too big and then we need to make it a separate topic, then the anchor ID prefix must be changed from section to topic, and that can break links. This is tech writing 101 and it's better to name the anchor IDs according to just the content instead of prefixing their location in the document.

4. The multiple toolchains are Middleman for our local staging, AsciiBinder for the upstream site, and ccutil/Docs 2.0 for the downstream site (RH customer portal). Like I mentioned before several times, even though OpenShift sometimes uses additional tags for code blocks, it's not defined in their conventions (and could actually be a problem that they're trying to address, for all we know). Better to keep the tagging simple or if you want to implement additional tags we should consult them if it's possible or how it will impact things.

5. The breadcrumb for the links is explained in the OpenShift documentation guidelines, I added the link to that section so you can see it and actually removed the text because people can read more about it in the guidelines.

6. I wouldn't restrict content to be one sentence per paragraph, because we try to implement short sentences so sometimes it's ok to have two sentences if they're not long. 

cc @LalatenduMohanty @Preeticp

hferentschik:
> but the conventions don't mention the double angle-brackets <<>> and we shouldn't use something that isn't defined in the conventions.

<<>> is the idiomatic internal reference style of asciidoctor. They are more concise than link tags. Also they are more readable, especially when looking at the outputsuffix we seem to append.  Effectively <<>> are xrefs. 

> if you remember when we talked to them a while ago we mentioned that xref doesn't work for us

I have no idea why xref would not work for us. In fact we are already using them at several places. Also to quote from the guidelines https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc:

    * all links to internal topics are created using xref and must have an anchor ID.
    * all links to external websites are created using link.

In the guidelines they use 'xref', but that's effectively the same as <<>>. 

@thatdocslady, so what would be the concrete reason for not using xref. Is there somewhere an open issue?

hferentschik:
> Related to point 1, I would actually avoid linking inside the same topic, because we use modular documentation so the topics are short enough, 

That's not a given and even in a shorter document it makes sometimes sense to cross reference. Does not have to be a section, but also could be for an example or table.

hferentschik:
> The section/topic prefix is not necessary and isn't defined in the conventions, 

IMO it makes sense and I made good experiences with it. However, I am happy to skip it.

> For example, if a section becomes too big and then we need to make it a separate topic, then the anchor ID prefix must be changed from section to topic, and that can break links. 

It will break links anyways, regardless of the prefix. You are moving into a new document. If you have any cross reference to this section you will need to adjust it.

> This is tech writing 101 

Do you have a reference?

hferentschik:
> The multiple toolchains are Middleman for our local staging, AsciiBinder for the upstream site, and ccutil/Docs 2.0 for the downstream site (RH customer portal).

But it is all AsciiDoc(tor) right?

> Like I mentioned before several times, even though OpenShift sometimes uses additional tags for code blocks, it's not defined in their conventions

So maybe their conventions are out of date? Clearly some blocks have a explicit source tag which makes sense imo. 

> (and could actually be a problem that they're trying to address, for all we know). 

Then we should find out.

>  or if you want to implement additional tags we should consult them if it's possible or how it will impact things.

I don't want to use anything which is not supported by Asciidoctor. To be clear adding [listing] does not change a thing. Not using it is just a shortcut. [listing.console] is different. This will add another css style attribute to the output. It does not matter atm, but I can see that this is controversial.

hferentschik:
> The breadcrumb for the links is explained in the OpenShift documentation guidelines

Which section exactly are you referring to?

hferentschik:
> I wouldn't restrict content to be one sentence per paragraph, because we try to implement short sentences so sometimes it's ok to have two sentences if they're not long.

This is vary seldom the case. One sentence a line makes a lot of things easier when it comes to maintaining the documentation. Just needs a bit to get used to. See also  https://developer.atlassian.com/blog/2015/11/totw-write-better-docs-and-blogs/

hferentschik:
> if you remember when we talked to them a while ago we mentioned that xref doesn't work for us

the only thing  I can think about are references into docs.openshift.org outside of Minisfhift. There xref is problematic for our local build, but I think it will still work when the overall docs are build. It is just harder for use to verify.

What we are talking about are cross referencing within our own documentation and there xref resp. <<>> should be used.

thatdocslady:
Status: All feedback implemented except a few questions, there is an issue in openshift-docs repo to address these https://github.com/openshift/openshift-docs/issues/4500

docs/source/developing/writing-docs.adoc
```diff
@@ -12,85 +12,156 @@ toc::[]
...
+a _topic_ and each topic is stored under a sub-folder based on the relevant category, such as
+getting started, using Minishift, or command reference.
+
+TIP: You can check out the raw file for this topic or other existing Minishift topics to see
+how information is structured and which tags to use.
+
+[[common-conventions]]
+=== Commonly-used conventions
+
+- Each file must contain a mandatory header section with metadata that is required for publication.
```
hferentschik: I think this still needs clarification. What is mandatory? Or we relax the sentence to:

"Each file starts with a level one heading, followed by any necessary asciidoc variable declarations."

docs/source/developing/writing-docs.adoc
```diff
@@ -10,29 +10,93 @@ toc::[]
...
+getting started, using Minishift, or command reference.
+
+TIP: You can check out the link:https://github.com/minishift/minishift/blob/master/docs/source/developing/writing-docs.adoc[raw file]
+of this topic or other existing Minishift topics to see examples of how the information is structured and which tags to use.
+
+[[common-conventions]]
+=== Commonly-used conventions
+
+- Each file must contain a mandatory header section with metadata that is required for publication.
+The only exception to this is the command reference topics, which are automatically generated and
```
hferentschik: I think we can skip these two sentences

docs/source/developing/writing-docs.adoc
```diff
@@ -10,29 +10,93 @@ toc::[]
...
+do not require this header. See the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#topic-metadata[topic metadata]
+section in the OpenShift documentation guidelines for more information.
+
+- Whenever we link to internal documentation topics, you should use relative links that can
```
hferentschik: Whenever _we_ ...., _you_ should ...?

we or you?

hferentschik: The key is to use xref, not that we are using relative links. The links will be relative automatically using xref

docs/source/developing/writing-docs.adoc
```diff
@@ -10,29 +10,93 @@ toc::[]
...
+be resolved during the publication workflow instead of hard-coding URLs. See the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#internal-cross-references[internal cross-references]
+section of the OpenShift documentation guidelines for more information and examples.
+
+- Because code blocks must be parsed through multiple toolchains, you should avoid complex
```
hferentschik: I think this paragraph is out of date - https://github.com/openshift/openshift-docs/issues/4500#issuecomment-305021585

Source formatting is ok

docs/source/developing/writing-docs.adoc
```diff
@@ -10,29 +10,93 @@ toc::[]
...
+or specialized source type tagging to prevent doc builds breaking. Instead, use simple four
+dashes (*----*) before and after the code block. See the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#formatting[Formatting]
+section of the OpenShift documentation guidelines for more information and examples.
+
+- Each section heading must be preceded by a unique anchor ID, to help make sure that
+cross-references resolve to the correct section. The anchor ID text should match the title
+as much as possible and it does not need to include the prefix _section_ or _topic_. See the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#unique-ids[unique IDs]
+section of the OpenShift documentation guidelines for more information and examples.
+
+- We use the variables file to generate version numbers. In the future we plan to expand
```
hferentschik: I don't think this is needed in this context.

docs/source/developing/writing-docs.adoc
```diff
@@ -10,29 +10,93 @@ toc::[]
...
+[[building-docs]]
+== Building the documentation
+
+Minishift documentation is located in the *_docs_* sub-folder. The
+documentation is a mix of generated
+https://en.wikipedia.org/wiki/Markdown[Markdown] files and manually
+maintained https://en.wikipedia.org/wiki/AsciiDoc[Asciidoc] files.
+
+By default, the documentation is built in a Docker container. This way
+>>>>>>> Issue #618 update writing and publishing docs topic
```
hferentschik: This looks like an unresolved merge conflict

docs/source/developing/writing-docs.adoc
```diff
@@ -110,4 +183,7 @@ $ cd ..
 $ rake build
 ----
 
-If the build completes successfully, the site is available under _preview/openshift-origin/latest/welcome/index.html_.
+If the build completes successfully, the site is available under *_preview/openshift-origin/latest/welcome/index.html_*.
+
+OpenShift runs a nightly build in which they update the documentation. After you update the documentation,
```
hferentschik: That's not correct. This has nothing to do with local docs development. This is a production/release thing.

docs/source/developing/writing-docs.adoc
thatdocslady: I don't think so, these guidelines were confirmed by Harrison and Jake from OpenShift and they are important enough to justify listing them.

docs/source/developing/writing-docs.adoc
thatdocslady: I mentioned the relative links because some people tend to copy and paste the published URL of a certain topic instead of the relative link to the raw file.

docs/source/developing/writing-docs.adoc
thatdocslady: This information was already in the topic, and the topic itself is titled "writing and publishing docs" so it can include both topics, unless you have a new suggestion on how to organize the content?

hferentschik: Sure, the problem is that it is plain wrong. It implies that you if you "update the documentation", somehow it magically get picked up by a nightly update. Neither is clear what "update the documentation" actually means, nor does it touch on the CentOS CI job which is the actual trigger for updating the docs on doc.openshift.org.

hferentschik: I'll create a pull request for your branch ...

hferentschik:
@thatdocslady See https://github.com/hferentschik/minishift/commit/f133653081ebd09937b79da4eeea8a807d825da6. this in particular makes changes to the deploying section.

@thatdocslady If you rebase your pull request on master, I can create a pull request against your feature branch.

thatdocslady:
@hferentschik I rebased again and pushed to the PR but I'm still seeing the "rebase and merge" button (disabled), what does this mean?

docs/source/developing/writing-docs.adoc
thatdocslady: Ok like I said, I didn't actually add this information so it was wrong before and I'll be happy to update it so that it's accurate. I was wondering myself how the missing piece of the process comes into play, so it's good that we're fixing this now.

hferentschik:
>  I rebased again and pushed to the PR

Worked fine. Here is the pull request I was talking about https://github.com/thatdocslady/minishift/pull/2

hferentschik:
> but I'm still seeing the "rebase and merge" button (disabled), what does this mean?

Well, right now there are still outstanding change request.

docs/source/developing/writing-docs.adoc
```diff
@@ -10,29 +10,85 @@ toc::[]
...
+
+- Each file must contain a topic metadata header with a top-level heading followed by required AsciiDoc variable declarations.
+The only exception to this is the command reference topics, which are automatically generated and do not require this header.
+You can view the raw file of this topic to see an example. Also, see
+the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#topic-metadata[topic metadata]
+section in the OpenShift documentation guidelines for more information.
+
+- Whenever you link to internal documentation topics, you should use relative links that can
+be resolved during the publication workflow instead of hard-coding URLs. See the link:https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#internal-cross-references[internal cross-references]
+section of the OpenShift documentation guidelines for more information and examples.
```
LalatenduMohanty: As @hferentschik mentioned in his PR to your feature branch. We should just use xref for all cross referencing

docs/source/developing/writing-docs.adoc
hferentschik: Have you guys looked at my pull request. I Changed this section quite a bit to make it clear what this is actually about.

thatdocslady:
Superseded by PR #1030

Issue #840, opened 2017-04-27
Failing master builds on CentOS CI


hferentschik:
https://ci.centos.org/job/minishift/121/console

The issues seems to be related to the docs build.

budhrg:
Failing locally as well.

hferentschik:
Merged via pull request #844

PR #844, opened 2017-04-28
Issue #840 Making sure Rake clean target does work even if files don'…

hferentschik:
…t exist

- Also allowing to specify a custom UID during build time
- Documenting providing custom UID

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -174,7 +174,7 @@ function artifacts_upload_on_pr_and_master_trigger() {
   # For PR build, GIT_BRANCH is set to branch name other than origin/master
   if [[ "$GIT_BRANCH" = "origin/master" ]]; then
     install_docs_prerequisite_packages;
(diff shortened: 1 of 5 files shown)


hferentschik:
@budhrg,  I updated the Dockerfile to make sure that a proper locale is set. Turns out the base image does not do that.

centos_ci.sh
```diff
@@ -222,13 +222,14 @@ else
...
-    make gen_adoc_tar
+    make gen_adoc_tar IMAGE_UID=$(id -u)
     docs_tar_upload $PASS
   else
     setup_kvm_docker_machine_driver;
     prepare;
     cd $GOPATH/src/github.com/minishift/minishift
     run_tests;
     artifacts_upload_on_pr_and_master_trigger $PASS;
+    make gen_adoc_tar IMAGE_UID=$(id -u)
```
budhrg: Extra line here

Issue #856, opened 2017-05-03
Pods unable to write to persistent volumes 


hferentschik:
Pods which make a persistent volume claim, cannot write to their assigned persistent volume.

Internally we rely on the 100 persistent volumes created by _cluster up_. In our case we set _host-pv_dir_ to _/var/lib/minishift/openshift.local.pv_. After provisioning is complete the created persistent volume directories have the following permissions:

    $ minishift ssh -- "sudo ls -la /var/lib/minishift/openshift.local.pv"
    drwxrwx---    2 root     root          4096 May  3 11:31 pv0001
    ...

When a persistent volume claim is made, the pod then cannot write to this directory (tested with MariaDB, MongoDB, Jenkins, etc).

One can get around this problem by running the following command after Minishfift started:

    $ minishift ssh -- "sudo chmod -R o+rwx /var/lib/minishift/openshift.local.pv/pv*"

Not sure what the proper permissions should be, but this works.

See also https://issues.jboss.org/projects/CDK/issues/CDK-120

hferentschik:
Not sure whether this is a 'cluster up' issue only or whether there is something else we should do on the Minishift side when using _host-pv_dir_. 

Not sure how to work around this for now either, since the persistent volumes are created in the background. We cannot just issue a _chmod_ command after _cluster up_ returns.

hferentschik:
@csrwng any insight on this? Is this a 'cluster up' issue or is there something we are missing or should be doing on the Minishift side?

Also, any idea on how we could at least temporarily work around this?

budhrg:
I was trying `oc cluster up` to see if it issue with it or not but seems like my setup is having some issue.
Keep getting message even setting the insecure registry in  `/etc/sysconfig/docker`
```
$ oc cluster up --host-pv-dir /var/lib/minishift/openshift.local.pv
-- Checking OpenShift client ... OK
-- Checking Docker client ... OK
-- Checking Docker version ... 
   WARNING: Cannot verify Docker version
-- Checking for existing OpenShift container ... OK
-- Checking for openshift/origin:v1.5.0 image ... OK
-- Checking Docker daemon configuration ... FAIL
   Error: did not detect an --insecure-registry argument on the Docker daemon
   Solution:

     Ensure that the Docker daemon is running with the following argument:
     	--insecure-registry 172.30.0.0/16

```

csrwng:
[...]

PR #859, opened 2017-05-04
Issue 856 Introduce method to port configure persistent volumes

hferentschik:
Addresses issue #856

- Once persistent volumes are created global permissions will be set to allow pods to write to persistent volume
- Silenced oc stdout for more consistent output diring bootstrap
- Adding echo statements to anyuid add-on to make it more explicit what the add-on does

### DIFF
--- a/addons/anyuid/anyuid.addon
+++ b/addons/anyuid/anyuid.addon
@@ -1,5 +1,10 @@
 # Name: anyuid
-# Description: Allows authenticated users to run images under a non pre-allocated UID
-# Url: https://access.redhat.com/documentation/en-us/openshift_enterprise/3.2/html/cluster_administration/admin-guide-manage-scc
+# Description: Changes the default security context constraints to allow pods to run as any user.
+# Url: https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-origin-specific-guidelines
 
 oc adm policy add-scc-to-group anyuid system:authenticated
+
+echo  Add-on '#{addon-name}' changed the default security context constraints to allow pods to run as any user.
+echo  Per default OpenShift runs containers using an arbitrarily assigned user ID.
+echo  Refer to https://docs.openshift.org/latest/architecture/additional_concepts/authorization.html#security-context-constraints and
+echo  https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-origin-specific-guidelines for more information.
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -23,6 +23,7 @@ import (
 
 	"strings"
 
+	"bytes"
 	units "github.com/docker/go-units"
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/drivers"
@@ -34,6 +35,7 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minikube/kubeconfig"
+	"github.com/minishift/minishift/pkg/minishift/addon/manager"
 	"github.com/minishift/minishift/pkg/minishift/cache"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
@@ -44,9 +46,11 @@ import (
 	"github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/minishift/minishift/pkg/version"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
+	"time"
 )
 
 const (
@@ -185,7 +189,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	//Create the host directories if not present
-	hostDirs := []string{viper.GetString(hostConfigDir), viper.GetString(hostDataDir), viper.GetString(hostVolumesDir)}
+	hostDirs := []string{viper.GetString(hostConfigDir), viper.GetString(hostDataDir), viper.GetString(hostVolumesDir), viper.GetString(hostPvDir)}
 	err = clusterup.EnsureHostDirectoriesExist(libMachineClient, hostDirs)
 	if err != nil {
 		fmt.Println("Error creating required host directories: ", err)
@@ -231,16 +235,76 @@ func postClusterUp(machineName string, ip string, port int, routingSuffix string
 		atexit.Exit(1)
 	}
 
-	applyAddOns(ip, routingSuffix, ocPath, kubeConfigPath, sshCommander)
+	addOnManager := addon.GetAddOnManager()
+	configurePersistentVolumes(addOnManager, sshCommander, ocRunner)
+	applyAddOns(addOnManager, ip, routingSuffix, ocPath, kubeConfigPath, sshCommander)
 }
 
-func applyAddOns(ip string, routingSuffix string, ocPath string, kubeConfigPath string, sshCommander provision.SSHCommander) {
-	addOnManager := addon.GetAddOnManager()
+func applyAddOns(addOnManager *manager.AddOnManager, ip string, routingSuffix string, ocPath string, kubeConfigPath string, sshCommander provision.SSHCommander) {
 	err := addOnManager.Apply(addon.GetExecutionContext(ip, routingSuffix, ocPath, kubeConfigPath, sshCommander))
 	if err != nil {
 		fmt.Println("Error executing addon commands: ", err)
 		atexit.Exit(1)
 	}
+}
+
+// TODO - persistent volume creation should really be fixed upstream, aka 'cluster up'. See https://github.com/openshift/origin/issues/14076 (HF)
+// configurePersistentVolumes makes sure that the default persistent volumes created by 'cluster up' have the right permissions - see https://github.com/minishift/minishift/issues/856
+func configurePersistentVolumes(addOnManager *manager.AddOnManager, sshCommander provision.SSHCommander, ocRunner *oc.OcRunner) error {
+	// don't apply this if anyuid is not enabled
+	anyuid := addOnManager.Get("anyuid")
+	if anyuid == nil || !anyuid.IsEnabled() {
+		return nil
+	}
+
+	fmt.Print("-- Waiting for persistent volumes to be created ... ")
+
+	hostPvDir := viper.GetString(hostPvDir)
+
+	var out, err *bytes.Buffer
+
+	// poll the status of the persistent-volume-setup job to determine when the persitent volume creates is completed
+	for {
+		out = new(bytes.Buffer)
+		err = new(bytes.Buffer)
+		exitStatus := ocRunner.Run("get job persistent-volume-setup -n default -o 'jsonpath={ .status.active }'", out, err)
+
+		if exitStatus != 0 || len(err.String()) > 0 {
+			return errors.New("Unable to monitor persistent volume creation")
+		}
+
+		if out.String() != "1" {
+			break
+		}
+
+		time.Sleep(1 * time.Second)
+	}
+
+	// verify the job succeeded
+	out = new(bytes.Buffer)
+	err = new(bytes.Buffer)
+	exitStatus := ocRunner.Run("get job persistent-volume-setup -n default -o 'jsonpath={ .status.succeeded }'", out, err)
+
+	if exitStatus != 0 || len(err.String()) > 0 || out.String() != "1" {
+		return errors.New("Persistent volume creation failed")
+	}
+
+	cmd := fmt.Sprintf("sudo chmod -R 777 %s/pv*", hostPvDir)
+	sshCommander.SSHCommand(cmd)
+
+	// if we have SELinux enabled we need to sort things out there as well
+	// 'cluster up' does this as well, but we do it here as well to have all required actions collected in one
+	// place, instead of relying on some implicit knowledge on what 'cluster up does (HF)
+	cmd = fmt.Sprintf("sudo which chcon; if [ $? -eq 0 ]; then chcon -R -t svirt_sandbox_file_t %s/*; fi", hostPvDir)
+	sshCommander.SSHCommand(cmd)
+
+	cmd = fmt.Sprintf("sudo which restorecon; if [ $? -eq 0 ]; then restorecon -R %s; fi", hostPvDir)
+	sshCommander.SSHCommand(cmd)
+
+	fmt.Println("OK")
+	fmt.Println()
+
+	return nil
 }
 
 func automountHostfolders(driver drivers.Driver) {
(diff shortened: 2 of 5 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -243,6 +244,30 @@ func applyAddOns(ip string, routingSuffix string, ocPath string, kubeConfigPath
...
 
+// TODO - persistent volume creation should really be fixed upstream, aka 'cluster up' (HF)
+// configurePeristentVolumes makes sure that the default persistent volumes created by 'cluster up' have the right permissions - see https://github.com/minishift/minishift/issues/856
+func configurePeristentVolumes(sshCommander provision.SSHCommander) {
+	fmt.Println("-- Waiting for persistent volumes to be ready")
+
+	hostPvDir := viper.GetString(hostPvDir)
+
+	// there are in total 101 directories when using ls. There is an additonal 'registry' directory
+	cmd := fmt.Sprintf("while [ $(ls %s | wc -l) -lt 101 ] ; do sleep 1; done", hostPvDir)
```
hferentschik: This is really the gist of things. We wait until the persistent volumes are created. We could easily create a loop in Go as well and combine it with some timeout, but the idea would stay the same. Wait until the persistent volumes are created and then modify the permissions.

hferentschik: BTW, it seems when 'cluster up' is used on a native Docker daemon, the problem does not exist. It seems to be an issue related to the fact that we are running in a VM. So I would feel a bit better, if I'd understand why that is.

@csrwng, any idea?

test/integration/features/basic.feature
```diff
@@ -44,15 +44,14 @@ Feature: Basic
...
-#  Test disabled - see https://github.com/minishift/minishift/issues/837
-#  Scenario: User has a pre-configured set of persitence volumnes
-#    When executing "oc get pv --as system:admin -o=name" retrying 5 times with wait period of 3 seconds
-#    Then stderr should be empty
-#     And exitcode should equal 0
-#     And stdout should contain
-#     """
-#     persistentvolume/pv0001
-#     """
+  Scenario: User has a pre-configured set of persitence volumnes
```
hferentschik: At least we could enable this test again ;-)

cmd/minishift/cmd/start.go
csrwng: @hferentschik not at the moment. I will need to look at things a bit closer when I get back from summit.

LalatenduMohanty:
@hferentschik the PR looks fine to me. However I would wait till we know why `cluster up` does not have the issue in some cases.

hferentschik:
>  However I would wait till we know why cluster up does not have the issue in some cases.

Sure, I am all up for better understanding this. Just want to make sure we have a contingency plan.

budhrg:
retest this please

hferentschik:
> However I would wait till we know why cluster up does not have the issue in some cases

@LalatenduMohanty, seems we have all the information we need now. IMO for now this pull request offers the best trade-off. Biggest drawback is the added bootstrap time.

gbraad:
Unable to review ?

addons/anyuid/anyuid.addon
```diff
@@ -1,5 +1,10 @@
 # Name: anyuid
-# Description: Allows authenticated users to run images under a non pre-allocated UID
-# Url: https://access.redhat.com/documentation/en-us/openshift_enterprise/3.2/html/cluster_administration/admin-guide-manage-scc
+# Description: Changes the default security context constraints to allow pods to run as any user.
+# Url: https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-origin-specific-guidelines
 
 oc adm policy add-scc-to-group anyuid system:authenticated
+
+echo  Add-on '#{addon-name}' changed the default security context constraints to allow pods to run as any user.
```
gbraad: +1

cmd/minishift/cmd/start.go
```diff
@@ -243,6 +244,31 @@ func applyAddOns(ip string, routingSuffix string, ocPath string, kubeConfigPath
...
 }
 
+// TODO - persistent volume creation should really be fixed upstream, aka 'cluster up'. See https://github.com/openshift/origin/issues/14076 (HF)
+// configurePeristentVolumes makes sure that the default persistent volumes created by 'cluster up' have the right permissions - see https://github.com/minishift/minishift/issues/856
+func configurePeristentVolumes(sshCommander provision.SSHCommander) {
+	fmt.Print("-- Waiting for persistent volumes to be created ... ")
+
+	hostPvDir := viper.GetString(hostPvDir)
+
+	// there are in total 101 directories when using ls. There is an additonal 'registry' directory
```
gbraad: -1 (at the moment we assume 101 folders), this does not work as expected when something has changed in the way openshift works.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -168,7 +168,7 @@ func (m *AddOnManager) String() string {
 }
 
 func (m *AddOnManager) applyAddOn(addOn addon.AddOn, context *command.ExecutionContext) error {
-	fmt.Print(fmt.Sprintf("Applying addon %s:", addOn.MetaData().Name()))
+	fmt.Print(fmt.Sprintf("-- Applying addon '%s':", addOn.MetaData().Name()))
```
gbraad: +1

gbraad: How about the first line of the description to show the Short Description, as Name might not be clear enough.

cmd/minishift/cmd/start.go
hferentschik: There is no other good option. There are several issues open for cluster up which will improve on this. In particular the ability to control the number of pas created as well as not creating them in the background. Once this is available we will adjust this code.

pkg/minishift/addon/manager/addon_manager.go
hferentschik: I think it is important to use the name. The name is the key for the various add-on commands. Also the description line can be long. 

Better to use the echo command to provide more context.

cmd/minishift/cmd/start.go
hferentschik: See also the TODO at the top of the method. We will adjust this as soon as 'cluster up' provides us with additional ways to control this. ATM, I don't see a way to make this dynamic

pkg/minishift/addon/manager/addon_manager.go
gbraad: At the moment there is no guarantee that the `Name'` is unique (however, the folder names always are). Is it possible to at least include this? I am sure this will prevent issue hunting later, as the addons enabling does not happen on the 'name', but folder... right?

cmd/minishift/cmd/start.go
```diff
@@ -231,6 +231,7 @@ func postClusterUp(machineName string, ip string, port int, routingSuffix string
 		atexit.Exit(1)
 	}
 
+	configurePeristentVolumes(sshCommander)
```
LalatenduMohanty: Because of this the `minishift start` is taking more time now. In my test it is almost taking a minute more i.e. 4m24.133s (vs 3m38.676s  without this PR). Can we call this only when anyuid is enabled?

LalatenduMohanty: I think it is time to raise an issue with cluster up to create only 10 directories and provide a flag to allow users to create more if user want. @hferentschik you might have raised an issue for this. Nor sure.

gbraad: this was my previous concern (as we now handle 100 directories), but this is an arbitrary number. It could even be this is just 10... and rather JUST ten. But is this configurable from the command line for `cluster up`?

hferentschik: > But is this configurable from the command line for cluster up

No, that's the point. It is hard coded in cluster up.

cmd/minishift/cmd/start.go
gbraad: See: https://github.com/openshift/origin/issues/14076 (as referenced in the TODO)

cmd/minishift/cmd/start.go
hferentschik: https://github.com/openshift/origin/issues/14061

hferentschik:
> you might have raised an issue for this. Nor sure.

I did, check the discussion on the issue

budhrg:
Working for me with binary compiled from this PR 
```
$ minishift start --memory 4096
Downloading OpenShift binary 'oc' version 'v1.5.0'
 19.96 MB / 19.96 MB [============================================================================================================================] 100.00% 0s
....

   To login as administrator:
       oc login -u system:admin

-- Waiting for persistent volumes to be created ... OK

-- Applying addon 'anyuid':.
   Add-on 'anyuid' changed the default security context constraints to allow pods to run as any user.
   Per default OpenShift runs containers using an arbitrarily assigned user ID.
   Refer to https://docs.openshift.org/latest/architecture/additional_concepts/authorization.html#security-context-constraints and
   https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-origin-specific-guidelines for more information.

$ minishift console
Opening the OpenShift Web console in the default browser...
``` 
screenshot from 2017-05-08 15-19-39

hferentschik:
@LalatenduMohanty I pushed an update which will not make the directory permissions changes when anyuid is not enabled. This helps on the community side of things until we might decide to enable _anyuid_ per default there as well. Please test.

LalatenduMohanty:
@hferentschik Thanks. Testing the PR.

hferentschik:
@LalatenduMohanty Also removed the commit related to the integration test again. Now with the optimisation in place it will fail again. I think better to take this separately, since I want to make some other changes to the test as well.

cmd/minishift/cmd/start.go
csrwng: why not set a watch on the pv dir job and wait for it to complete?

pkg/minishift/addon/parser/addon_parser_test.go
```diff
@@ -94,7 +94,7 @@ func Test_successful_parsing_of_addon_dir(t *testing.T) {
 		t.Fatalf("Unexpected addon name: '%s'. Expected '%s'", addOn.MetaData().Name(), expectedName)
 	}
 
-	expectedNumberOfCommands := 1
```
LalatenduMohanty: Interesting. Not sure how it used to pass before.

cmd/minishift/cmd/start.go
LalatenduMohanty: @csrwng you mean  we have separate process for pv dirs creation?

hferentschik: > why not set a watch on the pv dir job and wait for it to complete?

interesting, what "job" are we talking about? Is this a scheduled OpenShift job? If so, that would of course be an option. What _oc_ command is best to way to do this? _observe_?

csrwng: @LalatenduMohanty @hferentschik yes we do start a job. https://github.com/openshift/origin/blob/master/pkg/bootstrap/docker/openshift/pvsetup.go#L98

named https://github.com/openshift/origin/blob/master/pkg/bootstrap/docker/openshift/pvsetup.go#L19

and yes, I would use `oc observe` to watch for changes to it, if it is modified, then check its status and finish when the job is moved to complete status.

cmd/minishift/cmd/start.go
```diff
@@ -231,18 +235,76 @@ func postClusterUp(machineName string, ip string, port int, routingSuffix string
...
+	}
+
+	fmt.Print("-- Waiting for persistent volumes to be created ... ")
+
+	hostPvDir := viper.GetString(hostPvDir)
+
+	var out, err *bytes.Buffer
+
+	// pull the status of the persistent-volume-setup job to determine when the persitent volume creates is completed
+	for {
```
hferentschik: @csrwng @LalatenduMohanty @budhrg - here is the updated version of _ configurePersistentVolumes_, which does not rely on knowing the number of persistent volumes to create. Instead it polls the OpenShift job which creates these directories. 

This is obviously a bit neater, the bottom line stays the same.

cmd/minishift/cmd/start.go
```diff
@@ -231,18 +235,76 @@ func postClusterUp(machineName string, ip string, port int, routingSuffix string
...
+		return nil
+	}
+
+	fmt.Print("-- Waiting for persistent volumes to be created ... ")
+
+	hostPvDir := viper.GetString(hostPvDir)
+
+	var out, err *bytes.Buffer
+
+	// pull the status of the persistent-volume-setup job to determine when the persitent volume creates is completed
```
csrwng: s/pull/poll/

cmd/minishift/cmd/start.go
```diff
@@ -231,18 +235,76 @@ func postClusterUp(machineName string, ip string, port int, routingSuffix string
...
+	for {
+		out = new(bytes.Buffer)
+		err = new(bytes.Buffer)
+		exitStatus := ocRunner.Run("get job persistent-volume-setup -n default -o 'jsonpath={ .status.active }'", out, err)
+
+		if exitStatus != 0 || len(err.String()) > 0 {
+			return errors.New("Unable to monitor persistent volume creation")
+		}
+
+		if out.String() != "1" {
```
csrwng: Shouldn't this check if out.String() == "1" ?

csrwng: ahh nevermind, I read below :) please ignore

cmd/minishift/cmd/start.go
```diff
@@ -231,18 +235,76 @@ func postClusterUp(machineName string, ip string, port int, routingSuffix string
...
+
+	if exitStatus != 0 || len(err.String()) > 0 || out.String() != "1" {
+		return errors.New("Persistent volume creation failed")
+	}
+
+	cmd := fmt.Sprintf("sudo chmod -R 777 %s/pv*", hostPvDir)
+	sshCommander.SSHCommand(cmd)
+
+	// if we have SELinux enabled we need to sort things out there as well
+	cmd = fmt.Sprintf("sudo which chcon; if [ $? -eq 0 ]; then chcon -R -t svirt_sandbox_file_t %s/*; fi", hostPvDir)
```
csrwng: the create pv job already does this ... 
https://github.com/openshift/origin/blob/master/pkg/bootstrap/docker/openshift/pvsetup.go#L56

cmd/minishift/cmd/start.go

cmd/minishift/cmd/start.go
hferentschik: and this does not need to be re-run after modification of he permissions?

csrwng: afaik no, but I guess it doesn't hurt

hferentschik: my thinking as well :-)

hferentschik: @csrwng thanks for your eyes on this btw. Much appreciated. I

csrwng:
LGTM

hferentschik:
Tested with Boot2Docker and CentOS ISO w/ and w/o anyuid enabled. Works as expected.

gbraad:
@LalatenduMohanty please review again. I believe your comments have been addressed.

cmd/minishift/cmd/start.go
LalatenduMohanty: @hferentschik @csrwng If we do not need to set SELinux context because `cluster up` does it then we should  remove this as this would mean we are expecting `cluster up` to set the proper context which is the right thing to do IMO. It would also make this function's function much simpler.

SvenVD:
Experimenting with storage was my main motivator

praveenkumar:
> Experimenting with storage was my main motivator

@SvenVD it depends about how you are creating the PV and what permission you are providing to it, can you put all the steps here?

praveenkumar:
@SvenVD  I think we are talking on a closed PR, the best way to resolve any permission issue to create a new issue along with all the steps and we all can discuss it from there.

Issue #820, opened 2017-04-25
Extending go-dog execute step with 'succeed'


hferentschik:
We are having a lot of tests of the form:

    When executing "oc config current-context"
    Then stderr should be empty
     And exitcode should equal 0
     And stdout should contain
    """
    minishift
    """"

There is a lot of reception and the feature file gets unnecessarily long. 

We should introduce something like:

    When executing "oc config current-context" succeeds 
    Then stdout should contain
    """
    minishift
    """"

The 'succeeds' parameter in the step, will ensure that internally the command has an exit code of 0 and no output to standard out. This way we can reduce a lot of repetition in the feature file.

budhrg:
> The 'succeeds' parameter in the step, will ensure that internally the command has an exit code of 0 and no output to standard out. This way we can reduce a lot of repetition in the feature file.

:+1:   Good one , we did something similar for service-manager 
`When I successfully run <cmd>`

https://github.com/projectatomic/vagrant-service-manager/blob/master/features/cdk-openshift.feature#L29

hferentschik:
> Good one , we did something similar for service-manager

exactly

hferentschik:
Merged via pull request #864

PR #864, opened 2017-05-08
Issue #820: Extending go-dog execute step with 'succeed'

amitkrout:
Added steps with "succeed" to minimize the duplication of scenario steps. 

Addresses issue #820

### DIFF
--- a/test/integration/basic_test.go
+++ b/test/integration/basic_test.go
@@ -99,6 +99,18 @@ func (m *Minishift) executingOcCommand(command string) error {
 	return nil
 }
 
+func (m *Minishift) executingOcCommandSucceeds(command string) error {
+	err := m.executingOcCommand(command)
+	if err != nil {
+		return err
+	}
+	if lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr) != 0 {
+		return fmt.Errorf("Command did not execute successfully. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
+	}
+
+	return nil
+}
+
 func (m *Minishift) executingCommand(command string) error {
 	// TODO: there must be smarter way to destruct
 	cmdOut, cmdErr, cmdExit := m.runner.RunCommand(command)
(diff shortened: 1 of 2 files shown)


hferentschik:
Add to whitelist

test/integration/basic_test.go
```diff
@@ -108,11 +120,22 @@ func (m *Minishift) executingCommand(command string) error {
...
 	// Beware: you are responsible to verify the lastCommandOutput!
 	return nil
 }
 
+func (m *Minishift) executingMinishiftCommandSucceeds(command string) error {
+	err := m.executingCommand(command)
+	if err != nil {
+		return err
+	}
+	if lastCommandOutput.ExitCode != 0 && len(lastCommandOutput.StdErr) != 0 {
```
hferentschik: &&? I think it should be || here. If either of the two occurs the test should fail.

test/integration/basic_test.go
```diff
@@ -99,6 +99,18 @@ func (m *Minishift) executingOcCommand(command string) error {
 	return nil
 }
 
+func (m *Minishift) executingOcCommandSucceeds(command string) error {
+	err := m.executingOcCommand(command)
+	if err != nil {
+		return err
+	}
+	if lastCommandOutput.ExitCode != 0 && len(lastCommandOutput.StdErr) != 0 {
```
hferentschik: &&? I think it should be || here. If either of the two occurs the test should fail.

test/integration/basic_test.go
```diff
@@ -99,6 +99,18 @@ func (m *Minishift) executingOcCommand(command string) error {
...
+		return fmt.Errorf("cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
```
hferentschik: Maybe we can also add something "Command did not execute successfully. cmdExit: %d, cmdErr: %s"

test/integration/features/basic.feature
```diff
@@ -56,104 +50,84 @@ Feature: Basic
...
     Given Minishift has state "Running"
-     When executing "minishift stop"
+     When executing "minishift stop" succeeds
      Then Minishift should have state "Stopped"
 
   Scenario: Deleting Minishift
     Given Minishift has state "Stopped"
-     When executing "minishift delete"
+     When executing "minishift delete" succeeds
      Then Minishift should have state "Does Not Exist"
```
hferentschik: What's about adding more more tests after deletion, something like 'minishift ip'. This will actually return a non 0 exit code. This way we can keep a test for the case where we don't use the "succeeds" step. Not so much about testing Minishift, but more to show that this step also works.

test/integration/basic_test.go
amitkrout: Yes it should be ```||```. I checked it with truth table value.

test/integration/basic_test.go
amitkrout: Changes are updated

test/integration/basic_test.go
amitkrout: This is more explanatory. Changes are updated.

test/integration/features/basic.feature
amitkrout: Nice 👍 
```When executing "minishift ip"
Then exitcode should equal 1```
These two steps are added at the end of the ```Deleting Minishift``` scenario.

hferentschik:
Merged, thanks

hferentschik:
Merged via pull request #864

Issue #780, opened 2017-04-19
Add-on for openjdk imagestream and template


jorgemoralespou:
As we have an EAP7 addon, we need to have one for openjdk image which is the image we promote to develop Java microservices on openshift.

budhrg:
@jorgemoralespou  any idea where we can get the template for same ?

jorgemoralespou:
@budhrg yes. Same place you should get for EAP, as it's probably the most trustable source:

https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples/v1.5

Imagestream: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/files/examples/v1.5/xpaas-streams/jboss-image-streams.json#L371-L395

Template: https://github.com/openshift/openshift-ansible/blob/master/roles/openshift_examples/files/examples/v1.5/xpaas-templates/openjdk18-web-basic-s2i.json

praveenkumar:
@hferentschik @jorgemoralespou Since imagestream is same for EAP and openjdk `jboss-image-streams.json` should we add openjdk template as part of EAP addon or create new then duplicate image stream. I don't want to take a slice of imagestream ourself because then it's hard to maintain.

jorgemoralespou:
@praveenkumar If in the EAP add-on your installing more than EAP maybe the name should be changed. I use xPaaS, or jboss or jboss-mdw. Anything that reflects what will be installed.

praveenkumar:
@jorgemoralespou Ack so I will rename EAP addon to xPasS and add required template I think that is good way to go.

hferentschik:
@jorgemoralespou @praveenkumar - are we sure openjdk18-web-basic-s2i.json is ready for prime time. I tried to use the imported template, but I am not able to get a running app. 

    --> Scaling openjdk-app-1 to 1
    --> Waiting up to 10m0s for pods in rc openjdk-app-1 to become ready
    error: update acceptor rejected openjdk-app-1: pods for rc "openjdk-app-1" took longer than 600 seconds to become ready

Also the meta information seems odd:

![web](https://cloud.githubusercontent.com/assets/296354/25946494/b0b7348a-364b-11e7-95a6-43b790a5654f.png)

Why is there s2i in the name and no description? 

Form the other Java based templates  I can spin up a demo app, but not from this one, even though afaict it should use https://github.com/jboss-openshift/openshift-quickstarts

jorgemoralespou:
[...]

PR #875, opened 2017-05-10
Issue #780 Add openjdk template

praveenkumar:
Addresses issue #780

    - Renamed eap-template directory to xpaas
    - Renamed eap-template addon file to xpaas-template


hferentschik:
@praveenkumar I am not sure whether the added _openjdk18-web-basic-s2i_ is actually working. Have you tried to use it?

See also https://github.com/minishift/minishift/issues/780#issuecomment-300759194

praveenkumar:
@jorgemoralespou We had only those template when we worked on CDK-2.x if you think we should add all of those then someone need to verify those one by one before we included. 

@hferentschik I updated the imagestream.

jorgemoralespou:
@praveenkumar. What verification?
Those are the template that should need to be shipped, else someone will create another add-on and will open big issues since the add-on is called xPaaS but only EAP and jdk are installed.
Your missing EAP with SSL, EAP with postgresql, jdv, jdg, brms, bpms, tomcat, ...☹️

hferentschik:
@praveenkumar, regarding CDK, is it expected that Wildly would be appearing there as well? It does now and I am wondering whether this is correct?

addons/xpaas-templates/jboss-image-streams.json
```diff
@@ -367,6 +367,31 @@
...
             }
+        },
+        {
+            "kind": "ImageStream",
+            "apiVersion": "v1",
+            "metadata": {
+                "name": "redhat-openjdk18-openshift"
+            },
+            "spec": {
+                "dockerImageRepository": "registry.access.redhat.com/redhat-openjdk-18/openjdk18-openshift",
```
budhrg: I believe, this will not work with b2d iso right?

hferentschik:
Since we talked through our discussion about what happens when an image stream is already defined:

    -- Applying addon 'eap':.Error from server (AlreadyExists): error when creating "eap70-basic-s2i.json": templates "eap70-basic-s2i" already exists
    Error executing addon commands: Error executing command oc create -f eap70-basic-s2i.json -n openshift.

I think we said that this will just be ignored, but looking at this I am not so sure about it. If we are going to split into a jboss-openshift and eap add-on and in both cases the same image streams are used, we might run into issues.

jorgemoralespou:
Use oc apply, that will deploy only what's missing.

hferentschik:
Thanks for the tip.

praveenkumar:
@hferentschik @jorgemoralespou I added each and every template which is part of https://github.com/jboss-openshift/application-templates/tree/ose-v1.3.6 which I hope better for xpaas story.

addons/xpaas-templates/xpaas-templates.addon
```diff
@@ -0,0 +1,96 @@
+# Name: xpaas
+# Description: Imports xPaaS templates
+# Url: https://github.com/jboss-openshift/application-templates/tree/ose-v1.3.6
+
+oc create -f jboss-image-streams.json -n openshift
```
hferentschik: I am wondering whether we should not use 'oc apply' here as @jorgemoralespou suggested. WDYT?

hferentschik:
I cannot remember, did we say we stick to 'xpass' as name or 'jboss(-openshift)'?

addons/xpaas-templates/xpaas-templates.addon
praveenkumar: @hferentschik right now this is what a single image stream file we are using and we don't have any duplication for it. @jorgemoralespou suggested in case we need to split template but imagestream same then `apply` make more sense.

praveenkumar:
> I cannot remember, did we say we stick to 'xpass' as name or 'jboss(-openshift)'?

we stick with `jboss-openshift` but since now we put everything in same addon which complete `xpaas` scenario so I thought having it as xpaas might be better.

addons/xpaas-templates/xpaas-templates.addon
hferentschik: Sure, but apply seems to be the safer option in case other add-ons (custom or provided) are going to be added.

hferentschik: i see pro and cons for both, tend to sway more towards being more liberal

jorgemoralespou: Apply should be safer, in case an addon is applied when a template has already been created. Maybe in the update process, in case you are applying an update to the addon

jorgemoralespou:
Hi,
@praveenkumar @hferentschik I wonder some things:

- Why are you getting templates and imagestreams from [jboss-openshift repo](https://github.com/jboss-openshift/application-templates/tree/ose-v1.3.6) and not from (ansible installer repo)[https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples/v3.6]

- Executing import of the image streams can be time consuming, can this task be made a background task? Maybe in a Job? Maybe worth exploring doing similar bootstrappings in the background to prevent long startup times when starting a cluster with this (time consuming) addon.

- Why are you importing the secrets? Why are you importing them in the openshift namespace? The secrets should be imported in the project where you instantiate the template. I'm guessing here you have not realised what these are for.

hferentschik:
> Why are you getting templates and imagestreams from jboss-openshift repo and not from (ansible installer repo)[https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples/v3.6]

I guess mainly because there seems to be no consensus on what the authoritative source is for these image streams. We keep getting mixed messages here.  jboss-openshift seemed to contain what we needed, it properly tags the resources (essential imo) and it seems to be actively maintained. 

TBH, I was not aware of the Ansible examples. 

@jorgemoralespou, so why do you think we should use them? What would be the advantage? Being aligned with the Ansible installer which is the blessed way of installing OpenShift natively?

> Executing import of the image streams can be time consuming, can this task be made a background task? Maybe in a Job? 

I guess it could be done, but when running this pull request speed did not seem to be an issue.

> Maybe worth exploring doing similar bootstrappings in the background to prevent long startup times when starting a cluster with this (time consuming) addon.

I guess that would be a feature request for add-ons

> Why are you importing the secrets? 

That's a mistake.

jorgemoralespou:
>>Why are you getting templates and imagestreams from jboss-openshift repo and not from (ansible installer repo)[https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples/v3.6]

>I guess mainly because there seems to be no consensus on what the authoritative source is for these image streams. We keep getting mixed messages here. jboss-openshift seemed to contain what we needed, it properly tags the resources (essential imo) and it seems to be actively maintained.
>TBH, I was not aware of the Ansible examples.
>@jorgemoralespou, so why do you think we should use them? What would be the advantage? Being aligned with the Ansible installer which is the blessed way of installing OpenShift natively?

Mostly because of versioning alignment with openshift. While the jboss-openshift do version, they do not align with openshift, hence you do not know what alignment there is with respect software-ocp. Not saying that you maybe shouldn't use the latest from there, but in case a new release uses a function that doesn't work with a previous OCP it could break. Ansible should go through OCP QA, and be more appropriate. It's properly tagged with respect to the OCP version. Although, addons currently don't have a feature to pass arguments, as to what version should be installed (possibly aligning with the version of ocp in use).

>>Executing import of the image streams can be time consuming, can this task be made a background task? Maybe in a Job?
>I guess it could be done, but when running this pull request speed did not seem to be an issue.

Fair enough. Maybe it was me. To be honest, also I didn't run the PR, I was taking the plugin in oc-cluster wrapper as reference, although you ship all the resources, that may speed things up. Will try it whenever I can. 

>>Maybe worth exploring doing similar bootstrappings in the background to prevent long startup times when starting a cluster with this (time consuming) addon.
>I guess that would be a feature request for add-ons

I guess. I'll open a issue with this and possibly some others. Or would you prefer an issue per feature request?

>>Why are you importing the secrets?
>That's a mistake.

Great. Always better to have more eyes on this ;-)

praveenkumar:
@hferentschik @jorgemoralespou I updated this PR and now having templates from ansible repo. please have a look.

hferentschik:
ok, thanks

addons/xpaas-templates/xpaas-templates.addon
```diff
@@ -0,0 +1,102 @@
...
+oc create -f xpaas-templates/spring-boot-camel-template.json -n openshift
+oc create -f xpaas-templates/spring-boot-camel-xml-template.json -n openshift
+oc create -f xpaas-templates/spring-boot-cxf-jaxrs-template.json -n openshift
+oc create -f xpaas-templates/spring-boot-cxf-jaxws-template.json -n openshift
+
+oc create -f xpaas-templates/sso70-https.json -n openshift
+oc create -f xpaas-templates/sso70-mysql-persistent.json -n openshift
+oc create -f xpaas-templates/sso70-mysql.json -n openshift
+oc create -f xpaas-templates/sso70-postgresql-persistent.json -n openshift
+oc create -f xpaas-templates/sso70-postgresql.json -n openshift
```
hferentschik: Let's also add an echo at the end giving some sort of feedback what just happened.
Something like:

    echo Ansible OpenShift image streams and templates installed.
    echo See https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples/v3.6.

addons/xpaas-templates/xpaas-templates.addon
```diff
@@ -0,0 +1,102 @@
+# Name: xpaas
+# Description: Imports xPaaS templates
+# Url: https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples/v3.6
+
+oc create -f xpaas-streams/fis-image-streams.json -n openshift
```
hferentschik: Speaking of duplication, does this not conflict now with the _fuse_ add-on?

addons/xpaas-templates/xpaas-templates.addon
```diff
@@ -0,0 +1,102 @@
+# Name: xpaas
```
hferentschik: I'd call the file itself xpaas.addon we well, as well as the directory.

addons/xpaas-templates/xpaas-templates.addon
```diff
@@ -0,0 +1,102 @@
...
+oc create -f xpaas-templates/processserver63-amq-mysql-s2i.json -n openshift
+oc create -f xpaas-templates/processserver63-amq-postgresql-persistent-s2i.json -n openshift
+oc create -f xpaas-templates/processserver63-amq-postgresql-s2i.json -n openshift
+oc create -f xpaas-templates/processserver63-basic-s2i.json -n openshift
+oc create -f xpaas-templates/processserver63-mysql-persistent-s2i.json -n openshift
+oc create -f xpaas-templates/processserver63-mysql-s2i.json -n openshift
+oc create -f xpaas-templates/processserver63-postgresql-persistent-s2i.json -n openshift
+oc create -f xpaas-templates/processserver63-postgresql-s2i.json -n openshift
+
+oc create -f xpaas-templates/spring-boot-camel-amq-template.json -n openshift
```
hferentschik: These seem to be part of the fuse add-on as well. In fact the only thing missing from the fuse add-on are some karaf2 templates. What shall we do?

* Delete the fuse add-on?
  *  Ignoring Karaf2
  *  Adding Karaf2 to pass
* Keep both add-ons
   * As is
   * Rethink our use of 'oc create' and use 'oc apply'

praveenkumar: @hferentschik you are right, I would vote to remove fuse-templates complete if that already comes under `xpaas`, @jorgemoralespou any suggestion?

addons/xpaas-templates/xpaas-templates.addon
praveenkumar: @hferentschik yes it will, I didn't notice. l am tend to remove fuse completely if that is also part of `xpaas`.

addons/xpaas-templates/xpaas-templates.addon
praveenkumar: +1, will do that and other changes once get ack from @jorgemoralespou

addons/xpaas-templates/xpaas-templates.addon
praveenkumar: +1 Agree, will add it.

hferentschik:
@praveenkumar should we add xpass to the default addons in addon_install.go?

praveenkumar:
> should we add xpass to the default addons in addon_install.go?

Yes we should add it, now we have it from official openshift ansible repo.

hferentschik:
> Yes we should add it, now we have it from official openshift ansible repo.

+1

addons/xpaas-templates/xpaas-templates.addon
jorgemoralespou: @praveenkumar @hferentschik I would use "oc apply" in any case, as any user can contribute addons and it could conflict between what is provided OOTB and what he provides. I would try to avoid conflicts in what is provided OOTB as much as possible. User contributed addons are user's responsibility.

With regards to keeping or removing fis addon, the question boils down to the granularity you want to provide to the end user. Given that importing the imagestreams and templates seems not to be time consuming, I would have one xpaas addon that install everything, reducing complexity and maintenance efforts.

Then every user could just make a user addon more slim if they want/need.

jorgemoralespou:
> should we add xpass to the default addons in addon_install.go?
Yes.

praveenkumar:
@hferentschik @jorgemoralespou updated PR as per feedback also I put `apply` for image streams .

hferentschik:
Merged, thanks. Yeah :-)

PR #883, opened 2017-05-11
Issue #878 Fails to escape password with special characters

### LINKED ISSUE #878: Fails to escape password with special characters
When attempting to use a developer account with a special character in the password, minishift fails to escape it correctly during setup:

```
$./minishift --vm-driver=virtualbox --username=user@example.com --password="fakePWthat-has-)-in-it" start
E0510 14:37:54.600570   11169 start.go:176] Error starting the VM: Error registering the VM: ssh command error:
command : sudo -E subscription-manager register --auto-attach --username=user@example.com --password fakePWthat-has-)-in-it
err     : exit status 1
output  : bash: -c: line 0: syntax error near unexpected token `)'
```

Using MINISHIFT_PASSWORD has same effect.

Intentionally escaping yields a double escape:
```
$./minishift --vm-driver=virtualbox --username=user@example.com --password="fakePWthat-has-\)-in-it" start
E0510 14:37:54.600570   11169 start.go:176] Error starting the VM: Error registering the VM: ssh command error:
command : sudo -E subscription-manager register --auto-attach --username=user@example.com --password fakePWthat-has-\\)-in-it
err     : exit status 1
output  : bash: -c: line 0: syntax error near unexpected token `)'
```

budhrg:
Fix #878 

Found couple other special character which needs similar fix.

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -111,10 +111,21 @@ func (m MultiError) ToError() error {
 // EscapeStringForSSHUse will parse string and replace
 // shell special characters with escape
 func EscapeStringForSSHUse(s string) string {
-	r := strings.NewReplacer("$", "\\$",
+	r := strings.NewReplacer(
+		"$", "\\$",
 		`"`, `\"`,
+		`'`, `\'`,
 		"`", "\\`",
-		`\`, `\\`)
+		`\`, `\\`,
+		`(`, `\(`,
+		`)`, `\)`,
+		`{`, `\{`,
+		`}`, `\}`,
+		`[`, `\[`,
+		`]`, `\]`,
+		` `, `\ `,
+	)
+
 	return r.Replace(s)
 }
 
--- a/pkg/util/utils_test.go
+++ b/pkg/util/utils_test.go
@@ -65,7 +65,11 @@ func TestEscapeStringForSSHUse(t *testing.T) {
 		`Pa"ssw\ord$$`: `Pa\"ssw\\ord\$\$`,
 		`Password`:     `Password`,
 		"Passwo`rd":    "Passwo\\`rd",
+		"Passwo(rd":    `Passwo\(rd`,
+		"Passwo rd":    `Passwo\ rd`,
+		"Passwo'rd":    `Passwo\'rd`,
 	}
+
 	for pass, expected := range experimentstring {
 		if EscapeStringForSSHUse(pass) != expected {
 			t.Fatalf("Expected '%s' Got '%s'", expected, EscapeStringForSSHUse(pass))


pkg/util/utils.go
```diff
@@ -111,10 +111,20 @@ func (m MultiError) ToError() error {
...
-	r := strings.NewReplacer("$", "\\$",
+	r := strings.NewReplacer(
+		"$", "\\$",
 		`"`, `\"`,
 		"`", "\\`",
-		`\`, `\\`)
+		`\`, `\\`,
+		`(`, `\(`,
+		`)`, `\)`,
+		`{`, `\{`,
```
hferentschik: -1, this is not about escaping special shell characters. This is a specific escape for using a string as part of an ssh command. There is only a limited amount of characters which needs to be escaped there.

On top of this, even in a shell, {, }, [, ] would not have a special meaning.

I think we can actually close this pull request and resolve also the issue. The solution is to put the password into single quotes (instead of double) to prevent substitution within the shell.

LalatenduMohanty: > The solution is to put the password into single quotes (instead of double) to prevent substitution within the shell.

+1 I think this is a reasonable solution.

budhrg: >  The solution is to put the password into single quotes (instead of double) to prevent substitution within the shell.

I did that and I am getting same error message as mentioned in the issue. Doing the explicit escape as done in this PR fix that. Same for `;` too.

hferentschik: We even document using single quotes - https://docs.openshift.org/latest/minishift/using/troubleshooting.html#special-characters-passwords

The things which gets in the way here is the escaping rules of your shell. Once the password is passed on to us, we are dealing with the rules around what and what not is allowed  in terms of passing on the password as part of an ssh command. The rules differ. 

I tried with single quotes and for all I can tell things like ';' or spaces are not a problem.

@budhrg have you actually tested this end to end. I have a feeling you are breaking actual passwords in some cases.

budhrg: The workaround mentioned here https://docs.openshift.org/latest/minishift/using/troubleshooting.html#special-characters-passwords is not working for me when I run

### 1. minishift start --username foo --password 'foo bar'
```
E0512 13:52:02.292415   26563 start.go:176] Error starting the VM: Error registering the VM: ssh command error:
command : sudo -E subscription-manager register --auto-attach --username bgurung@redhat.com --password foo bar 
err     : exit status 64
output  : cannot parse argument: bar
. Retrying.
```

### 2. minishift start --username foo --password 'foo)bar'
```
E0512 13:57:01.094153   27974 start.go:176] Error starting the VM: Error registering the VM: ssh command error:
command : sudo -E subscription-manager register --auto-attach --username bgurung@redhat.com --password foo(bar 
err     : exit status 1
output  : bash: -c: line 0: syntax error near unexpected token `('
bash: -c: line 0: `sudo -E subscription-manager register --auto-attach --username bgurung@redhat.com --password foo(bar '
. Retrying.
```

Same for other brackets too.
Semi colon seems to be not affected.

budhrg:
Hi @praveenkumar , as you asked the single quote is not working now. Added fix for that in the PR too.

Getting following error before this PR. 

### $ minishift start --username foo --password "foo'bar"
```
E0512 14:07:23.846957   32671 start.go:176] Error starting the VM: Error registering the VM: ssh command error:
command : sudo -E subscription-manager register --auto-attach --username bgurung@redhat.com --password foo'bar 
err     : exit status 1
output  : bash: -c: line 0: unexpected EOF while looking for matching `''
bash: -c: line 1: syntax error: unexpected end of file
. Retrying.
```

@hferentschik @LalatenduMohanty , I will collaborate with @praveenkumar next week to have proper solution for this. Let's keep the PR till then.

hferentschik:
>  I will collaborate with @praveenkumar next week to have proper solution for this. Let's keep the PR till then.

+1, let's sort this out once and for all. I still wonder why we not pass the password quote to the ssh/sudo command. @praveenkumar was at the time very adamant that this is not needed.

gbraad:
Or use a way to also specify the password as part of the input as done with hostfolders. You could do this as an alternative way when not specified as an argument. This way you provide a way to get around the shell interpretation. Because in this case it can happen twice... when you specify it as an argument, and when executed on the host when again it is used as an argument. This will certainly causes issues to work around all of the edge cases... plus the fact that shell can behave differently.

hferentschik:
> Or use a way to also specify the password as part of the input as done with host folders.

@gbraad that's a different issue really -  https://issues.jboss.org/browse/CDK-110. We need this as well, but that's more for the case where a user does not want to use an env variable nor specify the password on the command line. 

We retrieve the password ok, or better we have to assume that the password we deal with was quoted and escaped by the user in the correct manner depending on his shell type. The problem here is moving forward on how we use this password as part of the ssh command issues to the VM in order to register. There it seems we don't do the right thing yet.

LalatenduMohanty:
Closing this in favor of #914 . Assuming @budhrg and @praveenkumar planned this way.

PR #894, opened 2017-05-14
Issue #814 major.minor.patch version format for jteeuwen/go-bindata

### LINKED ISSUE #814: Set explicit version for jteeuwen/go-bindata

### LINKED ISSUE #893: minishift fails to build against the latest github.com/google/go-github
On Fedora 25, minishift 1.0.0 build fails with the latest https://github.com/google/go-github/commit/4a42a1def4af554769703e8ac0ce5290c16371a2 and Fedora's 553fda4:

```
go install -pkgdir=/usr/share/gocode/src/github.com/minishift/minishift/out/bindata -ldflags="-X github.com/minishift/minishift/pkg/version.version=1.0.0 -X github.com/minishift/minishift/pkg/version.isoVersion=v1.0.2 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v1.5.0" ./cmd/minishift
# github.com/minishift/minishift/pkg/util/github
pkg/util/github/github.go:101: not enough arguments in call to client.Repositories.GetReleaseByTag
pkg/util/github/github.go:104: not enough arguments in call to client.Repositories.GetLatestRelease
pkg/util/github/github.go:123: not enough arguments in call to client.Repositories.DownloadReleaseAsset
pkg/util/github/github.go:309: not enough arguments in call to client.Repositories.DownloadReleaseAsset
```

I know that glide.lock uses an older dependency explicitly https://github.com/minishift/minishift/blob/ebcc3ee84489a71a710a54890c646a38b48f6dcf/glide.lock#L71 and minishift builds using it, but sooner or later one will need to update to a more recent github.com/google/go-github.

budhrg:
Fixes #814 

Also fixes #893 which seems to urgent and blocking from new users to build it properly.

The PR (https://github.com/minishift/minishift/pull/818) was closed not to have any changes in Makefile till 1.0.0.

Now, 1.0.0 is out, hence sending it again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,12 +50,12 @@ DOCS_SYNOPISIS_DIR = docs/source/_tmp
 # Start of the actual build targets
 
 .PHONY: $(GOPATH)/bin/minishift$(IS_EXE)
-$(GOPATH)/bin/minishift$(IS_EXE): vendor $(ADDON_ASSET_FILE)
+$(GOPATH)/bin/minishift$(IS_EXE): $(ADDON_ASSET_FILE) vendor
 	go install -pkgdir=$(ADDON_BINDATA_DIR) -ldflags="$(VERSION_VARIABLES)" ./cmd/minishift
 vendor:
 	glide install -v
 
-$(ADDON_ASSET_FILE): $(GOPATH)/bin/go-bindata vendor
+$(ADDON_ASSET_FILE): $(GOPATH)/bin/go-bindata
 	@mkdir -p $(ADDON_BINDATA_DIR)
 	go-bindata $(GO_BINDATA_DEBUG) -prefix $(ADDON_ASSETS) -o $(ADDON_ASSET_FILE) -pkg bindata $(ADDON_ASSETS)/...
 
--- a/glide.yaml
+++ b/glide.yaml
@@ -25,6 +25,7 @@ import:
   - libmachine/state
   - libmachine/swarm
 - package: github.com/google/go-github
+  version: 30a21ee1a3839fb4a408efe331f226b73faac379
   subpackages:
   - github
 - package: github.com/inconshreveable/go-update
@@ -44,7 +45,7 @@ import:
 - package: github.com/pkg/errors
   version: ^0.8.0
 - package: github.com/jteeuwen/go-bindata
-  version: ~3.0
+  version: v3.0.7
 - package: github.com/spf13/cobra
   subpackages:
   - doc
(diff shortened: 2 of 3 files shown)


glide.yaml
```diff
@@ -25,6 +25,7 @@ import:
   - libmachine/state
   - libmachine/swarm
 - package: github.com/google/go-github
+  version: 30a21ee1a3839fb4a408efe331f226b73faac379
```
praveenkumar: Why we need here version info as shasum?

hferentschik: There is no released version nor tag. What I don't understand is the relation to #893. AFAICT, you are explicitly setting now the version as it was determined by glide.lock. So there is no update here. Would the plan for #893 not be to actually update to the latest version of go-github, which would also require some code changes? Do you want to address this as part of a dedicated pull request for #893?

hferentschik: > which seems to urgent and blocking from new users to build it properly.

There is no issue building the project from scratch using our build. The problem lies in trying to build it from Fedora rpms. However, this approach needs to stop looking at glide.yaml, but rather look at glide.lock. Unless the exact pinned down versioned are used, whatever the build results, it is not Minishift.

Makefile
```diff
@@ -50,12 +50,12 @@ DOCS_SYNOPISIS_DIR = docs/source/_tmp
 # Start of the actual build targets
 
 .PHONY: $(GOPATH)/bin/minishift$(IS_EXE)
-$(GOPATH)/bin/minishift$(IS_EXE): vendor $(ADDON_ASSET_FILE)
+$(GOPATH)/bin/minishift$(IS_EXE): $(ADDON_ASSET_FILE) vendor
```
hferentschik: +1, This works, there was indeed a bit of an issue when deleting the glide.lock and trying to re-bootstrap. Now it works again :-)

glide.yaml
```diff
@@ -44,7 +45,7 @@ import:
 - package: github.com/pkg/errors
   version: ^0.8.0
 - package: github.com/jteeuwen/go-bindata
-  version: ~3.0
```
hferentschik: There was actually a reason I used ~3.0. I wanted to transparently be able to upgrade up to an 3.1. But anyways, ...

hferentschik:
Rebased and merged, thanks.

Issue #900, opened 2017-05-15
Allow print planning script to create data for closed milestones 


hferentschik:
Modify _scripts/planning/sprint-planning.sh_ to allow generation of csv data for closed GitHub milestones as well (for post processing sprints)

hferentschik:
Merged via pull request #901

PR #901, opened 2017-05-15
Issue #900 Change GitHub API call to retrieve issues of a milestone r…

hferentschik:
…egarless of issue state

Addresses issue #900


hferentschik:
@LalatenduMohanty as discussed, here is the missing changes to the sprint planning script.

PR #904, opened 2017-05-15
Issue #902 Fix the links in installation and installation guide

### LINKED ISSUE #902: Fix the links in installation and installation guide
Found that the KVM hyperlink is pointing to minishift.io here https://docs.openshift.org/latest/minishift/getting-started/installing.html#install-prerequisites 

and

Using -> Uninstalling   uses minishift.io here https://github.com/minishift/minishift/blob/master/docs/source/getting-started/uninstalling.adoc

budhrg:
Fix #902 

@hferentschik Could we merged so that it is available in the next doc sync with docs.openshift.org ?

### DIFF
--- a/docs/source/getting-started/uninstalling.adoc
+++ b/docs/source/getting-started/uninstalling.adoc
@@ -22,7 +22,7 @@ $ minishift delete
 +
 This command deletes everything in the
 `MINISHIFT_HOME/.minishift/machines/minishift` directory. Other cached data and
-the https://minishift.io/docs/managing-minishift.html#persistent-configuration[persistent configuration] are not removed.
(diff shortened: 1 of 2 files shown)


docs/source/getting-started/uninstalling.adoc
```diff
@@ -22,7 +22,7 @@ $ minishift delete
 +
 This command deletes everything in the
 `MINISHIFT_HOME/.minishift/machines/minishift` directory. Other cached data and
-the https://minishift.io/docs/managing-minishift.html#persistent-configuration[persistent configuration] are not removed.
+the link:../using/managing-minishift.html#persistent-configuration[persistent configuration] are not removed.
```
hferentschik: this really should be an xref, but we take this discussion later once we settled the doc writing guidelines.

Issue #874, opened 2017-05-10
Unable to run integration test for downstream bits


praveenkumar:
The current integration tests start with calling minishift start in order to bring up Minishift. This is not going to work with the CDK binary, since it requires to call setup-cdk first:

```
go test -timeout 3600s github.com/minishift/minishift/test/integration --tags=integration -v -args --binary /Users/hardy/work/developer-platform/cdk-3/go/bin/minishift
Running Integration test in:  /var/folders/39/70n784613gd370g63ggg62lh0000gn/T/minishift-integration-test-332904236
using binary:  /Users/hardy/work/developer-platform/cdk-3/go/bin/minishift

F---F---F---F--------F----F----F--F-- 37
--- Failed steps:
    Given  Minishift has state "Does Not Exist" # features/basic.feature:7

      Error: Minishift state did not match. Expected: Does Not Exist, Actual: You need to run 'minishift setup-cdk' first to install required CDK components.
```
One way of doing this is by replacing the 
```
When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT" succeeds
```
with an explicit
```
When starting Minshift with "<start command args>" succeeds
```

Basically we create a dedicated step for starting Minishift. This way we can prior to actually running the start command, check whether we are dealing with the CDK and if so run the setup command. We can use the `minishift-home` flag of the setup command to make sure the extraction happens into the right directory. The location we can get via `constants.MiniShiftHomeEnv`.

We can add some additional methods to the `MinishiftRunner` in util.go - `IsCdk()` and `NeedsSetup()`. The former checks whether we are dealing with the cdk. One way to do this would be to check the output of `minishift -h` to see whether there is a setup-cdk command. NeedsSetup() can check whether the CDK marker files exists. This might be useful when trying to call start (for example for test purposes) multiple times. 

 Source: https://issues.jboss.org/projects/CDK/issues/CDK-101

praveenkumar:
Fix by #905

PR #905, opened 2017-05-16
Issue #874 Implementation to run integration test for downstream bits

praveenkumar:
I did test and it works as expected but user should have RHN username/password as environment variable. 

One test scenario fail because it's specific to b2d image.

```cucumber
  Scenario: User is able to set custom Docker specific environment variables
     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
     Then stdout should contain
      """
      export "FOO=BAR"
      export "BAZ=BAT"
      """

--- Failed steps:

    Then  stdout should contain # features/basic.feature:62
      Error: Output did not match. Expected: export "FOO=BAR"
export "BAZ=BAT", Actual: 
```

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -6,7 +6,7 @@ Feature: Basic
 
   Scenario: Starting Minishift
     Given Minishift has state "Does Not Exist"
-     When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT" succeeds
+     When executing "minishift start" succeeds
      Then Minishift should have state "Running"
 
   Scenario: OpenShift is ready after startup
@@ -100,15 +100,6 @@ Feature: Basic
      Then stdout should contain
       """
       hello
-      """
-
-  Scenario: User is able to set custom Docker specific environment variables
-    Given Minishift has state "Running"
-     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
-     Then stdout should contain
-      """
-      export "FOO=BAR"
-      export "BAZ=BAT"
       """
 
   Scenario: User is able to retrieve host and port of OpenShift registry
--- a/test/integration/util/util.go
+++ b/test/integration/util/util.go
@@ -81,6 +81,19 @@ func (m *MinishiftRunner) Start() {
 	m.RunCommand(fmt.Sprintf("start %s", m.CommandArgs))
 }
 
+func (m *MinishiftRunner) CDKSetup() {
+	if (os.Getenv("MINISHIFT_USERNAME") == "") || (os.Getenv("MINISHIFT_PASSWORD") == "") {
+		fmt.Println("Either MINISHIFT_USERNAME or MINISHIFT_PASSWORD is not set as environment variable")
+		os.Exit(1)
+	}
+	m.RunCommand(fmt.Sprintf("setup-cdk --force --minishift-home %s", os.Getenv(constants.MiniShiftHomeEnv)))
+}
+
+func (m *MinishiftRunner) IsCDK() bool {
+	cmdOut, _, _ := m.RunCommand("-h")
+	return strings.Contains(cmdOut, "setup-cdk")
+}
+
 func (m *MinishiftRunner) EnsureRunning() {
 	if m.GetStatus() != "Running" {
 		m.Start()
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -108,6 +108,9 @@ func FeatureContext(s *godog.Suite) {
 
 	s.BeforeSuite(func() {
 		testDir = setUp()
+		if runner.IsCDK() {
+			runner.CDKSetup()
+		}
 		fmt.Println("Running Integration test in:", testDir)
 		fmt.Println("using binary:", givenPath)
 	})


test/integration/util/util.go
```diff
@@ -81,6 +81,18 @@ func (m *MinishiftRunner) Start() {
 	m.RunCommand(fmt.Sprintf("start %s", m.CommandArgs))
 }
 
+func (m *MinishiftRunner) NeedsSetup() {
```
hferentschik: NeedsSetup == SetupCDK?

test/integration/util/util.go
```diff
@@ -81,6 +81,18 @@ func (m *MinishiftRunner) Start() {
...
+	m.RunCommand(fmt.Sprintf("setup-cdk --force --minishift-home %s", os.Getenv(constants.MiniShiftHomeEnv)))
+}
+
+func (m *MinishiftRunner) IsCDK() bool {
```
hferentschik: +1

test/integration/util/util.go
praveenkumar: @hferentschik  or `NeedCdkSetup` ? it just rhyme with isCDK then `NeedCdkSetup` :) Just my thought (not a hard feeling around it).

hferentschik:
> I did test and it works as expected but user should have RHN username/password as environment variable.

We should add something into the branch where we detected that we are dealing with the CDK and check for the environment variables. If not set, we should fail the test. WDYT?

> One test scenario fail because it's specific to b2d image.

Let's get rid of this for now. I does not really fit so well in basic.feature right now. I am happy to change the feature to "When executing "minishift start" succeeds" and remove the  scenario "User is able to set custom Docker specific environment variables".

We can move this into a separate feature file and for example use an annotation @b2d to indicate that this test only works with Boot2Docker ISO.

That said, what is the issue? On CentOS ISO the variables are stored somewhere else? Not _/var/lib/boot2docker/profile_? Would there be a way to test in a generic way whether the Docker daemon has these variables defined? Either way, I vote to move this test into its own feature file.

agajdosi:
Works on my workstation. Only strange thing is that it seems like it ignores `MINISHIFT_PASSWORD` and `MINISHIFT_USERNAME` exported through ~/.bashrc, they need to be exported manually.

Looked on code and I think that in future it will be fine to move "setup of cdk" from beforeSuite to an actual scenario, because we want to test that also. For now it is fine I guess, I would like just to check if my understanding is correct?

praveenkumar:
> That said, what is the issue? On CentOS ISO the variables are stored somewhere else? Not /var/lib/boot2docker/profile? Would there be a way to test in a generic way whether the Docker daemon has these variables defined?

@hferentschik yes it's stored different location for RHEL/CentOS. I thought `docker info` will provide required info but that doesn't include docker env variables so for now I removed it from the test case. Tested again and worked as expected.

praveenkumar:
> Looked on code and I think that in future it will be fine to move "setup of cdk" from beforeSuite to an actual scenario, because we want to test that also. For now it is fine I guess, I would like just to check if my understanding is correct?

When you say move to a scenario that mean this scenario only going to run for CDK? Because what we have right now is check if binary related to CDK and then do addition required steps so all scenario can be run same way with downstream/upstream. My understanding is this is mostly a setup task.

agajdosi:
@praveenkumar 
> that mean this scenario only going to run for CDK?

Yes, it is a little bit strange to implement in upstream scenario for downstream feature, but if we are doing it in beforeScenario, then does not matter. I am having hard time formulating it but for example:

```
Scenario: User can configure minishift files, if setup command is packed and available
    Given: minishift home folder is empty
      And: Minishift status is not running
     When: user runs setup command
     Then: minishift is configured
```
But I am not sure right now how exactly skip scenario when `IsCDK()==false`.

gbraad:
Maybe we should not be testing:

```
    Then  stdout should contain # features/basic.feature:62
      Error: Output did not match. Expected: export "FOO=BAR"
export "BAZ=BAT", Actual: 
```

as this is only possible for b2d yes. we start the docker environment with passing variables in. Is this something we really need? Especially if we would move to make the CentOS iso the default, this becomes difficult/impossible.

test/integration/features/basic.feature
```diff
@@ -102,15 +102,6 @@ Feature: Basic
       hello
       """
```
gbraad: Can you also change:

```
  Scenario: Starting Minishift
     When executing "minishift start --docker-env=FOO=BAR --docker-env=BAZ=BAT" succeeds
```

As this is where we set the variables which are tested in this part.

test/integration/util/util.go
```diff
@@ -81,6 +81,18 @@ func (m *MinishiftRunner) Start() {
 	m.RunCommand(fmt.Sprintf("start %s", m.CommandArgs))
 }
 
+func (m *MinishiftRunner) CDKSetup() {
+	m.RunCommand(fmt.Sprintf("setup-cdk --force --minishift-home %s", os.Getenv(constants.MiniShiftHomeEnv)))
+}
+
+func (m *MinishiftRunner) IsCDK() bool {
+	cmdOut, _, _ := m.RunCommand("-h")
+	if strings.Contains(cmdOut, "setup-cdk") {
```
gbraad:
```go
return strings.Contains(cmdOut, "setup-cdk")
```

gbraad: Also rather see:

```
+                      if (os.Getenv("MINISHIFT_USERNAME") == "") || (os.Getenv("MINISHIFT_PASSWORD") == "") {
 +				fmt.Println("Either MINISHIFT_USERNAME or MINISHIFT_PASSWORD is not set as environment variable")
 +				os.Exit(1)
 +			}
```
here instead of the before suite. as it reduces the nesting in `BeforeSuite`, and is easier to reason about if the error occurs.

gbraad: Seems this is caused by the fact that no Println is shown for code outside of the feature setup.

test/integration/features/basic.feature
praveenkumar: @gbraad I am bit skeptical about it since atleast this steps making sure with `--docker-env` option minishift is up and running even we are not checking if those options present or not.

gbraad: @praveenkumar which seems to be moot. I could specify anything on the command line, but when untested they are not for sure relevant or correctly implemented. If you remove the previous test, I see no relevance to keep this in now. If however we will use this, we need to be sure it works... also for the CentOS image.

praveenkumar: @gbraad I got it, where are you coming from, I remove that part.

gbraad: @praveenkumar If I really wanted to be picky, I would request even a different commit. so a revert would be much easier... ;-)

praveenkumar: @gbraad Done added 2 commit messages.

gbraad:
If @budhrg can quickly review this, we can merge in my opinion

PR #911, opened 2017-05-17
Issue #912 Resolve error when no SSH installed on Windows

### LINKED ISSUE #912: Invalid handle when no SSH installed on Windows
See: https://github.com/minishift/minishift-centos-iso/issues/95

gbraad:
This will resolve #912

### DIFF
--- a/pkg/minishift/provisioner/minishift_provisioner.go
+++ b/pkg/minishift/provisioner/minishift_provisioner.go
@@ -35,7 +35,7 @@ import (
 
 func NewMinishiftProvisioner(osReleaseID string, d drivers.Driver) *MinishiftProvisioner {
 	systemdProvisioner := provision.NewSystemdProvisioner(osReleaseID, d)
-	systemdProvisioner.SSHCommander = provision.RedHatSSHCommander{Driver: d}
+	systemdProvisioner.SSHCommander = provision.GenericSSHCommander{Driver: d}
 	return &MinishiftProvisioner{
 		systemdProvisioner,
 	}


LalatenduMohanty:
@gbraad Nice. Can you please rebase your branch.

gbraad:
Rebased and changed commit message

gbraad:
retest please

budhrg:
`minishift.exe ssh` is not working though `minishift start` worked as expected. However, `minishift ss docker images` seems to be working.

```
PS C:\gowork\src\github.com\minishift\minishift> which ssh
which : The term 'which' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify
that the path is correct and try again.
At line:1 char:1
+ which ssh
+ ~~~~~
    + CategoryInfo          : ObjectNotFound: (which:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException


PS C:\gowork\src\github.com\minishift\minishift> minishift.exe start --iso-url file://C:/minishift-centos7.iso --show-libmachine-logs
Starting local OpenShift cluster using 'hyperv' hypervisor...
...
(minishift) Downloading C:\Users\budhram\.minishift\cache\boot2docker.iso from file://C:/minishift-centos7.iso...
(minishift) Creating SSH key...
(minishift) Creating VM...
(minishift) Using switch "minishift_switch"
...
-- Server Information ...
   OpenShift server started.
       https://192.168.1.105:8443

   You are logged in as:
       User:     developer
       Password: developer

   To login as administrator:
       oc login -u system:admin


PS C:\gowork\src\github.com\minishift\minishift> ssh
ssh : The term 'ssh' is not recognized as the name of a cmdlet, function, script file, or operable program. Check the spelling of the name, or if a path was included, verify that
the path is correct and try again.
At line:1 char:1
+ ssh
+ ~~~
    + CategoryInfo          : ObjectNotFound: (ssh:String) [], CommandNotFoundException
    + FullyQualifiedErrorId : CommandNotFoundException


PS C:\gowork\src\github.com\minishift\minishift> minishift.exe ssh    ##### Not Working


PS C:\gowork\src\github.com\minishift\minishift> minishift.exe ssh docker images
REPOSITORY                            TAG                 IMAGE ID            CREATED             SIZE
docker.io/openshift/origin-deployer   v1.5.1              0f23b83cad80        16 hours ago        617.5 MB
docker.io/openshift/origin            v1.5.1              a23dc456ea10        16 hours ago        617.5 MB
docker.io/openshift/origin-pod        v1.5.1              aad02d5e14b9        16 hours ago        1.143 MB

PS C:\gowork\src\github.com\minishift\minishift> minishift.exe ssh
```

gbraad:
@budhrg This has always been the case. It is not possible to have an interactive shell when no external ssh client is installed. This is expected.

Please verify this behaviour on Linux when no ssh client is available on path.

gbraad:
Same on Linux
```
$ sudo mv /usr/bin/ssh /usr/bin/.ssh
$ minishift start
$ minishift ssh    # no interactive shell
$ minishift stop
$ sudo mv /usr/bin/.ssh /usr/bin/ssh
```

gbraad:
recheck please

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -35,7 +35,7 @@ import (
 
 func NewMinishiftProvisioner(osReleaseID string, d drivers.Driver) *MinishiftProvisioner {
 	systemdProvisioner := provision.NewSystemdProvisioner(osReleaseID, d)
-	systemdProvisioner.SSHCommander = provision.RedHatSSHCommander{Driver: d}
+	systemdProvisioner.SSHCommander = provision.GenericSSHCommander{Driver: d}
```
praveenkumar: 👍

gbraad:
>
> I believe, we need to add documentation note for it as if user want to
> perform interactive ssh then they need get third party ssh client to
> perform it. This is quite not visible in Linux but in Windows.
>

@budhrg Agree, this means we need to modify the current documentation to
address this. I'd rather create a new issue for this, not to slow this
merge request down.

budhrg:
> I'd rather create a new issue for this, not to slow this merge request down.

Work for me for this PR :+1:

budhrg:
> I'd rather create a new issue for this, not to slow this merge request down.

Actually, I would really like to see the documentation for this one along with this PR because with this PR we are saying one doesn't need ssh binary but one should know what other things are coming up with this one.

Also, we have been following 'implementation + test + doc' changes as single commit so far which seems good with point that one one can see everything related to that task with that commit.

Let us know what is your opinion here and then we can discuss and follow :smile:

LalatenduMohanty:
> I believe, we need to add documentation note for it as if user want to perform interactive ssh then they need get third party ssh client to perform it. This is quite not visible in Linux but in Windows.

We can, however we do not want users to do `$ minishift ssh` i.e. not something we should recommend.


> Also, we have been following implementation + test + doc changes as single commit until the doc requirement is too divergent to make it as another issue.

+1. However in this case I think it is different as this PR does not change the older workflow for minishift in WIndows + we do not want users to ssh in to minishift VM and currently we do not have any documentation for  around ` $ minishift ssh` apart from command reference docs. Also we need test the ssh client e.g. OpenSSH (not sure if everyone agrees to use OpenSSH) along with documentation how to install it etc. So I think it would be better to open an new issue and do proper documentation around it. WDYT?

LalatenduMohanty:
The AppVeyor seems to fail with below error
```
-- FAIL: TestCacheOc (10.04s)
	oc_caching_test.go:69: Error attempting to download and cache oc: Cannot get the OpenShift release version v1.3.1: Get https://api.github.com/repos/openshift/origin/releases/tags/v1.3.1: net/http: TLS handshake timeout
FAIL
FAIL	github.com/minishift/minishift/pkg/minishift/cache	10.781s
?   	github.com/minishift/minishift/pkg/minishift/clusterup	[no test files]
```
Anyone seen this before?

budhrg:
retest this please

gbraad:
> retest

doesn't work here...

> `TLS handshake timeout`

Sounds like an issue on establishing a secure connection to the API backend. Feels like a glitch on GitHub's end. Not much we can do about it.

budhrg:
> Sounds like an issue on establishing a secure connection to the API backend. Feels like a glitch on GitHub's end. Not much we can do about it.

Thinking of re-triggering the job but look like I don't have access.

gbraad:
>  So I think it would be better to open an new issue and do proper documentation around it. WDYT?

I agree

gbraad:
For Windows I would like to recommend:

  * https://github.com/PowerShell/Win32-OpenSSH

But at the moment this is not production ready and has no 'installer', but it will integrate better with PowerShell.

budhrg:
@LalatenduMohanty 

> However in this case I think it is different as we do not have any documentation around $ minishift ssh apart from command reference docs.

We have only instruction for mounting [sshfs folder mount]( https://docs.openshift.org/latest/minishift/using/host-folders.html#sshfs-folder-mount) which contain `minishift ssh`.

If we are creating separate issue for doc, then let's change above intruction as : 
```
$ minishift ssh "sshfs <username>@<IP>:/Users/<username>/ /Users"
```

gbraad:
This should have been removed in the hostfolder documentation task.

gbraad:
Correction: only for sshfs it is relevant to keep this (for now). But.in this case people also need to have ssh installed (including sftp)

LalatenduMohanty:
> But.in this case people also need to have ssh installed (including sftp AFAIR)

@budhrg as @gbraad mentioned I think we need a proper sshfs client for it. So lets handle as part of the documentation task.

PR #914, opened 2017-05-17
Issue #878 Fail to escape password with special charachers

praveenkumar:
Basically the way sshCommander works is it will take the string as it is and then start ssh session. when it start session it is always be a new shell so if we want to escape special characters then we need to make sure string we we are passing should have `'<Password>'`.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -531,7 +531,7 @@ func ocSupportFlag(cmdName string, flag string) bool {
 
 func setSubcriptionManagerParameters() {
 	cluster.RegistrationParameters.Username = viper.GetString(username)
-	cluster.RegistrationParameters.Password = util.EscapeStringForSSHUse(viper.GetString(password))
+	cluster.RegistrationParameters.Password = viper.GetString(password)
 }
 
 func getConfigClusterName(ip string, port int) string {
--- a/docs/source/using/troubleshooting.adoc
+++ b/docs/source/using/troubleshooting.adoc
@@ -21,7 +21,9 @@ special characters can trigger variable interpolation and therefore
 cause passwords to fail.
 
 Workaround: When creating and entering passwords, wrap the string with
-single quotes in the following format: `<password>`
+single quotes in the following format: `'<password>'`. If your password
+contain single quote ( `'` ) then do check the escaping string for your
+current shell. For example bash have escape string `'"'"'` for single quote.
 
 [[minishift-delete-fails-undefine-snapshots]]
 == Undefining virsh snapshots fail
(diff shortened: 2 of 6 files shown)


praveenkumar:
@LalatenduMohanty we need password in single quote only if contain special characters otherwise without quote it also works and that is already mentioned in the  troubleshooting guide.

gbraad:
Would it hurt to always use single quotes?

gbraad:
Already saw you changed the command description, but how about documentation about this?

budhrg:
Looks like now having issue with password containing single quote. The password containing space, brackets seesm to be working.

```
E0518 11:30:00.955732   16556 start.go:176] Error starting the VM: Error registering the VM: ssh command error:
command : sudo -E subscription-manager register --auto-attach --username bgurung@redhat.com --password 'foo'bar' 
err     : exit status 1
output  : bash: -c: line 0: unexpected EOF while looking for matching `''
bash: -c: line 1: syntax error: unexpected end of file
. Retrying.
Waiting for SSH to be av
```

praveenkumar:
> Would it hurt to always use single quotes?

No it won't, but even you tell users then also they miss and only understand when face issue/failure :)

praveenkumar:
> Looks like now having issue with password containing single quote. The password containing space, brackets seesm to be working.

Single quote anyway need escape and that understood.

budhrg:
> Single quote anyway need escape and that understood.

I didn't get what it means. Is there anything here user need to do?

praveenkumar:
> I didn't get what it means. Is there anything here user need to do? I tried even passing it as "foo\'bar".

if you want to escape single quote then for bash it should be `'"'"'` as escaped sequence. Can you try `'foo'"'"'bar` which should work.

budhrg:
> With minishift start --username bgurung@redhat.com --password 'foo'"'"'bar'  

```
E0518 11:57:13.451351   21442 start.go:176] Error starting the VM: Error registering the VM: ssh command error:
command : sudo -E subscription-manager register --auto-attach --username bgurung@redhat.com --password 'foo'bar' 
err     : exit status 1
output  : bash: -c: line 0: unexpected EOF while looking for matching `''
bash: -c: line 1: syntax error: unexpected end of file
. Retrying.
Waiting for SSH to be available...
```

gbraad:
> No it won't, but even you tell users then also they miss and only understand when face issue/failure :)

Although you are right about this, maybe including this as part of the troubleshooting would help. "My password is not accepted? Here is what to do..."

praveenkumar:
> Although you are right about this, maybe including this as part of the troubleshooting would help. "My password is not accepted? Here is what to do..."

Agree might be we need bit change from troubleshooting guide, we do have now if password contain special characters then do that...

budhrg:
@praveenkumar , verified that single quote is working when passed as `"foo'\"'\"'bar"`

> Although you are right about this, maybe including this as part of the troubleshooting would help. "My password is not accepted? Here is what to do..."

:+1: , Might be a note in already existing troubleshooting guide that if your password contain single quote you need to pass it as per your shell. For bash, pass as `"foo'\"'\"'bar"`.

Issue #718, opened 2017-04-10
Downloading initial ISO's should support HTTP proxy


donovanmuller:
While `minishift start` does support providing a http/https proxy for OpenShift itself, it should also be able to provide a (or use the same) proxy when `minishift` downloads the initial ISO's.

```console
$ minishift start \
  --http-proxy http://22.151.94.51:3128 \
  --https-proxy http://22.151.94.51:3128
Starting local OpenShift cluster using 'xhyve' hypervisor...
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso'
E0410 09:13:49.693645   21752 start.go:160] Error starting the VM: Get https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso: dial tcp 192.30.253.113:443: i/o timeout. Retrying.
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso'
```

Sabartius:
As a Workaround, is it possible to trick Minishift, that theses ISO-Files are already downloaded?
I downloaded the iso with my browser and put it next to the minishift binary, but it still tries to download it.

donovanmuller:
@Sabartius I haven't tried this but you might be able to copy the ISO into the `.minishift/cache/iso` folder:

```console
$ ll ~/.minishift/cache/iso
...
-rw-r--r--  1 donovan  staff    40M May  1 21:35 minishift-b2d.iso
```
as deleting the `minishift-b2d.iso` file will force a re-download.
I'm not too sure what else would need to be downloaded as there are other `.iso`'s in the `~/.minishift/machines/minishift` directory which I'm assuming also need to be downloaded...

hferentschik:
Basically two things need to be downloaded, the ISO as @donovanmuller mentioned, but also the _oc_ binary. If you download both manually and put them at their appropriate places in the cache directory, things should be fine.

donovanmuller:
@hferentschik I also noticed that the OpenShift Docker image will have to be pulled (e.g. `openshift/origin:v1.5.0`) initially on start.

```console
Starting local OpenShift cluster using 'xhyve' hypervisor...
-- Checking OpenShift client ... OK
-- Checking Docker client ... OK
-- Checking Docker version ...
   WARNING: Cannot verify Docker version
-- Checking for existing OpenShift container ... OK
[...]

PR #921, opened 2017-05-18
Issue #718 Downloading initial ISO and oc binary should support HTTP proxy

praveenkumar:
This patch will modify `http` client in case of proxy and use it to download iso and oc binary. Please test it, for me working as expected I can see traffic is going from proxy server.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -95,6 +95,7 @@ var (
 	registryMirror   []string
 	openShiftEnv     []string
 	shellProxyEnv    string
+	proxyUrl         string
 )
 
 // startCmd represents the start command
@@ -163,6 +164,7 @@ func runStart(cmd *cobra.Command, args []string) {
 		HostOnlyCIDR:     viper.GetString(hostOnlyCIDR),
 		OpenShiftVersion: viper.GetString(openshiftVersion),
 		ShellProxyEnv:    shellProxyEnv,
+		ProxyUrl:         proxyUrl,
 	}
 
 	fmt.Printf("Starting local OpenShift cluster using '%s' hypervisor...\n", config.VMDriver)
@@ -347,9 +349,11 @@ func setOcProxy() {
 func setShellProxy() {
 	if viper.IsSet(httpProxy) {
 		shellProxyEnv += fmt.Sprintf("http_proxy=%s ", viper.GetString(httpProxy))
+		proxyUrl = viper.GetString(httpProxy)
 	}
 	if viper.IsSet(httpsProxy) {
 		shellProxyEnv += fmt.Sprintf("https_proxy=%s ", viper.GetString(httpsProxy))
+		proxyUrl = viper.GetString(httpsProxy)
 	}
 	if viper.IsSet(noProxyList) {
 		shellProxyEnv += fmt.Sprintf("no_proxy=%s", viper.GetString(noProxyList))
@@ -441,6 +445,7 @@ func clusterUp(config *cluster.MachineConfig, ip string) {
 	oc := cache.Oc{
 		OpenShiftVersion:  config.OpenShiftVersion,
 		MinishiftCacheDir: filepath.Join(constants.Minipath, "cache"),
+		ProxyUrl:          proxyUrl,
 	}
 	if err := oc.EnsureIsCached(); err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintln("Error starting the cluster: ", err))
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -188,6 +188,7 @@ type MachineConfig struct {
 	HostOnlyCIDR     string // Only used by the virtualbox driver
 	OpenShiftVersion string
 	ShellProxyEnv    string // Only used for proxy purpose
+	ProxyUrl         string // Only used for proxy purpose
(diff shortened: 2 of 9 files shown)


praveenkumar:
> Why do all CI fail?

I do need to update unit test also, will do that and add as [WIP] for it.

cmd/minishift/cmd/start.go
```diff
@@ -347,9 +349,11 @@ func setOcProxy() {
 func setShellProxy() {
 	if viper.IsSet(httpProxy) {
 		shellProxyEnv += fmt.Sprintf("http_proxy=%s ", viper.GetString(httpProxy))
+		proxyUrl = viper.GetString(httpProxy)
 	}
 	if viper.IsSet(httpsProxy) {
 		shellProxyEnv += fmt.Sprintf("https_proxy=%s ", viper.GetString(httpsProxy))
+		proxyUrl = viper.GetString(httpProxy)
```
LalatenduMohanty: should be viper.GetString(httpsProxy)

pkg/minikube/cluster/cluster.go
```diff
@@ -214,7 +215,13 @@ func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 	fmt.Println(fmt.Sprintf("Downloading ISO '%s'", m.MinikubeISO))
 
 	// store the iso inside the MINISHIFT_HOME dir
-	response, err := http.Get(m.MinikubeISO)
+	myClient := http.DefaultClient
```
LalatenduMohanty: 👍

gbraad: Code duplication. This is a candidate for a general download function that can be part of our utility library

```
Get(uri string, proxy ProxyInfo) error {}
```

pkg/util/github/github.go
```diff
@@ -120,13 +121,18 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
...
-	asset, url, err := client.Repositories.DownloadReleaseAsset("openshift", "origin", assetID)
+	asset, uri, err := client.Repositories.DownloadReleaseAsset("openshift", "origin", assetID)
 	if err != nil {
 		return errors.Wrap(err, fmt.Sprintf("Cannot download OpenShift release asset %d", assetID))
 	}
-	if len(url) > 0 {
+	if len(uri) > 0 {
 		fmt.Println(fmt.Sprintf("Downloading OpenShift binary '%s' version '%s'", binaryType.String(), *release.TagName))
-		httpResp, err := http.Get(url)
+		myClient := http.DefaultClient
```
gbraad: See above; code duplication

gbraad:
BTW, tested on Windows with Polipo as a caching server. Didn't want to dust off and test my Squid setup skills... but worked as expected.

pkg/util/github/github.go
```diff
@@ -120,13 +121,14 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
...
-	asset, url, err := client.Repositories.DownloadReleaseAsset("openshift", "origin", assetID)
+	asset, uri, err := client.Repositories.DownloadReleaseAsset("openshift", "origin", assetID)
 	if err != nil {
 		return errors.Wrap(err, fmt.Sprintf("Cannot download OpenShift release asset %d", assetID))
 	}
-	if len(url) > 0 {
+	if len(uri) > 0 {
 		fmt.Println(fmt.Sprintf("Downloading OpenShift binary '%s' version '%s'", binaryType.String(), *release.TagName))
-		httpResp, err := http.Get(url)
+		myClient := util.GetHttpClient(proxyUrl)
```
gbraad: Ah, I see what you did.
For some reason, I do think this code is more complex than it needs to be. Readability is not very nice...

LalatenduMohanty:
Changes looks good to me except that the openshift_caching.go is not used in the code. Had a discussion with @praveenkumar and he will create a followup issue for removing it. cc @hferentschik

praveenkumar:
@LalatenduMohanty @hferentschik #937 Issue created.

hferentschik:
@praveenkumar have you tried to configure the default transport? See also https://stackoverflow.com/questions/14661511/setting-up-proxy-for-http-client. This is probably even better than env variables. In fact this is what I had in mind first. If the proxy variables are set, we configure the default transport, so while the Minisfhit process is running, all requested HTTP clients (unless explicitly configured) use the default transport. This will then also work with the upcoming _update_ command which will also need to support proxies.

hferentschik:
More concretely, I would expect that we make this call when proxies are specified via the CLI

    http.DefaultTransport = &http.Transport{Proxy: http.ProxyURL(proxyUrl)}

praveenkumar:
@hferentschik Changing default transport is working (I did test it out) now how we should go about to add it. Should we revert committed code and then again send PR?

hferentschik:
> Changing default transport is working

There is a problem with it so, in terms how it deals with HTTP vs HTTPS proxies. I think the env approach is better, since it transparently makes the right decision on which proxy to use. 

 > (I did test it out) now how we should go about to add it. Should we revert committed code and then again send PR?

I think the best is to add this commit https://github.com/minishift/minishift/pull/942/commits/d046e2e24971dbca035f5742626dc2fe30b9811c which inverts the changes and either add another commit with the new implementation or even squash in a single commit.

Issue #906, opened 2017-05-16
nip.io status for deployed application


amitkrout:
In the current integration test we have ruby-ex application deployment scenario. Here the missing part is the response status of the deployed application over the HTTP port to the public domain and is not yet implemented.

In the existing code [1], url can easily be retrieved just by calling the function ```func GetServiceURL```. So the test scenario would look something like this

```cucumber
Scenario: User can create route for ruby-ex to make it visiable outside of the cluster
    When executing "oc expose svc/ruby-ex" succeeds
    Then stdout should contain
     """
     exposed
     """
    When accessing application "ruby-ex" entry point of namespace "ruby" with https "true" succeeds
    Then stdout should contain
     """
     200 OK
     """ 
```
[1] https://github.com/minishift/minishift/blob/master/pkg/minishift/openshift/service.go

budhrg:
>  When accessing application "ruby-ex" entry point of namespace "ruby" with https "true" succeeds

@amitkrout , is your problem of namespace solved ?

> 200 OK 

This depends on the implementation of your above step definition. What is expected here? Do you have the code staged in any of your branch?

agajdosi:
@amitkrout For http request you can use `func verifyHTTPResponse(partOfResponse, url, urlSuffix, assertion, expected string) error {}`, it got merged few hours ago. With current implementation you can check for both body and status code.

amitkrout:
@agajdosi i see. i have not checked that, definitely have a look to that 

@budhrg No the namespace problem is not yet solved. Currently i am doing some more investigation on it. There migit be a modification in step defination implimentation. Will push it to my local branch once i get a conclusion on that. Also i need to check what @agajdosi suggesting

hferentschik:
> When accessing application "ruby-ex" entry point of namespace "ruby" with https "true" succeeds

I don't understand this step? What does "entry point of namespace "ruby" with https "true" mean? I get "When accessing application "ruby-ex" which might be even clearer by "When accessing route for application "ruby-ex"", but what is the rest?

hferentschik:
> For http request you can use func verifyHTTPResponse(partOfResponse, url, urlSuffix, assertion, expected string) error {}, it got merged few hours ago. With current implementation you can check for both body and status code.

+1, at least have a look at it. There is for sure communality.

amitkrout:
[...]

PR #925, opened 2017-05-19
Issue #906 nip.io status for deployed application

amitkrout:


amitkrout:
This pr comprises only nio.io access to the deployed application. ```get route -o=custom-columns=URL:.spec.host``` is used to retrieve the routing url and further validation done as per the step definition ```s.Step(`(body|status code) of HTTP request to "([^"]*)" (contains|is equal to) "([^"]*)"`, verifyHTTPResponse)```

test/integration/features/basic.feature
```diff
@@ -147,6 +147,12 @@ Feature: Basic
      """
      exposed
      """
+
+  Scenario: User can access ruby-ex routing url
```
hferentschik: "User can access ruby-ex routing url" = "User can access ruby-ex route"

test/integration/features/basic.feature
```diff
@@ -147,6 +147,12 @@ Feature: Basic
...
+   Given Minishift has state "Running"
+     When status code of HTTP request to "Routing" is equal to "200"
+     Then body of HTTP request to "Routing" contains "ok"
```
hferentschik: That does not make sense here, or? Is this a cut&paste error from the health check scenario?

test/integration/features/basic.feature
```diff
@@ -147,6 +147,12 @@ Feature: Basic
      """
      exposed
      """
+
+  Scenario: User can access ruby-ex routing url
+   Given Minishift has state "Running"
+     When status code of HTTP request to "Routing" is equal to "200"
```
hferentschik: This really needs to be a parameter. How does one know which route to access? There might and will be more than one at some times.

test/integration/minishift.go
```diff
@@ -28,6 +28,11 @@ import (
 
 var lastCommandOutput CommandOutput
 
+const (
+	ConsoleURL   = "console --url"
```
hferentschik: Not sure whether I am a fan of using constants here. Are you planning to re-use these?

test/integration/minishift.go
```diff
@@ -105,6 +110,13 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
 }
 
 func (m *Minishift) getOpenShiftUrl() string {
-	cmdOut, _, _ := m.runner.RunCommand("console --url")
+	cmdOut, _, _ := m.runner.RunCommand(ConsoleURL)
 	return strings.TrimRight(cmdOut, "\n")
 }
+
+func (m *Minishift) getRoutingUrl() string {
```
hferentschik: _getApplicationRoute_ is probably a better name. That said, how can this even work w/o a parameter. You need to know the route or application name.

test/integration/integration_test.go
```diff
@@ -259,6 +261,8 @@ func verifyHTTPResponse(partOfResponse, url, urlSuffix, assertion, expected stri
 	switch url {
 	case "OpenShift":
 		url = minishift.getOpenShiftUrl() + urlSuffix
+	case "Routing":
```
hferentschik: Just "Routing" is not enough, it would need to be something like "route for <app-name>"

test/integration/minishift.go
```diff
@@ -105,6 +110,13 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
...
+	ocRunner := m.runner.GetOcRunner()
+	cmdOut, _, _ := ocRunner.RunCommand(URLCustomCol)
+	url := strings.Split((cmdOut), "\n")[1]
+	return "http://" + url
```
hferentschik: You can get the full url directly via the Minishift command "minishift openshift service ruby-ex -n ruby --url", but of course you need the the application name and project

test/integration/features/basic.feature
hferentschik: As it happens to be, the content contains accidentally "ok" as well, that's why it works

test/integration/features/basic.feature
amitkrout: Done

test/integration/features/basic.feature
amitkrout: Agree. In this case the step definition uses ```getRoutingUrl()``` and the constant ```get route -o=custom-columns=URL:.spec.host``` to get the route of deployed application. As ruby-ex is the only app deployed in this scenario, we wont get any problem in this case.  But its good idea to pass the argument to make it more specific.

test/integration/features/basic.feature
amitkrout: These steps are executed in namespace ```ruby``` and these steps are derived from same function ```verifyHTTPResponse```. Will check how it will be used in more specific way.

test/integration/minishift.go
amitkrout: Its not about reusing. IMO its a good practice to create a constant variable rather expanding the value inside the function. As the features are growing there might be a chance to reuse it in a different function.

test/integration/minishift.go
amitkrout: Done. As per the current implementation i can get the routing url  just be executing this ```get route -o=custom-columns=URL:.spec.host``` if there is only one application is deployed. Will check how it will be done in a better way.

test/integration/minishift.go
amitkrout: This command ```"minishift openshift service ruby-ex -n ruby --url"``` looks more specific. Will check it

test/integration/features/basic.feature
hferentschik: It is also about the term "Routing". It really does not make sense in this context. The application exposes a service as a route. That's the terminology really. One cannot make a request to "Routing".

test/integration/features/basic.feature
hferentschik: "body of HTTP request to (.*) contains", is a generic step, which allows you to make a HTTP request and make an assertion about the content of the response. "ok" makes sense for the health check earlier on this feature file, since the health check URL will literally just return "ok". Your test only works since the demo webpage contains accidentally the substring "ok".

test/integration/minishift.go
hferentschik: It requires one more lookup though. One additional level of indirection.

test/integration/minishift.go
hferentschik: But this only works, if there is only one app deployed. This will break once we have multiple routes. One needs to specify the route or application name.

test/integration/minishift.go
hferentschik: ok

test/integration/features/basic.feature
gbraad: Not formatted correctly... indentation is off

test/integration/minishift.go
```diff
@@ -28,6 +28,11 @@ import (
 
 var lastCommandOutput CommandOutput
 
+const (
```
gbraad: do not see a real relevance for both constants

test/integration/minishift.go
```diff
@@ -105,6 +110,13 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
...
 
 func (m *Minishift) getOpenShiftUrl() string {
-	cmdOut, _, _ := m.runner.RunCommand("console --url")
+	cmdOut, _, _ := m.runner.RunCommand(ConsoleURL)
 	return strings.TrimRight(cmdOut, "\n")
 }
+
+func (m *Minishift) getRoutingUrl() string {
+	ocRunner := m.runner.GetOcRunner()
+	cmdOut, _, _ := ocRunner.RunCommand(URLCustomCol)
```
gbraad: const usage... with a very unclear name

test/integration/features/basic.feature
amitkrout: I modified the steps in this way
```
Scenario: User can access ruby-ex route
    When status code of HTTP request to access "ruby-ex" application url in namespace "ruby" is equal to "200"
    Then body of HTTP request to access "ruby-ex" application url in namespace "ruby" contains "Welcome to your Ruby application on OpenShift"
     And body of HTTP request to access "ruby-ex" application url in namespace "ruby" contains "<title>Welcome to OpenShift</title>"
 ```

File: test/integration/features/basic.feature (hunk 2 above)
Comment (amitkrout) [inline test/integration/features/basic.feature]: Thanks @hferentschik. I did not realize that. I modified the step
```
Then body of HTTP request to access "ruby-ex" application url in namespace "ruby" contains "Welcome to your Ruby application on OpenShift"
```

File: test/integration/minishift.go (hunk 4 above)
Comment (amitkrout) [inline test/integration/minishift.go]: True..I removed it

File: test/integration/minishift.go (hunk 5 above)
Comment (amitkrout) [inline test/integration/minishift.go]: Done

File: test/integration/minishift.go (hunk 7 above)
Comment (amitkrout) [inline test/integration/minishift.go]: Done

File: test/integration/minishift.go (hunk 8 above)
Comment (amitkrout) [inline test/integration/minishift.go]: I removed and keep it within the function argument.

File: test/integration/minishift.go (hunk 9 above)
Comment (amitkrout) [inline test/integration/minishift.go]: Commented as above (I removed and keep it within the function argument.)

File: test/integration/features/basic.feature [hunk 10]
```diff
@@ -147,6 +147,12 @@ Feature: Basic
      """
      exposed
      """
+
+  Scenario: User can access ruby-ex route
```
hferentschik: I think this reads really awkwardly and requires us to pass information which from a test perspective should not be needed. Let's back off a bit and look like how this should/could read:

    Scenario: User can access application via HTTP
        And "ruby-ex" exposes route "foo"
       Then status code of HTTP request to "/" of "foo" is equal to "200"
        And body of HTTP request to "/" of "foo" contains "Welcome to your   Ruby application on OpenShift"

This is imo concise. It also takes hight for the fact, that one might want to hit something else than the root context "/". 

What's about trying to aim for something like that?

test/integration/features/basic.feature
```diff
@@ -147,6 +147,12 @@ Feature: Basic
...
+   Given Minishift has state "Running"
+    When status code of HTTP request to access "ruby-ex" application url in namespace "ruby" is equal to "200"
+    Then body of HTTP request to access "ruby-ex" application url in namespace "ruby" contains "Welcome to your Ruby application on OpenShift"
+     And body of HTTP request to access "ruby-ex" application url in namespace "ruby" contains "<title>Welcome to OpenShift</title>"
```
hferentschik: What's the difference to the above? Why two content checks?

test/integration/features/basic.feature
agajdosi: +1

test/integration/features/basic.feature
gbraad: Ditto... why is this needed?

gbraad:
Approved my changes, but still have questions

test/integration/features/basic.feature
amitkrout: Here the idea of matching two content is to maximize the scenario accuracy. That does not mean that we should validate all the page content, but the minimum content need to be validated. Which would be enough to identify the HTTP response of the deployed application webpage. IMO these two parameters ```page title``` and some part of ```page body``` would be good enough to increase the response accuracy.

test/integration/features/basic.feature
amitkrout: Considering ```"/"``` as any service and ```foo``` as projectname 

If i understand this correctly, the aim is to get the HTTP response of a application deployed in the name space irrespective of service name.

> that one might want to hit something else than the root context "/".

@hferentschik i don't understand when you say hit something else. Are you referring it to any service in the same project ? bit of explanation would be really helpful

test/integration/features/basic.feature
hferentschik: I think verifying "Welcome to your Ruby application on OpenShift" is more than enough. We are really not interested what exactly is displayed.

test/integration/features/basic.feature
hferentschik: "/" is the root of the application, but maybe in some cases you want to access a URL like /welcome or you maybe want to get pass some parameters. It does not always have to be the root context. Is that clearer now?

amitkrout: Got it. Also it looks more generic. Thanks hardy for the explanation. Will check the implementation same here.

test/integration/features/basic.feature
amitkrout: ok

test/integration/features/basic.feature
amitkrout: I spend couple of days to implement these steps but had some difficulties in implementation though it looks more generic. 

> And "ruby-ex" exposes route "foo"

Here i am seeing extra layer of validation. Implementation would be look like this 
```
func (m *Minishift) exposeRoute(serviceName, nameSpace) error {
	cmdOut, cmdErr, cmdExit := m.runner.RunCommand("project")
	if len(cmdErr) != 0 || cmdExit != 0 || !strings.Contains(cmdOut, nameSpace){
		return fmt.Errorf("Command did not execute successfully or in the wrong namespace. cmdExit : %d, cmdErr: %s, cmdOut : %s", cmdExit, cmdErr, cmdOut) //Extra layre of validation
	} else {
		cmdOut, cmdErr, _ := m.runner.RunCommand("expose svc/" + serviceName)
		if len(cmdErr) != 0 {
			return fmt.Errorf("Command did not execute successfully. cmdErr: %s", cmdErr)
		}
	}
	return nil
}
```
To avoid this we can use When executing "oc expose svc/ruby-ex" succeeds as it was previously.
> Then status code of HTTP request to "/" of "foo" is equal to "200"
And body of HTTP request to "/" of "foo" contains "Welcome to your Ruby application on OpenShift"

I tried some of ways/tricks to implement but it hard to manage as part of argument is concerned though the scenario steps looks more generic. I used some of other commands such as ```-o=custom-columns=URL:.spec.host``` and ```-o=custom-columns=NAME:.metadata.name,HOST:.spec.host``` but did get the proper match of arguments to implement the step. 

i modified the feature file keeping rest changes intact. Scenario would look like this 
```
Scenario: User can create route for ruby-ex to make it visiable outside of the cluster
    When executing "oc expose svc/ruby-ex" succeeds
    Then stdout should contain
     """
     exposed
     """
    When status code of HTTP request to service "ruby-ex" of namespace "ruby" is equal to "200"
    Then body of HTTP request to service "ruby-ex" of namespace "ruby" contains "Welcome to your Ruby application on OpenShift"
```

test/integration/features/basic.feature
```diff
@@ -172,6 +172,8 @@ Feature: Basic
      """
      exposed
      """
+    When status code of HTTP request to service "ruby-ex" of namespace "ruby" is equal to "200"
```
hferentschik: **And** status code of HTTP ...
**And** body of HTTP request ...

hferentschik: I guess what is missing regarding my initial comment is to parameterise the path. Atm, you only hit "/" of the app. What if you wanted to hit "/welcome" or you want to verify that there is a welcome page "/welcome"?

    And status code of HTTP request "/" of service "ruby-ex" in namespace "ruby" is equal to "200"

You then just need to extract this additional path parameter and pass it to _verifyHTTPResponse_. You already pass something, namely the empty string (aka the root "/"). You have all the pieces in place. The step definition just needs to capture one more variable.

amitkrout: Done. Please let me know if it needs any further code change.

test/integration/minishift.go
```diff
@@ -112,3 +112,20 @@ func (m *Minishift) getOpenShiftUrl() string {
 	cmdOut, _, _ := m.runner.RunCommand("console --url")
 	return strings.TrimRight(cmdOut, "\n")
 }
+
+func (m *Minishift) getRoutingUrlAndVerifyHTTPResponse(partOfResponse, urlRootSuffix, serviceName, nameSpace, assertion, expected string) error {
```
hferentschik: This is really bad separation of concern. The Minishift struct should just return the URL, it should not call _verifyHTTPResponse_. 

Really this method should be:

    func (m *Minishift) getRoute(serviceName, nameSpace) (string, error)

In integration_test.go you then add another method for the step verification:

    s.Step(`(body|status code) of HTTP request to "([^"]*)" of service "([^"]*)" in namespace "([^"]*)" (contains|is equal to) "([^"]*)"$`, verifyServiceHTTPRequest)

This method can then call minishift.getRoute and verifyHTTPResponse. 

If you have a generic getRoute in the Minishift struct, this method can be used in another context as well. If you "merge" the concern of getting the route with the concern of executing the HTTP request and making an assertion, your code becomes much more coupled and less re-usable.

amitkrout: This method can then call minishift.getRoute and verifyHTTPResponse.
> If you have a generic getRoute in the Minishift struct, this method can be used in another context as well. If you "merge" the concern of getting the route with the concern of executing the HTTP request and making an assertion, your code becomes much more coupled and less re-usable.

+1. Implemented the changes. Would you please have a look.

Issue #783, opened 2017-04-20
Documentation about how users can upgrade Minishift


LalatenduMohanty:
We should add documentation about how users can upgrade minishift to a new release when they are running an older version. 

Need to come up with the steps, it might me trivial but will confirm after my testing.

hferentschik:
Given that we have a pull request for #72 almost ready, I recommend to not do anything specific for now. Once the pull request is merged, the upgrade path is 'minishift update'. Documentation should come via the pull request.

budhrg:
Agree :+1:

hferentschik:
Merged via pull request #930

PR #930, opened 2017-05-19
Issue #72 Update command, unit tests and docs

### LINKED ISSUE #72: Add update command

budhrg:
Fix #72  and #783 

Thanks @dharmit for the `minishift update` implementation.
Added tests.

```
$ minishift update -h
Checks for the latest version of Minishift and prompt the user and update the binary if user answers with 'y'.
...
```

```
$ minishift update
Do you want to update from 1.0.0 to 1.0.1 now? [y/N]: y
Downloading https://github.com/minishift/minishift/releases/download/v1.0.1/minishift-1.0.1-linux-amd64.tgz
 3.43 MiB / 3.43 MiB [============================================================================================================================] 100.00% 0s
 65 B / 65 B [====================================================================================================================================] 100.00% 0s

Updated to minishift v1.0.1 successfully.
```

[Update]
Tested with 
- `minishift start` with old version
- `minishift stop`
- `minishift update` to latest
- `minishift start`

https://gist.github.com/budhrg/bd1759ccb6d688a267e25043099af223


LalatenduMohanty:
Marked it WIP as @budhrg is going to add documentation for the update command

LalatenduMohanty:
@budhrg @dharmit I tested the PR and it is working for me and the PR looks good to me. Thanks

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,305 @@
...
+	downloadLinkFormat := "https://github.com/" + githubOwner + "/" + githubRepo + "/releases/download/v%s/%s"
+
+	url := fmt.Sprintf(downloadLinkFormat, latestVersion, archiveName)
+	downloadedArchivePath, err := downloadAndVerifyArchive(url)
+	if err != nil {
+		return err
+	}
+
+	// Get the directory containing saved archive
+	dir := strings.Split(downloadedArchivePath, "minishift")[0]
```
hferentschik: This is prone to break and makes assumptions about the code in _downloadAndVerifyArchive_. Let's create the tmp directory here and pass it to _downloadAndVerifyArchive_. This was we don't have to make assumptions and know exactly we directory we need to delete on defer.

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,305 @@
...
+	defer os.RemoveAll(dir)
+
+	// Extract and replace the downloaded archive in place of current binary
+	binary, err := extractBinary(downloadedArchivePath)
```
hferentschik: binary == binaryPath?

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,305 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = updateBinary(binary)
+	if err != nil {
+		return err
+	}
+
+	fmt.Printf("\nUpdated to minishift v%s successfully.\n", latestVersion)
```
hferentschik: Is the point of passing the stdout (aka a writer) to use it for any output?

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,305 @@
...
+	if localVersion.Compare(latestVersion) < 0 {
+		return true
+	} else {
+		return false
+	}
+}
+
+// Update handles the update process by calling functions that download, verify, extract and replace the binary.
+// It returns an error if any of these functions fails at any point.
+func Update(output io.Writer, latestVersion semver.Version) error {
```
hferentschik: _output_ is an unused parameter as it stands now

pkg/minishift/update/update.go
hferentschik: Personally, I'd remove this message and let the cmd print it after successful update (if there is no error). In which case the _output_ parameter can be deleted

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,305 @@
...
+// getLatestVersionFromGitHub gets the latest version of minishift available on GitHub.
+// It returns the version and error.
+func getLatestVersionFromGitHub(githubOwner, githubRepo string) (semver.Version, error) {
+	var (
+		release *github.RepositoryRelease
+		resp    *github.Response
+		err     error
+	)
+
+	client := githubutils.Client()
```
hferentschik: So here we go, this will need to go via proxies as well, right? So internally it seems GitHub utils uses "httpClient = http.DefaultClient". So the aim is to make this call use proxies. We might beed to add the proxy flags to the update command, unless we assume that the env variables are set. For consistency with the start command we probably want to add the proxy flags.

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,305 @@
...
+		}()
+	}
+
+	archiveBytes, err := ioutil.ReadAll(updatedArchive)
+	if err != nil {
+		return "", errors.New(fmt.Sprintf("Unable to read downloaded archive: %s", err))
+	}
+
+	// Temporary directory to store archive contents
+	dir, err := ioutil.TempDir("", "download")
```
hferentschik: Best to create the tmp dir on the highest level. See earlier comment

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,305 @@
...
+		fmt.Println()
+		bar := pb.New64(checksumResp.ContentLength).SetUnits(pb.U_BYTES)
+		bar.Start()
+		checksum = bar.NewProxyReader(checksum)
+		defer func() {
+			<-time.After(2 * bar.RefreshRate)
+		}()
+	}
+
+	// Verify checksum
```
hferentschik: I still think the checksum verification could be done much simpler with a TeeReader - see https://github.com/minishift/minishift/blob/d288208583282074afa56efe5a66004957d649b0/pkg/util/github/github.go#L150-L150

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,305 @@
...
+
+	return hash.Sum([]byte{}), nil
+}
+
+// extractBinary extracts the downloaded archive and returns path to the extracted binary file.
+// It returns an error if extraction fails.
+func extractBinary(downloadedArchive string) (string, error) {
+	var binary string
+
+	dir := strings.Split(downloadedArchive, "minishift")[0]
```
hferentschik: gain, fragile. Let's create one tmp directory and pass it along.

pkg/minishift/update/update_test.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+var (
+	archivePath string
+	err         error
+	port        int
+)
+
+func TestDownloadAndVerifyArchive(t *testing.T) {
+	port = randomNum(1000, 9999)
```
hferentschik: Why not use the same approach we use already for _github.go_ - https://github.com/minishift/minishift/blob/435f9bc64b8d1054b8ddd045c411e7a3a24beccc/pkg/testing/http_mock.go#L53-L53

Instead of starting a file server we modify the HTTPRoundTripper to return something. This way, we don't have to use a custom HTTP server and we also don't run the risk of port conflicts. In your code we can get random errors, if a port is selected which is already occupied. This will be heaps of fun to debug.

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,305 @@
...
+	update "github.com/inconshreveable/go-update"
+	"github.com/minishift/minishift/pkg/util/archive"
+	pb "gopkg.in/cheggaaa/pb.v1"
+
+	githubutils "github.com/minishift/minishift/pkg/util/github"
+	"github.com/minishift/minishift/pkg/version"
+)
+
+const (
+	timeLayout  = time.RFC1123
```
hferentschik: btw, I think this is not used either

pkg/minishift/update/update.go
hferentschik: This is probably the way to do the proxy config:

    http.DefaultTransport = &http.Transport{Proxy: http.ProxyURL(proxyUrl)}

Which can be put into a util method which can be called from the command itself prior to calling _Update_. 

See also https://github.com/minishift/minishift/pull/921#issuecomment-303184845

budhrg: Looks like it is now dependent on https://github.com/minishift/minishift/issues/718 for this implementation.

pkg/minishift/update/update.go
budhrg: Done

pkg/minishift/update/update.go
budhrg: Done :+1:

pkg/minishift/update/update.go
budhrg: Moved to `cmd/update.go`.

pkg/minishift/update/update.go
budhrg: Not sure here, should I wait for https://github.com/minishift/minishift/pull/942 ?

hferentschik: > Not sure here, should I wait for #942 ?

I think it would make sense. With the changes of #942 and the new ProxyConfig struct it should be trivial to do. Basically becomes something like:

    proxyConfig, err := util.NewProxyConfig(viper.GetString(httpProxy), viper.GetString(httpsProxy), viper.GetString(noProxyList))

	if err != nil {
	}

	proxyConfig.SetToEnvironment()

pkg/minishift/update/update.go
hferentschik: +1

pkg/minishift/update/update_test.go
gbraad: Agree, as the current solution can still some up with a conflicting port. Randomization without check is not sufficient.

pkg/minishift/update/update.go
budhrg: Done

pkg/minishift/update/update.go
budhrg: Done

pkg/minishift/update/update.go
budhrg: Done, added to `cmd/update.go`

pkg/minishift/update/update.go
budhrg: The current implementation looks simpler as well.

pkg/minishift/update/update.go
budhrg: Done

pkg/minishift/update/update.go
budhrg: Removed

budhrg:
@hferentschik , Implemented your suggestions along with proxy options.

hferentschik:
> Implemented your suggestions along with proxy options.

Cool. Checking. What happened to the docs?

cmd/minishift/cmd/update.go
```diff
@@ -0,0 +1,83 @@
...
+		}
+	} else {
+		fmt.Printf("Nothing to update.\nAlready using latest version: %s.\n", latestVersion)
+	}
+}
+
+func init() {
+	RootCmd.AddCommand(updateCmd)
+	updateCmd.Flags().String(httpProxy, "", "HTTP proxy used for downloading binary (In the format of http://<username>:<password>@<proxy_host>:<proxy_port>). Overrides a potential HTTP_PROXY setting in the enviroment.")
+	updateCmd.Flags().String(httpsProxy, "", "HTTPS proxy used for downloading binary (In the format of https://<username>:<password>@<proxy_host>:<proxy_port>). Overrides a potential HTTPS_PROXY setting in the enviroment.")
```
praveenkumar: @budhrg make sure you do proxy flags changes  as we discussed before we are ready to push.

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,292 @@
...
+// IsNewerVersion compares the local and latest versions and returns a boolean
+func IsNewerVersion(localVersion, latestVersion semver.Version) bool {
+	if localVersion.Compare(latestVersion) < 0 {
+		return true
+	} else {
+		return false
+	}
+}
+
+// Update handles the update process by downloading, verifying, extracting and
```
hferentschik: Very nice with the Godog. Much appreciated. Big plus :-)

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,292 @@
...
+	"github.com/minishift/minishift/pkg/version"
+)
+
+const (
+	githubOwner = "minishift"
+	githubRepo  = "minishift"
+)
+
+// CurrentVersion returns the current version of minishift binary installed on the system
+func CurrentVersion() (semver.Version, error) {
```
hferentschik: +1

pkg/minishift/update/update.go
```diff
@@ -0,0 +1,292 @@
...
+func LatestVersion() (semver.Version, error) {
+	latestVersion, err := getLatestVersionFromGitHub(githubOwner, githubRepo)
+	if err != nil {
+		return semver.Version{}, err
+	}
+	return latestVersion, nil
+}
+
+// IsNewerVersion compares the local and latest versions and returns a boolean
+func IsNewerVersion(localVersion, latestVersion semver.Version) bool {
```
hferentschik: +1 Nice, the "public" API of _update.go_ looks also goo now

cmd/minishift/cmd/update.go
```diff
@@ -0,0 +1,88 @@
...
+var updateCmd = &cobra.Command{
+	Use:   "update",
+	Short: "Update to latest version of Minishift.",
+	Long:  `Checks for the latest version of Minishift, prompt the user and update the binary if user answers with 'y'.`,
+	Run:   runUpdate,
+}
+
+func runUpdate(cmd *cobra.Command, args []string) {
+	proxyConfig := handleProxies()
+	proxyConfig, err := util.NewProxyConfig(viper.GetString(httpProxy), viper.GetString(httpsProxy), viper.GetString(noProxyList))
```
hferentschik: This is not quite right, you introduced _updateHttpProxy_ and _updatesHttpProxy_, but you are here referencing the constants from start.go. Also, you should not be using no proxy constant. I'll take care of this.

cmd/minishift/cmd/update.go
```diff
@@ -0,0 +1,88 @@
...
+
+var updateCmd = &cobra.Command{
+	Use:   "update",
+	Short: "Update to latest version of Minishift.",
+	Long:  `Checks for the latest version of Minishift, prompt the user and update the binary if user answers with 'y'.`,
+	Run:   runUpdate,
+}
+
+func runUpdate(cmd *cobra.Command, args []string) {
+	proxyConfig := handleProxies()
```
hferentschik: This is not correct here. This is a call into start.go specific code

docs/source/_topic_map.yml
```diff
@@ -18,6 +18,8 @@ Topics:
         File: docker-machine-drivers
       - Name: Uninstalling Minishift
         File: uninstalling
+      - Name: Updating Minishift
```
hferentschik: index.adoc needs updating as well

hferentschik:
@budhrg from the update side of things, the code and docs are looking good. The proxy flags are not working though. If you have a look into this, could you please base any further work on https://github.com/hferentschik/minishift/tree/issue-72, since I made some minor adjustments.

hferentschik:
@budhrg Have a look at this commit - https://github.com/hferentschik/minishift/commit/27bfd288a0ad8f8547a8964ba75dddeaf99495a8#commitcomment-22296004

I think this works. We might want to clean this up, but the gist is that you need to share the actual flag instance.

budhrg:
> Have a look at this commit - hferentschik/minishift@27bfd28#commitcomment-22296004

I was also looking into it but couldn't into first attempt. Checking your commit.

budhrg:
> Have a look at this commit - hferentschik/minishift@27bfd28#commitcomment-22296004

Got the exact changes but it seems not working. Not able to see http(s) flags for update though it is there in start command.

budhrg:
> I think this works. We might want to clean this up, but the gist is that you need to share the actual flag instance.

Working now.

hferentschik:
> Working now.

:-) Was there something missing or did you have some other issues? Good that it works now. As said, this might need some cleanup, but then it should be ok

budhrg:
check 
screenshot from 2017-05-26 21-34-59

budhrg:
retest this please

budhrg:
retest this please

hferentschik:
@budhrg  @dharmit  - Merged, thanks. Thanks for having a lot of patience. We went though many iterations on this, but I think the result was worth it.

budhrg:
> I think the result was worth it.

Yes :+1

Issue #922, opened 2017-05-18
Some Godog step definitions are never executed, since preceding more general steps will match first


agajdosi:
So I was facing an issue that sometimes integration tests were failing for me. Today I realized this is because of oc binary download which fails during startup due to exceeded API limit on Github. Issue got resolved for me with setting up MINISHIFT_GITHUB_API_TOKEN...

But even though `minishift start` failed, cucumber step `When executing "minishift start" succeeds` passed. Right now I am not sure if this is problem of only this one step or if it is problem for whole implementation of `executing "xyz" succeeds`.

Log:
```
[agajdosi@dhcp-24-156 minishift]$ make integration
go install -pkgdir=/home/agajdosi/golang/src/github.com/minishift/minishift/out/bindata -ldflags="-X github.com/minishift/minishift/pkg/version.version=1.0.1 -X github.com/minishift/minishift/pkg/version.isoVer
sion=v1.0.2 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v1.5.1" ./cmd/minishift
go test -timeout 3600s github.com/minishift/minishift/test/integration --tags=integration -v -args --binary /home/agajdosi/golang/bin/minishift ""
Running Integration test in: /tmp/minishift-integration-test-715862899
using binary: /home/agajdosi/golang/bin/minishift
....F-----................................F....F......F........F..F..F 70
..F--..F..F..F........                                                 92


--- Failed steps:

    Then  status code of HTTP request to "OpenShift" at "/healthz" is equal to "200" # features/basic.feature:16
      Error: Server returned error: Get https://192.168.42.192:8443/healthz: dial tcp 192.168.42.192:8443: getsockopt: connection refused, on url: https://192.168.42.192:8443/healthz

    Then  stdout should contain # features/basic.feature:57
      Error: Output did not match. Expected: sudoer, Actual:

    Then  stdout should contain # features/basic.feature:70
      Error: Output did not match. Expected: minishift, Actual:

    [...] [...] [...]

22 scenarios (11 passed, 11 failed)
92 steps (74 passed, 11 failed, 7 skipped)
57.30857841s
testing: warning: no tests to run
PASS
exit status 1
FAIL    github.com/minishift/minishift/test/integration 57.316s
Makefile:143: recipe for target 'integration' failed
make: *** [integration] Error 1
```
`/tmp/minishift-integration-test-715862899/cache/oc` is then empty.

praveenkumar:
[...]

PR #931, opened 2017-05-19
Issue #922 Step '"minishift start" succeeds' does not fail when oc download fails

agajdosi:
Addresses issue #922

Fixing logical mistake in executingMinishiftCommandSucceedsOrFails() and renaming `success` variable to `failure` which better suits the result of `lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr`.

Moved `s.Step(`executing "minishift ([^"]*)" (.*)$`, executingMinishiftCommandSucceedsOrFails)` before step for executingMinishiftCommand as this was matching first and due to this fact executingMinishiftCommandSucceedsOrFails() has never been actually used.

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -80,31 +80,31 @@ func FeatureContext(s *godog.Suite) {
 	minishift = &Minishift{runner: runner}
 
 	// steps to execute minishift commands
-	s.Step(`Minishift (?:has|should have) state "([^"]*)"`, minishift.shouldHaveState)
-	s.Step(`executing "minishift ([^"]*)"`, minishift.executingMinishiftCommand)
-	s.Step(`executing "minishift ([^"]*)" (.*)$`, executingMinishiftCommandSucceedsOrFails)
-	s.Step(`([^"]*) of command "minishift ([^"]*)" is equal to "([^"]*)"`, commandReturnEquals)
-	s.Step(`([^"]*) of command "minishift ([^"]*)" contains "([^"]*)"`, commandReturnContains)
+	s.Step(`Minishift (?:has|should have) state "([^"]*)"$`, minishift.shouldHaveState)
+	s.Step(`executing "minishift ([^"]*)"$`, minishift.executingMinishiftCommand)
+	s.Step(`executing "minishift ([^"]*)" (succeeds|fails)$`, executingMinishiftCommandSucceedsOrFails)
+	s.Step(`([^"]*) of command "minishift ([^"]*)" is equal to "([^"]*)"$`, commandReturnEquals)
+	s.Step(`([^"]*) of command "minishift ([^"]*)" contains "([^"]*)"$`, commandReturnContains)
 
 	// steps for running oc
 	s.Step(`executing "oc ([^"]*)" retrying (\d+) times with wait period of (\d+) seconds$`, minishift.executingRetryingTimesWithWaitPeriodOfSeconds)
-	s.Step(`executing "oc ([^"]*)`, minishift.executingOcCommand)
-	s.Step(`executing "oc ([^"]*)" succeeds$`, minishift.executingOcCommandSucceeds)
+	s.Step(`executing "oc ([^"]*)"$`, minishift.executingOcCommand)
+	s.Step(`executing "oc ([^"]*)" (succeeds|fails)$`, minishift.executingOcCommandSucceedsOrFails)
 
 	// steps to verify stdout and stderr of commands executed
-	s.Step(`([^"]*) should contain ([^"]*)`, commandReturnShouldContain)
-	s.Step(`([^"]*) should contain`, commandReturnShouldContainContent)
-	s.Step(`([^"]*) should equal ([^"]*)`, commandReturnShouldEqual)
-	s.Step(`([^"]*) should equal`, commandReturnShouldEqualContent)
-	s.Step(`([^"]*) should be empty`, commandReturnShouldBeEmpty)
-	s.Step(`([^"]*) should be valid ([^"]*)`, shouldBeInValidFormat)
+	s.Step(`([^"]*) should contain ([^"]*)$`, commandReturnShouldContain)
+	s.Step(`([^"]*) should contain$`, commandReturnShouldContainContent)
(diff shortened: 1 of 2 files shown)


budhrg:
add to whitelist

test/integration/integration_test.go
```diff
@@ -239,14 +239,14 @@ func executingMinishiftCommandSucceedsOrFails(command, expectedResult string) er
 	if err != nil {
 		return err
 	}
-	success := (lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr) != 0)
+	failure := (lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr) != 0)
```
LalatenduMohanty: 👍

test/integration/integration_test.go
```diff
@@ -239,14 +239,14 @@ func executingMinishiftCommandSucceedsOrFails(command, expectedResult string) er
...
 	switch expectedResult {
 	case "succeeds":
-		if success {
+		if failure {
```
LalatenduMohanty: The switch case seems confusing to me.

gbraad:
Seems to be an error sneaked in. This should never happen during a refactor...

but I am as confused about the switch case, due to the naming... probably better to break it down a little bit more... ?

test/integration/integration_test.go
gbraad:
```diff
244		case "succeeds":
- 245			if success {
3 days ago @agajdosi Issue #922 Step "executing ... succeeds" doesn't fail	
+ 245			if failure {
6 days ago @hferentschik Issue #548 Refactoring existing code	
246				return fmt.Errorf("Command did not execute successfully. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
13 days ago @agajdosi Issue #548 Expanding basic.feature	
247			}
6 days ago @hferentschik Issue #548 Refactoring existing code	
248		case "fails":
- 249			if success {
3 days ago @agajdosi Issue #922 Step "executing ... succeeds" doesn't fail	
+ 249			if !failure {
250				return fmt.Errorf("Command executed successfully, however was expected to fail. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)		  			return fmt.Errorf("Command executed successfully, however was expected to fail. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
```

agajdosi: @gbraad @LalatenduMohanty TBH I do not share the "confusing" feeling, so I am quite unsure how to resolve this problem, but I am trying it like this, does this look better to you?

```
commandFailed := (lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr) != 0)
if (expectedResult == "succeeds" && commandFailed == true) {
    return fmt.Errorf("Command did not execute successfully. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
}
if (expectedResult == "fails" && commandFailed == false) {
    return fmt.Errorf("Command executed successfully, however was expected to fail. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
}
```
And the case where expectedResult is not "succeeds" nor "fails" I will cover in regex matching with group: `(succeeds|fails)`.

test/integration/integration_test.go
```diff
@@ -81,8 +81,8 @@ func FeatureContext(s *godog.Suite) {
 
 	// steps to execute minishift commands
 	s.Step(`Minishift (?:has|should have) state "([^"]*)"`, minishift.shouldHaveState)
+	s.Step(`executing "minishift ([^"]*)" (succeeds|fails)$`, executingMinishiftCommandSucceedsOrFails)
```
hferentschik: Order should not matter. If it does we do something wrong. We might for example need to match the end of line in '`executing "minishift ([^"]*)"`'

hferentschik: +1, this is already better than what we had. Explicitly list the potential states. I would have gone even a step further and added to dedicated steps. This would be even more explicit. Code wise we can still share some implementation details.

test/integration/integration_test.go
```diff
@@ -239,18 +239,12 @@ func executingMinishiftCommandSucceedsOrFails(command, expectedResult string) er
...
-		if success {
-			return fmt.Errorf("Command did not execute successfully. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
-		}
-	case "fails":
-		if success {
-			return fmt.Errorf("Command executed successfully, however was expected to fail. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
-		}
-	default:
-		return fmt.Errorf("Expected result: %s not recognized, please use: 'fails' or 'succeeds'", expectedResult)
+	commandFailed := (lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr) != 0)
```
hferentschik: +1  _commandFailed_ is better

test/integration/integration_test.go
hferentschik: Let's for sure try to avoid having to depend on order here

hferentschik:
Let's properly address the fact that order matters here. It should not. If it does we need to change the conflicting steps to be non ambitious.

test/integration/integration_test.go
agajdosi: @hferentschik My understanding was that it matches only if it matches whole string, not just its part. It does not. So I guess it will be better to add `^` and `$` to every step matching? WDYT?

test/integration/integration_test.go
```diff
@@ -239,18 +239,12 @@ func executingMinishiftCommandSucceedsOrFails(command, expectedResult string) er
...
+	if expectedResult == "succeeds" && commandFailed == true {
```
gbraad: Glad to remove the switch (removes the nesting) +1

hferentschik:
Merged, thanks

Issue #409, opened 2017-02-15
Refactor start.go


hferentschik:
More and more functionality is pushed into _start.go_, especially the 'oc cluster up' startup. Similar to _cluster.go_ which contains the VM management functionality, the 'cluster up' functionality should move into its own place. 

This is to enforce the separation between command and actual functionality. Also it will make testing easier.

budhrg:
To add one which I came across recently is the [`clusterUp`](https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/start.go#L311) function is downloading and caching 'oc' binary (or oc installation) and then performing oc cluster up command which IMO are two separate and individual operations.

This is also making the testing a bit harder.

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/commit/4b9ddbf40f525d63a86eac087b1406997b1a5ce9

PR #942, opened 2017-05-22
Issue #718 Creating explicit struct for dealing with proxy settings

hferentschik:
This reverts d28820858. Before merging this, we should probably make sure that an alternative approach works, but otherwise I suggest to reopen issue #718, apply this patch to revert the changes and then implement a different solution.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -95,7 +95,6 @@ var (
 	registryMirror   []string
 	openShiftEnv     []string
 	shellProxyEnv    string
-	proxyUrl         string
 )
 
 // startCmd represents the start command
@@ -120,9 +119,6 @@ var startFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
 // clusterUpFlagSet contains the command line switches which needs to be passed on to 'cluster up'
 var clusterUpFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
 
-// proxyFlagSet contains the command line switches for proxy
-var proxyFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
-
 // subscription Manager FlagSet contains username and password details
 var subscriptionManagerFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
 
@@ -145,14 +141,11 @@ func runStart(cmd *cobra.Command, args []string) {
 	defer libMachineClient.Close()
 
 	validateOpenshiftVersion()
-	validateProxyArgs()
+	setSubscriptionManagerParameters()
 
-	setDockerProxy()
-	setOcProxy()
-	setShellProxy()
-	setSubcriptionManagerParameters()
+	proxyConfig := handleProxies()
 
-	config := cluster.MachineConfig{
+	machineConfig := cluster.MachineConfig{
 		MinikubeISO:      viper.GetString(isoURL),
 		Memory:           viper.GetInt(memory),
 		CPUs:             viper.GetInt(cpus),
@@ -164,16 +157,15 @@ func runStart(cmd *cobra.Command, args []string) {
 		HostOnlyCIDR:     viper.GetString(hostOnlyCIDR),
 		OpenShiftVersion: viper.GetString(openshiftVersion),
 		ShellProxyEnv:    shellProxyEnv,
-		ProxyUrl:         proxyUrl,
 	}
 
-	fmt.Printf("Starting local OpenShift cluster using '%s' hypervisor...\n", config.VMDriver)
+	fmt.Printf("Starting local OpenShift cluster using '%s' hypervisor...\n", machineConfig.VMDriver)
 
 	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
 
 	var host *host.Host
 	start := func() (err error) {
-		host, err = cluster.StartHost(libMachineClient, config)
+		host, err = cluster.StartHost(libMachineClient, machineConfig)
 		if err != nil {
 			glog.Errorf("Error starting the VM: %s. Retrying.\n", err)
 		}
@@ -205,14 +197,47 @@ func runStart(cmd *cobra.Command, args []string) {
 		atexit.Exit(1)
 	}
 
+	if proxyConfig.IsEnabled() {
+		// once we know the IP, we need to make sure it is not proxied in a proxy environment
+		proxyConfig.AddNoProxy(ip)
+		proxyConfig.ApplyToEnvironment()
+	}
+
 	automountHostfolders(host.Driver)
 
-	clusterUp(&config, ip)
+	clusterUp(&machineConfig, ip)
 
 	if !isRestart {
 		sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
 		postClusterUp(constants.MachineName, ip, constants.APIServerPort, viper.GetString(routingSuffix), minishiftConfig.InstanceConfig.OcPath, constants.KubeConfigPath, "developer", "myproject", sshCommander)
 	}
+}
+
+func handleProxies() *util.ProxyConfig {
+	proxyConfig, err := util.NewProxyConfig(viper.GetString(httpProxy), viper.GetString(httpsProxy), viper.GetString(noProxyList))
+
+	if err != nil {
+		atexit.ExitWithMessage(1, err.Error())
+	}
+
+	if proxyConfig.IsEnabled() {
+		proxyConfig.ApplyToEnvironment()
+		dockerEnv = append(dockerEnv, proxyConfig.ProxyConfig()...)
+		shellProxyEnv += strings.Join(proxyConfig.ProxyConfig(), " ")
+
+		// It could be that the proxy config is retrieved from the environment. To make sure that
+		// proxy settings are properly passed to cluster up we need to explicitly set the values.
+		if proxyConfig.HttpProxy() != "" {
+			viper.Set(httpProxy, proxyConfig.HttpProxy())
+		}
+
+		if proxyConfig.HttpsProxy() != "" {
+			viper.Set(httpsProxy, proxyConfig.HttpsProxy())
+		}
+		viper.Set(noProxyList, proxyConfig.NoProxy())
+	}
+
+	return proxyConfig
 }
 
 // postClusterUp runs the Minishift specific provisioning after cluster up has run
@@ -315,58 +340,6 @@ func automountHostfolders(driver drivers.Driver) {
 	}
 }
 
-// Set Docker Proxy
-func setDockerProxy() {
-	if viper.IsSet(httpProxy) {
-		dockerEnv = append(dockerEnv, fmt.Sprintf("HTTP_PROXY=%s", viper.GetString(httpProxy)))
-	}
-	if viper.IsSet(httpsProxy) {
-		dockerEnv = append(dockerEnv, fmt.Sprintf("HTTPS_PROXY=%s", viper.GetString(httpsProxy)))
-	}
-	if viper.IsSet(noProxyList) {
-		dockerEnv = append(dockerEnv, fmt.Sprintf("NO_PROXY=%s,%s", updateNoProxyForDocker(),
-			viper.GetString(noProxyList)))
-	} else if viper.IsSet(httpProxy) || viper.IsSet(httpsProxy) {
-		dockerEnv = append(dockerEnv, fmt.Sprintf("NO_PROXY=%s", updateNoProxyForDocker()))
-	}
-
-}
-
-// Set OpenShiftProxy
-func setOcProxy() {
-	if viper.IsSet(httpProxy) {
-		clusterUpFlagSet.String(httpProxy, viper.GetString(httpProxy), "HTTP proxy to use for master and builds")
-	}
-	if viper.IsSet(httpsProxy) {
-		clusterUpFlagSet.String(httpsProxy, viper.GetString(httpsProxy), "HTTPS proxy to use for master and builds")
-	}
-	if viper.IsSet(noProxyList) {
-		clusterUpFlagSet.String(noProxyList, viper.GetString(noProxyList), "List of hosts or subnets for which a proxy shouldn't be use")
-	}
-}
-
-// Set shell proxy
-func setShellProxy() {
-	if viper.IsSet(httpProxy) {
-		shellProxyEnv += fmt.Sprintf("http_proxy=%s ", viper.GetString(httpProxy))
-		proxyUrl = viper.GetString(httpProxy)
-	}
-	if viper.IsSet(httpsProxy) {
-		shellProxyEnv += fmt.Sprintf("https_proxy=%s ", viper.GetString(httpsProxy))
-		proxyUrl = viper.GetString(httpsProxy)
-	}
-	if viper.IsSet(noProxyList) {
-		shellProxyEnv += fmt.Sprintf("no_proxy=%s", viper.GetString(noProxyList))
-	} else if viper.IsSet(httpProxy) || viper.IsSet(httpsProxy) {
-		dockerEnv = append(dockerEnv, fmt.Sprintf("NO_PROXY=%s", updateNoProxyForDocker()))
-	}
-}
-
-// update default no-proxy for docker
-func updateNoProxyForDocker() string {
-	return "localhost,127.0.0.1,172.30.1.1"
-}
-
 // calculateDiskSizeInMB converts a human specified disk size like "1000MB" or "1GB" and converts it into Megabits
 func calculateDiskSizeInMB(humanReadableDiskSize string) int {
 	diskSize, err := units.FromHumanSize(humanReadableDiskSize)
@@ -380,12 +353,10 @@ func calculateDiskSizeInMB(humanReadableDiskSize string) int {
 func init() {
 	initStartFlags()
 	initClusterUpFlags()
-	initProxyFlags()
 	initSubscriptionManagerFlags()
 
 	startCmd.Flags().AddFlagSet(startFlagSet)
 	startCmd.Flags().AddFlagSet(clusterUpFlagSet)
-	startCmd.Flags().AddFlagSet(proxyFlagSet)
 	startCmd.Flags().AddFlagSet(subscriptionManagerFlagSet)
 
 	viper.BindPFlags(startCmd.Flags())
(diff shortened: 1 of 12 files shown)


budhrg:
@hferentschik , are you saying that the fix for this in PR https://github.com/minishift/minishift/pull/921 not working or am I missing anything here?

gbraad:
he disagrees with some of the code introduced, such as the passimg of variables using the struct. I do not believe a revert is really necessary, although a fix can be introduced

LalatenduMohanty:
> but I do not believe a revert is really necessary, although a fix can be introduced.

@hferentschik I agree with @gbraad .  Why not introduce a fix ?

hferentschik:
> are you saying that the fix for this in PR #921 not working or am I missing anything here?

It is working, but the implementation is unfortunate, since it forces you to pass the proxy parameters through so many layers.

hferentschik:
>  I agree with removing the variables from the struct, and instead rely on the environment variables in the getHttpClient function. 

Well, I think there is even a solution which does not require a _ getHttpClient_ method at all. I think I have the solution.

> but I do not believe a revert is really necessary, although a fix can be introduced.

I am not planning to "just" revert the changes. The plan is to introduce a more suitable solution which does not force you to pass the proxy settings though all these layers breaking encapsulation and separation of concerns. I am going to add to add another commit which add a different approach at which case we want to make sure that the unnecessarily introduced variables get reverted. This commit and pull request will make sure this happens.

hferentschik:
>  Why not introduce a fix ?

that's what I am working on, but the fix will not have much in common with the current code and I want to make sure no obsolete code is left behind.

hferentschik:
Updated pull request with a new way of dealing with proxies. Most importantly there is a dedicated struct to deal with this. 

Also, environment variables are used to make sure http.Client gets the right proxy config.

gbraad:
This task is still assigned to @praveenkumar. Praveen, did you come up with a solution?

pkg/util/proxy.go
```diff
@@ -0,0 +1,115 @@
...
+package util
+
+import (
+	"fmt"
+	"github.com/pkg/errors"
+	"os"
+	"strings"
+)
+
+var defaultNoProxies = []string{"localhost", "127.0.0.1", "172.30.1.1"}
```
gbraad: why 172.30 ?

pkg/util/proxy.go
```diff
@@ -0,0 +1,115 @@
...
+
+func (p *ProxyConfig) NoProxy() string {
+	return strings.Join(p.noProxy, ",")
+}
+
+func (p *ProxyConfig) AddNoProxy(exclude string) {
+	p.noProxy = append(p.noProxy, exclude)
+}
+
+func (p *ProxyConfig) SetToEnvironment() {
```
gbraad: SetToEnvironment` grammatically indicates that internal parameters take the value from the environment variables?

I believe this is meant to be `SetEnvironmentVariables` ?

pkg/util/utils.go
```diff
@@ -71,7 +69,7 @@ func Retry(attempts int, callback func() error) (err error) {
 	return RetryAfter(attempts, callback, 0)
 }
 
-func ValidateProxyURI(uri string) bool {
+func ValidProxyURI(uri string) bool {
 	return govalidator.IsURL(uri)
```
gbraad: Proxy settings should 'always' specify a port, as there is no real default and we can not assume this; IsURL will not enforce this.

pkg/util/utils_test.go
```diff
@@ -70,8 +70,8 @@ func TestValidateProxyURI(t *testing.T) {
 		"https://foo:b@r@test.com:454": true,
 	}
 	for uri, val := range urlList {
-		if ValidateProxyURI(uri) != val {
-			t.Fatalf("Expected '%t' Got '%t'", val, ValidateProxyURI(uri))
+		if ValidProxyURI(uri) != val {
```
gbraad: urlList does not cover enough failing cases. In the test all cases include port specification.

cmd/minishift/cmd/start.go
```diff
@@ -145,12 +141,18 @@ func runStart(cmd *cobra.Command, args []string) {
 	defer libMachineClient.Close()
 
 	validateOpenshiftVersion()
-	validateProxyArgs()
 
-	setDockerProxy()
-	setOcProxy()
-	setShellProxy()
-	setSubcriptionManagerParameters()
+	proxyConfig, err := util.NewProxyConfig(viper.GetString(httpProxy), viper.GetString(httpsProxy), viper.GetString(noProxyList))
```
gbraad: Could this be moved to a separate function? We keep stuffing everything into the `runStart` and this function is becoming a monster. This can be handled outside of #409.

cmd/minishift/cmd/start.go
```diff
@@ -145,12 +141,18 @@ func runStart(cmd *cobra.Command, args []string) {
...
+
+	if err != nil {
+		atexit.ExitWithMessage(1, err.Error())
+	}
+
+	proxyConfig.SetToEnvironment()
```
gbraad: ditto...

praveenkumar:
> This task is still assigned to @praveenkumar. Praveen, did you come up with a solution?

I did try what @hferentschik suggested and updated https://github.com/minishift/minishift/pull/921#issuecomment-303383847

pkg/util/utils_test.go
gbraad: urlList should check the cases described here:

```
      --http-proxy string               HTTP proxy for virtual machine (In the format of http://<username>:<password>@<proxy_host>:<proxy_port>)
      --https-proxy string              HTTPS proxy for virtual machine (In the format of https://<username>:<password>@<proxy_host>:<proxy_port>)
```

this means, `username:password@` should be tested for

gbraad:
Check. @hferentschik has now been assigned to #718.

Tested on Windows (cmd and powershell) and on Linux. Both work

pkg/util/proxy.go
hferentschik: correct, that's the OpenShift registry.

> might be good to include a const, with a clear name, and include it as such in the defaultNoProxies

fair enough

pkg/util/utils.go
hferentschik: true. I basically kept the existing validation. That said, I checked against Go's internal checks there and they also just verify that the URL swell formed. They don't enforce a port.

pkg/util/utils_test.go
hferentschik: +1, I'll add some tests

cmd/minishift/cmd/start.go
```diff
@@ -315,58 +329,6 @@ func automountHostfolders(driver drivers.Driver) {
...
-	if viper.IsSet(httpProxy) {
-		dockerEnv = append(dockerEnv, fmt.Sprintf("HTTP_PROXY=%s", viper.GetString(httpProxy)))
-	}
-	if viper.IsSet(httpsProxy) {
-		dockerEnv = append(dockerEnv, fmt.Sprintf("HTTPS_PROXY=%s", viper.GetString(httpsProxy)))
-	}
-	if viper.IsSet(noProxyList) {
-		dockerEnv = append(dockerEnv, fmt.Sprintf("NO_PROXY=%s,%s", updateNoProxyForDocker(),
-			viper.GetString(noProxyList)))
-	} else if viper.IsSet(httpProxy) || viper.IsSet(httpsProxy) {
```
hferentschik: @praveenkumar here is a question for you. I basically tried to keep all of the semantics of the existing code, except for the part here. In your code, if --no-proxy is specified only this value is used for NO_PROXY and the default values are not applied. I think this is actually wrong and a bug which just had not surfaced yet. IMO, the default no proxy list needs to be added as soon as a proxy is used and they are additional to any explicitly specified no proxy hosts. WDYT?

pkg/util/utils.go
hferentschik: It seems some tools imply 1080 (curl for example). Also one might use the default port 80. Looking at the implementation of Transport in Go, the port stays completely opaque. If a proxy url is specified, it just tries to connect to this url.

That said, this does not have to stop us from adding constraints and enforcing a port to be specified. Just not sure whether it makes sense.

pkg/util/proxy.go
```diff
@@ -0,0 +1,139 @@
...
+	config := ProxyConfig{
+		httpProxy:  httpProxy,
+		httpsProxy: httpsProxy,
+		noProxy:    np,
+	}
+
+	return &config, nil
+}
+
+// ProxyConfig returns a the proxy configuration as a slice, one element for each of the potential settings HTTP_PROXY,
```
budhrg: Move `HTTP_PROXY` to next line for readability with other variables.

pkg/util/proxy.go
```diff
@@ -0,0 +1,139 @@
...
+		os.Setenv("HTTP_PROXY", p.httpProxy)
+	}
+	if p.httpsProxy != "" {
+		os.Setenv("HTTPS_PROXY", p.httpsProxy)
+	}
+	os.Setenv("NO_PROXY", p.NoProxy())
+}
+
+// Enabled returns true if at least one proxy (HTTP or HTTPS) is configured. Returns false otherwise.
+func (p *ProxyConfig) Enabled() bool {
```
budhrg: `Enabled()` to `IsEnabled()`  ?

pkg/util/proxy_test.go
```diff
@@ -0,0 +1,145 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
```
budhrg: Wrong licence header and year. It should of Red Hat as `proxy` is of Red Hat and 2017 year

pkg/util/utils.go
gbraad: there is no real 'default' port, as this could also be `3128` (squid). checking could introduce other issues, I understand; should it end with `/`, what if it contains a resource as `.php`/`.html`?

pkg/util/proxy.go
gbraad: @hferentschik this is set to the process itself, and not to the parent shell... should this somehow be mentioned in the function name? `SetEnvironmentVariablesForProcess`... since I want to document somehow that this function does NOT have the side-effect of polluting the shell environment that started the minishift command.

cmd/minishift/cmd/start.go
praveenkumar: @hferentschik you are right `--no-proxy` also should be used for `NO_PROXY` if proxy is specified. I might missed it when implementing this feature :( Good catch.

hferentschik: ok, thanks

praveenkumar:
Tested on Mac and works as expected 👍

cmd/minishift/cmd/start.go
```diff
@@ -422,13 +393,9 @@ func initClusterUpFlags() {
...
-}
-
-// initProxyFlags create the CLI flags which needs to be passed for proxy
-func initProxyFlags() {
-	proxyFlagSet.String(httpProxy, "", "HTTP proxy for virtual machine (In the format of http://<username>:<password>@<proxy_host>:<proxy_port>)")
-	proxyFlagSet.String(httpsProxy, "", "HTTPS proxy for virtual machine (In the format of https://<username>:<password>@<proxy_host>:<proxy_port>)")
-	proxyFlagSet.String(noProxyList, "", "List of hosts or subnets for which proxy should not be used.")
+	clusterUpFlagSet.String(httpProxy, "", "HTTP proxy used for downloading artefact and configure Docker as well as OpenShift (In the format of http://<username>:<password>@<proxy_host>:<proxy_port>). Overrides a potential HTTP_PROXY setting in the enviroment.")
+	clusterUpFlagSet.String(httpsProxy, "", "HTTPS proxy used for downloading artefact and configure Docker as well as OpenShift (In the format of https://<username>:<password>@<proxy_host>:<proxy_port>). Overrides a potential HTTPS_PROXY setting in the enviroment.")
+	clusterUpFlagSet.String(noProxyList, "", "List of hosts or subnets for which no proxy should be used.")
```
LalatenduMohanty: These are not entirely for cluster up as we use it in subscription manager. Is that the reason you changed it to clusterUpFlagSet

pkg/util/proxy.go
gbraad: Using a general function name like `Enabled()` can be seen as a function to turn functionality on, instead of requesting the state.

pkg/util/proxy.go
```diff
@@ -0,0 +1,154 @@
...
+	}
+}
+
+// AddNoProxy appends the specified host to the list of no proxied hosts.
+func (p *ProxyConfig) AddNoProxy(host string) {
+	p.noProxy = append(p.noProxy, host)
+}
+
+// Sets the current config as environment variables in the current process.
+func (p *ProxyConfig) SetToEnvironment() {
```
gbraad: Not clear in purpose: `SetEnvironmentVariables`

pkg/util/proxy_test.go
```diff
@@ -0,0 +1,178 @@
...
+		t.Fatalf("Unexpected no proxy list, expected '%s', but got '%s'", expectedNoProxy, proxyConfig.NoProxy())
+	}
+}
+
+func Test_validate_proxy_url(t *testing.T) {
+	urlList := map[string]bool{
+		"http://foo.com:3128":          true,
+		"htt://foo.com:3128":           false,
+		"http://127.0.0.1:3128":        true,
+		"http://foo:bar@test.com:324":  true,
```
gbraad: I expect more failing specifications: malformed or invalid username, wrong delimiter, etc.

pkg/util/proxy_test.go
gbraad: @budhrg please "Start review" and "Request changes"

LalatenduMohanty:
Tested with RHEL as well as B2D ISO , seems to work fine.

pkg/util/proxy_test.go
LalatenduMohanty: +1 lets add some more invalid URLs.

cmd/minishift/cmd/start.go
hferentschik: They do belong here. Adding them to this flag set allows me to get rid oh the awkward _setOcProxy_ which was just a way to make sure the flags are set in cluster up. We need it in clusterUpFalgSet, since the way we generate the parameters to pass to 'cluster up' is via iterating this set.

This has no bearing to the fact that the use the values somewhere else as well. This was and is handled via _shellProxyEnv_

pkg/util/proxy.go
hferentschik: hmm, not sure, in conjunction with the struct it belongs to, I actually prefer:
 
    proxyConfig.SetToEnvironment()

over the imo more generic 

    proxyConfig.SetEnvironmentVariables

I guess one could also use _applyToEnvironment_, would that be better?

pkg/util/proxy.go
hferentschik: Enabled != Enable 

My understanding is that the _is_ and _get_ prefixes are not so commonly used in Go as in say Java.

Personally I am trying to use let _get_ and _is_ method names these days, but if you feel strongly about this, I'll change it

pkg/util/proxy_test.go
hferentschik: As I mentioned before, this is a very weak validation here. All we do is to verify whether the string parses as URL. There is a whole lot there which will go through here which is probably not quite right. On the other hand this is aligned with the underlying implementation in the Transport.

cmd/minishift/cmd/start.go
LalatenduMohanty: @hferentschik sounds fine to me.

pkg/util/proxy.go
LalatenduMohanty: I think SetProxyToEnvironment would be better name as we are only setting proxy here. The function is also in proxy.go so I do not see it will be use for setting other environment variables in future.

gbraad: `ApplyProxy(Config)ToEnv(ironment)` is OK with me.

pkg/util/proxy.go
gbraad: I believe we all need to have a good look at a common or best practices for Golang. At the moment it looks we all have a .NET/C#, Java look at stuff. An agreed style guide would be appreciated. Suggestions to have a look at something of this nature?

/cc: @LalatenduMohanty @praveenkumar @budhrg

budhrg: IMO, `Enabled` sounds like "Enable is alway true". Adding `Is` as prefix modifies that thinking (at least to me :smile: ).

pkg/util/proxy.go
hferentschik: > I think SetProxyToEnvironment would be better name as we are only setting proxy here.

what we are setting is implicit, since we are calling the method on a ProxyConfig struct

>  The function is also in proxy.go

The filename is not a good measurement in Go, since it is not like in object oriented languages like Java, where one file most of the times implies one class. 

> Apply(Config)ToEnv(ironment) is OK with me

ApplyToEnvironment it is then.

> maybe using Proxy and/or Config again is not necessary, as this is in the name of the struct.

exactly

hferentschik: Done

pkg/util/proxy.go
hferentschik: I am going to change it to IsEnabled then. 

Regarding enforcing a style. I think we should be a bit flexible here. As long as the rest of the name makes sense I personally don't care so much whether we us _is_ or _get_ in a Java style manner. 
I have bigger issues with names which shoot ways past target.

hferentschik: renamed

pkg/util/proxy_test.go
gbraad: We will have to define testcases and actually make them fail appropriately long-term. Querystrings are allowed in the ValidURL check, but not for the proxy purpose. Does [ProxyURL]( https://golang.org/pkg/net/http/#ProxyURL) actually enforce something internally? If so, maybe we do not need these test (library code we do not have to test)?

pkg/util/proxy_test.go
```diff
@@ -0,0 +1,183 @@
...
+	"testing"
+)
+
+func Test_default_no_proxy_list(t *testing.T) {
+	proxyConfig, err := NewProxyConfig("http://foobar.com", "", "")
+	if err != nil {
+		t.Fatal("Unexpected error creating proxy config")
+	}
+
+	if proxyConfig.NoProxy() != "localhost,127.0.0.1,172.30.1.1" {
```
gbraad:
```go
const (
   registryIP = "172.30.1.1"
)
...
if proxyConfig.NoProxy() != "localhost,127.0.0.1,", registryIP {
```

Issue #939, opened 2017-05-22
Code should handle single quote in Password


praveenkumar:
Currently if password contain single quoted `'` then user need manually escape it using `'"'"'` which should be happened by code.

hferentschik:
Merged via pull request #943

PR #943, opened 2017-05-22
Issue #939 Add method to handle single quote string to password

praveenkumar:
I hope this will patch up all special character war in password string.

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -151,3 +151,8 @@ func GetHttpClient(proxyUrl string) *http.Client {
 	}
 	return myClient
 }
+
+func EscapeSingleQuote(s string) string {
+	r := strings.NewReplacer(`'`, `'"'"'`)
+	return r.Replace(s)
+}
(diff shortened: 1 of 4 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -537,7 +537,7 @@ func ocSupportFlag(cmdName string, flag string) bool {
 
 func setSubcriptionManagerParameters() {
 	cluster.RegistrationParameters.Username = viper.GetString(username)
-	cluster.RegistrationParameters.Password = viper.GetString(password)
+	cluster.RegistrationParameters.Password = util.EscapeSingleQuote(viper.GetString(password))
```
hferentschik: IMO this should be done here - https://github.com/minishift/minishift/blob/088338ecc7071f56c869eb2ffe7ec78df722f734/pkg/minishift/registration/redhat.go#L60-L60. This is just needed for using the password as part of this command.

pkg/util/utils.go
```diff
@@ -151,3 +151,8 @@ func GetHttpClient(proxyUrl string) *http.Client {
 	}
 	return myClient
 }
+
+func EscapeSingleQuote(s string) string {
+	r := strings.NewReplacer(`'`, `'"'"'`)
```
hferentschik: +1

praveenkumar:
> I take it you tested this end to end as well?

@hferentschik I did my best :)

hferentschik:
> I did my best :)

LOL, fair enough

Issue #843, opened 2017-04-28
Documentation related to virtual switch selection on Windows+Hyper-V


gbraad:
The documentation for the virtual switch selection can currently be found in the installing.adoc. I do not really believe this is the right location, unless we also link to the Note from either the troubleshooting guide or the running instructions.

Also, an example for `cmd` and/or `powershell` is preferred as the command itself can lead to a possible error. The name of the switch should be specified as entered in the Hyper-V manager, and NOT as the actually created virtual switch. Note: created as `External (Wireless)`, will lead to a bridge called `vEthernet (External (Wireless))` in the network adapters. Below you can see the difference in how to specify the name.

cmd:
```cmd
C:\> set HYPERV_VIRTUAL_SWITCH=External (Wireless)
... and not
C:\> set HYPERV_VIRTUAL_SWITCH="External (Wireless)" 
```

Notice that using quotes will result in errors of not finding the virtual switch:

      Error creating the VM. Error with pre-create check: "vswitch \"\\\"External (Wireless)\\\"\" not found"

powershell:
```powershell
PS> $env:HYPERV_VIRTUAL_SWITCH="External (Wireless)"
```

Related to: #590

Preeticp:
Considering that out of the 5 hypervisors Minishift supports, 2 are embedded in the Minishift and do not need additional installation steps(vbox and VM ware). Out of the other 3, KVM and Xhyve are covered in the Docker Machine Drivers doc. 
I suggest we expand the Docker Machine Drivers to include information about HyperV too. We can rename the topic to a generic - Installing Hypervisors. WDYT?

gbraad:
Yes... Or something like: "Setup hypervisors drivers"...

Preeticp:
Makes sense @gbraad

hferentschik:
> We can rename the topic to a generic - Installing Hypervisors. WDYT?

-1 for a rename. We choose this name explicitly, since it is the most appropriate one and is also the term which when used for example with Google gets you to the most relevant documentation.

Also a "hypervisor driver" is imo quite ambiguous.

hferentschik:
Merged via pull request #951

prbhadra:
We did setup an environment variable and getting the following error.

We have no clue on where things are going wrong.

C:\Users\prbha>echo %HYPERV_VIRTUAL_SWITCH%
External (Wireless)

C:\Users\prbha>minishift start
-- Starting profile 'minishift'
[...]

PR #951, opened 2017-05-24
Issue #843, expanded and renamed the docker machine driver doc to...

Preeticp:
…include information on Hyper-V.
Hyper-V information seemed to be expanding and as stated in issue #843, it was decided to go ahead with this approach.
I have updated the links to this doc in the Installation doc, index and topics doc and also made some grammatical corrections in these docs.

### DIFF
--- a/docs/source/_topic_map.yml
+++ b/docs/source/_topic_map.yml
@@ -14,8 +14,8 @@ Topics:
         File: installing
       - Name: Quickstart
         File: quickstart
-      - Name: Installing Docker Machine drivers
-        File: docker-machine-drivers
+      - Name: Setting up the driver plugin
+        File: setting-up-driver-plugin
       - Name: Uninstalling Minishift
         File: uninstalling
   - Name: Using Minishift
--- a/docs/source/getting-started/installing.adoc
+++ b/docs/source/getting-started/installing.adoc
@@ -16,10 +16,10 @@ The following section describes how to install Minishift and the required depend
 == Prerequisites
 
 Minishift requires a hypervisor to start the virtual machine on which the OpenShift cluster
-is provisioned. Make sure that the hypervisor of your choice is installed and enabled on
+is provisioned. Verify that the hypervisor of your choice is installed and enabled on
 your system before you start Minishift.
 
-Depending on your host OS, you have the choice of the following
+Depending on your host operating system, you have the choice of the following
 hypervisors:
 
 Mac OS X::
@@ -28,14 +28,14 @@ Mac OS X::
 - https://www.vmware.com/products/fusion[VMware Fusion]
 +
 NOTE: xhyve requires specific installation and configuration steps that are described
-in the link:../getting-started/docker-machine-drivers{outfilesuffix}[Installing Docker machine drivers] section.
+in the link:../getting-started/setting-up-driver-plugin{outfilesuffix}[Setting up the driver plugin] section.
 
 GNU/Linux::
 - https://en.wikipedia.org/wiki/Kernel-based_Virtual_Machine[KVM] (default)
 - https://www.virtualbox.org/wiki/Downloads[VirtualBox]
 +
 NOTE: KVM requires specific installation and configuration steps that are described
-in the link:../getting-started/docker-machine-drivers{outfilesuffix}[Installing Docker machine drivers] section.
+in the link:../getting-started/setting-up-driver-plugin{outfilesuffix}[Setting up the driver plugin] section.
 
 Windows::
 - https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/enable-hyper-v[Hyper-V] (default)
@@ -45,11 +45,7 @@ Windows::
 - To use Minishift with Hyper-V ensure that, after you
 https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/enable-hyper-v[install Hyper-V], you also
 https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/quick_start/walkthrough_virtual_switch[add a Virtual Switch]
-using the Hyper-V Manager.
-- Make sure that you pair the virtual switch
-with a __network card (wired or wireless) that is connected to the network__.
-- If you have multiple virtual switches , set the environment variable `HYPERV_VIRTUAL_SWITCH` to the name of the external virtual switch you want to use for Minishift. For further information see
-link:../using/managing-minishift{outfilesuffix}#driver-specific-environment-variables[setting driver specific environment variables].
+using the Hyper-V Manager. For specific configuration steps see the link:../getting-started/setting-up-driver-plugin{outfilesuffix}[Setting up the driver plugin] section.
 ====
 
 - https://www.virtualbox.org/wiki/Downloads[VirtualBox]
@@ -63,7 +59,6 @@ the link:../using/troubleshooting{outfilesuffix}[Troubleshooting] section.
 [[installing-instructions]]
 == Installing Minishift
 
-[[manual-install]]
 === Manually
 
 .  Download the archive for your operating system from the https://github.com/minishift/minishift/releases[Minishift releases page]
@@ -97,8 +92,7 @@ install the stable version of Minishift:
 [[homebrew-latest-install]]
 ==== Latest Beta
 
-If you want to install the latest beta version of Minishift you will
-need the homebrew-cask versions tap. After you install `homebrew-cask`,
+If you want to install the latest beta version of Minishift use the homebrew-cask versions tap. After you install `homebrew-cask`,
 run the following command:
 
 ----
--- a/docs/source/getting-started/index.adoc
+++ b/docs/source/getting-started/index.adoc
@@ -7,5 +7,5 @@ application deployment.
 
 - link:../getting-started/installing{outfilesuffix}[Installing Minishift]
 - link:../getting-started/quickstart{outfilesuffix}[Minishift Quickstart]
-- link:../getting-started/docker-machine-drivers{outfilesuffix}[Installing Docker Machine drivers]
+- link:../getting-started/setting-up-driver-plugin{outfilesuffix}[Setting up the driver plugin]
 - link:../getting-started/uninstalling{outfilesuffix}[Uninstalling Minishift]
--- a/docs/source/getting-started/setting-up-driver-plugin.adoc
+++ b/docs/source/getting-started/setting-up-driver-plugin.adoc
@@ -1,5 +1,5 @@
-[[docker-machine-driver-install]]
-= Installing Docker Machine drivers
+[[set-up-driver-plugin]]
+= Setting up the driver plugin
 :icons:
 :toc: macro
 :toc-title:
@@ -7,14 +7,13 @@
 
 toc::[]
 
-[[docker-machine-drivers-overview]]
+[[hypervisor-drivers-overview]]
 == Overview
-
 Minishift uses Docker Machine and its driver plugin architecture to
-provide a consistent way to manage the OpenShift VM. Minishift embeds
-VirtualBox and VMware Fusion drivers so no additional steps are required to use them.
+provide a consistent way to manage the OpenShift VM.
+Minishift embeds VirtualBox and VMware Fusion drivers so no additional steps are required to use them.
 
-However, other drivers require manual installation steps that are
+However, other drivers require manual installation and setup that are
 described in the following sections.
 
 [[kvm-driver-install]]
@@ -22,10 +21,15 @@ described in the following sections.
 
 Minishift is currently tested against `docker-machine-driver-kvm` version 0.7.0.
 
-Install and execute the KVM binary as follows:
-
+. Install the KVM binary as follows:
++
 ----
 $ sudo curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.7.0/docker-machine-driver-kvm -o /usr/local/bin/docker-machine-driver-kvm
+----
+
+. Make the binary executable as follows:
++
+----
 $ sudo chmod +x /usr/local/bin/docker-machine-driver-kvm
 ----
 
@@ -82,7 +86,7 @@ Minishift is currently tested against `docker-machine-driver-xhyve` version 0.3.
 [[homebrew-install]]
 === Homebrew install
 
-You can verify the installed version of the xhyve driver on your system before you install.
+You can verify the existing version of the xhyve driver on your system using:
 
 ----
 $ brew info --installed docker-machine-driver-xhyve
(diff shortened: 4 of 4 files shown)


gbraad:
Looks good. But missing examples to set the hyperv virtual switch? Maybe best to address in the same PR.

Preeticp:
Will do @gbraad

docs/source/_topic_map.yml
```diff
@@ -14,8 +14,8 @@ Topics:
         File: installing
       - Name: Quickstart
         File: quickstart
-      - Name: Installing Docker Machine drivers
-        File: docker-machine-drivers
+      - Name: Setting up the hypervisor drivers
```
hferentschik: why the rename from docker machine drivers to hypervisor drivers. We had a reason to stick to the term "Docker machine drivers". This is the proper term and also the term one needs to search for for finding appropriate information.

hferentschik: -1 from my side

docs/source/getting-started/Setting-up-hypervisor-driver.adoc
```diff
@@ -1,20 +1,20 @@
-[[docker-machine-driver-install]]
-= Installing Docker Machine drivers
+[[set-up-hypervisor-driver]]
+= Setting up the hypervisor drivers
```
hferentschik: -1 Same here, I am against this rename. We had this discussion and explicitly choose to use "Docker machine driver". If you enter this search string into Google, you get to the most appropriate documentation

docs/source/_topic_map.yml
LalatenduMohanty: Agree with @hferentschik we should stick to the old name. `docker-machine driver` is a string which is used in other projects and familiar to users.

docs/source/getting-started/Setting-up-hypervisor-driver.adoc
LalatenduMohanty: Same as above.

docs/source/getting-started/Setting-up-hypervisor-driver.adoc
```diff
@@ -1,20 +1,20 @@
...
 :icons:
 :toc: macro
 :toc-title:
 :toclevels: 1
 
 toc::[]
 
-[[docker-machine-drivers-overview]]
+[[hypervisor-drivers-overview]]
```
LalatenduMohanty: same as above

docs/source/_topic_map.yml
hferentschik: > docker-machine driver is a string which is used in other projects and familiar to users.

Exactly.

docs/source/getting-started/installing.adoc
```diff
@@ -62,8 +58,7 @@ the link:../using/troubleshooting{outfilesuffix}[Troubleshooting] section.
 
 [[installing-instructions]]
 == Installing Minishift
-
-[[manual-install]]
+that
```
hferentschik: there something went wrong, right?

docs/source/getting-started/Setting-up-hypervisor-driver.adoc
```diff
@@ -152,5 +152,17 @@ $ sudo chown root:wheel /usr/local/bin/docker-machine-driver-xhyve
...
 ====
+
+== Hyper-V
+To use Minishift with Hyper-V:
+
+. Install https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/enable-hyper-v[Hyper-V].
+. Add a https://msdn.microsoft.com/en-us/virtualization/hyperv_on_windows/quick_start/walkthrough_virtual_switch[Virtual Switch].
++
+NOTE: Verify that you pair the virtual switch
+with a __network card (wired or wireless) that is connected to the network__.
```
hferentschik: _connected to the network_ - I find this ambitious. To which _network_. My external LAN/WAN connection? Something which is connected to the internet?

docs/source/getting-started/Setting-up-hypervisor-driver.adoc
```diff
@@ -1,20 +1,20 @@
-[[docker-machine-driver-install]]
-= Installing Docker Machine drivers
+[[set-up-hypervisor-driver]]
```
hferentschik: regarding the filename, let's stick to all lowercase. In this case we should not even rename the file, just saying.

docs/source/_topic_map.yml
gbraad: I disagree, as Minikube calls this https://github.com/kubernetes/minikube/blob/master/docs/drivers.md "Driver plugin installation". The reasoning of the setup is related to the fact that also virtual switch settings are considered (due to Hyper-V). If at best, the page can be named: "Setting up the docker-machne drivers", but this is specifically referring to Docker and Docker machine, while these as technology should be mostly hidden from the user. Especially going forward, I believe we aregoing to pack them into the binary (if possible).

Even Docker Machine calls is Machine plugins and driver plugins to be precise: https://github.com/docker/docker.github.io/blob/master/machine/AVAILABLE_DRIVER_PLUGINS.md.

While I partially agree, I also disagree as much.

docs/source/getting-started/Setting-up-hypervisor-driver.adoc
gbraad: ditto

docs/source/getting-started/Setting-up-hypervisor-driver.adoc
gbraad: This has been part of our Installation instructions for already 2 months; https://github.com/minishift/minishift/blob/e1b5c98327f42d75b7feee3d4b9ad1078f492ea7/docs/source/getting-started/installing.adoc This originates from the Microsoft documentation: https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/quick-start/connect-to-network   "Connect to network" which details that to "Connect virtual machines to the internet" you need to do ... such and such, and statng in step 9: "Select the physical network card to be paired with the new virtual switch. This is the network card that is physically connected to the network." We choose not to differ from this

docs/source/getting-started/Setting-up-hypervisor-driver.adoc
gbraad: The order of using `setup` or `set-up` in the front of the filename is an improvement of having "install" at the end. 

Browsing the source tree for the docs from the computer or https://github.com/minishift/minishift/tree/master/docs/source is non-intuitive. 

Before:
install1

After:
install2

While I agree that the uppercase is not correct, the name indicates more meaning... `docker-machine-drivers.adoc` not so much

docs/source/_topic_map.yml
LalatenduMohanty: When I said projects I meant https://github.com/dhiltgen/docker-machine-kvm as @gbraad mentioned.   

> I'd rather align more with Minikube,

+1. So I take back what I said earlier and  +1 for `Driver plugin`

docs/source/getting-started/installing.adoc
Preeticp: Yes, will take care of it.

docs/source/_topic_map.yml
Preeticp: @gbraad has already mentioned this, just adding to it. The reasoning behind the rename was: Information wrt Hyper-V virtual switch was expanding. Installation doc did not seem to be the right place for this information. Given that we are already redirecting the users to the Docker Drivers doc section for other drivers, it made sense to add Hyper-V information in there too and provide a link to it from the Installation. However afaiu, please correct me if I am wrong, Hyper-V does not require a Docker machine driver, only KVM and xhyve need them, so I could not club Hyper-V under the same heading as Docker Machine Drivers. Hence the need for a rename. 
Based on above discussion it makes sense to give the doc a generic heading like "Setting up the Hypervisor driver plugin" as @gbraad suggested.
And maybe to be explicit, split the doc into one subheading talking about Hyper-V and another clubbing the KVM and xhyve info under the Docker Machine driver subheading. WDYT?

gbraad: I do not think splitting this up will be helpful, as it hides information that really belongs together. Although the driver plugins for KVM and xhyve need to be installed, they can also be configured/setup with additional options. At the moment we do not do this, but this might be added in future. The same would go for the virtualbox driver, although at this stage, no options are exposed.

docs/source/getting-started/Setting-up-hypervisor-driver.adoc
gbraad: With #718 merged and the coming #953, we only need the network to provide a DHCP address as the Internet part could be proxied.

docs/source/_topic_map.yml
Preeticp: Ok

Preeticp: Changed as per suggestions.

docs/source/_topic_map.yml
```diff
@@ -14,8 +14,8 @@ Topics:
         File: installing
       - Name: Quickstart
         File: quickstart
-      - Name: Installing Docker Machine drivers
-        File: docker-machine-drivers
+      - Name: Setting up the hypervisor driver plugin
+        File: setting-up-hypervisor-driver-plugin
```
budhrg: Why we are changing to `hypervisor driver plugin` ?

budhrg: I think I got it now after reading overview of "Setting up the hypervisor driver plugin" doc. Make sense IMO :+1:

docs/source/getting-started/setting-up-hypervisor-driver-plugin.adoc
```diff
@@ -1,24 +1,23 @@
...
+provide a consistent way to manage the OpenShift VM.
+Minishift embeds VirtualBox and VMware Fusion drivers so no additional steps are required to use them.
 
-However, other drivers require manual installation steps that are
+However, other drivers require manual installation and setup that are
 described in the following sections.
 
 [[kvm-driver-install]]
-== KVM driver
+=== KVM driver
```
budhrg: Adding triple `=` sign, remove it from table of content list. Check, the `View` button of this page.

budhrg: Also, under this section, please update following line: 
```
Install and execute the KVM binary as follows:
```
to 
```
Install the KVM binary and make it executable as follows:
```
as we are not executing the binary in the following step.

docs/source/getting-started/setting-up-hypervisor-driver-plugin.adoc
```diff
@@ -75,14 +74,14 @@ $ newgrp libvirt
 ----
 
 [[xhyve-driver-install]]
-== xhyve driver
+=== xhyve driver
```
budhrg: Double equal to sign only to make it visible as part of table of content. Click on `View` button of this page.

docs/source/getting-started/setting-up-hypervisor-driver-plugin.adoc
```diff
@@ -75,14 +74,14 @@ $ newgrp libvirt
...
 
 Minishift is currently tested against `docker-machine-driver-xhyve` version 0.3.1.
 
 [[homebrew-install]]
-=== Homebrew install
+==== Homebrew install
 
-You can verify the installed version of the xhyve driver on your system before you install.
+You can verify the existing version of the xhyve driver on your system before you install the latest version using:
```
budhrg: `You can verify the existing version of the xhyve driver on your system before you install the latest version using:` to
```
You can verify the existing version of the xhyve driver on your system using:
```
as above line only require if we have mandatory rule to keep the xhyve driver plugin latest. It's user choice or sometime as part of issue where user need to update the xhyve driver.

docs/source/_topic_map.yml
hferentschik: I don't get it? Did we not say we stick with "Docker machine driver"?

gbraad: @LalatenduMohanty +1 on my comment, @budhrg +1 on this change, and I was in favor from the beginning... (3-1 in favor of the change ?)

Preeticp: +1 to the name change from my side too, considering the addition of Hyper-V information.

docs/source/_topic_map.yml
hferentschik: > The reasoning behind the rename was: Information wrt Hyper-V virtual switch was expanding. Installation doc did not seem to be the right place for this information. Given that we are already redirecting the users to the Docker Drivers doc section for other drivers, it made sense to add Hyper-V information in there too and provide a link to it from the Installation. 

Sure, but that in itself does not give a reason for the rename. Hyper-V. We are also setting up/configuring the driver.

> However afaiu, please correct me if I am wrong, Hyper-V does not require a Docker machine driver,

It does. All hypervisors do. The driver is the API between docker/lib-machine and the hypervisor.

> only KVM and xhyve need them, so I could not club Hyper-V under the same heading as Docker Machine Drivers. 

You could have

> Based on above discussion it makes sense to give the doc a generic heading like "Setting up the Hypervisor driver plugin" 

"Hypervisor driver plugin" seems wrong to me. If rename, let drop the "docker-machine" part and align with Minikube. Personally I feel we are being less precise though.

> And maybe to be explicit, split the doc into one subheading talking about Hyper-V and another clubbing the KVM and xhyve info under the Docker Machine driver subheading. WDYT?

-1, that would not be needed.

docs/source/getting-started/setting-up-hypervisor-driver-plugin.adoc
```diff
@@ -152,5 +151,36 @@ $ sudo chown root:wheel /usr/local/bin/docker-machine-driver-xhyve
 $ sudo chmod u+s /usr/local/bin/docker-machine-driver-xhyve
 ----
 
-For more information, refer to the https://github.com/zchee/docker-machine-driver-xhyve#install[xhyve driver] documentation on GitHub.
+For more information, see the https://github.com/zchee/docker-machine-driver-xhyve#install[xhyve driver] documentation on GitHub.
 ====
+
+== Hyper-V
```
LalatenduMohanty: @gbraad @Preeticp IMO this does not fit here. May be we need to find a new home for this.

gbraad: You mean the setup for the Hyper-V driver? This is also aligned with minikube: https://github.com/kubernetes/minikube/blob/master/docs/drivers.md#hyperv-driver

Preeticp: This whole PR was about finding a better place for Hyper-V information and this discussion has been going on for a week here and in #843

docs/source/getting-started/setting-up-hypervisor-driver-plugin.adoc
```diff
@@ -1,35 +1,34 @@
...
-
+. Install the KVM binary as follows:
++
 ----
 $ sudo curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.7.0/docker-machine-driver-kvm -o /usr/local/bin/docker-machine-driver-kvm
 $ sudo chmod +x /usr/local/bin/docker-machine-driver-kvm
 ----
 
-For more information, see the GitHub documentation of the https://github.com/dhiltgen/docker-machine-kvm#quick-start-instructions[docker machine KVM driver].
+. Execute the KVM binary, for more information, see the GitHub documentation of the https://github.com/dhiltgen/docker-machine-kvm#quick-start-instructions[docker machine KVM driver].
```
budhrg: @Preeticp I think, putting `Execute the KVM binary` here is wrong here. 

I was thinking of this: 
. Install the KVM binary as follows:
```
$ sudo curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.7.0/docker-machine-driver-kvm -o /usr/local/bin/docker-machine-driver-kvm
```
. Make the binary executable as follows:
```
$ sudo chmod +x /usr/local/bin/docker-machine-driver-kvm
```
For more information, see the GitHub documentation of the [docker machine KVM driver](https://github.com/dhiltgen/docker-machine-kvm#quick-start-instructions).

Preeticp: ok

docs/source/getting-started/setting-up-hypervisor-driver-plugin.adoc
LalatenduMohanty: @Preeticp @gbraad . Sorry about it. As discused I was still not convinced  about it. I am still not very convinced but do not have a better suggestion either. So lets go with this.

hferentschik:
Merged, thanks.

PR #953, opened 2017-05-24
Issue #687 Extend configurable properties for config command

### LINKED ISSUE #687: Allow to add any start parameter as part of config.json
it will be nice to have the `routing-prefix` and `public-host-name` as part of the config properties so that user can set it once like other stuff `vm-driver` for their minishift. Right now its need that we add it every time during the minishift start

praveenkumar:
Fix #687


cmd/minishift/cmd/config/config.go
```diff
@@ -116,6 +116,59 @@ var settings []Setting = []Setting{
...
+		validations: []setFn{IsValidPath},
+	},
+	{
+		name:        "host-volumes-dir",
+		set:         SetString,
+		validations: []setFn{IsValidPath},
+	},
+	{
+		name: "http-proxy",
+		set:  SetString,
```
praveenkumar: I need to add proxy validation function also once #942 get merge.

cmd/minishift/cmd/config/config.go
```diff
@@ -116,6 +116,59 @@ var settings []Setting = []Setting{
...
+	},
+	{
+		name: "https-proxy",
+		set:  SetString,
```
praveenkumar: Ditto

cmd/minishift/cmd/config/config.go
hferentschik: +1

praveenkumar:
> No tests? 

Yes, I am extending unit test case for it.

> So here is a thought? What's about we change start.go a bit and put the flag names which we have currently in the const block as well into an array. Then we could iterate this array in a test and verify that the option is setable. WDYT?

It's not only the flag names we require but also type so we set a flag against that particular type (like int, string, bool ..etc) so we can map those details then yes we can iterate it.

praveenkumar:
> Not approving as still in WIP state

Removed WIP state.

LalatenduMohanty:
Looks good to me. I am yet to test it.

cmd/minishift/cmd/config/config.go
```diff
@@ -116,6 +116,61 @@ var settings []Setting = []Setting{
 		name: config.ReminderWaitPeriodInHours,
```
hferentschik: What's about we get rid of  _WantUpdateNotification_ and _ReminderWaitPeriodInHours_. AFAIC they are obsolete. WDYT?

praveenkumar: @hferentschik I do agree, I will remove it. 👍

cmd/minishift/cmd/config/config.go
```diff
@@ -109,11 +108,62 @@ var settings []Setting = []Setting{
...
+		set:         SetString,
+		validations: []setFn{IsValidProxy},
+	},
+	{
+		name:        "https-proxy",
+		set:         SetString,
+		validations: []setFn{IsValidProxy},
+	},
+	{
+		name: "insecure-registry",
```
praveenkumar: @hferentschik @LalatenduMohanty @budhrg @gbraad This is what I added and set a Slice method.

Now if I do like below then it does put value of insecure registry as slice but it neglected by minishift start.

```
$ ./minishift config set insecure-registry 17.0.0.1/32,63.0.0.1/23

17:40 $ ./minishift config view
- insecure-registry    : [17.0.0.1/32 63.0.0.1/23]
- show-libmachine-logs : true
```

gbraad: so in short you mean there is a serialization issue?

praveenkumar: @gbraad right that I think might be the issue but how do we solve it.

cmd/minishift/cmd/config/util.go
```diff
@@ -70,3 +71,9 @@ func SetBool(m MinishiftConfig, name string, val string) error {
 	m[name] = b
 	return nil
 }
+
+func SetSlice(m MinishiftConfig, name string, val string) error {
+	b := strings.Split(val, ",")
```
gbraad: Is this also called on rereading the file?
If that is the case, just also split on space... ? just like splitting on strings... or am I missing something (haven't been able to run the code yet)

cmd/minishift/cmd/config/config.go
```diff
@@ -44,6 +43,22 @@ type Setting struct {
...
+
+var (
+	SetUrl = Flags{
+		Name:        "iso-url",
+		set:         SetString,
+		validations: []setFn{IsValidUrl},
+		callbacks:   []setFn{RequiresRestartMsg},
+	}
+)
+
```
praveenkumar: @hferentschik FYI, this is how each flag will look when we have it in one place. Hope this is what we discussed. Currently I didn't put those options to a different file. Let me know if you have something else in mind.

cmd/minishift/cmd/start.go
```diff
@@ -144,7 +144,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	proxyConfig := handleProxies()
 
 	machineConfig := cluster.MachineConfig{
-		MinikubeISO:      viper.GetString(isoURL),
+		MinikubeISO:      viper.GetString(config.SetUrl.Name),
```
praveenkumar: @hferentschik This is changes which start.go need.

hferentschik:
@praveenkumar seems the CI builds are failing due to fmt check errors.

budhrg:
@praveenkumar , `Extent` or `Extend` in commit message.

cmd/minishift/cmd/start.go
gbraad: `Name` ? doesn't sound the same as the`url` (location).

cmd/minishift/cmd/start.go
```diff
@@ -365,7 +365,7 @@ func init() {
 
 // initStartFlags creates the CLI flags which needs to be passed on to 'libmachine'
 func initStartFlags() {
-	startFlagSet.String(isoURL, constants.DefaultIsoUrl, "Location of the minishift ISO.")
+	startFlagSet.String(config.SetUrl.Name, constants.DefaultIsoUrl, "Location of the minishift ISO.")
```
gbraad: Is this why we do:
```diff
+type Flags struct {
+	Name        string
```

if so, this is so meaningless (provides no clue what the relation is, and this also due to the usage `Name` when it is about an URL)

cmd/minishift/cmd/start.go
praveenkumar: @gbraad the way struct created. this is a sample not implemented completely, wanted to check with @hferentschik if this what he have in mind.

cmd/minishift/cmd/config/config.go
budhrg: @praveenkumar , did you find the solution here?

-

cmd/minishift/cmd/config/config.go
```diff
@@ -38,84 +37,66 @@ type setFn func(string, string) error
...
+	// Global flags
+	LogDir             = newFlags("log_dir", SetString, []setFn{IsValidPath}, nil, true)
+	ShowLibmachineLogs = newFlags("show-libmachine-logs", SetBool, nil, nil, true)
+
+	// Host Folders
+	HostFoldersMountPath = newFlags("hostfolders-mountpath", SetString, nil, nil, true)
+	HostFoldersAutoMount = newFlags("hostfolders-automount", SetBool, nil, nil, true)
+)
+
+func newFlags(name string, set func(MinishiftConfig, string, string) error, validations []setFn, callbacks []setFn, isApply bool) *Setting {
```
hferentschik: newFlag or createFlag (singular) or maybe even better newSetting

cmd/minishift/cmd/config/config.go
```diff
@@ -38,84 +37,66 @@ type setFn func(string, string) error
...
+	flag := Setting{
+		Name:        name,
+		set:         set,
+		validations: validations,
+		callbacks:   callbacks,
+	}
+	if isApply {
```
hferentschik: +1 right, a flag is enough. No need to pass the array itself. It would maybe be a bit more explicit, but ok

cmd/minishift/cmd/config/config.go
```diff
@@ -38,84 +37,66 @@ type setFn func(string, string) error
...
-	},
-	{
-		name: config.WantUpdateNotification,
-		set:  SetBool,
-	},
-	{
-		name: config.ReminderWaitPeriodInHours,
-		set:  SetInt,
-	},
+var flagsList []Setting
```
hferentschik: If we stick with _Setting_ we might want to call this settingsList

cmd/minishift/cmd/config/util.go
```diff
@@ -70,3 +71,9 @@ func SetBool(m MinishiftConfig, name string, val string) error {
 	m[name] = b
 	return nil
 }
+
+func SetSlice(m MinishiftConfig, name string, val string) error {
```
hferentschik: I guess this is obsolete for this pull request, right? You skipped the slice based parameters for now, no?

cmd/minishift/cmd/start.go
```diff
@@ -56,35 +57,6 @@ import (
 
 const (
 	commandName = "start"
-
-	// minishift
-	isoURL                = "iso-url"
```
hferentschik: nice, I like that we get rid of this here. Less lines in start.go. Together with #962 start.go start.go looks manageable again.

cmd/minishift/cmd/start.go
```diff
@@ -33,6 +33,7 @@ import (
 	"github.com/docker/machine/libmachine/provision"
 	"github.com/golang/glog"
 	"github.com/minishift/minishift/cmd/minishift/cmd/addon"
+	"github.com/minishift/minishift/cmd/minishift/cmd/config"
```
hferentschik: What's about we alias 'config' to make it more explicit. Something like 'startFlag' or 'flags' or 'settings'?

cmd/minishift/cmd/start.go
```diff
@@ -144,16 +116,16 @@ func runStart(cmd *cobra.Command, args []string) {
 	proxyConfig := handleProxies()
 
 	machineConfig := cluster.MachineConfig{
-		MinikubeISO:      viper.GetString(isoURL),
-		Memory:           viper.GetInt(memory),
-		CPUs:             viper.GetInt(cpus),
-		DiskSize:         calculateDiskSizeInMB(viper.GetString(humanReadableDiskSize)),
-		VMDriver:         viper.GetString(vmDriver),
+		MinikubeISO:      viper.GetString(config.ISOUrl.Name),
```
hferentschik: If we alias the import we might have something more understandable:

    viper.GetString(settings.ISOUrl.Name)

or 

    viper.GetString(startFlags.ISOUrl.Name)

hferentschik: bugger more rebase fun for me :-(

pkg/minikube/config/config.go
```diff
@@ -1,22 +0,0 @@
...
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
-package config
-
-const (
-	WantUpdateNotification    = "WantUpdateNotification"
```
hferentschik: +1, good to get rid of this

praveenkumar:
@hferentschik All suggestions are now incorporated.

hferentschik:
thanks

hferentschik:
Merged, thanks.

@praveenkumar could  you please create a follow up issue for the slice based flags? As part of the 'config set' command we use the comma separation part you already had. Needs to be documented for the user of course. Then we need to figure out why during start the slice values are not properly retrieved. For that we need to look at Viper itself, but also the code which iterated the set flags in start.go. It might be that we need a type based switch there.

We might also consider to just not use slices in the start command. Instead you comma separate multiple docker env settings. At least this would then be consistent with the 'config set' command. Unfortunately, it would break backwards compatibility.

Issue #637, opened 2017-04-03
Add explicit command to apply add-on


hferentschik:
At the moment addons are only applied during `minishift start`. There is no explicit way to run a given addon. 

There are two suggestions:

- An explicit `minishift addons run [ADDON-NAME]`
- Implicit use of `minishift addons enable  [ADDON-NAME]`

The reuse of the _enable_ command tries to avoid command proliferation. The idea is that if the VM is not running, the addon just gets enabled in the config (current behaviour). If the VM is running, however, it also gets run directly.

jorgemoralespou:
@hferentschik as you know I'm in favour of option 2 (whenever you enable the addon, it will install it if there is a VM running). But I just realized a race condition you should be aware, if you have a VM already created but stopped, and then enable an addon, it will not be applied, as install only happens at bootstrapping (or at command execution if this get implemented and the VM is running). Maybe there needs to be a mark, whether the addon was installed, so if it was enabled with the VM in stop mode, it will be applied on first next boot.

hferentschik:
I think an explicit _run_ is the better option. For example I could image a addon which I want to run multiple times. This is not really "enabling", but rather "execute"/"run". I think semantically a explicit "run" is needed. 

I am also not convinced yet, whether this implicit enable or enable and run/apply is really so intuitive.

LalatenduMohanty:
> think an explicit run is the better option. For example I could image a addon which I want to run multiple times. This is not really "enabling", but rather "execute"/"run". I think semantically a explicit "run" is needed.

+1

minishift addons enable is fine when the VM is not running and it automatically runs when minishift is started. However I think we need an explicit run command. It is similar to how `systemctl enable` and `systemctl start` work.

hferentschik:
Merged via pull request #958

Issue #665, opened 2017-04-05
User should enable/disable multiple addons at once


budhrg:
This is again a better user experience IMO for addon where user should enable/disable multiple addons through single commands.
ATM, it is just one addon per command.

```
$ minishift addons list                          
- anyuid         : disabled   P(0)
- cluster-admin  : disabled   P(0)
- registry       : disabled   P(0)

$ minishift addons enable anyuid cluster-admin registry

$ minishift addons list                                
- anyuid         : enabled    P(0)
- cluster-admin  : disabled   P(0)
- registry       : disabled   P(0)
```

We can also implement same for `install` subcommand though we are also looking to install multiple addons based on folder, see here https://github.com/minishift/minishift/issues/635#issuecomment-291547294.

praveenkumar:
@budhrg I agree that it's a better user experience but I don't think it should have priority as `major` because we are not breaking any usecase here.

budhrg:
Updated @praveenkumar 👍

hferentschik:
And how do you handle priorities?

hferentschik:
BTW, where is this issue coming from and why is it added to rc.2?

LalatenduMohanty:
The tricky part if how we handle the priorities when we enable multiple add-ons at one go. I think we have to assume that when we enable multiple add-ons , the add-ons have same priority. WDYT?

hferentschik:
>  I think we have to assume that when user enable multiple add-ons , the add-ons have same priority.

I guess.

gbraad:
for the moment yes... priorities can also be solved by dependency resolving (but this involves pre-processing)

praveenkumar:
I think we should drop this feature right now or if we implement then we need each field mandatory (addon name, priority etc.) Problem is if we enable this features then user might use it for different priority addons and then again hit the issue.

budhrg:
> The tricky part if how we handle the priorities when we enable multiple add-ons at one go. I think we have to assume that when user enable multiple add-ons , the add-ons have same priority. WDYT?

Yes. Agree. Also, depend on how much far we go with this feature too. As @gbraad  said, dependency resolving is key here.

hferentschik:
> I think we should drop this feature right now

Works for me

> or if we implement then we need each field mandatory (addon name, priority etc.) 

[...]

Issue #170, opened 2016-11-29
Minishift next


hferentschik:
The next community release of Minishift is coming up and the question is which version to use for it. It will be a major change, since provisioning will occur via 'oc cluster up' opposed to running the openshift binary on the VM. The current version is 0.9.0.

Do we go for 0.10.0? Or 1.0.0? IMO a major version change would make sense, especially since I'd like to start with semantic versioning asap. On the other hand, moving directly to 1.0.0 after making so substantial changes to the code might feel odd.

Thoughts?

For now I created a 1.0.0 milestone. Once we agreed on the Minishift next version, we can rename this to whatever we want to use.

jimmidyson:
How about a 1.0.0-beta.1? We should probably review commands, etc to make sure we agree on what we're going to remain strict on for backwards compatibility but that's fine.

hferentschik:
> How about a 1.0.0-beta.1

+1 Works for me

LalatenduMohanty:
I do not have strong opinion on the version we choose now. However another major version change in future will be to replace boot2docker ISO with the CentOS based ISO.

jimmidyson:
@LalatenduMohanty Do we think we could do that in an alpha/beta/rc cycle, with the change being mandatory for a 1.0.0 release?

praveenkumar:
I am favor to follow alpha/beta/rc cycle as features comes along.

LalatenduMohanty:
@jimmidyson  sure, we can try that.

hferentschik:
> I do not have strong opinion on the version we choose now. However another major version change in future will be to replace boot2docker ISO with the CentOS based ISO.

Why would that be a major version change? I would think this is just going from 1.0.0 to say 1.1.0. This change is really transparent to the user and just adds a new feature of optionally changing the ISO. 

We can discuss the version once we cross this bridge. I think this particular question is resolved for now.

PR #958, opened 2017-05-26
Issue #637 Adding a new addon sub-command "apply"

LalatenduMohanty:

### DIFF
--- a/cmd/minishift/cmd/addon/addon_apply.go
+++ b/cmd/minishift/cmd/addon/addon_apply.go
@@ -1,0 +1,92 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package addon
+
+import (
+	"fmt"
+
+	"github.com/docker/machine/libmachine"
+	"github.com/docker/machine/libmachine/provision"
+	"github.com/minishift/minishift/cmd/minishift/cmd/util"
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+const (
+	routingSuffix = "routing-suffix"
+)
+
+var addonsApplyCmd = &cobra.Command{
+	Use:   "apply ADDON_NAME ...",
+	Short: "Executes the specified add-ons.",
+	Long:  "Executes the specified add-ons. The command works with both enabled and disabled add-ons.",
+	Run:   runApplyAddon,
+}
+
+func init() {
+	AddonsCmd.AddCommand(addonsApplyCmd)
+}
+
+func runApplyAddon(cmd *cobra.Command, args []string) {
+
+	if len(args) == 0 {
+		atexit.ExitWithMessage(1, emptyAddOnError)
+	}
+
+	addOnManager := GetAddOnManager()
+	for i := range args {
+		addonName := args[i]
+		if !addOnManager.IsInstalled(addonName) {
+			atexit.ExitWithMessage(0, fmt.Sprintf(noAddOnMessage, addonName))
+		}
+	}
+
+	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+	defer api.Close()
+
+	util.ExitIfUndefined(api, constants.MachineName)
+
+	host, err := api.Load(constants.MachineName)
+	if err != nil {
+		atexit.ExitWithMessage(1, err.Error())
+	}
+
+	util.ExitIfNotRunning(host.Driver, constants.MachineName)
+
+	ip, err := host.Driver.GetIP()
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting IP: %s", err.Error()))
+	}
+
+	ocPath := minishiftConfig.InstanceConfig.OcPath
+	routingSuffix := viper.GetString(routingSuffix)
+	kubeConfigPath := constants.KubeConfigPath
+	sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
+
+	for i := range args {
+		addonName := args[i]
+		addon := addOnManager.Get(addonName)
+		addonContext := GetExecutionContext(ip, routingSuffix, ocPath, kubeConfigPath, sshCommander)
+		err = addOnManager.ApplyAddOn(addon, addonContext)
+		if err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprint("Error executing addon commands: ", err))
+		}
+	}
+}
--- a/cmd/minishift/cmd/util/util.go
+++ b/cmd/minishift/cmd/util/util.go
@@ -18,6 +18,7 @@ package util
 
 import (
 	"fmt"
+
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/state"
@@ -35,7 +36,7 @@ func VMExists(client *libmachine.Client, machineName string) bool {
 func ExitIfUndefined(client *libmachine.Client, machineName string) {
 	exists := VMExists(client, machineName)
 	if !exists {
-		atexit.ExitWithMessage(0, fmt.Sprintf("There is currently no '%s' VM defined.", machineName))
+		atexit.ExitWithMessage(0, fmt.Sprintf("The execution of this command requires an existing '%s' VM. However, there is currently none defined.", machineName))
 	}
 }
 
@@ -43,9 +44,9 @@ func IsHostRunning(driver drivers.Driver) bool {
 	return drivers.MachineInState(driver, state.Running)()
 }
 
-func ExitIfNotRunning(driver drivers.Driver) {
+func ExitIfNotRunning(driver drivers.Driver, machineName string) {
 	running := IsHostRunning(driver)
 	if !running {
-		atexit.ExitWithMessage(0, "Minishift VM is not running.")
+		atexit.ExitWithMessage(0, fmt.Sprintf("The execution of this command requires a running '%s' VM, but there is currently none.", machineName))
 	}
 }
--- a/docs/source/using/addons.adoc
+++ b/docs/source/using/addons.adoc
@@ -144,6 +144,7 @@ By default, this is not allowed in OpensShift.
 
 Add-ons are enabled with the link:../command-ref/minishift_addons_enable{outfilesuffix}[`minishift addons enable`]
 command and disabled with the link:../command-ref/minishift_addons_disable{outfilesuffix}[`minishift addons disable`] command.
+Enabled add-ons automatically get executed during link:../command-ref/minishift_start{outfilesuffix}[`minishift start`].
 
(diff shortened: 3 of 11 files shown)


hferentschik:
When running w/o an existing VM, I get:

    $ minishift addons apply anyuid
    There is currently no 'minishift' VM defined.

I think there is a bit of context missing. 

What's about something like:

    The command you are trying to run, requires an existing Minishift VM, but there is currently no 'minishift' VM defined.

Or we we want to get even more fancy:

    The <command-name> command requires an existing Minishift VM, but there is currently no 'minishift' VM defined.

In this case we need to add the command name to the _ExitIfNotRunning_ command. Looking at the other spots where this method is used, the change seems to be generally useful. WDYT?

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -0,0 +1,88 @@
...
+	AddonsCmd.AddCommand(addonsApplyCmd)
+}
+
+func runApplyAddon(cmd *cobra.Command, args []string) {
+
+	if len(args) == 0 {
+		atexit.ExitWithMessage(1, emptyAddOnError)
+	}
+
+	addonName := args[0]
```
hferentschik: If we do this in a loop, treating each argument as a add-on name, we can apply multiple add-ons in one call. 

The same should then be applied to 'enable', 'disable' and 'install'. This will improve the general usability of the add-on commands.

hferentschik: See also https://github.com/minishift/minishift/issues/665

LalatenduMohanty: @hferentschik agree. I was wondering something like `--all` but handling multiple addons is good for now.

LalatenduMohanty:
> In this case we need to add the command name to the ExitIfNotRunning command. Looking at the other spots where this method is used, the change seems to be generally useful. WDYT?

@hferentschik +1 to adding more context to the error.  Looking at the output, I think we do not need to add the command as it is understood. e.g. 
```
$ minishift addons apply anyuid
Requires a running Minishift VM, but there is currently no 'minishift' VM running.

```
Or
```
$ minishift addons apply anyuid
Requires a Minishift VM, but there is currently no 'minishift' VM defined
```
WDYT?

hferentschik:
> $ minishift addons apply anyuid
> Requires a Minishift VM, but there is currently no 'minishift' VM defined

I think there is something missing. If you are not passing the command name, it should be something like this imo:

    The execution of this command requires an existing Minishift VM. However, there is currently none defined.

That would work for me.

cmd/minishift/cmd/util/util.go
```diff
@@ -35,17 +36,17 @@ func VMExists(client *libmachine.Client, machineName string) bool {
...
 func IsHostRunning(driver drivers.Driver) bool {
 	return drivers.MachineInState(driver, state.Running)()
 }
 
-func ExitIfNotRunning(driver drivers.Driver) {
+func ExitIfNotRunning(driver drivers.Driver, machineName string) {
 	running := IsHostRunning(driver)
 	if !running {
-		atexit.ExitWithMessage(0, "Minishift VM is not running.")
+		atexit.ExitWithMessage(0, fmt.Sprintf("The execution of this command requires a running '%s' VM, but there is currently none running.", machineName))
```
hferentschik: The execution of this command requires a running '%s' VM, but there is currently none.

no need to repeat running

docs/source/using/addons.adoc
```diff
@@ -161,6 +161,21 @@ $ minishift addons enable anyuid
 $ minishift addons disable anyuid
 ----
 
+[[apply-addons]]
+== Applying add-ons
+Enabled add-ons automatically get executed during link:../command-ref/minishift_start{outfilesuffix}[`minishift start`].
+
+Also add-ons can be explictly executed with link:../command-ref/minishift_addons_apply{outfilesuffix}[`minishift addons apply`] command.
+
+The following examples show how to execute the _anyuid_ and _admin-user_ add-ons.
```
hferentschik: The following example shows how to explicitly apply the _anyuid_ and _admin-user_ add-ons.

docs/source/using/addons.adoc
```diff
@@ -161,6 +161,21 @@ $ minishift addons enable anyuid
 $ minishift addons disable anyuid
 ----
 
+[[apply-addons]]
+== Applying add-ons
+Enabled add-ons automatically get executed during link:../command-ref/minishift_start{outfilesuffix}[`minishift start`].
+
+Also add-ons can be explictly executed with link:../command-ref/minishift_addons_apply{outfilesuffix}[`minishift addons apply`] command.
```
hferentschik: However, you can also explicitly apply an add-on with the link:../command-ref/minishift_addons_apply{outfilesuffix}[`minishift addons apply`] command.

hferentschik: I suggest not to start a new paragraph. Right now we have three single sentences, each forming a paragraph. Looks odd.

cmd/minishift/cmd/util/util.go
LalatenduMohanty: Agree

docs/source/using/addons.adoc
LalatenduMohanty: Agree

docs/source/using/addons.adoc
LalatenduMohanty: agree

hferentschik:
So what's the intended behaviour for installed, but disabled add-ons? Can I apply them or not?

Right now the docs says:

    Enabled add-ons automatically get executed during minishift start.

The thing is, atm I can apply also disabled add-ons. So at the very least the docs and impl are not aligned. Let's settle on an intended behaviour.

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -0,0 +1,92 @@
...
+)
+
+const (
+	routingSuffix = "routing-suffix"
+)
+
+var addonsApplyCmd = &cobra.Command{
+	Use:   "apply ADDON_NAME...",
+	Short: "Executes the specified add-ons.",
+	Long:  "Executes the specified add-ons",
```
hferentschik: Missing full stop

hferentschik: You probably want to add that enabled but also disabled add-ons can be applied

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+const (
+	routingSuffix = "routing-suffix"
+)
+
+var addonsApplyCmd = &cobra.Command{
+	Use:   "apply ADDON_NAME...",
```
hferentschik: Is this the right "syntax" to imply that there can be multiple add-on names?

LalatenduMohanty: @hferentschik Honestly I could not find any good golang example to refer. In Linux/GNU distributions we use similar (e.g. $ cat --help or $ mkdir --help) and it also similar to variadic functions.

hferentschik: Good enough for me. But maybe put a space between ADDON_NAME and ...?

docs/source/using/addons.adoc
```diff
@@ -161,6 +161,21 @@ $ minishift addons enable anyuid
 $ minishift addons disable anyuid
 ----
 
+[[apply-addons]]
+== Applying add-ons
```
hferentschik: There should be an empty lines after a section heading

docs/source/using/addons.adoc
```diff
@@ -161,6 +161,21 @@ $ minishift addons enable anyuid
...
+Enabled add-ons automatically get executed during link:../command-ref/minishift_start{outfilesuffix}[`minishift start`].
+However, you can also explicitly apply an add-on with the link:../command-ref/minishift_addons_apply{outfilesuffix}[`minishift addons apply`] command.
+It works with both enabled and disabled add-ons.
```
hferentschik: Opposed to the implicit add-on executions during `minishift start` which only applies enabled add-ons, `minishift addon apply` allows to apply enabled as well as disabled add-ons. You can also apply multiple add-ons with a single command by specifying multiple add-on names separate by space.

docs/source/using/addons.adoc
```diff
@@ -161,6 +161,23 @@ $ minishift addons enable anyuid
...
 ----
 
+[[apply-addons]]
+== Applying add-ons
+
+Add-ons can be explicitly executed with the link:../command-ref/minishift_addons_apply{outfilesuffix}[`minishift addons apply`] command.
+You can use _add-on_ _apply_ command for both enabled and disabled add-ons. For applying multiple add-ons at once please specify multiple add-on names separated by space.
+
+_Tip_: To execute an add-on during link:../command-ref/minishift_start{outfilesuffix}[`minishift start`] you need to enable it using link:../command-ref/minishift_addons_enable{outfilesuffix}[`minishift addons enable`] command.
+
```
Preeticp: @LalatenduMohanty I think this information should go as the first line of enabling and disabling add-ons. It is a necessary action for the Add-on functionality.

I suggest starting the Enabling and disabling add-ons section with the sentence on line #170 and adding this as the second sentence: Enabled add-ons automatically get executed during link:../command-ref/minishift_start{outfilesuffix}[`minishift start`].
This would set the context for the Apply part.

docs/source/using/addons.adoc
```diff
@@ -161,6 +161,23 @@ $ minishift addons enable anyuid
 $ minishift addons disable anyuid
 ----
 
+[[apply-addons]]
+== Applying add-ons
+
+Add-ons can be explicitly executed with the link:../command-ref/minishift_addons_apply{outfilesuffix}[`minishift addons apply`] command.
```
Preeticp: Add-ons can also be explicitly executed

docs/source/using/addons.adoc
```diff
@@ -161,6 +161,23 @@ $ minishift addons enable anyuid
...
+You can use _add-on_ _apply_ command for both enabled and disabled add-ons. For applying multiple add-ons at once please specify multiple add-on names separated by space.
+
```
Preeticp: For applying multiple add-ons at once>For applying multiple add-ons witha single command,

docs/source/using/addons.adoc
LalatenduMohanty: @Preeticp I disagree  with the use of `also` as this is first sentence of the section.

docs/source/using/addons.adoc
```diff
@@ -161,6 +161,23 @@ $ minishift addons enable anyuid
 $ minishift addons disable anyuid
 ----
 
+[[apply-addons]]
+== Applying add-ons
+
+Add-ons can be explicitly executed with the link:../command-ref/minishift_addons_apply{outfilesuffix}[`minishift addons apply`] command.
+You can use _add-on_ _apply_ command for both enabled and disabled add-ons. For applying multiple add-ons at once please specify multiple add-on names separated by space.
```
hferentschik: I think we can drop "please". I know we want to be polite, but for the docs this is not needed imo. Also, I think we can drop "multiple" from "specify multiple add-on names"

docs/source/using/addons.adoc
Preeticp: Agreed, but I look at this as a continuation from the implicit use of Add-ons and lets the user know that apart from the implicit way you also have this option. Anyways it is subjective and we can live without using also.

docs/source/using/addons.adoc
```diff
@@ -194,6 +195,21 @@ the _eap_ add-on.
 NOTE: If two add-ons have the same priority the order in which they are getting applied is not
 determined.
 
+[[apply-addons]]
+== Applying add-ons
+
+Add-ons can be explicitly executed with the link:../command-ref/minishift_addons_apply{outfilesuffix}[`minishift addons apply`] command.
+You can use _add-on_ _apply_ command for both enabled and disabled add-ons. For applying multiple add-ons with a single command specify add-on names separated by space.
```
hferentschik: I don't get the formatting here - _add-on_ _apply_. Actual commands we put into backticks. So do you mean `addons apply`? Or do you want to say "You can use add-on `apply` command", in which case you generically refer to the add-on mechanism. No need for emphasis here. Or maybe just "You can use the `apply` command"?

hferentschik:
Merged, thanks.

Issue #849, opened 2017-04-28
Add reference to minishift-centos-iso in Minishift documentation


LalatenduMohanty:
We have not mentioned about minishift=centos ISO in Minishift documentation. We need documentation which will communicate user that apart form the default minishift -b2d ISO user has choice to run minishift with CentOS ISO.  We should also mention that CentOS ISO is of higher size as compared to minishift-b2d but it is an enterprise GNU/Linux distribution where as boot2docker is not. 

[1] https://github.com/minishift/minishift-centos-iso/

hferentschik:
Merged via pull request #973

Issue #428, opened 2017-02-18
Introduce short names for our maintained ISOs


hferentschik:
ATM, only the default Boot2Docker ISO gets automatically downloaded when nothing is cached. If someone wants to use the CentOS ISO he needs to specify the URL and he also has to be aware of any compatibility issues. 

It should be possible for the user to just specify that he wants to use the CentOS / RHEL image and it gets downloaded automatically. For this reason I suggest to introduce short names for our maintained images. Something along these lines

* `minishift start --iso-url=b2d` - that the same as our current default
* `minishift start --iso-url=centos` - will 'expand' to the latest compatible CentOS ISO
* `minishift start --iso-url=rhel` - get the latest RHEL image (if possible, would probably need to ask for credentials)

budhrg:
IMO, chosing `--iso-url` flag is not appropriate here. Now, the flexibility of chosing iso url is lost.

And, I also don't like to mix the values of `--iso-url` as :
```
- file://<iso-path>
- https://<iso-url>
-b2d / centos /rhel
```
Quite unclear and weird to me.

I would be happy if we have flag like `--iso-type` which IMO provide more clarity.
Now, lines are
```
minishift start --iso-type=b2d
minishift start --iso-type=centos
minishift start --iso-type=rhel

# And if user want to specify URL (most of the case it is always non-default one)
minishift start --iso-url=<url>
```

hferentschik:
The problem is with --iso-type and --iso-url, that you end up with conflicting settings. I have no problem with short names, after all they stand in for actual URLs. Interested to know how others think about it.

budhrg:
Resolved via https://github.com/minishift/minishift/pull/1033

PR #961, opened 2017-05-29
[Do not merge] Issue #849 Add documentation for minishift-centos-iso

thatdocslady:
Fixes issue #849 

This PR covers documentation for the Boot2Docker/CentOS images.

### DIFF
--- a/docs/source/using/managing-minishift.adoc
+++ b/docs/source/using/managing-minishift.adoc
@@ -53,6 +53,42 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
 The link:../command-ref/minishift_delete{outfilesuffix}[`minishift delete`] command deletes the OpenShift cluster,
 and also shuts down and deletes the Minishift VM. No data or state are preserved.
 
+[[choosing-iso-image]]
+== Choosing the ISO image
+
+When you start Minishift, it downloads a live ISO image that the hypervisor uses to
+provision the Minishift VM.
+
+The following ISO images are available:
+
+- link:https://github.com/minishift/minishift-b2d-iso/releases[Minishift Boot2Docker] (Default). This ISO image is
+based on link:https://github.com/boot2docker/boot2docker[Boot2Docker], which is a
+lightweight Linux distribution customized to run Docker containers. This ISO image size is small and
+optimized for development but not for production.
+
+- link:https://github.com/minishift/minishift-centos-iso/releases[Minishift CentOS]. This ISO image is based on
+link:https://www.centos.org/[CentOS], which is an enterprise-ready Linux distribution that more
+closely resembles a production environment. This ISO image size is larger than the Boot2Docker ISO image.
+
+By default, Minishift uses the **Minishift Boot2Docker** ISO image. To choose the Minishift CentOS ISO image instead,
+you can do one of the following actions:
+
+- Set the `--iso-url` flag when you run `minishift start` and enter the download URL of the
+Minishift CentOS ISO image. For example:
++
+----
+$ minishift start --iso-url <Web_location_of_ISO_image>
+----
+- Manually download the Minishift CentSO ISO image from the link:https://github.com/minishift/minishift-centos-iso/releases[releases page]
+and enter the path to the ISO image location in the following format:
++
+----
+$ minishift start --iso-url file:///<path_to_ISO_image>
+----
+
+NOTE: You cannot run Minishift with both ISO images concurrently. To switch between
+ISO images, delete the Minishift instance and start a new instance with the ISO image that you want to use.
+
 [[runtime-options]]
 == Runtime options
 


docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
 The link:../command-ref/minishift_delete{outfilesuffix}[`minishift delete`] command deletes the OpenShift cluster,
 and also shuts down and deletes the Minishift VM. No data or state are preserved.
 
+[[choosing-vm]]
+== Choosing the VM image
+
+When you start Minishift, it downloads an launches a VM image that runs the OpenShift cluster.
```
LalatenduMohanty: suggestion: s/it downloads an launches a VM image that runs the OpenShift cluster/it downloads an launches a VM image that provisions the OpenShift cluster in it/

hferentschik: "an" == "and"

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
 The link:../command-ref/minishift_delete{outfilesuffix}[`minishift delete`] command deletes the OpenShift cluster,
 and also shuts down and deletes the Minishift VM. No data or state are preserved.
 
+[[choosing-vm]]
+== Choosing the VM image
```
hferentschik: What's about making this a sub-section of the start command?

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+When you start Minishift, it downloads an launches a VM image that runs the OpenShift cluster.
+The VM image is a Live ISO that contains an operating system as well as all the dependencies to
+run the OpenShift cluster.
+
+The following VM images are available:
+
+- link:https://github.com/minishift/minishift-b2d-iso[Boot2Docker] (Default). This image is
```
hferentschik: "Boot2Docker ISO" and "CentOS ISO", especially in the later case, just having CentOS standalone seems confusing to me.

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+based on the link:https://github.com/boot2docker/boot2docker[Boot2Docker] image, which is a
+lightweight Linux distribution optimized for containerized application development.
+
+- link:https://github.com/minishift/minishift-centos-iso/[CentOS]. This image contains
+the link:https://www.centos.org/[CentOS] distribution, which is based on Red Hat Enterprise Linux.
```
hferentschik: "CentOS distribution, which is based on Red Hat Enterprise Linux" - it's the otehr way around, right?

thatdocslady: As far as I know it's Fedora>RHEL>CentOS (upstream>downstream>upstream).

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
 The link:../command-ref/minishift_delete{outfilesuffix}[`minishift delete`] command deletes the OpenShift cluster,
 and also shuts down and deletes the Minishift VM. No data or state are preserved.
 
+[[choosing-vm]]
```
LalatenduMohanty: It should be `choosing-iso`

docs/source/using/managing-minishift.adoc
thatdocslady: I considered that, but I figured if we want to really sign-post it, then it should have a 2heading and therefore show up in the TOC of the page. It's also something that you can persist IIRC so that means it's relevant to other sections in the page.

LalatenduMohanty: s/ Choosing the VM image/Choosing the ISO/

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+== Choosing the VM image
+
+When you start Minishift, it downloads an launches a VM image that runs the OpenShift cluster.
+The VM image is a Live ISO that contains an operating system as well as all the dependencies to
+run the OpenShift cluster.
```
LalatenduMohanty: I do not think we need this sentence as we should not explain why we need the VM in this section as we have already explained this in the getting started guide.

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+The following VM images are available:
```
LalatenduMohanty: s/VM images/ISOs/. Technically there is a difference between ISO and a VM image.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: I do no think we need to mention relationship between CentOS and RHEL here.

hferentschik: hmm, yes maybe. I guess I've read CentOS, but thought Fedora

docs/source/using/managing-minishift.adoc
LalatenduMohanty: @thatdocslady We should put the exact command in the documentation, so that user can copy paste and run the command.

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+- link:https://github.com/minishift/minishift-b2d-iso[Boot2Docker] (Default). This image is
+based on the link:https://github.com/boot2docker/boot2docker[Boot2Docker] image, which is a
+lightweight Linux distribution optimized for containerized application development.
+
+- link:https://github.com/minishift/minishift-centos-iso/[CentOS]. This image contains
+the link:https://www.centos.org/[CentOS] distribution, which is based on Red Hat Enterprise Linux.
++
+NOTE: The CentOS image size is larger than the Boot2Docker image, because the CentOS
```
praveenkumar: s/CentOS image size/CentOS ISO size/
s/Boot2Docker image/Boot2Docker ISO/

docs/source/using/managing-minishift.adoc
budhrg: +1

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+image is a full-featured enterprise Linux operating system.
```
praveenkumar: s/image is/ISO is/

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,32 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+To choose the CentOS image instead of the Boot2Docker image, you use the `--iso-url` flag
```
praveenkumar: s/CentOS image instead/CentOS ISO instead/

docs/source/using/managing-minishift.adoc
gbraad: I believe the way it used to be described was as the derived/based on the PNAELV offering. Not sure if mentioning RHEL is correct here.

LalatenduMohanty:
@thatdocslady There are two ways in which the CentOS ISO can be run with `--iso-url` command i.e.

1. directly pointing to the ISO in github release i.e. 
* ` $minishift start --iso-url https://github.com/minishift/minishift-centos-iso/releases/download/v1.0.0/minishift-centos7.iso` 
* Or download the ISO to the local host and then `--iso-url file:///<path to iso>`

docs/source/using/managing-minishift.adoc
thatdocslady: I don't agree because the ISO is just the packaging format for the host VM. If we put ourselves in the users' shoes, the ISO is the means to the VM which is the end.

docs/source/using/managing-minishift.adoc
thatdocslady: We can't assume that the users read all our docs from start to finish, so I try to give a little info here so that in case users land direclty on this page they don't have to skip around.

docs/source/using/managing-minishift.adoc
thatdocslady: What is the difference? IIRC the ISO is the packing format, like I wrote in a different comment.

docs/source/using/managing-minishift.adoc
thatdocslady: Following discussion, we decided to drop the URL example and write <location-of-iso> as a part of the command format syntax.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: @thatdocslady I disagree, I strongly feel that we do not need to explain this again here. When an user lands here he is just looking way to start Minishift with CentOS. If he wants to know more about Minishift then he should look in getting started section.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: The ISO is a disk format which contains the artifacts required for the VM. The hypervisor uses these and starts the VM.

LalatenduMohanty: Just to add, in general ISO can be used to boot physical machines and virtual machines.

docs/source/using/managing-minishift.adoc
hferentschik: > We should put the exact command in the documentation, so that user can copy paste and run the command.

+1 that would be my preferred option for now. We can make this partly dynamic by making the CentOS ISO a variable. We will need this most likely for other things as well later.

docs/source/using/managing-minishift.adoc
hferentschik: > I do no think we need to mention relationship between CentOS and RHEL here.

I think I prefer that as well

thatdocslady:
@LalatenduMohanty @budhrg @hferentschik I updated the PR based on your feedback.
Please note that the word "ISO" cannot exist by itself, the official term is "ISO image". Reference link: https://en.wikipedia.org/wiki/ISO_image

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,41 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+the CentOS ISO in the following format:
++
+----
+$ minishift start --iso-url <Web_location_of_ISO_image>
+----
+- Download the CentSO ISO image from the link:https://github.com/minishift/minishift-centos-iso/releases[releases page]
+and enter the path to the ISO image location in the following format:
++
+----
+$ minishift start --iso-url <file:///<path_to_ISO_image>
```
hferentschik: there is one '<' too much : <file:///<path_to_ISO_image>

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,41 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+full-featured enterprise Linux operating system.
+
+By default, Minishift uses the **Boot2Docker** image. To choose the CentOS ISO instead,
+you can do one of the following actions:
+
+- Set the `--iso-url` flag when you run `minishift start` and enter the download URL of
+the CentOS ISO in the following format:
++
+----
+$ minishift start --iso-url <Web_location_of_ISO_image>
```
hferentschik: Personally, I still would have preferred a single command line which can be but&paste by the user. Why did we not do that?

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,41 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
 The link:../command-ref/minishift_delete{outfilesuffix}[`minishift delete`] command deletes the OpenShift cluster,
 and also shuts down and deletes the Minishift VM. No data or state are preserved.
 
+[[choosing-iso-image-for-vm]]
+== Choosing the ISO image for the VM
```
hferentschik: "for the VM" seems superfluous to me

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,41 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+When you start Minishift, it downloads a live ISO image that the hypervisor uses to provision the host VM.
+This VM then locally provisions the OpenShift cluster.
```
hferentschik: "This VM then locally provisions the OpenShift cluster." - What does that mean. The VM itself does not provision anything. I recommend dropping the whole sentence

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,41 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+- link:https://github.com/minishift/minishift-b2d-iso/releases[Boot2Docker] (Default). This ISO is
+based on the link:https://github.com/boot2docker/boot2docker[Boot2Docker] ISO, which is a
+lightweight Linux distribution optimized for containerized application development.
+
+- link:https://github.com/minishift/minishift-centos-iso/releases[CentOS]. This ISO contains
+the link:https://www.centos.org/[CentOS] operating system, which is an enterprise-ready Linux
+distribution.
++
+NOTE: The CentOS ISO image size is larger than the Boot2Docker image, because CentOS is a
```
hferentschik: Should we not somewhere mentions why one should choose one over the other.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: `Choosing the ISO image for the VM` does not sound good. It is understood that the ISO is for VM.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: +1 @hferentschik .  I do not think we need to explain this here.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: +1. Lets mention that user can use one ISO at a time.

docs/source/using/managing-minishift.adoc
thatdocslady: Because when discussing this with @LalatenduMohanty and @gbraad on IRC it was determined that maintaining an example URL through the different ISO versions would be cumbersome (i.e. to make sure that it points to the correct release). Therefore, I added a link to the releases page where users can grab the latest URL, and used the format instead. Lala approved this step.

docs/source/using/managing-minishift.adoc
thatdocslady: I agree, what is indeed the use-case for changing the default?

docs/source/using/managing-minishift.adoc
thatdocslady: Ok, I was trying to illustrate the end-to-end case but I'm not emotionally attached to the latter part. It was more important for me to be able to describe what happens from ISO image > VM.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: There is no correlation between size and enterprise distribution. At least we must not not deduce this.  So we should mention that `  The CentOS ISO image size is larger than the Boot2Docker image, however CentOS is enterprise Linux distribution.`

docs/source/using/managing-minishift.adoc
LalatenduMohanty: s/ path to the ISO image/ path to the local copy of ISO image/

docs/source/using/managing-minishift.adoc
thatdocslady: @LalatenduMohanty in the description of issue #849 you specifically mentioned this correlation. Please double-check and let me know what you wish me to write here.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: @hferentschik We could not came up with a single command which will be applicable for all CentOS releases at this point. If you have a suggestion let us know.

hferentschik: > maintaining an example URL through the different ISO versions would be cumbersome 

Why is this cumbersome. The only moving part is the version, right? Why not use an variable for this? @LalatenduMohanty?

docs/source/using/managing-minishift.adoc
LalatenduMohanty: @thatdocslady WHat I mentioned in the issue is correct i.e.  `We should also mention that CentOS ISO is of higher size as compared to minishift-b2d but it is an enterprise GNU/Linux distribution where as boot2docker is not` AFAIU this very different than `The CentOS ISO image size is larger than the Boot2Docker image, because CentOS is a full-featured enterprise Linux operating system.`

You have replaced `but` with `because` and this completely changes the meaning.

thatdocslady: @LalatenduMohanty I replaced the words because to me it seemed to be implied that more features = larger image size, sorry if I misunderstood. 

To backtrack, I still say that the use of the word "but" doesn't make sense here, in English the word "but" goes usually to cancel or contradict the previous part of the sentence. This is why I'm asking what do you actually want the users to take away from this sentence, so that we can phrase it in a clearer way. Thanks for clarifying.

LalatenduMohanty: @thatdocslady The idea here is to tell users about the sizes of the ISO and also the advantage of both the ISO images so that user can take a conscious decision. When we compare both the ISOs , boot2docker ISO size is a lot less than CentOS. Hence users who are conscious about the size would go for boot2docker. However the downside is, Boot2Docker is not a Linux distribution which can be run in production. CentOS on the other hand  runs in production i.e. enterprise Linux/GNU distribution.  So user who want to have development environment (i.e. OpenShift on local workstation) similar to the production would opt for CentOS, even if it is higher size. Hope this clarify things.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: @hferentschik Yes, that sounds good. We can always have the command for tested version (mostly latest) CentOS ISO. However we need to change Makefile for this isn't it?

docs/source/using/managing-minishift.adoc
hferentschik: I agree with @LalatenduMohanty, one cannot conclude that the image size is bigger, because it is Enterprise Linux. On the other hand, I agree as well with @thatdocslady that the use of "but" is a bit unfortunate. Maybe we just need to elaborate more and split this into two or more sentences .

docs/source/using/managing-minishift.adoc
hferentschik: > However we need to change Makefile for this isn't it?

Yes, but that is trivial. All the mechanism are in place. We need to add the CentOS ISO version to the Makefile. This is something which we will need for #428 or whatever we come up with in order to let the user choose the CentOS image easier. 

We should in this context probably introduce CENTOS_ISO_VERSION and rename ISO_VERSION to B2D_ISO_VERSION

gbraad:
I'll do a review when most of the storm has calmed down. Just ping me later
when most of the comments have been dealt with.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: @hferentschik +1. Lets do that. Once @thatdocslady's PR get finalized , I will add another commit for the the change as we discussed and send for review i.e. the makefile change etc.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: @thatdocslady In case you are wondering why boot2docker is not  used in production , here is some pointers https://github.com/boot2docker/boot2docker#caveat-emptor

thatdocslady: Ok @LalatenduMohanty @hferentschik now it makes a lot more sense, and it also probably can address the "why choose one or the other" question that Hardy raised earlier.

thatdocslady:
@LalatenduMohanty @hferentschik updated again with new comments. I think it's better if we focus on the docs in this PR and then open a new issue + PR for adding the CentOS ISO image version to the makefile (easier tracking). Then we can just fix the link and add the variable as a part of that PR.

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,42 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+The following ISO images are available:
+
+- link:https://github.com/minishift/minishift-b2d-iso/releases[Boot2Docker] (Default). This ISO is
+based on the link:https://github.com/boot2docker/boot2docker[Boot2Docker] ISO, which is a
+lightweight Linux distribution optimized for containerized application development. This ISO image
+is optimized for development but not for production, and is much smaller in size from the CentOS ISO image.
+
+- link:https://github.com/minishift/minishift-centos-iso/releases[CentOS]. This ISO contains
+the link:https://www.centos.org/[CentOS] operating system, which is an enterprise-ready Linux
+distribution. This ISO image provides a production-ready environment.
```
hferentschik: I think we should be saying something like:

CentOS operating system, which is an enterprise-ready Linux more mirroring a production environment for OpenShift.

> This ISO image provides a production-ready environment.

I don't think we can say this. Minishift does not provide a production environment. It should never be used in a production setting. We should not imply this. The image mirrors more a production environment for an actual OpenShift install.

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,42 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+----
+- Download the CentSO ISO image from the link:https://github.com/minishift/minishift-centos-iso/releases[releases page]
+and enter the path to the ISO image location in the following format:
++
+----
+$ minishift start --iso-url file:///<path_to_ISO_image>
+----
+
+NOTE: You cannot run Minishift with both ISO images concurrently. To switch between
+ISO images, you must stop and restart Minishift with the ISO image that you want to use.
```
hferentschik: _stop and restart_, that won't help. This will just restart the existing VM. You need to delete the Minishift instance. So you need a `minishift delete` followed by a `minishift start` with the correct iso-url.

gbraad:
>
> *stop and restart*, that won't help. This will just restart the existing
> VM. You need to delete the Minishift instance.
>
Recreate would be a better way to describe the action at first, and then
explain the delete step involved

thatdocslady:
Updated based on latest feedback.

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,43 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+The following ISO images are available:
+
+- link:https://github.com/minishift/minishift-b2d-iso/releases[Boot2Docker] (Default). This ISO is
+based on the link:https://github.com/boot2docker/boot2docker[Boot2Docker] ISO, which is a
+lightweight Linux distribution optimized for containerized application development. This ISO image
+is optimized for development but not for production, and is much smaller in size from the CentOS ISO image.
+
+- link:https://github.com/minishift/minishift-centos-iso/releases[CentOS]. This ISO contains
+the link:https://www.centos.org/[CentOS] operating system, which is an enterprise-ready Linux
```
hferentschik: :+1 "Resembles", that's the word I was looking for

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,43 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+----
+- Download the CentSO ISO image from the link:https://github.com/minishift/minishift-centos-iso/releases[releases page]
+and enter the path to the ISO image location in the following format:
++
+----
+$ minishift start --iso-url file:///<path_to_ISO_image>
+----
+
+NOTE: You cannot run Minishift with both ISO images concurrently. To switch between
+ISO images, you must stop Minishift, delete the Minishift instance, and start a new
```
hferentschik: A lot of "Minishift". Would this not read better (no need to stop and delete, delete is enough)

"To switch between ISO images, you must delete the Minishift instance and start a new one with the ISO image you want to use."

hferentschik:
@LalatenduMohanty, so you want to take this over and add a commit on top of this taking care of the variable part?

docs/source/using/managing-minishift.adoc
thatdocslady: Aha so you can just delete without stopping. Cool, I'll fix that.

LalatenduMohanty:
> I think it's better if we focus on the docs in this PR and then open a new issue + PR for adding the CentOS ISO image version to the makefile (easier tracking). Then we can just fix the link and add the variable as a part of that PR.

@thatdocslady the change is related to this PR and should be straight forward. So lets handle this in this PR.  I will add just one more commit on top of this PR as per plan.

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,42 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+When you start Minishift, it downloads a live ISO image that the hypervisor uses to
+provision the host VM.
+
+The following ISO images are available:
+
+- link:https://github.com/minishift/minishift-b2d-iso/releases[Boot2Docker] (Default). This ISO is
+based on the link:https://github.com/boot2docker/boot2docker[Boot2Docker] ISO, which is a
+lightweight Linux distribution optimized for containerized application development. This ISO image
+is optimized for development but not for production, and is much smaller in size from the CentOS ISO image.
```
LalatenduMohanty: I think we should first tell that we have two ISO options i.e. minishift-b2d and minishift-centos for Minishift and then talk about difference between them. Else we are talking about the `minishift-centos` ISO when the reader does not know about it.

LalatenduMohanty: s/[Boot2Docker]/[minishift-b2d/ and similarly `minishift-centos` ISO. As these are modified for Minishift project.

LalatenduMohanty: Also we should improve the wordings e.g. `This ISO image is based on the link:https://github.com/boot2docker/boot2docker[Boot2Docker] ISO  which is a lightweight Linux distribution optimized for development but not recommended for production use.  The size of minishift-b2d ISO image is much smaller as compared to the minishift-centos ISO image.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: s/[CentOS]/minishift-centos/

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,42 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+- link:https://github.com/minishift/minishift-centos-iso/releases[CentOS]. This ISO contains
+the link:https://www.centos.org/[CentOS] operating system, which is an enterprise-ready Linux
+distribution that more closely resembles a production-ready environment.
```
LalatenduMohanty: s/production-ready/production/

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,42 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+By default, Minishift uses the **Boot2Docker** image. To choose the CentOS ISO instead,
```
LalatenduMohanty: s/Boot2docker/minishift-b2d/

LalatenduMohanty: s/image/ISO image/

LalatenduMohanty: s/ISO/ISO image/

LalatenduMohanty:
@thatdocslady I think I we should use actual ISO names i.e. minishift-b2d and minishift-centos instead of Boot2Docker and CentOS as these are modified for Minishift project. Not sure what others think cc @gbraad @hferentschik @praveenkumar

thatdocslady:
@LalatenduMohanty I recommend using more natural English names (rather than the technical shorthand), and initially I tried to match with the official names from the repos (adjusted for this topic):

https://github.com/minishift/minishift-b2d-iso - Minishift Boot2Docker ISO
https://github.com/minishift/minishift-centos-iso - Minishift CentOS VM

The adjustments I made were to remove "Minishift" because it is redundant in this context, and we agreed to use "ISO image" rather than "VM" so I didn't use that.

What you refer to as "actual ISO name" is in fact the name of the repo and the technical file name, which shouldn't be the benchmark for official documentation (similar to how we decided to capitalize "Minishift" rather than use "minishift" even though that is the command).

I hope this explanation makes sense, @hferentschik @gbraad @praveenkumar cc

LalatenduMohanty:
> What you refer to as "actual ISO name" is in fact the name of the repo and the technical file name, which shouldn't be the benchmark for official documentation. 

At the end we want the documentation to be technically accurate and should not mislead users.  The upstream Boot2Docker ISO is different than `minishift-b2d` ISO as we have changed it for Minishift. We should clearly communicate that this is a customized Boot2Docker ISO. User should not confuse this with upstream Boot2Docker ISO. Similarly for CentOS it is modified for Minishift. It is not an ISO which users can use for generic CentOS purpose.  At least we are using the correct links underneath so it is not very bad.

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,43 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
...
+
+When you start Minishift, it downloads a live ISO image that the hypervisor uses to
+provision the host VM.
+
+The following ISO images are available:
+
+- link:https://github.com/minishift/minishift-b2d-iso/releases[Boot2Docker] (Default). This ISO is
+based on link:https://github.com/boot2docker/boot2docker[Boot2Docker], which is a
+lightweight Linux distribution optimized for containerized application development. This ISO image
+size is small and optimized for development but not for production.
```
LalatenduMohanty: The word `optimized` and `development` is repeated and it is not looking good at the end. Also `optimized for containerized application development` is technically not correct as Boot2Docker was just created to run Docker containers. Refer [1]. That's the reason I have avoided these strings when I suggested it last time. 
[1] http://boot2docker.io/

docs/source/using/managing-minishift.adoc
```diff
@@ -53,6 +53,43 @@ the GitHub issue https://github.com/minishift/minishift/issues/179[#179].
 The link:../command-ref/minishift_delete{outfilesuffix}[`minishift delete`] command deletes the OpenShift cluster,
 and also shuts down and deletes the Minishift VM. No data or state are preserved.
 
+[[choosing-iso-image]]
+== Choosing the ISO image
+
+When you start Minishift, it downloads a live ISO image that the hypervisor uses to
+provision the host VM.
```
LalatenduMohanty: s/provision the host VM/provision the minishift VM/ or s/provision the host VM/provision the VM/. `host VM` seems incorrect to me.

hferentschik: Minishift VM? This is what we say at other places, right?

docs/source/using/managing-minishift.adoc
hferentschik: Maybe "This ISO is based on a lightweight Linux distribution made specifically to run Docker containers"

thatdocslady:
@LalatenduMohanty in that case please let me adjust the sentence around that code example so that it makes sense for the URL string. Will update the commit now.

hferentschik:
> the change is related to this PR and should be straight forward. So lets handle this in this PR. I will add just one more commit on top of this PR as per plan.

+1

LalatenduMohanty:
Superseded by https://github.com/minishift/minishift/pull/973

PR #962, opened 2017-05-29
Issue #409 Refactoring start.go

hferentschik:
- Introducing ClusterUpConfig struct to separate Viper flags processing from 'cluster up' call
- Moving ClusterUp and PostClusterUp into own package under pks, since they are now independent of the command
- Updated tests


hferentschik:
Need to work out some integration issues ...

praveenkumar:
@hferentschik you need to rebase and solve those conflicts :)

hferentschik:
> you need to rebase and solve those conflicts :)

I did another rebase

cmd/minishift/cmd/start_test.go
```diff
@@ -116,57 +128,62 @@ func (r *RecordingRunner) Output(command string, args ...string) ([]byte, error)
...
 
-var testDir string
-var testRunner *RecordingRunner
-var testMachineConfig = cluster.MachineConfig{
-	OpenShiftVersion: "v1.3.1",
-}
-var testIp = "192.168.99.42"
+var (
+	testConfig = &clusterup.ClusterUpConfig{
+		OpenShiftVersion: "v1.5.1",
```
praveenkumar: Finally version update for unit tests 👍

pkg/minishift/oc/oc.go
```diff
@@ -34,9 +37,9 @@ const (
 )
 
 type OcRunner struct {
-	ocPath         string
-	kubeConfigPath string
-	runner         util.Runner
+	OcPath         string
+	KubeConfigPath string
+	Runner         util.Runner
```
praveenkumar: 👍

pkg/testing/cli/harness.go
```diff
@@ -76,6 +76,14 @@ func CreateExitHandlerFunc(t *testing.T, tee *Tee, expectedExitCode int, expecte
 	return exitHandler
 }
 
+func PreventExitHandlerFunc(t *testing.T) func(int) bool {
```
praveenkumar: Does this mean it will return a function which have return type as `bool` and take a `int` parameter? Just want to understand.

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -75,15 +77,22 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting IP: %s", err.Error()))
 	}
 
-	ocPath := minishiftConfig.InstanceConfig.OcPath
 	routingSuffix := viper.GetString(routingSuffix)
-	kubeConfigPath := constants.KubeConfigPath
 	sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
+	ocPath := minishiftConfig.InstanceConfig.OcPath
+	kubeConfigPath := constants.KubeConfigPath
```
budhrg: Unless, `ocPath` or `kubeConfigPath` is used multiple times in later part of this file, we can directly pass these values.

cmd/minishift/cmd/addon/util.go
```diff
@@ -20,21 +20,14 @@ import (
...
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 )
 
-const (
-	ip_key             = "ip"
-	routing_suffix_key = "routing-suffix"
-)
-
-// getAddOnManager returns the addon manager
+// GetAddOnManager returns the addon manager
```
budhrg: Same change I did in my PR, looks like a conflict when I do rebase :wink:

cmd/minishift/cmd/start.go
```diff
@@ -115,36 +108,16 @@ func runStart(cmd *cobra.Command, args []string) {
...
-		InsecureRegistry: insecureRegistry,
-		RegistryMirror:   registryMirror,
-		HostOnlyCIDR:     viper.GetString(startFlags.HostOnlyCIDR.Name),
-		OpenShiftVersion: viper.GetString(startFlags.OpenshiftVersion.Name),
-		ShellProxyEnv:    shellProxyEnv,
-	}
-
-	fmt.Printf("Starting local OpenShift cluster using '%s' hypervisor...\n", machineConfig.VMDriver)
-
+	// to determine whether we need to run post cluster up actions,
```
budhrg: `to determine` to `Determine`. We have started comments with Caps (at least I am following it).

cmd/minishift/cmd/start.go
```diff
@@ -115,36 +108,16 @@ func runStart(cmd *cobra.Command, args []string) {
...
-		return err
-	}
-	err := util.Retry(3, start)
-	if err != nil {
-		fmt.Println("Error starting the VM: ", err)
-		atexit.Exit(1)
+	hostVm, ip := startHost(libMachineClient)
+
+	if proxyConfig.IsEnabled() {
+		// once we know the IP, we need to make sure it is not proxied in a proxy environment
```
budhrg: Same here, `once we know` to `Once we know`.

cmd/minishift/cmd/start.go
```diff
@@ -154,42 +127,50 @@ func runStart(cmd *cobra.Command, args []string) {
...
+		OcPath:           cacheOc(openShiftVersion),
 	}
 
-	automountHostfolders(host.Driver)
-
-	clusterUp(&machineConfig, ip)
+	clusterUpParams := determineClusterUpParameters(clusterUpConfig)
+	err = clusterup.ClusterUp(clusterUpConfig, clusterUpParams, &util.RealRunner{})
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error during'cluster up' execution: %v", err))
```
budhrg: Space missing after `Error during` word.

cmd/minishift/cmd/start.go
```diff
@@ -319,29 +243,15 @@ func automountHostfolders(driver drivers.Driver) {
...
-	startCmd.Flags().AddFlagSet(startFlagSet)
-	startCmd.Flags().AddFlagSet(clusterUpFlagSet)
-	startCmd.Flags().AddFlagSet(subscriptionManagerFlagSet)
-
-	viper.BindPFlags(startCmd.Flags())
-	RootCmd.AddCommand(startCmd)
-
-	provision.SetDetector(&provisioner.MinishiftProvisionerDetector{Delegate: provision.StandardDetector{}})
-}
+// initStartFlags creates the CLI flags which needs to be passed on to 'lib-machine'
```
budhrg: `lib-machine` to `libmachine`. I think `libmachine` is more familiar than `lib-machine`.

cmd/minishift/cmd/start.go
```diff
@@ -371,44 +285,31 @@ func initClusterUpFlags() {
 	clusterUpFlagSet.String(startFlags.NoProxyList.Name, "", "List of hosts or subnets for which no proxy should be used.")
 	clusterUpFlagSet.AddFlag(httpProxyFlag)
 	clusterUpFlagSet.AddFlag(httpsProxyFlag)
+
+	return clusterUpFlagSet
 }
 
 // initProxyFlags create the CLI flags which needs to be passed for proxy
```
budhrg: Wrong comment.

cmd/minishift/cmd/start.go
```diff
@@ -371,44 +285,31 @@ func initClusterUpFlags() {
...
-func initSubscriptionManagerFlags() {
+func initSubscriptionManagerFlags() *flag.FlagSet {
+	var subscriptionManagerFlagSet = flag.NewFlagSet(commandName, flag.ContinueOnError)
```
budhrg: Why not `subscriptionManagerFlagSet := flag.NewFlagSet(commandName, flag.ContinueOnError)`?

cmd/minishift/cmd/start_test.go
```diff
@@ -116,57 +128,62 @@ func (r *RecordingRunner) Output(command string, args ...string) ([]byte, error)
...
+		Ip:               "192.168.99.42",
+		OcPath:           "/home/john/.minishift/cache/oc/1.5.0/oc",
```
budhrg: `OcPath` should have `1.5.1` as version.

pkg/minishift/clusterup/clusterup.go
```diff
@@ -0,0 +1,227 @@
...
+	"github.com/minishift/minishift/pkg/minishift/oc"
+	"github.com/minishift/minishift/pkg/util"
+	"github.com/minishift/minishift/pkg/version"
+	"os"
+	"strings"
+	"time"
+)
+
+const (
+	ip_key             = "ip"
```
budhrg: Let change `ip_key` to `ipKey` and `routing_suffix_key` to `routingSuffixKey`.

pkg/minishift/clusterup/clusterup.go
```diff
@@ -0,0 +1,227 @@
...
+	if exitCode != 0 {
+		errors.New("Error starting the cluster.")
+	}
+	return nil
+}
+
+// PostClusterUp runs the Minishift specific provisioning after 'cluster up' has run
+func PostClusterUp(clusterUpConfig *ClusterUpConfig, sshCommander provision.SSHCommander, addOnManager *manager.AddOnManager) error {
+	err := kubeconfig.CacheSystemAdminEntries(clusterUpConfig.KubeConfigPath, getConfigClusterName(clusterUpConfig.Ip, clusterUpConfig.Port))
+	if err != nil {
```
budhrg:
```
if err != nil {
   return err
}
```
is the repeating pattern in this method. 

Seems to be some kind of code smelling here. Let do some refactoring here.

pkg/minishift/clusterup/clusterup.go
```diff
@@ -0,0 +1,227 @@
...
+
+	err = applyAddOns(addOnManager, clusterUpConfig.Ip, clusterUpConfig.RoutingSuffix, ocRunner, sshCommander)
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
+
+//EnsureHostDirectoriesExist Ensures the specified directories exist on the virtual machine and creates them if not
```
budhrg: Space missing in between `//` and `EnsureHostDirectoriesExist`

Also, comment should be
```
// EnsureHostDirectoriesExist ensures that the specified directories exist on the VM and creates them if does not exist
```

pkg/minishift/clusterup/clusterup.go
```diff
@@ -0,0 +1,227 @@
...
+	cmd := fmt.Sprintf("sudo mkdir -p %s", strings.Join(dirs, " "))
+	_, err := host.RunSSHCommand(cmd)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// DetermineOpenShiftVersion returns the OpenShift/oc version to use. If the requested version is
+// < the base line version we use baseline oc to provision the requested OpenShift version. If the requested
```
budhrg: Comma missing after `< the base line version`. Took me few  seconds to process this line :smile:  
Also, I feel if we write the comments one comment per line, then it will increase readability.

```
DetermineOpenShiftVersion returns the OpenShift/oc version to use. 
If the requested version is < the base line version we use baseline oc to provision the requested OpenShift version. 
If the requested OpenShift version is >= the baseline, we align the oc version with the requested OpenShift version.
```

pkg/minishift/clusterup/clusterup.go
```diff
@@ -0,0 +1,227 @@
...
+// OpenShift version is >= the baseline, we align the oc version with the requested OpenShift version.
+func DetermineOpenShiftVersion(requestedVersion string) string {
+	if !ValidateOpenshiftMinVersion(requestedVersion, version.GetOpenShiftVersion()) {
+		requestedVersion = version.GetOpenShiftVersion()
+	}
+
+	return requestedVersion
+}
+
+func GetExecutionContext(ip string, routingSuffix string, ocRunner *oc.OcRunner, sshCommander provision.SSHCommander) (*command.ExecutionContext, error) {
```
budhrg: Comment here will help more to understand this method well.

pkg/testing/cli/harness.go
hferentschik: So what's happening in some tests is, that if they are setup wrong and the code actually exits (using an atexit method for example), the program/test execution will stop directly. This means that no assertions will be run marking the test effectively as passed. Even worse, subsequent tests won't run. We had this in other test as well. 

Using this handler an os.Exit will be converted to a panic. This means the test will actually fail (with a panic, but at least it does not get unnoticed). 

I will try to find better name and add some godocs. Hope it becomes clearer then

cmd/minishift/cmd/addon/addon_apply.go
hferentschik: hmm, to humour you. Usually, I prefer shorter lines.

cmd/minishift/cmd/addon/util.go
hferentschik: :-)

cmd/minishift/cmd/start.go
hferentschik: fair enough

cmd/minishift/cmd/start.go
hferentschik: good catch

cmd/minishift/cmd/start.go
hferentschik: same same really. I guess I copied the 'var subscriptionManagerFlagSet' part originally from the top of the file. 

I guess we usually use the :=, so I am going to change it

cmd/minishift/cmd/start_test.go
hferentschik: that's really irrelevant here. I could have used /foo/bar here. In fact I was doing so initially, partly to make clear that this is not a real oc path. I am changing it to 1.5.1, but also considering using a pure dummy string.

pkg/minishift/clusterup/clusterup.go
hferentschik: Is it? That's just usual Go error handling to me. I am intrigued to know how you want to refactor this? I making various which potentially can return an error. I need to check whether an error occurred and act accordingly. In this case I just want to pass it on.

cmd/minishift/cmd/start_test.go
budhrg: @hferentschik , if that is the case, let's put dummy string here then. This path make me feel like we might be coming some version here :)

pkg/minishift/clusterup/clusterup.go
budhrg: I am thinking on this line.
```
func checkAndReturnError(err error) {
       if err != nil {
            return err
       }
}

// Call above instead of three lines
```
Choose the name accordingly :smile:

budhrg: Fair enough.
At least, if possible let's use this pattern : 
```
if err := ocRunner.AddSudoerRoleForUser(clusterUpConfig.User); err != nil {
	return err
}
```

pkg/minishift/clusterup/clusterup.go
```diff
@@ -0,0 +1,230 @@
...
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// DetermineOpenShiftVersion returns the OpenShift/oc version to use.
+// If the requested version is < the base line version we use baseline oc to provision the requested OpenShift version.
+// If the requested OpenShift version is >= the baseline, we align the oc version with the requested OpenShift version.
+// It also handles the case the user just specifies a version without the required 'v' prefix
```
budhrg: :+1:

pkg/minishift/clusterup/clusterup.go
```diff
@@ -0,0 +1,230 @@
...
+func DetermineOpenShiftVersion(requestedVersion string) string {
+	if !ValidateOpenshiftMinVersion(requestedVersion, version.GetOpenShiftVersion()) {
+		requestedVersion = version.GetOpenShiftVersion()
+	}
+
+	return requestedVersion
+}
+
+// GetExecutionContext createds an ExecutionContext for executing add-on.
```
budhrg: Typo here @hferentschik  :smile:

budhrg:
@hferentschik , I am done with PR review. Please fix the typo and this change https://github.com/minishift/minishift/pull/962#discussion_r119270241 .

Rest LGTM :+1:

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -75,15 +77,20 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
...
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error applying addon: %s", err.Error()))
+	}
 
 	for i := range args {
 		addonName := args[i]
 		addon := addOnManager.Get(addonName)
-		addonContext := GetExecutionContext(ip, routingSuffix, ocPath, kubeConfigPath, sshCommander)
+		addonContext, err := clusterup.GetExecutionContext(ip, routingSuffix, ocRunner, sshCommander)
+		if err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprint("Error executing addon commands: ", err))
```
LalatenduMohanty: I think you missed the `%s`.

LalatenduMohanty:
The code looks good. I did test the `minishift start` on KVM and it worked fine.

cmd/minishift/cmd/addon/addon_apply.go
praveenkumar: @LalatenduMohanty No %s required when you use `Sprintf`

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -75,15 +77,20 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting IP: %s", err.Error()))
 	}
 
-	ocPath := minishiftConfig.InstanceConfig.OcPath
 	routingSuffix := viper.GetString(routingSuffix)
-	kubeConfigPath := constants.KubeConfigPath
 	sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
+	ocRunner, err := oc.NewOcRunner(minishiftConfig.InstanceConfig.OcPath, constants.KubeConfigPath)
```
gbraad: I wonder why this is not used in the tests. At the moment we use: 	`"github.com/minishift/minishift/test/integration/util"`

pkg/minishift/clusterup/clusterup.go
hferentschik: not for me. If you prefer this syntax go for it, but I don't like it at all.

pkg/minishift/clusterup/clusterup.go
hferentschik: reads indeed better :-)

pkg/minishift/clusterup/clusterup.go
hferentschik: got it, thanks

cmd/minishift/cmd/addon/addon_apply.go
hferentschik: Right, not required here. Different method as @praveenkumar is saying

cmd/minishift/cmd/addon/addon_apply.go
hferentschik: It should be now. I think historically the "stub" OcRunner in the integration test was there first. The fully fleshed out runner as part of the production code is not around for so long. So yes, it could/should be used in the integration tests.

hferentschik:
Pushed another update.

cmd/minishift/cmd/addon/addon_apply.go
gbraad: @LalatenduMohanty you can actually write `"Hello", "World", "!"`
https://play.golang.org/p/3Mqog0waoH

cmd/minishift/cmd/start.go
```diff
@@ -83,68 +76,49 @@ var (
...
+	provision.SetDetector(&provisioner.MinishiftProvisionerDetector{Delegate: provision.StandardDetector{}})
 }
 
-// runStart is executed as part of the start command
+// runStart handles all command line arguments, launches the VM and provisions OpenShift
 func runStart(cmd *cobra.Command, args []string) {
 	libMachineClient := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer libMachineClient.Close()
 
+	ensureNotRunning(libMachineClient, constants.MachineName)
```
gbraad: +1 (does it do error reporting?)

cmd/minishift/cmd/start.go
```diff
@@ -154,42 +128,50 @@ func runStart(cmd *cobra.Command, args []string) {
...
 
 	//Create the host directories if not present
-	hostDirs := []string{viper.GetString(startFlags.HostConfigDir.Name),
-		viper.GetString(startFlags.HostDataDir.Name), viper.GetString(startFlags.HostVolumeDir.Name),
-		viper.GetString(startFlags.HostPvDir.Name)}
-	err = clusterup.EnsureHostDirectoriesExist(libMachineClient, hostDirs)
-	if err != nil {
-		fmt.Println("Error creating required host directories: ", err)
-		atexit.Exit(1)
+	hostDirs := []string{
```
gbraad: I prefer to extract to function.

cmd/minishift/cmd/start.go
```diff
@@ -154,42 +128,50 @@ func runStart(cmd *cobra.Command, args []string) {
...
+		viper.GetString(startFlags.HostConfigDir.Name),
+		viper.GetString(startFlags.HostDataDir.Name),
+		viper.GetString(startFlags.HostVolumeDir.Name),
+		viper.GetString(startFlags.HostPvDir.Name),
 	}
-
-	ip, err := host.Driver.GetIP()
+	err = clusterup.EnsureHostDirectoriesExist(hostVm, hostDirs)
```
gbraad: ... which can be used as the input for this function

cmd/minishift/cmd/start.go
```diff
@@ -154,42 +128,50 @@ func runStart(cmd *cobra.Command, args []string) {
...
 	if err != nil {
-		fmt.Println("Error determining host ip: ", err)
-		atexit.Exit(1)
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error creating required host directories: %v", err))
```
gbraad: and extract even further or combine as:
```
    if err := clusterup.EnsureHostDirectoriesExist(hostVm, getHostDirsFromViper()); err != nil {
        atexit.ExitWithMessage(1, fmt.Sprintf("Error creating required host directories: %v", err))
    }
```

cmd/minishift/cmd/start.go
```diff
@@ -154,42 +128,50 @@ func runStart(cmd *cobra.Command, args []string) {
...
 	}
 
-	if proxyConfig.IsEnabled() {
-		// once we know the IP, we need to make sure it is not proxied in a proxy environment
-		proxyConfig.AddNoProxy(ip)
-		proxyConfig.ApplyToEnvironment()
+	autoMountHostFolders(hostVm.Driver)
+
+	openShiftVersion := clusterup.DetermineOpenShiftVersion(viper.GetString(startFlags.OpenshiftVersion.Name))
```
gbraad: Suggest extracting into a meaningful function...

cmd/minishift/cmd/start.go
```diff
@@ -154,42 +128,50 @@ func runStart(cmd *cobra.Command, args []string) {
...
+		OpenShiftVersion: openShiftVersion,
+		MachineName:      constants.MachineName,
+		Ip:               ip,
+		Port:             constants.APIServerPort,
+		RoutingSuffix:    viper.GetString(startFlags.RoutingSuffix.Name),
+		HostPvDir:        viper.GetString(startFlags.HostPvDir.Name),
+		User:             defaultUser,
+		Project:          defaultProject,
+		KubeConfigPath:   constants.KubeConfigPath,
+		OcPath:           cacheOc(openShiftVersion),
```
gbraad: ... and use here; as it is likely the only place where it is used.

cmd/minishift/cmd/start.go
```diff
@@ -154,42 +128,50 @@ func runStart(cmd *cobra.Command, args []string) {
...
 	}
 
-	automountHostfolders(host.Driver)
-
-	clusterUp(&machineConfig, ip)
+	clusterUpParams := determineClusterUpParameters(clusterUpConfig)
+	err = clusterup.ClusterUp(clusterUpConfig, clusterUpParams, &util.RealRunner{})
```
gbraad: Make the code more stand out as related
```
if err := function(); err != nil {
   ...
}
```

cmd/minishift/cmd/start.go
```diff
@@ -154,42 +128,50 @@ func runStart(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error during 'cluster up' execution: %v", err))
+	}
 
 	if !isRestart {
```
gbraad: create a function for this...

cmd/minishift/cmd/start.go
```diff
@@ -215,101 +197,44 @@ func handleProxies() *util.ProxyConfig {
...
-
-	fmt.Println("OK")
-	fmt.Println()
-
-	return nil
+	return hostVm, ip
 }
 
-func automountHostfolders(driver drivers.Driver) {
+func autoMountHostFolders(driver drivers.Driver) {
```
gbraad: so you did want to change it ;-). had this same question when working on the task.

cmd/minishift/cmd/start.go
```diff
@@ -439,35 +353,35 @@ func setDefaultRoutingPrefix(ip string) {
...
 
-func validateOpenshiftVersion() {
-	if viper.IsSet(startFlags.OpenshiftVersion.Name) {
-		if !minishiftUtil.ValidateOpenshiftMinVersion(viper.GetString(startFlags.OpenshiftVersion.Name), constants.MinOpenshiftSuportedVersion) {
-			fmt.Printf("Minishift does not support Openshift version %s ."+
-				"You need to use a version >=%s\n", viper.GetString(startFlags.OpenshiftVersion.Name),
-				constants.MinOpenshiftSuportedVersion)
-			atexit.Exit(1)
-		}
+func ensureNotRunning(client *libmachine.Client, machineName string) {
```
gbraad: so the function does exit... I consider this kind of a side-effect. however the name `Ensure` kind of implies this...

cmd/minishift/cmd/start.go
hferentschik: :-)

cmd/minishift/cmd/start.go
hferentschik: Not quite sure what you are saying.

cmd/minishift/cmd/start.go
hferentschik: for what? the content in the if statement?

cmd/minishift/cmd/start.go
hferentschik: What do you mean? What do you want me to extract?

cmd/minishift/cmd/start.go
gbraad: @hferentschik related to an earlier comment: https://github.com/minishift/minishift/pull/962#discussion_r119299583

cmd/minishift/cmd/start.go
hferentschik: In case you are talking about openShiftVersion, it is used twice

cmd/minishift/cmd/start.go
hferentschik: I am not a big fan of this syntax for various reasons

cmd/minishift/cmd/start.go
gbraad: Yes . It seems to be something like `performPostClusterUpConfiguration() {}`

cmd/minishift/cmd/start.go
gbraad: but you do use it later on... it makes the code more cohesive.

cmd/minishift/cmd/start.go
gbraad:
```
function determineOpenShiftVersion() {
    return clusterup.DetermineOpenShiftVersion(viper.GetString(startFlags.OpenshiftVersion.Name))
}
```

cmd/minishift/cmd/start.go
gbraad: I see yeha, maybe the `cacheOc` should not happen here. It is a function that does too much

cmd/minishift/cmd/addon/addon_apply.go
LalatenduMohanty: Got it.  I thought it is `Sprintf`

cmd/minishift/cmd/start.go
hferentschik: It is part of the pre-flight checks we recently discussed. I noticed that if I had a machine running and did a 'minishift start', it would still do cluster up stuff. Effectively to no prevail. This is a check which will check whether the VM is running and if so, it will print a message of the form "Minishift VM already running".

cmd/minishift/cmd/start.go
hferentschik: Is clusterup.EnsureHostDirectoriesExist not the function you are after? The thing is, I need to provide this function with the set of directories to check/create. you want me to put 

	hostDirs := []string{
		viper.GetString(startFlags.HostConfigDir.Name),
		viper.GetString(startFlags.HostDataDir.Name),
		viper.GetString(startFlags.HostVolumeDir.Name),
		viper.GetString(startFlags.HostPvDir.Name),
	}

into a separate function?

hferentschik: I put it into a function. I guess the benefit is to a certain degree the gain in clarity the function name provides.

cmd/minishift/cmd/start.go
hferentschik: got you. Done.

cmd/minishift/cmd/start.go
hferentschik: I am lost here. We are are talking about a single line here, and you want me to make three lines and a function call out of this. AFAICT it also does not increase clarity here. 

> as it stands now, the GetString is likely used to override the value we seed. 

It does not override anything, it just gets it from Viper. The return value from clusterup.DetermineOpenShiftVersion does an override of sort, but I think that is well documented in the Godocs of clusterup.DetermineOpenShiftVersion.

cmd/minishift/cmd/start.go
hferentschik: Sure, I can do that as well. In this case it becomes something like:

       ocPath := cacheOc(openShiftVersion)
	clusterUpConfig := &clusterup.ClusterUpConfig{
		OpenShiftVersion: openShiftVersion,
		MachineName:      constants.MachineName,
		Ip:               ip,
		Port:             constants.APIServerPort,
		RoutingSuffix:    viper.GetString(startFlags.RoutingSuffix.Name),
		HostPvDir:        viper.GetString(startFlags.HostPvDir.Name),
		User:             defaultUser,
		Project:          defaultProject,
		KubeConfigPath:   constants.KubeConfigPath,
		OcPath:           ocPath,
	}

TBH, I prefer this as well. However, I got several other comments where people thing it is "better" to rather inlines a function call into another function call, over having two separate lines.

> maybe the cacheOc should not happen here. It is a function that does too much

it needs to happen here or just before creating the struct. _ocPath_ downloads and caches the _oc_ binary based on the OpenShift version we are after. It returns the path to the cached binary. I am not sure why you think it is doing too much.

That said, I am fine to split this into two lines. Note though, that I still need the openShiftVersion twice.

cmd/minishift/cmd/start.go
hferentschik: I usually never use this syntax. I actually think the majority of code is using the style I am using here.

cmd/minishift/cmd/start.go
hferentschik: +1

hferentschik: done

hferentschik:
@gbraad I pushed another update addressing most of your comments. I think I commented also on all of your comments, but if I missed something let me know.

LalatenduMohanty:
@gbraad Going to merge this as your comments are addressed.

Issue #826, opened 2017-04-26
Update docker-machine dependency to latest released


praveenkumar:
Currently we are consuming `0.9.0` version which is quite older and `0.11.0` is what right now latest and have quite of fixes/enhancement from driver and provisioner side. This is not going to be just a version change in the glide.yaml but also we need to work around our provisioner to make sure to use systemd drop-in files instead of over-writing the system units which changed during `0.10.0` release of docker machine code.

hferentschik:
Merged via pull request #969

PR #969, opened 2017-05-30
Issue #826 Update to docker-machine 0.11.0

praveenkumar:
Addresses issue #826

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -1,7 +1,7 @@
 package: github.com/minishift/minishift
 import:
 - package: github.com/docker/machine
-  version: 0.9.0
+  version: 0.11.0
   subpackages:
   - drivers/hyperv
   - drivers/virtualbox
--- a/pkg/minishift/provisioner/utils.go
+++ b/pkg/minishift/provisioner/utils.go
@@ -27,19 +27,8 @@ import (
 )
 
 var (
-	engineConfigTemplateRHEL = `[Unit]
-Description=Docker Application Container Engine
-Documentation=http://docs.docker.com
-After=network.target rc-local.service
-Requires=rhel-push-plugin.socket
-Requires=docker-cleanup.timer
-
-[Service]
-Type=notify
-NotifyAccess=all
-Environment=GOTRACEBACK=crash
-Environment=DOCKER_HTTP_HOST_COMPAT=1
-Environment=PATH=/usr/libexec/docker:/usr/bin:/usr/sbin
+	engineConfigTemplateRHEL = `[Service]
+ExecStart=
 ExecStart=/usr/bin/dockerd-current -H tcp://0.0.0.0:{{.DockerPort}} -H unix:///var/run/docker.sock \
            --authorization-plugin rhel-push-plugin \
            --selinux-enabled \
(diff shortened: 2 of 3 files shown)


pkg/minishift/provisioner/utils.go
```diff
@@ -27,19 +27,8 @@ import (
...
-After=network.target rc-local.service
-Requires=rhel-push-plugin.socket
-Requires=docker-cleanup.timer
-
-[Service]
-Type=notify
-NotifyAccess=all
-Environment=GOTRACEBACK=crash
-Environment=DOCKER_HTTP_HOST_COMPAT=1
-Environment=PATH=/usr/libexec/docker:/usr/bin:/usr/sbin
```
praveenkumar: These changes are required because docker machine is now using system drop-in for configuration https://github.com/docker/machine/commit/b6e85bfadf8f02dd2a85e928b49ed02300292861

LalatenduMohanty:
@praveenkumar nice.  I thought it would need more changes than this :)

glide.yaml
```diff
@@ -1,7 +1,7 @@
 package: github.com/minishift/minishift
 import:
 - package: github.com/docker/machine
-  version: 0.9.0
+  version: 0.11.0
```
hferentschik: Wau, I am also surprised. I seem to recall having several issues when trying this some time back.

hferentschik:
Looking good to me. I'll try it locally as well to verify it works.

pkg/minishift/provisioner/utils.go
hferentschik: So help me understand this. We had the need to add custom provisioning before before, basically adding systemd files, but that's not required anymore and hence can be removed? Correct?

pkg/minishift/provisioner/utils.go
```diff
@@ -52,31 +41,10 @@ ExecStart=/usr/bin/dockerd-current -H tcp://0.0.0.0:{{.DockerPort}} -H unix:///v
            --storage-driver {{.EngineOptions.StorageDriver}} --tlsverify --tlscacert {{.AuthOptions.CaCertRemotePath}} \
            --tlscert {{.AuthOptions.ServerCertRemotePath}} --tlskey {{.AuthOptions.ServerKeyRemotePath}} \
            {{ range .EngineOptions.Labels }}--label {{.}} {{ end }}{{ range .EngineOptions.InsecureRegistry }}--insecure-registry {{.}} {{ end }}{{ range .EngineOptions.RegistryMirror }}--registry-mirror {{.}} {{ end }}{{ range .EngineOptions.ArbitraryFlags }}--{{.}} {{ end }}
-ExecReload=/bin/kill -s HUP $MAINPID
```
hferentschik: So when we remove all this where does it go? All these options are not required anymore?

praveenkumar: @hferentschik yes we don't require those options any more.

pkg/minishift/provisioner/utils.go
hferentschik: > using system drop-in for configuration

what does that mean, systemd dropin?

glide.yaml
praveenkumar: @hferentschik @LalatenduMohanty I think issue was there we had kubernetes dependencies but now it was quite simple only had to take care of systemctl unit file bits.

pkg/minishift/provisioner/utils.go
praveenkumar: > what does that mean, systemd dropin?

https://coreos.com/os/docs/latest/using-systemd-drop-in-units.html have better explanation.

praveenkumar: > basically adding systemd files, but that's not required anymore and hence can be removed? Correct?

we still add some part of systemd files that why we didn't remove it completely but now they are consider as systemd drop-in units instead complete systemd file.

glide.yaml
hferentschik: >  I think issue was there we had kubernetes dependencies but now it was quite simple only had to take care of systemctl unit file bits.

that makes sense. We have a leaner dependency graph now

pkg/minishift/provisioner/utils.go
hferentschik: > https://coreos.com/os/docs/latest/using-systemd-drop-in-units.html have better explanation.

Ahh, that was what I was looking for, thanks

hferentschik: >we still add some part of systemd files that why we didn't remove it completely but now they are consider as systemd drop-in units instead complete systemd file.

ok

hferentschik:
@praveenkumar in which combinations have you tried this. I tried with OpenShift 1.5.1 and 3.6.0-alpha.1 as well as with the B2D and CentOS ISO. I see several startup errors. Right now I am not sure whether I forgot to clean up some state, so I need to retest.

praveenkumar:
> In which combinations have you tried this. I tried with OpenShift 1.5.1 and 3.6.0-alpha.1 as well as with the B2D and CentOS ISO.

I tried with 1.5.1 + b2d and 1.5.1 + centos on OSx (xhyve) and it worked as expected. Since integration test pass which mean it works for 1.5.1 + kvm also.

hferentschik:
I am still seeing issues, but that's seems to be caused by using 3.6.0-alpha.1. The problem I am seeing is also against master even though it does not seem to occur all the time, hence I had not noticed yet. I will need to look a bit deeper into this. 

The docker-machine upgrade seems to work fine.

hferentschik:
Rebased and merged.

hferentschik:
@praveenkumar Maybe now that we updated docker-machine, it makes sense to also look at the B2D ISO and get it to the latest version. We are quite behind. WDYT?

PR #973, opened 2017-05-31
Issue #849 Adding centos-iso-version to the ISO image download URL

LalatenduMohanty:


LalatenduMohanty:
I have not tested my commit as I am having some issue with `make` tasks for generating docs.

docs/source/using/managing-minishift.adoc
```diff
@@ -77,7 +77,7 @@ you can do one of the following actions:
 Minishift CentOS ISO image. For example:
 +
 ----
-$ minishift start --iso-url <Web_location_of_ISO_image>
+$ minishift start --iso-url https://github.com/minishift/minishift-centos-iso/releases/download/{centos-iso-version}/minishift-centos7.iso
 ----
```
LalatenduMohanty: This is not working as expected. Is this because `{centos-iso-version}` is inside a code block? @hferentschik

LalatenduMohanty:
This PR is not working at all  e.g. for `minishift-version`. If I remove the `DOC_VARIABLES` and revert ro the old way of doing things then `minishift-version` works fine but not `centos-iso-version`.

docs/source/using/managing-minishift.adoc
hferentschik: Yes, you need to add '[subs="attributes"]' above the ---- starting the listing. I thought we already used this somewhere.

Makefile
```diff
@@ -15,7 +15,10 @@
 # Various versions - Minishift, default OpenShift, default B2D ISO
 MINISHIFT_VERSION = 1.0.1
 OPENSHIFT_VERSION = v1.5.1
-ISO_VERSION = v1.0.2
+B2D_ISO_VERSION = v1.0.2
+CENTOS_ISO_VERSION = v1.0.0
+DOC_VARIABLES = '-e OPENSHIFT_VERSION=$(OPENSHIFT_VERSION) -e MINISHIFT_VERSION=$(MINISHIFT_VERSION) \
```
hferentschik: Let's move this variable under the comment "# Setup for the docs tasks". The variables are somewhat grouped together

Makefile
```diff
@@ -15,7 +15,10 @@
 # Various versions - Minishift, default OpenShift, default B2D ISO
 MINISHIFT_VERSION = 1.0.1
 OPENSHIFT_VERSION = v1.5.1
-ISO_VERSION = v1.0.2
+B2D_ISO_VERSION = v1.0.2
+CENTOS_ISO_VERSION = v1.0.0
```
hferentschik: 👍

hferentschik:
> This PR is not working at all e.g. for minishift-version. If I remove the DOC_VARIABLES in the Makefile and revert ro the old way of doing things then minishift-version works fine but not cantos-iso-version.

So what's not working? What's the content of the generated variables.adoc? Does it contain the right values? Is the issue just with the centos iso variable? If so, you are missing some asciidoc syntax to make variable substitution work within a code block.

LalatenduMohanty:
> So what's not working? What's the content of the generated variables.adoc? Does it contain the right values?

variables.adoc does not contain any value. 
```
$ cat docs/source/variables.adoc 
:minishift-version: unset
:openshift-version: unset
:centos-iso-version: unset
```

LalatenduMohanty:
> So what's not working? What's the content of the generated variables.adoc?



> If I remove the DOC_VARIABLES in the Makefile and revert ro the old way of doing things then minishift-version works fine but not cantos-iso-version.

@hferentschik  This is what I mean https://github.com/LalatenduMohanty/minishift/commit/7b18daac10ca2382d43151b31d88bd74a8718c24 . It variable substitution not working for me at this moment.

hferentschik:
> variables.adoc does not contain any value.

This is the default for when the underlying env variables are not available, Meaning they have not been passed along. We need to go back to the Makefile. What is the command executed by the Makefile at target execution?

LalatenduMohanty:
> What is the command executed by the Makefile at target execution?

The command is 
```
cd docs && docker run '-e OPENSHIFT_VERSION=v1.5.1 -e MINISHIFT_VERSION=1.0.1 -e CENTOS_ISO_VERSION=v1.0.0' -tiv /Users/lmohanty/code/gocode/src/github.com/minishift/minishift/docs:/home/docs:Z minishift/docs gen
```

> This is the default for when the underlying env variables are not available, Meaning they have not been passed along. We need to go back to the Makefile. What is the command executed by the Makefile at target execution?

Right. when we pass the env variable in quotes i.e. `$ docker run '-e OPENSHIFT_VERSION=v1.5.1' -it busybox` it does not work. The allowed way to pass multiple env variables is either through Dockerfile or through a file. 

I am going back to the old method of passing individual variables at this moment and find out why it is not working for me.

hferentschik:
> Right. when we pass the env variable in quotes i.e. $ docker run '-e OPENSHIFT_VERSION=v1.5.1' -it busybox it does not work. 

So why are they in quotes? Have you tried removing the quotes? You are using the quotes in the definition of the variable. Why? Have you tried removing it there?

> The allowed way to pass multiple env variables is either through Dockerfile or through a file.

No idea what you are saying. Even before this pull request we pass already two environment variables and that works fine. AFAICT you can specify -e multiple times.

> docker run '-e OPENSHIFT_VERSION=v1.5.1 -e MINISHIFT_VERSION=1.0.1 -e CENTOS_ISO_VERSION=v1.0.0' -tiv /Users/lmohanty/code/gocode/src/github.com/minishift/minishift/docs:/home/docs:Z minishift/docs get

Why the quotes, that does not look right

> I am going back to the old method of passing individual variables at this moment and find out why it is not working for me.

Please don't. Can you push back the version you had before? I think there was not so much missing to get it to work. I don't think we should waste the effort you already put into this to get it right?

LalatenduMohanty:
> So why are they in quotes?

> So why are they in quotes? Have you tried removing the quotes? You are using the quotes in the definition of the variable. Why? Have you tried removing it there?

After removing the quotes it started working.

hferentschik:
> After removing the quotes it started working.

👍

hferentschik:
Merged, thanks

Issue #975, opened 2017-05-31
TestStartClusterUpNoFlags fails regularly on travis CI


hferentschik:
This test also accesses the GitHub API, but is not excluded as the others which access GitHub leading to unwanted build failures. We should disable it as well.

hferentschik:
Actually the problem is the access to the tag info which is currently not mocked. So far we ignored mocking it, since we thought these request would not exceed the rate limit. I guess we now have to mock them as well.

hferentschik:
Merged via pull request #980

PR #980, opened 2017-06-01
Issue #975 Refactoring MockRoundTripper

hferentschik:


pkg/testing/http_mock.go
```diff
@@ -40,61 +45,101 @@ var DefaultRoundTripper http.RoundTripper = &http.Transport{
...
+		case SERVE_STRING:
+			response.Body = ioutil.NopCloser(strings.NewReader(cannedResponse.Response))
+		case SERVE_FILE:
+			file, err := os.Open(cannedResponse.Response)
+			if err != nil {
+				panic(err)
+			}
+			response.Body = ioutil.NopCloser(bufio.NewReader(file))
+		default:
+			panic("Unkown canned response type")
```
gbraad: Unknown

pkg/minishift/cache/oc_caching_test.go
```diff
@@ -24,11 +24,246 @@ import (
...
 	"os"
 	"path/filepath"
+	"runtime"
 	"testing"
 )
 
-var testDir string
-var testOc Oc
+var (
+	openShiftReleaseTagInfo = `{
```
budhrg: It would have been good if we extract this into separate json file and read it. 
I am not a big fan of embedding such a long testdata. 
Also, this would help in making this testdata as reusable (if in case).

pkg/minishift/update/update_test.go
```diff
@@ -110,3 +112,35 @@ func checkErr(t *testing.T, err error) {
 		t.Fatal(err)
 	}
 }
+
+func addMockedResponses(mockTransport *minitesting.MockRoundTripper) {
```
budhrg: `addMockResponses`. Hardly saw such names with 2nd/3rd form of the verb.

pkg/testing/http_mock.go
```diff
@@ -18,16 +18,21 @@ package testing
...
-	"path/filepath"
 	"regexp"
-	"runtime"
+	"strings"
 	"time"
 )
 
+const (
+	OCTET_STREAM = "application/octet-stream"
+	JSON         = "application/json; charset=utf-8"
```
budhrg: Can be rename `OCTET_STREAM`/`JSON` to `OCTET_STREAM_CONTENT_TYPE` / `JSON_CONTENT_TYPE`.

This feels too generic too me.

pkg/testing/http_mock.go
hferentschik: ok

pkg/minishift/cache/oc_caching_test.go
hferentschik: If someone else needs it, it is straight forward to extract. There are pro and cons for embedding. The moment another tests needs it lets extract it.

pkg/minishift/update/update_test.go
hferentschik: TBH, I see both names as equivalent. But if you prefer addMockResponses, so be it.

pkg/testing/http_mock.go
hferentschik: I find the names a bit too long then and I think they are telling enough.

hferentschik:
Pushed an update.

@budhrg I actually did pull out the tag json response :-)

pkg/minishift/cache/oc_caching_test.go
```diff
@@ -79,9 +86,27 @@ func setUp(t *testing.T) {
...
 }
 
-func EnsureGitHubApiAccessTokenSet(t *testing.T) {
-	if githubutils.GetGitHubApiToken() == "" {
-		t.Skip("Skipping GitHub API based test, because no access token is defined in the environment.\n " +
-			"To run this test check https://help.github.com/articles/creating-a-personal-access-token-for-the-command-line/ and set for example MINISHIFT_GITHUB_API_TOKEN (see github.go).")
+func addMockResponses(mockTransport *minitesting.MockRoundTripper) {
+	testDataDir := filepath.Join(basepath, "..", "..", "..", "test", "testdata")
+
+	mockTransport.RegisterResponse("https://api.github.com/repos/openshift/origin/releases/tags/v1.3.1", &minitesting.CannedResponse{
```
praveenkumar: @hferentschik Isn't we used v1.5.1 somewhere in test during refactoring?

gbraad: I noticed the same, but in this case it is just pinning on a known value, and not really the functionality provided by the version specified. it is mock and nothing more.

hferentschik: Right. The whole point is to not go over the wire (or at least as little as possible). The testdata we are serving back in these tests (checksums, actual binaries, release info son) are pinned to when these tests got created. No need to update them. 

For what it's worths, I think you are referring to some refactoring in start_test.go. The change there is also cosmetic. These tests don't do any HTTP requests at all anymore.

pkg/testing/http_mock.go
budhrg: >  find the names a bit too long then and I think they are telling enough.

For me, they are telling generic meaning. I am fine with a little longer name over to navigate over to  the definition to understand what a constant is doing. Let's get the feedback of others too. 
cc @gbraad @LalatenduMohanty @praveenkumar .

test/testdata/openshift-1.3.1-release-tag.json
```diff
@@ -0,0 +1,229 @@
+{
```
budhrg: :+1:  for pulling it out to separate file. This is best practice :)

pkg/minishift/cache/oc_caching_test.go
praveenkumar: @hferentschik @gbraad Ack 👍

pkg/testing/http_mock.go
LalatenduMohanty: @budhrg `OCTET_STREAM/JSON` sounds ok  to me. It almost mean the same thing.

Issue #837, opened 2017-04-27
Integration test scenario for persistence volumes fails on CentOS CI


hferentschik:
This is actually following up on #774 which should have fixed the issue.

hferentschik:
Merged via pull request #981

PR #981, opened 2017-06-01
Issue #837 Re-enabling persitent volume scenario

hferentschik:
Addresses issue #837 

Re-enalbing with retry. At least we know now why it failed. It would be nicer to properly wait for the job to finish, but we cannot do that for now.

The alternative is to also enable anyuid as part of the test scenario. In this case Minishift itself will wait for the creation of the persistent volumes and the 'oc get pv' command would need to return successfully directly.

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -1,8 +1,31 @@
 @basic
 Feature: Basic
-  In order to use Minishift
-  As a user
-  I need to be able to bring up a test environment
+  As a user I can perform basic operations of Minishift and OpenShift
+
+  Scenario: User can install default add-ons
+    Given Minishift has state "Does Not Exist"
+    When executing "minishift addons install --defaults" succeeds
+    Then stdout should contain
+     """
+     Default add-ons anyuid, admin-user, xpaas installed
+     """
+
+  Scenario: User can enable the anyuid add-on
+    Given Minishift has state "Does Not Exist"
+    When executing "minishift addons enable anyuid" succeeds
+    Then stdout should contain
+     """
+     Addon 'anyuid' enabled
+     """
+
+  Scenario: User can list enabled plugins
+    Given Minishift has state "Does Not Exist"
+    When executing "minishift addons list" succeeds
+    Then stdout should contain
+     """
+     - anyuid         : enabled    P(0)
+     - admin-user     : disabled   P(0)
+     """
 
   Scenario: Starting Minishift
     Given Minishift has state "Does Not Exist"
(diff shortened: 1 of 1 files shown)


gbraad:
From the suggested PR:
```
# Persistent volumes are created by a OpenShift job 'persistent-volume-setup' which can take time to
# It would be nice to use 'oc observce' (or similar) to explicitly wait for the persitent volumnes to be created,
# but that's currently not possible.
```
Not clear what the problem is. Refer to an issue which describes the problem...

hferentschik:
@gbraad, let me look once more into it

hferentschik:
I actually went for the add-on approach. This way we get some other useful test scenarios.

hferentschik:
retest this please

hferentschik:
Seems to work now

test/integration/features/basic.feature
```diff
@@ -1,8 +1,31 @@
...
+     """
+
+  Scenario: User can list enabled plugins
+    Given Minishift has state "Does Not Exist"
+    When executing "minishift addons list" succeeds
+    Then stdout should contain
+     """
+     - anyuid         : enabled    P(0)
+     - admin-user     : disabled   P(0)
+     """
```
LalatenduMohanty: I think we need to test test with add-ons and without add-ons. If we are testing only with add-ons than we might miss issues that might exist without add-ons . WDYT?

praveenkumar: @LalatenduMohanty only issue which I can see without addon is time taken to have pv created which is right now not predicted. ( @hferentschik tried with try and sleep time also). Another use oc command same way to wait till pv job is done `oc get job persistent-volume-setup -n default -o 'jsonpath={ .status.active }'`

hferentschik: > I think we need to test with add-ons and without add-ons. If we are testing only with add-ons than we might miss issues that might exist without add-ons

Without a doubt, both needs to be tested. But right now we still are trying to test a single basic usage. Eventually we should start creating multiple feature files.

hferentschik: > Another use oc command same way to wait till pv job is done oc get job persistent-volume-setup -n default -o 'jsonpath={ .status.active }'

Sure, except this command does not wait. It just displays the status. So we would need to execute this in a loop. For that we need to extend the infrastructure.

> only issue which I can see without addon is time taken to have pv created which is right now not predicted. 

IMO there are several pros with this. First of all we add some rudimentary tests around add-ons and secondly we have a full deterministic scenario. Without it, we either need to use retry or we need to extend the framework even further to allow execution in a loop until a certain event occurs (I think eventually we will need that as well). Right now, enabling anyuid seems to me the best option.

LalatenduMohanty:
> Without a doubt, both needs to be tested. But right now we still are trying to test a single basic usage. Eventually we should start creating multiple feature files.

What about an issue to track  this i.e. we need to test without addons and we are good to merge this one .

hferentschik:
IMO there are plenty of issues which deal with extending the Cucumber test harness. If you think there is an issue needed go ahead and create one.

hferentschik:
Rebased and merged.

Issue #396, opened 2017-02-13
Write flag whether VM is registered into persistent config


hferentschik:
#At the moment when stopping/deleting the VM, we cannot warn/inform the user whether un-registration was required, but did not occur. The problem is that atm we need a running VM to determine whether the VM is registered. If the VM is in an error state, we don't know whether un-registration should or should not occur. This can potentially lead to dangling subscription which the user does not know about. 

If we instead during registration write a flag into persistent config, we can check this falg during stop/delete and inform/warn the user accordingly.

Follows up on issue #318

praveenkumar:
> Basically I thought we read the config file as in root.go (we already do as part of the Viper config) and keep the landed config in memory behind some Config interface. 

Yes, we can use same config file which used for --iso-url and --vm-driver.

> When we modify the config we flush. For this particulate case we would write a isRegisteredVM flag. Maybe we put this into a nested currentVM context. 

I think we modify config file using set/unset param currently and not completely flush it but what we should do is make Registered as false initially and then change according to provioner.

> We will also need the current oc to be used persistent in there. This is required for the oc env command as well as for the registry exposure (issue #254). WDYT?

I didn't get it how it is relevant to our registration side of changes?

hferentschik:
> I didn't get it how it is relevant to our registration side of changes?

What I meant is, that the ability to read/write custom config entries is required for these tasks as well. So whichever tasks gets implemented first needs to put something into place which the other task can use.

hferentschik:
Merged via pull request #988

PR #982, opened 2017-06-02
Issue #688 Introduce --skip-registration and --skip-unregistration flag

### LINKED ISSUE #688: Provide --skip-registration and --skip-unregistration flag for subscription based iso's
Its observed that the RHN register and unregister is done every time the minishift is started/stopped. It will be nice only to do it when we delete and start the cluster.  Something similar to the options we have with vagrant registration.

praveenkumar:
Fix #688


pkg/minishift/registration/utils.go
```diff
@@ -26,12 +26,17 @@ import (
...
+	Username           string
+	Password           string
+	SkipRegistration   bool
+	SkipUnRegistration bool
 }
 
 // Register host VM
 func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
+	if param.SkipRegistration {
+		return nil
```
budhrg: Let's log here about skip registration for logging purpose.

pkg/minishift/registration/utils.go
```diff
@@ -60,6 +65,9 @@ func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
 
 // Unregister host VM
 func UnregisterHostVM(host *host.Host, param *RegistrationParameters) error {
+	if param.SkipUnRegistration {
+		return nil
```
budhrg: Same here as above.

pkg/minishift/registration/utils.go
hferentschik: What type of logging are you talking about here? glog? Something which appears when I turn on debug logs? Then yes, a general println to the console no.

budhrg: Yes glog one when we turn on debug logs, I forgot to mention that above. thanks

praveenkumar: @budhrg @hferentschik Done.

pkg/minishift/registration/redhat.go
```diff
@@ -43,7 +43,7 @@ func (registrator *RedHatRegistrator) CompatibleWithDistribution(osReleaseInfo *
 	if osReleaseInfo.ID != "rhel" {
 		return false
 	}
-	if _, err := registrator.SSHCommand("sudo -E subscription-manager version"); err != nil {
+	if _, err := registrator.SSHCommand("sudo -E subscription-manager"); err != nil {
```
LalatenduMohanty: @praveenkumar Interesting!

pkg/minishift/registration/utils.go
```diff
@@ -26,12 +26,18 @@ import (
...
+	Username           string
+	Password           string
+	SkipRegistration   bool
+	SkipUnRegistration bool
 }
 
 // Register host VM
 func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
+	if param.SkipRegistration {
+		log.Debug("User choosed to skip registration")
```
LalatenduMohanty: I think we need to push this to the console i.e. minishift start should print that `Registration is skipped and user need to manually register the VM`.

pkg/minishift/registration/utils.go
```diff
@@ -60,6 +66,10 @@ func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
 
 // Unregister host VM
 func UnregisterHostVM(host *host.Host, param *RegistrationParameters) error {
+	if param.SkipUnRegistration {
+		log.Debug("User choosed to skip un-registration")
```
LalatenduMohanty: Same here.

LalatenduMohanty: The caveat here is that , if user deletes the VM i.e `minishift delete` the entitlement will not be freed. So we should mention this to user IMO.

pkg/minishift/registration/utils.go
praveenkumar: @LalatenduMohanty Why do you think this something part of console because to achieve that user need to use `--skip-registration` flag which imply he choosed to do it. If that was something implicit we were doing than it did make sense but right now not to me.

praveenkumar:
> The caveat here is that , if user deletes the VM i.e minishift delete the entitlement will not be freed. So we should mention this to user IMO.

This is all user choice and we are not forcing. I am not sure if that is part of codebase of something we document for users.

pkg/minishift/registration/utils.go
hferentschik: I second @praveenkumar. This is imo classic info which should be logged through the logger. In fact we should refrain as much as possible from logging straight to the console from within anything in _pkg_. Otherwise it will be hard to control. 

@praveenkumar I suggest to formulate this a bit different though. "Skipping registration due to enabled --skip-registration flag". (or something like this). BTW, "choosed" does not exist ;-)

LalatenduMohanty: @praveenkumar partly agree. I think I should have been more explicit. The part which we think implicit is user knows that without registration things will not work while deploying application OpenShift cluster (in Minishift). But I believe not all users know this, hence we need to communicate this to user.

LalatenduMohanty:
Code looks good to me. 

> This is all user choice and we are not forcing. I am not sure if that is part of codebase of something we document for users.

Sure. But being explicit about the side effects of this is a good idea IMO. 

The perfect thing will be if `minishift delete` knows if the VM was in a registered or unregistered state and then accordingly convey the message to the user.

pkg/minishift/registration/utils.go
hferentschik: At this stage we should only log. 

@LalatenduMohanty If we want to better the user feedback, it should be in stop.go. The code would beed to check whether the VM is running and registered. If so and is stop is called with skip-unregistration, we can print a message. 

For my part I would be fine to merge, just with some updated log messages.

praveenkumar:
@hferentschik Updated debug message.

hferentschik:
> The perfect thing will be if minishift delete knows if the VM was in a registered or unregistered state and then accordingly convey the message to the user.

@LalatenduMohanty we know that, there is an issue for that -  #396

hferentschik:
@praveenkumar If I do a stop with '--skip-unregistration' and then start again. Will it transparently skip registration (aka it detects the VM is already registered) or do I need to explicitly set '--skip-registration' as well?

praveenkumar:
> If I do a stop with '--skip-unregistration' and then start again. Will it transparently skip registration (aka it detects the VM is already registered) or do I need to explicitly set '--skip-registration' as well?

Right now user need to explicit set `--skip-registration` to make it transparent some code-refactor needed from regratrator side.

LalatenduMohanty:
@praveenkumar @hferentschik If we are not communicating that OpenShift cluster in Minishift will not work as expected if the VM is not registered, we are leaving a loophole for the user to discover. Wondering how we expect user to know this piece of information.

LalatenduMohanty:
> @LalatenduMohanty we know that, there is an issue for that - #396

+1

hferentschik:
> If we are not communicating that OpenShift cluster in Minishift will not work as expected

What's not working as expected. AFAIU OpenShift is full functional. So provided you are building your own workload you would be fine, right?

 > if the VM is not registered, we are leaving a loophole for the user to discover. 

IMO this is a documentation issue for CDK. We probably should open a CDK issue to track that this flag(s) get properly documented. Also remember, the registrar framework even though atm only used by us, is generic so one could for example plug in say a Suse VM or similar. What skipping registration means will depend on the context.

hferentschik:
> Right now user need to explicit set --skip-registration

that's what I thought. 

> to make it transparent some code-refactor needed from regratrator side.

Right. It would be nice if this would be transparent, but as said, for now I am fine as is.

hferentschik:
Created https://issues.jboss.org/browse/CDK-132

LalatenduMohanty:
> @LalatenduMohanty do you want to do anything else right now?

Nope. We are fine now.

pkg/minishift/registration/utils.go
budhrg: :+1:

Issue #964, opened 2017-05-30
Add scenario to deploy a 'realistic' workload: Cool Store


gbraad:
Implement a scenario to deploy a workload that contains a frontend, API gateway and some services, which represent a realistic workload.

gbraad:
```cucumber
@coolstore
Feature: Cool Store
  In order to test Minishift under load
  I need to setup a test environment
  Using CentOS as image and 4G of memory
  After which I will import image streams
  Finally I will deploy the Cool Store
  During this deployment I will measure memory and CPU usage


  Scenario: Starting Minishift
    Given Minishift has state "Does Not Exist"
     When executing "minishift start --iso-url https://github.com/minishift/minishift-centos-iso/releases/download/v1.0.0/minishift-centos7.iso --memory 4096" succeeds
     Then Minishift should have state "Running"

  Scenario: Import JBoss imagestreams
    Given Minishift has state "Running"
     When executing "oc --as system:admin create -n openshift -f https://raw.githubusercontent.com/jboss-openshift/application-templates/master/jboss-image-streams.json" succeeds
     Then stdout should contain
      """

      """
      
  Scenario: Import JBoss imagestreams
    Given Minishift has state "Running"
     When executing "oc --as system:admin create -n openshift -f https://raw.githubusercontent.com/jboss-fuse/application-templates/master/fis-image-streams.json" succeeds
     Then stdout should contain
      """

      """

  Scenario: User can create new namespace coolstore
   Given Minishift has state "Running"
    When executing "oc new-project coolstore" succeeds
    Then stdout should contain
     """
     Now using project "coolstore"
     """
  
  Scenario: User can deploy application to namespace coolstore 
   Given Minishift has state "Running"
    When executing "oc process -f https://raw.githubusercontent.com/jbossdemocentral/coolstore-microservice/1.0.x/openshift/coolstore-template.yaml | oc create -f -" succeeds
    Then stdout should contain
     """
     Success
     """

  Scenario: Successfully built the cart service 
   Given Minishift has state "Running" 
    When executing "oc bc/cart --follow" succeeds
    Then stdout should contain
     """
     Push successful
     """

  Scenario: Successfully deployed cart service to namespace coolstore 
   Given Minishift has state "Running" 
    When executing "oc rollout status deploymentconfig cart --watch" succeeds
    Then stdout should contain
     """
     "cart" successfully rolled out
     """

  Scenario: Successfully rolled out gateway to namespace coolstore 
[...]

PR #986, opened 2017-06-05
Issue #964 Add Coolstore deployment scenario

gbraad:
Addresses issue #964

 * Added goroutines to verify rollout as tested in #1113 
 * Scenario variables can be set from executing a command

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -135,7 +135,7 @@ function prepare() {
 function run_tests() {
   make clean test cross fmtcheck prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration
+  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags basic -format pretty"
   echo "CICO: Tests ran successfully"
 }
 
--- a/test/integration/features/coolstore.feature
+++ b/test/integration/features/coolstore.feature
@@ -1,0 +1,72 @@
+@coolstore
+Feature: Cool Store
+  In order to test Minishift under load
+  I need to setup a test environment
+  Using CentOS and 4G of memory
+  After which I will import image streams
+  Finally I will deploy the Cool Store
+  
+  Scenario: User enables the 'xpaas' add-on and creates a Minishift instance
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift addons enable xpaas"
+     Then stdout should contain
+      """
+      Add-on 'xpaas' enabled
+      """
+     When executing "minishift start --memory 4096 --iso-url centos" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: User creates a project and deploys CoolStore
+   Given Minishift has state "Running"
+    When executing "oc new-project coolstore" succeeds
+    Then stdout should contain
+     """
+     Now using project "coolstore"
+     """
+    When executing "oc create -f ./templates/coolstore.yaml" succeeds
+    Then stdout should contain
+     """
+     template "coolstore" created
+     """
+    When executing "oc new-app --template=coolstore" succeeds
+    Then stdout should contain
+     """
+     Success
+     """
+    When services "web-ui, inventory, catalog, cart, coolstore-gw" rollout successfully
+    Then executing "oc status -v" succeeds
+     And stdout should not contain
+     """
+     has failed
+     """
+
+  Scenario: User is able to interact with the deployed services from the API gateway
+   Given Minishift has state "Running"
+    When setting scenario variable "coolstore-gw" to the stdout from executing "oc get pods -o name -l application=coolstore-gw"
+    Then scenario variable "coolstore-gw" should not be empty
+    When executing "oc rsh $(coolstore-gw) curl -sSL http://catalog:8080/api/products" succeeds
+    Then stdout should not contain
+     """
+     threw exception
+     """
+     And stdout should contain
+     """
+     Fedora
+     """
+    When executing "oc rsh $(coolstore-gw) curl -sSL http://inventory:8080/api/availability/329299" succeeds
+    Then stdout should contain
+     """
+     Raleigh
+     """
+    When executing "oc rsh $(coolstore-gw) curl -sSL http://cart:8080/api/cart/FOO" succeeds
+    Then stdout should contain
+     """
+     cartTotal
+     """
+
+  Scenario: User stops and deletes the Minishift instance
+    Given Minishift has state "Running"
+     When executing "minishift stop" succeeds
+     Then Minishift should have state "Stopped"
+     When executing "minishift delete" succeeds
+     Then Minishift should have state "Does Not Exist"
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -110,34 +110,69 @@ func FeatureContext(s *godog.Suite) {
 
 	minishift = &Minishift{runner: runner}
 
-	// steps to execute minishift commands
-	s.Step(`Minishift (?:has|should have) state "([^"]*)"$`, minishift.shouldHaveState)
-	s.Step(`executing "minishift ([^"]*)"$`, minishift.executingMinishiftCommand)
-	s.Step(`executing "minishift ([^"]*)" (succeeds|fails)$`, executingMinishiftCommandSucceedsOrFails)
-	s.Step(`([^"]*) of command "minishift ([^"]*)" is equal to "([^"]*)"$`, commandReturnEquals)
-	s.Step(`([^"]*) of command "minishift ([^"]*)" contains "([^"]*)"$`, commandReturnContains)
+	// steps to execute `minishift` commands
+	s.Step(`Minishift (?:has|should have) state "([^"]*)"$`,
+		minishift.shouldHaveState)
+	s.Step(`^executing "minishift ([^"]*)"$`,
+		minishift.executingMinishiftCommand)
+	s.Step(`^executing "minishift ([^"]*)" (succeeds|fails)$`,
+		executingMinishiftCommandSucceedsOrFails)
+	s.Step(`([^"]*) of command "minishift ([^"]*)" is equal to "([^"]*)"$`,
+		commandReturnEquals)
+	s.Step(`([^"]*) of command "minishift ([^"]*)" contains "([^"]*)"$`,
+		commandReturnContains)
 
-	// steps for running oc
-	s.Step(`executing "oc ([^"]*)" retrying (\d+) times with wait period of (\d+) seconds$`, minishift.executingRetryingTimesWithWaitPeriodOfSeconds)
-	s.Step(`executing "oc ([^"]*)"$`, minishift.executingOcCommand)
-	s.Step(`executing "oc ([^"]*)" (succeeds|fails)$`, minishift.executingOcCommandSucceedsOrFails)
+	// steps to execute `oc` commands
+	s.Step(`^executing "oc ([^"]*)" retrying (\d+) times with wait period of (\d+) seconds$`,
+		minishift.executingRetryingTimesWithWaitPeriodOfSeconds)
+	s.Step(`^executing "oc ([^"]*)"$`,
+		minishift.executingOcCommand)
+	s.Step(`^executing "oc ([^"]*)" (succeeds|fails)$`,
+		executingOcCommandSucceedsOrFails)
 
-	// steps to verify stdout and stderr of commands executed
-	s.Step(`([^"]*) should contain "([^"]*)"$`, commandReturnShouldContain)
-	s.Step(`([^"]*) should contain$`, commandReturnShouldContainContent)
-	s.Step(`([^"]*) should equal "([^"]*)"$`, commandReturnShouldEqual)
-	s.Step(`([^"]*) should equal$`, commandReturnShouldEqualContent)
-	s.Step(`([^"]*) should be empty$`, commandReturnShouldBeEmpty)
-	s.Step(`([^"]*) should be valid ([^"]*)$`, shouldBeInValidFormat)
+	// steps for scenario variables
+	s.Step(`setting scenario variable "([^"]*)" to the stdout from executing "oc ([^"]*)"$`,
+		minishift.setVariableExecutingOcCommand)
+	s.Step(`^scenario variable "([^"]*)" should not be empty$`,
+		variableShouldNotBeEmpty)
+
+	// steps for rollout check
+	s.Step(`^services "([^"]*)" rollout successfully$`,
+		minishift.rolloutServicesSuccessfully)
+
+	// steps to verify `stdout`, `stderr` and `exitcode` of commands executed
+	s.Step(`([^"]*) should contain ([^"]*)$`,
+		commandReturnShouldContain)
+	s.Step(`([^"]*) should not contain "([^"]*)"$`,
+		commandReturnShouldNotContain)
+	s.Step(`([^"]*) should contain$`,
+		commandReturnShouldContainContent)
+	s.Step(`([^"]*) should not contain$`,
+		commandReturnShouldNotContainContent)
+	s.Step(`([^"]*) should equal "([^"]*)"$`,
+		commandReturnShouldEqual)
+	s.Step(`([^"]*) should equal$`,
+		commandReturnShouldEqualContent)
+	s.Step(`([^"]*) should be empty$`,
+		commandReturnShouldBeEmpty)
+	s.Step(`([^"]*) should not be empty$`,
+		commandReturnShouldNotBeEmpty)
+	s.Step(`([^"]*) should be valid ([^"]*)$`,
+		shouldBeInValidFormat)
 
 	// step for HTTP requests for minishift web console
-	s.Step(`(body|status code) of HTTP request to "([^"]*)" (?:|at "([^"]*)" )(contains|is equal to) "([^"]*)"$`, verifyHTTPResponse)
+	s.Step(`(body|status code) of HTTP request to "([^"]*)" (?:|at "([^"]*)" )(contains|is equal to) "([^"]*)"$`,
+		verifyHTTPResponse)
+
 	// step for HTTP requests for accessing application
-	s.Step(`(body|status code) of HTTP request to "([^"]*)" of service "([^"]*)" in namespace "([^"]*)" (contains|is equal to) "([^"]*)"$`, getRoutingUrlAndVerifyHTTPResponse)
+	s.Step(`(body|status code) of HTTP request to "([^"]*)" of service "([^"]*)" in namespace "([^"]*)" (contains|is equal to) "([^"]*)"$`,
+		getRoutingUrlAndVerifyHTTPResponse)
 
 	// steps for verifying config file content
-	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value "(.*)"$`, configContains)
-	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)"(.*)$`, configContains)
+	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value "(.*)"$`,
+		configContains)
+	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)"(.*)$`,
+		configContains)
 
 	s.BeforeSuite(func() {
 		testDir = setUp()
@@ -217,6 +252,14 @@ func compareExpectedWithActualContains(expected string, actual string) error {
 	return nil
 }
 
+func compareExpectedWithActualNotContains(notexpected string, actual string) error {
+	if strings.Contains(actual, notexpected) {
+		return fmt.Errorf("Output did match. Not expected: %s, Actual: %s", notexpected, actual)
+	}
+
+	return nil
+}
+
 func compareExpectedWithActualEquals(expected string, actual string) error {
 	if actual != expected {
 		return fmt.Errorf("Output did not match. Expected: %s, Actual: %s", expected, actual)
@@ -225,7 +268,20 @@ func compareExpectedWithActualEquals(expected string, actual string) error {
 	return nil
 }
 
+func compareExpectedWithActualNotEquals(notexpected string, actual string) error {
+	if actual == notexpected {
+		return fmt.Errorf("Output did match. Not expected: %s, Actual: %s", notexpected, actual)
+	}
+
+	return nil
+}
+
+func getLastCommandOutput() CommandOutput {
+	return commandOutputs[len(commandOutputs)-1]
+}
+
 func selectFieldFromLastOutput(commandField string) string {
+	lastCommandOutput := getLastCommandOutput()
 	outputField := ""
 	switch commandField {
 	case "stdout":
@@ -273,8 +329,16 @@ func commandReturnShouldContain(commandField string, expected string) error {
 	return compareExpectedWithActualContains(expected, selectFieldFromLastOutput(commandField))
 }
 
+func commandReturnShouldNotContain(commandField string, notexpected string) error {
+	return compareExpectedWithActualNotContains(notexpected, selectFieldFromLastOutput(commandField))
+}
+
 func commandReturnShouldContainContent(commandField string, expected *gherkin.DocString) error {
 	return compareExpectedWithActualContains(expected.Content, selectFieldFromLastOutput(commandField))
+}
+
+func commandReturnShouldNotContainContent(commandField string, notexpected *gherkin.DocString) error {
+	return compareExpectedWithActualNotContains(notexpected.Content, selectFieldFromLastOutput(commandField))
 }
 
 func commandReturnShouldEqual(commandField string, expected string) error {
@@ -289,18 +353,46 @@ func commandReturnShouldBeEmpty(commandField string) error {
 	return compareExpectedWithActualEquals("", selectFieldFromLastOutput(commandField))
 }
 
-func executingMinishiftCommandSucceedsOrFails(command, expectedResult string) error {
-	err := minishift.executingMinishiftCommand(command)
+func commandReturnShouldNotBeEmpty(commandField string) error {
+	return compareExpectedWithActualNotEquals("", selectFieldFromLastOutput(commandField))
+}
+
+func variableShouldNotBeEmpty(variableName string) error {
+	return compareExpectedWithActualNotEquals("", minishift.GetVariableByName(variableName).Value)
+}
+
+type commandRunner func(string) error
+
+func executingOcCommandSucceedsOrFails(command string, expectedResult string) error {
+	return succeedsOrFails(minishift.executingOcCommand, command, expectedResult)
+}
+
+func executingMinishiftCommandSucceedsOrFails(command string, expectedResult string) error {
+	return succeedsOrFails(minishift.executingMinishiftCommand, command, expectedResult)
+}
+
+func succeedsOrFails(execute commandRunner, command string, expectedResult string) error {
+	err := execute(command)
 	if err != nil {
 		return err
 	}
-	commandFailed := (lastCommandOutput.ExitCode != 0 || len(lastCommandOutput.StdErr) != 0)
+
+	lastCommandOutput := getLastCommandOutput()
+	commandFailed := (lastCommandOutput.ExitCode != 0 ||
+		len(lastCommandOutput.StdErr) != 0)
+
 	if expectedResult == "succeeds" && commandFailed == true {
-		return fmt.Errorf("Command did not execute successfully. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
+		return fmt.Errorf("Command '%s' did not execute successfully. cmdExit: %d, cmdErr: %s",
+			lastCommandOutput.Command,
+			lastCommandOutput.ExitCode,
+			lastCommandOutput.StdErr)
 	}
 	if expectedResult == "fails" && commandFailed == false {
-		return fmt.Errorf("Command executed successfully, however was expected to fail. cmdExit: %d, cmdErr: %s", lastCommandOutput.ExitCode, lastCommandOutput.StdErr)
+		return fmt.Errorf("Command executed successfully, however was expected to fail. cmdExit: %d, cmdErr: %s",
+			lastCommandOutput.ExitCode,
+			lastCommandOutput.StdErr)
 	}
+
 	return nil
 }
 
--- a/test/integration/minishift.go
+++ b/test/integration/minishift.go
@@ -23,10 +23,12 @@ import (
 	"fmt"
 	"github.com/minishift/minishift/test/integration/util"
 	"strings"
+	"sync"
 	"time"
 )
 
-var lastCommandOutput CommandOutput
+var commandOutputs []CommandOutput
+var commandVariables []CommandVariable
 
 type CommandOutput struct {
 	Command  string
@@ -35,7 +37,13 @@ type CommandOutput struct {
 	ExitCode int
 }
 
+type CommandVariable struct {
+	Name  string
+	Value string
+}
+
 type Minishift struct {
+	mutex  sync.Mutex
 	runner util.MinishiftRunner
 }
 
@@ -54,6 +62,7 @@ func (m *Minishift) executingRetryingTimesWithWaitPeriodOfSeconds(command string
 		if err != nil {
 			return err
 		}
+		lastCommandOutput := getLastCommandOutput()
 		if lastCommandOutput.ExitCode == 0 {
 			break
 		}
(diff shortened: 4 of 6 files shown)


test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,157 @@
+@coolstore
+Feature: Cool Store
+  In order to test Minishift under load
+  I need to setup a test environment
+  Using CentOS as image and 4G of memory
+  After which I will import image streams
+  Finally I will deploy the Cool Store
+  During this deployment I will measure memory and CPU usage
+
+
```
budhrg: @gbraad This seems to be extra line to me.

gbraad: very well spotted... doesn't it make the file look nicer? ;-)

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,157 @@
...
+   Given Minishift has state "Running" 
+    When executing "oc status -v" succeeds
+    Then stdout should not contain
+     """
+     has failed
+     """
+
+  Scenario: Able to interact with the catalog from the coolstore gateway 
+   Given Minishift has state "Running" 
+    When set variable "apigw" executing "oc get pods -o name -l application=coolstore-gw"
```
hferentschik: What will eventually the scope of this variable? Global and re-usable for other scenarios or similar to the stdout and stderr content which gets overwritten all the time?

test/integration/features/coolstore.feature
budhrg: At least for me, it is breaking the consistency and increasing the file length by one :) 
It looks nice with 1 line too :)

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,157 @@
...
+     And executing "oc rsh $(apigw) curl http://catalog:8080/api/products" succeeds
+    Then stdout should contain
+     """
+     Fedora
+     """
+
+  Scenario: Able to interact with the inventory from the coolstore gateway 
+   Given Minishift has state "Running"
+    When set variable "apigw" executing "oc get pods -o name -l application=coolstore-gw"
```
hferentschik: You seem to be repeating the setting of the variable. Why not make this global and keep the set variables in a context? Once defined the variable can then be used in any step execution which supports variables.

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,157 @@
...
+     And executing "oc rsh $(apigw) curl http://inventory:8080/api/availability/329299" succeeds
+    Then stdout should contain
+     """
+     Raleigh
+     """
+
+  Scenario: Able to interact with the cart from the coolstore gateway 
+   Given Minishift has state "Running"
+    When set variable "apigw" executing "oc get pods -o name -l application=coolstore-gw"
+     And executing "oc rsh $(apigw) curl http://cart:8080/api/cart/FOO" succeeds
```
hferentschik: Why not test from the outside via the route? We have some step for this already.

hferentschik:
@gbraad For what I can tell, all deployments are kicked off at once, sending OpenShift into overdrive. My guess is that the template is defined in a way that it kicks off the initial deployments when the app is created. I guess the plan is actually to do the builds in sequence, right?

hferentschik:
For what it's worth, in my case the test actually timed out and got killed:

    *** Test killed with quit: ran too long (1h1m0s).

I think it might have eventually completed, since most of the services where at the time deployed. Only two pods were missing. Biggest issues affect was that all builds occurred at the same time, resulting in a lot of trashing and very little progress

agajdosi:
Mine results of `make integration GODOG_OPTS="-tags coolstore -format pretty"`:
```
  Scenario: Successfully built the inventory service        # features/coolstore.feature:64
    Given Minishift has state "Running"                     # integration_test.go:83 -> *Minishift.github.com/minishift/minishift/test/integration.shouldHaveState
    When executing "oc logs bc/inventory --follow" succeeds # integration_test.go:93 -> *Minishift.github.com/minishift/minishift/test/integration.executingOcCommandSucceedsOrFails
    Command 'logs bc/inventory --follow' did not execute successfully. cmdExit: 1, cmdErr: error: read tcp 192.168.42.1:42450->192.168.42.189:8443: read: connection timed out

    Then stdout should contain
      """
      Push successful
      """

  Scenario: Successfully built the catalog service        # features/coolstore.feature:72
    Given Minishift has state "Running"                   # integration_test.go:83 -> *Minishift.github.com/minishift/minishift/test/integration.shouldHaveState
    Minishift state did not match. Expected: Running, Actual: Paused
    When executing "oc logs bc/catalog --follow" succeeds
    Then stdout should contain
      """
      Push successful
      """
[...]
```
And all other scenarios fails with machine status: "Paused".

test/integration/features/coolstore.feature
gbraad: Not all routes should be publicly routed, as they are internal services. Only the API gateway should have access.

test/integration/features/coolstore.feature
gbraad: This is a know issue (copy paste); just wanted to see the result from CI first, as the paused state is puzzling at the moment. I believe it has to do with resource allocation, but wonder why manually this is NOT an issue. But yes, this is what is intended to be done; the variables should be refused in successive steps.

test/integration/features/coolstore.feature
gbraad: LOL. Note taken

test/integration/features/coolstore.feature
gbraad: What would you suggest. I would see it similarly to a shell variable, which is therefore global for the execution of that shell instance and all calls inherit. So global... And allow to be overwritten when a new set in initiated. This is not accounted for at this moment as I append...

gbraad:
The idea is to also allow command output and variables to be logged at the end of execution; you can image an output for free memory, or time of start/end, etc.

gbraad:
I see that the deployment in CI could actually pass. Just need to rephrase the test a little and fix an issue in the variables replace

gbraad:
```
14:37:45 using binary: /home/minishift_ci/gopath/bin/minishift
14:42:06 ...................................................................... 70
15:03:22 ...............................................................Killed by signal 15.
15:03:22 + rtn_code=124
15:03:22 + cico node done d6743a28
```

gbraad:
Added ` GODOG_OPTS="-tags coolstore"` to centos_ci.sh to only run coolstore for this PR(will be removed later)

gbraad:
purposely setting up 'set variable'-step to fail, as it seems it does not capture the output.

gbraad:
Current version is intended to fail... (need debug information)

gbraad:
got bitten by the same issue with regex and order. Have prefixed all executing steps as `^executing "oc`, etc.

hferentschik:
> got bitten by the same issue with regex and order. Have prefixed all executing steps as ^executing "oc, etc.

doh. Stupid "feature"

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,157 @@
...
+  During this deployment I intend to measure memory and CPU usage
+
+  Scenario: Starting Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start --memory 4096 --cpus 2 --iso-url https://github.com/minishift/minishift-centos-iso/releases/download/v1.0.0/minishift-centos7.iso" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: Import JBoss imagestreams
+    Given Minishift has state "Running"
+     When executing "oc --as system:admin create -n openshift -f https://raw.githubusercontent.com/jboss-openshift/application-templates/master/jboss-image-streams.json" succeeds
```
hferentschik: We probably would be better of the store the JSON file under source control instead of getting it over the web. At least a stable/tagged version needs to be used. Otherwise there might be a lot of debug fun coming up in case the content changes.

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,157 @@
...
+     Then stdout should contain
+      """
+      imagestream "redhat-openjdk18-openshift" created
+      """
+      
+  Scenario: Import JBoss imagestreams
+    Given Minishift has state "Running"
+     When executing "oc --as system:admin create -n openshift -f https://raw.githubusercontent.com/jboss-fuse/application-templates/master/fis-image-streams.json" succeeds
```
hferentschik: Same as above. I guess regarding the jboss image streams, we can skip importing them like this altogether and let the test install and enable the xpass add-on. The basic test has an example how this can be done

test/integration/features/coolstore.feature
gbraad: I agree, but the addons will solve the imagestreams issue. For templates we will have to do this

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,157 @@
...
+   Given Minishift has state "Running"
+    When executing "oc new-project coolstore" succeeds
+    Then stdout should contain
+     """
+     Now using project "coolstore"
+     """
+  
+  Scenario: User can create a template for coolstore 
+   Given Minishift has state "Running"
+    When executing "oc create -f https://raw.githubusercontent.com/jbossdemocentral/coolstore-microservice/1.0.x/openshift/coolstore-template.yaml" succeeds
```
hferentschik: I think we should have this a checked in test file

test/integration/features/coolstore.feature
gbraad: will be changed... ditto ^^

test/integration/features/coolstore.feature
gbraad: ditto ^^

test/integration/minishift.go
```diff
@@ -63,30 +83,79 @@ func (m *Minishift) executingRetryingTimesWithWaitPeriodOfSeconds(command string
 	return nil
 }
 
-func (m *Minishift) executingOcCommand(command string) error {
+// Variables (duplication)
+func (m *Minishift) setVariableExecutingOcCommand(name string, command string) error {
```
gbraad: will refactor to using `OcCommand` as a func()

test/integration/minishift.go
```diff
@@ -95,16 +164,20 @@ func (m *Minishift) executingOcCommandSucceedsOrFails(command, expectedResult st
...
 
-func (m *Minishift) executingMinishiftCommand(command string) error {
-	// TODO: there must be smarter way to destruct
-	cmdOut, cmdErr, cmdExit := m.runner.RunCommand(command)
-	lastCommandOutput = CommandOutput{
-		command,
-		cmdOut,
-		cmdErr,
-		cmdExit,
+func executingMinishiftCommandSucceedsOrFails(command, expectedResult string) error {
```
gbraad: will refactor to a general `SucceedOrFails(command function` by passing a func()

test/integration/minishift.go
```diff
@@ -35,10 +36,28 @@ type CommandOutput struct {
...
+type CommandVariable struct {
+	Name  string
+	Value string
+}
+
 type Minishift struct {
 	runner util.MinishiftRunner
 }
 
+func (m *Minishift) gatherInformation() error {
```
gbraad: to be removed...

budhrg:
retest this please

budhrg:
retest this please

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,161 @@
...
+
+  Scenario: Import JBoss imagestreams
+    Given Minishift has state "Running"
+     When executing "oc --as system:admin create -n openshift -f https://raw.githubusercontent.com/jboss-openshift/application-templates/master/jboss-image-streams.json" succeeds
+     Then stdout should contain
+      """
+      imagestream "redhat-openjdk18-openshift" created
+      """
+      
+  Scenario: Import JBoss imagestreams
```
budhrg: Can we rename this scenario or be more precise about JBoss here, the previous scenario seems to be same too. 

This scenario seems to import two imagestreams: 
```
imagestream "fis-java-openshift" created
imagestream "fis-karaf-openshift" created
```

hferentschik:
What's the status of this?

gbraad:
Currently it will run three tests, one for the basic integration, then to verify the variables functionality, and finally one for CoolStore. The variables feature will be removed if the functionality runs successfully.

CoolStore already deployed today several times but had issues with the 'cart' service not building properly. This is something that is also reported in their issue tracker: https://github.com/jbossdemocentral/coolstore-microservice/issues/49 (however, I had successful runs)

gbraad:
@agajdosi @budhrg The question is now how to enable this... this is likely not something we want to run on every checkin, but rather on nightlies? If so, this would be needed for the CentOS CI...

gbraad:
@budhrg Experiencing build timeout. Especially now relevant as we ALSO build a documentation image and generate the documentation with this. There is less time for the actual builds and tests...

gbraad:
Build was successful in previous run: https://ci.centos.org/job/minishift-pr/1393/

gbraad:
retest this please

gbraad:
In latest build: from 2:06 till 2:19 is related to compiling, unittest, and documentation (13mins)... which leaves too little time to run an integration test with coolstore. This needs to run as an extended test (as mentioned before, nightlies?), but allow to run for at least 1h.

agajdosi:
@gbraad I do not have any experience with Centos CI, but if it is possible to run nightly tests there, then +1 on adding cool store feature into nightlys. I am also interested in starting this test on downstream nightly builds. For PR checks I think there is no need to run Coolstore feature.

From practical point of view I think we could make `@basic` tag preselected when running `make integration` - `flag.StringVar(&godogTags, "tags", "basic", "Tags for godog test")`. When Coolstore feature is needed to run then one will do `make integration GODOG_OPTS="-tags coolstore"` which will run just Coolstore.

gbraad:
Totally agree. Currently it runs with tag @coolstore as a forced option,
but this is only for testing purposes. I believe budhram knows about
setting up CentOS CI...

-- 

   [ Doing Open Source Matters ]

gbraad:
retest this please

agajdosi:
trying this locally now

agajdosi:
Seems like it went fine for me on Fedora 24.
```
[agajdosi@dhcp minishift]$ make integration
[...]
51 scenarios (51 passed)
188 steps (188 passed)
30m59.982777035s
PASS
ok      github.com/minishift/minishift/test/integration 1859.998s
```

gbraad:
It really is a timeout issue... it just runs for a long time.
I think we need to enable it on a long running build/test anyways...

so, any comments on code? if not, I will re-enable basic, and want to merge this ASAP

hferentschik:
So this is otherwise ready to merge? I was waiting for the [WIP] status to be removed before doing another review. 

What's about the occurring deployment failures? This is out of our control? So the aim is to just have this as a test one can run on demand? It won't be enabled as part of the regular CI builds? Do we want to have a nightly (or on any other schedule) build which runs these tests?

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,137 @@
...
+   Given Minishift has state "Running" 
+    When executing "oc status -v" succeeds
+    Then stdout should not contain
+     """
+     has failed
+     """
+
+  Scenario: Should be able to get the pod identifier for the API gateway 
+   Given Minishift has state "Running" 
+    When set variable "coolstore-gw" executing "oc get pods -o name -l application=coolstore-gw"
```
hferentschik: I still having my issues with this step. Reading this as a whole sentence it feels something is missing ("succeeds"?) Or maybe:

    Given output of executing "oc get pods -o name -l application=coolstore-gw" is stored as environment variable "coolstore-gw"

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,137 @@
...
+    Then stdout should not be empty
+     And exitcode should equal "0"
+
+  Scenario: Able to interact with the catalog from the API gateway 
+   Given Minishift has state "Running"
+    When executing "oc rsh $(coolstore-gw) curl -sSL http://catalog:8080/api/products" succeeds
```
hferentschik: We probably need to document this variable approach somewhere. Which steps allows for interpolation? What is the scope of the variables?

hferentschik: So why the need to test the access internally? Why not hit the URL via the route? For that we have the infrastructure, right?

test/integration/features/variables.feature
```diff
@@ -0,0 +1,60 @@
...
+    Given Minishift has state "Running"
+     When executing "minishift stop" succeeds
+     Then Minishift should have state "Stopped"
+
+  Scenario: Deleting Minishift
+    Given Minishift has state "Stopped"
+     When executing "minishift delete" succeeds
+     Then Minishift should have state "Does Not Exist"
+
+  # Scenario: Gathering all information
```
hferentschik: Is this some left over?

test/integration/integration_test.go
```diff
@@ -112,22 +112,29 @@ func FeatureContext(s *godog.Suite) {
 
 	// steps to execute minishift commands
 	s.Step(`Minishift (?:has|should have) state "([^"]*)"$`, minishift.shouldHaveState)
-	s.Step(`executing "minishift ([^"]*)"$`, minishift.executingMinishiftCommand)
-	s.Step(`executing "minishift ([^"]*)" (succeeds|fails)$`, executingMinishiftCommandSucceedsOrFails)
+	// WIP: moving from minishift. to local functions?!?
```
hferentschik: What's the WIP part here?

test/integration/integration_test.go
```diff
@@ -289,18 +329,37 @@ func commandReturnShouldBeEmpty(commandField string) error {
 	return compareExpectedWithActualEquals("", selectFieldFromLastOutput(commandField))
 }
 
-func executingMinishiftCommandSucceedsOrFails(command, expectedResult string) error {
-	err := minishift.executingMinishiftCommand(command)
+func commandReturnShouldNotBeEmpty(commandField string) error {
```
hferentschik: 👍 I like how you split this into smaller pieces

test/integration/minishift.go
```diff
@@ -63,48 +70,84 @@ func (m *Minishift) executingRetryingTimesWithWaitPeriodOfSeconds(command string
 	return nil
 }
 
-func (m *Minishift) executingOcCommand(command string) error {
-	ocRunner := m.runner.GetOcRunner()
-	if ocRunner == nil {
-		return errors.New("Minishift is not Running")
+func getVariableByName(name string) *CommandVariable {
```
hferentschik: Should this be part of the Minishift struct.

test/integration/features/variables.feature
gbraad: i did not intent to keep the variables.feature, but if it is useful i can remove these comments. wdyt?

test/integration/features/coolstore.feature
gbraad: internal services, like cart, catalog, etc are supposed to be only exposed by the api gateway and not publicly by a route.

test/integration/integration_test.go
gbraad: missed...

test/integration/features/coolstore.feature
gbraad: suggestion is ok. just think the intent is more on the variable... but maybe the execution results in a value to store is clearer

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,137 @@
...
+   Given Minishift has state "Running"
+    When executing "oc new-project coolstore" succeeds
+    Then stdout should contain
+     """
+     Now using project "coolstore"
+     """
+  
+  Scenario: User can create a template for coolstore 
+   Given Minishift has state "Running"
+    When executing "oc create -f $(template_dir)/coolstore.yaml" succeeds
```
gbraad: it is defined in integration.go on start of the whole suite

gbraad:
[WIP] was removed from the commit... missed the PR title... :-s

test/integration/features/coolstore.feature
gbraad: https://github.com/minishift/minishift/pull/986/files#diff-a2af18ef513203ce3e4c765b60252e1cR159

LalatenduMohanty:
@gbraad There are some unwanted white spaces in the PR, you might want to fix those.

gbraad:
Can you be more specific about the where the unwanted whitespaces are?

LalatenduMohanty:
I can see the white spaces in `git show` command (in the bash shell) .  But does not show up in github diff. I am not sure if I have done anything specific to my bash environment for this. Let me know if you can see them in the `git show` command for your commit.

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,137 @@
...
+      """
+      Addon 'xpaas' enabled
+      """
+
+  Scenario: Starting Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start --memory 4096 --iso-url centos" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: User can create new namespace coolstore
```
LalatenduMohanty: 👍

LalatenduMohanty:
When I tested the PR locally I saw one failure, but the CentOS CI reported 3 failures. @gbraad do we know why CentOS CI failing?

```
$ make integration
go install -pkgdir=/Users/lmohanty/code/gocode/src/github.com/minishift/minishift/out/bindata -ldflags="-X github.com/minishift/minishift/pkg/version.minishiftVersion=1.2.0 -X github.com/minishift/minishift/pkg/version.b2dIsoVersion=v1.0.2 -X github.com/minishift/minishift/pkg/version.centOsIsoVersion=v1.0.0 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v1.5.1 -X github.com/minishift/minishift/pkg/version.commitSha=bd54c809" ./cmd/minishift
mkdir -p /Users/lmohanty/code/gocode/src/github.com/minishift/minishift/out/integration-test
go test -timeout 3600s github.com/minishift/minishift/test/integration --tags=integration -v -args --test-dir /Users/lmohanty/code/gocode/src/github.com/minishift/minishift/out/integration-test --binary /Users/lmohanty/code/gocode/bin/minishift ""
Running Integration test in: /Users/lmohanty/code/gocode/src/github.com/minishift/minishift/out/integration-test
Using binary: /Users/lmohanty/code/gocode/bin/minishift
Using template dir: /Users/lmohanty/code/gocode/src/github.com/minishift/minishift/test/integration/templates
...................................................................... 70
..........................................F........................... 140
................................................                       188


--- Failed steps:

    Then  stdout should contain # features/coolstore.feature:13
      Error: Output did not match. Expected: Addon 'xpaas' enabled, Actual: Add-on 'xpaas' enabled



51 scenarios (50 passed, 1 failed)
188 steps (187 passed, 1 failed)
43m4.35818322s
PASS
exit status 1
FAIL	github.com/minishift/minishift/test/integration	2584.391s
make: *** [integration] Error 1
```

test/integration/features/coolstore.feature
hferentschik: Sure, I kind of expected this to be defined somewhere, but it comes a bit out of the blue. 

@agajdosi and I had a discussions around using variables with the go-dog maintainer. Using variables the way we do now in this pull request was already once on the table when @agajdosi was making some changes -https://github.com/DATA-DOG/godog/issues/81

From a Cucumber perspective it is not recommended to do it. So far we did not introduce it and @agajdosi solved his problem in another way. Personally I also feel that variable interpolation (or what you want to call it now) is useful and eventually needed. There is a difference so between the explicit 'set variable' step and this "magic" $(template_dir). Why not specify a relative path?

test/integration/features/coolstore.feature
hferentschik: > but maybe the execution results in a value to store is clearer

I think so.  I am literally unclear abut the meaning/intend of "set variable "coolstore-gw" executing ..." (in terms of grammar and semantics)

test/integration/features/coolstore.feature
hferentschik: ok.

test/integration/features/variables.feature
hferentschik: Let's remove this

gbraad:
@LalatenduMohanty the whitespaces have been inherited from the basic.feature due to copy-paste. They have the same issue... but I am only fixing it now for this file.

test/integration/features/coolstore.feature
gbraad: Take note: this is NOT an environment variable. nowhere in the code do i use os.setenv or similar. These variables only live inside the feature execution.

gbraad: @agajdosi uses environment variables as:
```
// steps for environmental variables
s.Step(`^user sets environmental variable "([^"]*)" to "([^"]*)"$`, setEnvVariable)
s.Step(`^user unsets environmental variable "([^"]*)"$`, unsetEnvVariable)
```

which in a similar fashions set the a record by name with the result from another command. This is also my intention here... I think for consistency it is best to use the pattern [variable] [from], but perhaps dropping the 'user' in the phrasing

gbraad:
https://ci.centos.org/job/minishift-pr/1445/

gbraad:
Will now change the standard scenario to be basic again

centos_ci.sh
```diff
@@ -135,7 +135,7 @@ function prepare() {
 function run_tests() {
   make clean test cross fmtcheck prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration
+  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags basic -format pretty"
```
hferentschik: +1, I was thinking about this myself at some stage.

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,136 @@
...
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift addons enable xpaas"
+     Then stdout should contain
+      """
+      Add-on 'xpaas' enabled
+      """
+
+  Scenario: Starting Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start --memory 4096 --iso-url centos" succeeds
```
hferentschik: Nice with the alias here now :-)

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,136 @@
...
+   Given Minishift has state "Running"
+    When executing "oc new-project coolstore" succeeds
+    Then stdout should contain
+     """
+     Now using project "coolstore"
+     """
+  
+  Scenario: User can create a template for coolstore
+   Given Minishift has state "Running"
+    When executing "oc create -f ./templates/coolstore.yaml" succeeds
```
hferentschik: 👍 , I think this is better, thanks.

test/integration/integration_test.go
gbraad: I aim for re-use and short functions... but often it takes some time to reach this, and especially naming of functions is very important for this.

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,136 @@
...
+
+  Scenario: User can deploy the coolstore template to namespace 'coolstore'
+   Given Minishift has state "Running"
+    When executing "oc new-app --template=coolstore" succeeds
+    Then stdout should contain
+     """
+     Success
+     """
+
+  Scenario: Successfully deployed web-ui service to namespace 'coolstore'
```
hferentschik: On a side note, I think the test is a bit misleading. It somehow suggest that services are build in order and that you wait for each and single one. In reality though all builds are started and you don't know whether you have to wait at a given point. In the worst case, web-ui rolls out last, in which case all the following 'rollout watches become basically no-ops'. Correct me from wrong here.

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,136 @@
...
+     """
+
+  Scenario: User can deploy the coolstore template to namespace 'coolstore'
+   Given Minishift has state "Running"
+    When executing "oc new-app --template=coolstore" succeeds
+    Then stdout should contain
+     """
+     Success
+     """
+
```
hferentschik: Instead of this rather lengthy and imo ambigious rollout waiting, one could just add a new step:

    And services web-ui, catalog, cart rollout successfully

This way we are just saying that we expect the listed services to rollout. It avoids the ambiguity of implied order. Also it makes the test more concise. 

Internally one can still use the rollout command to watch the services. One can even do it "properly", reflecting what's really happening and go each watch in a go routine, waiting for each watch to complete/timeout/fail before moving on.

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,136 @@
...
+   Given Minishift has state "Running"
+    When executing "oc status -v" succeeds
+    Then stdout should not contain
+     """
+     has failed
+     """
+
+  Scenario: Should be able to get the pod identifier for the API gateway
+   Given Minishift has state "Running"
+    When sets scenario variable "coolstore-gw" to the result from executing "oc get pods -o name -l application=coolstore-gw"
```
hferentschik: Should it not be "setting"? and potentially a "succeeds" in the end?

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,136 @@
...
+    Then stdout should not be empty
+     And exitcode should equal "0"
+
+  Scenario: Able to interact with the catalog from the API gateway
+   Given Minishift has state "Running"
+    When executing "oc rsh $(coolstore-gw) curl -sSL http://catalog:8080/api/products" succeeds
```
hferentschik: I actually like the interpolation. I don't think we get around it in the long run, even if it might not be  Cucumber best practice.

test/integration/integration_test.go
```diff
@@ -112,22 +112,28 @@ func FeatureContext(s *godog.Suite) {
 
 	// steps to execute minishift commands
 	s.Step(`Minishift (?:has|should have) state "([^"]*)"$`, minishift.shouldHaveState)
-	s.Step(`executing "minishift ([^"]*)"$`, minishift.executingMinishiftCommand)
-	s.Step(`executing "minishift ([^"]*)" (succeeds|fails)$`, executingMinishiftCommandSucceedsOrFails)
+	s.Step(`^executing "minishift ([^"]*)"$`, minishift.executingMinishiftCommand)
```
hferentschik: LOL, +1. Reduce the risk of un-intendted matches.

test/integration/features/coolstore.feature
gbraad: Create a different branch and test PR for this. After this works, I can include it here...

test/integration/features/coolstore.feature
gbraad: ditto ^^

gbraad:
Scenario works, but still once in a while deployment of the cart service fails: https://ci.centos.org/job/minishift-pr/1490/console

hferentschik:
> Scenario works

I got it to pass for the first time now as well :-)

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,96 @@
...
+
+  Scenario: User can deploy the coolstore template to namespace 'coolstore' 
+   Given Minishift has state "Running"
+    When executing "oc new-app --template=coolstore" succeeds
+    Then stdout should contain
+     """
+     Success
+     """
+
+  Scenario: Successfully deployed services to namespace coolstore
```
hferentschik: I don't think this is a different scenario, since the deployment of the services is triggered by the _oc new-app_ above. I think it more like this:

    Scenario: User can deploy the coolstore app to namespace 'coolstore' 
       When executing "oc new-app --template=coolstore" succeeds
       Then stdout should contain
        """
        Success
        """
        And services "web-ui, inventory, catalog, cart, coolstore-gw" rollout successfully

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,96 @@
...
+    When services "web-ui, inventory, catalog, cart, coolstore-gw" rollout successfully
+     And executing "oc status -v" succeeds
+    Then stdout should not contain
+     """
+     has failed
+     """
+
+  Scenario: Should be able to get the pod identifier for the API gateway 
+   Given Minishift has state "Running"
+    When setting scenario variable "coolstore-gw" to the stdout from executing "oc get pods -o name -l application=coolstore-gw" when succeeds
```
hferentschik: i think here went something wrong:

"When setting scenario variable "coolstore-gw" to the stdout from executing "oc get pods -o name -l application=coolstore-gw" when succeeds"

What's about the 'when' before succeeds? Also 'of' instead of 'from':

    "When setting scenario variable "coolstore-gw" to the stdout of executing "oc get pods -o name -l application=coolstore-gw"

gbraad:
I created an issue for the failing cart service: https://github.com/jbossdemocentral/coolstore-microservice/issues/74 and will follow up on this with more info.

test/integration/minishift.go
```diff
@@ -117,3 +170,47 @@ func (m *Minishift) getRoute(serviceName, nameSpace string) string {
 	cmdOut, _, _ := m.runner.RunCommand("openshift service " + serviceName + " -n" + nameSpace + " --url")
 	return strings.TrimRight(cmdOut, "\n")
 }
+
+func (m *Minishift) checkServiceRolloutForSuccess(service string, done chan bool) {
+	command := fmt.Sprintf("rollout status deploymentconfig %s --watch", service)
+
+	ocRunner := m.runner.GetOcRunner()
+	cmdOut, cmdErr, cmdExit := ocRunner.RunCommand(command)
+	commandOutputs = append(commandOutputs,
```
hferentschik: I don't think append is thread safe this way. I this case you probably would have to use a mutex as well.

test/integration/features/coolstore.feature
gbraad: > Should it not be "setting"? and potentially a "succeeds" in the end?

This is what I had, but this was wrong? Actually I think the succeeds is unnecessary... linguistic it should be 'from' as it indicates the location or state it comes from, while 'of' would indicate the actual value, as it is a fixed given.

test/integration/features/coolstore.feature
gbraad:
```cucumber
  Scenario: User can deploy the coolstore template to namespace 'coolstore' 
    When executing "oc new-app --template=coolstore" succeeds
    Then stdout should contain
     """
     Success
     """
     And services "web-ui, inventory, catalog, cart, coolstore-gw" rollout successfully
     And executing "oc status -v" succeeds
     And stdout should not contain
     """
     has failed
     """
```

hferentschik: You can have another When

    When executing "oc status -v" succeeds
    Then stdout should not contain

But yes, in general yes

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,72 @@
...
+    When services "web-ui, inventory, catalog, cart, coolstore-gw" rollout successfully
+    Then executing "oc status -v" succeeds
+     And stdout should not contain
+     """
+     has failed
+     """
+
+  Scenario: User is able to interact with the deployed services from the API gateway
+   Given Minishift has state "Running"
+    When setting scenario variable "coolstore-gw" to the stdout from executing "oc get pods -o name -l application=coolstore-gw"
```
hferentschik: 👍

test/integration/minishift.go
```diff
@@ -117,3 +172,49 @@ func (m *Minishift) getRoute(serviceName, nameSpace string) string {
 	cmdOut, _, _ := m.runner.RunCommand("openshift service " + serviceName + " -n" + nameSpace + " --url")
 	return strings.TrimRight(cmdOut, "\n")
 }
+
+func (m *Minishift) checkServiceRolloutForSuccess(service string, done chan bool) {
+	command := fmt.Sprintf("rollout status deploymentconfig %s --watch", service)
+
+	ocRunner := m.runner.GetOcRunner()
+	cmdOut, cmdErr, cmdExit := ocRunner.RunCommand(command)
+	m.mutex.Lock()
```
hferentschik: :-) - one way of doing it

test/integration/features/coolstore.feature
```diff
@@ -0,0 +1,72 @@
+@coolstore
+Feature: Cool Store
```
hferentschik: I think the whole feature reads much better now. Things are better grouped imo

test/integration/minishift.go
```diff
@@ -23,10 +23,12 @@ import (
 	"fmt"
 	"github.com/minishift/minishift/test/integration/util"
 	"strings"
+	"sync"
 	"time"
 )
 
-var lastCommandOutput CommandOutput
+var commandOutputs []CommandOutput
```
hferentschik: IMO this should be part of the Minishift struct, since the way it is used from within the struct breaks imo encapsulation. But let's move on for now

hferentschik:
Rebased and merged. Awesome feature file!

test/integration/minishift.go
gbraad: I actually considered moving `commandVariables` out, and rename as `scenarioVariables`...

test/integration/minishift.go
```diff
@@ -117,3 +172,49 @@ func (m *Minishift) getRoute(serviceName, nameSpace string) string {
...
+
+	for i := 0; i < total; i++ {
+		success = success && <-done
+	}
+
+	if !success {
+		return fmt.Errorf("Not all successfully rolled out")
+	}
+	return nil
+}
```
LalatenduMohanty: @gbraad Interesting stuff!

PR #987, opened 2017-06-05
[WIP] Issue #639 Install default addons as part of start/update command

### LINKED ISSUE #639: Default add-ons bundled by Minishift should be installed automatically
At the moment a user has to explicitly install the default addons via `minishift addons install --defaults`. To be more user friendly we should install the bundled addons per default. This will require to keep some meta data somewhere in _$MINISHIFT_HOME_ which helps us to determine whether addons need to be installed. For example we need to distinguish the following cases:

- Fresh install (_$MINISHIFT_HOME_ does not exist)
- Existing _$MINISHIFT_HOME_ and user explicitly en- or disabled or in fact modified addons
- The Minishift binary got updated and there might be new versions of the addons

budhrg:
Fix #639 
@hferentschik , this PR works fine for Linux (tested locally). Might also work with macOS (untested).

This PR fails with Windows with usual error as we had seen `not supported by windows`.

Opened for review. Tests are yet to add, just want to make sure I am going in the right direction.

### DIFF
--- a/cmd/minishift/cmd/addon/addon.go
+++ b/cmd/minishift/cmd/addon/addon.go
@@ -17,6 +17,7 @@ limitations under the License.
 package addon
 
 import (
+	"github.com/minishift/minishift/out/bindata"
 	"github.com/spf13/cobra"
 )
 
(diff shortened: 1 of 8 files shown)


budhrg:
Closing PR as new implementation has been discussed and working on it now.

PR #988, opened 2017-06-05
Issue #396 Add registration flag to instance specific configuration

praveenkumar:


pkg/minishift/config/instanceconfig.go
```diff
@@ -27,6 +27,7 @@ var InstanceConfig *InstanceConfigType
 type InstanceConfigType struct {
 	FilePath string `json:"-"`
 	OcPath   string
+	Register bool
```
LalatenduMohanty: What about `IsRegistered` ?

pkg/minishift/registration/utils.go
```diff
@@ -60,6 +61,10 @@ func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
 		if err := registrator.Register(param); err != nil {
 			return err
 		}
+		minishiftConfig.InstanceConfig.Register = true
```
LalatenduMohanty: 👍

pkg/minikube/cluster/cluster.go
```diff
@@ -140,11 +144,23 @@ func DeleteHost(api libmachine.API) error {
 
 // Un-register host VM
 func unregister(host *host.Host) error {
+	if !minishiftConfig.InstanceConfig.Register {
+		return nil
```
LalatenduMohanty: Why not sending error string "Minishift VM is not registered"?

pkg/minikube/cluster/cluster.go
```diff
@@ -140,11 +144,23 @@ func DeleteHost(api libmachine.API) error {
...
+	}
 	if drivers.MachineInState(host.Driver, state.Running)() {
 		if err := registration.UnregisterHostVM(host, RegistrationParameters); err != nil {
 			return fmt.Errorf("Error unregistring the VM: %s", err)
 		}
 	}
+	if minishiftConfig.InstanceConfig.Register {
+		userConfirmation := util.ReadInputFromStdin("Current Minishift instance is registered with RHN, " +
+			"Still want to Delete/Stop VM (y/n)")
+		if strings.ToUpper(userConfirmation) == "Y" {
```
LalatenduMohanty: I think conformation from user should be in the runDelete() function . Example we should ask user confirmation when the VM is stopped + the VM is still in registered state and then user is trying to delete.

LalatenduMohanty: This is not right. As we are asking whether user wants to unregister after UnregisterHostVM(host, RegistrationParameters)

pkg/minikube/cluster/cluster.go
```diff
@@ -140,11 +144,23 @@ func DeleteHost(api libmachine.API) error {
...
+			return nil
+		} else {
+			return errors.New("Delete aborted by User")
```
LalatenduMohanty: This error user will get irrespective of stop or delete, isn't it? which is wrong IMO.

pkg/minishift/config/instanceconfig.go
praveenkumar: @LalatenduMohanty I think `Register` server purpose of the variable.

pkg/minikube/cluster/cluster.go
praveenkumar: @LalatenduMohanty Which error? This is when a system is already unregister why we want to send a error here?

pkg/minikube/cluster/cluster.go
praveenkumar: > This is not right. As we are asking whether user wants to unregister + delete/stop after UnregisterHostVM(host, RegistrationParameters)

@LalatenduMohanty I didn't understand what we are actually evaluating here below workflow.

1. start minishift VM 
2. stop minishift VM with --skip-unregistation flag (which will not reset the Register flag in the config)
3. delete minishift VM => which warn user about this.

pkg/minikube/cluster/cluster.go
LalatenduMohanty: @praveenkumar Fair enough.

cmd/minishift/cmd/delete.go
```diff
@@ -44,25 +44,31 @@ func runDelete(cmd *cobra.Command, args []string) {
 	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer api.Close()
 
+	removeConfigs()
```
hferentschik: _removeInstanceConfig(uration)_?

hferentschik: +1 though

hferentschik: I wonder though whether this should be done after _cluster.DeleteHost_?

cmd/minishift/cmd/stop.go
```diff
@@ -49,7 +49,6 @@ func runStop(cmd *cobra.Command, args []string) {
 	}
 
 	if err := cluster.StopHost(api); err != nil {
-		fmt.Println("Error stopping cluster: ", err)
```
hferentschik: +1

pkg/minikube/cluster/cluster.go
```diff
@@ -97,8 +99,10 @@ func StartHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 		return nil, fmt.Errorf("Error configuring authorization on host: %s", err)
 	}
 
-	if err := registration.RegisterHostVM(h, RegistrationParameters); err != nil {
-		return nil, fmt.Errorf("Error registering the VM: %s", err)
+	if !minishiftConfig.InstanceConfig.Register {
```
hferentschik: It is really unfortunate that this code reaches out to other parts of the code and relies on global variables being set and all. This breaks encapsulation. Really, the method should get passed all the information it needs and should be independent of minishiftConfig.InstanceConfig.

Can we maybe refactor this?

pkg/minikube/cluster/cluster.go
hferentschik: Same as above, it is unfortunate to create a dependency to minishiftConfig.InstanceConfig.Register.

pkg/minikube/cluster/cluster.go
hferentschik: Not sure what @LalatenduMohanty means, but user interaction should be in the command not here in _cluster.go_

pkg/minishift/config/instanceconfig.go
hferentschik: It depends whether you see this as an instruction/action for the system, aka 'minishift register this instance' or whether you describe a state, "this instance is registered".  I tend to think in the latter terms in this context.

pkg/minishift/registration/utils.go
hferentschik: We really should see whether we can avoid the global state aspect here.

pkg/minishift/config/instanceconfig.go
budhrg: Here, it seems to be state, so +1 for `IsRegistered`

cmd/minishift/cmd/delete.go
budhrg: > I wonder though whether this should be done after cluster.DeleteHost?

+1, if host couldn't deleted then we shouldn't delete those configs

hferentschik: BTW, I was expecting that here in _delete.go_ the user interaction occurs. That we check whether the VM is registered and potentially print a message to the user

pkg/minishift/registration/utils.go
```diff
@@ -60,6 +61,10 @@ func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
...
+		if err := minishiftConfig.InstanceConfig.Write(); err != nil {
+			return err
```
hferentschik: Hmm, I am wondering whether one should not just log a warning and continue. Is this really an abort situation? Not saying it is not, but I think we also should start asking ourselves once in a while whether each error leads to failure. It might be enough to log a warning or try to recover. Not sure.

cmd/minishift/cmd/delete.go
```diff
@@ -46,23 +48,41 @@ func runDelete(cmd *cobra.Command, args []string) {
 
 	util.ExitIfUndefined(api, constants.MachineName)
 
+	if err := cluster.UnRegister(api); err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error unregistering the VM: %s", err.Error()))
+	}
+
+	if minishiftConfig.InstanceConfig.IsRegister {
```
hferentschik: Does this not come to late now? You already made the unregister call. What if the user selects 'n' now?

cmd/minishift/cmd/delete.go
```diff
@@ -46,23 +48,41 @@ func runDelete(cmd *cobra.Command, args []string) {
...
+		userConfirmation := pkgUtil.ReadInputFromStdin("Current Minishift instance is registered with Subscription, " +
```
hferentschik: Why 'Subscription' with capital 'S'?

hferentschik: Otherwise +1 for doing the user interaction in the command now.

cmd/minishift/cmd/stop.go
```diff
@@ -48,8 +48,11 @@ func runStop(cmd *cobra.Command, args []string) {
 		cluster.RegistrationParameters.SkipUnRegistration = true
 	}
 
+	if err := cluster.UnRegister(api); err != nil {
```
hferentschik: Is this not a bit confusing now? First you do a check for skip registration and set a flag and then you still call _cluster.UnRegister(api)_. Should we not just move the UnRegister call into the if statement. This makes cluster.RegistrationParameters.SkipUnRegistration obsolete.

pkg/minikube/cluster/cluster.go
```diff
@@ -97,8 +98,10 @@ func StartHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 		return nil, fmt.Errorf("Error configuring authorization on host: %s", err)
 	}
 
-	if err := registration.RegisterHostVM(h, RegistrationParameters); err != nil {
-		return nil, fmt.Errorf("Error registering the VM: %s", err)
+	if !minishiftConfig.InstanceConfig.IsRegister {
```
hferentschik: so here we still rely on this global state. We could add the required information to MachineConfig or  wer do the registration separately in a dedicated method.

pkg/minishift/registration/utils.go
```diff
@@ -86,6 +91,10 @@ func UnregisterHostVM(host *host.Host, param *RegistrationParameters) error {
 		if err := registrator.Unregister(param); err != nil {
 			return err
 		}
+		minishiftConfig.InstanceConfig.IsRegister = false
```
hferentschik: The method could return the bool which then gets written into persistent config.

pkg/minishift/registration/utils.go
```diff
@@ -60,6 +61,10 @@ func RegisterHostVM(host *host.Host, param *RegistrationParameters) error {
 		if err := registrator.Register(param); err != nil {
 			return err
 		}
+		minishiftConfig.InstanceConfig.IsRegister = true
+		if err := minishiftConfig.InstanceConfig.Write(); err != nil {
```
hferentschik: The method could return the bool which then gets written into persistent config.

hferentschik: The actual Register and UnregisterHost methods don't have to know about our decision to persist wether a VM got registered. They should just be concerned about doing the actual registration/unregistration.

cmd/minishift/cmd/delete.go
praveenkumar: @hferentschik if unregister already happen then this doesn't even run because `unregister` will set this flag to false. We are using this flag for a situation only (below) for any other situation we don't actually warn user every time about `vm is registered`

1. If a VM is stopped but not unregistered (using --skip-unregister)
2. If a VM crashed and user going to delete it.

cmd/minishift/cmd/stop.go
praveenkumar: @hferentschik Good catch, this need to be in that if block.

cmd/minishift/cmd/delete.go
```diff
@@ -46,23 +48,41 @@ func runDelete(cmd *cobra.Command, args []string) {
 
 	util.ExitIfUndefined(api, constants.MachineName)
 
+	if err := cluster.UnRegister(api, minishiftConfig.InstanceConfig); err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error unregistering the VM: %s", err.Error()))
+	}
+
+	if minishiftConfig.InstanceConfig.IsRegister {
```
hferentschik: I think the problem is that this still does not compute in my mind w/o looking at the code for _cluster.UnRegister_. This can only work, because things are happening to minishiftConfig.InstanceConfig as you pass it into UnRegister. 

Why do you even pass _minishiftConfig.InstanceConfig_? Why not deal with the fact that we persist registration in instance config in the command itself, making this all explicit?

hferentschik: Here is what I mean:

    func runDelete(cmd *cobra.Command, args []string) {
    	if minishiftConfig.InstanceConfig.IsRegistered {
    		err := cluster.UnRegister(api)
    		if err != nil {
    			handleFailedUnRegistration()
    		} 
    		instanceConfig.IsRegister = false
    		// not sure whether we need to handle failure, since we are going to delete the
    		// instance config altogther a little bit later on!?
     		instanceConfig.Write()
    	}
    }
    
    func handleFailedUnRegistration() {
    		userConfirmation := pkgUtil.ReadInputFromStdin("Current Minishift instance is registered     with subscription, " +
    		"Still want to Delete VM (y/n)")
    		if strings.ToUpper(userConfirmation) != "Y" {
    			atexit.ExitWithMessage(0, fmt.Sprintln("Delete aborted by User"))
    		}	
    }
    
Might need some tuning, but I think something like this makes the intend and what's actually happening much clearer.

pkg/minikube/cluster/cluster.go
```diff
@@ -128,21 +121,53 @@ func DeleteHost(api libmachine.API) error {
...
+		return nil
+	}
 	if drivers.MachineInState(host.Driver, state.Running)() {
-		if err := registration.UnregisterHostVM(host, RegistrationParameters); err != nil {
-			return fmt.Errorf("Error unregistring the VM: %s", err)
+		unRegistered, err := registration.UnregisterHostVM(host, RegistrationParameters)
+		if err != nil {
+			return err
+		}
+		if unRegistered {
```
hferentschik: See https://github.com/minishift/minishift/pull/988/files#r120873570. I don't think  UnRegister needs to know anything off InstanceConfigType.

cmd/minishift/cmd/delete.go
```diff
@@ -46,23 +48,45 @@ func runDelete(cmd *cobra.Command, args []string) {
...
 
-	if err := minishiftConfig.InstanceConfig.Delete(); err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintln(fmt.Sprintf("Error deleting %s: ", minishiftConfig.InstanceConfig.FilePath), err))
+	removeInstanceConfigs()
+
+	fmt.Println("Minishift VM deleted.")
+}
+
+func handleFailedUnRegistration() {
+	userConfirmation := pkgUtil.ReadInputFromStdin("Current Minishift instance is registered with subscription, " +
```
hferentschik: I changed the message a bit. Have a look at https://github.com/hferentschik/minishift/commit/ba794b476691e2b1e079fb348a4706603b01d51d

I tried to create a pull request against your feature branch, but that was confusing, since I also rebased your pull request already.

praveenkumar:
> This lets me happily delete the VM. Instead it should tell me that the VM is still registered and whether I want to continue. I think the rub lies in UnRegister:

No, it's not happily delete VM. I did test that case many time with current code. but I think what you are referring about logic (I used `true` bool instead of `false` to build that logic) :)

```
$ ./minishift stop --skip-unregistration
Cluster stopped.
✔ ~/work/github/go_practice/bin 
11:45 $ ./minishift delete
Current Minishift instance is registered with subscription, Still want to Delete VM (y/n): n
Delete aborted by User
```

praveenkumar:
@hferentschik I updated PR as per suggestion.

hferentschik:
Merged, thanks.

PR #990, opened 2017-06-05
Issue #639 Install default addons as part of start/update command

### LINKED ISSUE #639: Default add-ons bundled by Minishift should be installed automatically

budhrg:
Fix #639 

@hferentschik , This PR is only missing tests, rest functionally working in Linux and **Windows**. You heard it correct :smile: 

Will write tests as well. Thanks for the suggestion (looks like better one) :+1: 

I have tested in the following way:
```
$ make
go install -pkgdir=/home/budhram/gowork/src/github.com/minishift/minishift/out/bindata -ldflags="-X github.com/minishift/minishift/pkg/version.version=1.0.0 -X github.com/minishift/minishift/pkg/version.isoVersion=v1.0.2 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v1.5.1" ./cmd/minishift

$ minishift version
minishift v1.0.0

$ minishift delete

$ rm -rf ~/.minishift

$ minishift start
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso'
...
# You might hit by issue https://github.com/minishift/minishift/issues/989 
...
-- Removing temporary directory ... OK
-- Checking container networking ... OK
-- Server Information ... 
   OpenShift server started.
       https://192.168.42.231:8443

   You are logged in as:
       User:     developer
       Password: developer

   To login as administrator:
       oc login -u system:admin

-- Installing default addons ... OK
 40.00 MiB / 40.00 MiB [==========================================================================================================================] 100.00% 0s


$ minishift update
Do you want to update from 1.0.0 to 1.0.1 now? [y/N]: y (doing Control + C to test --force)
Downloading https://github.com/minishift/minishift/releases/download/v1.0.1/minishift-1.0.1-linux-amd64.tgz
^C

$ minishift update --force
Downloading https://github.com/minishift/minishift/releases/download/v1.0.1/minishift-1.0.1-linux-amd64.tgz
 3.43 MiB / 3.43 MiB [============================================================================================================================] 100.00% 0s
 3.43 MiB / 3.43 MiB [============================================================================================================================] 100.00% 0s
 3.43 MiB / 3.43 MiB [============================================================================================================================] 100.00% 0s
Updated successfully to minishift v1.0.1.
Post update changes will be executed on the next call of any minishift command.

$ minishift version
Minishift version: 1.0.1

$ git diff Makefile
diff --git a/Makefile b/Makefile
index 248cce5..f4b00f9 100644
--- a/Makefile
+++ b/Makefile
@@ -13,7 +13,7 @@
 # limitations under the License.
 
-MINISHIFT_VERSION = 1.0.1
+MINISHIFT_VERSION = 1.0.0
 OPENSHIFT_VERSION = v1.5.1
 B2D_ISO_VERSION = v1.0.2
 CENTOS_ISO_VERSION = v1.0.0

#### Compile again to v1.0.0
$ make
go install -pkgdir=/home/budhram/gowork/src/github.com/minishift/minishift/out/bindata -ldflags="-X github.com/minishift/minishift/pkg/version.version=1.0.0 -X github.com/minishift/minishift/pkg/version.isoVersion=v1.0.2 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v1.5.1" ./cmd/minishift

#### Change the *.addon file here or delete the folder to test

#### Run any command
$ minishift version
Binary has been recently updated.
--- Updating default addons ...OK
minishift v1.0.0

```

### DIFF
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -25,15 +25,19 @@ import (
 
 	"path/filepath"
 
+	"encoding/json"
 	"github.com/docker/machine/libmachine/log"
 	"github.com/golang/glog"
 	"github.com/minishift/minishift/cmd/minishift/cmd/addon"
 	configCmd "github.com/minishift/minishift/cmd/minishift/cmd/config"
 	hostfolderCmd "github.com/minishift/minishift/cmd/minishift/cmd/hostfolder"
 	cmdOpenshift "github.com/minishift/minishift/cmd/minishift/cmd/openshift"
+	"github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
+	"github.com/minishift/minishift/pkg/util/filehelper"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/minishift/minishift/pkg/version"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
 	"github.com/spf13/viper"
@@ -68,7 +72,15 @@ var RootCmd = &cobra.Command{
 	Short: "Minishift is a tool for application development in local OpenShift clusters.",
 	Long:  `Minishift is a command-line tool that provisions and manages single-node OpenShift clusters optimized for development workflows.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		var err error
+		var (
+			err          error
+			isFreshStart bool
+		)
+
+		if !filehelper.Exists(constants.Minipath) || filehelper.IsEmptyDir(constants.Minipath) {
+			isFreshStart = true
+		}
+
 		for _, path := range dirs {
 			if err := os.MkdirAll(path, 0777); err != nil {
 				glog.Exitf("Error creating minishift directory: %s", err)
@@ -80,7 +92,6 @@ var RootCmd = &cobra.Command{
 		// Create all instances config
 		allInstanceConfigPath := filepath.Join(constants.Minipath, "config", "allinstances.json")
 		minishiftConfig.AllInstancesConfig, err = minishiftConfig.NewAllInstancesConfig(allInstanceConfigPath)
-
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating config for all instances: %s", err.Error()))
 		}
@@ -88,9 +99,25 @@ var RootCmd = &cobra.Command{
 		// Create MACHINE_NAME.json
 		instanceConfigPath := filepath.Join(constants.Minipath, "machines", constants.MachineName+".json")
 		minishiftConfig.InstanceConfig, err = minishiftConfig.NewInstanceConfig(instanceConfigPath)
-
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating config for VM: %s", err.Error()))
+		}
+
+		// Run the default addons on fresh minishift start
+		if isFreshStart {
+			fmt.Print("-- Installing default addons ... ")
+			if err := util.UnpackAddons(constants.MakeMiniPath("addons")); err != nil {
+				atexit.ExitWithMessage(1, fmt.Sprintf("Error installing default add-ons : %s", err))
+			}
+
+			fmt.Println("OK")
+		}
+
+		// Check marker file created by update command and perform post update execution steps
+		if filehelper.Exists(filepath.Join(constants.Minipath, constants.UpdateMarkerFileName)) {
+			if err := performPostUpdateExecution(filepath.Join(constants.Minipath, constants.UpdateMarkerFileName)); err != nil {
+				atexit.ExitWithMessage(1, fmt.Sprintf("Error in performing post update exeuction: %s", err))
+			}
 		}
 
 		shouldShowLibmachineLogs := viper.GetBool(showLibmachineLogs)
@@ -178,4 +205,32 @@ func ensureConfigFileExists(configPath string) {
 			glog.Exitf("Cannot encode config %s: %s", configPath, err)
 		}
 	}
+}
+
+// performPostUpdateExecution executes the post update actions like unpacking the default addons
+// if user chose to update addons during `minishift update` command.
+// It also remove the marker file created by update command to avoid repeating the post update execution process
+func performPostUpdateExecution(markerPath string) error {
+	var markerData UpdateMarker
+
+	file, err := ioutil.ReadFile(markerPath)
+	if err != nil {
+		return err
+	}
+
+	json.Unmarshal(file, &markerData)
+	if markerData.InstallAddon {
+		fmt.Println(fmt.Sprintf("Minishift was upgraded from v%s to v%s. Running post update actions.", markerData.PreviousVersion, version.GetVersion()))
+		fmt.Print("--- Updating default add-ons ... ")
+		util.UnpackAddons(constants.MakeMiniPath("add-ons"))
+		fmt.Println("OK")
+		fmt.Println(fmt.Sprintf("Default add-ons %s installed", strings.Join(util.DefaultAssets, ", ")))
+	}
+
+	// Delete the marker file once post update execution is done
+	if err := os.Remove(markerPath); err != nil {
+		return err
+	}
+
+	return nil
 }
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -388,7 +388,7 @@ func ensureNotRunning(client *libmachine.Client, machineName string) {
 func validateOpenshiftVersion() {
 	requestedVersion := viper.GetString(startFlags.OpenshiftVersion.Name)
 
-	valid, err := clusterup.ValidateOpenshiftMinVersion(requestedVersion, constants.MinOpenshiftSuportedVersion)
+	valid, err := clusterup.ValidateOpenshiftMinVersion(requestedVersion, constants.MinOpenshiftSupportedVersion)
 	if err != nil {
 		atexit.ExitWithMessage(1, err.Error())
 	}
@@ -396,7 +396,7 @@ func validateOpenshiftVersion() {
 	if !valid {
 		fmt.Printf("Minishift does not support Openshift version %s. "+
 			"You need to use a version >= %s\n", viper.GetString(startFlags.OpenshiftVersion.Name),
-			constants.MinOpenshiftSuportedVersion)
+			constants.MinOpenshiftSupportedVersion)
 		atexit.Exit(1)
 	}
 
--- a/cmd/minishift/cmd/update.go
+++ b/cmd/minishift/cmd/update.go
@@ -22,10 +22,28 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/update"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 
+	"encoding/json"
+	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/util"
+	"github.com/minishift/minishift/pkg/version"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
+	"os"
+	"path/filepath"
 	"strings"
+)
+
+type UpdateMarker struct {
+	InstallAddon    bool
+	PreviousVersion string
+}
+
+const (
+	addonForceFlag = "force"
+)
+
+var (
+	addonForce bool
 )
 
 var updateCmd = &cobra.Command{
@@ -34,6 +52,10 @@ var updateCmd = &cobra.Command{
 	Long:  `Checks for the latest version of Minishift, prompt the user and update the binary if user answers with 'y'.`,
 	Run:   runUpdate,
 }
+
+var (
+	confirm string
+)
 
 func runUpdate(cmd *cobra.Command, args []string) {
 	proxyConfig, err := util.NewProxyConfig(viper.GetString(httpProxy), viper.GetString(httpsProxy), "")
@@ -56,10 +78,10 @@ func runUpdate(cmd *cobra.Command, args []string) {
 	}
 
 	if update.IsNewerVersion(localVersion, latestVersion) {
-		fmt.Printf("A newer version of minishift is available.\nDo you want to update from %s to %s now? [y/N]: ", localVersion, latestVersion)
-
-		var confirm string
-		fmt.Scanln(&confirm)
+		if !addonForce {
+			fmt.Printf("A newer version of minishift is available.\nDo you want to update from %s to %s now? [y/N]: ", localVersion, latestVersion)
+			fmt.Scanln(&confirm)
+		}
 
 		if strings.ToLower(confirm) == "y" {
 			err := update.Update(latestVersion)
@@ -68,7 +90,22 @@ func runUpdate(cmd *cobra.Command, args []string) {
 			}
 
 			fmt.Printf("\nUpdated successfully to minishift v%s.\n", latestVersion)
+
+			markerData := UpdateMarker{InstallAddon: false, PreviousVersion: version.GetVersion()}
+
+			fmt.Print("\nDo you want to update default addons? [y/N]: ")
+			fmt.Scanln(&confirm)
+
+			if strings.ToLower(confirm) == "y" {
+				markerData.InstallAddon = true
+				fmt.Println("Default add-ons will be updated when executing the next minishift command.")
+			}
+			if err := createUpdateMarker(filepath.Join(constants.Minipath, constants.UpdateMarkerFileName), markerData); err != nil {
+				atexit.ExitWithMessage(1, "Failed to create update marker file.")
+			}
+
 		}
+
 	} else {
 		fmt.Printf("Nothing to update.\nAlready using latest version: %s.\n", latestVersion)
 	}
@@ -78,4 +115,21 @@ func init() {
 	RootCmd.AddCommand(updateCmd)
 	updateCmd.Flags().AddFlag(httpProxyFlag)
 	updateCmd.Flags().AddFlag(httpsProxyFlag)
+	updateCmd.Flags().BoolVar(&addonForce, addonForceFlag, false, "Force update the addons post binary update. Otherwise, prompt user to update addons.")
+}
+
+func createUpdateMarker(markerPath string, data UpdateMarker) error {
+	f, err := os.OpenFile(markerPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	defer f.Close()
+	if err != nil {
+		return err
+	}
+
+	b, err := json.Marshal(data)
+	if err != nil {
+		return err
+	}
+
+	f.Write(b)
+	return nil
 }
--- a/cmd/minishift/cmd/util/util.go
+++ b/cmd/minishift/cmd/util/util.go
@@ -23,6 +23,12 @@ import (
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/state"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
+
+	"github.com/minishift/minishift/out/bindata"
+)
+
+var (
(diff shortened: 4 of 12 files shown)


hferentschik:
retest this please

cmd/minishift/cmd/root.go
```diff
@@ -179,3 +199,34 @@ func ensureConfigFileExists(configPath string) {
 		}
 	}
 }
+
+func performPostUpdateExecution() error {
+	fmt.Println("Binary has been recently updated.")
```
hferentschik: we can formulate this a bit better I think. "Minishift was upgraded from xxx to yyy. Running post update actions" (we can write the previous version into the update file, then we can literally write this)

cmd/minishift/cmd/root.go
```diff
@@ -179,3 +199,34 @@ func ensureConfigFileExists(configPath string) {
...
+	if err != nil {
+		return err
+	}
+
+	var markerData UpdateMarker
+	var confirm = "y"
+
+	json.Unmarshal(file, &markerData)
+	if !markerData.Force {
+		fmt.Printf("Do you want to update the default addons? [y/N]: ")
```
hferentschik: Please don't use the 'f' versions (Print_f_) of methods when you are not interpolating anything.

cmd/minishift/cmd/root.go
```diff
@@ -179,3 +199,34 @@ func ensureConfigFileExists(configPath string) {
...
+		fmt.Scanln(&confirm)
+	}
+
+	if strings.ToLower(confirm) == "y" {
```
hferentschik: I am confused, didn't we say that the confirmation is in the update command? We want to keep root.go free from user interaction. Otherwise we break the ability to script Minishift commands.

I think we said that we ask for confirmation in the update command and store the answer into the update marker file.

cmd/minishift/cmd/start.go
```diff
@@ -160,6 +160,15 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	if !isRestart {
+		// Run the default addons if MINISHIFT_HOME dir does not exist or empty
```
hferentschik: Is this a left over? I don't anything should change in start.go

cmd/minishift/cmd/start.go
```diff
@@ -386,15 +395,15 @@ func ensureNotRunning(client *libmachine.Client, machineName string) {
 func validateOpenshiftVersion() {
 	requestedVersion := viper.GetString(startFlags.OpenshiftVersion.Name)
 
-	valid, err := clusterup.ValidateOpenshiftMinVersion(requestedVersion, constants.MinOpenshiftSuportedVersion)
+	valid, err := clusterup.ValidateOpenshiftMinVersion(requestedVersion, constants.MinOpenshiftSupportedVersion)
```
hferentschik: :-)

pkg/minikube/constants/constants.go
```diff
@@ -24,32 +24,22 @@ import (
 	"github.com/minishift/minishift/pkg/version"
 )
 
-// MachineName is the name to use for the VM.
-const MachineName = "minishift"
-
-// APIServerPort is the port that the API server should listen on.
-const APIServerPort = 8443
+const (
+	MachineName                  = "minishift"      // Name to use for the VM
```
hferentschik: +1 for the cleanup

pkg/util/filehelper/file.go
```diff
@@ -152,3 +152,20 @@ func CopyDir(src string, dst string) (err error) {
 
 	return
 }
+
+// IsDirEmpty checks whether directory is empty or not
```
hferentschik: What is the behaviour if the directory does not exist or is a file which is not a directory?

hferentschik:
retest this please

cmd/minishift/cmd/root.go
```diff
@@ -62,13 +64,26 @@ var viperWhiteList = []string{
...
 	Use:   "minishift",
 	Short: "Minishift is a tool for application development in local OpenShift clusters.",
 	Long:  `Minishift is a command-line tool that provisions and manages single-node OpenShift clusters optimized for development workflows.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		var err error
+
+		if filehelper.Exists(constants.Minipath) {
+			homeDirExists = true
+		}
+		if filehelper.IsDirEmpty(constants.Minipath) {
```
budhrg: No, package name is required here.

cmd/minishift/cmd/root.go
budhrg: ah got it thanks

cmd/minishift/cmd/start.go
budhrg: That is just comment right?

cmd/minishift/cmd/addon/addon.go
```diff
@@ -27,6 +28,10 @@ const (
 	noAddOnMessage  = "No add-on with the name %s is installed."
 )
 
+var (
```
budhrg: I thought this is common to addon and might be used at multiple places. But now I have moved it to `cmd/util/util.go` along with `UnpackAddons`

cmd/minishift/cmd/addon/addon.go
```diff
@@ -35,3 +40,11 @@ var AddonsCmd = &cobra.Command{
 		cmd.Help()
 	},
 }
+
+func UnpackAddons(addonsDir string) error {
```
budhrg: Done

cmd/minishift/cmd/root.go
budhrg: @hferentschik , I understood differently.  Will change that.

pkg/util/filehelper/file.go
budhrg: Need to update the `IsDirEmpty` to `IsEmpty` to accommodate both.

cmd/minishift/cmd/addon/addon.go
hferentschik: I think util is the better place. Otherwise I understand your intention with the move. Just asking ;-)

cmd/minishift/cmd/root.go
hferentschik: Might be a misunderstanding. I meant why Exist + IsEmptyDir? Hence my comment as well on the function as well. What is the outcome of IsEmptyDir is the path does not exist or is not a directory. I would assume it is _false_, unless you want to differentiate the case where the path does not exist with an error. Either way a call to IsEmptyDir should suffice here, right?

cmd/minishift/cmd/start.go
hferentschik: No, I mean the changes in start.go in general. We are doing the unpacking in root.go. We don't have to touch start.go (I think)

pkg/util/filehelper/file.go
hferentschik: I think IsEmptyDir is fine. IsEmpty is a bit ambiguous. It all comes down to what exact semantics you have in mind. Also you have to make a choice whether you treat a non existent path as _false_ (aka nope this is not a empty directory) or whether you want to treat this as error in a sense that this methods expects and existing path in which case it responds with _true_ or _false_, otherwise an error is thrown. Same for what do you do if this is actually a file and not a directory. Both approaches are valid. The other methods in this util file basically for simplicity try to avoid returning errors, but it comes down to the context in which you think we are going to use this function and whether or not it is important to differentiate.

cmd/minishift/cmd/start.go
budhrg: @hferentschik , It is needed to delay the installation of addons of `minishift start` specific. Otherwise, for any command, the installation of default addons will occur.

budhrg:
retest this please

budhrg:
@hferentschik , please review it as most of the suggestions are done and also add few unit tests.

cmd/minishift/cmd/root.go
```diff
@@ -69,6 +77,14 @@ var RootCmd = &cobra.Command{
 	Long:  `Minishift is a command-line tool that provisions and manages single-node OpenShift clusters optimized for development workflows.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		var err error
+
+		if filehelper.Exists(constants.Minipath) {
```
hferentschik: So this is obsolete, right

cmd/minishift/cmd/root.go
```diff
@@ -69,6 +77,14 @@ var RootCmd = &cobra.Command{
...
+			homeDirExists = true
+		}
+		if filehelper.IsEmpty(constants.Minipath) {
```
hferentschik: I think having _IsEmpty_ checking for empty directory as well as empty file content is quite unfortunate. I've not have came across an IO lib which does that. I think your initial IsEmptyDir was much better.

cmd/minishift/cmd/root.go
```diff
@@ -179,3 +200,31 @@ func ensureConfigFileExists(configPath string) {
 		}
 	}
 }
+
+// performPostUpdateExecution executes the post update actions like unpacking the default addons
+// if user chose to update addons during `minishift update` command.
+// It also remove the marker file created by update command to avoid repeating the post update execution process
+func performPostUpdateExecution(markerPath string) error {
```
hferentschik: +1

cmd/minishift/cmd/util/util.go
```diff
@@ -50,3 +63,30 @@ func ExitIfNotRunning(driver drivers.Driver, machineName string) {
...
+	for _, asset := range DefaultAssets {
+		if err := bindata.RestoreAssets(addonsDir, asset); err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
+func CreateUpdateMarker(markerPath string, data UpdateMarker) error {
```
hferentschik: Not sure whether I would put this method here, instead of _update.go_. Feels you just do it, because of the test.

cmd/minishift/cmd/util/util.go
```diff
@@ -50,3 +63,30 @@ func ExitIfNotRunning(driver drivers.Driver, machineName string) {
...
+	f, err := os.OpenFile(markerPath, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0644)
+	defer f.Close()
+	if err != nil {
+		return err
+	}
+
+	b, err := json.Marshal(data)
+	if err != nil {
+		fmt.Println("error:", err)
```
hferentschik: That seems wrong. Is this intended? You just log an error and continue. Should you not return the error?

cmd/minishift/cmd/start.go
hferentschik: > Otherwise, for any command, the installation of default addons will occur.

This is what we want, right? The idea is that you write the marker and then on execution of _any_ Minishift command we handle the post update commands. What if the user wants to "explore" any potential new add-ons via the 'minishift addons list' or maybe enable a new add-on prior to calling start. The whole idea of utilising the pre-run in root.go is to make sure that any outstanding changes we want to make as part of the update are executing as soon as possible. Does this make sense now?

hferentschik:
@budhrg AppVeyor failing is actually legit. Your TestMarkerFileRemovedPostUpdateExecution won't work on Windows. For simplicity reasons I'd vote for removing this in which case the UpdateMarkerFile method can move to where it belongs.

cmd/minishift/cmd/util/util.go
budhrg: Ya not intended, got by mistake probably. will fix it

budhrg:
>  For simplicity reasons I'd vote for removing this in which case the UpdateMarkerFile method can move to where it belongs.

Ack

cmd/minishift/cmd/root.go
budhrg: Yes after empty now

cmd/minishift/cmd/root.go
```diff
@@ -68,7 +72,15 @@ var RootCmd = &cobra.Command{
 	Short: "Minishift is a tool for application development in local OpenShift clusters.",
 	Long:  `Minishift is a command-line tool that provisions and manages single-node OpenShift clusters optimized for development workflows.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
-		var err error
+		var (
```
hferentschik: +1 :-)

cmd/minishift/cmd/root.go
```diff
@@ -68,7 +72,15 @@ var RootCmd = &cobra.Command{
...
+			err          error
+			isFreshStart bool
+		)
+
+		if filehelper.IsEmptyDir(constants.Minipath) {
```
hferentschik: :-)

cmd/minishift/cmd/root.go
```diff
@@ -80,19 +92,34 @@ var RootCmd = &cobra.Command{
...
+			fmt.Print("-- Installing default addons ... ")
+			if err := util.UnpackAddons(constants.MakeMiniPath("addons")); err != nil {
+				atexit.ExitWithMessage(1, fmt.Sprintf("Error installing default addons : %s", err))
+			}
+
+			fmt.Println("OK")
+		}
+
+		// Check marker file created by update command and perform post update execution steps
+		if filehelper.Exists(filepath.Join(constants.Minipath, constants.UpdateMarkerFileName)) {
```
hferentschik: +1

pkg/util/filehelper/file.go
```diff
@@ -152,3 +152,24 @@ func CopyDir(src string, dst string) (err error) {
...
+func IsEmptyDir(path string) bool {
+	if !Exists(path) {
+		return true
```
hferentschik: I don't think we can do this. If the path does not exist, it really should return _false_.

cmd/minishift/cmd/root.go
hferentschik: So in the light of https://github.com/minishift/minishift/pull/990/files#r121346230, this might need to be 

    if ! filehelper.Exists(constants.Minipath) || filehelper.IsEmptyDir(constants.Minipath)

pkg/util/filehelper/file.go
budhrg: @hferentschik , if path doesn't exists then we assume file is empty and hence true.

If we return _false_, the meaning would be "If there doesn't exist path for directory, return _IsEmptyDir_ as _false_ which means "Directory contains something".

budhrg:
> When executing minishift status I would have expected the update action to take place.

The binary you used here is latest which doesn't have code to perform post update execution. Please compile the binary again with 1.1.0 version and perform `minishift status` again.

hferentschik:
> The binary you used here is latest which doesn't have code to perform post update execution. Please compile the binary again with 1.1.0 version and perform minishift status again.

Ahh, bugger, got you. That's a trap indeed ;-)

cmd/minishift/cmd/root.go
budhrg: @hferentschik , are you saying we should extract out the logic of checking file existance from `IsEmptyDir` ?

hferentschik:
@budhrg, so yes, it actually works. Obviously the binary I "upgraded" to did not have the required changes. Good stuff.

cmd/minishift/cmd/root.go
```diff
@@ -179,3 +206,31 @@ func ensureConfigFileExists(configPath string) {
...
+	var markerData UpdateMarker
+
+	file, err := ioutil.ReadFile(markerPath)
+	if err != nil {
+		return err
+	}
+
+	json.Unmarshal(file, &markerData)
+	if markerData.InstallAddon {
+		fmt.Println(fmt.Sprintf("Minishift was upgraded from v%s to v%s. Running post update actions", markerData.PreviousVersion, version.GetVersion()))
```
hferentschik: Full stop after "Running post update actions"

cmd/minishift/cmd/root.go
```diff
@@ -80,19 +92,34 @@ var RootCmd = &cobra.Command{
...
-
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating config for VM: %s", err.Error()))
 		}
 
+		// Run the default addons on fresh minishift start
+		if isFreshStart {
+			fmt.Print("-- Installing default addons ... ")
+			if err := util.UnpackAddons(constants.MakeMiniPath("addons")); err != nil {
+				atexit.ExitWithMessage(1, fmt.Sprintf("Error installing default addons : %s", err))
```
hferentschik: "default add-ons"

cmd/minishift/cmd/root.go
```diff
@@ -179,3 +206,31 @@ func ensureConfigFileExists(configPath string) {
...
+
+	file, err := ioutil.ReadFile(markerPath)
+	if err != nil {
+		return err
+	}
+
+	json.Unmarshal(file, &markerData)
+	if markerData.InstallAddon {
+		fmt.Println(fmt.Sprintf("Minishift was upgraded from v%s to v%s. Running post update actions", markerData.PreviousVersion, version.GetVersion()))
+		fmt.Print("--- Updating default addons ... ")
```
hferentschik: add-ons

cmd/minishift/cmd/update.go
```diff
@@ -68,7 +90,22 @@ func runUpdate(cmd *cobra.Command, args []string) {
...
 			fmt.Printf("\nUpdated successfully to minishift v%s.\n", latestVersion)
+
+			markerData := UpdateMarker{InstallAddon: false, PreviousVersion: version.GetVersion()}
+
+			fmt.Print("\nDo you want to update default addons? [y/N]: ")
+			fmt.Scanln(&confirm)
+
+			if strings.ToLower(confirm) == "y" {
+				markerData.InstallAddon = true
+				fmt.Println("Default addons will be updated on the call of next minishift command.")
```
hferentschik: "Default add-ons will be updated when executing the next minishift command."

cmd/minishift/cmd/update.go
```diff
@@ -68,7 +90,22 @@ func runUpdate(cmd *cobra.Command, args []string) {
 			}
 
 			fmt.Printf("\nUpdated successfully to minishift v%s.\n", latestVersion)
+
+			markerData := UpdateMarker{InstallAddon: false, PreviousVersion: version.GetVersion()}
+
+			fmt.Print("\nDo you want to update default addons? [y/N]: ")
```
hferentschik: add-ons

pkg/util/filehelper/file.go
hferentschik: The questions is "Is this an empty directory" which is really to questions, is this a directory and if so is it empty. A non existent path cannot be an empty directory.

cmd/minishift/cmd/root.go
hferentschik: No, just return false

hferentschik: In IsEmptyDir, _false_ should be returned in case the path does not exist. You just need another check here

budhrg:
@hferentschik , I have pushed the wordings changes and updated _IsEmptyDir_ as per our discussion.

cmd/minishift/cmd/root.go
```diff
@@ -80,19 +92,34 @@ var RootCmd = &cobra.Command{
...
 		instanceConfigPath := filepath.Join(constants.Minipath, "machines", constants.MachineName+".json")
 		minishiftConfig.InstanceConfig, err = minishiftConfig.NewInstanceConfig(instanceConfigPath)
-
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating config for VM: %s", err.Error()))
 		}
 
+		// Run the default addons on fresh minishift start
+		if isFreshStart {
+			fmt.Print("-- Installing default addons ... ")
```
hferentschik: add-ons

hferentschik:
Merged, fixing the "add-on" typo.

PR #998, opened 2017-06-06
[WIP] Issue #1000 Adds the launcher-template addon

### LINKED ISSUE #1000: Add launch.openshift.io template to minishift addon list
As discussed in the devtools-program mailing list, it would be nice to have the Openshift template used in launch.openshift.io as part of the Minishift/CDK catalog

gastaldi:
Fix #1000 
This PR adds the launch.openshift.io template to the minishift catalog

### DIFF
--- a/addons/launcher-template/launcher-template.addon
+++ b/addons/launcher-template/launcher-template.addon
@@ -1,0 +1,4 @@
+# Name: launcher-template
+# Description: Imports the launch.openshift.io template
+
+oc create -f launchpad-template.yaml -n openshift
--- a/addons/launcher-template/launchpad-template.yaml
+++ b/addons/launcher-template/launchpad-template.yaml
@@ -1,0 +1,442 @@
+kind: Template
+apiVersion: v1
+metadata:
+  name: launchpad
+  annotations:
+    openshift.io/display-name: "Launchpad (Frontend + Backend + Mission Control)"
+    description: This application consists of the combination of Mission Control and the Backend and Frontend
+    tags: instant-app,launchpad
+    iconClass: icon-shadowman
+    template.openshift.io/long-description: This application consists of the combination of Mission Control, Backend and Frontend
+    template.openshift.io/provider-display-name: Red Hat, Inc.
+    template.openshift.io/documentation-url: https://github.com/openshiftio
+    template.openshift.io/support-url: https://access.redhat.com
+message: "The following service(s) have been created in your project: mission-control, launchpad-backend, launchpad-frontend.\n\nFor more information about using this template, including OpenShift considerations, see https://github.com/openshiftio"
+labels:
+  template: launchpad
+objects:
+- apiVersion: v1
+  kind: Service
+  metadata:
+    labels:
+      app: launchpad-missioncontrol
+    name: launchpad-missioncontrol
+  spec:
+    ports:
+    - name: 8080-tcp
+      port: 8080
+      protocol: TCP
+      targetPort: 8080
+    selector:
+      app: launchpad-missioncontrol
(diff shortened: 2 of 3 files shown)


gbraad:
CI fails because of the following:
```
--- FAIL: Test_create_addon_manager (0.00s)
	addon_manager_test.go:70: Unexpected number of addons. Expected 4, got 5
=== RUN   Test_installing_addon_for_non_existing_directory_returns_an_error
--- PASS: Test_installing_addon_for_non_existing_directory_returns_an_error (0.00s)
=== RUN   Test_invalid_addons_get_skipped
--- PASS: Test_invalid_addons_get_skipped (0.03s)
FAIL
FAIL	github.com/minishift/minishift/pkg/minishift/addon/manager	0.111s
```

you can easily fix this by incrementing a number. ;-). But it also shows that the test is kinda faulty, as it assumes actual code to also be the testdata.

addons/launcher-template/launcher-template.addon
```diff
@@ -0,0 +1,4 @@
+# Name: launcher-template
+# Description: Imports the launch.openshift.io template
+
+oc create -f https://raw.githubusercontent.com/openshiftio/launchpad-templates/master/openshift/launchpad-template.yaml -n openshift
```
LalatenduMohanty: @gastaldi  I somehow not comfortable with pulling the template for the runtime from https://github.com/openshiftio/launchpad-templates/tree/master/openshift . I would be ok for with pulling it from internet if it is pinned to a particular version (which is not possible with add-ons). The issue is how to make sure `https://raw.githubusercontent.com/openshiftio/launchpad-templates/master/openshift/launchpad-template.yaml` is a valid and correct URL always. Also how to make sure  launchpad-template.yaml always stays sane.

hferentschik:
add to whitelist

addons/launcher-template/launcher-template.addon
```diff
@@ -0,0 +1,4 @@
+# Name: launcher-template
```
LalatenduMohanty: @gastaldi is launcher is the official name?

gastaldi: The official name is launch.openshift.io, but I couldn't use it as the addon name, so I adopted as "launcher" instead

gastaldi: Probably the template needs to be updated to remove any reference to `launchpad`. That's why I was pointing to the URL before

addons/launcher-template/launcher-template.addon
```diff
@@ -0,0 +1,4 @@
...
+# Description: Imports the launch.openshift.io template
+
```
praveenkumar: @gastaldi Also in the description please add from which git repo we have those templates. It would be great if you also include tag/release so when a issue  come up regarding template we can easily track it.

LalatenduMohanty:
Marking the PR as work in progress as @gastaldi suggested to wait for some more time as they are working on finalizing the name of the project.

addons/launcher-template/launcher-template.addon
jorgemoralespou: And try to version using the same versioning schema as origin. My recommendation is to use branches and not tags, as branches can be updated. Then you can tag a specific item if required. [Here's](https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples) an example repository doing that

hferentschik:
@gastaldi What's the state of this pull request? We have in the mean time created a dedicated repo for collecting community provided add-ons. See https://github.com/minishift/minishift-addons/issues/3.

It would be nice, if you could help us seeding this repo :-)

I will be closing the pull request and I am looking forward to see a new one in the new repo ;-)

Issue #919, opened 2017-05-18
Document SSH usage with Minishift


gbraad:
It is possible to perform SSH-based interaction with the VM instance, for this we provide a subcommand.In our documentation it should reflect the following:

   * SSH is not a requirement (not on Windows, Linux and/or Mac,
      * but this means that the interaction using `minishift ssh [command]` is non-interactive. This is standard behaviour from the native client
      * not possible to use `minishift ssh` (see previous comment)
   * if SSH is installed
      * using `minishift ssh` will open an interactive shell using this external client
      * using `minishift ssh [command]` will run a single command using the external client
 

Do we need to describe the possible install methods?

   * Windows:  
      `choco install openssh`  (which is the powershell-openssh-win32 distribution)
   * Linux; distro provided
   * macOS: distro provided

Do we need to describe the requirements for `sshfs` ?

Ref: #911, #912

gbraad:
https://github.com/PowerShell/Win32-OpenSSH/issues/738

gbraad:
Due to the reported issue with OpenSSH-Win32 it is probably better for the moment to try:

```powershell
PS> choco install git -params "/GitAndUnixToolsOnPath"
```

gbraad:
@thatdocslady We know why the 'suggested' `choco install openssh` does not work, and we know it will be eventually fixed. BUT, for the moment it is a known-issue... how to record it as such?

thatdocslady:
@gbraad I guess it depends on what "eventually" means :) The safest place we can document it is probably in the troubleshooting guide + link from the relevant usage doc to that known issue. Then when it's fixed we can remove the content from the troubleshooting guide.

gbraad:
I am trying to get a release date, but this is at the moment not clear.
Even if a new binary is available, it is not automatically synced to the
Chocolatey repository.

I just want to be sure this is a recorded known issue for upstream AND
downstream.

thatdocslady:
@gbraad for upstream we can definitely get it into the troubleshooting guide, and IIRC the downstream writers are pulling in that doc into the CDK doc set so it should make it there :)

gbraad:
let's do that!

thatdocslady:
@gbraad awesome, so I'll create SSH content for the "normal" use and add the known issue in the troubleshooting guide. Will ping with questions as needed :)

hferentschik:
[...]

PR #1001, opened 2017-06-07
[WIP] Issue #919 Document SSH usage

thatdocslady:
Fixes issue #919 

This PR covers usage scenario for the `minishift ssh` command, and includes some notes about using external SSH clients.

Note: Following a discussion with @gbraad it was deemed better to document which Windows client is supported rather than go into the technical details of why OpenSSH client is not supported (as this is out of our scope and we don't know when a fix will be implemented).

### DIFF
--- a/docs/source/using/managing-minishift.adoc
+++ b/docs/source/using/managing-minishift.adoc
@@ -275,3 +275,29 @@ Use the `openshift-version` option to request a specific version of OpenShift. Y
 
 The Minishift VM is exposed to the host system with a host-only IP address that
 can be obtained with the `minishift ip` command.
+
+[[connecting-with-ssh]]
+== Connecting to the Minishift VM with SSH
+
+You can use the xref:../command-ref/minishift_ssh{outfilesuffix}[`minishift ssh`] command to interact with the Minishift VM.
+
+When you run this command, it checks whether you have an external SSH client installed on your machine. An SSH client is not a
+prerequisite, but it impacts the behavior of the command.
+
+If you have an external SSH client::
+You can run `minishift ssh` without a sub-command to open an interactive shell using your external SSH client.
+Then, you can run commands on the Minishift VM in the same way that you run commands interactively on any remote machine using SSH.
+
+You can also run the command with a sub-command. The native Minishift client then sends the command to the Minishift VM
+and returns the result to your local shell.
+
+Make sure that the path to the SSH client is specified in the _PATH_ environment variable.
+
+NOTE: You can use most OpenSSH-based clients. However, if you use the Windows operating system, only *cygwin* is supported.
+
+If you do not have an external SSH client::
+You must specify a sub-command when you run `minishift ssh`. The native Minishift client then sends the command to the Minishift VM
+and returns the result to your local shell.
+
+This behavior is considered non-interactive and you must run the command with a sub-command
+each time you want to perform an operation on the Minishift VM.


gbraad:
the issue of support on Windows is deal with by https://github.com/PowerShell/Win32-OpenSSH/issues/666 and the release: https://github.com/PowerShell/Win32-OpenSSH/releases/tag/v0.0.15.0 Will need to be verified and find it's way into chocolatey. Until then, better to describe what we recommend.

docs/source/using/managing-minishift.adoc
```diff
@@ -275,3 +275,29 @@ Use the `openshift-version` option to request a specific version of OpenShift. Y
...
+[[connecting-with-ssh]]
+== Connecting to the Minishift VM with SSH
+
+You can use the xref:../command-ref/minishift_ssh{outfilesuffix}[`minishift ssh`] command to interact with the Minishift VM.
+
+When you run this command, it checks whether you have an external SSH client installed on your machine. An SSH client is not a
+prerequisite, but it impacts the behavior of the command.
+
+If you have an external SSH client::
+You can run `minishift ssh` without a sub-command to open an interactive shell using your external SSH client.
```
hferentschik: An example would be nice to clarify this, aka `minishift ssh`

hferentschik: "You can run `minishift ssh` without a sub-command to open an interactive shell using using the external SSH client."

docs/source/using/managing-minishift.adoc
```diff
@@ -275,3 +275,29 @@ Use the `openshift-version` option to request a specific version of OpenShift. Y
...
+Then, you can run commands on the Minishift VM in the same way that you run commands interactively on any remote machine using SSH.
+
+You can also run the command with a sub-command. The native Minishift client then sends the command to the Minishift VM
```
hferentschik: Example on what this means. `minishift ssh -- ls /` (or whatever we decide on)

hferentschik: "The native Minishift client", that's not really a thing. "Minishift uses the external SSH client to send the command to the VM and returns its results"

docs/source/using/managing-minishift.adoc
```diff
@@ -275,3 +275,29 @@ Use the `openshift-version` option to request a specific version of OpenShift. Y
...
+and returns the result to your local shell.
+
+Make sure that the path to the SSH client is specified in the _PATH_ environment variable.
+
+NOTE: You can use most OpenSSH-based clients. However, if you use the Windows operating system, only *cygwin* is supported.
```
hferentschik: "...only Cygwin's SSH implementation..."

Cygwin is a whole suite of tools. And is this correct? I thought at least for Windows 10 OpenSSH was an option?

docs/source/using/managing-minishift.adoc
```diff
@@ -275,3 +275,29 @@ Use the `openshift-version` option to request a specific version of OpenShift. Y
...
+
+If you do not have an external SSH client::
+You must specify a sub-command when you run `minishift ssh`. The native Minishift client then sends the command to the Minishift VM
```
hferentschik: Same as above.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: Also we should start with when an user does not have a SSH client. Because thats the more probable scenario. Then we should explain the difference when user has an SSH client.

docs/source/using/managing-minishift.adoc
LalatenduMohanty: I think an example will be useful.

hferentschik:
I am also wondering where this limitation of the native client comes from https://github.com/minishift/minishift/issues/919#issuecomment-306904466 

Is this a confirmed issue?

docs/source/using/managing-minishift.adoc
LalatenduMohanty: Linux and OSX would have ssh clients by default, so the scenario of having no SSH clients valid for Windows hosts. So this is not correct.

LalatenduMohanty: May be we change the note to add that most ssh clients will work fine but on Windows, OpenSSH has an issue and point to the issue.

docs/source/using/managing-minishift.adoc
thatdocslady: I don't think grammatically there is a problem with "your" and it's also consistent with the rest of the phrasing (including the title of this section). Is there a specific reason that you're proposing this change?

docs/source/using/managing-minishift.adoc
thatdocslady: "the native client" is something that I took from the issue description, to differentiate between the external SSH client and the Minishift client itself. If you think we don't need to be this explicit I can rephrase.

docs/source/using/managing-minishift.adoc
thatdocslady: Are you sure about this? In one of the other comments you mentioned that Linux/Mac come with an SSH client by default, so from that statement we can conclude that having an SSH client is actually more common. Also, the only reason that we needed to work around no SSH client in Windows is because of a bug in OpenSSH, which means that in the very near future (after this bug is fixed) we might even change this text to recommend/require an SSH client.

thatdocslady:
@hferentschik @LalatenduMohanty @gbraad can we decide on an example for running the minishift ssh command with a sub-command? I'm not an expert on ssh (although I did use it occasionally to test things), so I'd appreciate your expertise in coming up with an example that can reflect a typical operation that Minishift users might want to perform on the VM.

Also, some responses to feedback in the comments.

hferentschik:
I think we should put this on ice until we have some clarity around #1005. Here we describe things in a way that it is expected that you cannot use an interactive session with the native SSH implementation. This is not correct. This is really a regression somewhere.

docs/source/using/managing-minishift.adoc
gbraad: native is the Go implementation, as we observed different behaviour between the use of the external client, and the way docker-machine can provide this with the Go's native implementation. However, it is believed there is an issue with the way how minikube/minishift work => #1005

docs/source/using/managing-minishift.adoc
gbraad: For Windows `win32-openssh` (openssh-win32) is not an option unless the fixed release with https://github.com/PowerShell/Win32-OpenSSH/issues/666 is available on chocolatey.

Cygwin, choco git+unixtools, git bash all can work, but at the moment install using choco is not recommended.

gbraad:
I have fixed the issue #1005 (with #1009, and reported to minkube as https://github.com/kubernetes/minikube/pull/1571). This means the difference with native client and external client does not need to be described anymore. I will update the original task to reflect this.

This means that the only difference is:
  * interactive shell
  * subcommand given

docs/source/using/managing-minishift.adoc
LalatenduMohanty: Yes, user who have SSH client already i.e. Linux and OS X will not look in to the doc as it will work as expected. However @gbraad's PR we might not need this documentation.

LalatenduMohanty:
Marking this PR as `Do Not Merge` because we need wait for #1005 and as @gbraad mentioned we might not need documentation for this.

hferentschik:
> Marking this PR as Do Not Merge because we need to wait for #1005 

Pull request for #1005 is merged.

> and as @gbraad mentioned we might not need documentation for this.

On IRC we discussed that it might still make sense to document the various forms of interaction one can use with the _ssh_ command. Aka that one can either start an interactive shell or that one can execute once of command using 'minishift ssh -- <my command>'. I think it is fair enough to document this. Adding a note about which ssh client we recommend (let's try to avoid "support") on Windows makes sense. Maybe as a note. 

I think this might be useful, but can be convinced to drop this completely as well.

docs/source/using/managing-minishift.adoc
hferentschik: > Also, the only reason that we needed to work around no SSH client in Windows is because of a bug in OpenSSH

That's not quite correct. On Windows there were issues with external clients (aka OpenSSH) as well as the native impl (#1005).

LalatenduMohanty:
> On IRC we discussed that it might still make sense to document the various forms of interaction one can use with the ssh command. Aka that one can either start an interactive shell or that one can execute once of command using 'minishift ssh -- '. I think it is fair enough to document this.

Fair enough.

hferentschik:
@thatdocslady what's about we close this pull request and you create a new one which takes into account there is no differentiation between external and native client needed, since they behave the same. 

I think it would be easier than keep on discussing on this pull request.

thatdocslady:
@hferentschik @LalatenduMohanty @gbraad just recording here that we agreed that I will create a new PR with a new branch to address the updated issue and will reference this PR for history.

thatdocslady:
Superseded by PR #1013

Issue #1005, opened 2017-06-08
Using Go's native SSH implementation will not open interactive shell


gbraad:
## Steps to reproduce
```
$ sudo mv /usr/bin/ssh /usr/bin/-ssh
$ minishift start
$ minishift ssh
$ 
```
this behaviour is also observed in Minikube

while:
```
$ docker-machine start default
$ docker-machine ssh
# 
```

works.

gbraad:
Notes:
  * https://github.com/docker/machine/blob/master/libmachine/host/host.go#L75
  * https://github.com/docker/machine/blob/e1a03348ad83d8e8adb19d696bc7bcfb18ccd770/commands/ssh.go#L18
  * https://github.com/kubernetes/minikube/blob/master/pkg/minikube/cluster/cluster.go#L537

budhrg:
I had mentioned that earlier here https://github.com/minishift/minishift/pull/911#issuecomment-302043786. 

+1 for fixing it.

gbraad:
Still not sure why the QnD approach is needed, as it seems https://github.com/docker/machine/blob/master/libmachine/ssh/client.go#L318 actually checks for the length of the args.

gbraad:
@budhrg the understanding was that the native-client did not support interactive shell, but this is because we inherited it from Minikube. :-s Still looking for a cleaner fix.

gbraad:
We invoke the shell as follows https://github.com/minishift/minishift/blob/master/pkg/minikube/cluster/cluster.go#L511
```go
	return client.Shell(strings.Join(args, " "))
```
What happens is that we provde an empty string instead of no value: https://play.golang.org/p/z7DDUWnuYm (you can verify this with:

```go
package main

import (
	"strings"
	"fmt"
	
)

func main() {
	args := []string {}
	command := strings.Join(args, " ")
	fmt.Printf("Type: %T, Length: %v, Value: %v.\n", command, len(command), command)
}
```

At https://github.com/docker/machine/blob/master/libmachine/ssh/client.go#L269 the use of `args ...string` will result in args having a length of 1.

```go
func (client *NativeClient) Shell(args ...string) error {
```

this means that the check at https://github.com/docker/machine/blob/master/libmachine/ssh/client.go#L318 wrongly assumes that arguments are given, and it will execute https://github.com/docker/machine/blob/master/libmachine/ssh/client.go#L324

```go
		session.Run(strings.Join(args, " "))
```

[...]

PR #1009, opened 2017-06-09
Issue #1005 Interactive SSH

gbraad:
#1005

### DIFF
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -508,7 +508,8 @@ func CreateSSHShell(api libmachine.API, args []string) error {
 	if err != nil {
 		return err
 	}
-	return client.Shell(strings.Join(args, " "))
+
+	return client.Shell(args...)
 }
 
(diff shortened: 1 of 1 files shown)


gbraad:
This is just a quick and dirty fix to communicate and check if the functionality now works as expected accross the board

pkg/minikube/cluster/cluster.go
```diff
@@ -508,7 +508,12 @@ func CreateSSHShell(api libmachine.API, args []string) error {
 	if err != nil {
 		return err
 	}
-	return client.Shell(strings.Join(args, " "))
+
+	command := strings.Join(args, " ")
```
budhrg: Awesome! Looks like we can contribute back to minikube :)

budhrg:
Working fine for me :+1:

LalatenduMohanty:
Now we wont have to be dependent of any SSH Client, which is really cool.

hferentschik:
Merged, thanks.

Issue #996, opened 2017-06-06
"Stop" command should check state of minishift VM


agajdosi:
We should improve around the way in which Minishift handles states of VM and commands which change these states. 

Great fact is that it is not possible to start a running VM anymore if the VM is running. However `minishift stop` command seem to be still not handling the situation correctly: 

```
[agajdosi@dhcp-24-156 ~]$ minishift status
Does Not Exist
[agajdosi@dhcp-24-156 ~]$ minishift stop
Stopping local OpenShift cluster...
Error stopping cluster:  Host does not exist: "minishift"
Error stopping cluster: Host does not exist: "minishift"
```

Similar situation is in case when VM is already stopped. I think it could be useful to prevent stop command from trying like it happens in `delete`:

```
[agajdosi@dhcp-24-156 ~]$ minishift delete
The execution of this command requires an existing 'minishift' VM. However, there is currently none defined.
```

LalatenduMohanty:
@agajdosi Thanks. I agree that ` $ minishift stop` should check the state of VM and accordingly should provide the error.

agajdosi:
We can also add a feature which will describe behaviour of Minishift when handling states with `start`, `stop` and `delete` commands. Focusing only on problematic cases it could look like this:

```cucumber
Feature: Minishift handles states of VM effectively
Minishift is able to change states of VM - can create, stop or delete a VM,
is able to report current state to a user and do not execute state changing commands
in command-state cases which are not supported - eg. starting a running VM. 

  Scenario: Deleting non existing VM
    Given Minishift has state "Does Not Exist"
     When executing "minishift delete"
     Then Minishift should have state "Does Not Exist"
      And stdout should contain
      """
      The execution of this command requires an existing 'minishift' VM. However, there is currently none defined.
      """
  Scenario: Stopping non existing VM
    Given Minishift has state "Does Not Exist"
     When executing "minishift stop"
     Then Minishift should have state "Does Not Exist"
      And stdout should contain
      """
      The execution of this command requires an existing 'minishift' VM. However, there is currently none defined.
      """

  Scenario: Stopping stopped VM
    Given Minishift has state "Stopped"
     When executing "minishift stop"
     Then Minishift should have state "Stopped"
      And stdout should contain
      """
      Minishift is already stopped.
      """

  Scenario: Starting running VM
    Given Minishift has state "Running"
     When executing "minishift start"
     Then Minishift should have state "Running"
      And stdout should contain
      """
[...]

PR #1010, opened 2017-06-09
Issue #996 - `minishift stop` should check status of VM

dharmit:
Addresses issue #996

### DIFF
--- a/cmd/minishift/cmd/stop.go
+++ b/cmd/minishift/cmd/stop.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/docker/machine/libmachine"
+	"github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
@@ -40,16 +41,29 @@ VM but does not delete any associated files. To start the cluster again, use the
 }
 
 func runStop(cmd *cobra.Command, args []string) {
-	fmt.Println("Stopping local OpenShift cluster...")
 	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer api.Close()
+
+	// if VM does not exist, exit with error
+	util.ExitIfUndefined(api, constants.MachineName)
+
+	hostVm, err := api.Load(constants.MachineName)
+	if err != nil {
+		atexit.ExitWithMessage(1, err.Error())
+	}
+
+	// check if VM is already in stopped state
+	if util.IsHostStopped(hostVm.Driver) {
+		atexit.ExitWithMessage(0, fmt.Sprintf("The '%s' VM is already stopped.", constants.MachineName))
+	}
+
+	fmt.Println("Stopping local OpenShift cluster...")
 
 	if skipUnRegistration {
 		cluster.RegistrationParameters.SkipUnRegistration = true
 	}
 
 	if err := cluster.StopHost(api); err != nil {
-		fmt.Println("Error stopping cluster: ", err)
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error stopping cluster: %s", err.Error()))
 	}
 	fmt.Println("Cluster stopped.")
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -192,6 +192,15 @@ Feature: Basic
      When executing "minishift stop" succeeds
      Then Minishift should have state "Stopped"
 
+  Scenario: Stopping an already stopped VM
+    Given Minishift has state "Stopped"
+     When executing "minishift stop"
+     Then Minishift should have state "Stopped"
+      And stdout should contain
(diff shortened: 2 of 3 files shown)


budhrg:
retest this please

budhrg:
@dharmit , thanks could you add Integration test for it ?

cmd/minishift/cmd/stop.go
```diff
@@ -40,16 +41,29 @@ VM but does not delete any associated files. To start the cluster again, use the
...
+	// if VM does not exist, exit with error
+	util.ExitIfUndefined(api, constants.MachineName)
+
+	hostVm, err := api.Load(constants.MachineName)
+	if err != nil {
+		atexit.ExitWithMessage(1, err.Error())
+	}
+
+	// check if VM is already in stopped state
+	if util.IsHostStopped(hostVm.Driver) {
```
LalatenduMohanty: Why can't we use ExitIfNotRunning() . IMO we do not need `IsHostStopped`.

hferentschik:
Add to whitelist

cmd/minishift/cmd/stop.go
hferentschik: There is a semantic difference. _ExitIfNotRunning_ will just check for a running VM. There are more states than "stopped" and "running". If we use _ExitIfNotRunning_, Minishift won't try to attempt to stop if the VM is for example in state Error (which is probably should).

I think it is right to not use ExitIfNotRunning(). Whether it is enough to use check against "stopped" state is a different question. I think it will do though.

cmd/minishift/cmd/stop.go
```diff
@@ -40,16 +41,29 @@ VM but does not delete any associated files. To start the cluster again, use the
...
+		atexit.ExitWithMessage(1, fmt.Sprintf("The '%s' VM is already stopped.", constants.MachineName))
```
hferentschik: This is not an error condition. We should not exit with '1', but rather '0'

hferentschik:
I am basically fine this this change. As @budhrg is saying maybe extend basic.feature

dharmit:
I've extended `basic.feature` based on the comment https://github.com/minishift/minishift/issues/996#issuecomment-306468364 by @agajdosi. This commit is dealing only with stopped state so I've not added all the scenarios he's mentioned there. But let me know if I should because the other two conditions already seem to be satisfied by existing code.

test/integration/features/basic.feature
```diff
@@ -192,6 +192,24 @@ Feature: Basic
      When executing "minishift stop" succeeds
      Then Minishift should have state "Stopped"
 
+  Scenario: Stopping non existing VM
```
hferentschik: that does not work. After stopping the VM is still there, so the Given clause will fail.

test/integration/features/basic.feature
```diff
@@ -192,6 +192,24 @@ Feature: Basic
...
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift stop"
+     Then Minishift should have state "Does Not Exist"
+      And stdout should contain
+      """
+      The execution of this command requires an existing 'minishift' VM. However, there is currently none defined.
+      """
+
+  Scenario: Stopping stopped VM
```
hferentschik: +1 

Maybe call the scenario "Stopping an already stopped VM"

test/integration/features/basic.feature
```diff
@@ -192,6 +192,24 @@ Feature: Basic
...
+    Given Minishift has state "Stopped"
+     When executing "minishift stop"
```
hferentschik: When executing "minishift stop" succeeds

this way we check the error code as well

hferentschik:
What we want to add as test is:

  Scenario: Stopping already stopped VM
    Stopping an already stopped VM does print a message and exits without error
     When executing "minishift stop" succeeds
      And stdout should contain
      """
      Minishift is already stopped.
      """

dharmit:
@hferentschik that looks incomplete near the end. This part:
> And stdout should contain
         """
         Minis

hferentschik:
> that looks incomplete near the end. 

Fixed

hferentschik:
The assertion of the message is not quite right:

     Expected: Minishift is already stopped., Actual: The 'minishift' VM is already stopped.

I am going to adjust this while rebasing on master.

cmd/minishift/cmd/stop.go
LalatenduMohanty: > There is a semantic difference. ExitIfNotRunning will just check for a running VM. There are more states than "stopped" and "running". If we use ExitIfNotRunning, Minishift won't try to attempt to stop if the VM is for example in state Error (which is probably should).
> 
> I think it is right to not use ExitIfNotRunning(). Whether it is enough to use check against "stopped" state is a different question. I think it will do though.

Right, there seems to be a lot of state. I think it is fine with just check if it is already stopped.

```
var states = []string{
	"",
	"Running",
	"Paused",
	"Saved",
	"Stopped",
	"Stopping",
	"Starting",
	"Error",
	"Timeout",
}
```

hferentschik:
@dharmit, I merged your pull request. Thanks for another contribution. Much appreciated.

Issue #938, opened 2017-05-22
Interactively ask for username and password during registration


praveenkumar:
If a user doesn't use `MINISHIFT_USERNAME` and `MINISHIFT_PASSWORD` as environment variable + they don't want to use username/password flag then there should be an interactive shell which ask user about `username and password`

LalatenduMohanty:
CDK issue  https://issues.jboss.org/browse/cdk-110

gbraad:
Note: current implementation for password input does not echo the input with `*`. This is due the implementations that were available did not work as expected for a Windows console.

rctrh:
A sub-case that I think should be also covered - If the user specifies --username but doesn't specify password, then the user should be prompted for the password, or minimally an error should be generated since registration won't succeed with only a username.

hferentschik:
Resolved via pull request #1014

shuawest:
I just installed CDK 3.1.0-1. While using the environment variable method of install I say my username and password printed in plain text. 

```
- Command "sudo -E subscription-manager register --auto-attach --username jowest@redhat.com --password 'notmyrealpassword' " took 35.9s
```

When you use the prompt model the same happens.  

```
Red Hat Developers or Red Hat Subscription Management (RHSM) username: jowest@redhat.com
Red Hat Developers or Red Hat Subscription Management (RHSM) password: [HIDDEN]
| Command "sudo -E subscription-manager register --auto-attach --username jowest@redhat.com --password 'notmyrealpassword' " took 17s
```

The expected behavior is that is the password is stored in a more secure way, and never shows up in plain text in an log file, or console output. 

This issue should be reopened.

gbraad:
@shuawest Can you provide more information about your start command?

I guess you used `--show-libmachine-logs -v5`, and if so, there is no way for us to prevent this behaviour, unless we start messing with the way `libmachine` handles this. This is because we use an SSH connection to do the subscription command, and libmachine abstracts all these calls and allow debug logging. The same would be for all the ssh connections that might contain a password, such as hostfolders.

shuawest:
> Can you provide more information about your minishift start command or configuration?

The logging level was set to 4. 

```
minishift config set cpus 2
minishift config set disk-size 30G
minishift config set memory 8192
minishift config set vm-driver virtualbox
minishift config set server-loglevel 4
minishift start -v 4
```

gbraad:
It therefore worked as expected. Newer version will not log this at
[...]

PR #1014, opened 2017-06-12
Issue #938 Interactive ask username/password during registration

praveenkumar:


cmd/minishift/cmd/start.go
```diff
@@ -416,6 +420,20 @@ func setSubscriptionManagerParameters() {
 	cluster.RegistrationParameters.Password = viper.GetString(startFlags.Password.Name)
 }
 
+func checkRegistrationCredential(host *host.Host) {
+	needRegistration, _ := registration.NeedRegistration(host)
+	if needRegistration {
+		if cluster.RegistrationParameters.Username == "" {
+			username := pkgUtil.ReadInputFromStdin("Username to register to subscription-manager")
```
hferentschik: "Username to register VM"

subscription-manager is really RHEL specific, right?

cmd/minishift/cmd/start.go
```diff
@@ -416,6 +420,20 @@ func setSubscriptionManagerParameters() {
...
+			cluster.RegistrationParameters.Username = username
+		}
+		if cluster.RegistrationParameters.Password == "" {
+			password := pkgUtil.ReadPasswordFromStdin("Password to register to subscription-manager")
```
hferentschik: Same as above.

hferentschik: ReadPasswordFromStdin, nice

cmd/minishift/cmd/start.go
```diff
@@ -35,15 +39,14 @@ import (
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
 	"github.com/minishift/minishift/pkg/minishift/hostfolder"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
+	"github.com/minishift/minishift/pkg/minishift/registration"
 	"github.com/minishift/minishift/pkg/util"
+	pkgUtil "github.com/minishift/minishift/pkg/util"
```
hferentschik: minishiftUtil?

cmd/minishift/cmd/start.go
praveenkumar: @hferentschik no, it actually used for opensuse also `subscription-manager` is common but `RHN` is specific to RHEL.

cmd/minishift/cmd/start.go
praveenkumar: this is outside of minishift pkg and part of pkg/util instead pkg/minishift/util so I used this. Same what I used in previous PR https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/delete.go#L29 let me know if you still think this should be `minishiftUtil` so I wll change it to other occurrence also.

cmd/minishift/cmd/start.go
LalatenduMohanty: @praveenkumar Is there any way we can check which `register` we are using i.e. `Redhat` for for a RHEL ISO? If so we should make the string as `Enter Red Hat Username to register the VM`

LalatenduMohanty: I have got this feedback in the past for vagrant-registration that  when we ask for username and password it is not inventive which username and password user supposed to enter.

LalatenduMohanty:
@praveenkumar Also incase user enters a wrong username and password, user should be given couple of tries more i.e. 3 attempts in total.  It is something we had got as feedback in the past and had implemented in vagrant-registration.

praveenkumar:
> Is there any way we can check which register we are using i.e. Redhat for for a RHEL ISO? If so we should make the string as Enter Red Hat Username to register the VM

Right now we only check if a iso need registration or not (iso might be any distro). With current string `Username to register to subscription-manager` it clear which username user suppose to put I think.

praveenkumar:
> Also incase user enters a wrong username and password, user should be given couple of tries more i.e. 3 attempts in total. It is something we had got as feedback in the past and had implemented in vagrant-registration.

With current implementation we can't do right now because registration is kind of abstract from start command. If we need to implement this feature then might be refactor some code base because as per @hferentschik each user interaction should happen in the cmd level not any package level.

cmd/minishift/cmd/start.go
hferentschik: >  no, it actually used for opensuse also subscription-manager is common but RHN is specific to RHEL.

What if it where not SUSE?

hferentschik: > I have got this feedback in the past for vagrant-registration that when we ask for username and password it is not intuitive which username and password user supposed to enter.

If we want to be specific, we need to bake this into the registration framework. In it's simplest form, one would leave the values in RegistrationParameters nil and let the registrator take care of this. In our RedHatRegistrator in redhead.go. It becomes then the responsibility of the reregistrar to deal with non specified credentials. Works as well for me. Might be even nicer, since it allows you to be more specific in the message you provide. 

To integrate this nicely with the registration framework, the RegistrationParameters could provide a pointer to a function which can be used to interactively receive credentials. This way the solution would not only work for the Red Hat registration. We provide the right function to all registrars in case the credentials are not set via environment or command lines. It is up to the registrar then to either call the function or abort registration.

cmd/minishift/cmd/start.go
hferentschik: I see the ambiguity to minishift/util, especially if you would try to use both in the same go file. Happy to use something else as well. pkgUtil seems wrong, since it seems to be based on the _pkg_ directory in which it is located, but that is not really part of the package names. 

_generalUtils_? _inputUtils_ (since this is the context in which you are using this in this case)?

cmd/minishift/cmd/start.go
praveenkumar: @hferentschik Below is something you proposed ?

```go
////(registration.go)
// Registration defines distribution specific actions
type Registrator interface {
	provision.SSHCommander

	// Register
	Register(param *RegistrationParameters) error

	// Return the auth options used to configure remote connection for the daemon.
	Unregister(param *RegistrationParameters) error

	// Figure out whether this is a matching registrar
	CompatibleWithDistribution(osReleaseInfo *provision.OsRelease) bool

	// Use Interactive registration
	UseInteractiveRegistration() bool

	// Provide Username and Password
	GetInteractiveUsername() string
	GetInteractivePassword() string
}

//////(redhat.go)
func (registrator *RedHatRegistrator) UseInteractiveRegistration() bool {
	return true
}

func (registrator *RedHatRegistrator) GetInteractiveUsername() string {
	username := pkgUtil.ReadInputFromStdin("Username to register to subscription-manager")
	return username
}

func (registrator *RedHatRegistrator) GetInteractivePassword() string {
	password := pkgUtil.ReadPasswordFromStdin("Password to register to subscription-manager")
	return password
}
```

hferentschik: Almost ;-) This is option 2 I mentioned, where one extends the Registrator interface. In this case imo, the method to add are:

    GetInteractiveUsernameMessage() string
    GetInteractivePasswordMessage() string

We just need the message to display to make it specific (provided UseInteractiveRegistration is _true_). The actual retrieval of the data can still happen in the command.

.gitignore
```diff
@@ -35,6 +35,7 @@ _testmain.go
 *.iso
 /.idea
 *.iml
+/.vscode
```
hferentschik: Did you switch editor or how come you add this so late?

cmd/minishift/cmd/start.go
```diff
@@ -36,14 +40,12 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/hostfolder"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
 	"github.com/minishift/minishift/pkg/util"
+	inputUtils "github.com/minishift/minishift/pkg/util"
```
hferentschik: +1 IMO much better alias

cmd/minishift/cmd/start.go
```diff
@@ -414,6 +416,16 @@ func validateOpenshiftVersion() {
 func setSubscriptionManagerParameters() {
 	cluster.RegistrationParameters.Username = viper.GetString(startFlags.Username.Name)
 	cluster.RegistrationParameters.Password = viper.GetString(startFlags.Password.Name)
+	cluster.RegistrationParameters.GetUsernameInteractive = getUsernameInteractive
```
hferentschik: +1 :-)

cmd/minishift/cmd/start.go
```diff
@@ -414,6 +416,16 @@ func validateOpenshiftVersion() {
...
+	cluster.RegistrationParameters.GetPasswordInteractive = getPasswordInteractive
+}
+
+func getUsernameInteractive() string {
```
hferentschik: Right, except this method should take a parameter as well, the message. This is what got us here, remember? We wanted to have more specific messages when asking for credentials. For example now we can say "Please provide your RHEL subscription username" (or whatever @LalatenduMohanty mentioned in an earlier comment).

cmd/minishift/cmd/start.go
```diff
@@ -414,6 +416,16 @@ func validateOpenshiftVersion() {
...
+	return inputUtils.ReadInputFromStdin("Username to register to subscription-manager")
+}
+
+func getPasswordInteractive() string {
+	return inputUtils.ReadPasswordFromStdin("Password to register to subscription-manager")
```
hferentschik: Same as above. We want the registrar to specify the message. If we want we can add a check for empty string in which case a default this default message is used.

pkg/minishift/registration/redhat.go
```diff
@@ -55,6 +56,12 @@ func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) er
 		return err
 	} else {
 		if strings.Contains(output, "not registered") {
+			if param.Username == "" {
```
hferentschik: +1

pkg/minishift/registration/redhat.go
```diff
@@ -55,6 +56,12 @@ func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) er
...
+				param.Username = param.GetUsernameInteractive()
+			}
+			if param.Password == "" {
```
hferentschik: +1

pkg/minishift/registration/redhat.go
```diff
@@ -55,6 +56,12 @@ func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) er
 		return err
 	} else {
 		if strings.Contains(output, "not registered") {
+			if param.Username == "" {
+				param.Username = param.GetUsernameInteractive()
```
hferentschik: Here you want to pass your registrar specific message.

pkg/minishift/registration/utils.go
```diff
@@ -22,12 +22,20 @@ import (
...
 	"github.com/docker/machine/libmachine/provision"
-	"github.com/pkg/errors"
 )
 
 type RegistrationParameters struct {
-	Username string
-	Password string
+	Username               string
+	Password               string
+	GetUsernameInteractive func() string
```
hferentschik: GetUsernameInteractive func(string) string

cmd/minishift/cmd/start.go
hferentschik: return inputUtils.ReadInputFromStdin(fmt.Sprintf"%s :", message)

LalatenduMohanty: > "Please provide your RHEL subscription username"

Sounds good.

LalatenduMohanty:
> @LalatenduMohanty how badly do you want the retry?

I strongly feel retry needs to be part of this PR.  This is something we have learned from CDK2.x

pkg/minishift/registration/redhat.go
```diff
@@ -51,17 +52,36 @@ func (registrator *RedHatRegistrator) CompatibleWithDistribution(osReleaseInfo *
...
+	output, err := registrator.SSHCommand("sudo -E subscription-manager version")
+	if err != nil {
 		return err
-	} else {
-		if strings.Contains(output, "not registered") {
+	}
+	if strings.Contains(output, "not registered") {
+		for i := 1; i < 4; i++ {
+			if param.Username == "" {
+				param.Username = param.GetUsernameInteractive("Username to register to RHN")
```
LalatenduMohanty: -1. RHN is going to be shutdown in July 2017. Check https://access.redhat.com/products/red-hat-subscription-management/#migration 

I am fine with RHSM or just `Red Hat`.

praveenkumar:
@LalatenduMohanty updated to `RHSM` for message now.

pkg/minishift/registration/redhat.go
```diff
@@ -51,17 +52,36 @@ func (registrator *RedHatRegistrator) CompatibleWithDistribution(osReleaseInfo *
...
 func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) error {
-	if output, err := registrator.SSHCommand("sudo -E subscription-manager version"); err != nil {
+	output, err := registrator.SSHCommand("sudo -E subscription-manager version")
+	if err != nil {
 		return err
-	} else {
-		if strings.Contains(output, "not registered") {
+	}
+	if strings.Contains(output, "not registered") {
+		for i := 1; i < 4; i++ {
```
hferentschik: Interesting that you initialise the loop with 1.

pkg/minishift/registration/redhat.go
```diff
@@ -51,17 +52,36 @@ func (registrator *RedHatRegistrator) CompatibleWithDistribution(osReleaseInfo *
...
+			}
 			subscriptionCommand := fmt.Sprintf("sudo -E subscription-manager register --auto-attach "+
 				"--username %s "+
 				"--password '%s' ", param.Username, util.EscapeSingleQuote(param.Password))
-			if _, err := registrator.SSHCommand(subscriptionCommand); err != nil {
+			_, err = registrator.SSHCommand(subscriptionCommand)
+			if err == nil {
+				return nil
+			}
+			if strings.Contains(err.Error(), "Invalid username or password") {
```
hferentschik: So the idea is that if registration fails with explicitly specified credentials (env or CLI), it switches into an interactive mode? I guess that works. @LalatenduMohanty wdyt?

pkg/minishift/registration/redhat.go
```diff
@@ -51,17 +52,36 @@ func (registrator *RedHatRegistrator) CompatibleWithDistribution(osReleaseInfo *
...
+	output, err := registrator.SSHCommand("sudo -E subscription-manager version")
+	if err != nil {
 		return err
-	} else {
-		if strings.Contains(output, "not registered") {
+	}
+	if strings.Contains(output, "not registered") {
+		for i := 1; i < 4; i++ {
+			if param.Username == "" {
+				param.Username = param.GetUsernameInteractive("Username to register to RHSM")
```
hferentschik: I see, is RHSM the official acronym for the subscription manager? I was wondering whether it should say something like RedHat Developers credentials, but that is not really required, right? Other subscriptions would work as well. I guess the confusion for people coming from the Red Hat Developers has always been that they sign up with Red Hat Developers and get under the hood a Red Hat subscription. For them this is not transparent though and for them RHSM might not make much sense.

pkg/minishift/registration/redhat.go
LalatenduMohanty: @hferentschik yup. Works for me.

pkg/minishift/registration/redhat.go
LalatenduMohanty: > s RHSM the official acronym for the subscription manager?

Yes, I believe so.

> I was wondering whether it should say something like RedHat Developers credentials, but that is not really required, right? Other subscriptions would work as well. I guess the confusion for people coming from the Red Hat Developers has always been that they sign up with Red Hat Developers and get under the hood a Red Hat subscription. For them this is not transparent though and for them RHSM might not make much sense.

Valid points. I am not sure how to address this. Thats why I suggested `Username to register to Red Hat`. @praveenkumar any suggestions?

LalatenduMohanty:
@praveenkumar I like the way the PR woks now. Thanks. We can merge this once we decide on https://github.com/minishift/minishift/pull/1014#discussion_r121987846

pkg/minishift/registration/redhat.go
praveenkumar: https://access.redhat.com/articles/63269 that is official acronym. 

> Thats why I suggested Username to register to Red Hat. @praveenkumar any suggestions?

`Username to register to Red Hat` does look way to generic but I can go with something like `Username to register to RHSM/RHD` if that make sense?

LalatenduMohanty: @praveenkumar @hferentschik I think `Username for Red Hat subscription`  would make sense for all use cases.

gbraad: Are you sure Subscription and Username should use capitals?

hferentschik: I am fine with RHSM, but I think we should spell it out. Acronyms are always a pitfall for users. What's about:

"Username for Red Hat Subscription Management (RHSM):"

As said before, this still does not solve the ambiguity for RedHat Developers. Maybe:

"Red Hat Developers or Red Hat Subscription Management (RHSM) username:"

This somehow covers these two bases.

hferentschik: > Are you sure Subscription and Username should use capitals?

Not in "Red Hat Subscription Username"

praveenkumar: > "Red Hat Developers or Red Hat Subscription Management (RHSM) username:"

I can live with this, if all agree.

budhrg: > I can live with this, if all agree.

:+1:

PR #1016, opened 2017-06-12
Issue #1006 Implementation and unit tests

### LINKED ISSUE #1006: oc-env command errors on fish shell
```fish
➜ ~ eval (minishift oc-env)
Unexpected end of string, quotes are not balanced
- (line 1): begin; set -gx PATH $PATH /home/q/.minishift/cache/oc/v1.5.1"; # Run this command to configure your shell:  # eval (minishift oc-env) 
                                                                             ^
from sourcing file -
	called on line 60 of file /usr/share/fish/functions/eval.fish

in function “eval”
	called on standard input

source: Error while reading file “-”

```
This appears to be due to an unnecessary closing quote on https://github.com/minishift/minishift/blob/master/pkg/util/shell/shell.go#L77. If that is indeed the solution
I shall submit a PR shortly.

budhrg:
Fix #1006 

cc @hferentschik

### DIFF
--- a/pkg/util/shell/shell.go
+++ b/pkg/util/shell/shell.go
@@ -17,9 +17,15 @@ limitations under the License.
 package shell
 
 import (
+	"errors"
 	"fmt"
 	"github.com/docker/machine/libmachine/shell"
 	"os"
+	"strings"
+)
+
+var (
+	supportedShell = []string{"bash", "fish", "powershell", "cmd", "emacs", "tcsh", "zsh"}
 )
 
 type ShellConfig struct {
@@ -30,9 +36,21 @@ type ShellConfig struct {
 
 func GetShell(userShell string) (string, error) {
 	if userShell != "" {
+		if !isSupportedShell(userShell) {
+			return "", errors.New(fmt.Sprintf("'%s' is not a supported shell.\nSupported shells are %s.", userShell, strings.Join(supportedShell, ", ")))
+		}
 		return userShell, nil
 	}
 	return shell.Detect()
+}
+
+func isSupportedShell(userShell string) bool {
+	for _, shell := range supportedShell {
+		if userShell == shell {
+			return true
+		}
+	}
+	return false
 }
 
 func FindNoProxyFromEnv() (string, string) {
@@ -67,7 +85,7 @@ func GenerateUsageHint(userShell, cmdLine string) string {
 		cmd = fmt.Sprintf("eval $(%s)", cmdLine)
 	}
 
-	return fmt.Sprintf("%s Run this command to configure your shell: \n%s %s\n", comment, comment, cmd)
+	return fmt.Sprintf("%s Run this command to configure your shell:\n%s %s\n", comment, comment, cmd)
 }
 
 func GetPrefixSuffixDelimiterForSet(userShell string, pathVar bool) (prefix, suffix, delimiter string) {
@@ -77,7 +95,7 @@ func GetPrefixSuffixDelimiterForSet(userShell string, pathVar bool) (prefix, suf
 		suffix = "\";\n"
 		delimiter = " \""
 		if pathVar {
-			suffix = "\" $PATH;\n"
+			suffix = "\" \"$PATH\";\n"
 		}
 	case "powershell":
 		prefix = "$Env:"
--- a/pkg/util/shell/shell.go
+++ b/pkg/util/shell/shell.go
@@ -77,7 +77,7 @@ func GetPrefixSuffixDelimiterForSet(userShell string, pathVar bool) (prefix, suf
 		suffix = "\";\n"
 		delimiter = " \""
 		if pathVar {
-			delimiter = " $PATH "
+			suffix = "\" $PATH;\n"
 		}
 	case "powershell":
 		prefix = "$Env:"
--- a/pkg/util/shell/shell_test.go
+++ b/pkg/util/shell/shell_test.go
@@ -1,0 +1,125 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package shell
+
+import (
+	"os"
+	"testing"
+)
+
+type shellTestCase struct {
+	name         string
+	cmdLine      string
+	expectedHint string
+}
+
+type prefixSuffixDelimiterTestCase struct {
+	shellName string
+	prefix    string
+	suffix    string
+	pathVar   bool
+	delimiter string
+}
+
+func TestUnknownShell(t *testing.T) {
+	expectedShellName := "foo"
+	defer func(shell string) { os.Setenv("SHELL", shell) }(os.Getenv("SHELL"))
+	os.Setenv("SHELL", "/bin/"+expectedShellName)
+
+	shell, err := GetShell("")
+	if err != nil {
+		t.Fatal("Unexpected err : ", err.Error())
+	}
+
+	if shell != expectedShellName {
+		t.Fatalf("Expected shell to be %s but %s", expectedShellName, shell)
+	}
+}
+
+func TestKnownShell(t *testing.T) {
+	expectedShellName := "bash"
+	shell, err := GetShell("bash")
+	if err != nil {
+		t.Fatal("Unexpected err : ", err.Error())
+	}
+
+	if shell != expectedShellName {
+		t.Fatalf("Expected shell to be %s but %s", expectedShellName, shell)
+	}
+}
+
+func TestNoProxyFromEnvWhenUsedLowerCase(t *testing.T) {
+	expectedNoProxyVarName := "no_proxy"
+	expectedNoProxyVarValue := "FOO_BAR"
+	defer func(val string) { os.Setenv(expectedNoProxyVarName, val) }(os.Getenv(expectedNoProxyVarName))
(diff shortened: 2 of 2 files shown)


pkg/util/shell/shell_test.go
```diff
@@ -0,0 +1,217 @@
...
+	defer func(val string) { os.Setenv(expectedNoProxyVarName, val) }(os.Getenv(expectedNoProxyVarName))
+	os.Setenv(expectedNoProxyVarName, expectedNoProxyVarValue)
+
+	_, noProxyValue := FindNoProxyFromEnv()
+	if noProxyValue != expectedNoProxyVarValue {
+		t.Fatalf("Expected no proxy var value as %s but got %s", expectedNoProxyVarValue, noProxyValue)
+	}
+}
+
+func TestGenerateUsageHint(t *testing.T) {
```
hferentschik: I was hoping to test the actual final output of _oc-env_ and _docker-env_. This should not be so hard really. Rename _executeOcTemplateStdout_ to _executeOcTemplate_ and pass the writer into the function. For production code it is _os.Stdout_. For test purposes you can pass a string buffer. This way you can test the final output of the commands.

You should be able to do something like:

     shellCfg, err = getOcShellConfig("/home/john/.minishift/oc/1.5.1/oc", "fish")
     executeOcTemplateStdout(shellCfg, buffer)
    // make assertions against the buffer

I think such a test would be more valuable. It is almost like an integration test, except that we can skip the running VM.

WDYT?

hferentschik: obviously the test would be on the command level in this case

pkg/util/shell/shell_test.go
```diff
@@ -0,0 +1,217 @@
...
+	shellName string
+	prefix    string
+	suffix    string
+	pathVar   bool
+	delimiter string
+}
+
+func TestUnknownShell(t *testing.T) {
+	expectedShellName := "foo"
+	defer func(shell string) { os.Setenv("SHELL", shell) }(os.Getenv("SHELL"))
```
praveenkumar: @budhrg  I am not sure what this defer is going to do, does it set the default shell back to environment?

LalatenduMohanty:
Auto-detect is not working for fish shell. Is this expected?
```
~/c/g/s/g/m/minishift> minishift oc-env
export PATH="/Users/lmohanty/.minishift/cache/oc/v1.5.1:$PATH"
# eval $(minishift oc-env)
lmohanty@Lalatendus-MacBook-Pro ~/c/g/s/g/m/minishift> eval $(minishift oc-env)
fish: $(...) is not supported. In fish, please use '(minishift)'.
eval $(minishift oc-env)
     ^
```

LalatenduMohanty:
The auto-detection is not working because `$SHELL` returning bash in Fish shell
```
~/c/g/s/g/m/minishift> echo $SHELL
/bin/bash
```

pkg/util/shell/shell_test.go
budhrg: yes @praveenkumar

budhrg:
> The auto-detection is not working because $SHELL returning bash in Fish shell for me. It is because I am running Fish on top of bash.

Yes @LalatenduMohanty , verified by making fish shell as default login shell `chsh -s /bin/fish <username>`

pkg/util/shell/shell_test.go
budhrg: @hferentschik , I think those are part of `oc_env_test.go` and `env_test.go` files. I don't think we should mix the testing of such sort here. `shell.go` is a top level general implementation and we should write tests specific to it.

> I was hoping to test the actual final output of oc-env and docker-env. This should not be so hard really. Rename executeOcTemplateStdout to executeOcTemplate and pass the writer into the function. For production code it is os.Stdout. For test purposes you can pass a string buffer. This way you can test the final output of the commands.

For this, I would like to raise it in new issue and implementation as part of `oc_env_test.go` and `env_test.go`.

pkg/util/shell/shell_test.go
```diff
@@ -0,0 +1,217 @@
...
+	}
+
+	if shell != expectedShellName {
+		t.Fatalf("Expected shell to be %s but %s", expectedShellName, shell)
+	}
+}
+
+func TestKnownShell(t *testing.T) {
+	expectedShellName := "bash"
+	shell, err := GetShell("bash")
```
hferentschik: what happens if a user uses '/bin/bash'? I am wondering whether the impl of GetShell should not do something like:

    filepath.Base(shell), nil

As for the case where the shell is detected.

pkg/util/shell/shell_test.go
```diff
@@ -0,0 +1,217 @@
...
+	if err != nil {
+		t.Fatal("Unexpected err : ", err.Error())
+	}
+
+	if shell != expectedShellName {
+		t.Fatalf("Expected shell to be %s but %s", expectedShellName, shell)
+	}
+}
+
+func TestNoProxyFromEnvWhenUsedLowerCase(t *testing.T) {
```
hferentschik: where is this coming from?

pkg/util/shell/shell_test.go
hferentschik: >  I think those are part of oc_env_test.go and env_test.go files

Exactly that's what I meant

hferentschik: > For this, I would like to raise it in new issue and implementation as part of oc_env_test.go and env_test.go

Why? The very issue was from the start that _oc-env_ returned invalid output for _fish_. So I would very much think that a _oc_env_test.go_ is in scope here.

pkg/util/shell/shell_test.go
```diff
@@ -0,0 +1,217 @@
...
+			name: "cmd", cmdLine: "foo", expectedHint: `REM Run this command to configure your shell:
+REM 	@FOR /f "tokens=*" %i IN ('foo') DO @call %i
+`,
+		},
+		{
+			name: "emacs", cmdLine: "foo", expectedHint: `;; Run this command to configure your shell:
+;; (with-temp-buffer (shell-command "foo" (current-buffer)) (eval-buffer))
+`,
+		},
+	}
```
hferentschik: what's about the empty string case and the unknown shell type case?

pkg/util/shell/shell_test.go
```diff
@@ -0,0 +1,217 @@
...
+		hint := GenerateUsageHint(tt.name, tt.cmdLine)
+		if tt.expectedHint != hint {
+			t.Fatalf("Expected hint to be \n%s but got \n%s", tt.expectedHint, hint)
+		}
+
+	}
+}
+
+func TestGetPrefixSuffixDelimiterForSet(t *testing.T) {
+	testData := []prefixSuffixDelimiterTestCase{
```
hferentschik: I find this test not terribly useful. You are effectively just repeating what's setup in GetPrefixSuffixDelimiterForSet. IMO we can safely remove this test.

pkg/util/shell/shell_test.go
budhrg: @hferentschik , it is using `filepath.Base(shell), nil` only. `GetShell` is invoking `machine/libmachine/shell/shell.go#Detect()` under the hood.

pkg/util/shell/shell_test.go
budhrg: We have different behavior for `http_proxy` and `HTTP_PROXY` env vars. Hence, two test cases.

pkg/util/shell/shell_test.go
budhrg: @hferentschik , the `bash` one is checking unknown shell type case.

pkg/util/shell/shell_test.go
budhrg: Got it, will add `oc_env_test.go` :+1:

pkg/util/shell/shell_test.go
hferentschik: sure my point is that one might pass as parameter '/bin/bash' in which case it would make sense to just use the base similar to what the detect method does. 

I guess you assume the value passed by the user is one of 'fish, cmd, powershell, tcsh, bash, zsh'. Note though that you don't verify the parameters. If you add a check in the command that only these values are allowed, you will indeed never end up with something like '/bin/bash', but right now I think it is possible to do something like:

    $ minishift oc-env --shell /bin/bash 

That would not give the desired result right?

pkg/util/shell/shell_test.go
hferentschik: So you added these tests as part now, right? Just trying to figure out where they came from.

budhrg: Yes, one to test lower case and one for upper case HTTP_PROXY

pkg/util/shell/shell_test.go
budhrg: @hferentschik , the above one works as `GenerateUsageHint` accepts any shell name and if it is not "fish", or some defined shell, it default to bash like hint.

```
$ minishift oc-env --shell /bin/fish
export PATH="/home/budhram/.minishift/cache/oc/v1.5.1:$PATH"
# eval $(minishift oc-env)

$ minishift oc-env --shell fish
set -gx PATH "/home/budhram/.minishift/cache/oc/v1.5.1" $PATH;
# eval (minishift oc-env)
```

budhrg: The best way here is to deal the shell name verification at `GetShell`.

budhrg: Added ! Now we get error message as below:
```
$ minishift oc-env --shell fisha
Error running oc-env command: 'fisha' is not a supported shell.
```

pkg/util/shell/shell.go
```diff
@@ -17,9 +17,15 @@ limitations under the License.
...
 import (
+	"errors"
 	"fmt"
 	"github.com/docker/machine/libmachine/shell"
 	"os"
+	"strings"
+)
+
+var (
+	supportedShell = []string{"bash", "fish", "powershell", "cmd", "emacs", "tcsh", "zsh"}
```
hferentschik: +1 we could have gotten the list of shells also from shells.go, but fair enough.

hferentschik:
Rebased and merged. Thanks.

Issue #1004, opened 2017-06-08
Make basic.feature work for upstream and downstream


praveenkumar:
We need to make sure our basic features works for upstream and downstream without any additional patching. Currently we have below scenario doesn't work with downstream.

``` gherkin
  Scenario: User can list enabled plugins
    Given Minishift has state "Does Not Exist"
    When executing "minishift addons list" succeeds
    Then stdout should contain
     """
     - anyuid         : enabled    P(0)
     - admin-user     : disabled   P(0)
     """
```

agajdosi:
If I move my perspective from addons to whole Minishift I think there is need to design the way how to handle these differences. For particular CDK only features/scenarios we can have `@cdk` tag, but still there will be particular spheres where both Minishift and CDK will differ - list of addons, expected version of oc, list of example apps available in OpenShift... I did some research today to find possible ways of handling this aspect of keeping tests valid for both upstream and downstream.

**Opt. 1 - config file**
One option is to use some sort of config file and for those steps and their definitions which are different for upstream and downstream use values from config. In this case we would keep this config up to date for upstream so `make integration` would stay executable out of the box. For this purpose we probably should use [cucumber.yaml](https://github.com/cucumber/cucumber/wiki/cucumber.yml) file. (We could also probably save some lines/chars from make file when doing this.)

```cucumber
  Scenario: User can list enabled plugins
    Given Minishift has state "Does Not Exist"
    When executing "minishift addons list" succeeds
    Then stdout should contain config:addons_after_
```
This way we loose some readability, but as there are not many differences between Minishift and CDK it could be fine - most steps will be just as they are now.

**Opt. 2 - excluding scenarios by cucumber tags
This is just idea I had which can be maybe done (info for example [here](https://stackoverflow.com/questions/28408962/how-to-exclude-cucumber-tags)). Idea is to have both variants of same scenario for Minishift and CDK next to each other and exclude the one which is not relevant by tags. So we would have set B of basic scenarios tagged `@basic`, then set M of upstream valid scenarios tagged `@minishift` and set C of downstream valid scenarios tagged `@cdk`. For basic scenario with Minishift we would run: `B \ C`, for downstream we would just run `B \ M`. 

This solution seems ideal to me and I can investigate more tomorrow if nobody is against it.

agajdosi:
So logical operations are possible on Godog tags. This solution is working with our current code:
[...]

PR #1019, opened 2017-06-13
Issue #1004 Make basic.feature work for upstream and downstream

agajdosi:
Adding support for `@minishift` and `@cdk` tags - these are used on Minishift or CDK specific scenarios or their variants for Minishift/CDK. So if we have a scenario which is different for both, then we will have it twice - one with `@cdk` tag and values for CDK, other for Minishift... When integration tests starts, code checks if binary is Minishift or CDK and based on that it adds `&&~cdk` to tags, or respectively just `~cdk` if no tags are specified by user, these removes all cdk related cases so only those which are not specific or are specifically tagged `@minishift` remains and are run.

Addresses issue #1004

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -3,7 +3,7 @@ Feature: Basic
   As a user I can perform basic operations of Minishift and OpenShift
 
   Scenario: User can install default add-ons
-    Given Minishift has state "Does Not Exist"
+   Given Minishift has state "Does Not Exist"
     When executing "minishift addons install --defaults" succeeds
     Then stdout should contain
      """
@@ -11,15 +11,16 @@ Feature: Basic
      """
 
   Scenario: User can enable the anyuid add-on
-    Given Minishift has state "Does Not Exist"
+   Given Minishift has state "Does Not Exist"
     When executing "minishift addons enable anyuid" succeeds
     Then stdout should contain
      """
      Addon 'anyuid' enabled
      """
 
+  @minishift-only
   Scenario: User can list enabled plugins
-    Given Minishift has state "Does Not Exist"
+   Given Minishift has state "Does Not Exist"
     When executing "minishift addons list" succeeds
     Then stdout should contain
      """
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -55,6 +55,16 @@ func TestMain(m *testing.M) {
 	var godogPaths = flag.String("paths", "./features", "")
 	flag.Parse()
 
+	if *godogTags != "" {
+		*godogTags += "&&"
+	}
+	runner := util.MinishiftRunner{CommandPath: givenPath}
+	if runner.IsCDK() {
+		*godogTags += "~minishift-only"
+	} else {
+		*godogTags += "~cdk-only"
+	}
+
 	status := godog.RunWithOptions("minishift", func(s *godog.Suite) {
 		FeatureContext(s)
 	}, godog.Options{


test/integration/features/basic.feature
```diff
@@ -3,30 +3,42 @@ Feature: Basic
...
-    Given Minishift has state "Does Not Exist"
+   Given Minishift has state "Does Not Exist"
     When executing "minishift addons list" succeeds
     Then stdout should contain
      """
      - anyuid         : enabled    P(0)
      - admin-user     : disabled   P(0)
      """
 
+  @cdk
```
budhrg: :+1:

test/integration/integration_test.go
```diff
@@ -55,6 +55,16 @@ func TestMain(m *testing.M) {
...
 
+	if *godogTags != "" {
+		*godogTags += "&&"
+	}
+	runner := util.MinishiftRunner{CommandPath: givenPath}
+	if runner.IsCDK() {
+		*godogTags += "~minishift"
+	} else {
+		*godogTags += "~cdk"
+	}
```
praveenkumar: Shouldn't it other way around?

```go
	if runner.IsCDK() {
		*godogTags += "~cdk"
	} else {
		*godogTags += "~minishift"
```

agajdosi: @praveenkumar No, `~` is used in godog to exclude tagged scenarios.

I am using it this way due to fact that as whole feature is tagged `@basic` I am not able to add `@minishift` tagged scenarios, they are already in. And so I am here excluding `@minishift` scenarios for CDK use and vice versa.

This way we are not forced to repetively tag every scenario cdk or minishift, but we tag just few of them which we want to exclude. But this is at cost of less readability.

praveenkumar: @agajdosi ack thanks.

test/integration/features/basic.feature
hferentschik: I am not so sure about this for several reasons. 

* This makes CDK visible in our tests. I thought we said, we would keep CDK specific tests in the downstream repo
* If we have several of these cases, the feature file will become unwieldy 
* It might be confusing for contributors. It is not clear what 'CDK' means from a community perspective. 

Maybe we should discuss the strategy moving forward.

test/integration/integration_test.go
```diff
@@ -55,6 +55,16 @@ func TestMain(m *testing.M) {
 	var godogPaths = flag.String("paths", "./features", "")
 	flag.Parse()
 
+	if *godogTags != "" {
+		*godogTags += "&&"
+	}
+	runner := util.MinishiftRunner{CommandPath: givenPath}
+	if runner.IsCDK() {
```
hferentschik: I guess I am ok, to add these hooks into the code. This will allow us then to add CDK specific tests downstream. 

BTW, should the annotation maybe be names '@cdk-only' and 'minishift-only'?

agajdosi: @hferentschik +1 on `@minishift-only`, going to change it next update.

> This makes CDK visible in our tests. I thought we said, we would keep CDK specific tests in the downstream repo

It is possible I missed this, lets discuss this soon.

> If we have several of these cases, the feature file will become unwieldy

Agree. But hopefully we would not need many of these cases in basic.feature file. I still have the idea of ideally having some cdk.feature or more features tagged whole as `@cdk-only` which would focus on aspects of CDK. And then keep `@cdk-only` on individual scenarios/cases (added up or downstream) to be a rare case.

> It might be confusing for contributors. It is not clear what 'CDK' means from a community perspective.

If documented it could be fine, but still it would be a little bit "strange".

test/integration/features/basic.feature
```diff
@@ -3,23 +3,24 @@ Feature: Basic
...
   Scenario: User can enable the anyuid add-on
-    Given Minishift has state "Does Not Exist"
+   Given Minishift has state "Does Not Exist"
     When executing "minishift addons enable anyuid" succeeds
     Then stdout should contain
      """
      Addon 'anyuid' enabled
      """
 
+  @minishift-only
```
hferentschik: +1

test/integration/integration_test.go
```diff
@@ -55,6 +55,16 @@ func TestMain(m *testing.M) {
 	var godogPaths = flag.String("paths", "./features", "")
 	flag.Parse()
 
+	if *godogTags != "" {
```
hferentschik: Looking good now

hferentschik:
Rebased and merged, thanks.

hferentschik:
Will try to sync this downstream today, latest tomorrow ...

agajdosi:
@hferentschik thanks, Hardy.

Issue #1002, opened 2017-06-08
Allow to specify additional default and blocked docker registries


sabre1041:
Add support to specify additional default registries and blocked registries at startup

LalatenduMohanty:
@sabre1041 Can you please explain the use case for `default and blocked registries`. 

It seems it would be useful for users when they want to completely work with 3rd party or private registries. However it is tricky as we download OpenShift images from docker hub and if users blocks it during startup `minishift start` will fail. 

Also do you want to have these flags for both secure and in-secure registry?

jorgemoralespou:
In RHEL based docker (RHEL and centos) these flags exist. Upstream docker
don't support them.

This is useful in an Enterprise environment when they want to prevent
developers or deployments from images coming from dockerhub or any other
registry not trusted by the organization. The default registry prevents
having to fully qualify the registry, so the images can be hosted in places
like satellite that makes the organization not dependent on internet and
prevents having to configure docker proxies.

This only makes sense in downstream minishift.

praveenkumar:
Having it upstream will fail for b2d iso which doesn't have RHEL based docker. If we implement those flags here then user might be hit bug for b2d.

jorgemoralespou:
@praveenkumar Agree with you, but for CDK you might end up in trouble. So either you support all the bootstrapping and don't act in case of B2D image, or don't support it and run into CDK not supporting same features as OCP. I would think the former is better :-D, Although I guess it needs more thought than what @sabre1041 has provided.

sabre1041:
@jorgemoralespou @praveenkumar I would like to at least see this in the RHEL based CDK.

I can configure these options to only apply to the RHEL template if it is detected

praveenkumar:
> I  can configure these options to only apply to the RHEL template if it is detected

@sabre1041 For me it need some more time to figure out if we can have those options only passed to our provision templates.

> but for CDK you might end up in trouble. So either you support all the bootstrapping and don't act in case of B2D image, or don't support it and run into CDK not supporting same features as OCP. I would think the former is better :-D

@jorgemoralespou I hear you and will try to make to make sure those options only work for rhel/centos ISO.

hferentschik:
So what have we decided is the way forward on this. Where do we add what type of functionality?

praveenkumar:
> So what have we decided is the way forward on this. Where do we add what type of functionality?

[...]

PR #1027, opened 2017-06-15
Issue #1002 Allow to specify additional default and blocked Docker registries

praveenkumar:
I will create a downstream doc request also to put content around it.

Addresses issue #873, #1002

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -54,6 +54,10 @@ var (
 	VmDriver         = createFlag("vm-driver", SetString, []setFn{IsValidDriver}, []setFn{RequiresRestartMsg}, true)
 	OpenshiftVersion = createFlag("openshift-version", SetString, nil, nil, true)
 	HostOnlyCIDR     = createFlag("host-only-cidr", SetString, []setFn{IsValidCIDR}, nil, true)
+	DockerEnv        = createFlag("docker-env", SetSlice, nil, nil, true)
+	DockerEngineOpt  = createFlag("docker-opt", SetSlice, nil, nil, true)
+	InsecureRegistry = createFlag("insecure-registry", SetSlice, nil, nil, true)
+	RegistryMirror   = createFlag("registry-mirror", SetSlice, nil, nil, true)
 
 	// cluster up
 	SkipRegistryCheck = createFlag("skip-registry-check", SetBool, nil, nil, true)
--- a/cmd/minishift/cmd/config/util.go
+++ b/cmd/minishift/cmd/config/util.go
@@ -19,6 +19,7 @@ package config
 import (
 	"fmt"
 	"strconv"
+	"strings"
 )
 
 // Runs all the validation or callback functions and collects errors
@@ -68,5 +69,16 @@ func SetBool(m MinishiftConfig, name string, val string) error {
 		return err
 	}
 	m[name] = b
+	return nil
+}
+
+func SetSlice(m MinishiftConfig, name string, val string) error {
+	var tmpSlice []string
+	if val != "" {
+		for _, v := range strings.Split(val, ",") {
+			tmpSlice = append(tmpSlice, strings.TrimSpace(v))
+		}
+	}
+	m[name] = tmpSlice
 	return nil
 }
--- a/cmd/minishift/cmd/config/util_test.go
+++ b/cmd/minishift/cmd/config/util_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package config
 
 import "testing"
+import "reflect"
+import "strings"
 
 var minikubeConfig = MinishiftConfig{
 	"vm-driver":            "kvm",
@@ -73,5 +75,20 @@ func TestSetBool(t *testing.T) {
 	}
 	if !val {
 		t.Fatalf("SetBool set value is incorrect")
+	}
+}
+
+func TestSetSlice(t *testing.T) {
+	expectedSlice := []string{"172.0.0.1/16", "mycustom.registry.com/3030"}
+	err := SetSlice(minikubeConfig, "insecure-registry", strings.Join(expectedSlice, ","))
+	if err != nil {
+		t.Fatal("Cannot set Slice value in config: %s", err)
+	}
+	val, ok := minikubeConfig["insecure-registry"].([]string)
+	if !ok {
+		t.Fatalf("Type is not set to slice")
+	}
+	if !reflect.DeepEqual(expectedSlice, val) {
+		t.Fatalf("Expected: %+v, Got: %+v", expectedSlice, val)
 	}
 }
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -58,11 +58,14 @@ const (
 	b2dIsoAlias    = "b2d"
 	centOsIsoAlias = "centos"
 
+	defaultInsecureRegistry = "172.30.0.0/16"
+
 	unsupportedIsoUrlFormat = "Unsupported value for iso-url. It can be an URL, file URI or one of the following short names: [b2d centos]."
 )
 
 var (
 	dockerEnv        []string
+	dockerEngineOpt  []string
 	insecureRegistry []string
 	registryMirror   []string
 	openShiftEnv     []string
@@ -221,6 +224,25 @@ func handleProxies() *util.ProxyConfig {
 	return proxyConfig
 }
 
+func getSlice(key string) []string {
+	if viper.IsSet(key) {
+		return viper.GetStringSlice(key)
+	}
+	return nil
+}
+
+func determineInsecureRegistry(key string) []string {
+	s := getSlice(key)
+	if s != nil {
+		for _, v := range s {
+			if v == defaultInsecureRegistry {
+				return s
+			}
+		}
+	}
+	return append(s, defaultInsecureRegistry)
+}
+
 func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
 	machineConfig := &cluster.MachineConfig{
 		MinikubeISO:      determineIsoUrl(viper.GetString(startFlags.ISOUrl.Name)),
@@ -228,9 +250,10 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
 		CPUs:             viper.GetInt(startFlags.CPUs.Name),
 		DiskSize:         calculateDiskSizeInMB(viper.GetString(startFlags.DiskSize.Name)),
 		VMDriver:         viper.GetString(startFlags.VmDriver.Name),
-		DockerEnv:        dockerEnv,
-		InsecureRegistry: insecureRegistry,
-		RegistryMirror:   registryMirror,
+		DockerEnv:        getSlice(startFlags.DockerEnv.Name),
+		DockerEngineOpt:  getSlice(startFlags.DockerEngineOpt.Name),
+		InsecureRegistry: determineInsecureRegistry(startFlags.InsecureRegistry.Name),
+		RegistryMirror:   getSlice(startFlags.RegistryMirror.Name),
 		HostOnlyCIDR:     viper.GetString(startFlags.HostOnlyCIDR.Name),
 		ShellProxyEnv:    shellProxyEnv,
 	}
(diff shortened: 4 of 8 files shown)


LalatenduMohanty:
@praveenkumar Please add issue number to your commit message.

cmd/minishift/cmd/start.go
```diff
@@ -276,6 +278,7 @@ func initStartFlags() *flag.FlagSet {
 	startFlagSet.String(startFlags.DiskSize.Name, constants.DefaultDiskSize, "Disk size to allocate to the Minishift VM. Use the format <size><unit>, where unit = b, k, m or g.")
 	startFlagSet.String(startFlags.HostOnlyCIDR.Name, "192.168.99.1/24", "The CIDR to be used for the minishift VM. (Only supported with VirtualBox driver.)")
 	startFlagSet.StringArrayVar(&dockerEnv, "docker-env", nil, "Environment variables to pass to the Docker daemon. Use the format <key>=<value>.")
+	startFlagSet.StringSliceVar(&engineOpt, "docker-opt", nil, "Arbitrary flags to include with Docker daemon in the form flag=value")
```
LalatenduMohanty: `docker-opt` is not communicating properly about the purpose of the flag. My suggestion would be `docker-opt-flag` i.e.optional flag for  docker daemon

praveenkumar:
> Also I think it would be useful if user can save it using minishift config set docker-opt-flag AddRegistry=ip/hostname:port

@LalatenduMohanty No, using config set we can't have slice value. We did try but it not work atm that's the reason we don't have `insecure-registry` also part of our config set.

cmd/minishift/cmd/start.go
praveenkumar: `docker-opt` is something which is used for minikube also this term actually comes from docker-machine side and well understood. https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/start.go#L293

LalatenduMohanty:
> @LalatenduMohanty No, using config set we can't have slice value. We did try but it not work atm that's the reason we don't have insecure-registry also part of our config set.

Right.  I remember.

cmd/minishift/cmd/start.go
LalatenduMohanty: > docker-opt is something which is used for minikube also

I was not aware of that. So we are fine then.

hferentschik:
Well, maybe it is time now to look at this again. I think this is just some badly written code on our end. IMO it makes sense to look into this again as part of this issue.

cmd/minishift/cmd/start.go
hferentschik: but I think we can call the variable dockerEngineOpt to make things a bit clearer

hferentschik:
@praveenkumar could you please add the issue key to the commit message?

cmd/minishift/cmd/start.go
```diff
@@ -221,6 +222,7 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
 		DiskSize:         calculateDiskSizeInMB(viper.GetString(startFlags.DiskSize.Name)),
 		VMDriver:         viper.GetString(startFlags.VmDriver.Name),
 		DockerEnv:        dockerEnv,
+		EngineOpt:        engineOpt,
```
hferentschik: "Engine" seems a bit too generic to me. I think either "DockerOpt" or "DockerEngineOpt" to make it clear to what this really applies.

hferentschik:
@praveenkumar so where and how are these Docker engine options then picked up?

cmd/minishift/cmd/start.go
budhrg: +1 to change to `dockerEngineOpt`

cmd/minishift/cmd/start.go
budhrg: Same here for changing the name to `dockerEngineOpt`

cmd/minishift/cmd/config/config.go
```diff
@@ -54,6 +54,10 @@ var (
 	VmDriver         = createFlag("vm-driver", SetString, []setFn{IsValidDriver}, []setFn{RequiresRestartMsg}, true)
 	OpenshiftVersion = createFlag("openshift-version", SetString, nil, nil, true)
 	HostOnlyCIDR     = createFlag("host-only-cidr", SetString, []setFn{IsValidCIDR}, nil, true)
+	DockerEnv        = createFlag("docker-env", SetSlice, nil, nil, true)
```
hferentschik: :-) nice one. I love when things become more consistent

cmd/minishift/cmd/config/util.go
```diff
@@ -70,3 +71,10 @@ func SetBool(m MinishiftConfig, name string, val string) error {
 	m[name] = b
 	return nil
 }
+
+func SetSlice(m MinishiftConfig, name string, val string) error {
+	if val != "" {
```
hferentschik: 👍 

I think we discussed we need to document this somewhere. I would think somewhere in the persistent config section.

I am also wondering whether one could not write some standalone unit tests for this SetXYZ.

cmd/minishift/cmd/start.go
```diff
@@ -57,6 +57,7 @@ const (
 
 var (
 	dockerEnv        []string
+	dockerEngineOpt  []string
```
hferentschik: Still not quite sure where this is then picked up. Do we not need any changes to our provisioner now?

cmd/minishift/cmd/start.go
```diff
@@ -215,16 +216,25 @@ func handleProxies() *util.ProxyConfig {
 	return proxyConfig
 }
 
+func getSlice(key string) []string {
+	viper.Set(startFlags.InsecureRegistry.Name, []string{"172.30.0.0/16"})
```
hferentschik: That seems fugly. You are calling a getter, but internally the whole impl has side effects, since it calls a viper.Set as well. 

Also you are setting the value for keys which are completely unrelated to _startFlags.InsecureRegistry.Name_,  not to mention that you do it multiple times.

I think this needs to be handled explicitly.

cmd/minishift/cmd/start.go
```diff
@@ -215,16 +216,25 @@ func handleProxies() *util.ProxyConfig {
...
 	machineConfig := &cluster.MachineConfig{
 		MinikubeISO:      viper.GetString(startFlags.ISOUrl.Name),
 		Memory:           viper.GetInt(startFlags.Memory.Name),
 		CPUs:             viper.GetInt(startFlags.CPUs.Name),
 		DiskSize:         calculateDiskSizeInMB(viper.GetString(startFlags.DiskSize.Name)),
 		VMDriver:         viper.GetString(startFlags.VmDriver.Name),
-		DockerEnv:        dockerEnv,
-		InsecureRegistry: insecureRegistry,
-		RegistryMirror:   registryMirror,
+		DockerEnv:        getSlice(startFlags.DockerEnv.Name),
```

cmd/minishift/cmd/start.go
praveenkumar: @hferentschik This is auto picked up by libmachine and we don't need to change our provisioner for this.

hferentschik: ahh, that part I have been missing. Thanks.

cmd/minishift/cmd/start.go
```diff
@@ -300,9 +315,10 @@ func initStartFlags() *flag.FlagSet {
 	startFlagSet.Int(startFlags.CPUs.Name, constants.DefaultCPUS, "Number of CPU cores to allocate to the Minishift VM.")
 	startFlagSet.String(startFlags.DiskSize.Name, constants.DefaultDiskSize, "Disk size to allocate to the Minishift VM. Use the format <size><unit>, where unit = b, k, m or g.")
 	startFlagSet.String(startFlags.HostOnlyCIDR.Name, "192.168.99.1/24", "The CIDR to be used for the minishift VM. (Only supported with VirtualBox driver.)")
-	startFlagSet.StringArrayVar(&dockerEnv, "docker-env", nil, "Environment variables to pass to the Docker daemon. Use the format <key>=<value>.")
-	startFlagSet.StringSliceVar(&insecureRegistry, "insecure-registry", []string{"172.30.0.0/16"}, "Non-secure Docker registries to pass to the Docker daemon.")
-	startFlagSet.StringSliceVar(&registryMirror, "registry-mirror", nil, "Registry mirrors to pass to the Docker daemon.")
+	startFlagSet.StringArrayVar(&dockerEnv, startFlags.DockerEnv.Name, nil, "Environment variables to pass to the Docker daemon. Use the format <key>=<value>.")
+	startFlagSet.StringSliceVar(&dockerEngineOpt, startFlags.DockerEngineOpt.Name, nil, "Arbitrary flags to include with Docker daemon in the form flag=value")
```
hferentschik: I think "to include with" is a bit badly formulated. Qinikube's wording is a bit better: "Specify arbitrary flags to pass to the Docker daemon." (aka passing something to the daemon)

By the way, there is also a fullstop missing.

cmd/minishift/cmd/config/util_test.go
```diff
@@ -75,3 +77,18 @@ func TestSetBool(t *testing.T) {
 		t.Fatalf("SetBool set value is incorrect")
 	}
 }
+
+func TestSetSlice(t *testing.T) {
```
hferentschik: :-)

cmd/minishift/cmd/config/util_test.go
```diff
@@ -75,3 +77,18 @@ func TestSetBool(t *testing.T) {
...
+	err := SetSlice(minikubeConfig, "insecure-registry", strings.Join(expectedSlice, ","))
+	if err != nil {
+		t.Fatal("Cannot set Slice value in config: %s", err)
+	}
+	val, ok := minikubeConfig["insecure-registry"].([]string)
+	if !ok {
+		t.Fatalf("Type is not set to slice")
+	}
+	if !reflect.DeepEqual(expectedSlice, val) {
+		t.Fatalf("Expected: %+v, Got: %+v", expectedSlice, val)
```
hferentschik: Not so sure about the deep equal, but I guess in will work

cmd/minishift/cmd/start.go
```diff
@@ -221,16 +223,29 @@ func handleProxies() *util.ProxyConfig {
...
 		Memory:           viper.GetInt(startFlags.Memory.Name),
 		CPUs:             viper.GetInt(startFlags.CPUs.Name),
 		DiskSize:         calculateDiskSizeInMB(viper.GetString(startFlags.DiskSize.Name)),
 		VMDriver:         viper.GetString(startFlags.VmDriver.Name),
-		DockerEnv:        dockerEnv,
-		InsecureRegistry: insecureRegistry,
-		RegistryMirror:   registryMirror,
+		DockerEnv:        getSlice(startFlags.DockerEnv.Name),
+		DockerEngineOpt:  getSlice(startFlags.DockerEngineOpt.Name),
+		InsecureRegistry: determineInsecureRegistry(startFlags.InsecureRegistry.Name),
```

cmd/minishift/cmd/start.go
```diff
@@ -221,16 +223,29 @@ func handleProxies() *util.ProxyConfig {
...
+func getSlice(key string) []string {
+	if viper.IsSet(key) {
+		return viper.GetStringSlice(key)
+	}
+	return nil
+}
+
+func determineInsecureRegistry(key string) []string {
+	s := getSlice(key)
+	return append(s, defaultInsecureRegistry...)
```
hferentschik: This will not de-dup though, right? If the user specified "172.30.0.0/16" himself, it will end up twice in the final slice.

LalatenduMohanty:
Tested this and works for me on Linux/KVM.

cmd/minishift/cmd/start.go
praveenkumar: @hferentschik yes but that is already handed by docker daemon, I tested this scenario explicit and worked as expected.

cmd/minishift/cmd/config/util_test.go
praveenkumar: Yes tested and it did work, bit slow but works :)

cmd/minishift/cmd/start.go
hferentschik: See the problem is that now we depend on the Docker Daemon doing it, instead of doing it ourself. People might be stumbling over this and wondering the same thing and it is not apparent that de-duping is handled by the Docker Daemon (if we want to rely on this, we probably should add at least a comment). Also, we don't know whether this behaviour is going to change or not. This comes again down to the [Robustness principle](https://en.wikipedia.org/wiki/Robustness_principle) - "Be conservative in what you do, be liberal in what you accept from others".

In this case, conservative means, doing something as simple as de-duping ourselves, making the code explicit and not reliant on things which are not part of any sort of defined API. 

Just explaining where I am coming from here. If you think you want to rely on this de-duping being handled somewhere else, so be it.

hferentschik:
Merged thanks.

Issue #887, opened 2017-05-11
Expose node ports to host


durandom:
I'm trying to expose a non-http port to my host (e.g. postgresql or in my case kafka 9092) and follow https://docs.openshift.org/latest/dev_guide/getting_traffic_into_cluster.html#using-externalIP

oc create -f loadbalancer.yml

```yaml
apiVersion: v1
kind: Service
metadata:
  name: kafka-ingress
spec:
  ports:
  - name: kafka
    port: 9092
  type: LoadBalancer
  selector:
    name: apache-kafka
```

creates this service:

```yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: 2017-05-11T14:43:47Z
  name: kafka-ingress
  namespace: myproject
  resourceVersion: "2110"
  selfLink: /api/v1/namespaces/myproject/services/kafka-ingress
  uid: 3d74c24c-3658-11e7-a52d-5add44d633b6
spec:
  clusterIP: 172.30.119.105
  deprecatedPublicIPs:
  - 172.29.8.80
  externalIPs:
  - 172.29.8.80
  ports:
  - name: kafka
    nodePort: 30528
    port: 9092
    protocol: TCP
    targetPort: 9092
  selector:
    name: apache-kafka
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer:
    ingress:
    - ip: 172.29.8.80
```

now how do I get to the ip, which seems to be local to the minishift vm

I'm on osx, xhyve

hferentschik:
Have you tried accessing the node port under the Minishift IP? Something like `$(minishift ip): 30528`. I am not 100% sure whether this is going to work. The easier is to use _oc port-forward_.

Something like:

    $ oc port-froward POD [LOCAL_PORT:]REMOTE_PORT

Does that work for you?

hferentschik:
@durandom, I actually think connecting via VM IP + node port should work. Here is what I just tried, using MariaDB as an example:

    $ minishift start
    $ eval $(minishift oc-env)
    $ oc new-app -e MYSQL_ROOT_PASSWORD=foo https://raw.githubusercontent.com/openshift/origin/master/examples/db-templates/mariadb-persistent-template.json
    $ oc rollout status -w dc/mariadb
    $ oc expose dc mariadb --type=LoadBalancer --name=mariadb-ingress
    $ oc export svc mariadb-ingress
     ....
    ports:
        - nodePort: 30907
     ....
[...]

PR #1035, opened 2017-06-16
Issue #887 Adding a Exposing Services topic

hferentschik:
- Also extracting OpenShift related topics into a shared 'openshift' directory
- Refactoring navigation structure

Addresses issue #887


docs/source/_topic_map.yml
```diff
@@ -31,16 +31,23 @@ Topics:
...
+    Topics:
+      - Name: Overview
+        File: index
+      - Name: Using the OpenShift client binary
+        File: openshift-client-binary
+      - Name: Exposing Services
+        File: exposing-services
+      - Name: Accessing the OpenShift Docker registry
+        File: openshift-docker-registry
+  - Name: Contributing to Minishift
```
budhrg: :+1:

docs/source/openshift/exposing-services.adoc
```diff
@@ -0,0 +1,63 @@
...
+oc expose svc/frontend --hostname=www.example.com
+----
+
+To see a full example of creating an application and exposing it via a route, refer to the xref:../getting-started/quickstart.adoc#deploy-sample-app[quickstart].
+
+== NodePorts
+
+In the case that the service you want to expose is not HTTP based, you can create a link:https://docs.openshift.org/latest/architecture/core_concepts/pods_and_services.html#service-nodeport[_NodePort_] service.
+In this case each OpenShift node will proxy that port into your service.
+To access this port on your Minishfift VM, you need to configure an link:https://docs.openshift.org/latest/dev_guide/getting_traffic_into_cluster.html#using-ingress-IP-self-service[Ingress IP Self-Service] using `oc expose` with the parameter `type=LoadBalancer`.
```
budhrg: Typo for `minishift` word here.

docs/source/openshift/openshift-client-binary.adoc
```diff
@@ -46,30 +39,25 @@ for example, logging into another OpenShift instance, you can run:
...
-to authenticate against the local cluster. This means any non-empty user name and password can
-be used to login to the local cluster. The recommended user name and password combination is
-developer/developer, because it already has a default project _myproject_ and also can
-https://docs.openshift.org/latest/architecture/additional_concepts/authentication.html#authentication-impersonation[impersonate]
-the administrator. This allows you to run administrator commands using the `--as system:admin` parameter.
+By default, _cluster up_ uses https://docs.openshift.org/latest/install_config/configuring_authentication.html#AllowAllPasswordIdentityProvider[AllowAllPasswordIdentityProvider] to authenticate against the local cluster.
+This means any non-empty user name and password can be used to login to the local cluster.
+The recommended user name and password combination is developer/developer, because it already has a default project _myproject_ and also can https://docs.openshift.org/latest/architecture/additional_concepts/authentication.html#authentication-impersonation[impersonate] the administrator.
```
budhrg: The recommended user name and password combination is _developer_ and _developer_ respectively since it already has a default project ...

hferentschik: TBH, I find your wording even harder to process, especially the "developer and developer respectively" part.

docs/source/_topic_map.yml
hferentschik: I hope this makes thing clearer. I thought about the ambiguity here before, but now that we have an actual case showing that this is actually happening, it is time to change. I think also pulling up the OpenShift topic helps.

docs/source/openshift/exposing-services.adoc
```diff
@@ -0,0 +1,63 @@
...
+[[overview]]
+== Overview
+
+There are several ways you can expose your service once deployed on OpenShift.
+The following sections describes the various alternatives and when to use them.
+
+[[application-routes]]
+== Routes
+
+Provided you are deploying some form of web application, the most common way to expose it is by a so called link:https://docs.openshift.org/latest/dev_guide/routes.html[route].
```
Preeticp: Suggest using: If you want to deploy a web application, the most common way to expose it is by using routes.

docs/source/openshift/exposing-services.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+----
+oc expose svc/frontend --hostname=www.example.com
+----
+
+To see a full example of creating an application and exposing it via a route, refer to the xref:../getting-started/quickstart.adoc#deploy-sample-app[quickstart].
+
+== NodePorts
+
+In the case that the service you want to expose is not HTTP based, you can create a link:https://docs.openshift.org/latest/architecture/core_concepts/pods_and_services.html#service-nodeport[_NodePort_] service.
```
Preeticp: In case the service you...

docs/source/openshift/openshift-client-binary.adoc
Preeticp: The recommended user name and password is _developer_ and _developer_ respectively since it already has a default project...

docs/source/openshift/exposing-services.adoc
hferentschik: It might not be a web application per se. It might be a micro service which falls also into this category. Which part does bother you here? The "some form of"?

docs/source/openshift/exposing-services.adoc
hferentschik: Ok

Preeticp:
@thatdocslady request you too to have a look at this PR since it introduces changes to the doc structure. I feel it is alright, but it would be great to have your opinion too.

docs/source/openshift/openshift-client-binary.adoc
hferentschik: Hmm, ok

docs/source/openshift/exposing-services.adoc
Preeticp: Use of active voice in the first part of the sentence is missing. Why are we using 'so called' it could imply that it is 'Incorrectly or falsely termed' as routes.

Preeticp: How about : If you want to deploy some form of web application, the most common way to expose it is to use routes.

docs/source/openshift/index.adoc
```diff
@@ -0,0 +1,13 @@
+[[interacting-with-openshift]]
+= Interacting with OpenShift
+:icons:
+
+Minishift creates a virtual machine (VM) and provisions a local, single-node OpenShift
```
thatdocslady: Missing Overview H2

docs/source/using/index.adoc
```diff
@@ -1,14 +1,12 @@
-[[using-index]]
+[[using-minishift]]
 = Using Minishift
```
thatdocslady: Maybe change this to "Managing Minishift", now that we separated the OpenShift topics? I'm not terribly stubborn about this suggestion, btw.

docs/source/using/index.adoc
```diff
@@ -1,14 +1,12 @@
...
 :icons:
 
 This section describes how to manage your Minishift VM and interact with your
 OpenShift cluster. It also helps troubleshooting common issues.
```
thatdocslady: now this section doesn't describe how to interact with openshift anymore :)

hferentschik:
> Looks great overall, I dig the separated openshift/minishift topics. 

Thanks. The driving force was adding this new "Exposing services" and the realisation that the grouping did not quite work anymore. Especially since I think there will be more topics in the line of "Exposing services" coming up. 

I think these are the valuables things we need to document, not how to deploy application X, Y or Z.

> Added a couple of really tiny comments/suggestions besides.

I'm on it.

docs/source/openshift/index.adoc
hferentschik: On second thought. None of the index pages has a second level heading and "Overview" section. They are also not including a TOC. So I think this is actually correct, at least looking at the current usage.

docs/source/using/index.adoc
hferentschik: Let's stick with "Using" for now. You are managing a server, but using s CLI. That's at least where I am coming from atm.

docs/source/using/index.adoc
hferentschik: +1

docs/source/openshift/index.adoc
hferentschik: Double checking with the rest of docs.openshift.org, it seems we are doing the same as other topics.

docs/source/using/index.adoc
hferentschik: Let's defer this for a bit, if possible. I think we go with managing we would need to say something like "Managing the Minishift VM". For now I suggest to keep it, but let's keep this in mind.

hferentschik:
@thatdocslady @Preeticp - I pushed an update with a change to _index.adoc_ for the using topic. WDYT can we merge this now?

thatdocslady:
@hferentschik @Preeticp LGTM, the only thing that I just noticed is the file name for the OpenShift registry topic doesn't actually need the word "the", but this is an enhancement for convenience (shorter name) rather than a convention so up to you.

hferentschik:
Ok, I am going to merge then

hferentschik:
>  I just noticed is the file name for the OpenShift registry topic doesn't actually need the word "the"

It is not in the filename, but in the anchor. Whether or not one adds it, is a question of how close you want to match the section title.

PR #1041, opened 2017-06-20
Issue #1011 Fix Adding developer to sudoers when using OpenShift 3.6.x

### LINKED ISSUE #1011: Adding developer to sudoers fails when using 3.6.x version of OpenShift
Looks like we have some code regression the way `minishift_kubeconfig` created and this make minishift failed during adding `developer` user to sudoers.

@jorgemoralespou 

```
$ ./minishift start --openshift-version v3.6.0-alpha.2
[...]
Image pull complete
OpenShift server started.

    https://192.168.42.71:8443

You are logged in as:
    User:     developer

To login as administrator:
    oc login -u system:admin

Error during post cluster up configuration: Unable to add sudoer role
```

praveenkumar:
Fix #1011

### DIFF
--- a/pkg/minikube/kubeconfig/config.go
+++ b/pkg/minikube/kubeconfig/config.go
@@ -60,7 +60,7 @@ func GetConfigPath() string {
 }
 
 // Cache system admin entries to be used to run oc commands
-func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string) error {
+func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string, userName string) error {
 	config, err := Read(GetConfigPath())
 	if err != nil {
 		return errors.New(fmt.Sprintf("Error reading config file %s", systemEntriesConfigPath))
@@ -82,9 +82,8 @@ func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string) error 
 		}
 	}
 
-	userName := fmt.Sprintf("system:admin/%s", clusterName)
 	for k, v := range config.Users {
-		if v.Name == userName || v.Name == "system:admin/127-0-0-1:8443" {
+		if v.Name == userName {
 			targetConfig.Users = append(targetConfig.Users, config.Users[k])
 			break
 		}
--- a/pkg/minishift/clusterup/clusterup.go
+++ b/pkg/minishift/clusterup/clusterup.go
@@ -20,9 +20,14 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+
 	"github.com/blang/semver"
 	"github.com/docker/machine/libmachine/host"
 	"github.com/docker/machine/libmachine/provision"
+
+	"os"
+	"strings"
+	"time"
 
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minikube/kubeconfig"
@@ -31,9 +36,6 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/oc"
 	"github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/version"
-	"os"
-	"strings"
-	"time"
 )
 
 const (
@@ -75,7 +77,16 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 
 // PostClusterUp runs the Minishift specific provisioning after 'cluster up' has run
 func PostClusterUp(clusterUpConfig *ClusterUpConfig, sshCommander provision.SSHCommander, addOnManager *manager.AddOnManager) error {
-	err := kubeconfig.CacheSystemAdminEntries(clusterUpConfig.KubeConfigPath, getConfigClusterName(clusterUpConfig.Ip, clusterUpConfig.Port))
+	isOpenshift3_6 := util.VersionOrdinal(clusterUpConfig.OpenShiftVersion) >= util.VersionOrdinal("v3.6.0")
+
+	clusterName := getConfigClusterName(clusterUpConfig.Ip, clusterUpConfig.Port)
+	userName := fmt.Sprintf("system:admin/%s", clusterName)
+	// See issue https://github.com/minishift/minishift/issues/1011. Needs to be revisted whether needed with final OpensShift 3.6 (HF)
+	if isOpenshift3_6 {
+		userName = fmt.Sprintf("system:admin/127-0-0-1:%d", clusterUpConfig.Port)
+	}
+
+	err := kubeconfig.CacheSystemAdminEntries(clusterUpConfig.KubeConfigPath, getConfigClusterName(clusterUpConfig.Ip, clusterUpConfig.Port), userName)
 	if err != nil {
 		return err
 	}
--- a/pkg/minikube/kubeconfig/config_test.go
+++ b/pkg/minikube/kubeconfig/config_test.go
@@ -18,11 +18,12 @@ package kubeconfig
 
 import (
 	"fmt"
-	"gopkg.in/yaml.v2"
 	"io/ioutil"
 	"os"
 	"strings"
 	"testing"
+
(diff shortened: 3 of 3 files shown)


budhrg:
@praveenkumar , you mean by `3.6.*` instead of `1.3.6` in commit message right? Need update if so.

budhrg:
This PR is working as expected for me with https://ci.appveyor.com/api/buildjobs/0w0waa3d1w8mfgvn/artifacts/out%2Fwindows-amd64%2Fminishift.exe. 

@hferentschik , please integrate this with https://github.com/minishift/minishift/pull/1017 before merging it and see if the issue still exists or not.
For me, I am still getting same error when cherry-picking this PR commit into your PR.

hferentschik:
> please integrate this with #1017 before merging it and see if the issue still exists or not.

Why? The problem is fully reproducible and testable w/o #1017.

pkg/minishift/clusterup/clusterup.go
```diff
@@ -75,7 +77,11 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 
 // PostClusterUp runs the Minishift specific provisioning after 'cluster up' has run
 func PostClusterUp(clusterUpConfig *ClusterUpConfig, sshCommander provision.SSHCommander, addOnManager *manager.AddOnManager) error {
-	err := kubeconfig.CacheSystemAdminEntries(clusterUpConfig.KubeConfigPath, getConfigClusterName(clusterUpConfig.Ip, clusterUpConfig.Port))
+	isVersion3_6, err := ValidateOpenshiftMinVersion(clusterUpConfig.OpenShiftVersion, "v3.6.0-alpha.0")
```
hferentschik: will this work with ocp?

pkg/minikube/kubeconfig/config.go
```diff
@@ -59,7 +60,7 @@ func GetConfigPath() string {
 }
 
 // Cache system admin entries to be used to run oc commands
-func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string) error {
+func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string, isVersion3_6 bool) error {
```
hferentschik: I don't get why you have to pass _isVersion3_6_ into this method. IMO you can keep this method as is and just change the clusterName parameter clusterup.go. Then the need to check the version stays in one spot.

hferentschik:
> This PR is working as expected for me 

How did you verify this? For me it still does not work as expected. Here is what I did.

* `rm ~/.kube/config`
* `minishift start --vm-driver xhyve` # OK will provision 1.5.1
* `minishift delete`
* `minishift start --vm-driver xhyve --openshift-version v3.6.0-alpha.2` # OK will provision 3.6.0-alpha.2
* `minishift delete`
* `minishift start --vm-driver xhyve` # FAIL "Error during post cluster up configuration: Unable to add sudoer role"

@budhrg, @praveenkumar - Is the above sequence working for you?

budhrg:
@hferentschik , looks like someone reproduce it here https://github.com/minishift/minishift/issues/1053.

hferentschik:
I have been over this - https://github.com/minishift/minishift/pull/1041#issuecomment-309879842 - multiple times and I keep getting the same behaviour. Even when I delete the full ~/.kube directory. 

Once it fails, I'll delete ~/.kube/config and start again and it works. 

At that stage I will need to add more debug trace to see what the exact error is.

budhrg:
@LalatenduMohanty @praveenkumar , can one of you try this PR in both Mac or Linux ? 
Well I don't want to say "it can be OS specific" but let's confirm it too. Also, James seems to have reproduce it here  https://github.com/minishift/minishift/pull/1041 in Mac. 

I am not able to reproduce in Linux (F25).
Sequences to test : 
* `rm ~/.kube/config`
* `minishift start`
* `minishift delete`
* `minishift start --openshift-version v3.6.0-alpha.2`
* `minishift delete`
* `minishift start`

pkg/minikube/kubeconfig/config.go
```diff
@@ -81,8 +81,10 @@ func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string) error
 			break
 		}
 	}
-
 	userName := fmt.Sprintf("system:admin/%s", clusterName)
+	if isVersion3_6 {
+		userName = "system:admin/127-0-0-1:8443"
+	}
 	for k, v := range config.Users {
 		if v.Name == userName || v.Name == "system:admin/127-0-0-1:8443" {
```
hferentschik: @praveenkumar @budhrg there bug in the code is here: 

    if v.Name == userName || v.Name == "system:admin/127-0-0-1:8443"

To start with the _v.Name == "system:admin/127-0-0-1:8443_ is generally not needed anymore with the introduction of the _isVersion3_6_ flag. If __isVersion3_6_ is true, both condition are actually identical. However, what is worse, it makes things fail independent of the OpenShift version. config.Users is iterating over a map which does not guarantee any order. Remember the 127-0-0-1:8443 is not gone form the config. It seems that on Linux the iteration order is different than on OS X. Removing _|| v.Name == "system:admin/127-0-0-1:8443"_ should do the trick.

pkg/minikube/kubeconfig/config.go
```diff
@@ -60,7 +60,7 @@ func GetConfigPath() string {
 }
 
 // Cache system admin entries to be used to run oc commands
-func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string) error {
+func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string, isVersion3_6 bool) error {
```
hferentschik: maybe _isOpenShift3_6_ as flag name.

pkg/minikube/kubeconfig/config.go
LalatenduMohanty: I am little confused. The change in logic for all versions of Origin > = 3.6 , right? But the code seems to be on only 3.6 version..

LalatenduMohanty:
@budhrg @praveenkumar It is failing for me 

```
-- Creating initial project "myproject" ... OK
-- Removing temporary directory ... OK
-- Checking container networking ... OK
-- Server Information ... 
   OpenShift server started.
       https://192.168.64.7:8443

   You are logged in as:
       User:     developer
       Password: developer

   To login as administrator:
       oc login -u system:admin

Error during post cluster up configuration: Unable to add sudoer role
```

budhrg:
> It is failing for me on macOS.

and in Fedora ?

pkg/minikube/kubeconfig/config.go
praveenkumar: @hferentschik looks like I made a mistake remove that from my current PR.

budhrg:
@LalatenduMohanty Is the explanation of bug https://github.com/minishift/minishift/pull/1041#discussion_r123334995.

The reason I was not able to reproduce is I had the previous commit where I didn't have the follow code snippet.
```
v.Name == "system:admin/127-0-0-1:8443"
```

cc @hferentschik

hferentschik:
> The reason I was not able to reproduce is I had the previous commit where I didn't have the follow code snippet.

then I have no idea what you tested?!

Either way, it looks like the code is updated.

pkg/minikube/kubeconfig/config.go
```diff
@@ -60,7 +60,7 @@ func GetConfigPath() string {
 }
 
 // Cache system admin entries to be used to run oc commands
-func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string) error {
+func CacheSystemAdminEntries(systemEntriesConfigPath, clusterName string, isOpenshift3_6 bool) error {
```
hferentschik: I am still not happy about the fact that the version check "leaks" even into _CacheSystemAdminEntries_. Here is a proposal for a change - https://github.com/praveenkumar/minishift/pull/1

hferentschik:
Merged, thanks.

PR #1043, opened 2017-06-20
[WIP] Issue #1043 Add fabric8 add-on

### LINKED ISSUE #1034: Create a fabric8 add-on
lets create an addon so its easy to install fabric8 into minishift from the command line.

We're getting close to a repeatable install on minishift; so I should be able to submit a PR for a first spike at an addon next week.

Is there a way to specify a parameter to an addon? e.g. I'd like the fabric8 addon to use the latest release of fabric8 but it might be nice to let a user override which version to use via the minishift CLI?

jstrachan:
Fix #1034 

first spike of a fabric8 addon. Currently has some missing issues
* does not enable CORS
* does not pass in the github OAuth client ID or secret (e.g. from env vars?)
* routing-suffix does not seem to work?

### DIFF
--- a/addons/fabric8/fabric8.addon
+++ b/addons/fabric8/fabric8.addon
@@ -1,0 +1,39 @@
+# Name: fabric8
+# Description: Installs fabric8 locally on MiniShift
+
+oc login -u system:admin
+
+echo Adding the OAuthClient and cluster-admin role to the init-tenant service account
+oc new-app --param URL="https://keycloak-fabric8.#{ip}.#{routing-suffix}/auth/realms/fabric8/broker/openshift-v3/endpoint" https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-oauthclient-template.yml
+oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:fabric8:init-tenant
+
+oc login -u developer -p developer
+
+oc new-project developer
+oc new-project fabric8
+
+echo Installing the Fabric8 System using GitHub Client ID ${GITHUB_OAUTH_CLIENT_SECRET}
+#oc new-app --param APISERVER_HOSTPORT="#{ip}:8443" --param GITHUB_OAUTH_CLIENT_SECRET=${GITHUB_OAUTH_CLIENT_SECRET} --param GITHUB_OAUTH_CLIENT_ID=${GITHUB_OAUTH_CLIENT_ID} https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-system-openshift.yml
+oc new-app --param APISERVER_HOSTPORT="#{ip}:8443" https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-system-openshift.yml
+
+echo Please wait while the pods all startup!
+echo
+echo To watch this happening you can type:
+echo   oc get pod -l provider=fabric8 -w
+echo
+echo Or you can watch in the OpenShift console via:
+echo   minishift console
+echo
+echo When the pods are all running please click on the following URLs in your browser, then ADVANCED, then click the URL at the bottom
+echo To approve the certs
+echo
+echo   https://keycloak-fabric8.#{ip}.#{routing-suffix}/
+echo   https://wit-fabric8.#{ip}.#{routing-suffix}/api/status
+echo   https://forge-fabric8.#{ip}.#{routing-suffix}/forge/version
+echo   https://fabric8-fabric8.#{ip}.#{routing-suffix}/
+echo
+echo
+echo Then you should be able the open the fabric8 console here:
+echo   https://fabric8-fabric8.#{ip}.#{routing-suffix}/
+
+
--- a/addons/fabric8/fabric8.addon
+++ b/addons/fabric8/fabric8.addon
@@ -1,5 +1,7 @@
 # Name: fabric8
-# Description: Installs fabric8 locally on MiniShift
+# Description: Installs fabric8 locally on Minishift
+
+# TODO enable CORS!
 
 oc login -u system:admin
 
@@ -13,6 +15,7 @@ oc new-project developer
 oc new-project fabric8
 
 echo Installing the Fabric8 System using GitHub Client ID ${GITHUB_OAUTH_CLIENT_SECRET}
+# TODO lets uncomment when we figure out parameter passing!
 #oc new-app --param APISERVER_HOSTPORT="#{ip}:8443" --param GITHUB_OAUTH_CLIENT_SECRET=${GITHUB_OAUTH_CLIENT_SECRET} --param GITHUB_OAUTH_CLIENT_ID=${GITHUB_OAUTH_CLIENT_ID} https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-system-openshift.yml
 oc new-app --param APISERVER_HOSTPORT="#{ip}:8443" https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-system-openshift.yml
 
@@ -24,8 +27,8 @@ echo
 echo Or you can watch in the OpenShift console via:
 echo   minishift console
 echo
-echo When the pods are all running please click on the following URLs in your browser, then ADVANCED, then click the URL at the bottom
-echo To approve the certs
+echo When the pods are all running please click on the following URLs in your browser
+echo then ADVANCED, then click the URL at the bottom to approve the certs:
 echo
 echo   https://keycloak-fabric8.#{ip}.#{routing-suffix}/
 echo   https://wit-fabric8.#{ip}.#{routing-suffix}/api/status
@@ -35,5 +38,3 @@ echo
 echo
 echo Then you should be able the open the fabric8 console here:
 echo   https://fabric8-fabric8.#{ip}.#{routing-suffix}/
-
-
--- a/addons/fabric8/fabric8.addon
+++ b/addons/fabric8/fabric8.addon
@@ -1,38 +1,37 @@
 # Name: fabric8
 # Description: Installs fabric8 locally on Minishift
+# Required-Vars: GITHUB_OAUTH_CLIENT_ID,GITHUB_OAUTH_CLIENT_SECRET
+
+echo NOTE: To delete this addon, execute:
+echo   oc delete project/fabric8 project/developer oauthclient/fabric8-online-platform
 
 echo "Enabling CORS"
 openshift ex config patch /var/lib/origin/openshift.local.config/master/master-config.yaml --patch='{"corsAllowedOrigins": [".*"]}'
 docker restart origin
 
+# TODO how best to wait for openshift to be back?
+sleep 10
+
+oc adm new-project developer
+oc adm new-project fabric8
+
+echo Installing the Fabric8 System using GitHub Client ID #{GITHUB_OAUTH_CLIENT_SECRET}
+oc new-app --param APISERVER_HOSTPORT="#{ip}:8443" --param GITHUB_OAUTH_CLIENT_SECRET=#{GITHUB_OAUTH_CLIENT_SECRET} --param GITHUB_OAUTH_CLIENT_ID=#{GITHUB_OAUTH_CLIENT_ID} https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-system-openshift.yml -n fabric8
+
+oc adm policy add-role-to-user admin developer -n developer
+oc adm policy add-role-to-user admin developer -n fabric8
+
 echo Adding the OAuthClient and cluster-admin role to the init-tenant service account
-oc --as system:admin new-app --param URL="https://keycloak-fabric8.#{ip}.#{routing-suffix}/auth/realms/fabric8/broker/openshift-v3/endpoint" https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-oauthclient-template.yml
+oc --as system:admin new-app --param URL=keycloak-fabric8.#{routing-suffix}/auth/realms/fabric8/broker/openshift-v3/endpoint https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-oauthclient-template.yml
 oc --as system:admin adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:fabric8:init-tenant
 
-oc new-project developer
-oc new-project fabric8
-
-echo Installing the Fabric8 System using GitHub Client ID ${GITHUB_OAUTH_CLIENT_SECRET}
-# TODO lets uncomment when we figure out parameter passing!
-#oc new-app --param APISERVER_HOSTPORT="#{ip}:8443" --param GITHUB_OAUTH_CLIENT_SECRET=${GITHUB_OAUTH_CLIENT_SECRET} --param GITHUB_OAUTH_CLIENT_ID=${GITHUB_OAUTH_CLIENT_ID} https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-system-openshift.yml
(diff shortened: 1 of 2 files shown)


budhrg:
add to whitelist

addons/fabric8/fabric8.addon
```diff
@@ -0,0 +1,39 @@
+# Name: fabric8
+# Description: Installs fabric8 locally on MiniShift
```
budhrg: @jstrachan , could you rename it to `Minishift` with which we are branding it?

addons/fabric8/fabric8.addon
```diff
@@ -0,0 +1,39 @@
...
+oc new-app --param URL="https://keycloak-fabric8.#{ip}.#{routing-suffix}/auth/realms/fabric8/broker/openshift-v3/endpoint" https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-oauthclient-template.yml
+oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:fabric8:init-tenant
+
+oc login -u developer -p developer
+
+oc new-project developer
+oc new-project fabric8
+
+echo Installing the Fabric8 System using GitHub Client ID ${GITHUB_OAUTH_CLIENT_SECRET}
+#oc new-app --param APISERVER_HOSTPORT="#{ip}:8443" --param GITHUB_OAUTH_CLIENT_SECRET=${GITHUB_OAUTH_CLIENT_SECRET} --param GITHUB_OAUTH_CLIENT_ID=${GITHUB_OAUTH_CLIENT_ID} https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-system-openshift.yml
```
budhrg: So this is just a comment or you forgot to uncomment it :smile:

addons/fabric8/fabric8.addon
```diff
@@ -0,0 +1,39 @@
...
+echo
+echo   https://keycloak-fabric8.#{ip}.#{routing-suffix}/
+echo   https://wit-fabric8.#{ip}.#{routing-suffix}/api/status
+echo   https://forge-fabric8.#{ip}.#{routing-suffix}/forge/version
+echo   https://fabric8-fabric8.#{ip}.#{routing-suffix}/
+echo
+echo
+echo Then you should be able the open the fabric8 console here:
+echo   https://fabric8-fabric8.#{ip}.#{routing-suffix}/
+
```
budhrg: Extra line here.

addons/fabric8/fabric8.addon
```diff
@@ -0,0 +1,39 @@
...
+echo Please wait while the pods all startup!
+echo
+echo To watch this happening you can type:
+echo   oc get pod -l provider=fabric8 -w
+echo
+echo Or you can watch in the OpenShift console via:
+echo   minishift console
+echo
+echo When the pods are all running please click on the following URLs in your browser, then ADVANCED, then click the URL at the bottom
```
budhrg: May be split into multiple lines for readability.

```
echo When the pods are all running, please click on following URLS in your browser, then ADVANCED
echo And then click the URL at the bottom to approve the certs
```

budhrg:
Not sure why CI are failing here.

budhrg:
retest this please

budhrg:
@jstrachan , we have some valid unit test failures reported by CIs

```
00:03:19.649 --- FAIL: Test_create_addon_manager (0.00s)
00:03:19.649 	addon_manager_test.go:70: Unexpected number of addons. Expected 4, got 5
```

Please fix by updating it to 5 here https://github.com/minishift/minishift/blob/master/pkg/minishift/addon/manager/addon_manager_test.go#L67 ?

addons/fabric8/fabric8.addon
jstrachan: I'll uncomment it (replacing the next line) once we figure out how to add parameters ;)

addons/fabric8/fabric8.addon
jstrachan: good point

addons/fabric8/fabric8.addon
```diff
@@ -1,5 +1,7 @@
 # Name: fabric8
-# Description: Installs fabric8 locally on MiniShift
+# Description: Installs fabric8 locally on Minishift
+
+# TODO enable CORS!
```
budhrg: :+1:

addons/fabric8/fabric8.addon
```diff
@@ -24,8 +27,8 @@ echo
 echo Or you can watch in the OpenShift console via:
 echo   minishift console
 echo
-echo When the pods are all running please click on the following URLs in your browser, then ADVANCED, then click the URL at the bottom
-echo To approve the certs
+echo When the pods are all running please click on the following URLs in your browser
```
budhrg: :+1:

budhrg:
Also, we are following below pattern for commit message: 
```
Issue #1043 Add fabric8 addon 
```

It would be great to have for keeping consistency in commit messages :smile:

jstrachan:
@budhrg ah cool, had missed that ;)

addons/fabric8/fabric8.addon
```diff
@@ -0,0 +1,40 @@
+# Name: fabric8
+# Description: Installs fabric8 locally on Minishift
+
+# TODO enable CORS!
+
+oc login -u system:admin
```
LalatenduMohanty: -1. I would suggest `oc --as system:admin`

addons/fabric8/fabric8.addon
```diff
@@ -0,0 +1,40 @@
...
+
+echo Adding the OAuthClient and cluster-admin role to the init-tenant service account
+oc new-app --param URL="https://keycloak-fabric8.#{ip}.#{routing-suffix}/auth/realms/fabric8/broker/openshift-v3/endpoint" https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-oauthclient-template.yml
+oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:fabric8:init-tenant
+
+oc login -u developer -p developer
```
LalatenduMohanty: This will not be required if you use `oc --as system:admin`

addons/fabric8/fabric8.addon
jstrachan: ah great, thanks for the heads up. You learn something new about `oc` every day it seems ;)

jstrachan: @LalatenduMohanty hmm, `oc --as system:admin` doesn't seem to work
```
$ oc --as system:admin  get oauthclient
error: error processing template "fabric8/fabric8-oauthclient-template": User "developer" cannot "impersonate" "systemusers.user.openshift.io" with name "system:admin" in project ""

or when trying to use it instead of `oc login -u system:admin` I now get:
```
oc --as system:admin adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:fabric8:init-tenant
Error from server (Forbidden): User "developer" cannot "impersonate" "systemusers.user.openshift.io" with name "system:admin" in project ""
```

Comment (jstrachan) [inline addons/fabric8/fabric8.addon]: unless this is a new issue with 3.6.x?

Comment (jstrachan) [inline addons/fabric8/fabric8.addon]: Ah yeah - lemme split that off as a separate issue: https://github.com/minishift/minishift/issues/1053

**Comment by LalatenduMohanty** [comment]
@jstrachan I came across https://github.com/sabre1041/cdk-minishift-utils/blob/master/addons/cors-allowed-origins-wildcard/cors-allowed-origins-wildcard.addon

**Comment by hferentschik** [comment]
> But I think we need a better way to achieve the same.

Why? This is how you would do if you script it via a shell as well. From the CLI we offer a shortcut for this via `minishift openshift patch`, but under the hood we do exactly that.

**Comment by LalatenduMohanty** [comment]
> Why?
Somehow I feel the sed commands are not for every one who wants to change the master config. There should be some better/easier way to do it.

**Comment by gbraad** [comment]
> There should be some better/easier way to do it.

+1 the added value of using Minishift (for development purpose)...

File: addons/fabric8/fabric8.addon [hunk 9]
```diff
@@ -0,0 +1,38 @@
...
+echo To watch this happening you can type:
+echo   oc get pod -l provider=fabric8 -w
+echo
+echo Or you can watch in the OpenShift console via:
+echo   minishift console
+echo
+echo When the pods are all running please click on the following URLs in your browser
+echo then ADVANCED, then click the URL at the bottom to approve the certs:
+echo
+echo   https://keycloak-fabric8.#{ip}.#{routing-suffix}/
```
LalatenduMohanty: With #1056 routing-suffix will have ip+nip.io cc @hferentschik

jorgemoralespou:
@jstrachan I have taken a stab at your template. Here's a more functional version. I give some comments right after:

````
# Name: fabric8
# Description: Installs fabric8 locally on Minishift

echo NOTE: To delete this addon, execute:
echo   oc delete project/fabric8 project/developer oauthclient/fabric8-online-platform

echo "Enabling CORS"
openshift ex config patch /var/lib/origin/openshift.local.config/master/master-config.yaml --patch='{"corsAllowedOrigins": [".*"]}'
docker restart origin

sleep 3

echo Adding the OAuthClient and cluster-admin role to the init-tenant service account
oc new-app --param URL="https://keycloak-fabric8.#{routing-suffix}/auth/realms/fabric8/broker/openshift-v3/endpoint" https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-oauthclient-template.yml
oc adm policy add-cluster-role-to-user cluster-admin system:serviceaccount:fabric8:init-tenant

oc adm new-project developer
oc adm new-project fabric8

echo Installing the Fabric8 System using GitHub Client ID ${GITHUB_OAUTH_CLIENT_SECRET}
# TODO lets uncomment when we figure out parameter passing!
#oc new-app --param APISERVER_HOSTPORT="#{ip}:8443" --param GITHUB_OAUTH_CLIENT_SECRET=${GITHUB_OAUTH_CLIENT_SECRET} --param GITHUB_OAUTH_CLIENT_ID=${GITHUB_OAUTH_CLIENT_ID} https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-system-openshift.yml -n fabric8
oc new-app --param APISERVER_HOSTPORT="#{ip}:8443" https://raw.githubusercontent.com/fabric8io/fabric8-resources/master/fabric8-system-openshift.yml -n fabric8

oc adm policy add-role-to-user admin developer -n developer
oc adm policy add-role-to-user admin developer -n fabric8

echo Please wait while the pods all startup!
echo -
echo To watch this happening you can type:
echo   oc get pod -l provider=fabric8 -n fabric8 -w
echo -
echo Or you can watch in the OpenShift console via:
echo   minishift console
echo -
echo When the pods are all running please click on the following URLs in your browser
echo then ADVANCED, then click the URL at the bottom to approve the certs:
echo -
echo   https://keycloak-fabric8.#{routing-suffix}/
echo   https://wit-fabric8.#{routing-suffix}/api/status
echo   https://forge-fabric8.#{routing-suffix}/forge/version
echo   https://fabric8-fabric8.#{routing-suffix}/
echo -
echo -
echo Then you should be able the open the fabric8 console here:
echo   https://fabric8-fabric8.#{routing-suffix}/
````

cc/ @hferentschik @LalatenduMohanty 

Comments:
- Addons don't provide a way to uninstall. When an addon fails, if there's some stuff that needs to be removed for the next try to not fail, you need to know what or there should be an "addon remove" command. Not existent now.
- For this reason, addons should need to be "idempotent", but this is really difficult. If a resource already exists, oc will fail. If you remove it in the addon, but it didn't exist, oc will fail. I would recommend to provide a way to not fail if the command fails. "oc-not-fail" or similar. As I can not just ignore the errors 2> /dev/null
- Echo with empty string does not work. This is a BUG, and should be fixed. I have for now, just printed a "-" so that the output is nicer than what gets printed by default which is "echo".
-  I have used "oc adm new-project" so that the user does not change context to that project. Not really needed when bootstrapping project as admin to be used by a user.
- I have added a admin role to the user "developer" so he owns both projects. But he's not logged in by default on any of those projects. Trying to be non-intrusive with addons applying.
- I have added a sleep to the docker restart, but it would be good to provide a way to validate that origin is up and running.

Hope this helps,

**Comment by hferentschik** [comment]
> Addons don't provide a way to uninstall. When an addon fails, if there's some stuff that needs to be removed for the next try to not fail, you need to know what or there should be an "addon remove" command. Not existent now.

I think you will have to take the manual approach or just delete the VM and start over. I guess you see which step fails and it is up to you to decide/know what to do. I don't think a 'addon remove' is feasible, at least not in a sense that we would know what to do to get back to the original state. 

What I could think of is a 'addon remove' where in an add-on package one can also define an uninstall add-on file which would be called on 'addon remove foo'. It would be up to the user to provide appropriate uninstall action. Not sure whether this would be useful?

> Echo with empty string does not work. This is a BUG, and should be fixed. I have for now, just printed a "-" so that the output is nicer than what gets printed by default which is "echo".

See #1064

**Comment by jorgemoralespou** [comment]
@hferentschik  indeed this is what I have in my wrapper for the same reason. It does make sense.

But still, I would think that I would like to execute oc commands and obviate if they fail and keep going.

**Comment by hferentschik** [comment]
@jstrachan, you can grab the 1.2.0 release of Minishift - https://github.com/minishift/minishift/releases/tag/v1.2.0. That should solve your interpolation issue. If you are already using 1.1.0, you can try `minishift update` to get the latest version. 

Assuming interpolation is fixed and given the tips from @jorgemoralespou, what are the outstanding issues with this add-on?

**Comment by jstrachan** [comment]
@hferentschik @jorgemoralespou thanks folks!

@hferentschik things are looking good, we just need env vars (or something like it) so we can pass in/configure 2 strings otherwise we're pretty much there now I think: https://github.com/minishift/minishift/issues/1045

**Comment by budhrg** [comment]
> things are looking good, we just need env vars (or something like it) so we can pass in/configure 2 strings otherwise we're pretty much there now I think: #1045

I have tested this last week in the cooperation with @jstrachan . 
Except for env vars (GITHUB_OAUTH_CLIENT_*) most of the things are working. 

Are you also updating addon with `oc login -u developer -u developer` ? otherwise the commands will run through system as we notice.

**Comment by jstrachan** [comment]
I'm a bit snowed in stuff until Friday but I can retest the addon with the update from @jorgemoralespou on minishift 1.2.0  on Friday to double check everything's all looking good (hacking around the lack of environment variables for now)

**Comment by jstrachan** [comment]
OK I've taken the latest from @jorgemoralespou and got it working with the latest greatest minishift + fabric8 along with using the new environment variable stuff.

I first started minishift via:
```
minishift start --vm-driver=xhyve --memory=5000 --cpus=5 --disk-size=50g
```
Then after installing a local copy of the fabric8 addon I was able to apply it via:
```
minishift addon apply --addon-env GITHUB_OAUTH_CLIENT_ID=$GITHUB_OAUTH_CLIENT_ID --addon-env GITHUB_OAUTH_CLIENT_SECRET=$GITHUB_OAUTH_CLIENT_SECRET fabric8
```

hferentschik:
@jstrachan, sorry for the long turnaround time. What is the state on this? Is it working now. I recall you were saying that you need the ability to interpolate env variables. This has been possible since 1.3.0. Have you tried? 

I suggest to close this pull request and create a new pull request in https://github.com/minishift/minishift-addons. This is a dedicated repository for collecting community provided add-ons.

jstrachan:
@hferentschik I've tested it on 1.3.0 with the new env var interpolation and its all working fine. I was  hoping we could merge it ;)

hferentschik:
> I've tested it on 1.3.0 with the new env var interpolation and its all working fine.

Cool

 > I was hoping we could merge it ;)

+1, just into a different repo.

jstrachan:
PR here: https://github.com/minishift/minishift-addons/pull/5

Issue #426, opened 2017-02-17
Document CI platforms being used by Minishift


budhrg:
See here https://github.com/minishift/minishift/issues/415#issuecomment-280667400

thatdocslady:
@budhrg it looks like the comment you referenced discusses the CentOS CI doc which now exists (awesome!). Can you provide a bit more details about the other CI systems, not necessarily generic info but how we incorporate them in our specific build flow (I will include links to generic Travis/AppVeyor docs too).

budhrg:
@thatdocslady http://bit.ly/minishiftci is the slides I prepare for one of my talk. Let me know if you need more help.

Slides definitely missing steps to setup any particular CI.

I can add steps to setup CentOS CI and for other CI @praveenkumar and @hferentschik can add those details.

Preeticp:
The existing docs for CentOS CI are here: https://github.com/minishift/minishift/blob/master/docs/source/developing/ci.adoc
@hferentschik, @praveenkumar request you to add initial steps for other CI such as appveyor and circle CI as mentioned by @budhrg so that I can add it to the docs.

praveenkumar:
@Preeticp For appveyor and CircleCI

```
You can find the appveyor CI job for Minishift here <https://ci.appveyor.com/project/hferentschik/minishift-o61ou>

The artifacts of a successful pull request build can be found at https://ci.appveyor.com/project/hferentschik/minishift-o61ou/build/<build_id>/aritfacts

Artifacts of a successful master build can be found at https://ci.appveyor.com/project/hferentschik/minishift-o61ou/branch/master/artifacts

```

```
You can find the circle-CI job for Minishift here <https://circleci.com/gh/minishift/minishift>

The artifacts of a successful pull request build can be found at https://circleci.com/gh/minishift/minishift/<build_id>#artifacts/  under below directory structure.
distiller/
    src/
        github.com/
            minishift/
                minishift/
                    out/

Artifacts of a successful master build can be found at https://circleci.com/gh/minishift/minishift/tree/master
```

budhrg:
Also, @thatdocslady @Preeticp , also, it would be great if we mention which CI is used for testing which bits.

Like, CentOS CI is used specifically to test integration tests, unit tests, code format tests and cross build tests in Linux (CentOS). Other CI's AFAIR are:
- Circle CI : Used for testing unit tests, code format tests and cross build tests in macOS
[...]

PR #1048, opened 2017-06-20
Issue #691 Clean up doc set

thatdocslady:
Fixes issue #691 

This PR intends to group together style and formatting conventions based on recent work on #618.

Currently the PR contains only the topics in the Getting Started section, as well as the topic map.

Question for @hferentschik @LalatenduMohanty: it seems to take about a day for fixing up a category (i.e. Getting Started) and testing everything. Should I create multiple commits, each with its own category fixes? I am trying to avoid merge/push conflicts if other people are working on the same topics.

### DIFF
--- a/docs/source/contributing/developing.adoc
+++ b/docs/source/contributing/developing.adoc
@@ -22,12 +22,12 @@ NOTE: You should be able to develop Minishift on any operating system, such as G
 The Windows operating system might require additional steps or have some limitations.
 
 [[set-up-dev-env]]
-== Setting up the development environment
+== Setting Up the Development Environment
 
 [[create-go-workspace]]
-=== Creating the Go workspace
+=== Creating the Go Workspace
 
-We highly recommend to setup a default Go https://golang.org/doc/code.html#Workspaces[workspace].
+We highly recommend to setup a default Go link:https://golang.org/doc/code.html#Workspaces[workspace].
 Even though it might require an adjustment in your work processes, the goal is to have a single workspace for all Go development.
 
 . Create the following directories in the base directory you want to use, such as *_$HOME/work_*:
@@ -41,8 +41,8 @@ Contains package objects
 /src::
 Contains Go source files
 
-. Add the path of the root workspace directory to the `GOPATH` environment
-variable. For example:
+. Add the path of the root workspace directory to the `GOPATH` environment variable.
+For example:
 +
 ----
 $ export GOPATH=$HOME/work
@@ -57,9 +57,9 @@ $ export PATH=$PATH:$GOPATH/bin
 NOTE: On Windows operating systems, you use the UI or use `setx` to set the environment variables.
 
 [[cloning-repository]]
-=== Cloning the repository
+=== Cloning the Repository
 
-Get the Minishift sources from GitHub.
+Get the Minishift sources from GitHub:
 
 ----
 $ cd $GOPATH/src
@@ -70,15 +70,15 @@ $ git clone https://github.com/minishift/minishift.git github.com/minishift/mini
 === Using an IDE
 
 You can use any editor you want.
-However, most of the core maintainers of Minishift use IntelliJ https://www.jetbrains.com/idea/[IDEA] with the latest Go plugin.
-This IDE indexes your whole workspace and allows for easy navigation of the sources, and also integrates with the Go debugger https://github.com/derekparker/delve[Delve].
+However, most of the core maintainers of Minishift use IntelliJ link:https://www.jetbrains.com/idea/[IDEA] with the latest Go plug-in.
+This IDE indexes your whole workspace and allows for easy navigation of the sources, and also integrates with the Go debugger link:https://github.com/derekparker/delve[Delve].
 
-For instructions on setting up IDEA, see http://hadihariri.com/2015/09/30/setting-up-go-on-intellij/[Setting up Go on IntelliJ].
+For instructions on setting up IDEA, see link:http://hadihariri.com/2015/09/30/setting-up-go-on-intellij/[Setting up Go on IntelliJ].
 
 [[manage-dependencies]]
-== Dependency management
+== Dependency Management
 
-Minishift uses https://github.com/Masterminds/glide[Glide] for dependency management.
+Minishift uses link:https://github.com/Masterminds/glide[Glide] for dependency management.
 
 [[install-glide]]
 === Installing Glide
@@ -89,11 +89,11 @@ Before you can use Glide you need to download and install it from GitHub:
 $ go get github.com/Masterminds/glide
 ----
 
-This will install the `glide` binary into *_$GOPATH/bin_*.
-Make sure to use Glide version _0.12.3_ or later.
+This will install the *glide* binary into *_$GOPATH/bin_*.
+Make sure to use Glide version 0.12.3 or later.
 
 [[bootstrap-dependencies]]
-=== Bootstrapping dependencies
+=== Bootstrapping Dependencies
 
 After a clean checkout or after a `make clean`, there won't be a *_vendor_* directory containing the needed Minishift dependencies.
 
@@ -111,18 +111,18 @@ $ glide install -v
 ----
 
 [[update-dependencies]]
-=== Updating dependencies
+=== Updating Dependencies
 
 If your work requires a change to the dependencies, you need to update the Glide configuration.
 
 . Edit *_glide.yaml_* to change the dependencies as needed.
 
-. Delete *_glide.lock_*  and re-create the vendor directory by running `make vendor`.
+. Delete *_glide.lock_* and re-create the vendor directory by running `make vendor`.
  Glide will recognize that there is no lock file and recalculate the required dependencies.
 
-. Check-in the updated *_glide.yaml_* and *_glide.lock_*.
+. Check-in the updated *_glide.yaml_* and *_glide.lock_* files.
 
-. Test that everything still compiles with the new lock file in place: `make clean && make`.
+. Test that everything still compiles with the new lock file in place by running `make clean && make`.
 
 TIP: In some cases the Glide cache located under *_~/.glide/cache_* can get corrupted.
 If you seeing Glide errors during `make vendor`, you can clear the Glide cache via `glide cc`.
@@ -131,10 +131,9 @@ If you seeing Glide errors during `make vendor`, you can clear the Glide cache v
 == Building Minishift
 
 [[build-minishift-binary]]
-=== Building the Minishift binary
+=== Building the Minishift Binary
 
-Run the following command to create a platform-specific binary and copy it
-to *_$GOPATH/bin_*.
+Run the following command to create a platform-specific binary and copy it to *_$GOPATH/bin_*.
 
 ----
 $ make
@@ -143,7 +142,7 @@ $ make
 NOTE: Use `make cross` to cross-compile for other platforms.
 
 [[run-minishift-binary]]
-=== Running the Minishift binary
+=== Running the Minishift Binary
 
 Start the OpenShift cluster with your built minishift binary:
 
@@ -151,8 +150,7 @@ Start the OpenShift cluster with your built minishift binary:
 $ minishift start
 ----
 
-This command will run Minishift from *_$GOPATH/bin/minishift_*, if you set up your Go
-workspace as described in the xref:./developing.adoc#create-go-workspace[Creating the Go workspace] section.
+This command will run Minishift from *_$GOPATH/bin/minishift_*, if you set up your Go workspace as described in the xref:create-go-workspace[Creating the Go workspace] section.
 
 You can also execute the binaries directly from the *_out_* directory of the checkout.
 Depending on your operating system, the binary is in one of the following directories:
@@ -161,12 +159,13 @@ Depending on your operating system, the binary is in one of the following direct
 * *_out/linux-amd64_*
 * *_out/windows-amd64_*
 
-For more minishift commands and flags, see the xref:../command-ref/minishift.adoc#minishift[Minishift command reference] documentation.
+For more Minishift commands and flags, see the xref:../command-ref/minishift.adoc#minishift[Minishift command reference] documentation.
 
 [[unit-tests]]
-=== Running unit tests
+=== Running Unit Tests
 
-Unit tests run on Travis before code is merged. To run tests during the development cycle:
+Unit tests run on *Travis* before the code is merged.
+To run tests during the development cycle:
 
 ----
 $ make test
@@ -194,11 +193,11 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
 For more information about test options, run the `go test --help` command and review the documentation.
 
 [[integration-tests]]
-=== Running integration tests
+=== Running Integration Tests
 
-Integration tests utilize https://github.com/DATA-DOG/godog[`godog`], which uses Gherkin (Cucumber) to define test cases.
-The test cases are defined in *_test/integration/+*+.feature_*.
-Per default, the tests are executed against the binary created by `make build`, that is *_$GOPATH/bin/minishift_*.
+Integration tests utilize link:https://github.com/DATA-DOG/godog[Godog], which uses Gherkin (Cucumber) to define test cases.
+The test cases are defined in *_test/integration/features_*.
+By default, the tests are executed against the binary created by `make build`, that is *_$GOPATH/bin/minishift_*.
 To run the tests, run:
 
 ----
@@ -215,30 +214,29 @@ Additional properties for Godog runner can be specified with `GODOG_OPTS` argume
 The following options are available:
 
 - `tags`: Use `tags` to ensure that scenarios and features containing at least one of the selected tags are executed.
-- `paths`: Use `paths` to define paths to different feature files or folders containing feature files. This can be used to run feature files
-outside of the _test/integration/features_ folder.
-- `format`:  Use `format` to change the format of Godog's output, for example you can use `pretty` format instead of the native `progress`.
+- `paths`: Use `paths` to define paths to different feature files or folders containing feature files.
+This can be used to run feature files outside of the *_test/integration/features_* folder.
+- `format`:  Use `format` to change the format of Godog's output.
+For example, you can use `pretty` format instead of the native `progress`.
 - `stop-on-failure`: Set `stop-on-failure` to true to stop integration tests on failure.
 - `no-colors`: Set `no-colors` to true to disable ansi colors of Godog's output.
 - `definitions`: Set `definitions` to true to print all available step definitions.
 
-For example, to run integration tests on two specific feature files, using only `@basic` and `@openshift` tags and without ansi colors,
-the following command can be used:
+For example, to run integration tests on two specific feature files using only `@basic` and `@openshift` tags and without ansi colors, the following command can be used:
 
 ----
 $ make integration GODOG_OPTS="-paths ~/tests/custom.feature,~/my.feature -tags basic,openshift -no-colors true"
 ----
 
-NOTE: When multiple values are used for options in `GODOG_OPTS`, then they have to be separated by a comma, without whitespace.
-While `-tags basic,openshift` will be parsed properly by make, `-tags basic, openshift` will
-result in only `@basic` being used.
+NOTE: When multiple values are used for options in `GODOG_OPTS`, then they have to be separated by a comma without whitespace.
+While `-tags basic,openshift` will be parsed properly by make, `-tags basic, openshift` will result in only `@basic` being used.
 
-[[format-source]]
-=== Formatting the source
+[[format-source-code]]
+=== Formatting the Source Code
 
-Minishift adheres to the Go formatting https://golang.org/doc/effective_go.html#formatting[guidelines].
-Code with incorrect formatting will fail the CI builds. You can check whether any of your
-files violate the guidelines with the following command:
+Minishift adheres to the Go link:https://golang.org/doc/effective_go.html#formatting[formatting guidelines].
+Code with incorrect formatting will fail the CI builds.
+You can check whether any of your files violate the guidelines with the following command:
 
 ----
 $ make fmtcheck
@@ -262,6 +260,6 @@ $ make clean
 [[godoc]]
 == Godoc
 
-When developing on Minishift it is encouraged to use link:https://godoc.org/golang.org/x/tools/cmd/godoc[Godoc] to document the source code.
+When developing Minishift, it is encouraged to use link:https://godoc.org/golang.org/x/tools/cmd/godoc[Godoc] to document the source code.
 You can find guidelines on how to use godoc in link:https://blog.golang.org/godoc-documenting-go-code[this blog post].
-You can browse the Minishift godoc documentation under link:https://godoc.org/github.com/minishift/minishift[https://godoc.org/github.com/minishift/minishift].
+You can browse the Minishift Godoc documentation under link:https://godoc.org/github.com/minishift/minishift[https://godoc.org/github.com/minishift/minishift].
--- a/docs/source/getting-started/index.adoc
+++ b/docs/source/getting-started/index.adoc
@@ -1,12 +1,11 @@
 [[getting-started]]
-= Getting started with Minishift
+= Getting Started with Minishift
 :icons:
 
-This section contains information about installing Minishift, quickstart steps, and sample
-application deployment.
+This section contains information about installing Minishift, quickstart steps, and sample application deployment.
 
-- link:../getting-started/installing{outfilesuffix}[Installing Minishift]
-- link:../getting-started/quickstart{outfilesuffix}[Minishift Quickstart]
-- link:../getting-started/setting-up-driver-plugin{outfilesuffix}[Setting up the driver plugin]
-- link:../getting-started/updating{outfilesuffix}[Updating Minishift]
-- link:../getting-started/uninstalling{outfilesuffix}[Uninstalling Minishift]
+- xref:../getting-started/installing.adoc#installing-minishift[Installing Minishift]
+- xref:../getting-started/quickstart.adoc#quickstart[Minishift Quickstart]
+- xref:../getting-started/setting-up-driver-plugin.adoc#set-up-driver-plugin[Setting Up the Driver Plug-in]
+- xref:../getting-started/updating.adoc#update-minishift[Updating Minishift]
+- xref:../getting-started/uninstalling.adoc#uninstall-minishift[Uninstalling Minishift]
--- a/docs/source/getting-started/quickstart.adoc
+++ b/docs/source/getting-started/quickstart.adoc
@@ -1,7 +1,7 @@
 include::variables.adoc[]
 
 [[quickstart]]
-= Quickstart
+= Minishift Quickstart
 :icons:
 :toc: macro
 :toc-title:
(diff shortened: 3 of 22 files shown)


hferentschik:
> Should I create multiple commits, each with its own category fixes?

Sure multiple commits is fine. One commit per category is fine. We can squash in the end (if we want).

docs/source/getting-started/index.adoc
```diff
@@ -5,8 +5,8 @@
 This section contains information about installing Minishift, quickstart steps, and sample
```
hferentschik: What's about we take care of "line per sentence" as well. A fair bit of the docs are already written this way. It would be nice to align the rest as well.

thatdocslady: Sure, I just realized now that we never settled on this, but the openshift folks said it doesn't matter to publishing so I can fix it here. We should also add this to our own doc guidelines probably, no?

thatdocslady:
Ok I powered through and managed to fix up everything (tomorrow I have a late start and I want to try and get a review going ASAP). Removing the WIP because it's now ready for review.

A few notes to considering during the review:

- There are 4 commits, one for each category/directory. The only exception is that the Getting Started commit also has the topic map, I couldn't figure out how to extract it after committing it, and I have a feeling it's not a huge deal?
- There's no command reference topic for `minishift oc-env`, but there are some references to it in the docs. Not sure what happened to it or whether we replaced it with something else?
- Legally we are supposed to use "container registry" and not "docker registry" when talking about downstream offerings, but I'm not sure this is relevant for upstream/community editions. Thoughts?
- I didn't touch the CI topic in Contributing because there's an open PR for issue #426, will fix this later after that PR is merged.
- While I was fixing the style and formatting, I also tweaked a few terms that needed standardizing. Please let me know if you have questions about any of these (i.e. changed "plugin" to "plug-in").

hferentschik:
> There are 4 commits, one for each category/directory. The only exception is that the Getting Started commit also has the topic map, I couldn't figure out how to extract it after committing it, and I have a feeling it's not a huge deal?

It is not. In fact I think when we apply this we will squash anyways. 

> There's no command reference topic for minishift oc-env, but there are some references to it in the docs. Not sure what happened to it or whether we replaced it with something else?

I would think you have not re-generated the command ref files since _oc-env_ got added. 

    $ make clean clean_docs 
    $ make gen_docs

After that you should see them as well.

> Legally we are supposed to use "container registry" and not "docker registry" when talking about downstream offerings, but I'm not sure this is relevant for upstream/community editions. Thoughts?

I am fine to stick with Docker registry for now.

> I didn't touch the CI topic in Contributing because there's an open PR for issue #426, will fix this later after that PR is merged.

+1 In fact I would assume that won't be necessary. @Preeticp is re-writing the whole thing. I would think it will in the end adhere to the standards as well.

> While I was fixing the style and formatting, I also tweaked a few terms that needed standardizing. Please let me know if you have questions about any of these (i.e. changed "plugin" to "plug-in").

Sounds good.

docs/source/contributing/developing.adoc
```diff
@@ -111,18 +111,18 @@ $ glide install -v
...
 If your work requires a change to the dependencies, you need to update the Glide configuration.
 
 . Edit *_glide.yaml_* to change the dependencies as needed.
 
-. Delete *_glide.lock_*  and re-create the vendor directory by running `make vendor`.
+. Delete *_glide.lock_* and re-create the vendor directory by running `make vendor`.
  Glide will recognize that there is no lock file and recalculate the required dependencies.
 
-. Check-in the updated *_glide.yaml_* and *_glide.lock_*.
+. Check-in the updated *_glide.yaml_* and *_glide.lock_* files.
```
hferentschik: _files_ is imo superfluous here. It just makes it sounds odd, even though it is correct. I am saying "I check-in start.go" not "I am checking in the start.go file". That's implicit.

docs/source/contributing/developing.adoc
```diff
@@ -194,11 +193,11 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
 
 [[integration-tests]]
-=== Running integration tests
+=== Running Integration Tests
 
-Integration tests utilize https://github.com/DATA-DOG/godog[`godog`], which uses Gherkin (Cucumber) to define test cases.
-The test cases are defined in *_test/integration/+*+.feature_*.
-Per default, the tests are executed against the binary created by `make build`, that is *_$GOPATH/bin/minishift_*.
+Integration tests utilize link:https://github.com/DATA-DOG/godog[Godog], which uses Gherkin (Cucumber) to define test cases.
+The test cases are defined in *_test/integration/features_*.
```
hferentschik: That's not correct anymore. Why did you change this? There is no directory _test/integration/features_, there is a __test/integration_ and it contains several files with the feature suffix. This is what was expressed in the existing expression. Also, it needs the extra '+' since one needs to escape the '*'. Let's change this back please.

docs/source/getting-started/quickstart.adoc
```diff
@@ -50,17 +45,15 @@ Starting local OpenShift cluster using 'kvm' hypervisor...
...
+- The IP is dynamically generated for each OpenShift cluster.
+To check the IP, run the `minishift ip` command.
+- By default, Minishift uses the driver most relevant to the host OS.
+To use a different driver, set the `--vm-driver` flag in `minishift start`.
+For example, to use VirtualBox instead of KVM on GNU/Linux operating systems, run `minishift start --vm-driver=virtualbox`.
+For more information about `minishift start` options, see the xref:../command-ref/minishift_start.adoc#minishift-start[`minishift start`] command.
 ====
 
-. Use link:../command-ref/minishift_oc-env{outfilesuffix}[`minishift oc-env`] to display the command you need to type into your shell in order to add the _oc_ binary to your PATH.
+. Use xref:../command-ref/minishift_oc-env.adoc#minishift_oc-env[`minishift oc-env`] to display the command you need to type into your shell in order to add the *oc* binary to your `PATH` environment variable.
```
hferentschik: should be \`oc\` instead of \*oc\*. It's a binary.

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -1,25 +1,23 @@
...
+= Setting Up the Driver Plug-in
 :icons:
 :toc: macro
 :toc-title:
 :toclevels: 1
 
 toc::[]
 
-[[hypervisor-drivers-overview]]
+[[setup-driver-plugin-overview]]
```
hferentschik: good catch. There is no need so to make the overview anchor unique. 'overview' will just work fine.

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -108,25 +105,23 @@ $ sudo chown root:wheel $(brew --prefix)/opt/docker-machine-driver-xhyve/bin/doc
...
-`/usr/local/bin` is most likely a good choice, since it is the default installation
-directory for Docker Machine binaries.
+To manually install the xhyve driver, you need to download and install the *docker-machine-driver-xhyve* binary and place it in a directory which is on your `PATH` environment variable.
+The directory *_/usr/local/bin_* is most likely a good choice, since it is the default installation directory for Docker Machine binaries.
 
-The following steps explain the installation of the `docker-machine-driver-xhyve` binary to the `/usr/local/bin/` directory:
+The following steps explain the installation of the *docker-machine-driver-xhyve* binary to the *_/usr/local/bin/_* directory:
 
-. Download the `docker-machine-driver-xhyve` binary using:
+. Download the *docker-machine-driver-xhyve* binary using:
```
hferentschik: back-ticks, we are talking about a binary here.

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -108,25 +105,23 @@ $ sudo chown root:wheel $(brew --prefix)/opt/docker-machine-driver-xhyve/bin/doc
...
 +
 ----
 $ sudo curl -L  https://github.com/zchee/docker-machine-driver-xhyve/releases/download/v0.3.1/docker-machine-driver-xhyve -o /usr/local/bin/docker-machine-driver-xhyve
 ----
 
-. Enable root access for the `docker-machine-driver-xhyve` binary and add it to the default wheel group:
+. Enable root access for the *docker-machine-driver-xhyve* binary and add it to the default wheel group:
```
hferentschik: Same as above. Why are you changing the formatting here?

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -60,13 +57,13 @@ $ newgrp libvirtd
 [[kvm-driver-fedora]]
 === On Fedora
 
-.  Install `libvirt` and `qemu-kvm` on your system:
+.  Install *libvirt* and *qemu-kvm* on your system:
```
hferentschik: hmm, not sure whether one should consider libvirt and qemu-kvm as actual binaries.

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -140,9 +135,9 @@ $ sudo chmod u+s /usr/local/bin/docker-machine-driver-xhyve
 
 [NOTE]
 ====
-The downloaded `docker-machine-driver-xhyve` binaries are compiled against a specific
-version of OS X. It is possible that the driver fails to work after an OS X version
-upgrade. In this case you can try to compile the driver from source:
+The downloaded *docker-machine-driver-xhyve* binaries are compiled against a specific version of OS X.
```
hferentschik: Ok, let's align also how we refer to OS X. According to the latest, it should be macOS. I believe we might use it once or twice, but as we see here, not across the board. Let's do a find&replace of OS X and replace it with macOS.

docs/source/openshift/openshift-client-binary.adoc
```diff
@@ -12,12 +12,13 @@ toc::[]
...
 
-The `minishift start` command creates an OpenShift instance using the https://github.com/openshift/origin/blob/master/docs/cluster_up_down.md[cluster up] approach.
+The `minishift start` command creates an OpenShift cluster using the link:https://github.com/openshift/origin/blob/master/docs/cluster_up_down.md[cluster up] approach.
+For this purpose it copies the *oc* binary onto your host.
 
-For this purpose it copies the `oc` binary onto your host.
-It is located in the  _~/.minishift/cache/oc/{openshift-version}_ directory, provided you use Minishift's default version of OpenShift.
-You can add this binary to your `PATH` using xref:../command-ref/minishift_oc-env.adoc#minishift-oc-env[`minishift oc-env`] which displays the command you need to type into your shell.
-The output of `oc-env` differs depending on the operating system and the shell type.
+The *oc* binary is located in the  *_~/.minishift/cache/oc/{openshift-version}_* directory, assuming that you use Minishift's default version of OpenShift.
```
hferentschik: backticks for _oc_

docs/source/openshift/openshift-client-binary.adoc
```diff
@@ -27,26 +28,26 @@ export PATH="/Users/john/.minishift/cache/oc/v1.5.0:$PATH"
 ----
 
 [[minishift-oc-context]]
-== Minishift oc context
+== Minishift oc Context
 
-As part of the `minishift start` command there is also a `minishift` https://docs.openshift.org/latest/cli_reference/manage_cli_profiles.html[oc context] created.
-This context contains the configuration to communicate with your OpenShift cluster.
+As a part of the `minishift start` command, a `minishift` context is also created.
+This context contains the configuration to communicate with your OpenShift cluster, and is similar to an OpenShift https://docs.openshift.org/latest/cli_reference/manage_cli_profiles.html[CLI profile].
```
hferentschik: "and is similar"? It is not similar to a CLI profile, it **is** a CLI profile.

docs/source/openshift/openshift-client-binary.adoc
```diff
@@ -27,26 +28,26 @@ export PATH="/Users/john/.minishift/cache/oc/v1.5.0:$PATH"
 ----
 
 [[minishift-oc-context]]
-== Minishift oc context
+== Minishift oc Context
```
hferentschik: Not really happy about the capitalisation here.

docs/source/openshift/openshift-client-binary.adoc
```diff
@@ -109,36 +108,29 @@ $ minishift logs
...
 
-For details about the `view` command, see the link:../command-ref/minishift_openshift_config_view{outfilesuffix}[synopsis]
-command reference.
+For details about the `view` sub-command, see the xref:../command-ref/minishift_openshift_config_view.adoc#minishift-openshift-config-view[minishift openshift config view] synopsis.
 
 NOTE: After you update the OpenShift configuration, OpenShift will transparently restart.
 
 [[example-config-cors]]
-=== Example: Configuring cross-origin resource sharing
+.Example: Configuring cross-origin resource sharing
```
hferentschik: This makes the entry disappear from the TOC and makes this example hard to find. I think it is better to keep these as sections. It gives us entries in the TOC and I also feel it is a bit too much content for the '.Example' form.

docs/source/openshift/openshift-client-binary.adoc
```diff
@@ -148,42 +140,37 @@ $  minishift openshift config set --patch '{"corsAllowedOrigins": [".*"]}'
...
+----
 $ minishift openshift config set --patch "{\"corsAllowedOrigins\": [\".*\"]}"
-....
+----
 
 ====
 
 [[example-change-openshift-routing-suffix]]
-=== Example: Changing the OpenShift routing suffix
+.Example: Changing the OpenShift routing suffix
```
hferentschik: Same as above.

docs/source/openshift/openshift-docker-registry.adoc
```diff
@@ -11,37 +11,42 @@ toc::[]
...
+== Logging Into the Registry
 
-[[login-to-openshift-docker-registry]]
-== Registry login
-
-. Start Minishift and add the `oc` binary to the PATH. Refer link:../getting-started/quickstart{outfilesuffix}[Minishift quickstart] for details.
-. Make sure your shell is configured to  link:../using/docker-daemon{outfilesuffix}[reuse the Minishift docker daemon].
-. Login to the docker registry
-
+. Start Minishift and add the *oc* binary to the PATH. For a detailed example, see the xref:../getting-started/quickstart.adoc#quickstart[Minishift Quickstart] section.
```
hferentschik: formatting of 'oc'

docs/source/using/addons.adoc
```diff
@@ -31,65 +31,60 @@ oc adm policy add-scc-to-group anyuid system:authenticated
...
-This is similar to running link:../command-ref/minishift_ssh{outfilesuffix}[`minishift ssh`] and
-then executing any command on the VM.
+If the add-on command starts with `ssh`, you can run any command within the Minishift-managed VM.
+This is similar to running xref:../command-ref/minishift_ssh.adoc#minishift-ssh[`minishift ssh`] and then executing any command on the VM.
+For more information about the `minishift ssh` command usage, see xref:../using/managing-minishift.adoc#connecting-with-ssh[Connecting to the Minishift VM with SSH].
 
 oc::
-If the add-on command starts with _oc_, it uses the _oc_ binary that is cached on your host to execute
-the specified `oc` command. This is similar to running `oc --as system:admin ...` from the command-line.
+If the add-on command starts with `oc`, it uses the *oc* binary that is cached on your host to execute the specified `oc` command.
```
hferentschik: I think it formatting should be the other way around. First use of oc \*oc\* (here we are referring to the "text" oc), second use \`oc\` since we are referring to the binary/command

docs/source/using/addons.adoc
```diff
@@ -31,65 +31,60 @@ oc adm policy add-scc-to-group anyuid system:authenticated
...
-link:../command-ref/minishift_docker-env{outfilesuffix}[`minishift docker-env`].
+If the add-on command starts with `docker`, it executes a `docker` command against the Docker daemon within the Minishift VM.
+This is the same daemon on which the single-node OpenShift cluster is running as well.
+This is similar to running `eval $(minishift docker-env)` on your host and then executing any `docker` command.
+See also xref:../command-ref/minishift_docker-env.adoc#minishift-docker-env[`minishift docker-env`].
 
 echo::
-If the add-on command starts with _echo_, the arguments following the _echo_ command are printed
-to the console. This can be used to provide additional feedback during add-on execution.
+If the add-on command starts with `echo`, the arguments following the `echo` command are printed to the console.
```
hferentschik: \*echo\*, we are not referring to binary or command here.

docs/source/using/addons.adoc
```diff
@@ -31,65 +31,60 @@ oc adm policy add-scc-to-group anyuid system:authenticated
...
 
 Minishift also allows the use of variables as part of the supported commands.
-Variables have the from _#{<variable-name>}_. <<example-addon-routing-variable>>
-shows how the routing suffix can be interpolated into a _openshift_ command
-to create a new certificate as part of securing the OpenShift registry.
+Variables have the format `#{<variable-name>}`.
 
 [[example-addon-routing-variable]]
 .Example: Usage of the routing suffix variable
+
```
hferentschik: I am not a big fan of pulling this into the example. The example is the console listing, not the explaining text which imo should live outside. This applies to the other examples as well.

docs/source/using/addons.adoc
```diff
@@ -100,53 +95,52 @@ $ openshift admin ca create-server-cert \
   --key=/etc/secrets/registry.key
 ----
 
-<<table-supported-addon-variables>> shows the currently supported variables
-which are available for interpolation.
-
 [[table-supported-addon-variables]]
 .Supported add-on variables
+
```
hferentschik: -1, I don't think this is a good form.

docs/source/using/addons.adoc
```diff
@@ -100,53 +95,52 @@ $ openshift admin ca create-server-cert \
...
 
 ----
 $ minishift addons list --verbose=true
 ----
 
-This command prints a list of installed add-ons. You should at least see the _anyuid_ add-on listed.
+This command prints a list of installed add-ons. You should at least see the *anyuid* add-on listed.
 This is an important add-on that allows you to run images that do not use a pre-allocated UID.
-By default, this is not allowed in OpensShift.
+By default, this is not allowed in OpenShift.
```
hferentschik: :-)

docs/source/using/addons.adoc
```diff
@@ -181,48 +176,46 @@ By default, an add-on has the priority 0. Add-ons with a lower priority value ar
 [[example-priority-list]]
 .Example: List command output with explicit priorities
 
+In this example, the *anyuid*, *registry*, and *eap* add-ons are enabled with the respective priorities of 0, 5 and 10.
```
hferentschik: Same as with other tables and examples. This text belongs outside of the example/table scope. I double checked with the rest of docs.openshift.org and there it is also outside.

docs/source/using/host-folders.adoc
```diff
@@ -12,60 +12,71 @@ toc::[]
 
 Host folders are directories on the host which are shared between the host and the Minishift VM.
 They allow for a two way file synchronization between host and VM.
-The following sections discuss the various types of host folders, driver provided host folders, as well as the Minishift `hostfolder` command.
+The following sections discuss the various types of host folders, driver provided host folders, as well as the `minishift hostfolder` command.
```
hferentschik: I guess both works.

docs/source/using/host-folders.adoc
```diff
@@ -12,60 +12,71 @@ toc::[]
...
-On Windows CIFS is the default technology for sharing directories.
-For example, on Windows 10 the _C:\Users_ folder is shared by default and can be accessed by locally-authenticated users.
+To use the `minishift hostfolder` command, you need to be able to share directories using CIFS.
+On Windows, CIFS is the default technology for sharing directories.
+For example, on Windows 10 the *_C:\Users_* folder is shared by default and can be accessed by locally-authenticated users.
 
 It is also possible to use CIFS on OS X and Linux.
-On the former you can enable CIFS based shares under _System Preferences > Sharing_.
-Refer to link:https://support.apple.com/en-us/HT204445[How to connect with File Sharing on your Mac] for detailed setup instructions.
+On OS X, you can enable CIFS-based shares under *System Preferences > Sharing*.
```
hferentschik: Here is another OS X

docs/source/using/host-folders.adoc
```diff
@@ -186,26 +199,23 @@ No host folders defined
...
+You can use SSHFS-based host folders if you have an SSH daemon running on your host.
 Normally, this prerequisite is met by default on Linux and Mac OS X.
 
-Most Linux distributions have an SSH daemon installed. If not, follow the instructions
-for your specific distribution to install an SSH daemon.
+Most Linux distributions have an SSH daemon installed. If not, follow the instructions for your specific distribution to install an SSH daemon.
 
-Mac OS X also has a built-in
-SSH server. To use it, make sure that _Remote Login_ is enabled in _System Preferences > Sharing_.
+Mac OS X also has a built-in SSH server.
```
hferentschik: OS X

hferentschik:
@thatdocslady very nice indeed. Things are definitely taking shape.

docs/source/contributing/developing.adoc
thatdocslady: You're right that when you refer to the proper noun you don't need a qualifier, but in this case you used the definite article which does require it, and you're using an adjective ("updated") which requires the definite article. These are cascading grammatical dependencies. 

And a general comment, even though you personally use certain language to discuss things, try to remember that not everyone does, and we need to be prepared for different types of readers, which means keeping the language and grammar as explicit as possible.

docs/source/contributing/developing.adoc
thatdocslady: Actually when I checked why the weird string, I saw that there is this directory in my local minishift repo, which is why I changed it. Attaching screenshot, notice the breadcrumb. I can add back the wildcard annotation but the directory structure looks ok to me.

thatdocslady: screenshot from 2017-06-22 15-38-35

docs/source/getting-started/quickstart.adoc
thatdocslady: I'm replying once here but it should cover all of the instances where you commented about this. In the openshift doc guidelines, the instruction is to tag "System term, GUI menu items and buttons, daemon, service, or software package" with **, while "Inline commands, operations, literal values, environment variables, parameters, settings, flags, and user input" should use ``. AFAIK a binary is a software package so that's why it gets the ** tag. When referring to env vars or commands I use ``. 

Reference: https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#quick-markup-reference

docs/source/getting-started/setting-up-driver-plugin.adoc
thatdocslady: Thanks, and actually even overview anchor IDs need to be unique across the entire repository (as in no matter the hierarchy). This is according to the openshift guidelines: https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#unique-ids

docs/source/getting-started/setting-up-driver-plugin.adoc
thatdocslady: Ugh, I am not impressed with this capitalization, but I don't work at Apple so not much I can do to protest ;) will search and replace.

docs/source/openshift/openshift-client-binary.adoc
thatdocslady: We're supposed to use title caps in all headings, according to the openshift guidelines https://github.com/openshift/openshift-docs/blob/master/contributing_to_docs/doc_guidelines.adoc#topic-titles-and-section-headings

In fact, I'd like to suggest that we ditch the term "oc context", now that I dug a little into it and you also confirmed that it's a CLI Profile, and I would propose that we rename it as such so that we're aligned with the terms. I don't know where the term "oc context" came from" but it's not accurate, even though the command uses that term (which IMO should have been aligned in the first place but I guess it's too late now to change that).

docs/source/openshift/openshift-client-binary.adoc
thatdocslady: Good point, I'll convert these back to headings and rephrase the lead-in sentence a bit to work with it.

docs/source/using/addons.adoc
thatdocslady: You're right, I missed that one. Will pull the text out and rephrase to flow into the example heading.

docs/source/using/addons.adoc
thatdocslady: What do you mean "not good form"? I didn't change the heading of the table or the structure, I only removed the unnecessary definite articles, the text is clear without them and there's no need for "The" in the beginning of every line.

docs/source/using/host-folders.adoc
thatdocslady: I made similar changes in lots of places where people wrote only the sub-command without the parent, and I did this because of the likelihood that someone will copy and paste the command text, and if we only write something like "get" instead of "minishift config get" then the command wouldn't work.

thatdocslady:
@hferentschik I reviewed your feedback and replied inline wherever possible. The only thing about the oc-env comment, is that I actually I regenerated the docs many times (because I had to test every single link and make sure everything still compiles after changing the formatting), it's minishift itself (i.e. make clean) that I didn't run lately, I didn't think that I needed to clean up that deep. Now it works :)

docs/source/contributing/developing.adoc
hferentschik: I take this back. The directory does exist. I believe this got refactored at some stage. The idea with *.feature was to make explicit what the test files are, but I guess we don't need it.

docs/source/getting-started/quickstart.adoc
hferentschik: fair enough

docs/source/getting-started/setting-up-driver-plugin.adoc
hferentschik: Well, it is not required and I think the section you are referring to is misleading. Uniqueness is given by the full context. There would be no loss in information here, nor would there be an issue with other tooling. But let's move on

docs/source/openshift/openshift-client-binary.adoc
hferentschik: > We're supposed to use title caps in all headings, according to the openshift guidelines

I know, however, in this context it is a bit unfortunate that 'oc' is lowercase. Never mind.

> In fact, I'd like to suggest that we ditch the term "oc context", now that I dug a little into it and you also confirmed that it's a CLI Profile, and I would propose that we rename it as such so that we're aligned with the terms.

I am happy to rename the section "Minishift CLI Profile", however, we also need to keep the term 'context'. The term is correct as well and probably closer to what actually happens on a technical level - https://docs.openshift.org/latest/cli_reference/manage_cli_profiles.html

So +1 for changing the title, -1 for removing the mention to 'context'.

docs/source/openshift/openshift-client-binary.adoc
hferentschik: thanks

docs/source/using/addons.adoc
hferentschik: The comment is the same as for the example ones. You are pulling text into the table, similar as to the examples above.

hferentschik:
>  I reviewed your feedback and replied inline wherever possible. Let me know if you want to bjn about any of these, I hope my replies are clear and helpful.

I am fine. I get your formatting of 'oc' now. There are still several other things which needs updating, in particulate the examples and tables. Once you pushed an update I can have another look and if there are still question we can have a chat. WDYT?

thatdocslady:
@hferentschik implemented your feedback, and specifically:

- Aligned macOS term
- Changed oc context section title and rephrased to lead into "context" terminology
- Fixed examples and tables

I think this and a few other fixes basically cover what you asked for, I'm happy with how things are looking so if it looks good to you I think we can merge it. Of course rebased etc. :)

docs/source/contributing/developing.adoc
hferentschik: > And a general comment, even though you personally use certain language to discuss things, try to remember that not everyone does, and we need to be prepared for different types of readers

Agree. On the same note, please remember that we are targeting developers who have partly their own lingo. What might make sense for a non technical user or non-dev, will sound awkward for a developer.

hferentschik:
> I think this and a few other fixes basically cover what you asked for

+1, thanks for applying the fixes

> I'm happy with how things are looking 

Same here

> So if it looks good to you I think we can merge it.

Will rebase, squash and merge.

hferentschik:
Rebased, squashed and merged. Thanks.

docs/source/contributing/developing.adoc
```diff
@@ -131,10 +131,9 @@ If you seeing Glide errors during `make vendor`, you can clear the Glide cache v
 == Building Minishift
 
 [[build-minishift-binary]]
-=== Building the Minishift binary
+=== Building the Minishift Binary
```
LalatenduMohanty: @thatdocslady I confused why `the` has small `t` but other words have capital letter as starting letter.

gbraad: Articles are usually written as lowercase; 'a', 'an', and 'the'.

The point here would be, consistency. In the documentation there is
  * Building the Minishift binary
  * Running the Minishift binary

thatdocslady: Right @LalatenduMohanty , the rules for title-case capitalization are not to capitalize articles and also some prepositions such as "in", "on". 

@gbraad I don't know what you are referring to, I see the titles on the master branch and they're capitalized correctly. Where did you spot this?

gbraad: @thatdocslady I mention that they are consistent. as the report would have only pick one of the both cases.

Issue #1046, opened 2017-06-20
`#{routing-suffix}` does not get populated


jstrachan:
here's a sample PR
https://gist.github.com/jstrachan/0d9bb130ffbb4f2f64316fb20d8f6daf

the routing-suffix does not get replaced with `nip.io` though - its blank https://github.com/jstrachan/minishift/blob/91814cf1403e5721931dd7ffc06cc74f7e44b8ff/addons/fabric8/fabric8.addon#L7

hferentschik:
#{routing-prefix} is not a valid variable name. There is the #{routing-suffix} which gets interpolated. Is this what you want? And if not, what do you expect #{routing-prefix} to expand to?

jstrachan:
sorry bad typing on the issue - I copy pasted the name `routing-suffix` from the docs into the source code and thats the one that doesn't work - I didn't copy the text into the issue though ;) just editted it

jstrachan:
I just confirmed this error occurs on the default openshift 3.5.x as well as 3.6.x; it also doesn't matter if I deploy some Routes first before I try and use the `#{routing-suffix}` or not

hferentschik:
@jstrachan Are you using `minishift addons apply fabric8` when you test this? My guess is yes. If you enable the plugin and run `minishift start` you most likely will see that the substitution will work. The problem is not so much that interpolation does not occur, but rather that the routing-suffix is not properly set when `addons apply` is executed.

Aside of this, I noticed something along the lines of `#{ip}.#{routing-suffix}` on the add-on. This way you will end up with the IP twice. The routing-suffix is not just 'nip.io' it is ip+'.nip.io'. So you only want to interpolate the routing suffix. 

I try to see whether there is an easy way to fix this.

hferentschik:
@jstrachan FYI https://github.com/minishift/minishift/pull/1056

jstrachan:
@hferentschik so previously I'd done something like this...
```
minishift start --vm-driver=xhyve --memory=7000 --cpus=4 --disk-size=50g 
minishift addons install --force addons/fabric8
minishift addons apply
```
after your comment I realised maybe the way you're meant to use addons is to install & enable them first then just run minishift start. So I tried that...
```
$ minishift addons install --force addons/fabric8
$ minishift addons enable fabric8
$ minishift start --vm-driver=xhyve --memory=7000 --cpus=4 --disk-size=50g --openshift-version=v3.6.0-alpha.1
[...]

PR #1056, opened 2017-06-21
Issue #1046 Making sure routing suffix is provided in context for 'add…

hferentschik:
…ons apply'

Addresses issue #1046

### DIFF
--- a/cmd/minishift/cmd/addon/addon_apply.go
+++ b/cmd/minishift/cmd/addon/addon_apply.go
@@ -20,15 +20,18 @@ import (
 	"fmt"
 
 	"github.com/docker/machine/libmachine"
+	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/provision"
 	"github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
+	"github.com/minishift/minishift/pkg/minishift/docker"
 	"github.com/minishift/minishift/pkg/minishift/oc"
+	"github.com/minishift/minishift/pkg/minishift/openshift"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
-	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
 )
 
 const (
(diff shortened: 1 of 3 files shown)


hferentschik:
@budhrg @praveenkumar - this should be good to go now. Give it a go if you can.

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -97,3 +100,28 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
...
+
+	var config map[interface{}]interface{}
+	err = yaml.Unmarshal([]byte(raw), &config)
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to parse OpenShift master configuration: %s", err.Error()))
+	}
+
+	// making assumptions about the master config here. In case the config structure changes, the code might panic here
+	return config["routingConfig"].(map[interface{}]interface{})["subdomain"].(string)
+}
```
LalatenduMohanty: I think this the best we can do now.

cmd/minishift/cmd/start.go
```diff
@@ -374,8 +374,7 @@ func determineClusterUpParameters(config *clusterup.ClusterUpConfig) map[string]
 	viper.Set(startFlags.HostDataDir.Name, viper.GetString(startFlags.HostDataDir.Name))
 	viper.Set(startFlags.HostVolumeDir.Name, viper.GetString(startFlags.HostVolumeDir.Name))
 	viper.Set(startFlags.HostPvDir.Name, viper.GetString(startFlags.HostPvDir.Name))
-
-	setDefaultRoutingPrefix(config.Ip)
+	viper.Set(startFlags.RoutingSuffix.Name, config.RoutingSuffix)
```
LalatenduMohanty: Just to confirm, you are doing this to make sure if user has set any routing suffix it gets passed to startFlags right?

cmd/minishift/cmd/addon/addon_apply.go
hferentschik: Not sure what you mean. Are you having potential concerns on the method we retrieve the routing suffix or how I deal with the YAML. In case of the former, I believe this is by far the best way and I really like the we benefit from the infrastructure we have in place.

In case of the latter. There are always two ways to deal with this. I could also makes this a typed un-marshall, but imo this is not required here (and can fail as well, in case the underlying YAML changes). In the worst case the user sees an error message.

cmd/minishift/cmd/start.go
hferentschik: Yes

PR #1061, opened 2017-06-25
Issue #320 Add --follow option to 'minishift logs' command

### LINKED ISSUE #320: Add --follow option to `minishift logs` subcommand

budhrg:
Fix #320 

@hferentschik This work functionally IMO. Also, others have verified it.

### DIFF
--- a/cmd/minishift/cmd/logs.go
+++ b/cmd/minishift/cmd/logs.go
@@ -27,6 +27,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+var (
+	follow bool
+)
+
 // logsCmd represents the logs command
 var logsCmd = &cobra.Command{
 	Use:   "logs",
@@ -35,7 +39,7 @@ var logsCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
-		s, err := cluster.GetHostLogs(api)
+		s, err := cluster.GetHostLogs(api, follow)
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error getting logs: %s", err.Error()))
 		}
@@ -44,5 +48,6 @@ var logsCmd = &cobra.Command{
 }
 
 func init() {
+	logsCmd.Flags().BoolVarP(&follow, "follow", "f", false, "Continously print the logs entries")
 	RootCmd.AddCommand(logsCmd)
 }
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -43,11 +43,13 @@ import (
 	"github.com/minishift/minishift/pkg/util"
 	minishiftOs "github.com/minishift/minishift/pkg/util/os"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/pkg/errors"
 	pb "gopkg.in/cheggaaa/pb.v1"
 )
 
 var (
 	logsCmd             = "docker logs origin"
+	logsCmdFollow       = logsCmd + " -f"
 	dockerAPIVersionCmd = "docker version --format '{{.Server.APIVersion}}'"
 )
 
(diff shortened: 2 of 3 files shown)


pkg/minikube/cluster/cluster.go
```diff
@@ -43,11 +43,13 @@ import (
 	"github.com/minishift/minishift/pkg/util"
 	minishiftOs "github.com/minishift/minishift/pkg/util/os"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/pkg/errors"
 	pb "gopkg.in/cheggaaa/pb.v1"
 )
 
 var (
 	logsCmd             = "docker logs origin"
+	logsCmdFollow       = "docker logs origin -f"
```
gbraad: Not a big problem, but reusing `logsCmd` makes sense

cmd/minishift/cmd/logs.go
```diff
@@ -44,5 +48,6 @@ var logsCmd = &cobra.Command{
 }
 
 func init() {
+	logsCmd.Flags().BoolVarP(&follow, "follow", "f", false, "Continously print the logs entries")
```
gbraad: +1 `-f` ;-)

pkg/minikube/cluster/cluster.go
budhrg: Done :+1:

cmd/minishift/cmd/logs.go
budhrg: @gbraad , that's not required as cobra take care of it  

```
$ minishift logs -h
Gets the logs of the running OpenShift cluster. The logs do not contain information about your application code.

Usage:
  minishift logs [flags]

Flags:
  -f, --follow   Continously print the logs entries
  -h, --help     help for logs
```

gbraad: I meant, it is good to have the shorthand alias.

pkg/minikube/cluster/cluster.go
```diff
@@ -468,16 +470,32 @@ func GetHostDockerEnv(api libmachine.API) (map[string]string, error) {
...
 // GetHostLogs gets the openshift logs of the host VM.
-func GetHostLogs(api libmachine.API) (string, error) {
+// If follow is specified, it will tail the logs
+func GetHostLogs(api libmachine.API, follow bool) (string, error) {
 	host, err := CheckIfApiExistsAndLoad(api)
 	if err != nil {
 		return "", err
 	}
+
+	if follow {
```
hferentschik: Can we not pull this into its own method and make the switch on the _follow_ flag more apparent?  Really we have an is/else here, right?

pkg/minikube/cluster/cluster_test.go
```diff
@@ -50,3 +53,52 @@ func TestLocalBoot2DockerURL(t *testing.T) {
...
+		},
+		{
+			description: "logs -f",
+			follow:      true,
+		},
+	}
+
+	for _, test := range tests {
+		t.Run(test.description, func(t *testing.T) {
+			t.Parallel()
```
hferentschik: Why parallel. It does not serve any purpose here and makes it harder to to understand what you are testing.

pkg/minikube/cluster/cluster_test.go
```diff
@@ -50,3 +53,52 @@ func TestLocalBoot2DockerURL(t *testing.T) {
 		t.Fatalf("Expected URL : %s", localISOUrl)
 	}
 }
+
+func TestHostGetLogs(t *testing.T) {
```
hferentschik: I think we discussed this before, I am not so sure of the usefulness of the test. At the very least we should make explicit what this actually does. It verifies that the '-f' flag if specified gets passed to the VM (via ssh). This is by no means a test to verify whether streaming the logs actually works or not.

budhrg:
@hferentschik  Updated the PR as per our discussion.

hferentschik:
Merged, thanks.

hferentschik:
Merged via pull request #1061

Issue #1064, opened 2017-06-26
Empty echo command in add-on does not print empty line


hferentschik:
Adding an empty "echo" command in an add-on to print an empty lines fails. Should just print the empty line.

LalatenduMohanty:
@hferentschik It is not failing for me. However there seems to be a bug for sure. For example if I execute an addon with following content
```
# Name: foo
# Description: Changes the default security context constraints to allow pods to run as any user.
# Url: https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-origin-specific-guidelines

echo  hello
echo
echo  bye
~         
```
The out put contains the `echo` as below. 
```
-- Applying addon 'foo':
   hello
  echo
   bye
```

budhrg:
This is happening in this PR as well https://github.com/minishift/minishift/pull/1043

hferentschik:
@LalatenduMohanty, so it is "failing", right?

hferentschik:
Merged via pull request #1079

PR #1079, opened 2017-06-28
Issue #1064 Fixed the echo add-on command to print empty line

LalatenduMohanty:

### DIFF
--- a/pkg/minishift/addon/command/echo_command.go
+++ b/pkg/minishift/addon/command/echo_command.go
@@ -34,11 +34,19 @@ func NewEchoCommand(command string) *EchoCommand {
 }
 
 func (c *EchoCommand) doExecute(ec *ExecutionContext) error {
-	// split off the actual 'echo' command. As we need to print rest of the string
-	echo := strings.Replace(c.rawCommand, "echo ", "", 1)
-	echo = ec.Interpolate(echo)
+	var err error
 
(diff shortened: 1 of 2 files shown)


LalatenduMohanty:
@hferentschik Agree, will add test.

pkg/minishift/addon/command/echo_command.go
```diff
@@ -34,11 +34,19 @@ func NewEchoCommand(command string) *EchoCommand {
...
+	var err error
 
-	_, err := fmt.Print("\n  " + echo)
+	// split off the actual 'echo' command. As we need to print rest of the string
+	echoString := strings.Replace(c.rawCommand, "echo", "", 1)
+
+	//Remove leading and trailing spaces
+	echoString = strings.TrimSpace(echoString)
+	if echoString == "" {
+		_, err = fmt.Print("\n")
```
praveenkumar: Why can't we just use `fmt.Println()` which auto append newline?

LalatenduMohanty: @praveenkumar I think it is almost same thing. I used `Print("\n")` for consistency in the existing code and it is explicit in a way.

hferentschik: Right, it is more explicit. I definitely get it. It is like _stressing_ a word.

hferentschik:
Merged, thanks.

Issue #850, opened 2017-04-28
Edit the doc strings for newly added commands


thatdocslady:
Since the first round of doc string edits, some new commands were added to minishift and therefore need to be reviewed. For example minishift openshift, addon-apply, update.

thatdocslady:
@hferentschik I can see a few new commands that haven't gone thru a doc string edit yet (i.e. update, addon-apply, etc). Should I rephrase this issue to cover all new commands, or is it better to handle each command set in a separate issue?

hferentschik:
Please all in one :-)

hferentschik:
Merged via pull request #1088

PR #1088, opened 2017-06-29
Issue #850 Edited doc strings for latest commands

thatdocslady:
Fixes issue #850 

Since the last time the doc strings were added, new commands were incorporated in to Minishift. This PR reviews and cleans up the doc strings.

Out of scope by request: *_test.go files

### DIFF
--- a/cmd/minishift/cmd/addon/addon.go
+++ b/cmd/minishift/cmd/addon/addon.go
@@ -23,15 +23,15 @@ import (
 const (
 	addOnConfigKey = "addons"
 
-	emptyAddOnError = "You must specify an add-on name. Use `minishift addons list` to view installed add-ons."
+	emptyAddOnError = "You must specify an add-on name. Run `minishift addons list` to view installed add-ons."
 	noAddOnMessage  = "No add-on with the name %s is installed."
 )
 
 var AddonsCmd = &cobra.Command{
 	Use:     "addons SUBCOMMAND [flags]",
 	Aliases: []string{"addon"},
-	Short:   "Manages Minishift add-ons",
-	Long:    "Manages Minishift add-ons. You can install, list, enable or disable Minishift add-ons.",
+	Short:   "Manages Minishift add-ons.",
+	Long:    "Manages Minishift add-ons. Use the sub-commands to install, list, enable or disable Minishift add-ons.",
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
 	},
--- a/cmd/minishift/cmd/addon/addon_apply.go
+++ b/cmd/minishift/cmd/addon/addon_apply.go
@@ -40,8 +40,8 @@ const (
 
 var addonsApplyCmd = &cobra.Command{
 	Use:   "apply ADDON_NAME ...",
-	Short: "Executes the specified add-ons.",
-	Long:  "Executes the specified add-ons. The command works with both enabled and disabled add-ons.",
+	Short: "Applies the specified add-ons.",
+	Long:  "Applies the specified add-ons. You can specify one or more add-ons, regardless of whether the add-on is enabled or disabled.",
 	Run:   runApplyAddon,
 }
 
@@ -77,14 +77,14 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
 
 	ip, err := host.Driver.GetIP()
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting IP: %s", err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting the IP address: %s", err.Error()))
 	}
 
 	routingSuffix := determineRoutingSuffix(host.Driver)
 	sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
 	ocRunner, err := oc.NewOcRunner(minishiftConfig.InstanceConfig.OcPath, constants.KubeConfigPath)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Error applying addon: %s", err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error applying the add-on: %s", err.Error()))
 	}
 
 	for i := range args {
@@ -92,11 +92,11 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
 		addon := addOnManager.Get(addonName)
 		addonContext, err := clusterup.GetExecutionContext(ip, routingSuffix, ocRunner, sshCommander)
 		if err != nil {
-			atexit.ExitWithMessage(1, fmt.Sprint("Error executing addon commands: ", err))
+			atexit.ExitWithMessage(1, fmt.Sprint("Error applying the add-on: ", err))
 		}
 		err = addOnManager.ApplyAddOn(addon, addonContext)
 		if err != nil {
-			atexit.ExitWithMessage(1, fmt.Sprint("Error executing addon commands: ", err))
+			atexit.ExitWithMessage(1, fmt.Sprint("Error applying the add-on: ", err))
 		}
 	}
 }
@@ -104,7 +104,7 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
 func determineRoutingSuffix(driver drivers.Driver) string {
 	defer func() {
 		if r := recover(); r != nil {
-			atexit.ExitWithMessage(1, "Unable to determine routing suffix from OpenShift master config.")
+			atexit.ExitWithMessage(1, "Cannot determine the routing suffix from the OpenShift master configuration.")
 		}
 	}()
 
@@ -113,13 +113,13 @@ func determineRoutingSuffix(driver drivers.Driver) string {
 
 	raw, err := openshift.ViewConfig(openshift.MASTER, dockerCommander)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to retrieve OpenShift master configuration: %s", err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot get the OpenShift master configuration: %s", err.Error()))
 	}
 
 	var config map[interface{}]interface{}
 	err = yaml.Unmarshal([]byte(raw), &config)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to parse OpenShift master configuration: %s", err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot parse the OpenShift master configuration: %s", err.Error()))
 	}
 
 	// making assumptions about the master config here. In case the config structure changes, the code might panic here
--- a/cmd/minishift/cmd/addon/addon_enable.go
+++ b/cmd/minishift/cmd/addon/addon_enable.go
@@ -38,7 +38,7 @@ var addonsEnableCmd = &cobra.Command{
 }
 
 func init() {
-	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of the add-on when it is applied.")
+	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of the add-on.")
 	AddonsCmd.AddCommand(addonsEnableCmd)
 }
 
@@ -60,9 +60,9 @@ func runEnableAddon(cmd *cobra.Command, args []string) {
 func enableAddon(addOnManager *manager.AddOnManager, addOnName string, priority int) {
 	addOnConfig, err := addOnManager.Enable(addOnName, priority)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to enable add-on %s: %s", addOnName, err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot enable the add-on %s: %s", addOnName, err.Error()))
 	} else {
-		fmt.Println(fmt.Sprintf("Addon '%s' enabled", addOnName))
+		fmt.Println(fmt.Sprintf("Add-on '%s' enabled", addOnName))
 	}
 
 	addOnConfigMap := GetAddOnConfiguration()
--- a/cmd/minishift/cmd/addon/addon_install.go
+++ b/cmd/minishift/cmd/addon/addon_install.go
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	unspecifiedSourceError   = "You need to specify the source for the add-on."
+	unspecifiedSourceError   = "You must specify the source of the add-on."
 	failedPluginInstallation = "Add-on installation failed with the error: %s"
 
 	forceFlag    = "force"
--- a/cmd/minishift/cmd/addon/util.go
+++ b/cmd/minishift/cmd/addon/util.go
@@ -32,7 +32,7 @@ func GetAddOnManager() *manager.AddOnManager {
 	addOnConfigs := GetAddOnConfiguration()
 	m, err := manager.NewAddOnManager(constants.MakeMiniPath("addons"), addOnConfigs)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to initialize the add-on manager: %s", err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot initialize the add-on manager: %s", err.Error()))
 	}
 
 	return m
@@ -41,14 +41,14 @@ func GetAddOnManager() *manager.AddOnManager {
 func WriteAddOnConfig(addOnConfigMap map[string]*addon.AddOnConfig) {
 	c, err := config.ReadConfig()
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to read the Minishift configuration: %s", err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot read the Minishift configuration: %s", err.Error()))
 	}
 
 	c[addOnConfigKey] = addOnConfigMap
 
 	err = config.WriteConfig(c)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to write the Minishift configuration: %s", err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot write the Minishift configuration: %s", err.Error()))
 	}
 }
 
@@ -57,7 +57,7 @@ func WriteAddOnConfig(addOnConfigMap map[string]*addon.AddOnConfig) {
 func GetAddOnConfiguration() map[string]*addon.AddOnConfig {
 	c, err := config.ReadConfig()
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to read the Minishift configuration: %s", err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot read the Minishift configuration: %s", err.Error()))
 	}
 
 	var configSlice map[string]interface{}
--- a/cmd/minishift/cmd/image/export.go
+++ b/cmd/minishift/cmd/image/export.go
@@ -31,8 +31,8 @@ import (
 
 var imageExportCmd = &cobra.Command{
 	Use:   "export [image ...]",
-	Short: "Exports container images (experimental).",
-	Long:  "Exports container images (experimental).",
+	Short: "Exports the specified container images (experimental).",
+	Long:  "Exports the specified container images (experimental).",
 	Run:   exportImage,
 }
 
@@ -44,21 +44,21 @@ func exportImage(cmd *cobra.Command, args []string) {
 	defer api.Close()
 
 	if len(args) < 1 {
-		atexit.ExitWithMessage(0, "No image specifications provided")
+		atexit.ExitWithMessage(0, "You must specify at least one container image.")
 	}
 
 	util.ExitIfUndefined(api, constants.MachineName)
 
 	host, err := api.Load(constants.MachineName)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Error creating VM client: %v", err))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error creating the VM client: %v", err))
 	}
 
 	util.ExitIfNotRunning(host.Driver, constants.MachineName)
 
 	handler, err := image.NewDockerImageHandler(host.Driver)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to create image handler: %v", err))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot create the image handler: %v", err))
 	}
 
 	imageCacheConfig := &image.ImageCacheConfig{
@@ -69,7 +69,7 @@ func exportImage(cmd *cobra.Command, args []string) {
 	}
 	err = handler.ExportImages(imageCacheConfig)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Export of container images failed: %v", err))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Failed to export the container images: %v", err))
 	}
 }
 
@@ -79,7 +79,7 @@ func createLogFile() *os.File {
 	logFilePath := filepath.Join(constants.MakeMiniPath("logs"), fmt.Sprintf("image-export-%s.log", timeStamp))
 	logFile, err := os.Create(logFilePath)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to create log file for image export: %v", err))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot create the log file of the image export: %v", err))
 	}
 
 	return logFile
--- a/cmd/minishift/cmd/oc_env.go
+++ b/cmd/minishift/cmd/oc_env.go
@@ -65,11 +65,11 @@ func executeOcTemplateStdout(shellCfg *OcShellConfig) error {
 
 var ocEnvCmd = &cobra.Command{
 	Use:   "oc-env",
-	Short: "Sets path of 'oc' binary.",
-	Long:  `Sets path of OpenShift client binary, 'oc'.`,
+	Short: "Sets the path of the 'oc' binary.",
+	Long:  `Sets the path of OpenShift client binary 'oc'.`,
 	Run: func(cmd *cobra.Command, args []string) {
 		if config.InstanceConfig.OcPath == "" {
-			atexit.ExitWithMessage(1, "Unable to find OpenShift client binary.\nPlease make sure that OpenShift has been provisioned successfully.")
+			atexit.ExitWithMessage(1, "Cannot find the OpenShift client binary.\nMake sure that OpenShift was provisioned successfully.")
 		}
 
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
@@ -85,7 +85,7 @@ var ocEnvCmd = &cobra.Command{
 
 		shellCfg, err = getOcShellConfig(config.InstanceConfig.OcPath, forceShell)
 		if err != nil {
-			atexit.ExitWithMessage(1, fmt.Sprintf("Error running oc-env command: %s", err.Error()))
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error running the oc-env command: %s", err.Error()))
 		}
 
 		executeOcTemplateStdout(shellCfg)
--- a/cmd/minishift/cmd/openshift/config.go
+++ b/cmd/minishift/cmd/openshift/config.go
@@ -27,8 +27,8 @@ const (
 
 var configCmd = &cobra.Command{
 	Use:   "config SUBCOMMAND [flags]",
-	Short: "Displays or patches OpenShift configuration.",
-	Long:  "Displays or patches OpenShift master or node configuration. Patches are supplied in JSON format.",
+	Short: "Displays or patches the OpenShift configuration.",
+	Long:  "Displays or patches the OpenShift master or node configuration. Patches must be in a valid JSON format.",
 	Run: func(cmd *cobra.Command, args []string) {
 		cmd.Help()
 	},
--- a/cmd/minishift/cmd/openshift/openshift.go
+++ b/cmd/minishift/cmd/openshift/openshift.go
@@ -22,8 +22,8 @@ import (
 
 var OpenShiftCmd = &cobra.Command{
 	Use:   "openshift SUBCOMMAND [flags]",
-	Short: "Interact with an Openshift Cluster",
-	Long:  "Interact with an Openshift cluster, check available versions and modify master config resources",
+	Short: "Interacts with your local OpenShift cluster.",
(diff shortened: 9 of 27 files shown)


cmd/minishift/cmd/addon/addon_install.go
```diff
@@ -44,7 +44,7 @@ var (
 var addonsInstallCmd = &cobra.Command{
 	Use:   "install [SOURCE]",
 	Short: "Installs the specified add-on.",
-	Long:  "Installs the add-on from the specified file path and verifies the installation.",
+	Long:  "Installs the specified add-on from the specified file path and verifies the installation.",
```
hferentschik: why "specified" add-on? You are not specifying and add-on, you are specifying an installation source in form of a path.

cmd/minishift/cmd/image/export.go
```diff
@@ -69,7 +69,7 @@ func exportImage(cmd *cobra.Command, args []string) {
 	}
 	err = handler.ExportImages(imageCacheConfig)
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Export of container images failed: %v", err))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Failed to export the container image: %v", err))
```
hferentschik: plural "images"

cmd/minishift/cmd/oc_env.go
```diff
@@ -65,11 +65,11 @@ func executeOcTemplateStdout(shellCfg *OcShellConfig) error {
 
 var ocEnvCmd = &cobra.Command{
 	Use:   "oc-env",
-	Short: "Sets path of 'oc' binary.",
-	Long:  `Sets path of OpenShift client binary, 'oc'.`,
+	Short: "Sets the path of the 'oc' binary.",
+	Long:  `Sets the path of OpenShift client binary, or 'oc'.`,
```
hferentschik: "or"? Where s that coming from? The OpenShift client binary = oc

cmd/minishift/cmd/openshift/config.go
```diff
@@ -27,8 +27,8 @@ const (
 
 var configCmd = &cobra.Command{
 	Use:   "config SUBCOMMAND [flags]",
-	Short: "Displays or patches OpenShift configuration.",
-	Long:  "Displays or patches OpenShift master or node configuration. Patches are supplied in JSON format.",
+	Short: "Displays or updates the OpenShift configuration.",
```
hferentschik: The process is called "patching". This is the right word, not updating. This is what I am trying to say. We are targeting developers and in the terminology of software development patches are applied.

cmd/minishift/cmd/openshift/config.go
```diff
@@ -27,8 +27,8 @@ const (
...
+	Long:  "Displays or updates the OpenShift master or node configuration. Patches must be in a valid JSON format.",
```
hferentschik: Same as above.

cmd/minishift/cmd/openshift/openshift.go
```diff
@@ -22,8 +22,8 @@ import (
 
 var OpenShiftCmd = &cobra.Command{
 	Use:   "openshift SUBCOMMAND [flags]",
-	Short: "Interact with an Openshift Cluster",
-	Long:  "Interact with an Openshift cluster, check available versions and modify master config resources",
+	Short: "Interacts with your local OpenShift cluster.",
```
hferentschik: Technically the 'openshift' command does not do such thing. It is the container command for the actual subcommands which interact with OpenShift.

cmd/minishift/cmd/openshift/openshift.go
```diff
@@ -22,8 +22,8 @@ import (
...
+	Long:  "Interacts with your local OpenShift cluster. Use the sub-commands to check available OpenShift versions, access services, and modify configuration resources.",
```
hferentschik: I think we need to reword this. `minishift openshift` does not stand alone. It is the actual sub-commands which do something.

cmd/minishift/cmd/openshift/version.go
```diff
@@ -32,8 +32,8 @@ import (
 // version command represent current running openshift version and available one.
 var versionCmd = &cobra.Command{
 	Use:   "version [command] [flags]",
-	Short: "Print the current running openshift version to stdout",
-	Long:  `Print the current running openshift version to stdout`,
+	Short: "Prints the current running OpenShift version to the console.",
```
hferentschik: Actually we need to be careful here. Technically stdout is actually better than console, since you might be re-directing your stdout in which case nothing comes out to the console.

cmd/minishift/cmd/update.go
```diff
@@ -89,16 +89,16 @@ func runUpdate(cmd *cobra.Command, args []string) {
...
 			markerData := UpdateMarker{InstallAddon: false, PreviousVersion: version.GetMinishiftVersion()}
 
-			fmt.Print("\nDo you want to update default addons? [y/N]: ")
+			fmt.Print("\nDo you want to update the default add-ons? [y/N]: ")
 			fmt.Scanln(&confirm)
 
 			if strings.ToLower(confirm) == "y" {
 				markerData.InstallAddon = true
-				fmt.Println("Default add-ons will be updated when executing the next minishift command.")
+				fmt.Println("Default add-ons will be updated the next time you run any `minishift` command.")
```
hferentschik: I think we agreed to not use backticks in the help messages. This is not markdown.

cmd/minishift/cmd/addon/addon_install.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	unspecifiedSourceError   = "You need to specify the source for the add-on."
+	unspecifiedSourceError   = "You must specify the source file path of the add-on."
```
LalatenduMohanty: We need to provide path to the directory which contains the addon file. So we need to reword this e.g. `You must specify the source directory path of the add-on` or something similar.

cmd/minishift/cmd/addon/util.go
```diff
@@ -41,14 +41,14 @@ func GetAddOnManager() *manager.AddOnManager {
 func WriteAddOnConfig(addOnConfigMap map[string]*addon.AddOnConfig) {
 	c, err := config.ReadConfig()
 	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Unable to read the Minishift configuration: %s", err.Error()))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot read the Minishift configuration: %s", err.Error()))
```
LalatenduMohanty: What is the difference between `unable` or `cannot` here. Just trying to understand.

hferentschik:
@thatdocslady, your pull request broke the integration tests, since you updated some strings which were assessed in the integration test - https://ci.centos.org/job/minishift-pr/1382/console

cmd/minishift/cmd/addon/util.go
thatdocslady: This is for consistency and conciseness. "Unable to" is considered in some parts more "high language" and "Cannot" is more universally understood by non-native English speakers. We also use "cannot" already in other commands, so I'm aligning the terminology and phrasing.

cmd/minishift/cmd/update.go
thatdocslady: I don't remember we agreed on this, and we should have a way to indicate commands and flags. I actually see/use straight quotes in other places, so I can change the backticks to those.

cmd/minishift/cmd/openshift/openshift.go
thatdocslady: We already use this phrasing format with other parent/container commands, so I was just aligning with those (i.e. hostfolder, config, addon).

cmd/minishift/cmd/oc_env.go
thatdocslady: that's the same thing but I can think of another way of indicating this that doesn't involve the word "or" if it bothers you.

cmd/minishift/cmd/image/export.go
thatdocslady: so is it all images? usually I see usage of one (specified) thing or all things.

cmd/minishift/cmd/openshift/config.go
thatdocslady: We already use "update" in other config commands so I'm aligning with existing (agreed upon) terminology.

thatdocslady:
@hferentschik @LalatenduMohanty implemented some feedback and replied to some open issues. I am not sure how to fix the integration tests, this didn't happen last time I edited the doc strings. Maybe you can show me how to fix this?

cmd/minishift/cmd/openshift/openshift.go
hferentschik: I know. I was just pointing it out here.

cmd/minishift/cmd/openshift/config.go
hferentschik: I don't think this is correct and we had the same discussion for the docs. In this case we are talking about patching which means more than updating. I cannot recall where we agreed that patching == updating

cmd/minishift/cmd/image/export.go
hferentschik: > so is it all images? usually I see usage of one (specified) thing or all things.

We are exporting multiple images.

hferentschik:
In the console output you see the falling test - https://ci.centos.org/job/minishift-pr/1382/console:

    Error: Output did not match. Expected: Addon 'anyuid' enabled, Actual: Add-on 'anyuid' enabled

You need to locate this assertion in [basic.feature](https://github.com/minishift/minishift/blob/master/test/integration/features/basic.feature) and align it with your text changes. Then you can run:

    $ make integration 

to verify the integration tests are passing now.

cmd/minishift/cmd/openshift/openshift.go
hferentschik: Sure, I am just saying that we maybe could phrase this a bit better. Here and then potentially at the other commands as well.

cmd/minishift/cmd/image/export.go
hferentschik: > so is it all images? usually I see usage of one (specified) thing or all things.

Not quite able to parse your comment, but yes we are exporting multiple images.

cmd/minishift/cmd/openshift/config.go
hferentschik: I don't think so. At one point we are saying "updating .... with a patch", but this still keeps the reference to patching. Patching means more than updating. Updating can happen in many ways, patching is the process of updating something (config, software, etc) using the diff format.

I am pretty sure we had this discussion before when eating the asciidoctor docs.

hferentschik:
> Out of scope by request: *_test.go files

I think we generally agreed that there is no need to touch the test files.

cmd/minishift/cmd/image/export.go
thatdocslady: sorry if my comment wasn't clear, what I mean is that we should clarify how many images we are referring to, whether the users can select certain images, only one image, or all images? these are the types of questions that are most likely to be asked by users.

cmd/minishift/cmd/openshift/config.go
thatdocslady: First of all, "updating configuration" doesn't always happen with a patch, sometimes you set properties in a config file. This specific usage is with a patch, yes, but the end-goal or overall goal is to update.

We discussed the usage of update/patch before, but if you look at the other config command doc strings you will see that we do use "update" to describe the high-level use-case, and these are already merged and in use on master so I would think that we agreed on this (otherwise we wouldn't have merged those doc strings to begin with).

If you want to change the doc strings across the board we need to discuss further, but it's reverting a convention we already set the last time we reviewed doc strings.

thatdocslady: Sorry, I take this back. In other config command we use "modify" as well as "update". Let me take a look again and see what the best way to standardize.

thatdocslady: Ok I will just use "patching" because tbh it's not that critical and I already spent too much energy trying to work around it. Will update the PR after I fix the integration test and other feedback.

cmd/minishift/cmd/addon/addon.go
```diff
@@ -23,15 +23,15 @@ import (
...
+	emptyAddOnError = "You must specify an add-on name. Run `minishift addons list` to view installed add-ons."
 	noAddOnMessage  = "No add-on with the name %s is installed."
 )
 
 var AddonsCmd = &cobra.Command{
 	Use:     "addons SUBCOMMAND [flags]",
 	Aliases: []string{"addon"},
-	Short:   "Manages Minishift add-ons",
-	Long:    "Manages Minishift add-ons. You can install, list, enable or disable Minishift add-ons.",
+	Short:   "Manages Minishift add-ons using sub-commands.",
```
hferentschik: "Manages Minishift add-ons with its sub-commands."?

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -77,34 +77,34 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
...
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error applying the add-on: %s", err.Error()))
 	}
 
 	for i := range args {
 		addonName := args[i]
 		addon := addOnManager.Get(addonName)
 		addonContext, err := clusterup.GetExecutionContext(ip, routingSuffix, ocRunner, sshCommander)
 		if err != nil {
-			atexit.ExitWithMessage(1, fmt.Sprint("Error executing addon commands: ", err))
+			atexit.ExitWithMessage(1, fmt.Sprint("Error running add-on commands: ", err))
```
hferentschik: Applying instead of running?

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -77,34 +77,34 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
...
 		}
 		err = addOnManager.ApplyAddOn(addon, addonContext)
 		if err != nil {
-			atexit.ExitWithMessage(1, fmt.Sprint("Error executing addon commands: ", err))
+			atexit.ExitWithMessage(1, fmt.Sprint("Error running add-on commands: ", err))
```
hferentschik: Same as above

cmd/minishift/cmd/addon/addon_enable.go
```diff
@@ -38,7 +38,7 @@ var addonsEnableCmd = &cobra.Command{
 }
 
 func init() {
-	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of the add-on when it is applied.")
+	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of the add-on. Determines the order in which to apply the add-on.")
```
hferentschik: "Used to determine the order in which add-ons get applied"

cmd/minishift/cmd/addon/addon.go
thatdocslady: Not a good idea, since "its" is a possessive and therefore it makes the sentence ambiguous (are the sub-commands belonging to add-ons, to Minishift, or to the command? etc). Leaving it out means the sub-commands stand on their own.

cmd/minishift/cmd/image/export.go
```diff
@@ -31,8 +31,8 @@ import (
 
 var imageExportCmd = &cobra.Command{
 	Use:   "export [image ...]",
-	Short: "Exports container images (experimental).",
-	Long:  "Exports container images (experimental).",
+	Short: "Exports the specified container image (experimental).",
```
hferentschik: images

cmd/minishift/cmd/image/export.go
```diff
@@ -44,21 +44,21 @@ func exportImage(cmd *cobra.Command, args []string) {
 	defer api.Close()
 
 	if len(args) < 1 {
-		atexit.ExitWithMessage(0, "No image specifications provided")
+		atexit.ExitWithMessage(0, "You must specify a container image.")
```
hferentschik: You must specify at least one container image.

cmd/minishift/cmd/image/export.go
hferentschik: This is an error message indicating something went wrong during the export of the configured image**s**. The concrete error is in _err_.

test/integration/features/basic.feature
```diff
@@ -15,7 +15,7 @@ Feature: Basic
     When executing "minishift addons enable anyuid" succeeds
     Then stdout should contain
      """
-     Addon 'anyuid' enabled
+     Add-on 'anyuid' enabled
```
hferentschik: +1, no need for a second commit though. You can just squash the commits.

cmd/minishift/cmd/addon/addon_apply.go
thatdocslady: No, you either run the command or apply the add-on. Because you used "execute" and "command", I updated the phrasing to reflect that. If you want me to change it to "apply", then I need to remove the word "command" because it's incorrect.

cmd/minishift/cmd/addon/addon_enable.go
thatdocslady: Passive voice usage isn't recommended (tech writing 101, let's talk about it someday), and the meaning is basically the same so I don't see why we need to rephrase it.

cmd/minishift/cmd/image/export.go
thatdocslady: Again, you're not clarifying how to determine which images to export and how can users decide whether they want to export one, several, or all images.

cmd/minishift/cmd/addon/addon_apply.go
hferentschik: I don't think there is a problem here. An add-on command is one of the documented DSL commands. And I think we aligned on that we apply add-ons and their commands.

cmd/minishift/cmd/addon/addon_enable.go
hferentschik: Because it is not quite correct. This is does not determine the order of this single add-on. To determine an order I need to take more than one thing into account.

cmd/minishift/cmd/image/export.go
hferentschik: Right now the user does not anything with this command. In fact it is hidden. Image coaching is experimental. Either way, if you look at the if statement, it just checks that there is at least one argument, aka one image spec. There can be multiple.

cmd/minishift/cmd/addon/addon.go
hferentschik: I am missing the ambiguity here. It refers to the command, right?  It renders like this:

    Available Commands:
       addons      Manages Minishift add-ons with its sub-commands.

hferentschik: so _its_ refers to the 'addon command'. Where is the  ambiguity?

cmd/minishift/cmd/addon/addon_apply.go
hferentschik: Actually both messages are not quite right, the first should be:

    atexit.ExitWithMessage(1, fmt.Sprintf("Error creating execution context for applying addon %s: %v", addonName, err))

And the second:

    atexit.ExitWithMessage(1, fmt.Sprintf("Error applying addon %s: %v", addonName, err))

cmd/minishift/cmd/addon/addon.go
LalatenduMohanty: IMO we do not need to mention sub-commands because the context is correct when we say  `Manages Minishift add-ons`. Anyway the user will see the sub-commands in the usage of `minishift addons --help`. 

Without this change the the short description shows up in `minishift` usage i.e. `minishift --help` which looks correct to me.
```
$ minishift 
Minishift is a command-line tool that provisions and manages single-node OpenShift clusters optimized for development workflows.

Usage:
  minishift [command]

Available Commands:
  addons      Manages Minishift add-ons
  config      Modifies Minishift configuration properties.
  console     Opens or displays the OpenShift Web Console URL.
  delete      Deletes the Minishift VM.
  docker-env  Sets Docker environment variables.
  help        Help about any command
  hostfolder  Manages host folders for use by the OpenShift cluster.
  ip          Gets the IP address of the running cluster.
  logs        Gets the logs of the running OpenShift cluster.
  oc-env      Sets path of 'oc' binary.
  openshift   Interact with an Openshift Cluster
  ssh         Log in to or run a command on a Minishift VM with SSH.
  start       Starts a local OpenShift cluster.
  status      Gets the status of the local OpenShift cluster.
  stop        Stops the running local OpenShift cluster.
  update      Update to latest version of Minishift.
  version     Gets the version of Minishift.

Flags:
  -h, --help                             help for minishift
      --show-libmachine-logs             Show logs from libmachine.
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
  -v, --v Level                          log level for V logs
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging

Use "minishift [command] --help" for more information about a command.
```

cmd/minishift/cmd/addon/addon_apply.go
LalatenduMohanty: Agree with @hferentschik. Lets fix this to 
```
atexit.ExitWithMessage(1, fmt.Sprintf("Error applying add-on %s: %v", addonName, err))
```

cmd/minishift/cmd/addon/addon_apply.go
LalatenduMohanty: @hferentschik What you suggested iis technically correct. The reason I had written this because I wanted an easier message  to understand.  Do you think we need to change the current message? or we can just leave it as it is.  

@thatdocslady by add-on command we mean https://docs.openshift.org/latest/minishift/using/addons.html#addon-commands , I think `executing addon commands` is ok.  However you can replace it with `apply`  and thats fine too.

cmd/minishift/cmd/addon/addon_enable.go
LalatenduMohanty: @thatdocslady `Determines the order in which to apply the add-on.`  does not sound right. Also it makes it difficult to understand the actual meaning for sure.

Just to be clear (as @hferentschik mentioned) `priority` is something comes to play when we have enabled mutiple add-ons and we want the add-ons to apply in a specific order as there might be dependency issues.

cmd/minishift/cmd/image/export.go
LalatenduMohanty: +1.

cmd/minishift/cmd/addon/addon_enable.go
hferentschik: A single priority cannot determine the order of a single add-on. More than one add-ons are needed to establish an order.

cmd/minishift/cmd/addon/addon.go
hferentschik: > IMO we do not need to mention sub-commands because the context is correct when we say Manages Minishift add-ons. Anyway the user will see the sub-commands in the usage of minishift addons --help.

IMO it gives valuable context. I can look at `minishift -h` and see directly which commands have subcommands. But I am ok, with not having it.

cmd/minishift/cmd/addon/addon_apply.go
hferentschik: > The reason I had written this because I wanted an easier message to understand. Do you think we need to change the current message? or we can just leave it as it is.

I think it needs to change. Whether we mention execution context or whether we got with the more generic "Error applying add-on %s: %v" I don't mind.

thatdocslady:
@hferentschik @LalatenduMohanty implemented the latest feedback, hopefully this wraps up the review.

hferentschik:
retest this please

cmd/minishift/cmd/addon/addon_enable.go
```diff
@@ -38,7 +38,7 @@ var addonsEnableCmd = &cobra.Command{
 }
 
 func init() {
-	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of the add-on when it is applied.")
+	addonsEnableCmd.Flags().IntVar(&priority, priorityFlag, 0, "The priority of the add-on. If multiple add-ons are enabled, this value is used by Minishift to determine the order in which to apply the add-ons.")
```
hferentschik: Just "The priority of the add-on.". The second sentence is actually not accurate and also to long.

cmd/minishift/cmd/addon/addon_install.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	unspecifiedSourceError   = "You need to specify the source for the add-on."
+	unspecifiedSourceError   = "You must specify the source directory path of the add-on."
```
hferentschik: It intentionally only says source here without further qualifications. We are about to implement other source types. "You must specify the source of the add-on."

cmd/minishift/cmd/openshift/registry.go
```diff
@@ -29,8 +29,8 @@ import (
 
 var registryCmd = &cobra.Command{
 	Use:   "registry [flags]",
-	Short: "Prints host and port of the OpenShift registry.",
-	Long:  `Prints host and port of the OpenShift docker registry in the format 'host:port'.`,
+	Short: "Prints the host name and port number of the OpenShift registry to the console.",
```
hferentschik: "to standard output". This is also what we use in for example the `ip` command. The output goes to standard output which per default will be the console.

cmd/minishift/cmd/openshift/restart.go
```diff
@@ -33,7 +33,7 @@ import (
 var restartCmd = &cobra.Command{
 	Use:   "restart",
 	Short: "Restarts the OpenShift cluster.",
-	Long:  "Restarts the OpenShift cluster, keeping the VM alive.",
+	Long:  "Restarts the OpenShift cluster, and maintains the running state of the Minishift VM.",
```
hferentschik: just "state", running is not required

cmd/minishift/cmd/openshift/set.go
```diff
@@ -37,8 +37,8 @@ const (
 	patchFlag  = "patch"
 
 	unknownPatchTargetError = "Unkown patch target. Only 'master' and 'node' are supported."
-	emptyPatchError         = "You need to specify a patch using --patch."
-	invalidJSONError        = "The specified patch need to be valid JSON."
+	emptyPatchError         = "You must specify a patch using the --patch flag."
+	invalidJSONError        = "You must specify a patch with a valid JSON format."
```
hferentschik: " with a valid JSON format" sounds wrong to me. "Needs to be valid JSON". Also there is only one valid JSON format, not multiple which 'a' would imply.

cmd/minishift/cmd/util/util.go
```diff
@@ -34,15 +34,15 @@ var (
...
+		atexit.ExitWithMessage(1, "Cannot determine the state of Minishift VM.")
 	}
 	return exists
 }
 
 func ExitIfUndefined(client *libmachine.Client, machineName string) {
 	exists := VMExists(client, machineName)
 	if !exists {
-		atexit.ExitWithMessage(0, fmt.Sprintf("The execution of this command requires an existing '%s' VM. However, there is currently none defined.", machineName))
+		atexit.ExitWithMessage(0, fmt.Sprintf("Running this command requires an existing '%s' VM, but no VMs are currently defined.", machineName))
```
hferentschik: "but no VMs are currently defined", why VMs. We are dealing with one VM.

cmd/minishift/cmd/addon/addon_install.go
LalatenduMohanty: @hferentschik I think we should make the change `You must specify the source of the add-on.` when we add the code to implement other source types.

cmd/minishift/cmd/openshift/restart.go
LalatenduMohanty: `without changing the state of the Minishift VM." might be a better phrase , but not sure.

cmd/minishift/cmd/addon/addon_install.go
hferentschik: This is an error message. I bet you it will be forgotten then. There is hardly any loss in clarity here by being less specific. Also, installing from URL is high on the trio list. I would expect this to land in the next sprint or so. What's the point of "toggling" the message then?

cmd/minishift/cmd/openshift/restart.go
hferentschik: that works also. I think the "running" part is what makes it sound wrong.

thatdocslady: I'll go with @hferentschik suggestion since it keeps a positive voice and is more concise.

cmd/minishift/cmd/openshift/set.go
thatdocslady: The article 'a' isn't a quantifier in this case, rather it makes it explicit that the JSON format needs to be valid. However, to help be even more explicit, we can write something like "a valid JSON file". Simply writing "valid JSON" is incomplete grammatically and needs a qualifier.

cmd/minishift/cmd/util/util.go
thatdocslady: Ok this wasn't clear from the original text, will fix.

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -40,8 +40,8 @@ const (
 
 var addonsApplyCmd = &cobra.Command{
 	Use:   "apply ADDON_NAME ...",
-	Short: "Executes the specified add-ons.",
-	Long:  "Executes the specified add-ons. The command works with both enabled and disabled add-ons.",
+	Short: "Applies the specified add-ons.",
+	Long:  "Applies the specified add-ons. You can run this command on enabled or disabled add-ons.",
```
thatdocslady: @LalatenduMohanty none of what you're saying is clear from the original text. This information wasn't available to me when I edited this string, and most of the time the word "and" is incorrectly used when "or" should be used (meaning, you can run it on "any").

Now that you added this new information, I suggest the following: "You can specify more than one add-on to apply. The add-ons can be enabled or disabled."

This way it's more clear that you can apply multiple add-ons and they can have different states. WDYT?

cmd/minishift/cmd/addon/addon.go
```diff
@@ -23,15 +23,15 @@ import (
...
 	noAddOnMessage  = "No add-on with the name %s is installed."
 )
 
 var AddonsCmd = &cobra.Command{
 	Use:     "addons SUBCOMMAND [flags]",
 	Aliases: []string{"addon"},
-	Short:   "Manages Minishift add-ons",
-	Long:    "Manages Minishift add-ons. You can install, list, enable or disable Minishift add-ons.",
+	Short:   "Manages Minishift add-ons.",
+	Long:    "Manages Minishift add-ons. Use the sub-commands to install, list, enable or disable Minishift add-ons.",
```
hferentschik: Still think it should be "its sub-commands". "the" is w/o reference here imo

LalatenduMohanty: I am ok with this as the `Long:` shows up in the `minishift addons --help`.

thatdocslady: I think there is enough context here to indicate that the sub-commands belong to the parent command, especially if you see it in the --help. The word "its" is not recommended in docs because it's a possessive contraction and usually used to indicate ownership (and commands, unlike humans, can't "own" things). I already removed some instances of that word across the docs in past reviews.

LalatenduMohanty:
@thatdocslady @hferentschik Looks fine to me now.

hferentschik:
Merged, thanks.

Issue #1078, opened 2017-06-28
Feedback when clearing cache


gbraad:
The command `minishift delete --clear-cache` does not provide feedback about the action happening or not.

certainlynotbarton:
Can I have this as my first-time contribution ?

https://github.com/minishift/minishift/blob/7544af22b6bd3953d516d3fc2f49827b04093ab0/cmd/minishift/cmd/delete.go#L63

gbraad:
Worked on by @dharmit

hferentschik:
Merged via pull request #1089

PR #1089, opened 2017-06-30
Issue #1078 Provide feedback about cache getting deleted

dharmit:
Addresses issue #1078

### DIFF
--- a/cmd/minishift/cmd/delete.go
+++ b/cmd/minishift/cmd/delete.go
@@ -18,8 +18,8 @@ package cmd
 
 import (
 	"fmt"
-
 	"os"
+	"strings"
 
 	"github.com/docker/machine/libmachine"
 	"github.com/minishift/minishift/cmd/minishift/cmd/util"
@@ -30,7 +30,6 @@ import (
 	"github.com/minishift/minishift/pkg/util/filehelper"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
-	"strings"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


cmd/minishift/cmd/delete.go
```diff
@@ -61,6 +60,7 @@ func runDelete(cmd *cobra.Command, args []string) {
 	}
 
 	if clearCache {
+		fmt.Println("Deleting Minishift cache")
```
LalatenduMohanty: I think we need to be more verbose than that e.g. We should say what files are getting deleted or atleast which directories.

hferentschik:
add to whitelist

cmd/minishift/cmd/delete.go
```diff
@@ -61,6 +60,7 @@ func runDelete(cmd *cobra.Command, args []string) {
 	}
 
 	if clearCache {
+		fmt.Printf("Deleting Minishift ISO, OpenShift container images and OpenShift origin binaries from %s/cache\n", constants.Minipath)
```
hferentschik: hmm, whether you have images will depend really. I thought more about literally logging what gets deleted. The concrete file names. Not sure what @LalatenduMohanty had in mind.

LalatenduMohanty: @dharmit @hferentschik I think we would be fine with the absolute path of the cache directory as mentioning individual files will make the output cluttered. I modified the code little bit (as below) and it looks fine to me.
```
if clearCache {
		cachePath := constants.MakeMiniPath("cache")
		err := os.RemoveAll(cachePath)
		if err != nil {
			atexit.ExitWithMessage(1, fmt.Sprintf("Error deleting Minishift cache: %v", err))
		} else {
			fmt.Printf("Removed the cache at: %s\n", cachePath)
		}
	}
```
I moved the print statement after deletion as I think saying deleting the cache would means we should again mention that it is deleted successfully.  Also as user has explicitly used --clear-cache we can just mention after deleting and no need to say we are deleting the cache. 
```
$ minishift delete --clear-cache
Removed the cache at: /home/lmohanty/.minishift/cache
Deleting the Minishift VM...
Minishift VM deleted.
```

gbraad: The message should be concise (and simple): "Removed cache", or as suggested above "Removed cache at ~/.minishift/cache"

hferentschik:
Merged, thanks @dharmit for another contribution. Keep it coming :-)

PR #1093, opened 2017-07-03
Issue #876 Update 'openshift service list' command to handle NodePorts

### LINKED ISSUE #876: Update 'openshift service list' command to handle NodePort type
This is a follow up issue of [Update service command to be based on app routes](https://github.com/minishift/minishift/issues/342) to also serve URLs in the format of NodePort.

praveenkumar:
Fix #876 

Using current PR service list now looks like

```
$ ./minishift openshift service list
|-----------|-----------------|----------------------------------------------------|---------------------------|
| NAMSEPACE |      NAME       |                     ROUTE-URL                      |         NODEPORT          |
|-----------|-----------------|----------------------------------------------------|---------------------------|
| default   | docker-registry | No route for service                               | No node port              |
| default   | kubernetes      | No route for service                               | No node port              |
| default   | router          | No route for service                               | No node port              |
| myproject | simpleservice   | http://simpleservice-myproject.192.168.42.8.nip.io | No node port              |
| myproject | testservice     | No route for service                               | No node port              |
| myproject | mariadb         | No route for service                               | No node port              |
| myproject | mariadb-ingress | No route for service                               | http://192.168.42.8:31145 |
|-----------|-----------------|----------------------------------------------------|---------------------------|
```

### DIFF
--- a/cmd/minishift/cmd/openshift/service.go
+++ b/cmd/minishift/cmd/openshift/service.go
@@ -20,11 +20,13 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/pkg/browser"
-	"github.com/spf13/cobra"
-
+	"github.com/docker/machine/libmachine"
+	"github.com/minishift/minishift/cmd/minishift/cmd/util"
+	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/openshift"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/pkg/browser"
+	"github.com/spf13/cobra"
 )
 
 var (
@@ -39,13 +41,18 @@ var serviceCmd = &cobra.Command{
 	Short: "Opens the URL for the specified service in the browser or prints it to the console.",
 	Long:  `Opens the URL for the specified service and namespace in the default browser or prints it to the console. If no namespace is provided, 'default' is assumed.`,
 	Run: func(cmd *cobra.Command, args []string) {
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		defer api.Close()
+
+		util.ExitIfUndefined(api, constants.MachineName)
+
 		if len(args) == 0 || len(args) > 1 {
 			atexit.ExitWithMessage(1, "You must specify the name of the service.")
 		}
 
 		service := args[0]
 
-		url, err := openshift.GetServiceURL(service, namespace, https)
+		url, err := openshift.GetServiceSpec(service, namespace, https)
 		if err != nil {
 			atexit.ExitWithMessage(1, err.Error())
 		}
--- a/cmd/minishift/cmd/openshift/service_list.go
+++ b/cmd/minishift/cmd/openshift/service_list.go
@@ -17,8 +17,13 @@ limitations under the License.
 package openshift
 
 import (
+	"fmt"
 	"os"
+	"strings"
 
+	"github.com/docker/machine/libmachine"
+	"github.com/minishift/minishift/cmd/minishift/cmd/util"
+	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/openshift"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/olekukonko/tablewriter"
@@ -33,18 +38,52 @@ var serviceListCmd = &cobra.Command{
 	Short: "Gets the URLs of the services in your local OpenShift cluster.",
 	Long:  `Gets the URLs of the services in your local OpenShift cluster.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		urls, err := openshift.GetServiceURLs(serviceListNamespace)
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		defer api.Close()
+
+		util.ExitIfUndefined(api, constants.MachineName)
+
+		host, err := api.Load(constants.MachineName)
+		if err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		}
+
+		util.ExitIfNotRunning(host.Driver, constants.MachineName)
+
+		ip, err := host.Driver.GetIP()
+		if err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error getting IP: %s", err.Error()))
+		}
+
+		serviceSpecs, err := openshift.GetServiceSpecs(serviceListNamespace)
 		if err != nil {
 			atexit.ExitWithMessage(1, err.Error())
 		}
 
 		var data [][]string
-		for _, url := range urls {
-			data = append(data, []string{url.Namespace, url.Name, url.URL})
+		namespace := make(map[string]bool)
+		for _, serviceSpec := range serviceSpecs {
+			if _, ok := namespace[serviceSpec.Namespace]; ok {
+				serviceSpec.Namespace = ""
+			} else {
+				namespace[serviceSpec.Namespace] = true
+			}
+			var urls, weights string
+			nodePortURL := serviceSpec.NodePort
+			if nodePortURL != "" {
+				nodePortURL = fmt.Sprintf("%s:%s", ip, nodePortURL)
+			}
+			if serviceSpec.URL != nil {
+				urls = strings.Join(serviceSpec.URL, "\n")
+			}
+			if serviceSpec.Weight != nil {
+				weights = strings.Join(serviceSpec.Weight, "\n")
+			}
+			data = append(data, []string{serviceSpec.Namespace, serviceSpec.Name, nodePortURL, urls, weights})
 		}
 
 		table := tablewriter.NewWriter(os.Stdout)
-		table.SetHeader([]string{"Namsepace", "Name", "URL"})
+		table.SetHeader([]string{"Namsepace", "Name", "NodePort", "Route-URL", "Weight"})
 		table.SetBorders(tablewriter.Border{Left: true, Top: true, Right: true, Bottom: true})
 		table.SetCenterSeparator("|")
 		table.AppendBulk(data) // Add Bulk Data
@@ -53,6 +92,6 @@ var serviceListCmd = &cobra.Command{
 }
 
 func init() {
-	serviceListCmd.Flags().StringVarP(&serviceListNamespace, "namespace", "n", "default", "The namespace of the services.")
+	serviceListCmd.Flags().StringVarP(&serviceListNamespace, "namespace", "n", "", "The namespace of the services.")
 	serviceCmd.AddCommand(serviceListCmd)
 }
--- a/pkg/minishift/openshift/service.go
+++ b/pkg/minishift/openshift/service.go
@@ -18,13 +18,18 @@ package openshift
 
 import (
 	"fmt"
+	"net/url"
+	"path/filepath"
+	"strconv"
+	"strings"
+
+	"encoding/json"
+
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	instanceState "github.com/minishift/minishift/pkg/minishift/config"
 	"github.com/minishift/minishift/pkg/util"
 	"github.com/pkg/errors"
-	"path/filepath"
-	"regexp"
-	"strings"
+	"sort"
 )
 
 var (
@@ -36,69 +41,111 @@ func init() {
 	systemKubeConfigPath = filepath.Join(constants.Minipath, "machines", constants.MachineName+"_kubeconfig")
 }
 
-type ServiceURL struct {
+type Service struct {
+	Items []struct {
+		Metadata struct {
+			Name string `json:"name"`
+		} `json:"metadata"`
+		Spec struct {
+			Ports []struct {
+				NodePort int `json:"nodePort"`
+			} `json:"ports"`
+		} `json:"spec"`
+	} `json:"items"`
+}
+
+type Route struct {
+	Items []struct {
+		Spec struct {
+			AlternateBackends []struct {
+				Name   string `json:"name"`
+				Weight int    `json:"weight"`
+			} `json:"alternateBackends"`
+			Host string `json:"host"`
+			To   struct {
+				Name   string `json:"name"`
+				Weight int    `json:"weight"`
+			} `json:"to"`
+		} `json:"spec"`
+	} `json:"items"`
+}
+
+type ServiceWeight struct {
+	Name   string
+	Weight string
+}
+
+type ServiceSpec struct {
 	Namespace string
 	Name      string
-	URL       string
+	URL       []string
+	NodePort  string
+	Weight    []string
 }
 
 const (
-	URLCustomCol      = "-o=custom-columns=URL:.spec.host"
-	URLsCustomCol     = "-o=custom-columns=NAME:.metadata.name,HOST:.spec.host"
 	ProjectsCustomCol = "-o=custom-columns=NAME:.metadata.name"
 )
 
 // Get the route for service
-func GetServiceURL(service, namespace string, https bool) (string, error) {
+func GetServiceSpec(service, namespace string, https bool) (string, error) {
 	urlScheme := "http://"
 	if https {
 		urlScheme = "https://"
 	}
 
-	if !isProjectExists(namespace) {
-		return "", errors.New(fmt.Sprintf("Namespace %s doesn't exits", namespace))
-	}
-
-	cmdArgText := fmt.Sprintf("get route/%s -n %s --config=%s %s", service, namespace, systemKubeConfigPath, URLCustomCol)
-	tokens := strings.Split(cmdArgText, " ")
-	cmdName := instanceState.InstanceConfig.OcPath
-	cmdOut, err := runner.Output(cmdName, tokens...)
+	serviceSpecs, err := GetServiceSpecs(namespace)
 	if err != nil {
-		return "", errors.New(fmt.Sprintf("No service with name '%s' defined in namespace '%s'", service, namespace))
+		return "", err
 	}
 
-	url := strings.Split(byteArrayToString(cmdOut), "\n")[1] // second element contain actual URL content
-	return urlScheme + url, nil
+	for _, serviceSpec := range serviceSpecs {
+		sort.Strings(serviceSpec.URL)
+		if serviceSpec.Name == service {
+			if serviceSpec.URL != nil {
+				u, _ := url.Parse(serviceSpec.URL[0])
+				return urlScheme + u.Host, nil
+			} else {
+				return "", errors.New(fmt.Sprintf("Service '%s' in namespace '%s' does not have route associated which can be opened in the browser.", service, namespace))
+			}
+		}
+	}
+
+	return "", errors.New(fmt.Sprintf("Service %s does not exist in namespace %s.", service, namespace))
 }
 
-// Get the available routes to user
-func GetServiceURLs(serviceListNamespace string) ([]ServiceURL, error) {
-	var serviceURLs []ServiceURL
+// GetServiceSpecs takes Namespace string and return route/nodeport/name/weight in a ServiceSpec structure
+func GetServiceSpecs(serviceNamespace string) ([]ServiceSpec, error) {
+	var serviceSpecs []ServiceSpec
 
-	if serviceListNamespace != "" && !isProjectExists(serviceListNamespace) {
-		return serviceURLs, errors.New(fmt.Sprintf("Namespace %s doesn't exits", serviceListNamespace))
+	if serviceNamespace != "" && !isProjectExists(serviceNamespace) {
+		return serviceSpecs, errors.New(fmt.Sprintf("Namespace %s doesn't exits", serviceNamespace))
 	}
 
-	namespaces, err := getValidNamespaces(serviceListNamespace)
+	namespaces, err := getValidNamespaces(serviceNamespace)
 	if err != nil {
-		return serviceURLs, err
+		return serviceSpecs, err
 	}
 
-	// iterate over namespaces, get command output, format route in ServiceURL
+	// iterate over namespaces, get command output, format route and nodePort
 	for _, namespace := range namespaces {
-		outputData, err := getServiceURLsOutput(namespace)
+		serviceNodePort, err := getService(namespace)
 		if err != nil {
-			return serviceURLs, err
+			return serviceSpecs, err
 		}
-		if !strings.Contains(outputData, "No resources found") {
-			serviceURLs = filterAndUpdateServiceURLS(outputData, namespace)
+		routeSpec, err := getRouteSpec(namespace)
+		if err != nil {
+			return serviceSpecs, err
+		}
+		if serviceNodePort != nil {
+			serviceSpecs = append(serviceSpecs, filterAndUpdateServiceSpecs(routeSpec, namespace, serviceNodePort)...)
 		}
 	}
-	if len(serviceURLs) == 0 {
-		return serviceURLs, errors.New(fmt.Sprintf("No services defined in namespace '%s'", serviceListNamespace))
+	if len(serviceSpecs) == 0 {
+		return serviceSpecs, errors.New(fmt.Sprintf("No services defined in namespace '%s'", serviceNamespace))
 	}
 
-	return serviceURLs, nil
+	return serviceSpecs, nil
 }
 
 // Check whether project exists or not
@@ -136,33 +183,94 @@ func getValidNamespaces(serviceListNamespace string) ([]string, error) {
 	return namespaces, nil
 }
 
-func getServiceURLsOutput(namespace string) (string, error) {
-	cmdArgText := fmt.Sprintf("get route -n %s --config=%s %s", namespace, systemKubeConfigPath, URLsCustomCol)
+// getRouteSpec take valid namespace and return map which have route as key and
+// value be ServiceWeight struct
+func getRouteSpec(namespace string) (map[string][]ServiceWeight, error) {
+	routeSpec := make(map[string][]ServiceWeight)
+	cmdArgText := fmt.Sprintf("get route -o json --config=%s -n %s", systemKubeConfigPath, namespace)
 	tokens := strings.Split(cmdArgText, " ")
 	cmdName := instanceState.InstanceConfig.OcPath
 	cmdOut, err := runner.Output(cmdName, tokens...)
 	if err != nil {
-		return "", err
+		return nil, err
 	}
 
-	return byteArrayToString(cmdOut), nil
+	if strings.Contains(string(cmdOut), "No resources found") {
+		return nil, nil
+	}
+
+	var data Route
+	err = json.Unmarshal(cmdOut, &data)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, item := range data.Items {
+		totalWeight := item.Spec.To.Weight
+		for _, alternateBackend := range item.Spec.AlternateBackends {
+			totalWeight += alternateBackend.Weight
+		}
+		routeSpec[item.Spec.Host] = []ServiceWeight{{Name: item.Spec.To.Name, Weight: calculateWeight(float64(item.Spec.To.Weight), float64(totalWeight))}}
+		for _, alternateBackend := range item.Spec.AlternateBackends {
+			routeSpec[item.Spec.Host] = append(routeSpec[item.Spec.Host], ServiceWeight{Name: alternateBackend.Name,
+				Weight: calculateWeight(float64(alternateBackend.Weight), float64(totalWeight))})
+		}
+	}
+	return routeSpec, nil
 }
 
-func filterAndUpdateServiceURLS(data, namespace string) []ServiceURL {
-	var serviceURLs []ServiceURL
+func calculateWeight(a float64, b float64) string {
+	weightPercentage := int(a / b * 100)
+	if weightPercentage != 100 {
+		return strconv.Itoa(weightPercentage) + "%"
+	}
+	return ""
+}
 
-	re_whtsp_inside := regexp.MustCompile(`[\s\p{Zs}]{2,}`)
-	data = re_whtsp_inside.ReplaceAllString(data, " ") // replace all extra whitespaces
-	contents := strings.Split(data, "\n")              // split on new lines
-	contents = emptyFilter(contents[1:])               // remove the header "NAME HOST" and empty elements
+func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace string, serviceNodePort map[string]string) []ServiceSpec {
+	var serviceSpecs []ServiceSpec
+	serviceSpecMap := make(map[string]*ServiceSpec)
 
-	for _, content := range contents {
-		// split content on white space to separate NAME and HOST
-		data := strings.Split(content, " ")
-		serviceURLs = append(serviceURLs, ServiceURL{Namespace: namespace, Name: data[0], URL: "http://" + data[1]})
+	for routeURL, v := range routeSpec {
+		for _, service := range v {
+			// split content on colon to separate NAME, HOST and Weight
+			if _, ok := serviceNodePort[service.Name]; ok {
+				if _, ok := serviceSpecMap[service.Name]; !ok {
+					serviceSpecMap[service.Name] = &ServiceSpec{Namespace: namespace,
+						Name:     service.Name,
+						URL:      []string{fmt.Sprintf("http://%s", routeURL)},
+						NodePort: serviceNodePort[service.Name],
+						Weight:   []string{service.Weight},
+					}
+				} else {
+					serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf("http://%s", routeURL))
+					serviceSpecMap[service.Name].Weight = append(serviceSpecMap[service.Name].Weight, service.Weight)
+				}
+				delete(serviceNodePort, service.Name)
+			} else {
+				if _, ok := serviceSpecMap[service.Name]; !ok {
+					serviceSpecMap[service.Name] = &ServiceSpec{Namespace: namespace,
+						Name:     service.Name,
+						URL:      []string{fmt.Sprintf("http://%s", routeURL)},
+						NodePort: "",
+						Weight:   []string{service.Weight},
+					}
+				} else {
+					serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf("http://%s", routeURL))
+					serviceSpecMap[service.Name].Weight = append(serviceSpecMap[service.Name].Weight, service.Weight)
+				}
+			}
+		}
 	}
 
-	return serviceURLs
+	for k, v := range serviceNodePort {
+		serviceSpecs = append(serviceSpecs, ServiceSpec{Namespace: namespace, Name: k, URL: nil, NodePort: v, Weight: nil})
+	}
+	for _, v := range serviceSpecMap {
+		serviceSpecs = append(serviceSpecs, *v)
+	}
+
+	return serviceSpecs
 }
 
 // Discard empty elements
@@ -178,6 +286,39 @@ func emptyFilter(data []string) []string {
 	return res
 }
 
+// Get service provide service name and node-port as map data structure
+func getService(namespace string) (map[string]string, error) {
+	serviceNodePort := make(map[string]string)
+	cmdArgText := fmt.Sprintf("get svc -o json --config=%s -n %s", systemKubeConfigPath, namespace)
+	tokens := strings.Split(cmdArgText, " ")
+	cmdName := instanceState.InstanceConfig.OcPath
+	cmdOut, err := runner.Output(cmdName, tokens...)
+	if err != nil {
+		return nil, err
+	}
+
+	if strings.Contains(string(cmdOut), "No resources found") {
+		return nil, nil
+	}
+
+	var data Service
+	err = json.Unmarshal(cmdOut, &data)
+	if err != nil {
+		return nil, err
+	}
+
+	for _, item := range data.Items {
+		for _, port := range item.Spec.Ports {
+			nodePort := ""
+			if port.NodePort != 0 {
+				nodePort = fmt.Sprintf("%d", port.NodePort)
+			}
+			serviceNodePort[item.Metadata.Name] = nodePort
+		}
+	}
+	return serviceNodePort, nil
+}
+
 // Get all projects a user belongs to
 func getProjects() ([]string, error) {
 	cmdArgText := fmt.Sprintf("get projects --config=%s %s", systemKubeConfigPath, ProjectsCustomCol)
@@ -188,11 +329,6 @@ func getProjects() ([]string, error) {
 		return []string{}, err
 	}
 
-	contents := strings.Split(byteArrayToString(cmdOut), "\n")
+	contents := strings.Split(string(cmdOut), "\n")
 	return emptyFilter(contents[1:]), nil
-}
-
-// Convert byte array to string
-func byteArrayToString(data []byte) string {
-	return string(data)
 }
--- a/pkg/minishift/openshift/service_test.go
+++ b/pkg/minishift/openshift/service_test.go
@@ -1,0 +1,177 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openshift
+
+import (
+	"fmt"
+	"sort"
+	"testing"
+
+	instanceState "github.com/minishift/minishift/pkg/minishift/config"
+	test "github.com/minishift/minishift/pkg/testing"
+	testdata "github.com/minishift/minishift/pkg/testing/testdata"
+)
+
+//
+type ByName []ServiceSpec
+
+func (a ByName) Len() int           { return len(a) }
+func (a ByName) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
+func (a ByName) Less(i, j int) bool { return a[i].Name < a[j].Name }
+
+func Test_getServiceSpec_if_service_exist_with_route(t *testing.T) {
+	namespace := "bar"
+	service := "guestbook-v1"
+	https := false
+
+	setup(t, namespace)
+	defer teardown()
+
+	got, err := GetServiceSpec(service, namespace, https)
+
+	if err != nil {
+		t.Fatalf("%+v", err)
+	}
+
+	expected := "http://guestbook-myproject.192.168.64.2.nip.io"
+	if https {
+		expected = "https://guestbook-myproject.192.168.64.2.nip.io"
+	}
+
+	if got != expected {
+		t.Fatalf("Expected: %+v, Got: %+v", expected, got)
+	}
+}
+
+func Test_getServiceSpec_if_service_exist_without_route(t *testing.T) {
+	namespace := "bar"
+	service := "guestbook-v1-np"
+	https := false
+
+	setup(t, namespace)
+	defer teardown()
+
+	_, err := GetServiceSpec(service, namespace, https)
+
+	if err == nil {
+		t.Fatalf("Error should be: %+v", err.Error())
+	}
+
+	expectedErrorText := fmt.Sprintf("Service '%s' in namespace '%s' does not have route associated which can be opened in the browser.", service, namespace)
+	if expectedErrorText != err.Error() {
+		t.Fatalf("Expected: %+v, Got: %+v", expectedErrorText, err.Error())
+	}
+
+}
+
+func Test_getServiceSpec_if_service_not_exist(t *testing.T) {
+	namespace := "bar"
+	service := "nonExistService"
+	https := false
+
+	setup(t, namespace)
+	defer teardown()
+
+	_, err := GetServiceSpec(service, namespace, https)
+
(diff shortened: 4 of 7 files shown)


hferentschik:
> Using current PR service list now looks like

Was not so hard after all, right ;-)

A few things looking at the output:

* `minishift openshift service list` does not list all services from all namespaces. It used to force you to specify a namespace. Is this intentional? Are you aligning with Minikube here? Really we have three choices 
  * Enforce specifying namespace (as it used to be)
  * List all namespaces (as it is now)
  * Use the current context of the user (I think this has been discussed as well and fair this was a preferred choice by some)
* I am not a big fan of the repetitive "No route for service " and "No node port". IMO that is just noise. I'd rather keep the column empty or have a less intrusive '-'.
* Using an URL for a NodePort is wrong. Looking at the example output you defend a NodePort for MaraiDB. This is not a URL.

@LalatenduMohanty, @budhrg, @gbraad, @jorgemoralespou,

cmd/minishift/cmd/openshift/service.go
```diff
@@ -39,6 +41,11 @@ var serviceCmd = &cobra.Command{
 	Short: "Opens the URL for the specified service in the browser or prints it to the console",
 	Long:  `Opens the URL for the specified service and namespace in the browser or prints it to the console. If no namespace is provided, 'default' is assumed.`,
 	Run: func(cmd *cobra.Command, args []string) {
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
```
hferentschik: 👍

cmd/minishift/cmd/openshift/service_list.go
```diff
@@ -33,18 +37,35 @@ var serviceListCmd = &cobra.Command{
 	Short: "Gets the URLs of the services in your local cluster.",
 	Long:  `Gets the URLs of the services in your local cluster.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		urls, err := openshift.GetServiceURLs(serviceListNamespace)
+		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
```
hferentschik: 👍

pkg/minishift/openshift/service.go
```diff
@@ -72,7 +89,7 @@ func GetServiceURL(service, namespace string, https bool) (string, error) {
 }
 
 // Get the available routes to user
-func GetServiceURLs(serviceListNamespace string) ([]ServiceURL, error) {
+func GetServiceURLs(serviceListNamespace string, ip string) ([]ServiceURL, error) {
```
hferentschik: Do we really have to pass the IP? This could be added at the table displaying level. We know that NodePorts are exposed in our context on the VM IP.

hferentschik: I think we should rename the _ServiceURL_ struct. It contains much more than a URL to a service. Maybe _ServiceSpec_?

hferentschik: Function needs to be renamed in this case as well.

pkg/minishift/openshift/service.go
```diff
@@ -86,12 +103,20 @@ func GetServiceURLs(serviceListNamespace string) ([]ServiceURL, error) {
 
 	// iterate over namespaces, get command output, format route in ServiceURL
 	for _, namespace := range namespaces {
-		outputData, err := getServiceURLsOutput(namespace)
+		serviceNodePort, err := getService(namespace, ip)
```
hferentschik: So now you make requests per namespace. You see not issue with this anymore. We talked about getting all service info in one go. Just wondering and  trying to understand your reasoning.

pkg/minishift/openshift/service.go
```diff
@@ -86,12 +103,20 @@ func GetServiceURLs(serviceListNamespace string) ([]ServiceURL, error) {
...
+		if err != nil {
+			fmt.Println("Here is errror")
```
hferentschik: I don't think you want to have this print statement in the merged code, right?

pkg/minishift/openshift/service.go
```diff
@@ -86,12 +103,20 @@ func GetServiceURLs(serviceListNamespace string) ([]ServiceURL, error) {
...
+			return serviceURLs, err
+		}
+		outputData, err := getRouteURLsOutput(namespace)
```
hferentschik: _routes_ instead of _ outputData_ and _getRoutes_ instead of _getRouteURLsOutput_ ?

pkg/minishift/openshift/service.go
```diff
@@ -136,7 +161,7 @@ func getValidNamespaces(serviceListNamespace string) ([]string, error) {
 	return namespaces, nil
 }
 
-func getServiceURLsOutput(namespace string) (string, error) {
+func getRouteURLsOutput(namespace string) (string, error) {
```
hferentschik: getRoute(s) or getRoute(s)ForNamespace.

hferentschik: Could there not be more than one route? How does that work?

pkg/minishift/openshift/service.go
```diff
@@ -148,7 +173,7 @@ func getServiceURLsOutput(namespace string) (string, error) {
 	return byteArrayToString(cmdOut), nil
```
hferentschik: hmm, I am not sure about the value of _byteArrayToString_. I think it is confusing for a Go programmer, after all it is just _string(data)_

pkg/minishift/openshift/service.go
```diff
@@ -178,6 +212,39 @@ func emptyFilter(data []string) []string {
...
+		return nil, err
+	}
+
+	if strings.Contains(byteArrayToString(cmdOut), "No resources found") {
+		return nil, nil
+	}
+
+	var data Service
+	err = json.Unmarshal(cmdOut, &data)
+	if err != nil {
```
hferentschik: what are the errors returned here. On a site note, I am concerned about our default mode of handling exceptions. Just checking for it and then passing it on. This might make sense in some cases, but we should also always try to look what type of error can get returned and whether we can do some better error handling.

pkg/minishift/openshift/service.go
```diff
@@ -178,6 +212,39 @@ func emptyFilter(data []string) []string {
...
+		return nil, err
+	}
+
+	for _, item := range data.Items {
+		for _, port := range item.Spec.Ports {
+			nodePort := "No node port"
+			if port.NodePort != 0 {
+				nodePort = fmt.Sprintf("http://%s:%d", ip, port.NodePort)
```
hferentschik: -1, Let's not munge data together to early. A ServiceSpec (or ServiceURL as it is called atm) struct should keep the data normalised. Than I can decide at display time how I want to display the data.

pkg/minishift/openshift/service.go
hferentschik: The godocs are not right

hferentschik: Also _getService_ has wrong godog. There it talk about returning JSON, which I  believe is not quite correct.

cmd/minishift/cmd/openshift/service_list.go
```diff
@@ -53,6 +74,6 @@ var serviceListCmd = &cobra.Command{
 }
 
 func init() {
-	serviceListCmd.Flags().StringVarP(&serviceListNamespace, "namespace", "n", "default", "The namespace of the services.")
+	serviceListCmd.Flags().StringVarP(&serviceListNamespace, "namespace", "n", "", "The namespace of the services.")
```
budhrg: @hferentschik might have some opinion keeping `default` as default namespace as it was added by him when I kept it empty in my last PR.

jorgemoralespou:
@praveenkumar it would be good to try some weird (but possible) usecases, to see how they look.

- A/B routing: A route with weighted weights to 2+ services
- A service with multiple routes
- Multiple nodeports for a service (a service with multiple ports exposed as nodePorts)

Also:
I agree with @hferentschik that nodePort is not a url but an IP and a port. And I think that the command should use current namespace by default (if that's the case) and specify an optional namespace or "--all-namespaces --as=system:admin" to get them all (although these are oc command options).

pkg/minishift/openshift/service.go
praveenkumar: @hferentschik It actually not the routes url but contain string output which have route name and it's url so I think instead `outputData` I will change it to `routeSpec` and `getRouteSpec`.

pkg/minishift/openshift/service.go
praveenkumar: I just removed this method and used standard `string(byte_data)`

praveenkumar:
> it would be good to try some weird (but possible) usecases, to see how they look.
    A/B routing: A route with weighted weights to 2+ services
    A service with multiple routes
    Multiple nodeports for a service (a service with multiple ports exposed as nodePorts)

Can you point me some existing service templates to try it out? you can also use https://ci.appveyor.com/project/hferentschik/minishift-o61ou/build/1879/artifacts (any platform specific binary) to test it out.

> nodePort is not a url but an IP and a port. 

Done

> I think that the command should use current namespace by default (if that's the case) and specify an optional namespace or "--all-namespaces --as=system:admin" to get them all (although these are oc command options).

I will work on this improvement.

jorgemoralespou:
@praveenkumar I have created you an example [here](https://gist.github.com/jorgemoralespou/333b784c4a9966fe67ed838028576e90)

This is the output I get:

````
./minishift-svc-test openshift service list
|-----------|-------------------|--------------------------------------------------------|---------------------------|
| NAMSEPACE |       NAME        |                       ROUTE-URL                        |         NODEPORT          |
|-----------|-------------------|--------------------------------------------------------|---------------------------|
| default   | docker-registry   |                                                      - |                         - |
| default   | kubernetes        |                                                      - |                         - |
| default   | router            |                                                      - |                         - |
| myproject | guestbook         | http://guestbook-myproject.192.168.64.2.nip.io         | 192.168.64.2:No node port |
| myproject | guestbook-v1      | http://guestbook-v1-myproject.192.168.64.2.nip.io      |                         - |
| myproject | guestbook-v1-3002 | http://guestbook-v1-3002-myproject.192.168.64.2.nip.io | 192.168.64.2:No node port |
| myproject | guestbook-v2      | http://guestbook-v2-myproject.192.168.64.2.nip.io      |                         - |
| myproject | guestbook-v2-3002 | http://guestbook-v2-3002-myproject.192.168.64.2.nip.io | 192.168.64.2:No node port |
| myproject | guestbook-v1-np   |                                                      - | 192.168.64.2:30757        |
| myproject | guestbook-v2-np   |                                                      - | 192.168.64.2:31819        |
|-----------|-------------------|--------------------------------------------------------|---------------------------|
````

And this is the services I have:
````
$ oc get services
NAME              CLUSTER-IP      EXTERNAL-IP   PORT(S)                      AGE
guestbook-v1      172.30.240.13   <none>        3000/TCP,3001/TCP,3002/TCP   4d
guestbook-v1-np   172.30.25.27    <nodes>       11001:30757/TCP              4d
guestbook-v2      172.30.149.49   <none>        3000/TCP,3001/TCP,3002/TCP   4d
guestbook-v2-np   172.30.230.95   <nodes>       11002:31819/TCP              4d
````


And the routes:
````
oc get routes
NAME                HOST/PORT                                         PATH      SERVICES                              PORT       TERMINATION   WILDCARD
guestbook           guestbook-myproject.192.168.64.2.nip.io                     guestbook-v1(50%),guestbook-v2(50%)   3000-tcp                 None
guestbook-v1        guestbook-v1-myproject.192.168.64.2.nip.io                  guestbook-v1                          3000-tcp                 None
guestbook-v1-3002   guestbook-v1-3002-myproject.192.168.64.2.nip.io             guestbook-v1                          3002-tcp                 None
guestbook-v2        guestbook-v2-myproject.192.168.64.2.nip.io                  guestbook-v2                          3000-tcp                 None
guestbook-v2-3002   guestbook-v2-3002-myproject.192.168.64.2.nip.io             guestbook-v2                          3002-tcp                 None
````

As you can see, this is not totally correct. The services are 4, not 7. The routes are 5 (which is correct) and node Ports are 2.

Also, see the typo in __NAMSEPACE__ header.

**Comment by praveenkumar** [comment]
> As you can see, this is not totally correct. The services are 4, not 7. The routes are 5 (which is correct) and node Ports are 2.

@hferentschik @jorgemoralespou Since we are displaying route and service both in a single table then it is as expected otherwise how we change the service display count? `NAME` field contain all the service/route name and sometime happen that service name and route name is same so it will only have single field otherwise different (in this case).


```
$ ./minishift openshift service list -n test
|-----------|-------------------|--------------------------------------------------------|--------------------|
| NAMSEPACE |       NAME        |                       ROUTE-URL                        |      NODEPORT      |
|-----------|-------------------|--------------------------------------------------------|--------------------|
| test      | guestbook         | http://guestbook-myproject.192.168.64.2.nip.io         |                  - |
| test      | guestbook-v1      | http://guestbook-v1-myproject.192.168.64.2.nip.io      |                  - |
| test      | guestbook-v1-3002 | http://guestbook-v1-3002-myproject.192.168.64.2.nip.io |                  - |
| test      | guestbook-v2      | http://guestbook-v2-myproject.192.168.64.2.nip.io      |                  - |
| test      | guestbook-v2-3002 | http://guestbook-v2-3002-myproject.192.168.64.2.nip.io |                  - |
| test      | guestbook-v1-np   |                                                      - | 192.168.42.8:31291 |
| test      | guestbook-v2-np   |                                                      - | 192.168.42.8:32093 |
|-----------|-------------------|--------------------------------------------------------|--------------------|

$ oc get service
NAME              CLUSTER-IP       EXTERNAL-IP   PORT(S)                      AGE
guestbook-v1      172.30.251.26    <none>        3000/TCP,3001/TCP,3002/TCP   51s
guestbook-v1-np   172.30.92.203    <nodes>       11001:31291/TCP              51s
guestbook-v2      172.30.178.125   <none>        3000/TCP,3001/TCP,3002/TCP   51s
guestbook-v2-np   172.30.102.172   <nodes>       11002:32093/TCP              51s

$ oc get route
NAME                HOST/PORT                                         PATH      SERVICES                              PORT       TERMINATION   WILDCARD
guestbook           guestbook-myproject.192.168.64.2.nip.io                     guestbook-v1(50%),guestbook-v2(50%)   3000-tcp                 None
guestbook-v1        guestbook-v1-myproject.192.168.64.2.nip.io                  guestbook-v1                          3000-tcp                 None
guestbook-v1-3002   guestbook-v1-3002-myproject.192.168.64.2.nip.io             guestbook-v1                          3002-tcp                 None
guestbook-v2        guestbook-v2-myproject.192.168.64.2.nip.io                  guestbook-v2                          3000-tcp                 None
guestbook-v2-3002   guestbook-v2-3002-myproject.192.168.64.2.nip.io             guestbook-v2                          3002-tcp                 None
```

**Comment by jorgemoralespou** [comment]
This should be the expected output, for that example:

````
$ ./minishift openshift service list -n test
|-----------|-------------------|--------------------------------------------------------|--------|--------------------|
| NAMSEPACE |       NAME        |                       ROUTE-URL                        | WEIGHT |      NODEPORT      |
|-----------|-------------------|--------------------------------------------------------|--------|--------------------|
| test      | guestbook-v1      | http://guestbook-myproject.192.168.64.2.nip.io         |   50%  |                  - |
| test      |                   | http://guestbook-v1-myproject.192.168.64.2.nip.io      |        |                  - |
| test      |                   | http://guestbook-v1-3002-myproject.192.168.64.2.nip.io |        |                  - |
| test      | guestbook-v2      | http://guestbook-myproject.192.168.64.2.nip.io         |   50%  |                  - |
| test      |                   | http://guestbook-v2-myproject.192.168.64.2.nip.io      |        |                  - |
| test      |                   | http://guestbook-v2-3002-myproject.192.168.64.2.nip.io |        |                  - |
| test      | guestbook-v1-np   |                                                      - |        | 192.168.42.8:31291 |
| test      | guestbook-v2-np   |                                                      - |        | 192.168.42.8:32093 |
|-----------|-------------------|--------------------------------------------------------|--------|--------------------| 
````

**Comment by praveenkumar** [comment]
@jorgemoralespou So how would we get this 50% weight using oc binary, I tried to check with exported json of route definition but no luck :(

```
$ oc get routes -o=custom-columns=NAME:.spec.to.name,HOST:.spec.host,Weight:.spec.to.weight
NAME           HOST                                              Weight
guestbook-v1   guestbook-myproject.192.168.64.2.nip.io           1
guestbook-v1   guestbook-v1-myproject.192.168.64.2.nip.io        100
guestbook-v1   guestbook-v1-3002-myproject.192.168.64.2.nip.io   100
guestbook-v2   guestbook-v2-myproject.192.168.64.2.nip.io        100
guestbook-v2   guestbook-v2-3002-myproject.192.168.64.2.nip.io   100
```

**Comment by praveenkumar** [comment]
@hferentschik @jorgemoralespou with latest commit on this PR now we have below output (still need to figure out weight stuff)

```
$ ./minishift openshift service list -n test
|-----------|-----------------|--------------------------------------------------------|--------------------|
| NAMSEPACE |      NAME       |                       ROUTE-URL                        |      NODEPORT      |
|-----------|-----------------|--------------------------------------------------------|--------------------|
| test      | guestbook-v1-np |                                                        | 192.168.42.8:31291 |
| test      | guestbook-v2-np |                                                        | 192.168.42.8:32093 |
| test      | guestbook-v1    | http://guestbook-myproject.192.168.64.2.nip.io         |                    |
|           |                 | http://guestbook-v1-myproject.192.168.64.2.nip.io      |                    |
|           |                 | http://guestbook-v1-3002-myproject.192.168.64.2.nip.io |                    |
| test      | guestbook-v2    | http://guestbook-v2-myproject.192.168.64.2.nip.io      |                    |
|           |                 | http://guestbook-v2-3002-myproject.192.168.64.2.nip.io |                    |
|-----------|-----------------|--------------------------------------------------------|--------------------|

$ ./minishift openshift service list 
|-----------|-----------------|--------------------------------------------------------|--------------------|
| NAMSEPACE |      NAME       |                       ROUTE-URL                        |      NODEPORT      |
|-----------|-----------------|--------------------------------------------------------|--------------------|
| default   | docker-registry |                                                        |                    |
| default   | kubernetes      |                                                        |                    |
| default   | router          |                                                        |                    |
| myproject | mariadb         |                                                        |                    |
| myproject | mariadb-ingress |                                                        | 192.168.42.8:31145 |
| myproject | testservice     |                                                        |                    |
| myproject | simpleservice   | http://simpleservice-myproject.192.168.42.8.nip.io     |                    |
| test      | guestbook-v2-np |                                                        | 192.168.42.8:32093 |
| test      | guestbook-v1-np |                                                        | 192.168.42.8:31291 |
| test      | guestbook-v1    | http://guestbook-myproject.192.168.64.2.nip.io         |                    |
|           |                 | http://guestbook-v1-myproject.192.168.64.2.nip.io      |                    |
|           |                 | http://guestbook-v1-3002-myproject.192.168.64.2.nip.io |                    |
| test      | guestbook-v2    | http://guestbook-v2-myproject.192.168.64.2.nip.io      |                    |
|           |                 | http://guestbook-v2-3002-myproject.192.168.64.2.nip.io |                    |
|-----------|-----------------|--------------------------------------------------------|--------------------|
```

**Comment by jorgemoralespou** [comment]
@praveenkumar with this you can get an idea:

````
oc get route -o jsonpath='{range .items[*]}{.metadata.name}-->[{.spec.to.name}:{.spec.to.weight}]--[{.spec.alternateBackends[].name}:{.spec.alternateBackends[].weight}]{"\n"}{end}{"\n"}'
````

**Comment by praveenkumar** [comment]
@jorgemoralespou right but if I have multiple alternate backend service then it will only show only first one.

```
$ oc get routes  
NAME                HOST/PORT                                         PATH      SERVICES                                                PORT       TERMINATION   WILDCARD
guestbook           guestbook-myproject.192.168.64.2.nip.io                     guestbook-v1(50%),guestbook-v2(25%),guestbook-v3(25%)   3000-tcp                 None
guestbook-v1        guestbook-v1-myproject.192.168.64.2.nip.io                  guestbook-v1                                            3000-tcp                 None
guestbook-v1-3002   guestbook-v1-3002-myproject.192.168.64.2.nip.io             guestbook-v1                                            3002-tcp                 None
guestbook-v2        guestbook-v2-myproject.192.168.64.2.nip.io                  guestbook-v2                                            3000-tcp                 None
guestbook-v2-3002   guestbook-v2-3002-myproject.192.168.64.2.nip.io             guestbook-v2                                            3002-tcp                 None

$ oc get route -o jsonpath='{range .items[*]}{.metadata.name}-->[{.spec.to.name}:{.spec.to.weight}]--[{.spec.alternateBackends[].name}:{.spec.alternateBackends[].weight}]{"\n"}{end}{"\n"}'
guestbook-->[guestbook-v1:2]--[guestbook-v2:1]  ====> No entry for guestbook-v3
guestbook-v1-->[guestbook-v1:100]--[:]
guestbook-v1-3002-->[guestbook-v1:100]--[:]
guestbook-v2-->[guestbook-v2:100]--[:]
guestbook-v2-3002-->[guestbook-v2:100]--[:]
```

File: cmd/minishift/cmd/openshift/service_list.go (hunk 11 above)
Comment (LalatenduMohanty) [inline cmd/minishift/cmd/openshift/service_list.go]: @praveenkumar @hferentschik IMO we should keep the current name  space(i.e. the name space user is using currently) space to list services. Refer #927

**Comment by praveenkumar** [comment]
@jorgemoralespou With latest commit you have now below output hope that makes you happy :) 

Yet to add unit test for same.

```
$ ./minishift openshift service list
|-----------|-----------------|--------------------------------------------------------|--------------------|--------|
| NAMSEPACE |      NAME       |                       ROUTE-URL                        |      NODEPORT      | WEIGHT |
|-----------|-----------------|--------------------------------------------------------|--------------------|--------|
| default   | docker-registry |                                                        |                    |        |
| default   | kubernetes      |                                                        |                    |        |
| default   | router          |                                                        |                    |        |
| myproject | mariadb         |                                                        |                    |        |
| myproject | mariadb-ingress |                                                        | 192.168.42.8:31145 |        |
| myproject | testservice     |                                                        |                    |        |
| myproject | simpleservice   | http://simpleservice-myproject.192.168.42.8.nip.io     |                    |        |
| test      | guestbook-v2-np |                                                        | 192.168.42.8:31287 |        |
| test      | guestbook-v1-np |                                                        | 192.168.42.8:32502 |        |
| test      | guestbook-v1    | http://guestbook-myproject.192.168.64.2.nip.io         |                    | 50%    |
|           |                 | http://guestbook-v1-myproject.192.168.64.2.nip.io      |                    |        |
|           |                 | http://guestbook-v1-3002-myproject.192.168.64.2.nip.io |                    |        |
| test      | guestbook-v2    | http://guestbook-myproject.192.168.64.2.nip.io         |                    | 25%    |
|           |                 | http://guestbook-v2-myproject.192.168.64.2.nip.io      |                    |        |
|           |                 | http://guestbook-v2-3002-myproject.192.168.64.2.nip.io |                    |        |
| test      | guestbook-v3    | http://guestbook-myproject.192.168.64.2.nip.io         |                    | 25%    |
|-----------|-----------------|--------------------------------------------------------|--------------------|--------|
```

**Comment by hferentschik** [comment]
Without looking at the actual impl, here are a few thoughts purely based on the table. 

* What's about switching the route and node port column? This would put the route and weight closer together where they are specified. 
* Where there is no weight, does it mean 100%? Should it say so?
* The way the namespace is repeated in the first column is not consistent. It is a mix of always displaying and displaying it when the value changes. What's about always printing the namespace or doing something like this: 

```
$ ./minishift openshift service list
|-----------|-----------------|--------------------------------------------------------|--------------------|--------|
| NAMSEPACE |      NAME       |                       ROUTE-URL                        |      NODEPORT      | WEIGHT |
|-----------|-----------------|--------------------------------------------------------|--------------------|--------|
| default   | docker-registry |                                                        |                    |        |
|               | kubernetes      |                                                        |                    |        |
|               | router          |                                                        |                    |        |
| myproject | mariadb         |                                                        |                    |        |
|                   | mariadb-ingress |                                                        | 192.168.42.8:31145 |        |
|                  | testservice     |                                                        |                    |        |
|               | simpleservice   | http://simpleservice-myproject.192.168.42.8.nip.io     |                    |        |
| test      | guestbook-v2-np |                                                        | 192.168.42.8:31287 |        |
|      | guestbook-v1-np |                                                        | 192.168.42.8:32502 |        |
|      | guestbook-v1    | http://guestbook-myproject.192.168.64.2.nip.io         |                    | 50%    |
|           |                 | http://guestbook-v1-myproject.192.168.64.2.nip.io      |                    |        |
|           |                 | http://guestbook-v1-3002-myproject.192.168.64.2.nip.io |                    |        |
|       | guestbook-v2    | http://guestbook-myproject.192.168.64.2.nip.io         |                    | 25%    |
|           |                 | http://guestbook-v2-myproject.192.168.64.2.nip.io      |                    |        |
|           |                 | http://guestbook-v2-3002-myproject.192.168.64.2.nip.io |                    |        |
|      | guestbook-v3    | http://guestbook-myproject.192.168.64.2.nip.io         |                    | 25%    |
|-----------|-----------------|--------------------------------------------------------|--------------------|--------|
```

**Comment by jorgemoralespou** [comment]
Sooooo happy :-D

But it's not about me ;-)

**Comment by praveenkumar** [comment]
> What's about switching the route and node port column? This would put the route and weight closer together where they are specified.

@hferentschik That's doable.

> Where there is no weight, does it mean 100%? Should it say so?

So by default it's 100 % and even oc route doesn't show default same here.

> The way the namespace is repeated in the first column is not consistent. It is a mix of always displaying and displaying it when the value changes. What's about always printing the namespace or doing something like this

You are asking too much :) current data structure already way to complicated and if we have this then again need to change or this is something done in command side.

**Comment by hferentschik** [comment]
> current data structure already way to complicated and if we have this then again need to change or this is something done in command side.

I would the whole table to be driven on the command side. I think I commented on the code already, but really there should be a list/set of ServiceSpec instance be returned from the implementation of the query side. On the command side where we print the table, it should be a simple matter of iterating (at potentially sorting) over the ServiceSpecs and deciding what to display. Changing how the table renders should be independent on how this data is retrieved. This is model & view pattern really. If you are saying that it is so hard to change this, I fear too much of aggregation is happening too early.

**Comment by praveenkumar** [comment]
> I would the whole table to be driven on the command side. I think I commented on the code already, but really there should be a list/set of ServiceSpec instance be returned from the implementation of the query side.

@hferentschik I agree, as I said that from command side it easy to manipulate and I implemented it.

```
$ ./minishift openshift service list
|-----------|-----------------|--------------------|--------------------------------------------------------|--------|
| NAMSEPACE |      NAME       |      NODEPORT      |                       ROUTE-URL                        | WEIGHT |
|-----------|-----------------|--------------------|--------------------------------------------------------|--------|
| default   | docker-registry |                    |                                                        |        |
|           | kubernetes      |                    |                                                        |        |
|           | router          |                    |                                                        |        |
| myproject | mariadb         |                    |                                                        |        |
|           | mariadb-ingress | 192.168.42.8:31145 |                                                        |        |
|           | testservice     |                    |                                                        |        |
|           | simpleservice   |                    | http://simpleservice-myproject.192.168.42.8.nip.io     |        |
| test      | guestbook-v2-np | 192.168.42.8:31287 |                                                        |        |
|           | guestbook-v1-np | 192.168.42.8:32502 |                                                        |        |
|           | guestbook-v1    |                    | http://guestbook-v1-myproject.192.168.64.2.nip.io      |   50%  |
|           |                 |                    | http://guestbook-v1-3002-myproject.192.168.64.2.nip.io |        |
|           |                 |                    | http://guestbook-myproject.192.168.64.2.nip.io         |        |
|           | guestbook-v2    |                    | http://guestbook-v2-myproject.192.168.64.2.nip.io      |   25%  |
|           |                 |                    | http://guestbook-v2-3002-myproject.192.168.64.2.nip.io |        |
|           |                 |                    | http://guestbook-myproject.192.168.64.2.nip.io         |        |
|           | guestbook-v3    |                    | http://guestbook-myproject.192.168.64.2.nip.io         | 25%    |
|-----------|-----------------|--------------------|--------------------------------------------------------|--------|
```

**Comment by praveenkumar** [comment]
>  @praveenkumar @hferentschik IMO we should keep the current name space(i.e. the name space user is using currently) to list services by default. Refer #927

@hferentschik @LalatenduMohanty Now we are doing each `oc` operation using `~/.minishift/machine/minishift_kubeconfig` which will always provide us `default` as current namespace.

```
$ oc project -q
test

$ oc project -q --config ~/.minishift/machines/minishift_kubeconfig 
default
```

**Comment by budhrg** [comment]
> Now we are doing each oc operation using ~/.minishift/machine/minishift_kubeconfig which will always provide us default as current namespace.

Yes, it is designed that way. Hence, for service, we have explicit `-n`.

**Comment by hferentschik** [comment]
> Now we are doing each oc operation using ~/.minishift/machine/minishift_kubeconfig which will always provide us default as current namespace.

The idea would be to to determine the current namespace first and then use this information to drive the rest. This might sound easier than it is. The current context for _oc_ does not have to be pointing to the Minishift instance. It might be that I am temporarily interacting with my OpenShift online instance. So, if we go down this route, we also need to add some check that the namespace we "extract", maps to the Minishift instance.

Easier is to say that you either have to specify a namespace or that we list all namespaces if none is specified. The latter might not be the worst, since the list command gives you in this case a full view over the configured services. If we enforced the namespace and also used the current namespace as default, there would be no way to get this full view. One would always  limit to a namespace.

This needs to be thought trough in terms of consistency in terms of the other commands.

**Comment by hferentschik** [comment]
I have been playing a bit with the pull request and here are some things I think we need to consider. BTW, the output of `minishift openshift service list` works for me now. This is basically what I had in mind. I also think it is ok to display all namespaces per default and use '-n' to limit them. The current output gives a nice overview of the services in my cluster. 

One thing I noticed is that if a service has neither route nor node port, it is still listed. Do we want this. Personally, I think this is ok. It can be helpful for the user to determine which service he wants/needs to expose. We need to align with `minishift openshift service` though. Taking the router for example, 

    $ minishift openshift service router -n default
    No service with name 'router' defined in namespace 'default'

That's not correct. The router "just" does not have a route. At the very least it needs to say:"Service 'router' in namespace 'default' does not have a route which can be opened in the browser" (something like that).

Also how do we incorporate NodePort into `minishift openshift service` (if at all)? The current behaviour is to open the route of a service in the default browser, provided one is defined.  Is this what we want? Should we change this? Also in the light of that a service can have multiple routes, how do we deal with that?

**Comment by praveenkumar** [comment]
> That's not correct. The router "just" does not have a route. At the very least it needs to say:"Service 'router' in namespace 'default' does not have a route which can be opened in the browser" (something like that).

Right, I didn't touch the `service` part but it should be changed as you suggested.

> Also how do we incorporate NodePort into minishift openshift service (if at all)? 

Since NodePort is not something which should be open to browser so I don't think we should have it part of `minishift openshift service`

> The current behaviour is to open the route of a service in the default browser, provided one is defined. Is this what we want? Should we change this?

What else you want doesn't that why we have minishift openshift service <service-name> to open it to browser if route associated to it.
 
>  Also in the light of that a service can have multiple routes, how do we deal with that?

I am not sure right now but what about we use first one?

**Comment by jorgemoralespou** [comment]
@praveenkumar @hferentschik and what would, in the example above, this command do:

````
minishift openshift service guestbook-v1 -n test
````

That service has 3 routes. I think there is a default route in the web UI, but not sure what logic we're using to decide which is the default route to show/use in case of multiple.

Adding @spadgett and @jwforres who are OpenShift UI engineers to see if they can give an answer to this.

**Comment by jwforres** [comment]
In the console we use a scoring algorithm that takes several factors into account https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/routes.js#L106

File: pkg/minishift/openshift/service.go [hunk 12]
```diff
@@ -67,38 +107,42 @@ func GetServiceURL(service, namespace string, https bool) (string, error) {
...
 
-	url := strings.Split(byteArrayToString(cmdOut), "\n")[1] // second element contain actual URL content
+	url := strings.Split(string(cmdOut), "\n")[1] // second element contain actual URL content
 	return urlScheme + url, nil
 }
 
-// Get the available routes to user
-func GetServiceURLs(serviceListNamespace string) ([]ServiceURL, error) {
-	var serviceURLs []ServiceURL
+// Get the Service spec like route/nodeport/name in a ServiceSpec structure
```
hferentschik: Let's use proper Godoc style docs and start with the function name GetServiceSpecs

pkg/minishift/openshift/service.go
```diff
@@ -136,33 +180,92 @@ func getValidNamespaces(serviceListNamespace string) ([]string, error) {
 	return namespaces, nil
 }
 
-func getServiceURLsOutput(namespace string) (string, error) {
-	cmdArgText := fmt.Sprintf("get route -n %s --config=%s %s", namespace, systemKubeConfigPath, URLsCustomCol)
+func getRouteSpec(namespace string) (map[string][]ServiceWeight, error) {
```
hferentschik: some godocs would be nice

pkg/minishift/openshift/service.go
```diff
@@ -136,33 +180,92 @@ func getValidNamespaces(serviceListNamespace string) ([]string, error) {
...
-	data = re_whtsp_inside.ReplaceAllString(data, " ") // replace all extra whitespaces
-	contents := strings.Split(data, "\n")              // split on new lines
-	contents = emptyFilter(contents[1:])               // remove the header "NAME HOST" and empty elements
+func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace string, serviceNodePort map[string]string) []ServiceSpec {
+	var serviceSpecs []ServiceSpec
+	serviceSpecMap := make(map[string]*ServiceSpec)
+
+	for routeURL, v := range routeSpec {
+		for _, service := range v {
+			// split content on colon to separate NAME, HOST and Weight
```
hferentschik: for my taste to many nested levels. Usually extracting a method or two helps.

pkg/minishift/openshift/service_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"strings"
+	"testing"
+
+	test "github.com/minishift/minishift/pkg/testing"
+
+	instanceState "github.com/minishift/minishift/pkg/minishift/config"
+)
+
+func Test_getProject(t *testing.T) {
+	setup()
```
hferentschik: that's a good pattern 

    setup()
    defer tearDown()

especially if all your tests need the same setup

pkg/minishift/openshift/service_test.go
```diff
@@ -0,0 +1,56 @@
...
+	defer teardown()
+
+	fakeRunner, _, cmdToOutput := test.NewFakeRunner()
```
hferentschik: I think a more explicit way of configuring the fake runner would be more transparent . Right now you returns two maps which it is hard to know what to do with, especially since there are no godocs either. 

Why not:

    fakeRunner := test.NewFakeRunner(t)
    fakeRunner.expectAndReturn("my 1st command", "my 1st expected response")
    fakeRunner.expectAndReturn("my 2nd command", "my 2nd expected response")
     // ...
 
     // make actual test call
     got, err := getProjects()

     // verify what we got is correct

     // verify all expected calls in the right order where received by the FakeRunner
     fakeRunner.verify()

praveenkumar:
@hferentschik Another update to test package. I will take care other suggestion also. From test  package side only thing I didn't get how to implement `verify` method might be we can discuss more today after standup.

pkg/minishift/openshift/service.go
praveenkumar: @hferentschik I did try to have it in a separate method but all those are too dependent that I endup having those in single place,

LalatenduMohanty:
> The idea would be to to determine the current namespace first and then use this information to drive the rest. This might sound easier than it is. The current context for oc does not have to be pointing to the Minishift instance. It might be that I am temporarily interacting with my OpenShift online instance. So, if we go down this route, we also need to add some check that the namespace we "extract", maps to the Minishift instance.

Sounds good to me.  I think it make more sense as we can expect users to use multiple instances of OpenShift along with Minishift.

hferentschik:
> Sounds good to me. I think it make more sense as we can expect users to use multiple instances of OpenShift along with Minishift.

@LalatenduMohanty, so what is it you are after. You think we should determine the current namespace? You don't think the current output of listing all namespaces is the way to go? If not, how does one get this output in your scenario? Not at all? By adding another switch '--all'?

jorgemoralespou:
I would agree here with @hferentschik that this command should not try to guess the current namespace of the user, since it would be way to complex, given how .kube/config works, I might be logged to many different servers. I could look into the minishift context, but once you also introduce multiple clusters, you need to know what context to look for as minishift will probably not be the right context. I would leave that decision to the user, and use this command as an administrative command. Else, the user can always use plain "oc".

hferentschik:
> I would agree here with @hferentschik that this command should not try to guess the current namespace of the user, since it would be way to complex, given how .kube/config works

+1

pkg/minishift/openshift/service_test.go
```diff
@@ -0,0 +1,96 @@
...
+	service := "foo"
+	namespace := "bar"
+	https := false
+
+	fakeRunner := test.NewFakeRunner()
+	runner = fakeRunner.Runner
+
+	args := fmt.Sprintf("get projects --config=%s %s", systemKubeConfigPath, ProjectsCustomCol)
+	expectString := fmt.Sprintf("NAME\n%s", namespace)
+	fakeRunner.ExpectAndReturn(args, expectString)
```
hferentschik: +1

LalatenduMohanty:
> @LalatenduMohanty, so what is it you are after. You think we should determine the current namespace? You don't think the current output of listing all namespaces is the way to go? If not, how does one get this output in your scenario? Not at all? By adding another switch '--all'?

The current implementation for `service list` works for me.   However I was talking about `minishift openshift service` command. 

The current behavior is as below which needs to change as we discussed above.
```
$ minishift openshift service jenkins
No service with name 'jenkins' defined in namespace 'default'
```

jorgemoralespou:
@LalatenduMohanty correct. For that one, I would not use the user in kubeconfig_admin but the one the user is logged in with, as long as it matches the current "context" name.
I guess that minikube is using the context of the user for those commands.

hferentschik:
> The current behavior is as below which needs to change as we discussed above i.e. We need to findout the current name space and not guess it.

I discussed with @praveenkumar the following plan. We keep `openshift service [name]` semantically as is, just fixing the most obvious error in the message. So if the service, exists but does not define a route, it would say "service exists, but is not exposed as route". I think we agree this is not the final solution, but it keeps existing semantics and allows us to take things in steps. 

@praveenkumar wants to create a follow up issue which will list the various alternatives we discussed for the `openshift service [name]` behaviour and then we can decide which one to implement. It would be nice though to get this in into the same sprint.

praveenkumar:
@hferentschik updated test package, right now each unit test fail because map return unstructured object :( but will handle it some way.

pkg/minishift/openshift/service_test.go
```diff
@@ -0,0 +1,133 @@
...
+		{Namespace: namespace, Name: "guestbook-v2-np", URL: []string(nil), NodePort: "30485", Weight: []string(nil)},
+		{Namespace: namespace, Name: "guestbook-v1", URL: []string{"http://guestbook-v1-3002-myproject.192.168.64.2.nip.io", "http://guestbook-myproject.192.168.64.2.nip.io", "http://guestbook-v1-myproject.192.168.64.2.nip.io"}, NodePort: "", Weight: []string{"", "50%", ""}},
+		{Namespace: namespace, Name: "guestbook-v2", URL: []string{"http://guestbook-v2-myproject.192.168.64.2.nip.io", "http://guestbook-v2-3002-myproject.192.168.64.2.nip.io", "http://guestbook-myproject.192.168.64.2.nip.io"}, NodePort: "", Weight: []string{"", "", "50%"}}}
+
+	if err != nil {
+		t.Fatalf("%+v", err)
+	}
+
+	if !reflect.DeepEqual(got, expected) {
+		t.Fatalf("\nExpected: %+v\nGot: %+v", expected, got)
```
praveenkumar: Here is why test case failing because `string{"http://guestbook-v2-myproject.192.168.64.2.nip.io", "http://guestbook-v2-3002-myproject.192.168.64.2.nip.io", "http://guestbook-myproject.192.168.64.2.nip.io"}` this is something part of map value when iterate so I need to find a way either compare bits manually or any smart way.

cmd/minishift/cmd/openshift/service.go
```diff
@@ -39,13 +41,18 @@ var serviceCmd = &cobra.Command{
...
+		util.ExitIfUndefined(api, constants.MachineName)
+
 		if len(args) == 0 || len(args) > 1 {
 			atexit.ExitWithMessage(1, "You must specify the name of the service.")
 		}
 
 		service := args[0]
 
-		url, err := openshift.GetServiceURL(service, namespace, https)
+		url, err := openshift.GetServiceSpec(service, namespace, https)
```
hferentschik: So this pul request does not address the error message of the service command yet, right?

pkg/minishift/openshift/service.go
```diff
@@ -67,38 +107,42 @@ func GetServiceURL(service, namespace string, https bool) (string, error) {
...
 
-	url := strings.Split(byteArrayToString(cmdOut), "\n")[1] // second element contain actual URL content
+	url := strings.Split(string(cmdOut), "\n")[1] // second element contain actual URL content
 	return urlScheme + url, nil
 }
 
-// Get the available routes to user
-func GetServiceURLs(serviceListNamespace string) ([]ServiceURL, error) {
-	var serviceURLs []ServiceURL
+// GetServiceSpecs takes Namespace string and return route/nodeport/name/weight in a ServiceSpec structure
```
hferentschik: technically it takes a list/slice of namespaces and returns a list/slice of ServiceSpec instances

pkg/minishift/openshift/service.go
```diff
@@ -136,33 +180,94 @@ func getValidNamespaces(serviceListNamespace string) ([]string, error) {
 	return namespaces, nil
 }
 
-func getServiceURLsOutput(namespace string) (string, error) {
-	cmdArgText := fmt.Sprintf("get route -n %s --config=%s %s", namespace, systemKubeConfigPath, URLsCustomCol)
+// getRouteSpec take valid namespace and return map which have route as key and
```
hferentschik: 👍

pkg/minishift/openshift/service_test.go
```diff
@@ -0,0 +1,173 @@
...
+		expected = "https://" + strings.Split(expectString, "\n")[1]
+	}
+
+	if got != expected {
+		t.Fatalf("Expected: %+v, Got: %+v", expected, got)
+	}
+
+}
+
+func Test_getServiceSpecs(t *testing.T) {
```
hferentschik: I am big advocate to make the test "tell" you what is tested, e.g. Test_get_ServiceSpecs_with_multiple_nodeports_and_routes. Something like this. Or the special condition you try to verify.

pkg/minishift/openshift/service_test.go
```diff
@@ -0,0 +1,173 @@
...
+
+	fakeRunner := test.NewFakeRunner(t)
+	runner = fakeRunner.Runner
+
+	args := fmt.Sprintf("get projects --config=%s %s", systemKubeConfigPath, ProjectsCustomCol)
+	expectString := fmt.Sprintf("NAME\n%s", namespace)
+	fakeRunner.ExpectAndReturn(args, expectString)
+
+	args = fmt.Sprintf("get svc -o json --config=%s -n %s", systemKubeConfigPath, namespace)
+	expectString = testdata.ServiceSpec
```
hferentschik: nice, using some real test data. This way the whole JSON processing gets tested +1

pkg/testing/runner_mock.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (f *FakeRunner) ExpectAndReturn(input string, output string) {
+	expectCommands = append(expectCommands, ExpectedCommand{commandArgs: input, commandToOutput: output})
+}
+
+/*func (f *FakeRunner) Verify() {
+
+}*/
+
+func (f *Runner) Output(command string, args ...string) ([]byte, error) {
```
hferentschik: +1, going to work. There many ways to do this, this is one of them

praveenkumar:
@hferentschik  Good to go now.

pkg/minishift/openshift/service.go
```diff
@@ -64,41 +104,45 @@ func GetServiceURL(service, namespace string, https bool) (string, error) {
 	cmdName := instanceState.InstanceConfig.OcPath
 	cmdOut, err := runner.Output(cmdName, tokens...)
 	if err != nil {
-		return "", errors.New(fmt.Sprintf("No service with name '%s' defined in namespace '%s'", service, namespace))
+		return "", errors.New(fmt.Sprintf("Service with name '%s' in namespace '%s' doesn't have route associated", service, namespace))
```
hferentschik: I don't think this is going to work. Now it displays the right thing for say the router service, but I also get 'Service with name 'foo' in namespace 'default' doesn't have route associated' for some random non existent service. Really we need to differentiate somewhere.

hferentschik: There is a full stop missing in the message. 
Maybe something like this sounds better: 

"Service  '%s' in namespace '%s' does not have route associated which can be opened in the browser."

hferentschik: Would it not be easiest to let _ GetServiceSpec_ call _ GetServiceSpecs_ and then based on the returned  ServiceSpec decide on either the URL to return or the appropriate error message?

pkg/minishift/openshift/service.go
```diff
@@ -36,69 +41,111 @@ func init() {
...
+		return "", err
 	}
 
-	cmdArgText := fmt.Sprintf("get route/%s -n %s --config=%s %s", service, namespace, systemKubeConfigPath, URLCustomCol)
-	tokens := strings.Split(cmdArgText, " ")
-	cmdName := instanceState.InstanceConfig.OcPath
-	cmdOut, err := runner.Output(cmdName, tokens...)
-	if err != nil {
-		return "", errors.New(fmt.Sprintf("No service with name '%s' defined in namespace '%s'", service, namespace))
+	for _, serviceSpec := range serviceSpecs {
```
hferentschik: :-) 👍

hferentschik:
Merged, thanks.

hferentschik:
Let's see whether we can fix #1106 as part of this release as well.

Issue #1084, opened 2017-06-29
Create integration test for proxy


agajdosi:
Upstream version for [CDK-90](https://issues.jboss.org/browse/CDK-90).

**About**
Integration tests needs to cover "behind the proxy" situation, which gave us some problems in past. Current plan is to use [go-proxy](https://github.com/elazarl/goproxy) to setup proxy directly on developer's computer to keep integration tests close to developers.

**Current state**
In what I have now: feature is able to start proxy, then start minishift with --http-proxy - and start runs successfully. However IP address of network device is still needed. Will try to automate this process. Another big task is to take care of all the logging and checking that the traffic really goes through the proxy.

hferentschik:
merged via pull request #1095

PR #1095, opened 2017-07-03
Issue #1084 Create integration test for proxy

agajdosi:
Adds:
- detection of host IP (need verification)
- creation of proxy server
- env var set/unset steps
- destruction of proxy listener
- logging + checking proxy log

Work in progress - I mostly interested how detection of host IP works. For minishift one can't use localhost, but needs IP of host, tested on Fedora when connected via WiFi, not sure how this will work on ethernet, or ethernet + wifi combinations etc.

Thankful for help and feedback.

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -260,6 +260,7 @@ function setup_build_environment() {
   install_core_deps;
   install_kvm_virt;
   install_docker;
+  prepare_for_proxy;
   runuser -l minishift_ci -c "/bin/bash centos_ci.sh"
 }
 
@@ -308,6 +309,11 @@ function build_and_test() {
   echo "CICO: Tests ran successfully"
 
   artifacts_upload_on_pr_and_master_trigger $1;
+}
+
+function prepare_for_proxy() {
+  export INTEGRATION_PROXY_CUSTOM_PORT=8181 #needs to be an unused port
+  firewall-cmd --zone=public --add-port=$INTEGRATION_PROXY_CUSTOM_PORT/tcp;
 }
 
 if [[ "$UID" = 0 ]]; then
--- a/docs/source/contributing/developing.adoc
+++ b/docs/source/contributing/developing.adoc
@@ -192,12 +192,26 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
 For more information about test options, run the `go test --help` command and review the documentation.
 
 [[integration-tests]]
-=== Running Integration Tests
+=== Integration Tests
 
-Integration tests utilize link:https://github.com/DATA-DOG/godog[Godog], which uses Gherkin (Cucumber) to define test cases.
-The test cases are defined in *_test/integration/features_*.
-By default, the tests are executed against the binary created by `make build`, that is *_$GOPATH/bin/minishift_*.
-To run the tests, run:
+Integration tests utilize link:https://github.com/DATA-DOG/godog[Godog], which uses Gherkin (Cucumber) to define sets of test cases, in Gherkin terminology known as _features_.
+The features are located in *_test/integration/features_* folder.
+Features for Minishift follow these basic concepts:
+
+User stories::
+Features which follow a happy path of user.
+For example, _basic.feature_ or _coolstore.feature_.
+
+Feature and command coverage::
+Features which focuses on specific fields of Minishift functionality or individual commands.
+For example, _proxy.feature_ or _cmd-version.feature_.
+
+
+[[running-integration-tests]]
+==== Running Integration Tests
+
+By default, the tests are being run against the binary created by `make build`, which is *_$GOPATH/bin/minishift_*. 
+To run the tests, use the following command:
 
 ----
 $ make integration
@@ -209,17 +223,31 @@ To run integration tests against a Minishift binary in a different location you 
 $ make integration MINISHIFT_BINARY=<path-to-custom-binary>
 ----
 
-Additional properties for Godog runner can be specified with `GODOG_OPTS` argument.
+[[godog-options]]
+==== Using Tags and Other Godog Options
+
+Additional properties for Godog runner can be specified with the `GODOG_OPTS` argument. 
 The following options are available:
 
-- `tags`: Use `tags` to ensure that scenarios and features containing at least one of the selected tags are executed.
-- `paths`: Use `paths` to define paths to different feature files or folders containing feature files.
+Tags::
+Use `tags` to ensure that scenarios and features containing at least one of the selected tags are executed.
+
+Paths::
+Use `paths` to define paths to different feature files or folders containing feature files.
 This can be used to run feature files outside of the *_test/integration/features_* folder.
-- `format`:  Use `format` to change the format of Godog's output.
+
+Format::
+Use `format` to change the format of Godog's output.
 For example, you can use `pretty` format instead of the native `progress`.
-- `stop-on-failure`: Set `stop-on-failure` to true to stop integration tests on failure.
-- `no-colors`: Set `no-colors` to true to disable ansi colors of Godog's output.
-- `definitions`: Set `definitions` to true to print all available step definitions.
+
+Stop-on-failure::
+Set `stop-on-failure` to true to stop integration tests on failure.
+
+No-colors::
+Set `no-colors` to true to disable ansi colors of Godog's output.
+
+Definitions::
+Set `definitions` to true to print all available step definitions.
 
 For example, to run integration tests on two specific feature files using only `@basic` and `@openshift` tags and without ansi colors, the following command can be used:
 
--- a/test/integration/features/proxy.feature
+++ b/test/integration/features/proxy.feature
@@ -1,0 +1,61 @@
+@proxy @minishift-only
+Feature: Minishift can run behind proxy
+  As a user I can use Minishift behind a proxy.
+  INFO: Code behind this feature will start a proxy server in background
+        on port 8181 and will try to get IP of host machine automatically
+        to set MINISHIFT_HTTP_PROXY for this feature only.
+        However one can override port number or set IP manually if needed.
+        To use custom port number run this test with environmental variable
+        INTEGRATION_PROXY_CUSTOM_PORT.       
+        To use custom IP address please set INTEGRATION_PROXY_CUSTOM_IP
+        environmental variable.
+
+  Scenario: Start behind the proxy
+   Given user starts proxy server and sets MINISHIFT_HTTP_PROXY variable
+    When executing "minishift start" succeeds
+    Then proxy log should contain "Accepting CONNECT to registry-1.docker.io:443"
+     And proxy log should contain "Accepting CONNECT to auth.docker.io:443"
+
+  Scenario: Proxy environmental variable is set in created VM
+    When executing "minishift ssh -- cat /var/lib/boot2docker/profile" succeeds
+    Then stdout should match "HTTP_PROXY=http\:\/\/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\:\d{1,5}"
+     And stdout should match "NO_PROXY=localhost,127.0.0.1,172.30.1.1"
+
+  Scenario: User can login to the server
+   Given Minishift has state "Running"
+    When executing "oc login --username=developer --password=developer" succeeds
+    Then stdout should contain
+     """
+     Login successful
+     """
+
+  Scenario: User can create new namespace ruby for application ruby-ex
+   Given Minishift has state "Running"
+    When executing "oc new-project ruby" succeeds
+    Then stdout should contain
+     """
+     Now using project "ruby"
+     """
+
+  Scenario: User can deploy application ruby-ex to namespace ruby
+    Given Minishift has state "Running"
+     When executing "oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-ex.git" succeeds
+     Then stdout should contain
+      """
+      Success
+      """
+    When executing "oc rollout status deploymentconfig ruby-ex --watch" succeeds
+    Then stdout should contain
+     """
+     "ruby-ex-1" successfully rolled out
+     """
+     And proxy log should contain "Accepting CONNECT to registry-1.docker.io:443"
+     And proxy log should contain "Accepting CONNECT to bundler.rubygems.org:443"
+     And proxy log should contain "Accepting CONNECT to rubygems.org:443"
+
+  Scenario: Delete behind the proxy
+    When executing "minishift delete" succeeds
+    Then Minishift has state "Does Not Exist"
+
+  Scenario: User stops proxy
+   Given user stops proxy server and unsets MINISHIFT_HTTP_PROXY variable
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -19,13 +19,18 @@ limitations under the License.
 package integration
 
 import (
+	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io/ioutil"
 	"net"
+	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"testing"
@@ -33,13 +38,9 @@ import (
 	"github.com/DATA-DOG/godog"
 	"github.com/DATA-DOG/godog/gherkin"
 
-	"crypto/tls"
-	"errors"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	testProxy "github.com/minishift/minishift/test/integration/proxy"
 	"github.com/minishift/minishift/test/integration/util"
-	"net/http"
-	"path/filepath"
-	"regexp"
 )
 
 var (
@@ -94,7 +95,7 @@ func parseFlags() {
 
 	flag.StringVar(&testDir, "test-dir", "", "Path to the directory in which to execute the tests")
 
-	flag.StringVar(&godogFormat, "format", "progress", "Sets which format godog will use")
+	flag.StringVar(&godogFormat, "format", "pretty", "Sets which format godog will use")
 	flag.StringVar(&godogTags, "tags", "", "Tags for godog test")
 	flag.BoolVar(&godogShowStepDefinitions, "definitions", false, "")
 	flag.BoolVar(&godogStopOnFailure, "stop-on-failure ", false, "Stop when failure is found")
@@ -141,33 +142,43 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^services "([^"]*)" rollout successfully$`,
 		minishift.rolloutServicesSuccessfully)
 
+	// steps for proxying
+	s.Step(`^user starts proxy server and sets MINISHIFT_HTTP_PROXY variable$`,
+		testProxy.SetProxy)
+	s.Step(`^user stops proxy server and unsets MINISHIFT_HTTP_PROXY variable$`,
+		testProxy.UnsetProxy)
+	s.Step(`^proxy log should contain "([^"]*)"$`,
+		proxyLogShouldContain)
+	s.Step(`^proxy log should contain$`,
+		proxyLogShouldContainContent)
+
 	// steps to verify `stdout`, `stderr` and `exitcode` of commands executed
-	s.Step(`([^"]*) should contain "([^"]*)"$`,
+	s.Step(`^(stdout|stderr|exitcode) should contain "([^"]*)"$`,
 		commandReturnShouldContain)
-	s.Step(`([^"]*) should not contain "([^"]*)"$`,
+	s.Step(`^(stdout|stderr|exitcode) should not contain "([^"]*)"$`,
 		commandReturnShouldNotContain)
-	s.Step(`([^"]*) should contain$`,
+	s.Step(`^(stdout|stderr|exitcode) should contain$`,
 		commandReturnShouldContainContent)
-	s.Step(`([^"]*) should not contain$`,
+	s.Step(`^(stdout|stderr|exitcode) should not contain$`,
 		commandReturnShouldNotContainContent)
-	s.Step(`([^"]*) should equal "([^"]*)"$`,
+	s.Step(`^(stdout|stderr|exitcode) should equal "([^"]*)"$`,
 		commandReturnShouldEqual)
-	s.Step(`([^"]*) should equal$`,
+	s.Step(`^(stdout|stderr|exitcode) should equal$`,
 		commandReturnShouldEqualContent)
-	s.Step(`([^"]*) should be empty$`,
+	s.Step(`^(stdout|stderr|exitcode) should be empty$`,
 		commandReturnShouldBeEmpty)
-	s.Step(`([^"]*) should not be empty$`,
+	s.Step(`^(stdout|stderr|exitcode) should not be empty$`,
 		commandReturnShouldNotBeEmpty)
-	s.Step(`([^"]*) should be valid ([^"]*)$`,
+	s.Step(`^(stdout|stderr|exitcode) should be valid ([^"]*)$`,
 		shouldBeInValidFormat)
 	// steps for matching stdout, stderr or exitcode with regular expression
-	s.Step(`([^"]*) should match "([^"]*)"$`,
+	s.Step(`^(stdout|stderr|exitcode) should match "([^"]*)"$`,
 		commandReturnShouldMatchRegex)
-	s.Step(`([^"]*) should not match "([^"]*)"$`,
+	s.Step(`^(stdout|stderr|exitcode) should not match "([^"]*)"$`,
 		commandReturnShouldNotMatchRegex)
-	s.Step(`([^"]*) should match$`,
+	s.Step(`^(stdout|stderr|exitcode) should match$`,
 		commandReturnShouldMatchRegexContent)
-	s.Step(`([^"]*) should not match$`,
+	s.Step(`^(stdout|stderr|exitcode) should not match$`,
 		commandReturnShouldNotMatchRegexContent)
 
 	// step for HTTP requests for minishift web console
@@ -196,6 +207,11 @@ func FeatureContext(s *godog.Suite) {
 	s.AfterSuite(func() {
 		minishift.runner.EnsureDeleted()
 	})
+
+	s.AfterScenario(func(interface{}, error) {
+		testProxy.ResetLog(false)
+	})
+
 }
 
 func setUp() string {
@@ -361,13 +377,11 @@ func selectFieldFromLastOutput(commandField string) string {
 		outputField = lastCommandOutput.StdErr
 	case "exitcode":
 		outputField = strconv.Itoa(lastCommandOutput.ExitCode)
-	default:
-		fmt.Errorf("Incorrect field type specified for comparison: %s", commandField)
 	}
 	return outputField
 }
 
-func shouldBeInValidFormat(commandField, format string) error {
+func shouldBeInValidFormat(commandField string, format string) error {
 	result := selectFieldFromLastOutput(commandField)
 	result = strings.TrimRight(result, "\n")
 	switch format {
@@ -380,18 +394,16 @@ func shouldBeInValidFormat(commandField, format string) error {
 		if net.ParseIP(result) == nil {
 			return fmt.Errorf("%s of previous command is not a valid IP address: %s", commandField, result)
 		}
-	default:
-		return fmt.Errorf("Format %s not implemented.", format)
 	}
 	return nil
 }
 
-func commandReturnEquals(commandField, command, expected string) error {
+func commandReturnEquals(commandField string, command string, expected string) error {
 	minishift.executingMinishiftCommand(command)
 	return compareExpectedWithActualEquals(expected+"\n", selectFieldFromLastOutput(commandField))
 }
 
-func commandReturnContains(commandField, command, expected string) error {
+func commandReturnContains(commandField string, command string, expected string) error {
 	minishift.executingMinishiftCommand(command)
 	return compareExpectedWithActualContains(expected, selectFieldFromLastOutput(commandField))
 }
@@ -483,7 +495,7 @@ func succeedsOrFails(execute commandRunner, command string, expectedResult strin
 	return nil
 }
 
-func verifyHTTPResponse(partOfResponse, url, urlSuffix, assertion, expected string) error {
+func verifyHTTPResponse(partOfResponse string, url string, urlSuffix string, assertion string, expected string) error {
 	switch url {
 	case "OpenShift":
 		url = minishift.getOpenShiftUrl() + urlSuffix
@@ -523,7 +535,7 @@ func verifyHTTPResponse(partOfResponse, url, urlSuffix, assertion, expected stri
 	return nil
 }
 
-func getRoutingUrlAndVerifyHTTPResponse(partOfResponse, urlRoot, serviceName, nameSpace, assertion, expected string) error {
+func getRoutingUrlAndVerifyHTTPResponse(partOfResponse string, urlRoot string, serviceName string, nameSpace string, assertion string, expected string) error {
 	url := minishift.getRoute(serviceName, nameSpace)
 	if urlRoot == "/" {
 		return verifyHTTPResponse(partOfResponse, url, "", assertion, expected)
@@ -534,4 +546,12 @@ func getRoutingUrlAndVerifyHTTPResponse(partOfResponse, urlRoot, serviceName, na
 		return fmt.Errorf("Wrong input format : %s. Input must start with /", urlRoot)
 	}
 	return nil
+}
+
+func proxyLogShouldContain(expected string) error {
+	return compareExpectedWithActualContains(expected, testProxy.GetLog())
+}
+
+func proxyLogShouldContainContent(expected *gherkin.DocString) error {
+	return compareExpectedWithActualContains(expected.Content, testProxy.GetLog())
 }
--- a/test/integration/proxy/proxy.go
+++ b/test/integration/proxy/proxy.go
@@ -1,0 +1,126 @@
+// +build integration
+
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package proxy
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"log"
+	"net"
+	"net/http"
+	"os"
+	"strings"
+
+	"github.com/elazarl/goproxy"
+	minishiftUtil "github.com/minishift/minishift/pkg/minishift/util"
+)
+
+var (
+	proxyListener net.Listener
+	proxyLog      *bytes.Buffer
+)
+
+const DEFAULT_PROXY_PORT string = "8181"
+
+func GetLog() string {
+	var log string
+	if proxyLog != nil {
+		log = proxyLog.String()
+	}
+
+	return log
+}
+
+func ResetLog(verbose bool) {
+	if GetLog() != "" {
+		if verbose == true {
+			fmt.Printf("beginning of proxy log >>>\n%s<<< end of proxy log\n", GetLog())
+		}
+		proxyLog.Reset()
+	}
+
+	return
+}
+
+func startProxy(proxyPort string) error {
+	var err error
+	proxyListener, err = net.Listen("tcp", ":"+proxyPort)
+	if err != nil {
+		return err
+	}
+
+	proxy := goproxy.NewProxyHttpServer()
+	proxy.Verbose = true
+	proxyLog = bytes.NewBufferString("")
+	proxy.Logger = log.New(proxyLog, "", log.Ldate)
+
+	http.Serve(proxyListener, proxy)
+	fmt.Println("Proxy has stopped.")
+
+	return nil
+}
+
+func getPort() string {
+	if value, present := os.LookupEnv("INTEGRATION_PROXY_CUSTOM_PORT"); present == true {
+		return value
+	}
+
+	return DEFAULT_PROXY_PORT
+}
+
+func getIP() (string, error) {
+	if value, present := os.LookupEnv("INTEGRATION_PROXY_CUSTOM_IP"); present == true {
+		return value, nil
+	}
+
+	ips := minishiftUtil.HostIPs()
+	if ips == nil {
+		return "", errors.New(`No IP found. This might be an error in automated detection of available network devices.
+							   You can use INTEGRATION_PROXY_CUSTOM_IP to set IP manually.`)
+	}
+
+	ip := strings.Split(ips[0], "/")[0]
+	return ip, nil
+}
+
+func SetProxy() error {
+	port := getPort()
+	ip, err := getIP()
+	if err != nil {
+		return err
+	}
+
+	go startProxy(port)
+
+	address := fmt.Sprintf("http://%v:%v", ip, port)
+	os.Setenv("MINISHIFT_HTTP_PROXY", address)
+	fmt.Printf("  INFO: this feature is now using MINISHIFT_HTTP_PROXY=%s\n", os.Getenv("MINISHIFT_HTTP_PROXY"))
+
+	return nil
+}
+
+func UnsetProxy() error {
+	os.Unsetenv("MINISHIFT_HTTP_PROXY")
+	if proxyListener != nil {
+		proxyListener.Close()
+	}
+
+	return nil
+}
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 9331dcdebd3c68898e6841b0bc140c756c478f2325064216b89d035060cdc066
-updated: 2017-06-23T15:31:41.602757701+05:30
+hash: 8a468283a9a4675c087c7d25a00ce4a72c4fc0f5f05abe60eec93bfce22be14b
+updated: 2017-07-10T11:56:37.047777606+02:00
 imports:
 - name: github.com/asaskevich/govalidator
   version: 7b3beb6df3c42abd3509abfc3bcacc0fbfb7c877
@@ -60,6 +60,8 @@ imports:
   - libmachine/version
   - libmachine/versioncmp
   - version
+- name: github.com/elazarl/goproxy
+  version: aacba83f36a55ac31cbb71c06547a328c0cd1604
 - name: github.com/fatih/color
   version: 62e9147c64a1ed519147b62a56a14e83e2be02c1
 - name: github.com/fsnotify/fsnotify
--- a/glide.yaml
+++ b/glide.yaml
@@ -60,6 +60,7 @@ import:
   version: 5
 - package: github.com/DATA-DOG/godog
   version: v0.6.2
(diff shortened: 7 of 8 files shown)


test/integration/integration_test.go
```diff
@@ -181,6 +196,72 @@ func ensureTestDirEmpty() {
...
+	}
+	proxy := goproxy.NewProxyHttpServer()
+	proxy.Verbose = true
+	proxyLog = bytes.NewBufferString("proxy log:\n")
+	proxy.Logger = log.New(proxyLog, "", log.Ldate)
+	http.Serve(proxyListener, proxy)
+	return nil
+}
+
+func getHostIP() net.IP {
```
gbraad: a similar function already exists in the minishift utils library. Please see if this is suitable and works for you.

test/integration/integration_test.go
```diff
@@ -181,6 +196,72 @@ func ensureTestDirEmpty() {
...
+			case *net.IPAddr:
+				ip = v.IP
+			}
+			return ip
+		}
+	}
+	return nil
+}
+
+func setEnvVariable(varName, varValue string) error {
```
gbraad: varName is of type string, so please describe it as such.

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,15 @@
+@proxy
+Feature: Minishift can run behind proxy
+  As a user I can use Minishift behind a proxy.
+
+  Scenario: Start behind the proxy
+   Given user starts proxy server and sets MINISHIFT_HTTP_PROXY variable
+    When executing "minishift start" succeeds
+    Then proxy log should contain "Accepting CONNECT to registry-1.docker.io:443"
```
gbraad: Maybe this is not the right place. see below

test/integration/integration_test.go
```diff
@@ -234,6 +315,8 @@ func selectFieldFromLastOutput(commandField string) string {
 		outputField = lastCommandOutput.StdErr
 	case "exitcode":
 		outputField = strconv.Itoa(lastCommandOutput.ExitCode)
+	case "proxy log":
```
gbraad: This is not the right place, as this function should only deal with the last command output, which is a typed struct which contains command specific output.

It is probably better to define a specific function to deal with this instead. This might mean that you need to define a step definition as:

`"^proxy log should contain", content` instead... which can then call it's own function.

gbraad:
add to whitelist

test/integration/integration_test.go
gbraad: see `/pkg/minishift/util/hostip.go` which returns all the hosts defined IP addresses as a `[]string`

hferentschik:
Add to whitelist

test/integration/integration_test.go
agajdosi: wanted to be brief, my bad, going to change it

test/integration/integration_test.go
agajdosi: @gbraad I started this way, but then I have realized that `^proxy log should contain` will be caught by `([^"]*) should contain$` if it is not first. Will be dependent on order or I will need to change `([^"]*)` for `(stdout|stderr|exitcode)`.

I agree with you that naming is bad in this case. But I am still wondering if it is better to create new functions with duplicate code here or if it would be possible to rename the existing functions and make them more re-usable. I would vote for the second, because:

```golang
func commandReturnShouldContain(commandField string, expected string) error {
	return compareExpectedWithActualContains(expected, selectFieldFromLastOutput(commandField))
}
```
there is really nothing "last command output" specific. And here:
```golang
func selectFieldFromLastOutput(commandField string) string {
	outputField := ""
	switch commandField {
	case "stdout":
		outputField = lastCommandOutput.StdOut
	case "stderr":
		outputField = lastCommandOutput.StdErr
	case "exitcode":
		outputField = strconv.Itoa(lastCommandOutput.ExitCode)
	case "proxy log":
		outputField = proxyLog.String()
	default:
		fmt.Errorf("Incorrect field type specified for comparison: %s", commandField)
	}
	return outputField
}
```

the code specific for command output is only in switch and these cases are on just one line. WDYT? Would like to ask @hferentschik also.

test/integration/integration_test.go
agajdosi: Thanks @gbraad, will take a look on it.

test/integration/integration_test.go
gbraad: I do not believe we duplicate code, as when you write the step:

```
"^proxy log should contain$"
"^([^"]*) should contain$"
```

You can have the function just do:
```
func proxyLogShouldContain(expected string) error {
  return compareExpectedWithActualContains(expected, proxyLog.String())
}
```

agajdosi: @gbraad Neat. Thanks.

Going to make all the changes and update the PR.

test/integration/integration_test.go
agajdosi: @gbraad Having some weak moment now, so I would like to ask you if I am on right track or completely lost:

In order to try func HostIPs() from file /pkg/minishift/util/hostip.go
I import the package util as minishiftUtil:
```golang
import (
  ...
  "github.com/minishift/minishift/test/integration/util"
  minishiftUtil "github.com/minishift/minishift/pkg/util"
  ...
)
```
Then hostIPs() should be accessible with: minishiftUtil.hostIPs()
```golang
ips := minishiftUtil.HostIPs()
```
But I am getting
```
test/integration/integration_test.go:205: undefined: "github.com/minishift/minishift/pkg/util".HostIPs
```
with which I am not sure what to do.

agajdosi: Hmmm, I have to not forget /minishift/ in path and import correct package..
```
minishiftUtil "github.com/minishift/minishift/pkg/minishift/util"
```

gbraad:
After pushing the changes to #986, I will spend a little bit more time on this.
If you made the change of `(exitcode|stdout|...)` I do not need to refactor this in my PR.

The `HostIP` suggestion I gave (existing code) should be enough to progress I guess...

hferentschik:
What's the sate here. We want to first merge the Coolstore suff and then continue working on this, right?

agajdosi:
@hferentschik @gbraad Hi, I have done '(exitcode|stdout|...)' change in this PR. HostIP function works for me also. There is need for some minor changes and then I can push on Monday morning. However after all code changes, there will be need for addition of more scenarios as now it is just bare minimum.

agajdosi:
It is still WIP - afterScenario contains proxy log printing, which is used for debugging and will be removed later. Also this PR needs more scenarios to be added, working on it today/tomorrow.

agajdosi:
Tested on Windows 7 in VM, tests passed, proxy works as expected just it asks for network privileges. I need to investigate why it is having troubles in centos-ci - as log is empty and does not even have "proxy log:" header it seems, that it could got stuck somewhere in proxy goroutine even before the log is populated with header. Second guess is firewall blocking port 3128.

test/integration/features/proxy.feature
agajdosi: Resolved now, using newly created func configContains(), thanks.

agajdosi:
So the problems with this PR and central-ci, on central-ci proxy log contains after startup just this:
```
2017/07/14 [001] INFO: Running 0 CONNECT handlers
2017/07/14 [001] INFO: Accepting CONNECT to github.com:443
2017/07/14 [002] INFO: Running 0 CONNECT handlers
2017/07/14 [002] INFO: Accepting CONNECT to github-production-release-asset-2e65be.s3.amazonaws.com:443
2017/07/14 [003] INFO: Running 0 CONNECT handlers
2017/07/14 [003] INFO: Accepting CONNECT to api.github.com:443
2017/07/14 [004] INFO: Running 0 CONNECT handlers
2017/07/14 [004] INFO: Accepting CONNECT to github-production-release-asset-2e65be.s3.amazonaws.com:443
```
these two calls are I believe downloads for oc and b2d. Until here it is fine. But why it shows nothing for pulls of OpenShift images? Does the proxy die here? How is then possible that start is successful?

My local log contains:

```
2017/07/14 [001] INFO: Running 0 CONNECT handlers
2017/07/14 [001] INFO: Accepting CONNECT to github.com:443
2017/07/14 [002] INFO: Running 0 CONNECT handlers
2017/07/14 [002] INFO: Accepting CONNECT to github-production-release-asset-2e65be.s3.amazonaws.com:443
2017/07/14 [003] INFO: Running 0 CONNECT handlers
2017/07/14 [003] INFO: Accepting CONNECT to api.github.com:443
2017/07/14 [004] INFO: Running 0 CONNECT handlers
2017/07/14 [004] INFO: Accepting CONNECT to github-production-release-asset-2e65be.s3.amazonaws.com:443
2017/07/14 [005] INFO: Running 0 CONNECT handlers
2017/07/14 [005] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [006] INFO: Running 0 CONNECT handlers
2017/07/14 [006] INFO: Accepting CONNECT to auth.docker.io:443
2017/07/14 [007] INFO: Running 0 CONNECT handlers
2017/07/14 [007] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [008] INFO: Running 0 CONNECT handlers
2017/07/14 [008] INFO: Accepting CONNECT to dseasb33srnrn.cloudfront.net:443
2017/07/14 [009] INFO: Running 0 CONNECT handlers
2017/07/14 [009] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [010] INFO: Running 0 CONNECT handlers
2017/07/14 [010] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [011] INFO: Running 0 CONNECT handlers
2017/07/14 [011] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [012] INFO: Running 0 CONNECT handlers
2017/07/14 [012] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [013] INFO: Running 0 CONNECT handlers
2017/07/14 [013] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [014] INFO: Running 0 CONNECT handlers
2017/07/14 [014] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [015] INFO: Running 0 CONNECT handlers
2017/07/14 [015] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [016] INFO: Running 0 CONNECT handlers
2017/07/14 [016] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [017] INFO: Running 0 CONNECT handlers
2017/07/14 [017] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [018] INFO: Running 0 CONNECT handlers
2017/07/14 [018] INFO: Accepting CONNECT to registry-1.docker.io:443
2017/07/14 [019] INFO: Running 0 CONNECT handlers
2017/07/14 [019] INFO: Accepting CONNECT to registry-1.docker.io:443
```
^here interesting fact is that if I do not delete the out folder containing the minishift home folder for integration tests both calls on github are not present - seems like oc and b2d are preserved through test runs.

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,87 @@
+@proxy @minishift-only
+Feature: Minishift can run behind proxy
+  As a user I can use Minishift behind a proxy.
+
```
hferentschik: So how do I run this test. I figure I have to set some environment variables, but I am not sure which ones exactly. I got asked for a password in the middle of the test . I thought it was the CDK registration and set MINISHIFT_USER and MINISHIFT_PASSWORD. But now I am wondering whether this was the password for the proxy.

Maybe the feature header should mention which env variables this test expects to be set.

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,87 @@
...
+  Scenario: Start behind the proxy
+   Given user starts proxy server and sets MINISHIFT_HTTP_PROXY variable
```
hferentschik: Or maybe you don't need to set anything at all? I see that setProxt sets the proxy variable.

test/integration/integration_test.go
```diff
@@ -48,6 +53,10 @@ var (
 
 	testDir string
 
+	//proxy
+	proxyListener net.Listener
+	proxyLog      *bytes.Buffer
```
hferentschik: :+1 writing the logs into a buffer

hferentschik: I would extract all proxy related functions into a separate file proxy.go or something

test/integration/integration_test.go
```diff
@@ -436,7 +520,7 @@ func verifyHTTPResponse(partOfResponse, url, urlSuffix, assertion, expected stri
 	return nil
 }
 
-func getRoutingUrlAndVerifyHTTPResponse(partOfResponse, urlRoot, serviceName, nameSpace, assertion, expected string) error {
+func getRoutingUrlAndVerifyHTTPResponse(partOfResponse string, urlRoot string, serviceName string, nameSpace string, assertion string, expected string) error {
```
hferentschik: :-) Both syntaxes are valid. Personally I also prefer this one.

test/integration/integration_test.go
gbraad: agree, but maybe after we know what goes wrong in CI. the benefit of having it in `integration_test.go` is the possible use of `println` ;-)

test/integration/integration_test.go
gbraad: We actually have some others that also miss the type... I forgot to fix them.

hferentschik:
I like where this is going. Trying to run this on my machine atm. I'll let you know whether it works for me.

One thing we should be aware of is that what this shows is that there are proxy requests made. There is no guarantee that everything is properly working yet. For that we would need to block requests being made to the default non proxy ports aka 80 and 443. 

For Minishift itself this can be easily done by changing some HTTP config settings as part of the test setup. This would work for testing that the ISO and oc downloads fully go through the proxy. 

Things are getting more complicated when the CDK comes into play and you want for example ensure that all the RHEL registration goes via the proxy. The requests are made from within the VM, so modifying the Go network classes won't help. In this case one needs to find other ways, to either block requests to the non proxy ports or at least have a possibility to verify that no requests are going through there.

We had actually exactly this problem before. We had some (manual) testing regarding the CDK registration code. When inspecting the logs we could see that some requests towards the registration backend were made, but as it turned out later, this was just the initial request. There was more communication going on as part of registration and that happily bypassed the proxy w/o being noticed. 

+1 For where this is heading, just be aware that one can not deduct with 100% certainty from a test like this that proxying works.

gbraad:
we could configure libvirt to create a host-only network, so we have to rely on the proxy to provide download. let's do this as an improvement after we have more visibility on why this fails, or even after a merge.

hferentschik:
For what it's worth, here is my result:

    --- Failed scenarios:
    
        features/proxy.feature:12
        features/proxy.feature:22
    
    13 scenarios (11 passed, 2 failed)
    37 steps (33 passed, 2 failed, 2 skipped)

I also got failures, but apparently less than on Central CI. Not sure whether they are "expected".

hferentschik:
> we could configure libvirt to create a host-only network, so we have to rely on the proxy to provide download.

Interesting. Definitely an option. The other thing we could do is to enable a set of iptables rules before the test run and then reset them on test tear down.

> let's do this as an improvement after we have more visibility on why this fails, or even after a merge.

definitely. I would be more than happy to get this merged once we figured out what the problem is right now. We can then refine/extend later. I just wanted to raise the issue.

hferentschik:
once this works, I think it is time to setup the nightly build we have been talking about which would run all integration tests.

gbraad:
Fails from the beginning at the moment
```
  Scenario: Start behind the proxy                                                # features/proxy.feature:5
    Given user starts proxy server and sets MINISHIFT_HTTP_PROXY variable         # integration_test.go:261 -> github.com/minishift/minishift/test/integration.setProxy
    When executing "minishift start" succeeds                                     # integration_test.go:454 -> github.com/minishift/minishift/test/integration.executingMinishiftCommandSucceedsOrFails
    Then proxy log should contain "Accepting CONNECT to registry-1.docker.io:443" # integration_test.go:536 -> github.com/minishift/minishift/test/integration.proxyLogShouldContain
    Output did not match. Expected: Accepting CONNECT to registry-1.docker.io:443, Actual: <nil>
    And proxy log should contain "Accepting CONNECT to auth.docker.io:443"
```

did something change?

test/integration/features/proxy.feature
agajdosi: Wanted to describe the behaviour and processes which are being done there - the step starts proxy and sets the env variable. Felt like it will just complicate things if these were in 2 separate steps.

test/integration/integration_test.go
agajdosi: @hferentschik 

> I would extract all proxy related functions into a separate file proxy.go or something

I am notorious one-filer :( but I agree, will move it later into separate file after CI problem is resolved as @gbraad suggests.

agajdosi:
> Not sure whether they are "expected".

@hferentschik 12 and 22 are steps for curl from the box, I am not sure why they fails: `When executing "minishift ssh curl www.midi.cz" succeeds`. When I execute same command manually behind the proxy it works fine, also I am having troubles with -L flag, so I am not curl-ing redhat.com or github.com but just some random web server.

> we could configure libvirt to create a host-only network, so we have to rely on the proxy to provide download. let's do this as an improvement after we have more visibility on why this fails, or even after a merge.

@gbraad Really interesting, will look on that. Agree that it could be added later after merging basics, but it could be handy to use it also for manual testing, so totally going to get more information there.

> Fails from the beginning at the moment

Works for me (if not counting steps 12 and 22) - maybe could be bad NIC or port chosen, I am still wondering here which one to pick. For me on Fedora 24 all NICs found by HostIP works, on Windows I just used first IP in returned list and it went fine. On ports I think I had problems with 8080 on Fedora. For CI I tried all three NICs' IPs which it returns with 3128, 8080, 8181 and some random ports number without any success.

What I really do not understand how is it possible that it downloads iso and oc through proxy and then on OpenShift images it just stops using proxy - but as start is successful, it downloads them, just not using proxy.

test/integration/features/proxy.feature
agajdosi: For now it is not using any environmental variables, but I am starting to think that it could probably use PROXY_PORT_NUMBER and maybe PROXY_IP, now detection of IP is fine on Fedora and Windows - but @gbraad is maybe having troubles with that, so adding IP and port as parameters is an option.

To your problem, I can imagine that if you have MINISHIFT_USER set, maybe minishift will ask you for password even if the box does not need the registration - tested interactive registration with nightly build of CDK so I am not sure with behaviour on minishift. Can try if I can replicate this idea. For this test I am using minishift 1.2.0.

agajdosi: No, that was not the case. So I am completely unsure how is it possible that it asked you for a password.

+1 on keeping test setup information in feature header, seems practical. Will add it.

gbraad:
The behaviour in CI does not make sense, unless it crashes. But locally I see why it might fail, as port 8181 is occupied.

Might be good to introduce a variable to easily specify the port for deployment.

gbraad:
I also tried to add some logic to determine which port can be used, BUT... the minishift proxy needs to be set before the node is up... and `minishift ip` only works after the machine is created, so no way to get the intersection of the network interface IPs. :-s. I guess locally it is related to an occupied port... just now it is to figure out how to get debug info from the CentOS CI to track this down...

hferentschik:
Might be worth reading reaching out to the CentOS CI guys. I think our contact is Brian. @budhrg knows best.

gbraad:
> I think our contact is Brian

Already emailed him about this...

-- 

   [ Doing Open Source Matters ]

hferentschik:
+1 :-)

hferentschik:
Is there an update on this? It would be nice to get this ready and merged. Are we still unclear on why this fails in CentOS CI?

gbraad:
So far, no... :-s
I have started to test this locally again, but CentOS CI remains a problem.

Luckily @budhrg is back...

hferentschik:
> Luckily @budhrg is back...

👍

agajdosi:
So I need to open the port in firewall, which I forgot to do as it looked open - the proxy worked from start and then stopped. However how I realized, this could be info just about fact, that there was connect request incoming not that it actually served the connection.

hferentschik:
> So I need to open the port in firewall, which I forgot to do as it looked open - the proxy worked from start and then stopped. However how I realized, this could be info just about fact, that there was connect request incoming not that it actually served the connection.

ahhh progress :-)

agajdosi:
@hferentschik @gbraad So this was because of a firewall, test now passes on centos-ci. Going to implement suggested code changes.

hferentschik:
> So this was because of a firewall, test now passes on centos-ci. Going to implement suggested code changes.

+1 Looking forward to integrate it :-)

gbraad:
great, looking forward to have this test, together with the flags test to
be integrated soon

agajdosi:
I have moved proxy functionality into proxy package as suggested by @hferentschik. Also I have added support for setting up custom port for proxy and custom IP of network interface controller. So now tests natively will select NIC automatically and will use port 8181. If one needs to use another port (8181 he uses for something else and is occupied) then one can use INTEGRATION_PROXY_CUSTOM_PORT environmental variable. If there is a problem with automatic detection of NIC's IP, then one can run for example `ifconfig` and choose another IP which can be passed to tests with INTEGRATION_PROXY_CUSTOM_IP. This test should take care and do it for you, but is tweakable if needed.

agajdosi:
Will add documentation in next iteration.

test/integration/integration_test.go
```diff
@@ -141,33 +142,43 @@ func FeatureContext(s *godog.Suite) {
...
+	s.Step(`^user stops proxy server and unsets MINISHIFT_HTTP_PROXY variable$`,
+		testProxy.UnsetProxy)
+	s.Step(`^proxy log should contain "([^"]*)"$`,
+		proxyLogShouldContain)
+	s.Step(`^proxy log should contain$`,
+		proxyLogShouldContainContent)
+
 	// steps to verify `stdout`, `stderr` and `exitcode` of commands executed
-	s.Step(`([^"]*) should contain "([^"]*)"$`,
+	s.Step(`^(stdout|stderr|exitcode) should contain "([^"]*)"$`,
```
hferentschik: :+1 I like that this is getting more explicit. This becomes more and more like Aruba. I like.

centos_ci.sh
```diff
@@ -304,12 +312,17 @@ function build_and_test() {
   cd $GOPATH/src/github.com/minishift/minishift
   make prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags ~coolstore -format pretty"
+  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags proxy -format pretty"
```
budhrg: @agajdosi Are we running coolstore test suite as well?

agajdosi: @budhrg This is just to test if this passes centos-ci, on merge it should be probably changed to `MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags ~coolstore,~proxy -format pretty"`. @hferentschik

hferentschik: > This is just to test if this passes centos-ci, on merge it should be probably changed to MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags ~coolstore,~proxy -format pretty"

The problem is really the time it takes to run all this and the fact that we run it as part of pull request builds which should complete in a reasonable time. I guess _proxy_ in itself runs in a reasonable time. _proxy_ + _basic_ might become a border case. I guess we can see how long it takes. _coolstore_ is a no-go for now, since provisioning takes 30min+.

That said, why do we not add the nightly build at this stage. The default could then just run _basic_ and in the nightly we build all features we have.

agajdosi:
retest this please

centos_ci.sh
budhrg: Ack @agajdosi

centos_ci.sh
```diff
@@ -310,6 +311,11 @@ function build_and_test() {
   artifacts_upload_on_pr_and_master_trigger $1;
 }
 
+function prepare_for_proxy() {
+  export INTEGRATION_PROXY_CUSTOM_PORT=8181 #needs to be an unused port
```
budhrg: :+1:

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
 [[integration-tests]]
-=== Running Integration Tests
+=== Integration Tests
 
-Integration tests utilize link:https://github.com/DATA-DOG/godog[Godog], which uses Gherkin (Cucumber) to define test cases.
-The test cases are defined in *_test/integration/features_*.
-By default, the tests are executed against the binary created by `make build`, that is *_$GOPATH/bin/minishift_*.
-To run the tests, run:
+Integration tests utilize link:https://github.com/DATA-DOG/godog[Godog], which uses Gherkin (Cucumber) to define sets of test cases,
+aka features in Gherkin terminology. The features are defined in *_test/integration/features_*. Features for Minishift are following
```
budhrg: I think the path for features should be `_test/integration/features/*.feature`. And next line should be

The feature file for Minishift follows two basic concepts:

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
+2 basic concepts:
+
+- *user stories* - follows a happy path of user, for example _basic.feature_ or _coolstore.feature_,
+- *topic and command coverage* - for _example proxy.feature_ or _cmd-version.feature_.
```
budhrg: example is inside `_` underscore scope. Should be   `for example _proxy.feature_ ...`

test/integration/integration_test.go
```diff
@@ -361,13 +377,11 @@ func selectFieldFromLastOutput(commandField string) string {
...
 	case "exitcode":
 		outputField = strconv.Itoa(lastCommandOutput.ExitCode)
-	default:
-		fmt.Errorf("Incorrect field type specified for comparison: %s", commandField)
 	}
 	return outputField
 }
 
-func shouldBeInValidFormat(commandField, format string) error {
+func shouldBeInValidFormat(commandField string, format string) error {
```
budhrg: This was not needed. It is usual pattern in Golang.

test/integration/proxy/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+var (
+	proxyListener net.Listener
+	proxyLog      *bytes.Buffer
+)
+
+func GetLog() string {
+	var log string
+	if proxyLog != nil {
+		log = proxyLog.String()
+	} else {
```
budhrg: Else part not needed. Zero/default value of string is empty. See https://tour.golang.org/basics/12.

test/integration/proxy/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+
+	return nil
+}
+
+func getPort() string {
+	if value, present := os.LookupEnv("INTEGRATION_PROXY_CUSTOM_PORT"); present == true {
+		return value
+	}
+
+	return "8181"
```
budhrg: Would be good to make it constant above variable declaration in the file and return here.
Something like `PROXY_PORT`

test/integration/integration_test.go
budhrg: I prefer the former. Want to write more Go idiomatic code.

budhrg:
LGTM :+1: with suggestions/comment resolved.

docs/source/contributing/developing.adoc
hferentschik: > I think the path for features should be _test/integration/features/*.feature

Maybe, maybe not. Does not really matter so much imo. We are aiming though for a sentence a line. So, there is a "wrong" line break after the command and a missing one after _ terminology_.

docs/source/contributing/developing.adoc
hferentschik: what's 'topic'?

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,81 @@
+@proxy @minishift-only
```
hferentschik: what makes this test Minishift specific?

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,81 @@
...
+    Then stdout should contain
+     """
+     Success
+     """
+    When executing "oc rollout status deploymentconfig ruby-ex --watch" succeeds
+    Then stdout should contain
+     """
+     "ruby-ex-1" successfully rolled out
+     """
+     And proxy log should contain "Accepting CONNECT to registry-1.docker.io:443"
```
hferentschik: nice nice

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,81 @@
...
+   Given Minishift has state "Running"
+    When executing "oc expose svc/ruby-ex" succeeds
+    Then stdout should contain
+     """
+     exposed
+     """
+    And status code of HTTP request to "/" of service "ruby-ex" in namespace "ruby" is equal to "200"
+    And body of HTTP request to "/" of service "ruby-ex" in namespace "ruby" contains "Welcome to your Ruby application on OpenShift"
+
+  Scenario: User can delete namespace ruby
```
hferentschik: Is this really needed? How is this related to proxy? I believe this is covered somewhere else, right? The tests should be to the point and as much as possible only test what's relevant for a given feature. One might as well delete here directly.

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,81 @@
...
+  Scenario: User can log out the session
+   Given Minishift has state "Running"
+    When executing "oc logout" succeeds
+    Then stdout should contain
+     """
+     Logged "developer" out
+     """
+
+  Scenario: Delete behind the proxy
+    When executing "minishift delete" succeeds
```
hferentschik: "     Then Minishift should have state "Does Not Exist"

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,81 @@
...
+    When executing "oc rollout status deploymentconfig ruby-ex --watch" succeeds
+    Then stdout should contain
+     """
+     "ruby-ex-1" successfully rolled out
+     """
+     And proxy log should contain "Accepting CONNECT to registry-1.docker.io:443"
+     And proxy log should contain "Accepting CONNECT to bundler.rubygems.org:443"
+     And proxy log should contain "Accepting CONNECT to rubygems.org:443"
+
+  Scenario: User can create route for ruby-ex to make it visiable outside of the cluster
```
hferentschik: why the route scenario? How is it relevant?

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,81 @@
+@proxy @minishift-only
+Feature: Minishift can run behind proxy
+  As a user I can use Minishift behind a proxy.
+  INFO: Code behind this feature will start a proxy server in background
```
hferentschik: :+1 Nice with docs!

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,81 @@
...
+     """
+
+  Scenario: User can create new namespace ruby for application ruby-ex
+   Given Minishift has state "Running"
+    When executing "oc new-project ruby" succeeds
+    Then stdout should contain
+     """
+     Now using project "ruby"
+     """
+
```
hferentschik: What I am missing is a another simple check. When proxies are specified we actually also set the environment variables HTTP(S)_PROXY an NO_PROXY in the VM environment. So I believe that something like:

    $ minishift ssh -- echo $HTTP_PROXY

should return the configured proxy.

test/integration/integration_test.go
hferentschik: Not needed, but we usually use the more explicit form.

test/integration/proxy/proxy.go
hferentschik: DEFAULT_PROXY_PORT

docs/source/contributing/developing.adoc
agajdosi: good eyes, thanks

test/integration/proxy/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+		log = ""
+	}
+
+	return log
+}
+
+func ResetLog(verbose bool) {
+	if GetLog() != "" {
+		if verbose == true {
+			fmt.Printf("beggining of proxy log >>>\n%s<<< end of proxy log\n", GetLog())
```
hferentschik: typo - "beginning"

test/integration/integration_test.go
agajdosi: @budhrg There was a suggestions above in the PR about using this format and not the shorted form. So I followed it and made it the same in whole file. Personally I like `name1, name2, name3 type` a little bit more :)

test/integration/proxy/proxy.go
agajdosi: @budhrg +1, thanks for tip

test/integration/features/proxy.feature
agajdosi: @hferentschik In current form which does not block all connections but more it checks the log, the minishift specific part lies in checks for download of OpenShift images, for CDK it needs to check for RedHat's registry.

test/integration/features/proxy.feature
agajdosi: @hferentschik Going to look on it, thanks for suggestion.

agajdosi:
@hferentschik For some reason I am unable to reply on your comment.

> what's 'topic'?

By topic I meant for example a usage of Minishift behind the proxy. However there could be better word, will try to figure something out.

docs/source/contributing/developing.adoc
hferentschik: Maybe "feature" is better in this context?

test/integration/proxy/proxy.go
agajdosi: @budhrg That is neat! Did not realized this shortcut, thanks!

docs/source/contributing/developing.adoc
agajdosi: @hferentschik +1

test/integration/features/proxy.feature
agajdosi: @hferentschik So it seems that for non-interactive shell the environmental variables are not loaded from /var/lib/boot2docker/profile:
```
$ minishift ssh -- echo $HTTP_PROXY

$
```

Thinking if this is an issue or just normal behaviour? Can be here an user case which would require non-interactive shell and will need all parts of profile executed? I know that this is quite common around .bash_rc or .bash_profile, not sure here - can be resolved somehow, but for some files it is just known and expected to behave like this.  

For now I think it could be just fine to do:
```cucumber
  Scenario: Proxy environmental variable is set in created VM
    When executing "minishift ssh -- cat /var/lib/boot2docker/profile" succeeds
    Then stdout should match "HTTP_PROXY=http\:\/\/\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3}\:\d{1,5}"
     And stdout should match "NO_PROXY=localhost,127.0.0.1,172.30.1.1"
```

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
 For more information about test options, run the `go test --help` command and review the documentation.
 
 [[integration-tests]]
-=== Running Integration Tests
+=== Integration Tests
```
agajdosi: I did small changes here, checked by Hardy and Budhram already, but could I please ask you for an additional quick check, @thatdocslady?

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
 
-Integration tests utilize link:https://github.com/DATA-DOG/godog[Godog], which uses Gherkin (Cucumber) to define test cases.
-The test cases are defined in *_test/integration/features_*.
-By default, the tests are executed against the binary created by `make build`, that is *_$GOPATH/bin/minishift_*.
-To run the tests, run:
+Integration tests utilize link:https://github.com/DATA-DOG/godog[Godog], which uses Gherkin (Cucumber) to define sets of test cases, aka _features_ in Gherkin terminology.
```
thatdocslady: we don't use abbreviations such as "aka", please rephrase to something like: "... test cases, which are parallel to _features_ in Gherkin terminology", or even just use the word "or", if it's simpler: "... test cases, or _features_..."

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
+The features are located in *_test/integration/features_* folder.
+Features for Minishift are following 2 basic concepts:
```
thatdocslady: "are following 2" > "follow these"
(reasons: "are following" is ambiguous, and 2 is hardcoding a number, we try to avoid that in case this changes in the future)

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
+
+- *user stories* - follows a happy path of user, for example _basic.feature_ or _coolstore.feature_,
```
thatdocslady: Please use sentence case for "User stories and "Features". Also, I recommend using deflist tagging:

User stories::
<para>

Features and.. ::
<para>

thatdocslady: What is a "happy path"? is this a known industry term? I think we can use a more accurate and less emotional term here.

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
+- *feature and command coverage* - for example _proxy.feature_ or _cmd-version.feature_.
```
thatdocslady: Please capitalize sentences in the beginning and use a comman after "For example,".

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
+
+===== Running integration tests
```
thatdocslady: * over-indentation - if you use === in the previous heading, then this heading should have ====
* Missing anchod ID above the heading, each heading should have an anchor ID that is unique and that matches the title text, so in this case it would be [[running-integration-tests]]

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
+
+By default, the tests are executed against the binary created by `make build`, that is *_$GOPATH/bin/minishift_*.
```
thatdocslady: "that is" is confusing and ambiguous here, what do you mean?

docs/source/contributing/developing.adoc
```diff
@@ -192,12 +192,19 @@ $go test -v <relative path of package> -run "Test<Regex pattern to match tests>"
...
+The features are located in *_test/integration/features_* folder.
+Features for Minishift are following 2 basic concepts:
+
+- *user stories* - follows a happy path of user, for example _basic.feature_ or _coolstore.feature_,
+- *feature and command coverage* - for example _proxy.feature_ or _cmd-version.feature_.
+
+===== Running integration tests
+
+By default, the tests are executed against the binary created by `make build`, that is *_$GOPATH/bin/minishift_*. 
+To run the tests, execute:
```
thatdocslady: We try not to use "execute" for commands, please change to "use the following command" or "run the following command" (although since we're running tests, "use" is probably better)

docs/source/contributing/developing.adoc
```diff
@@ -209,7 +216,9 @@ To run integration tests against a Minishift binary in a different location you
 $ make integration MINISHIFT_BINARY=<path-to-custom-binary>
 ----
 
-Additional properties for Godog runner can be specified with `GODOG_OPTS` argument.
+===== Using tags and other Godog options
```
thatdocslady: Same comment about indentation level and anchor ID

docs/source/contributing/developing.adoc
```diff
@@ -209,7 +216,9 @@ To run integration tests against a Minishift binary in a different location you
...
+
+Additional properties for Godog runner can be specified with `GODOG_OPTS` argument.
```
thatdocslady: "specified with the..."

docs/source/contributing/developing.adoc
hferentschik: "which is" might be a bit better, but I think the overall meaning is clear.

docs/source/contributing/developing.adoc
hferentschik: > What is a "happy path"? is this a known industry term? 

Very much so - https://en.wikipedia.org/wiki/Happy_path

docs/source/contributing/developing.adoc
agajdosi: @thatdocslady Thanks for all the feedback, I really appreciate it. 

Could this case be changed to `to define sets of test cases known as _features_ in Gherkin terminology.`?

thatdocslady: @agajdosi yup that works too :)

agajdosi:
@hferentschik PR could be fine now. On merge there will be probably just need to change L:308 in centos_ci.sh to exclude proxy.feature from running on PRs. I won't be here tomorrow.

hferentschik:
>  PR could be fine now. On merge there will be probably just need to change L:308 in centos_ci.sh to exclude proxy.feature from running on PRs.

Sure, I'll take a look. Depending on the time it takes we might as well run this test on CI.

test/integration/features/proxy.feature
hferentschik: ahh, got you

gbraad:
As configured in the current way, it would only run the proxy test case.
Better to go back to basic, and make the proxy test run when the task for
'make integration' is worked on.

-- 

   [ Doing Open Source Matters ]

gbraad:
>> @@ -0,0 +1,81 @@
> +@proxy @minishift-only

Or this...

-- 

   [ Doing Open Source Matters ]

test/integration/features/proxy.feature
```diff
@@ -0,0 +1,61 @@
...
+        environmental variable.
+
+  Scenario: Start behind the proxy
+   Given user starts proxy server and sets MINISHIFT_HTTP_PROXY variable
+    When executing "minishift start" succeeds
+    Then proxy log should contain "Accepting CONNECT to registry-1.docker.io:443"
+     And proxy log should contain "Accepting CONNECT to auth.docker.io:443"
+
+  Scenario: Proxy environmental variable is set in created VM
+    When executing "minishift ssh -- cat /var/lib/boot2docker/profile" succeeds
```
hferentschik: I believe this is now ISO specific. I will change this a bit to make it ISO independent.

hferentschik:
Merged with some minor amendment for the test checking that proxy vars are applied on the VM

hferentschik:
Actually right now we are running all Cucumber tests, except coolstore. Let's see how this goes and whether the time for a build result is too long. If so we can adjust as part of #1216

gbraad:
Timeout is currently set to 45 minutes IIRC (@budhrg)...
Actually this is long, but then again... we kinda have unittests for the
quick feedback loop.
Wish we had an intermediate smoke test option... or more of a pipeline with
different stages.

hferentschik:
We are below that for now. We are looking at ~30min. Given that the other CI system respond much faster, this might be ok. It is a trade-off really. Any additional test on integration level run, gives us more security. 

What hurts imo right now is that `make integration` on localhost runs all. There we really need to have an easy option to just run _basic_, otherwise devs won't run integration tests. I would like this be addresses as part of #1216. Also I think as part of #1216 we can add the nightly build we have been discussing for such a long time.

Issue #950, opened 2017-05-23
minishift start --help should document environment variables applicable in Minishift


rctrh:
I think it will come up frequently enough that `minishift start --help` should document the environment variables that can be used to store the registration username and password.

hferentschik:
I think we need to add something to the generic usage of  the help output, not only the username and password. All the parameters are configurable by env variables https://docs.openshift.org/latest/minishift/using/managing-minishift.html#environment-variables. This generic approach should be mentioned.

gbraad:
I believe `minikube` doesn't specify this either, but only refers to it from their documentation.  I could be wrong, but my 2c? I doubt `-h`/`--help` output is the place for this... it will clutter the help output. I do not think people will look for this functionality there.

Minikube docs reference:
Originally https://github.com/kubernetes/minikube/pull/1094/files, but moved to https://github.com/kubernetes/minikube/blob/master/docs/env_vars.md

praveenkumar:
>  I doubt -h/--help output is the place for this... it will clutter the help output. I do not think people will look for this functionality there

I do agree. Either we create a separate command option for this which provide info about what can be set as environment variable. Also as per `minikube` document it says only config which are part of `config view` can be set as env but for us this list is very big now.

hferentschik:
Merged via pull request #1096

PR #1096, opened 2017-07-04
Issue #950  Adding information of env and persistent config settings 

hferentschik:
- Making 'b2d' default for ISO URL to shorten help display

Addresses issue #950.


hferentschik:
>  However we need to change https://docs.openshift.org/latest/minishift/using/managing-minishift.html#environment-variables to tell users how they can use values for flags in minishift start from environment variables i.e. similar to the this PR. At least we should fix the docs to reflect the same.

You are loosing me here. The docs contain this information. I basically took it from there. A online docs are more verbose, but effectively the information is the same.

LalatenduMohanty:
> You are loosing me here. The docs contain this information. I basically took it from there. A online docs are more verbose, but effectively the information is the same.

I feel the text in this PR is relatively easy to understand as compared to the docs. Hence the suggestion.

hferentschik:
> I feel the text in this PR is relatively easy to understand as compared to the docs.

I agree. IMO it contains all the information one needs. It does not have to be more complicated than this. The current text on the online docs is a result of making the docs "easier" to understand. Whether this is the case or not depends most likely on the point of view.

LalatenduMohanty:
> Whether this is the case or not depends most likely on the point of view.

@thatdocslady @gbraad @praveenkumar @budhrg WDYT? Should we change the docs similar to the PR text?

hferentschik:
@LalatenduMohanty Just to be clear, what you want is to replace this (in the online docs):

    Minishift allows you to specify command-line flags you commonly use through environment variables. To do so, apply the following rules to the flag you want to set as an environment variable.

    * Apply MINISHIFT_ as a prefix to the flag you want to set as an environment variable. For example, the vm-driver flag of the minishift start command becomes MINISHIFT_vm-driver.

    * Use uppercase characters for the flag, so MINISHIFT_vm-driver in the above example becomes MINISHIFT_VM-DRIVER.

    * Replace - with _, so MINISHIFT_VM-DRIVER becomes MINISHIFT_VM_DRIVER.

With basically a single sentence:

"Prefix the flag with MINISHIFT_, uppercase characters and replace '-' with '_', for example MINISHIFT_VM_DRIVER"

praveenkumar:
To me both looks understandable. I do agree on doc side it' bit complicated but it also provide better context about how actually convert happen.

LalatenduMohanty:
@hferentschik  Actually 
```
* Apply MINISHIFT_ as a prefix to the flag you want to set as an environment variable. For example, the vm-driver flag of the minishift start command becomes MINISHIFT_vm-driver.

* Use uppercase characters for the flag, so MINISHIFT_vm-driver in the above example becomes MINISHIFT_VM-DRIVER.

* Replace - with _, so MINISHIFT_VM-DRIVER becomes MINISHIFT_VM_DRIVER.
```
With

"Prefix the flag with MINISHIFT_, uppercase characters and replace '-' with '_', for example MINISHIFT_VM_DRIVER"

and then proof reading the paragraph again and fix if required.

hferentschik:
> and then proof reading the paragraph again and fix if required.

Sure.

jorgemoralespou:
l liked that the current format has an example along the explanation, so as long as you add the example afterwards, WFM.

PR #1113, opened 2017-07-10
[WIP] Rollout check with goroutines

gbraad:
DO NOT MERGE

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -135,7 +135,7 @@ function prepare() {
 function run_tests() {
   make clean test cross fmtcheck prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration
+  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags basic -format pretty"
   echo "CICO: Tests ran successfully"
 }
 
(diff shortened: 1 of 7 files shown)


gbraad:
Basic scenario works: https://ci.centos.org/job/minishift-pr/1463/console

gbraad:
retest this please

gbraad:
Requesting two services fails, while in a small scale test this works as expected. Probably a stupid typo...

gbraad:
when dealing with channels... do not do:

```go
if strings.Contains(output, expected) {
   done <- true
}
done <- false
```

but ALWAYS do:

```go
if strings.Contains(output, expected) {
   done <- true
} else {
   done <- false
}
```

gbraad:
I broke it down into a smaller application: https://gist.github.com/gbraad/c2cba6cde46eb7968c5e8be974371dde

and as mentioned in the previous message. If you do not use an `else` condition block, the channel will always receive `false` for the successive calls.

gbraad:
retest this please

hferentschik:
Will this supersede #986?

gbraad:
@hferentschik nope. it is a bad changeset anyways, as it got branched of a wrong point and changes some of the logic. If I am happy with how it works I'll incorporate it in the other PR.

hferentschik:
Got you

gbraad:
OK, so it can run to completion now.
Time to optimize the goroutine and prevent it from blocking on the channel receive.

gbraad:
Cart failed... as usual I should say... ?

gbraad:
Integrated in the CoolStore PR #986

PR #1114, opened 2017-07-10
Issue #1106 Incorporate NodePort into minishift openshift service command

### LINKED ISSUE #1106: Incorporate NodePort into minishift openshift service
This is a follow up issue for #876.

The current behaviour is to open the route of a service in the default browser, provided one is defined. Is this what we want? 

Should we change this  so that when a user asks for a specific service then we display all associated routes and nodeports. To open in browser  `--inbrowser` needs to be explicitly specified.

Something like

```
$ minishift openshift service guestbook-v2 -n myproject
|-----------------|----------------------|--------------------------------------------------------|--------|
|      NAME       |       NODEPORT       |                       ROUTE-URL                        | WEIGHT |
|-----------------|----------------------|--------------------------------------------------------|--------|
| guestbook-v2    |                      | http://guestbook-v2-3002-myproject.192.168.64.2.nip.io |  50%   |
|                 |                      | http://guestbook-myproject.192.168.64.2.nip.io         |        |
|                 |                      | http://guestbook-v2-myproject.192.168.64.2.nip.io      |        |
|-----------------|----------------------|--------------------------------------------------------|--------|

$ minishift openshift service guestbook-v2 -n myproject --inbrowser
< Open a routes in browser>
```

Also in the light of that a service can have multiple routes, how do we deal with that?

@jorgemoralespou  asked openshift folks about how to display the route and they provided below explanation but I think for our minishift purpose we don't have to put all in to account and just open first from index, let me know if that is wrong.

*In the console we use a scoring algorithm that takes several factors into account https://github.com/openshift/origin-web-console/blob/master/app/scripts/services/routes.js#L106*

praveenkumar:
Fix #1106


praveenkumar:
retest this please

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,78 @@ var serviceCmd = &cobra.Command{
...
+
+		ip, err := host.Driver.GetIP()
+		if err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error getting IP: %s", err.Error()))
+		}
+
 		service := args[0]
 
-		url, err := openshift.GetServiceSpec(service, namespace, https)
+		serviceSpec, err := openshift.GetServiceSpec(service, namespace, https)
```
hferentschik: I would have thought we can get rid off _openshift.GetServiceSpec_ and only use _openshift.GetServiceSpecs_, no?

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,78 @@ var serviceCmd = &cobra.Command{
...
 		if err != nil {
 			atexit.ExitWithMessage(1, err.Error())
 		}
 
-		if urlMode {
-			fmt.Fprintln(os.Stdout, url)
+		if url {
```
hferentschik: Would we not get rid of the url flag in this case?

hferentschik: Or is this just printing the URL to the console, instead of the table format? I guess that's an option.

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,78 @@ var serviceCmd = &cobra.Command{
...
+			stdOutURL(serviceSpec)
+		} else if inbrowser {
+			openInBrowser(serviceSpec)
 		} else {
-			fmt.Fprintln(os.Stdout, "Opening the service "+namespace+"/"+service+" in the default browser...")
-			browser.OpenURL(url)
+			var data [][]string
```
hferentschik: Maybe do this in a function

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,78 @@ var serviceCmd = &cobra.Command{
...
 	serviceCmd.Flags().StringVarP(&namespace, "namespace", "n", "default", "The namespace of the service.")
-	serviceCmd.Flags().BoolVar(&urlMode, "url", false, "Access the service in the command-line console instead of in the default browser.")
+	serviceCmd.Flags().BoolVar(&inbrowser, "inbrowser", false, "Access the service in the default browser.")
+	serviceCmd.Flags().BoolVar(&url, "url", false, "Access the service url in the command-line console.")
 	serviceCmd.Flags().BoolVar(&https, "https", false, "Access the service with HTTPS instead of HTTP.")
 	OpenShiftCmd.AddCommand(serviceCmd)
 }
+
+func openInBrowser(serviceSpec openshift.ServiceSpec) {
+	if serviceSpec.URL != nil {
```
hferentschik: I am wondering whether we should allow opening a NodePort in browser. See https://github.com/minishift/minishift/issues/1106#issuecomment-314100093

pkg/minishift/openshift/service.go
```diff
@@ -88,30 +88,30 @@ const (
 )
 
 // Get the route for service
-func GetServiceSpec(service, namespace string, https bool) (string, error) {
+func GetServiceSpec(service, namespace string, https bool) (ServiceSpec, error) {
+	var serviceSpec ServiceSpec
 	urlScheme := "http://"
```
hferentschik: I think this ever only made sense for a NodePort, no?

cmd/minishift/cmd/openshift/service.go
praveenkumar: @hferentschik we need a way for our integration test where we need url so for automation purpose we need it.

cmd/minishift/cmd/openshift/service.go
praveenkumar: @hferentschik I do like to have `GetServiceSpec` even it just return serviceSpec for specific service looks much cleaner then. (Internally it does use GetServiceSpecs only)

cmd/minishift/cmd/openshift/service.go
hferentschik: > we need a way for our integration test where we need url so for automation 

purpose we need it.

good point

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,90 @@ var serviceCmd = &cobra.Command{
...
-			browser.OpenURL(url)
+			printToStdOut(serviceSpec, ip)
 		}
 	},
 }
 
 func init() {
 	serviceCmd.Flags().StringVarP(&namespace, "namespace", "n", "default", "The namespace of the service.")
-	serviceCmd.Flags().BoolVar(&urlMode, "url", false, "Access the service in the command-line console instead of in the default browser.")
+	serviceCmd.Flags().BoolVar(&inbrowser, "inbrowser", false, "Access the service in the default browser.")
```
hferentschik: What do you think about 'in-browser' as flag name?

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,90 @@ var serviceCmd = &cobra.Command{
...
+	serviceCmd.Flags().BoolVar(&url, "url", false, "Access the service url in the command-line console.")
```
hferentschik: "Print the service URL to standard output."

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,90 @@ var serviceCmd = &cobra.Command{
...
 	serviceCmd.Flags().BoolVar(&https, "https", false, "Access the service with HTTPS instead of HTTP.")
 	OpenShiftCmd.AddCommand(serviceCmd)
 }
+
+func openInBrowser(serviceSpec openshift.ServiceSpec, https bool, ip string) {
+	if serviceSpec.URL != nil {
+		fmt.Fprintln(os.Stdout, "Opening the service route "+serviceSpec.URL[0]+" in the default browser...")
```
hferentschik: I think we can skip 'service' - "Opening route "+serviceSpec.URL[0]+" in the default browser..."

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,90 @@ var serviceCmd = &cobra.Command{
...
+		browser.OpenURL(serviceSpec.URL[0])
+
+	} else if serviceSpec.NodePort != "" {
+		nodePortURL := fmt.Sprintf("%s:%s", ip, serviceSpec.NodePort)
+		urlScheme := "http://"
+		if https {
+			urlScheme = "https://"
+		}
+		fmt.Fprintln(os.Stdout, "Opening the service nodeport "+urlScheme+nodePortURL+" in the default browser...")
```
hferentschik: "Opening NodePort "+urlScheme+nodePortURL+" in the default browser..."

I don't think 'service is needed here. Also, everywhere I see node ports mentioned, it is in the format "NodePort" never "nodeport".

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,90 @@ var serviceCmd = &cobra.Command{
...
+func stdOutURL(serviceSpec openshift.ServiceSpec) {
+	if serviceSpec.URL != nil {
+		fmt.Fprintln(os.Stdout, serviceSpec.URL[0])
+
+	} else {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Service '%s' in namespace '%s' does not have route associated which can be opened in the browser.", serviceSpec.Name, serviceSpec.Namespace))
+	}
+}
+
+func printToStdOut(serviceSpec openshift.ServiceSpec, ip string) {
```
hferentschik: 👍

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +53,90 @@ var serviceCmd = &cobra.Command{
...
+		atexit.ExitWithMessage(1, fmt.Sprintf("Service '%s' in namespace '%s' does not have route or nodeport associated which can be opened in the browser.", serviceSpec.Name, serviceSpec.Namespace))
+	}
+}
+
+func stdOutURL(serviceSpec openshift.ServiceSpec) {
+	if serviceSpec.URL != nil {
+		fmt.Fprintln(os.Stdout, serviceSpec.URL[0])
+
+	} else {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Service '%s' in namespace '%s' does not have route associated which can be opened in the browser.", serviceSpec.Name, serviceSpec.Namespace))
```
hferentschik: I think here we are a bit inconsistent:

      $ minishift openshift service mariadb-ingress -n myproject --inbrowser

The above will open in browser, but below:

     $ minishift openshift service mariadb-ingress -n myproject --url
     Service 'mariadb-ingress' in namespace 'myproject' does not have route associated which can be opened in the browser.

I think the same argument applied for --url as for --inbrowser. The user asks for it, let's give it to him. I think this is more consistent.

pkg/minishift/openshift/service.go
```diff
@@ -88,30 +88,22 @@ const (
 )
 
 // Get the route for service
-func GetServiceSpec(service, namespace string, https bool) (string, error) {
-	urlScheme := "http://"
```
hferentschik: +1

hferentschik:
So I have been playing with this for a while now and apart of some minor comments in the review, I have to agree that it sucks to having to type the namespace (as also @jorgemoralespou mentioned).

We always said one reason for that is that it is hard to determine the "current" namespace. What if we looked at this the wrong way? We can access all namespaces anyways (`service list`). Why bother with determining the namespace. We just get all service and if we have a service match we open/list it. Only if we have more than one service with the same name, we need to display something like:"Service x is defined in multiple namespaces. You need to explicitly use -n [namespace]". WDYT?

jorgemoralespou:
@hferentschik to be honest, I don't really understand what would be so difficult for minishift to get current namespace from the user's KUBECONFIG, the same way oc or kubectl does.

I'm also concerned that this can be an issue bigger when we look at supporting multiple profiles, so I would take that "work" into consideration for every change you do until then.

Anyway, you should be able to:
- identify which "minishift context/profile" you're talking to (whenever profiles are introduced this will be needed), so the .minishift should know what profiles are active.
- Get the namespace from context defined in the KUBECONFIG
- Use that namespace, and as the admin user (using the profile's admin_kubeconfig) and pass it transparently to these minishift commands

There was some discussion yesterday on the DevExp meeting that is relevant to all this and there's the need to better handle context management. I'll try to hook you into those conversations, whenever they progress.

cmd/minishift/cmd/openshift/service.go
LalatenduMohanty: @praveenkumar @hferentschik  keeping --url just for integration tests seems unfair to me as it is part of the user experience we are trying to fix. If I have 2 routes which route the url gives? I guess the first one.

praveenkumar:
> keeping --url just for integration tests seems unfair to me as it is part of the user experience we are trying to fix. If I have 2 routes which route the url gives? I guess the first one.

Why do you think it's unfair, It is perfect for scripting part if a user have 2 routes for a service then it will print first one. For user experience side we are already fixed the default stdout which provide tabular notification about serviceSpec.

LalatenduMohanty:
> @LalatenduMohanty Why do you think it's unfair, It is perfect for scripting part if a user have 2 routes for a service then it will print first one. For user experience side we are already fixed the default stdout which provide tabular notification about serviceSpec.

The issue is --url only seems to be useful for the integration tests but not for scripting in general. For users it is not required as we fixed `minishift openshift service` command. 

There are two things 
1. Making commands script friendly.
2. Providing command line flags only for integration testing. 

I am supportive of 1 but not for 2. Rather we should fix our integration tests. For example in this case we can use `oc get routes` in the place of `--url` with some code change in integration tests. 

If we want to make `minishift openshift service` script friendly then that would be a different discussion.

hferentschik:
> The issue is --url only seems to be useful for the integration tests but not for scripting in general.

It is also useful for the user, in case he wants to script something. The integration tests is just one of our use cases. Without it we have not alternative to the already existing --url functionality. The current default is a format which is hard/impossible to process in a script.

hferentschik:
@LalatenduMohanty Also with _oc_ we need additional processing to get the pure URL. It's definitely useful.

cmd/minishift/cmd/openshift/service.go
hferentschik: This is not only for integration test, but simply to get the single url to standard out. What you do with it is up to you. The table format is nice to read, but almost impossible to process in a script. That only the first URL is returned is a decision we made to keep things simpler. --url option or not this "issue" exists. If you open in browser you the same "issue". We decided to open the first.

cmd/minishift/cmd/openshift/service.go
```diff
@@ -50,25 +54,120 @@ var serviceCmd = &cobra.Command{
...
+	}
+	return namespceList
+}
+
+func printToStdOut(serviceSpecs []openshift.ServiceSpec, ip string) {
+	var data [][]string
+	var urls, weights string
+
+	for _, serviceSpec := range serviceSpecs {
+		fmt.Println(serviceSpec.Name)
```
hferentschik: I think here is an unwanted print statement:

    router
    docker-registry
    kubernetes
    mariadb
    mariadb-ingress
    nodejs-ex
    hawkular-cassandra
    hawkular-cassandra-nodes
    heapster
    hawkular-metrics
    |-----------|-----------|----------|--------------------------------------------------|--------|
    | NAMSEPACE |   NAME    | NODEPORT |                    ROUTE-URL                     | WEIGHT |
    |-----------|-----------|----------|--------------------------------------------------|--------|
    | myproject | nodejs-ex |          | http://nodejs-ex-myproject.192.168.99.145.nip.io |        |
    |-----------|-----------|----------|--------------------------------------------------|--------|

pkg/minishift/openshift/service.go
```diff
@@ -87,33 +85,6 @@ const (
 	ProjectsCustomCol = "-o=custom-columns=NAME:.metadata.name"
 )
 
-// Get the route for service
-func GetServiceSpec(service, namespace string, https bool) (string, error) {
```
hferentschik: Yeah, +1 for getting rid of this

hferentschik:
Getting there. Apart of the unwanted print statement, it works for the cases where there is no ambiguity. However, I think the case where the service matches more than one namespace is not handled consistently. 

    $minishift openshift service list
    |-----------------|--------------------------|----------------------|-----------------------------------------    -------------|--------|
    |    NAMSEPACE    |           NAME           |       NODEPORT       |                          ROUTE-URL                       | WEIGHT |
    |-----------------|--------------------------|----------------------|-----------------------------------------    -------------|--------|
    | default         | docker-registry          |                          |                                                      |        |
    |                 | kubernetes               |                          |                                                      |        |
    |                 | router                   |                          |                                                      |        |
    | foo             | nodejs-ex                |                      | http://    nodejs-ex-foo.192.168.99.145.nip.io           |        |
    | myproject       | mariadb                  |                          |                                                      |        |
    |                 | mariadb-ingress          | 192.168.99.145:31030     |                                                      |        |
    |                 | nodejs-ex                |                      | http://    nodejs-ex-myproject.192.168.99.145.nip.io     |        |
    | openshift-infra | heapster                 |                          |                                                      |        |
    |                 | hawkular-cassandra       |                          |                                                      |        |
    |                 | hawkular-cassandra-nodes |                          |                                                      |        |
    |                 | hawkular-metrics         |                      | http://    metrics-openshift-infra.192.168.99.145.nip.io |        |
    |-----------------|--------------------------|----------------------|-----------------------------------------    -------------|--------|



    $ minishift     openshift service nodejs-ex

    |-----------|-----------|----------|--------------------------------------------------|--------|
    | NAMSEPACE |   NAME    | NODEPORT |                    ROUTE-URL                     | WEIGHT |
    |-----------|-----------|----------|--------------------------------------------------|--------|
    | foo       | nodejs-ex |          | http://nodejs-ex-foo.192.168.99.145.nip.io       |        |
    | myproject | nodejs-ex |          | http://nodejs-ex-myproject.192.168.99.145.nip.io |        |
    |-----------|-----------|----------|--------------------------------------------------|--------|

Here I would also expect to get a message telling me that I need to specify the namespace, otherwise it gets inconsistent on how this works with --url or --inbrowser.
    
    $ minishift     openshift service nodejs-ex --url
    Service nodejs-ex exists in [foo myproject] namespaces, you need to chose a specific namespace using -n <    namespace_name>    

I think you use the default rendering of a string slice which read awkward. Something like:
"Service nodejs-ex exists in the namespaces foo, bar and myproject.  You need to chose a specific namespace using -n <namespace_name>"

or to make it simpler

"Service nodejs-ex exists in multiple namespaces (foo, bar, my project).  You need to chose a specific namespace using -n <namespace_name>"

LalatenduMohanty:
> It is also useful for the user, in case he wants to script something. The integration tests is just one of our use cases. Without it we have not alternative to the already existing --url functionality. The current default is a format which is hard/impossible to process in a script.

I am not convinced that it is useful because how do we assume that route is the only think user wants to script and not NodePort. Also `--url` just picks the first route, so it is not useful to users who wants a different route (if the app has multiple route) or user wants to go through the routes. 

I agree that the current output format (with table) is almost impossible to parse in a script. Hence my suggestion if we want to make it script friendly we should take it up in a separate issue.

hferentschik:
What I don't  understand is that your argument is around the _url_ flag. IMO your issue is about multiple routes which is inducement of --url. It also applies to --inbrowser. 

We had a discussion around this issue where we said that to allow to select a specific route, we would have to introduce some sort of order and index. That seems awkward and still does help you if you really want to access via a selected route. We decided to start with choosing the first one for now and see whether there is a need for something more. 

Also, even with multiple routes, it is backed by one service. So if your intention is to "just" access the app, it won't matter which one is chosen. If it does matter, you would need to specify the explicit route anyway, since that's the distinguishing factor. Introducing an index would be artificial data added by us.

> I am not convinced that it is useful because how do we assume that route is the only think user wants to script and not NodePort.

You can script the NodePort as well. --inbrowser as well as --url work for NodePort as well. Also a service won't ever have a route as well as NodePort.

> so it is not useful to users who wants a different route (if the app has multiple route) or user wants to go through the routes.

We are not saying that this command tailors all use cases. But here is the thing, if you want to hit the app via a specific route, how do you "select" this via the CLI without specifying precisely this route, in which case the whole command becomes useless. 

In this case you are probably better of using myapp.$(minishift ip).nip.io/myspecificroute. In this example you know which URL you need/want to access. The unknown factor is the VM IP.

> I agree that the current output format (with table) is almost impossible to parse in a script. Hence my suggestion if we want to make it script friendly we should take it up in a separate issue.

The current default format is great imo and aligns well with `minishift openshift service list`. --url and --inbrowser are specialisation of these command. The aim is to provide equivalent functionality to `minikube service` and the ability to get scriptable output.

LalatenduMohanty:
I can not help if you do not see the inconsistency we have with `minishift opensift service` which lists  route, nodeport etc in console , `--url` which only prints the first route in the console, then `--inbrowser` which opens a route in browser.  To me --route is redundant and it is not required at all.  I do not want to drag this further. If you feel strongly about it then ignore my previous comments.

hferentschik:
@praveenkumar Can you address the outstanding change request, most importantly the unwanted Println?

hferentschik:
Rebased and merged.

Issue #1090, opened 2017-06-30
Running minishift update gives "Updated successfully" erroneously 


gastaldi:
Running on Fedora 25, when I run `minishift update` in Minishift 1.1.0 I get the following output:

```
$ minishift update
A newer version of minishift is available.
Do you want to update from 1.1.0 to 1.2.0 now? [y/N]: y
Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/minishift-1.2.0-linux-amd64.tgz
 3.68 MiB / 3.68 MiB [============================================================================================================================================================] 100.00% 0s
 65 B / 65 B [====================================================================================================================================================================] 100.00% 0s

Updated successfully to minishift v1.2.0.

Do you want to update default addons? [y/N]: y
Default add-ons will be updated when executing the next minishift command.
$ minishift status
Minishift was upgraded from v1.1.0 to v1.1.0. Running post update actions.
--- Updating default add-ons ... OK
Default add-ons anyuid, admin-user, xpaas installed
Running
$ minishift version
minishift v1.1.0
```

Notice that when I do `minishift version`, the version is not really updated as the command output says. 

It works if I run it again with `sudo minishift update`

hferentschik:
We had this report once before. I think as well on Linux. Needs some investigation. Not sure why sudo would be required here. How did you install 1.1.0?

gastaldi:
@hferentschik I downloaded the 1.1.0 release manually from https://github.com/minishift/minishift/releases and moved the minishift binary to `/usr/bin` (which requires sudo permission)

LalatenduMohanty:
@gastaldi That explains it.  The binary actually gets replaced so If you have kept the old binary at a place which needs sudo access you need to call it with sudo.

LalatenduMohanty:
I can reproduce this. 

```
$ minishift version
minishift v1.1.0
$ which minishift 
/usr/bin/minishift
 
[lmohanty@LalatenduM-laptop minishift-test-update]A newer version of minishift is available.
Do you want to update from 1.1.0 to 1.2.0 now? [y/N]: y
Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/minishift-1.2.0-linux-amd64.tgz
 3.68 MiB / 3.68 MiB [========================================================================] 100.00% 0s
 65 B / 65 B [========================================================================] 100.00% 0s

Updated successfully to minishift v1.2.0.

$ minishift version
minishift v1.1.0
```

[...]

PR #1115, opened 2017-07-10
Issue #1090 Handle file permission error during update

praveenkumar:
Now it will handle file permission error as expected. We did ignore error during implementation which caused this issue.

```
$ minishift update
Do you want to update from 1.1.0 to 1.2.0 now? [y/N]: y
Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/minishift-1.2.0-linux-amd64.tgz
 3.68 MiB / 3.68 MiB [===========================================================================================================================================] 100.00% 0s
 65 B / 65 B [===================================================================================================================================================] 100.00% 0s
Update failed: open /usr/bin/.minishift.new: permission denied
```

### DIFF
--- a/docs/source/using/troubleshooting.adoc
+++ b/docs/source/using/troubleshooting.adoc
@@ -11,6 +11,28 @@ toc::[]
 
 This section contains solutions to common problems that you might encounter while using Minishift.
 
+[[minshift-update-failed-due-to-permission-denied]]
+== Permission denied error when updating Minishift
+
+When updating Minishift using `minishift update`, the update process needs write permissions for the Minishift binary as well as the directory in which it is located.
+Without these permisions `minishift update` will fail.
+For example, this issue might occur when installing Minishift as root.
+
+Workaround: When updating minishift, use `sudo minishift update` (Linux/MacOS).
+
+----
+$ which minishift
+/usr/bin/minishift
+
+$ minishift update
+A newer version of minishift is available.
+Do you want to update from 1.1.0 to 1.2.0 now? [y/N]: y
+Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/minishift-1.2.0-linux-amd64.tgz
+ 3.68 MiB / 3.68 MiB [===========================================================================================================================================] 100.00% 0s
+ 65 B / 65 B [===================================================================================================================================================] 100.00% 0s
+Update failed: open /usr/bin/.minishift.new: permission denied
+---- 
+
 [[special-characters-passwords]]
 == Special characters cause passwords to fail
 
--- a/pkg/minishift/update/update.go
+++ b/pkg/minishift/update/update.go
@@ -282,11 +282,11 @@ func updateBinary(binary string) error {
 		Hash: crypto.SHA256,
 	})
 	if err != nil {
-		fmt.Errorf("Cannot apply binary update: %s", err)
-		err := update.RollbackError(err)
-		if err != nil {
-			return errors.New(fmt.Sprintf("Failed to rollback update: %s", err))
+		rollbackErr := update.RollbackError(err)
+		if rollbackErr != nil {
+			return errors.New(fmt.Sprintf("Failed to rollback update: %s", rollbackErr))
 		}
+		return err
 	}
 
 	return nil


pkg/minishift/update/update.go
```diff
@@ -282,11 +282,11 @@ func updateBinary(binary string) error {
 		Hash: crypto.SHA256,
 	})
 	if err != nil {
-		fmt.Errorf("Cannot apply binary update: %s", err)
```
hferentschik: So this line did not really do anything. It returns an error which got ignored.

pkg/minishift/update/update.go
```diff
@@ -282,11 +282,11 @@ func updateBinary(binary string) error {
...
-		err := update.RollbackError(err)
-		if err != nil {
-			return errors.New(fmt.Sprintf("Failed to rollback update: %s", err))
+		rollbackErr := update.RollbackError(err)
+		if rollbackErr != nil {
```
hferentschik: So how does this fix things? Even in the old code _RollbackError_ is called. I have a hard time seeing the difference. Can you explain. I get that the _fmt.Errorf_ is bogus, but how does this fix the issue.

With this in place what happens now when the binary is owned by root?

praveenkumar: > So how does this fix things? Even in the old code RollbackError is called. I have a hard time seeing the difference.

As you said before that `fmt.Errorf` gets ignored and we were not returning it. Now we have err and rollbackErr variable. Here `err` variable keep what `error` returned by `Apply` and rollbackErr keep what returned by `RollbackError`. Previously we were not returning err which came from `Apply` and overwriting it with `RollbackError` but now we do return this err and if rollback fails then Rollback one.

hferentschik: Ahh, now I got it, thanks.

So do we need to do anything else? Is this error enough? Should it be more explicit? We could inspect the error and if it is the permission error we could create a more specific error. This makes, however, most sense when the permission error is its own error type, which I don't think it is. One would have to inspect the message itself, which is error prone.

That said, I think we should at least add something to the troubleshooting guide as well. Aka, if the minishift binary is owned by root, one needs to update as root as well (`sudo minishift update`). That's working, right?

praveenkumar: >So do we need to do anything else? Is this error enough? Should it be more explicit? We could inspect the error and if it is the permission error we could create a more specific error. This makes, however, most sense when the permission error is its own error type, which I don't think it is. One would have to inspect the message itself, which is error prone.

That's enough, we don't have to do anything else. 

> That said, I think we should at least add something to the troubleshooting guide as well. Aka, if the minishift binary is owned by root, one needs to update as root as well (sudo minishift update). That's working, right?

I put some documents to troubleshooting guide.

LalatenduMohanty: @praveenkumar good catch. 👍

docs/source/using/troubleshooting.adoc
```diff
@@ -11,6 +11,26 @@ toc::[]
 
 This section contains solutions to common problems that you might encounter while using Minishift.
 
+[[minshift-update-failed-due-to-permission-denied]]
+== Permission cause update to fail
+
+Depending on your operating system if you have minishift binary installed in location where current user does not have write permission then `minishift update` will fail.
+
```
LalatenduMohanty: `minishift update` command replaces the old binary with the new binary using the current user.  So if the user does not have access permission on the file system to replace the minishift binary then the `update` will fail.

docs/source/using/troubleshooting.adoc
```diff
@@ -11,6 +11,26 @@ toc::[]
...
+Workaround: When updating minishift use `sudo minishift update` (Linux/Unix).
```
LalatenduMohanty: Workaround: Please provide the user with necessary access permission or you can run the user in an elevated permission mode e.g. `sudo minishift update` in GNU/Linux.

LalatenduMohanty: Ideally we should add example for Windows or macOS. But not sure if you have them handy.

gbraad: the permission issue will not be same for Windows in the first place.

docs/source/using/troubleshooting.adoc
hferentschik: > o if the user does not have access permission on the file system to replace the minishift binary then the update will fail.

That's not quite correct. The user can have permissions to the binary and the update will still fail, if the directory the binary lives in is not writable.

docs/source/using/troubleshooting.adoc
hferentschik: > user in an elevated permission mode

IMO "elevated permission mode" is Windows terminology.

hferentschik: > the permission issue will not be same for Windows in the first place (needs to be tested more?)

I think if the Windows problem arises we create a new issue.

hferentschik:
@praveenkumar https://github.com/praveenkumar/minishift/pull/2

docs/source/using/troubleshooting.adoc
LalatenduMohanty: @hferentschik If the directory is not writable at all then it is a different issue isn't it? Otherwise how the minishift binary is present in the location.

hferentschik: It is not a different issue. In fact the error we are seeing is because the user does not have permissions to write to the directory. Have a closer look at the error message. The binary itself + the directory need to be writable to the user executing the update. 

> Otherwise how the minishift binary is present in the location.

Because you installed it as root.

docs/source/using/troubleshooting.adoc
```diff
@@ -11,6 +11,28 @@ toc::[]
...
 This section contains solutions to common problems that you might encounter while using Minishift.
 
+[[minshift-update-failed-due-to-permission-denied]]
+== Permission denied error when updating Minishift
+
+When updating Minishift using `minishift update`, the update process needs write permissions for the Minishift binary as well as the directory in which it is located.
+Without these permisions `minishift update` will fail.
+This is for example an issue when installing Minishift as root.
+
+Workaround: When updating minishift use `sudo minishift update` (Linux/Unix).
```
LalatenduMohanty: I do not think we need to add Linux/Unix as just Linux is fine. with Unix it most likely not consistent with other documentation.

hferentschik: I actually think it is not needed at all, if it comes down to it.

thatdocslady: "When updating Minishift, use..." (needs a comma)

praveenkumar: MacOS is based on Unix so I don't see any harm to mention it.

docs/source/using/troubleshooting.adoc
```diff
@@ -11,6 +11,28 @@ toc::[]
 
 This section contains solutions to common problems that you might encounter while using Minishift.
 
+[[minshift-update-failed-due-to-permission-denied]]
+== Permission denied error when updating Minishift
+
+When updating Minishift using `minishift update`, the update process needs write permissions for the Minishift binary as well as the directory in which it is located.
+Without these permisions `minishift update` will fail.
+This is for example an issue when installing Minishift as root.
```
thatdocslady: "For example, this issue might occur when installing Minishift as root." (grammar/syntax fix)

docs/source/using/troubleshooting.adoc
hferentschik: Right, so imo either is ok.

LalatenduMohanty: @praveenkumar @hferentschik  We have not mentioned `unix` in our existing documentation AFAICR, so why introducing it now? For macOS we mention macOS or OS X but not Unix.

hferentschik:
Merged, thanks.

Issue #1070, opened 2017-06-27
config set does not support new centos iso-url option


jimmidyson:
```bash
$ minishift config set iso-url centos
[centos url is not valid: parse centos: invalid URI for request]
```

centos needs to be added as a valid value to the validator.

Great release btw!

hferentschik:
> centos needs to be added as a valid value to the validator.

Good catch, thank you.

> Great release btw!

Thanks :-) On On

hferentschik:
Merged via pull request #1118

PR #1118, opened 2017-07-11
Issue #1070 Add centos and b2d alias to config set

praveenkumar:
Addresses issue  #1070

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -45,6 +45,11 @@ type Setting struct {
 
 var settingsList []Setting
 
+const (
+	B2dIsoAlias    = "b2d"
+	CentOsIsoAlias = "centos"
+)
+
 var (
 	// minishift
 	ISOUrl           = createFlag("iso-url", SetString, []setFn{IsValidUrl}, []setFn{RequiresRestartMsg}, true)
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -21,6 +21,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"os"
+
 	"github.com/asaskevich/govalidator"
 	units "github.com/docker/go-units"
 	"github.com/docker/machine/libmachine"
@@ -49,7 +51,6 @@ import (
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
-	"os"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


cmd/minishift/cmd/config/config.go
```diff
@@ -43,7 +43,11 @@ type Setting struct {
 	callbacks   []setFn
 }
 
-var settingsList []Setting
+var (
+	settingsList   []Setting
+	B2dIsoAlias    = "b2d"
```
gbraad: why `var` and not `const` ?

cmd/minishift/cmd/start.go
```diff
@@ -362,9 +360,9 @@ func determineIsoUrl(iso string) string {
 	iso = strings.ToLower(iso)
 
 	switch iso {
-	case b2dIsoAlias, "":
+	case startFlags.B2dIsoAlias, "":
```
gbraad: in this case we might also be able to use a internal variable to make `""` refer to a named option such as `DefaultIsoOption` to make it clearer why we have this in the case select.

cmd/minishift/cmd/start.go
```diff
@@ -58,9 +59,6 @@ const (
 	defaultProject = "myproject"
 	defaultUser    = "developer"
 
-	b2dIsoAlias    = "b2d"
-	centOsIsoAlias = "centos"
-
```
LalatenduMohanty: 👍

pkg/minikube/constants/constants.go
```diff
@@ -35,6 +35,7 @@ const (
 	DefaultCPUS                  = 2
 	DefaultDiskSize              = "20g"
 	UpdateMarkerFileName         = "updated"
+	DefaultISO                   = ""
```
gbraad: The risk we have with this is that a constant seems to need a value... and when they give it one, it fails the intended function. Better to leave it as a local named variable

LalatenduMohanty: This does not look right as B2D is the default ISO. I understand that the default thing is handled in start.go  but the question is if any other code uses this if yes, then we should fix it accordingly and see if we need this at all.

cmd/minishift/cmd/start.go
```diff
@@ -360,11 +358,12 @@ func calculateDiskSizeInMB(humanReadableDiskSize string) int {
 
 func determineIsoUrl(iso string) string {
 	iso = strings.ToLower(iso)
+	defaultISO := ""
```
hferentschik: not sure what you gain here with this variable. I think at least the name of the variable is confusing. It is not _isoNotSpecified_

hferentschik:
Rebased and merged, thanks.

Issue #1119, opened 2017-07-11
Add webhook to circleci for irc notification.


praveenkumar:

hferentschik:
> We do need to have a hook link from minibot like it have for travis and appveyor.

Sure. However, nothing will happen until we implement the receiving end. Do you have a link to the web hook format CircleCI sends? Are you by any chance planning to add the hook processing to Minibot?

hferentschik:
This is the matching issue in the Minibot repo - https://github.com/minishift/minibot/issues/9

hferentschik:
Circle CI web hook docs - https://circleci.com/docs/1.0/configuration/#notify

PR #1120, opened 2017-07-11
Issue #1119 Add webhook notifications to Circle CI

praveenkumar:
Depend on https://github.com/minishift/minibot/issues/20

Addresses issue #1119

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -39,3 +39,8 @@ test:
 general:
     artifacts:
         - "../src/github.com/minishift/minishift/out"
+
+## Notifications
+notify:
+    webhooks:
+        - url: http://minibot.34e99f76.svc.dockerapp.io:9009/hubot/circleci


circle.yml
```diff
@@ -39,3 +39,8 @@ test:
 general:
     artifacts:
         - "../src/github.com/minishift/minishift/out"
+
+## Notifications
+notify:
+    webhooks:
+        - url: http://minibot.34e99f76.svc.dockerapp.io:9009/hubot/circleci
```
hferentschik: So are you going to write the receiving end as well ;-)

LalatenduMohanty:
@hferentschik @praveenkumar  -1 for sending CI results to #minishift IRC. Personally I think it clutters the IRC.

hferentschik:
> -1 for sending CI results to #minishift IRC. Personally I think it clutters the IRC

We had some issues for a few days, where too much content was posted as well as duplicate content. I think this is under control now. Personally I like the instant feedback for various reasons. We can try to fine-tune the precise format more as well. 

But anyways, what does the rest of @minishift/minishift-dev think?

hferentschik:
Instead of canning this completely, here are some suggestions:

* Reduce number of lines 
   * We could have a single line (as we used to have in Travis)
   * We can skip the commit message and build URL, just "pull request #xyz passed"  
* Skip quotes

gbraad:
> Reduce number of lines
> Skip quotes

The quotes are not informational at all... and it is a first to reduce the amount of lines that are published

LalatenduMohanty:
> Personally I like the instant feedback for various reasons.

Do you like it because you do not have to go the web to see the status of the CI runs? Or something else. I am wondering if using minibot to see CI status for a PR is something you will be interested or will be beneficial for other folks e.g. `minibot pr-status 1108`

LalatenduMohanty:
In general, the only case which will be worthy (IMO) to send to IRC is when the master is broken.

hferentschik:
> Do you like it because you do not have to go the web to see the status of the CI runs?

Yes. Also the fact that I get notified instead of me having to check. It also is a bit of an ongoing standup. One sees who is doing what.

> I am wondering if using minibot to see CI status for a PR is something you will be interested or will be beneficial for other folks e.g. minibot pr-status 1108

You mean just the overall status of the pull request? That would be useful for sure. One could use the GitHub API for that. Still it is not the same. First it is again an active check from my side (in which case I might just refresh my browser) and even more unless I first manually check or unless I got a notification that a given pull request exist, I cannot ask for its status.

hferentschik:
> In general, the only case which will be worthy (IMO) to send to IRC is when the master is broken.

That's another distinction I forgot to mention in https://github.com/minishift/minishift/pull/1120#issuecomment-314367617. One can differentiate between master pushed and pull request builds. Maybe we can at least agree that notifications for punish to master as useful. IMO regardless of whether the build as successful or not. Given that master pushes happen not so often, I would think notifications would not considered to be clutter.

hferentschik:
> The quotes are not informational at all...

Sure, that's not their intend either. 

> and it is a first to reduce the amount of lines that are published

+1, I think disabling notifications in general is not the right approach, but finding something which strikes a balance

gbraad:
> Personally I like the instant feedback for various reasons.

I am only at all the times and this is not instant feedback. There is a very high signal-to-noise ratio at the moment. Actually, I am unable to track the status of my own build this way. I'd rather suggest a dashboard view for this... and this is what I also use for local builds for my gitlab instance.

hferentschik:
> I am only at all the times and this is not instant feedback.

Not sure what you mean. Maybe you missed a word? Or two?

> There is a very high signal-to-noise ratio at the moment

Fair enough.

> I'd rather suggest a dashboard view for this..

Which means? Can you elaborate?

gbraad:
Typos... distracted.... "online all the time"

A dashboard that shows build states pretty much as done at: https://github.com/minishift/minishift/pulls, but in slightly more detail, and I am sure this should be possible from the Github API, as each pull actually shows their state.

hferentschik:
You mean in textual form integrated into IRC? Or are you talking about building another HTML dashboard? And how does it get triggered? You ask for it? That's the status check as suggested by @LalatenduMohanty, right?

gbraad:
Pretty much... in this case a trigger can be an easy `<meta http-equiv="refresh" content="15">` if needed.

hferentschik:
@minishift/minishift-dev, so I reduced the lines printed per pull request notification to a single line. Let me know whether you still find this obstructive.  I will miss Chuck so ;-)

praveenkumar:
> I reduced the lines printed per pull request notification to a single line. Let me know whether you still find this obstructive. 

I can live with it since now it should not flooding channel I hope.

> A dashboard that shows build states pretty much as done at: https://github.com/minishift/minishift/pulls, but in slightly more detail, and I am sure this should be possible from the Github API, as each pull actually shows their state.

Where we should host it if that would be a dashboard (gh pages?). Then to get feedback I need to open it every-time but on irc I don't need any other things to be opened.

hferentschik:
> I can live with it since now it should not flooding channel I hope.

Let's see what others think now. 

> Where we should host it if that would be a dashboard (gh pages?). Then to get feedback I need to open it every-time but on irc I don't need any other things to be opened.

Right, my thinking as well. If it is just another website, then I might as well reload the GitHub pulls page.

hferentschik:
I merged this pull request. As for now it does not do anything, since we are missing the integration in Minibot - https://github.com/minishift/minibot/issues/9. 

Having the hook enabled though allows us to add integration when we want. We can then control things from the Minibot side.

Issue #1045, opened 2017-06-20
Allow add-ons to refer to environment variables


jstrachan:
it would be nice to be able to refer to environment variables inside an addon to aid parameterisation.

e.g. we'd like to pass in a github oauth app client id / secret when installing fabric8

hferentschik:
+1, variable support is high on my list of features as well.

hferentschik:
So two options really. 

1. Just place all environment variables into the interpolation context and allow interpolation without any further actions needed. This means, something like this will just work (assuming you want to interpolate say the Git Hub Auth token into an installed and enabled add-on):

        $ export GITHUB_TOKEN=1234567890
        $ minishift start
        # or
        $ minishift addons apply foo

1. Alternatively we make variables explicit. Instead of adding all environment variables into the interpolation context, you need to explicitly specify variables (aligning a bit with _docker_env_ and _openshift_env_):

        $ export GITHUB_TOKEN=1234567890
        $ minishift start --addon-env GITHUB_TOKEN=$GITHUB_TOKEN --addon-env FOO=BAR
        # or
        $ minishift addons apply --addon-env GITHUB_TOKEN=$GITHUB_TOKEN --addon-env FOO=BAR foo

    This can then also be paired with the config:

        $ minishift config set addon-env GITHUB_TOKEN=$GITHUB_TOKEN,FOO=BAR
        $ minishift start
        # or
        $ minishift addons apply foo

Obviously the latter is a bit more explicit. 

Bonus question, what should happen if the add-on has a token ${FOO} for which there is no value defined in the context?

1. Ignore and leave un-replaced
1. Ignore and leave un-replaced and print warning
1. Abort add-on application (treat as error)
1. Replace with empty string
1. ?

jstrachan:
I prefer 1 so that the user doesn't have to explicitly import env vars. Maybe env vars should be prefixed with something like `#{env.FOO}` in the addon file so its more clear its an env var and avoids clashes with addon variables?

It might be nice to support some syntax for a default value if the env var is not defined. e.g. something like
`#{env.FOO:-defaultValue}` kinda like bash or `#{env.FOO|defaultValue}`?

I can see some folks wanting to fail if an environment variable is not provided with a nice error message; so maybe we should use some syntax for that too - something like `#{!env.FOO}` or `#{mandatoryEnv.FOO}`?

hferentschik:
Gee, you really want it all ;-) I admit so, that a lot makes sense. 

[...]

PR #1122, opened 2017-07-11
Issue #1045 Introducing --addon-env for 'addon apply' and 'start' com…

hferentschik:
…mand

Addresses issue #1045

### DIFF
--- a/cmd/minishift/cmd/addon/addon_apply.go
+++ b/cmd/minishift/cmd/addon/addon_apply.go
@@ -31,26 +31,25 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/openshift"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
 	"gopkg.in/yaml.v2"
 )
 
-const (
-	routingSuffix = "routing-suffix"
+var (
+	addonsApplyCmd = &cobra.Command{
+		Use:   "apply ADDON_NAME ...",
+		Short: "Applies the specified add-ons.",
+		Long:  "Applies the specified add-ons. You can specify one or more add-ons, regardless of whether the add-on is enabled or disabled.",
+		Run:   runApplyAddon,
+	}
 )
 
-var addonsApplyCmd = &cobra.Command{
-	Use:   "apply ADDON_NAME ...",
-	Short: "Applies the specified add-ons.",
-	Long:  "Applies the specified add-ons. You can specify one or more add-ons, regardless of whether the add-on is enabled or disabled.",
-	Run:   runApplyAddon,
-}
-
 func init() {
+	addonsApplyCmd.Flags().AddFlag(util.AddOnEnvFlag)
 	AddonsCmd.AddCommand(addonsApplyCmd)
 }
 
 func runApplyAddon(cmd *cobra.Command, args []string) {
-
 	if len(args) == 0 {
 		atexit.ExitWithMessage(1, emptyAddOnError)
 	}
@@ -90,7 +89,7 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
 	for i := range args {
 		addonName := args[i]
 		addon := addOnManager.Get(addonName)
-		addonContext, err := clusterup.GetExecutionContext(ip, routingSuffix, ocRunner, sshCommander)
+		addonContext, err := clusterup.GetExecutionContext(ip, routingSuffix, viper.GetStringSlice(util.AddOnEnv), ocRunner, sshCommander)
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprint("Error applying the add-on: ", err))
 		}
--- a/docs/Rakefile
+++ b/docs/Rakefile
@@ -6,7 +6,12 @@ task :default => :gen
 BUILD_DIR          = "build"
 MARKDOWN_FILES     = Rake::FileList.new("source/_tmp/*.md")
 TOPIC_MAP          = "#{BUILD_DIR}/_topic_map.yml"
-ADOC_VARIABLES     = "source/variables.adoc"
+ADOC_VARIABLES     = FileList.new('source/variables.adoc',
+ 'source/contributing/variables.adoc',
+ 'source/getting-started/variables.adoc',
+ 'source/openshift/variables.adoc',
+ 'source/using/variables.adoc'
+)
 GENERATED_ADOC_DIR = "source/command-ref"
 ADOC_FILES = Rake::FileList.new do |list|
 	MARKDOWN_FILES.each do |f|
@@ -96,10 +101,13 @@ task :adoc_variables => :init do
   minishift_version  = ENV['MINISHIFT_VERSION'] || 'unset'
   openshift_version  = ENV['OPENSHIFT_VERSION'] || 'unset'
   centos_iso_version = ENV['CENTOS_ISO_VERSION'] || 'unset'
-  File.open(ADOC_VARIABLES, "w+") do |f|
-    f.write(":minishift-version: #{minishift_version}\n")
-    f.write(":openshift-version: #{openshift_version}\n")
-    f.write(":centos-iso-version: #{centos_iso_version}\n")
+
+  ADOC_VARIABLES.each do |f|
+    File.open(f, "w+") do |f|
+      f.write(":minishift-version: #{minishift_version}\n")
+      f.write(":openshift-version: #{openshift_version}\n")
+      f.write(":centos-iso-version: #{centos_iso_version}\n")
+    end
   end
 end
 
--- a/docs/source/using/addons.adoc
+++ b/docs/source/using/addons.adoc
@@ -11,13 +11,13 @@ toc::[]
 
 NOTE: This feature is still considered experimental and might change in future releases.
 
-Minishift allows to configure the vanilla OpenShift setup provided by *cluster up* with an add-on mechanism.
+Minishift allows to extend the vanilla OpenShift setup provided by *cluster up* with an add-on mechanism.
 
-Add-ons are directories that contain one or more text files with the *_.addon_* extension.
-The directory can also have other resource files such as JSON template files.
-However, only one *_.addon_* file is allowed in each directory.
+Add-ons are directories that contain a text file with the *_.addon_* extension.
+The directory can also contain other resource files such as JSON template files.
+However, only one *_.addon_* file is allowed per add-on.
 
-The following example shows the contents of the add-on text file, including the name and description of the add-on, additional metadata, and the actual commands to run.
+The following example shows the contents of an add-on file, including the name and description of the add-on, additional metadata, and the actual add-on commands to apply.
 
 [[example-anyuid-addon]]
 .Example: anyuid add-on definition file
@@ -25,22 +25,24 @@ The following example shows the contents of the add-on text file, including the 
 ----
 # Name: anyuid                                                                          // <1>
 # Description: Allows authenticated users to run images under a non pre-allocated UID   // <2>
+# Required-vars: ACME_TOKEN                                                             // <3>
 
-oc adm policy add-scc-to-group anyuid system:authenticated                              // <3>
+oc adm policy add-scc-to-group anyuid system:authenticated                              // <4>
 ----
 <1> (Required) Name of the add-on.
 <2> (Required) Description of the add-on.
-<3> Actual add-on command. In this case, the command executes the *oc* binary.
+<3> (Optional) Comma separated list of required interpolation variables. See xref:../using/addons.adoc#addon-variable-interpolation[Variable Interpolation]
+<4> Actual add-on command. In this case, the command executes the *oc* binary.
 
-NOTE: Comment lines can be inserted at anywhere in the file. Comment lines must start with the '#' character.
+NOTE: Comment lines, starting with the '#' character, can be inserted at anywhere in the file.
 
 Enabled add-ons are applied during xref:../command-ref/minishift_start.adoc#[`minishift start`], immediately after the initial cluster provisioning is successfully completed.
 
 [[addon-commands]]
 == Add-on Commands
 
-This section describes the command types that an add-on file can contain and what the commands do.
-The following command types are available forming a sort of mini-DSL for add-ons:
+This section describes the commands that an add-on file can contain.
+They are forming a mini-DSL for Minishift add-ons:
 
 ssh::
 If the add-on command starts with `ssh`, you can run any command within the Minishift-managed VM.
@@ -71,18 +73,18 @@ sleep::
 If the add-on command starts with `sleep`, it waits for the specified number of seconds.
 This can be useful in cases where you know that a command such as `oc` might take a few seconds before a certain resource can be queried.
 
-NOTE: Trying to add a command that is not listed will cause an error when the add-on gets parsed.
+NOTE: Trying to use an undefined command will cause an error when the add-on gets parsed.
 
 [[addon-variable-interpolation]]
-=== Variable Interpolation
+== Variable Interpolation
 
-Minishift also allows the use of variables as part of the supported commands.
+Minishift allows the use of variables within the add-on commands.
 Variables have the format `#{<variable-name>}`.
-
-The following example shows how the routing suffix can be interpolated into an `openshift` command to create a new certificate as part of securing the OpenShift registry.
+The following example shows how the OpenShift routing suffix can be interpolated into an `openshift` command to create a new certificate as part of securing the OpenShift registry.
+The used variable `#{routing-suffix}` is part of the built-in add-on variables.
 
 [[example-addon-routing-variable]]
-.Example: Usage of the routing suffix variable
+.Example: Usage of the routing-suffix variable
 
 ----
 $ openshift admin ca create-server-cert \
@@ -94,10 +96,14 @@ $ openshift admin ca create-server-cert \
   --key=/etc/secrets/registry.key
 ----
 
-The following table shows the supported variables that are available for interpolation.
+[[addon-built-in-variables]]
+=== Built-in Variables
+
+There exist several built-in variables which are available for interpolation at all times.
+The following table shows these variables.
 
 [[table-supported-addon-variables]]
-.Supported add-on variables
+.Supported built-in add-on variables
 
 |===
 |Variable |Description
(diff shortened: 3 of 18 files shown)


hferentschik:
retest this please

jorgemoralespou:
Here are my tests and comments:


Tested with following addon (test.addon):

````
echo FOO=#{FOO} #FOO
echo BAR=#{BAR} #BAR
echo BAZ=#{BAZ} #BAZ
````

Test 1 - Explicit addon enablement:
````
minishift-1122 addons apply test --addon-env FOO=BAR --addon-env BAR=$USER --addon-env BAZ=\$USER
-- Applying addon 'test':
  FOO=BAR #FOO
  BAR=jmorales #BAR
  BAZ= #BAZ
````

Test 2- Explicit addon enablement with no values set:
````
minishift-1122 addons apply test
-- Applying addon 'test':
  FOO=#{FOO} #FOO
  BAR=#{BAR} #BAR
 BAZ=#{BAZ} #BAZ
````

CONFIGURING DEFAULTS:
````
minishift-1122 config set addon-env FOO=BAR,BAR=$USER,BAZ=\$USER
cat ~/.minishift/config/config.json {
    "addon-env": [
        "FOO=BAR",
        "BAR=jmorales",
        "BAZ=$USER"
    ],
    "cpus": 2,
    "disk-size": "40g",
    "image-caching": true,
    "memory": 8192,
    "vm-driver": "xhyve"
}
````

Test 3 - Explicit addon enablement with configured defaults:
````
minishift-1122 addons apply test --addon-env FOO=BAR --addon-env BAR=$HOME --addon-env BAZ=\$HOME
-- Applying addon 'test':
  FOO=BAR #FOO
  BAR=/Users/jmorales #BAR
  BAZ= #BAZ
````
Test 4- Explicit addon enablement with no values set with configured defaults:
````
minishift-1122 addons apply test
-- Applying addon 'test':
  FOO=BAR #FOO
  BAR=jmorales #BAR
  BAZ= #BAZ
````

Comments:

- I would like a shortcut for --addon-env (-e) as it's too much writing
- I would like interpolation to happen also at resolution, so I can resolve a install time (example3 for BAZ)
- I failed  initially to set the ENVs within {} to resemble bash. It's OK, and it's documented, but wanted to comment.

Else, great job.

**Comment by hferentschik** [comment]
> I would like a shortcut for --addon-env (-e) as it's too much writing

Ok, ok. What's about '-a' instead though? '-e' is taken for --openshift-env. IMO this short form flags fall short of making the intend clear, but ok.

**Comment by hferentschik** [comment]
> I would like interpolation to happen also at resolution, so I can resolve a install time (example3 for BAZ)

Not quite sure I understand what you mean, but I guess you mean you want to see literal $HOME as value for BAZ, right. Do you mind trying something like \\\$HOME? I think the problem is that under the hood we are using regular expressions for the interpolation and '$' will be treated as meta character.

**Comment by jorgemoralespou** [comment]
> Not quite sure I understand what you mean, but I guess you mean you want to see literal $HOME as value for BAZ, right. Do you mind trying something like \$HOME? I think the problem is that under the hood we are using regular expressions for the interpolation and '$' will be treated as meta character.

So, what I would like is to have in the "config" \$VAR or something that will resolve at install time. (e.g. when I want to inject an env that will be changing with time, so I don't want to resolve the value when I add the config but when the addon is installed. If I have create a minishift instance with that addon installed now will have the value it has now, if I create another minishift instance with a new value in my shell, I want that new value to be picked up. 

e.g:
````
$ minishift-1122 config set addon-env BAZ=\$MY_KEY
$ export MY_KEY=123
$ minishift start
$ minishift-1122 addon enable test 
BAZ=123
$ minishift stop
export MY_KEY=567
$ minishift start
$ minishift-1122 addon enable test 
BAZ=567
```` 

And the syntax you refer is what I tried for BAZ, so it's using \$HOME when setting the config, although then it's not in the config file.
If I edit manually the config file to add the \, I get:

````
minishift-1122 addons apply test
Cannot read the Minishift configuration: Cannot decode config /Users/jmorales/.minishift/config/config.json: invalid character '$' in string escape code
````

**Comment by hferentschik** [comment]
Pushed an update adding '-a' as shortcut. Also rebased.

**Comment by hferentschik** [comment]
> So, what I would like is to have in the "config" $VAR or something that will resolve at install time. (e.g. when I want to inject an env that will be changing with time, so I don't want to resolve the value when I add the config but when the addon is installed.

Got you now. This is a bit tricky in terms of syntax. I know '$' is obvious for Linux and macOS, but on Windows 10 it would for example be '$env:VAR'. That's for powershell, cmd would be different again. So the question is whether one needs to handle all this cases or just uses "$" or "env.", either way, this would be Minishift specific syntax and not be driven by the shell's built-in interpolation. Not sure whether we should really go down this road.

File: docs/source/using/addons.adoc [hunk 1]
```diff
@@ -112,6 +112,23 @@ The following table shows the supported variables that are available for interpo
 |Name of the current add-on.
 |===
 
+The commands xref:../command-ref/minishift_start.adoc#[`minishift start`] as well as xref:../command-ref/minishift_addons_apply.adoc#[`minishift addons apply`] also provide a `--addon-env` flag which allow to pass variables for interpolation dynamically, for example:
+
+----
+$ minishift addons apply --addon-env ACME_TOKEN=1234 acme
+----
+
+The `--addon-env` flag can be specified multiple times to define more than one variables for interpolation.
```
LalatenduMohanty: You should mention the short form as well i.e. -a . Also we should mention about how it can take variables from environment.

hferentschik: > You should mention the short form as well i.e. -a .

I don't think this is required. We do this at no other point in the docs. Things just get cumbersome to describe and sentences unnecessarily complicated. People who want this option will check the command ref.

> Also we should mention about how it can take variables from environment

What do you mean? This would then again go into variable interpolation of the shell which is a topic outside of the scope of Minishift.

hferentschik:
Pushed another update which fixes the issue with '$' in the replacement value which was treated as special character. So the result for BAZ in the example would be `BAZ=$USER`. For now I suggest to leave it so and not try to do environment substitution based on the '$'.

jorgemoralespou:
I'll think about this and give you a reply once I'm back from a customer
meeting. But I'm the way I've thought of that we should have a flag, either
on the variable used or on the add-on description of what variables are
required, so that if there's no value provided the addon don't try to be
installed. Some of these variables will be usually mandatory/required and
we should allow to prevent errors.

hferentschik:
> I'll think about this and give you a reply once I'm back from a customer meeting

Ok

> But I'm the way I've thought of that we should have a flag, either on the variable used or on the add-on description of what variables are required, so that if there's no value provided the addon don't try to be installed. 

I guess that ties into the discussion we had in the issue and supporting something like #{!FOO}, which would imply, fail if FOO cannot  be replaced. Maybe even something like #{!FOO:My custom error message in case FOO cannot be replaced}. This would be one way of solving this. The other is as you say to add addition meta data to the add-on header. Possible as well, but it might not be as obvious, since now the use of the variable is de-coupled from where it is defined as required. I am definitely not against this feature, but it requires a bit more work in how we replace things. Maybe this would be better to track in a different issue?

jorgemoralespou:
As you think it's better tracked, then let it be, but I'll push for this one, since I'm thinking on start creating addons and this would be a must.

I would prefer to fail fast, meaning that if I don't specify the variable, don't execute the addon to the point the variable is defined and fail there' as that leaves the addon in an inconsistent state, and there's no way to recover (so far, like addon remove).

jorgemoralespou:
> Pushed another update which fixes the issue with '$' in the replacement value which was treated as special character. So the result for BAZ in the example would be BAZ=$USER. For now I suggest to leave it so and not try to do environment substitution based on the '$'.

So to this one, my opinion. I would like to be able to do substitution at the time of instantiation and not at the time of configuration, and using config as where the value would be set would not work.

If having it cross platform is an issue because of the $, let it be something else, as #{env.USER} would replace at instantiation time from the environmnet in the proper way depending on the os minishift is run. 

So that would leave that, if I want inmediate replacement at apply, I would do:

````
minishift addons apply test -a BAZ=$MY_KEY
````

If I want to add it into the config, I would do:

````
minishift config add addon-env BAZ=env.MY_KEY
````

or

````
minishift config add addon-env BAZ=$MY_KEY
````
 
Depending on whether I want MY_KEY to be replaced at config or instantiation time. This would add some power to those environment variables that change with time. Not many, but some.

File: docs/source/using/addons.adoc (hunk 1 above)
Comment (LalatenduMohanty) [inline docs/source/using/addons.adoc]: >What do you mean? This would then again go into variable interpolation of the shell which is a topic outside of the scope of Minishift.

Sure, we do not need to go the details which is out of scope of Minishift, but what about mentioning a single or couple of sentence about it. The idea should be to tell users that this capability also exists. If you are saying that its not a recommended method then it is fine not to mention in the docs.

**Comment by hferentschik** [comment]
> If having it cross platform is an issue because of the $, let it be something else, 

Ok, we agree on this.

> So that would leave that, if I want inmediate replacement at apply, I would do:

         minishift addons apply test -a BAZ=$MY_KEY

Right, so that's what we have now. Whether it is $ or $env or whatever, depends on your shell type. This is standard interpolation offered by your OS / shell.

> If I want to add it into the config, I would do:

        minishift config add addon-env BAZ=env.MY_KEY

Right, so this is one option. I think we really need to introduce in this case some Minishift specific notation, to avoid confusion with shell behaviour which differs depending on OS and shell type. This is definitely an option. 

Question is whether in this case it would make more sense to push this into to the add-on itself (see discussion on the original issue). So in the case where one wants to dynamically replace with an environment value, you can place #{env.MY_KEY} into the add-on. In this case you would not even have to specify --addon-env (just trying to think through some alternatives).

**Comment by hferentschik** [comment]
> As you think it's better tracked, then let it be, but I'll push for this one, since I'm thinking on start creating addons and this would be a must.

Fair enough. We agree that we want this in one form or the other. 

> I would prefer to fail fast, meaning that if I don't specify the variable, don't execute the addon to the point the variable is defined and fail there' as that leaves the addon in an inconsistent state, and there's no way to recover (so far, like addon remove).

That's a good point. It requires to change a bit how we process add-ons. ATM, we do it line based , but to fail early we would have to add a pre-apply step where we check whether all variables can be substituted. I think that's doable. If we were to add this, early failure in case an add-on contains variables which there is no replacement defined, would that be enough for you for now?

**Comment by jorgemoralespou** [comment]
> Question is whether in this case it would make more sense to push this into to the add-on itself (see discussion on the original issue). So in the case where one wants to dynamically replace with an environment value, you can place #{env.MY_KEY} into the add-on. In this case you would not even have to specify --addon-env (just trying to think through some alternatives).

The main concern I have about this is that I think there will be 2 types of envs:

- those provided at instantiation time and required (typically will be user provided, although could fall back to a real OS ENV, by using --addon-env BAZ=$FOO or in config by having BAZ=env.FOO)
- Those provided at config time. These will always go to the config file, and would have the real value in the config file via BAZ=MY_VALUE or instruct to be replaced at instantiation by env.FOO

I think there's 2 different sides to this, one the addon creator, that will expect/require a VAR and the addon instantiation (the final user) that will want to provide the value in whatever form he wants. Forcing him to have the VAR defined as an env by using #{env.BAR} notation in the addon is taking a decision for him we shouldn't take. IMHO

Whether you have a pre-apply step that looks into the defined vars, with a #{!BAZ} notation or a descriptor comment of # Required-vars: BAZ is an implementation detail that is up to you, but yes, this would be enough, with the caveat that "there might be some variables that could be empty and some that not", so I would say that it needs to be implicitly defined which variables can not fail.

**Comment by jorgemoralespou** [comment]
+1

**Comment by hferentschik** [comment]
@jorgemoralespou, @LalatenduMohanty - I pushed another update. This will now allow to use env.FOO to dynamically replace with the value of the environment variable at the time the add-on gets applied.

**Comment by jorgemoralespou** [comment]
@hferentschik I tried:

with this text.addon:
````
# Name: test
# Description: Tests addons-env
# Url:

echo FOO=#{FOO} #FOO
echo BAR=#{BAR} #BAR
echo BAZ=#{BAZ} #BAZ
echo BAT=#{env.USER} #BAZ
````

With this config:

````
    "addon-env": [
        "FOO=BAR",
        "BAR=jmorales",
        "BAZ=env.HOME"
    ],
````

Got this:
````
minishift-1122 addons apply test
-- Applying addon 'test':
  FOO=BAR #FOO
  BAR=jmorales #BAR
  BAZ=/Users/jmorales #BAZ
  BAT=#{env.USER} #BAZ
````

Expected this:
````
minishift-1122 addons apply test
-- Applying addon 'test':
  FOO=BAR #FOO
  BAR=jmorales #BAR
  BAZ=/Users/jmorales #BAZ
  BAT=jmorales #BAZ
````

**Comment by hferentschik** [comment]
I thought we said, that we don't support env.VAR in the add-on itself?

You where expecting that

    BAT=#{env.USER} 

Gets replaced with 

   BAT=jmorales

Where 'jmorales' the value of your $USER environment variable. 

This is not how this works atm. You would have to add _BAT_ to your config as well.

**Comment by jorgemoralespou** [comment]
I was expecting that using env. notation would serve as a marker to resolve to the environment in both places. But I guess I can live with this, although I think that it would be more consistent.

Don't really want to make much noise on this. This is excellent job so far. I'm loving it.

**Comment by hferentschik** [comment]
Now with 'Required-vars' as meta data header :-), e.g.

    # Required-vars: ACCESS_TOKEN

Will still add some docs and maybe a couple of tests.

**Comment by hferentschik** [comment]
Now with updated docs as all. I changed 'Required-vars' to 'Required-Vars'.

**Comment by LalatenduMohanty** [comment]
@hferentschik  With respect to Required-Vars it does not work as expected for me. Not sure why.

```
$ minishift addons apply --addon-env GOPATH=env.GOPATH anyuid
-- Applying addon 'anyuid':
  /Users/lmohanty/code/gocode
  #{xyz}
```

the addon content is

```
 # Name: anyuid
 # Description: Changes the default security context constraints to allow pods to run as any user.
 # Url: https://docs.openshift.org/latest/creating_images/guidelines.html#openshift-origin-specific-guidelines
Required-Vars: xyz

  echo #{GOPATH}
  echo #{xyz}
  echo #{abc}
```

**Comment by LalatenduMohanty** [comment]
@hferentschik Nevermind I missed the `#` before Required-Vars

```
$ minishift addons apply --addon-env GOPATH=env.GOPATH anyuid
-- Applying addon 'anyuid':Error applying the add-on: The variable(s) xyz are required by the add-on, but are not defined in the context.
```

File: docs/Rakefile [hunk 2]
```diff
@@ -6,7 +6,12 @@ task :default => :gen
 BUILD_DIR          = "build"
 MARKDOWN_FILES     = Rake::FileList.new("source/_tmp/*.md")
 TOPIC_MAP          = "#{BUILD_DIR}/_topic_map.yml"
-ADOC_VARIABLES     = "source/variables.adoc"
+ADOC_VARIABLES     = FileList.new('source/variables.adoc',
+ 'source/contributing/variables.adoc',
+ 'source/getting-started/variables.adoc',
+ 'source/openshift/variables.adoc',
+ 'source/using/variables.adoc'
+)
```
LalatenduMohanty: You broke it down because you thought there would be name/vale clash for variables?

praveenkumar:
@hferentschik let me know if below is expected. My understanding is we don't support `env.VAR` as part of addon file for interpolation now either a user have to set each var as part of config or he need to specific each var as part of `addon-env` options. 

```
$ cat test.addon 
# Name: test
# Description: Tests addons-env
# Url:

echo FOO=#{FOO} #FOO
echo BAR=#{BAR} #BAR
echo BAZ=#{BAZ} #BAZ
echo BAT=#{BAT} #BAT

$ ./minishift config view
- addon-env            : [FOO=bar BAR=foo BAZ=hello]

$ ./minishift addons --addon-env BAT=env.USER apply test
-- Applying addon 'test':
  FOO=#{FOO} #FOO
  BAR=#{BAR} #BAR
  BAZ=#{BAZ} #BAZ
  BAT=prkumar #BAT => Since we used addon-env, config part is ignored.

$ ./minishift addons apply test
-- Applying addon 'test':
  FOO=bar #FOO
  BAR=foo #BAR
  BAZ=hello #BAZ
  BAT=#{BAT} #BAT
```

docs/Rakefile
hferentschik: > You broke it down because you thought there would be name/vale clash for variables?

No. The problem is with integration of Middleman and Asciibinder (docs.openshift.org). The latter requires includes to be relative to the root of the docs whereas Middleman does it relative to the current directory. Originally we had some customisation in _config.rb_, to make Middleman change the include path. This worked ok. Then recently we made it possible to add images as well. There things are relative, in this case also in AsciiBinder. I actually removed at this stage the customisation. This made image includes work, but broke accidentally the includes of variables. 

We could try to get the customisations back, but then we would somehow have to differentiate between image includes and adoc includes. I think this would be hard. The alternative is to duplicate the variables file, so that it works in all cases. In the case of Middleman, it uses the variables.adoc in each directory. docs.openshift.org will only use the top level one. Since they are all the same, no harm is done. Seemed to be the easiest way to "fix" it.

hferentschik:
>  My understanding is we don't support env.VAR as part of addon file for interpolation now either a user have to set each var as part of config or he need to specific each var as part of addon-env options.

Right

jorgemoralespou:
@praveenkumar, I felt the same at first, but thinking more about it, I like making the user more conscious where the car comes from.

docs/Rakefile
hferentschik: > You broke it down because you thought there would be name/vale clash for variables?

To be clear, the files are identical and need to be. This is not about potentially having different content in each of them. This is about making sure that variable interpolation works for us using Middleman as well as integration into docs.openshift.org.

hferentschik:
> but thinking more about it, I like making the user more conscious where the var comes from.

Right. That's my thinking as well. Also as we discussed before, if we enable the use of env.FOO in the add-on, add-ons can interpolate whatever, w/o the user knowing what's going on. Right now, everything is explicit w/o really limiting functionality.

Also there is a slight semantic difference. Take:

    ./minishift addons --addon-env BAT=env.USER apply test

So _BAT_ is the key and the value is env.USER where we say that using the prefix _env_ makes the value special in terms of that we retrieve the underlying environment variable. 

Now assume we add  'env.USER' to the add-on:

    echo Hello #{env.User}

So here env.User is used as key. If we enable this, we are basically saying that keys and values need to expand against environment variables. So there is a difference.

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -31,26 +31,25 @@ import (
...
 
-const (
-	routingSuffix = "routing-suffix"
+var (
+	addonsApplyCmd = &cobra.Command{
+		Use:   "apply ADDON_NAME ...",
+		Short: "Applies the specified add-ons.",
+		Long:  "Applies the specified add-ons. You can specify one or more add-ons, regardless of whether the add-on is enabled or disabled.",
+		Run:   runApplyAddon,
+	}
```
praveenkumar: 👍

docs/source/using/addons.adoc
```diff
@@ -112,6 +118,45 @@ The following table shows the supported variables that are available for interpo
...
+----
+$ minishift config set addon-env PROJECT_USER=env.USER        // <1>
+$ minishift addons apply acme                                 // <2>
+----
+<1> Using the _env_ prefix ensures that instead of literally replacing '#{PROJECT_USER}' with 'env.USER', the value of the environment variable `USER` is used.
+If the environment variable is not set not, interpolation does not occur.
+<2> When the add-on is applied, each occurrence of `#{PROJECT_USER}` within an add-on command gets replaced with the value of the environment variable `USER`.
+
+As add-on developer, you can enforce that a given variable is provided by the user during add-on application.
+This can be achieved by adding the _Required-Vars_ metadata header:
```
hferentschik: > What is the suggested way to add multiple variables to Required-Vars, comma separated right?

Added it to docs. Comma separated

hferentschik: > suggestion You can enforce that a variable value needs to be provided when the add-on is applied by adding it to the _Required-Vars_ metadata header:

Changed the wording

hferentschik:
Merged.

Issue #344, opened 2017-01-30
Incomplete ISO download makes startup fail


cmoulliard:
Version used : 1.0.0.Beta3

error 

```
minishift start --openshift-version=v1.4.1 --memory=4000 --vm-driver=virtualbox --iso-url=https://github.com/minishift/minishift-centos-iso/releases/download/v1.0.0-rc.1/minishift-centos7.iso --docker-env=[storage-driver=devicemapper]
Starting local OpenShift cluster using 'virtualbox' hypervisor...
Downloading ISO 'https://github.com/minishift/minishift-centos-iso/releases/download/v1.0.0-rc.1/minishift-centos7.iso'
 295.52 MB / 329.00 MB [==================================================================================================================================================================>------------------]  89.82% 13s
E0130 11:01:36.831166    3152 start.go:135] Error starting the VM: read tcp 192.168.1.80:53769->52.216.64.240:443: read: operation timed out. Retrying.
^C
dabou-macosx:~$ minishift start --openshift-version=v1.4.1 --memory=4000 --vm-driver=virtualbox --iso-url=https://github.com/minishift/minishift-centos-iso/releases/download/v1.0.0-rc.1/minishift-centos7.iso --docker-env=[storage-driver=devicemapper]
Starting local OpenShift cluster using 'virtualbox' hypervisor...
E0130 11:16:50.500950    3660 start.go:135] Error starting the VM: Error starting stopped host: Too many retries waiting for SSH to be available.  Last error: Maximum number of retries (60) exceeded. Retrying.
Password:
```

hferentschik:
ok, maybe a comment from issue #343 was actually supposed to go on this issue. I think we have a bug here.

It looks like the following happened. The ISO downloaded started and got aborted. My guess is the file was partially downloaded and existed on the file system. Our current checks only verify the ISO based on file name, so on the second attempt, we assume there is a valid ISO, but it is actually corrupted. See also https://github.com/minishift/minishift/blob/9d320e02683c857aa395e2adcc65fcf42a68db64/pkg/minikube/cluster/cluster.go#L261-L261

So our algorithm needs to be a bit smarter and also does checksum verification. If the sha is not there, the download needs to restart, if it is there, we need to compare the sha against the ISO and if they don't match re-trigger the download...

budhrg:
[...]

PR #1124, opened 2017-07-13
Issue #344 Add tmp file to contain ISO download

praveenkumar:


pkg/minikube/cluster/cluster.go
```diff
@@ -224,6 +225,12 @@ func createDriverOptions(driver drivers.Driver, explicitOptions map[string]inter
 
 func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 	fmt.Println(fmt.Sprintf("Downloading ISO '%s'", m.MinikubeISO))
+	tmpISOFile, err := ioutil.TempFile("", "tmpISOFile")
+	fmt.Println("TMP FILE IS", tmpISOFile.Name())
```
hferentschik: I guess that's not intended for the user, right?

pkg/minikube/cluster/cluster.go
```diff
@@ -224,6 +225,12 @@ func createDriverOptions(driver drivers.Driver, explicitOptions map[string]inter
...
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpISOFile.Name())
 
 	// store the iso inside the MINISHIFT_HOME dir
```
hferentschik: I think this comment is a bit misleading here. Maybe instead of this comment we should add a proper GoDoc description on what this method does.

pkg/minikube/cluster/cluster.go
```diff
@@ -248,12 +255,29 @@ func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 		}()
 	}
 
+	if _, err = io.Copy(tmpISOFile, iso); err != nil {
```
hferentschik: We have a ffilehelper.CopyFile method which does all of this.

pkg/minikube/cluster/cluster.go
```diff
@@ -224,6 +225,12 @@ func createDriverOptions(driver drivers.Driver, explicitOptions map[string]inter
...
 	fmt.Println(fmt.Sprintf("Downloading ISO '%s'", m.MinikubeISO))
+	tmpISOFile, err := ioutil.TempFile("", "tmpISOFile")
+	fmt.Println("TMP FILE IS", tmpISOFile.Name())
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpISOFile.Name())
 
 	// store the iso inside the MINISHIFT_HOME dir
 	response, err := http.Get(m.MinikubeISO)
```
hferentschik: Maybe we could also append 'sha256' to the URL and check whether there is a checksum. If there is, we can also add checksum verification of the download capturing potential transfer problems.

pkg/minikube/cluster/cluster.go
hferentschik: For what it's worth, using something like ISO filename + ".part" might be a bit more effective. You basically download already into the right directory, just with the suffix '.part' or '.tmp'. Once the download is complete and verified you just rename the file. Renaming is cheaper than another copy and probably less error prone. I guess we can do either one.

pkg/minikube/cluster/cluster.go
praveenkumar: @hferentschik yes we can do but even for our github iso download url we don't have checksum appended so I didn't included it atm.

pkg/minikube/cluster/cluster.go
praveenkumar: @hferentschik make sense.

pkg/minikube/cluster/cluster.go
praveenkumar: @hferentschik ah, it's not for intended user, I was just testing something :(

hferentschik: thought so :-)

pkg/minikube/cluster/cluster.go
hferentschik: I think we do have it - https://github.com/minishift/minishift-b2d-iso/releases, https://github.com/minishift/minishift-centos-iso/releases

pkg/minikube/cluster/cluster.go
```diff
@@ -222,10 +223,16 @@ func createDriverOptions(driver drivers.Driver, explicitOptions map[string]inter
...
 
+// CacheMinikubeISOFromURL download minishift ISO from a given URI.
+// It also checks sha256sum if present and then put ISO to cached directory.
 func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 	fmt.Println(fmt.Sprintf("Downloading ISO '%s'", m.MinikubeISO))
+	tmpISOFile, err := ioutil.TempFile(filepath.Join(constants.Minipath, "cache", "iso"), "minishift.part")
+	if err != nil {
+		return err
+	}
+	defer os.RemoveAll(tmpISOFile.Name())
```
hferentschik: If you do the '.part' approach you don't want to delete the tmp file. You literally rename. See also https://github.com/minishift/minishift/blob/83e6821864e371e6ab968e950e4a04ccd489083b/pkg/minishift/docker/image/image_handler.go#L246-L246

pkg/minikube/cluster/cluster.go
```diff
@@ -248,12 +255,33 @@ func (m *MachineConfig) CacheMinikubeISOFromURL() error {
...
+		return err
+	}
+
+	checksumURL := fmt.Sprintf(m.MinikubeISO + ".sha256")
+	checksumResp, err := http.Get(checksumURL)
+	defer checksumResp.Body.Close()
+	if err != nil {
+		return err
+	}
+	if checksumResp.StatusCode != 404 {
```
hferentschik: I think that might fall a bit short. There might be many other things which can go wrong. 404 means file not found, but there might be other HTTP errors. I think you need to define this the other around, anything except a 200 is in this context a failure.

pkg/util/utils.go
```diff
@@ -152,3 +160,42 @@ func FriendlyDuration(d time.Duration) time.Duration {
 	d2 := (d / time.Nanosecond) * (time.Nanosecond)
 	return d2
 }
+
+// ChecSha256Sum takes a file handler and a http-response and match the sha256sum
+// and return error if not equal
+func CheckSha256Sum(source *os.File, checksumResp *http.Response) error {
+	if !crypto.SHA256.Available() {
```
hferentschik: +1 nice one

pkg/minikube/cluster/cluster.go
```diff
@@ -222,10 +223,15 @@ func createDriverOptions(driver drivers.Driver, explicitOptions map[string]inter
 	return checkFlags, nil
 }
 
+// CacheMinikubeISOFromURL download minishift ISO from a given URI.
+// It also checks sha256sum if present and then put ISO to cached directory.
 func (m *MachineConfig) CacheMinikubeISOFromURL() error {
 	fmt.Println(fmt.Sprintf("Downloading ISO '%s'", m.MinikubeISO))
+	tmpISOFile, err := ioutil.TempFile(filepath.Join(constants.Minipath, "cache", "iso"), "minishift.part")
```
hferentschik: Let's make this 

    tmpISOFile, err := os.Create(m.GetISOCacheFilepath() + ".part")

Then one has some understandable filenames while download is in process, but also when it fails.

hferentschik:
Merged, thanks.

Issue #485, opened 2017-02-27
Improve feedback to the user during startup


gbraad:
As part of an improved user experience, the user should be told what defaults has been chosen for him, or which values he has given, to make the starting process easier to understand.

For instance,
  * initial/default values can be conveyed to the user
  * Misinterpreted values become visible (use as 4 or 4096, when the user expects something to be in GB?)
  * Progress tracking, which can indicate a point of failure/slowdown to be user, without having to resort to the libmachine logs.

bronhaim:
- The defaults are presented as part of the help doc which is enough imo. 
- start --memory flag accepts only int value for amount of mb - so --memory 4GB doesn't exist.
- The progress tracking - do you mean some kind of ./minishift events watch command ? not sure its really needed. there is status command which is good enough. maybe we can improve the output there.

Maybe this issue should be about improving "start" output. currently its the content below. maybe it should print also vm's details - how much memory, cpus it uses for example.

Although this info is quite easy to reach using virsh dumpxml.. let me know if you want that


--- snip
Starting local OpenShift cluster using 'kvm' hypervisor...
Provisioning OpenShift via '/home/ybronhaim/.minishift/cache/oc/v1.4.1/oc [cluster up --use-existing-config --host-config-dir /var/lib/minishift/openshift.local.config --host-data-dir /var/lib/minishift/hostdata --host-volumes-dir /var/lib/minishift/openshift.local.volumes]'
-- Checking OpenShift client ... OK
-- Checking Docker client ... OK
-- Checking Docker version ... OK
-- Checking for existing OpenShift container ... 
   Deleted existing OpenShift container
-- Checking for openshift/origin:v1.4.1 image ... OK
-- Checking Docker daemon configuration ... OK
-- Checking for available ports ... OK
-- Checking type of volume mount ... 
   Using Docker shared volumes for OpenShift volumes
-- Creating host directories ... OK
-- Finding server IP ... 
   Using 192.168.42.143 as the server IP
-- Starting OpenShift container ... 
   Starting OpenShift using container 'origin'
   Waiting for API server to start listening
   OpenShift server started
-- Removing temporary directory ... OK
-- Server Information ... 
   OpenShift server started.
   The server is accessible via web console at:
       https://192.168.42.143:8443

   To login as administrator:
       oc login -u system:admin

gbraad:
[...]

Issue #418, opened 2017-02-16
Hyper-V virtual networking; NAT and IPv6


gbraad:
When a VM is started on Hyper-V, connected to a virtual switch that is configured for NAT, the virtual machine will not receive an IPv4 address (as no DHCP server runs), but gets assigned a IPv6 address. Eg. when using 'DockerNAT'.

However, this leads to unexpected behaviour. The VM is 'Running', but SSH can not connect (and the process seems to be hanging). On a separate cmd prompt the following can be seen:
```
C:\Users\gbraad\Downloads>minishift ip
fe80::215:5dff:fe15:7804
```

Related to: #417

lucastheisen:
I think this may be related to my problem as well... Th only difference is that when i run start it never gets past this:
```
PS C:\Tools\minishift> minishift start
Starting local OpenShift cluster using 'hyperv' hypervisor...
```
Using hyperv manager, i can _Connect..._ to the vm and watch it boot all the way to the root prompt:
```
![image](https://user-images.githubusercontent.com/1952786/27358025-b23c2188-55e2-11e7-85b8-982d44d0f8b0.png)
```
Also, inside of hyperv manager i can see an IPV6 address assigned under the network tab...  But then it just hangs.

gbraad:
@lucastheisen we have since provided a way to select the networking switch https://docs.openshift.org/latest/minishift/getting-started/setting-up-driver-plugin.html#hyper-v hopefully this helps. Do you have Docker for Windows installed?

lucastheisen:
@gbraad, I did set up my network switch in that fashion.  Created one according to the instructions you linked to on the Microsoft page...  The only difference being that I created it as an internal Network with NAT...

gbraad:
@lalatenduM @hferentschik one of the things we can do on startup is to throw an error when the host has two (or no) virtual switches, and of course report which one is used

gbraad:
@lucastheisen The text mentions "set the environment variable HYPERV_VIRTUAL_SWITCH to the name of the **external** virtual switch"

At the moment we do not support NAT or Internal, as they do not provide the necessary DHCP functionality.

lucastheisen:
@gbraad, I did see that, but thought this issue was about getting NAT support, so I hoped that little please was outdated (given that docker for Windows has this feature).  Without NAT, my team is out of luck, as corporate won't assign ips to non-corporate managed images.  If I keep watch on this ticket will it let us know when NAT is supported?

[...]

PR #1139, opened 2017-07-17
[WIP] Startup feedback improvements

gbraad:
Work in progress for #485

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -18,8 +18,12 @@ package cmd
 
 import (
 	"fmt"
+	"net"
 	"path/filepath"
 	"strings"
+
+	"os"
+	"os/exec"
 
 	"github.com/asaskevich/govalidator"
 	units "github.com/docker/go-units"
@@ -42,6 +46,7 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/hostfolder"
 	"github.com/minishift/minishift/pkg/minishift/openshift"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
+	miniutil "github.com/minishift/minishift/pkg/minishift/util"
 	"github.com/minishift/minishift/pkg/util"
 	inputUtils "github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
@@ -49,7 +54,6 @@ import (
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
-	"os"
 )
 
 const (
@@ -124,6 +128,9 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	ensureNotRunning(libMachineClient, constants.MachineName)
 	validateOpenshiftVersion()
+
+	preflightChecksBeforeStartHost()
+
 	setSubscriptionManagerParameters()
 
 	proxyConfig := handleProxies()
@@ -132,8 +139,15 @@ func runStart(cmd *cobra.Command, args []string) {
 	// we need to determine whether this is a restart prior to potentially creating a new VM
 	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
 
+	print("-- Starting local OpenShift cluster")
 	hostVm, ip := startHost(libMachineClient)
 	registerHost(libMachineClient)
+
+	// preflight checks
+	if net.ParseIP(ip).To4() == nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("VM has been assigned an unusable IP address: %s", ip))
+	}
+	preflightChecksAfterStartHost(hostVm.Driver)
 
 	if proxyConfig.IsEnabled() {
 		// once we know the IP, we need to make sure it is not proxied in a proxy environment
@@ -267,7 +281,7 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
 		ShellProxyEnv:    shellProxyEnv,
 	}
 
-	fmt.Printf("Starting local OpenShift cluster using '%s' hypervisor...\n", machineConfig.VMDriver)
+	fmt.Printf(" using '%s' hypervisor ...\n", machineConfig.VMDriver)
 	var hostVm *host.Host
 	start := func() (err error) {
 		hostVm, err = cluster.StartHost(libMachineClient, *machineConfig)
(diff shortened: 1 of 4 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -132,9 +139,16 @@ func runStart(cmd *cobra.Command, args []string) {
 	// we need to determine whether this is a restart prior to potentially creating a new VM
 	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
 
+	print("-- Starting local OpenShift cluster")
 	hostVm, ip := startHost(libMachineClient)
 	registerHost(libMachineClient)
 
+	// preflight checks
+	if net.ParseIP(ip).To4() == nil {
```
hferentschik: What do you mitigate here? In which cases have we seen non IP4 IPs?

cmd/minishift/cmd/start.go
```diff
@@ -281,11 +295,14 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error starting the VM: %v", err))
 	}
 
+	print("-- Checking for IP ... ")
```
hferentschik: Did you not already do an IP check above?

cmd/minishift/cmd/start.go
```diff
@@ -564,3 +581,168 @@ func IsOpenShiftRunning(driver drivers.Driver) bool {
...
+
+ */
+
+/*
+
+startupCheck("Checking for ... ", func() bool, shouldFail)
+
+*/
+
+type preflightCheckRunner func() bool
```
hferentschik: I would think all this goes into some new file/package?

cmd/minishift/cmd/start.go
```diff
@@ -564,3 +581,168 @@ func IsOpenShiftRunning(driver drivers.Driver) bool {
...
+	} else {
+		println("FAIL")
+	}
+}
+
+func preflightChecksBeforeStartHost() {
+	// TODO: to be split-up in individual tests that can be enabled, disabled or ignored
+
+	// to extract
+	vmdriver := viper.GetString(startFlags.VmDriver.Name)
```
hferentschik: I would suggest to pass the driver instead of getting it from Viper. Especially if we are going to extract this code.

cmd/minishift/cmd/start.go
```diff
@@ -564,3 +581,168 @@ func IsOpenShiftRunning(driver drivers.Driver) bool {
...
+			false, "")
+	case "hyperv":
+		succeedsOrFails(checkHypervDriver,
+			"-- Checking of 'hyperv' driver configured ...",
+			true,
+			"   Hyper-V virtual switch not set")
+	}
+}
+
+func checkXhyveDriver() bool {
```

cmd/minishift/cmd/start.go
```diff
@@ -564,3 +581,168 @@ func IsOpenShiftRunning(driver drivers.Driver) bool {
...
+func checkHypervDriver() bool {
+	switchEnv := os.Getenv("HYPERV_VIRTUAL_SWITCH")
+	if switchEnv == "" {
+		return false
+	}
+	return true
+}
+
+func preflightChecksAfterStartHost(driver drivers.Driver) {
+	succeedsOrFailsWithDriver(checkIPConnectivity, driver, "-- Checking host '8.8.8.8' reachable from VM ... ")
```
hferentschik: Not sure whether this is a good test. Why do you assume 8.8.8.8 needs to be reachable. There is really no requirement for this from our side. Who says that 8.8.8.8 is not blocked for some reasons?

cmd/minishift/cmd/start.go
```diff
@@ -564,3 +581,168 @@ func IsOpenShiftRunning(driver drivers.Driver) bool {
...
+	succeedsOrFailsWithDriver(checkDNSConnectivity, driver, "-- Checking host 'www.redhat.com' reachable from VM ... ")
```
hferentschik: redhat.com? Really?

pkg/minikube/cluster/cluster.go
```diff
@@ -54,7 +56,8 @@ var (
 )
 
 const (
-	fileScheme = "file"
+	fileScheme    = "file"
```
hferentschik: So far we tried to stay away from changing things in this packages. It will make things a lot harder when trying to change the model on how we consume Minikube.

pkg/minikube/cluster/cluster.go
```diff
@@ -328,20 +333,33 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 	h.HostOptions.AuthOptions.StorePath = constants.Minipath
 	h.HostOptions.EngineOptions = engineOptions(config)
 
+	println("-- Instance will be configured with ...")
```
hferentschik: We can add this logging in start.go. Then we don't have to touch cluster.go

cmd/minishift/cmd/start.go
gbraad: See comments in issue #485. Any address will do, but google will not work in China, index.docker.io can not be pinged, etc... so, any suggestion can be recorded.

gbraad:
This PR is not ready for review...

cmd/minishift/cmd/start.go
gbraad: As mentioned in #485, values for this test (and the DNS check) are to be configurable for override by config settings. At the moment, `8.8.8.8` seems to be a good test to perform to see if something is remotely reachable. But yes, when proxy is concerned, this might fail. This test is therefore also a WARNING, and not an ERROR.

cmd/minishift/cmd/start.go
gbraad: sure, ... actually, I want to be sure this value is the same as actually used by the `CreateHost` call. So extracting is maybe a good way... since we can pass this into a general preflightCheck function.

cmd/minishift/cmd/start.go
gbraad: This will be moved to a new file... and maybe even pkg. Not sure yet... but as part of tests it lives here, as I need to test on three platforms to make sure I get the 'same' results.

cmd/minishift/cmd/start.go
gbraad: IPv6 is supported on Hyper-V and is not working well #418

cmd/minishift/cmd/start.go
gbraad: Already concluded this when looking at the code here: https://github.com/gbraad/minishift/commit/802ea86e070e194d49deef88bc863e48f80f427d. It seems the original code came TOO late to determine an IP was not assigned. Will correct this...

PR #1140, opened 2017-07-17
Issue #1121 Add support registration bool flag

### LINKED ISSUE #1121: Possible error in delete/registration code
When deleting a CentOS based image using `minishift delete`, I get the following error:

```
$ minishift delete
Current Minishift VM is registered, but unregistration failed. Do you still want to delete the VM [y/N]?: y
Deleting the Minishift VM...
Minishift VM deleted.
```

This seems wrong, as the VM is not `registered`.

praveenkumar:
Fixes #1121; error in registration code

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -544,11 +544,14 @@ func registerHost(libMachineClient *libmachine.Client) {
 		log.Debug("Skipping registration due to enabled --skip-registration flag")
 		return
 	}
-	if err := cluster.Register(libMachineClient); err != nil {
+	supportRegistration, err := cluster.Register(libMachineClient)
+	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error to register VM: %v", err))
-	} else {
+	}
+	if supportRegistration {
 		minishiftConfig.InstanceConfig.IsRegistered = true
 		minishiftConfig.InstanceConfig.Write()
+		return
 	}
 }
 
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -133,15 +133,16 @@ func DeleteHost(api libmachine.API) error {
 }
 
 // Register host VM
-func Register(api libmachine.API) error {
+func Register(api libmachine.API) (bool, error) {
 	host, err := api.Load(constants.MachineName)
 	if err != nil {
-		return err
+		return false, err
 	}
-	if err := registration.RegisterHostVM(host, RegistrationParameters); err != nil {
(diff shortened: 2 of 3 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -544,11 +544,14 @@ func registerHost(libMachineClient *libmachine.Client) {
...
-	if err := cluster.Register(libMachineClient); err != nil {
+	supportRegistration, err := cluster.Register(libMachineClient)
+	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error to register VM: %v", err))
-	} else {
+	}
+	if supportRegistration {
 		minishiftConfig.InstanceConfig.IsRegistered = true
 		minishiftConfig.InstanceConfig.Write()
+		return
```
LalatenduMohanty: The only issue I have with the code is we are passing supportRegistration variable from two level down deep i.e. all the way from RegisterHostVM function. Is there any specific reason we can not move the below code to cluster.go i.e. Register function

```
 		minishiftConfig.InstanceConfig.IsRegistered = true
 		minishiftConfig.InstanceConfig.Write()
```

praveenkumar: @LalatenduMohanty we shouldn't because cluster.go is part of minikube pkg and moving forward we are thinking to consume it as library so moving this code doesn't make sense.

gbraad: I had the same concern, but this is also introducing consistency between the `Register` and `UnRegister` function. I believe we will need to address this is in later code refactors.

hferentschik: > we shouldn't because cluster.go is part of minikube pkg and moving forward we are thinking to consume it as library so moving this code doesn't make sense.

In fact I am wondering what Register and Unregister to in cluster.go at all. There is no driving force here and they could be as easily be placed in the Minishift registration package. That would make a better separation of concerns and also reduce one more pain point thinking of splitting out all Minikube related. 

> Is there any specific reason we can not move the below code to cluster.go i.e. Register function

There is a good reason imo. The commands are in a certain way our orchestrators. The fact that we store whether a machine is registered is really orthogonal to the actual act of registration and un-registrating it. If you move the config handling into the Register methods, you now do to things in one methods. Makes it harder to test and also to re-use.

registration.RegisterHostVM

gbraad:
> we could move Register and Unregister as well

Agree, as this is specific Minishift code and only related to our registration process.
If additional functionality is needed from `cluster.go`, we should introduce hooks for precreate and postcreate

hferentschik:
Rebased and merged, thanks.

Issue #1133, opened 2017-07-14
Function configContains in integration_test.go does not work properly


agajdosi:
Function was meant to cover 4 cases:
1. key has exact value
2. key has not specified value
4. key does not exist
3. key exists with any value

Case number 4 is broken:

```gherkin
  Scenario: Setup-cdk sets RHEL7 iso to be used                                                 # features/test.feature:4
    When executing "minishift config set disk-size 33" succeeds                                 # integration_test.go:454 -> github.com/minishift/minishift/test/integration.executingMinishiftCommandSucceedsOrF
ails
    Then JSON config file "config/config.json" contains key "disk-size" with value "33"         # integration_test.go:309 -> github.com/minishift/minishift/test/integration.configContains
    Then JSON config file "config/config.json" does not contain key "disk-size" with value "66" # integration_test.go:309 -> github.com/minishift/minishift/test/integration.configContains
    Then JSON config file "config/config.json" does not contain key "nonsense"                  # integration_test.go:309 -> github.com/minishift/minishift/test/integration.configContains
    Then JSON config file "config/config.json" contains key "disk-size"                         # integration_test.go:309 -> github.com/minishift/minishift/test/integration.configContains
    For key 'disk-size' config contains unexpected value '33'
```

I would like to use it in downstream for `iso-url` value check as part of CDK-142, maybe would be nice to implement regex also.

hferentschik:
> I would like to use it in downstream for iso-url value check as part of CDK-142, maybe would be nice to implement regex also.

+1

hferentschik:
Merged via pull request #1143

PR #1143, opened 2017-07-17
Issue #1133 Fix and split jsonConfigContains, add regexp support

agajdosi:
Splits configContains into 3 separate functions: jsonConfigContains compares actual and expected key values, jsonConfigHasKey - checks whether key has any value, getConfigValue - gets the actual value.

Adds support for regular expressions in field of expected key values.

### DIFF
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -49,12 +49,12 @@ Feature: Basic
     and its subcommands, changing values stored in "config/config.json".
     Given Minishift has state "Running"
      When executing "minishift config set <property> <value>" succeeds
-     Then JSON config file "config/config.json" contains key "<property>" with value "<value>"
+     Then JSON config file "config/config.json" contains key "<property>" with value matching "<value>"
       And stdout of command "minishift config get <property>" is equal to "<value>"
       And stdout of command "minishift config view --format {{.ConfigKey}}:{{.ConfigValue}}" contains "<property>:<value>"
      When executing "minishift config unset <property>" succeeds
      Then stdout of command "minishift config get <property>" is equal to "<nil>"
-      And JSON config file "config/config.json" does not contain key "<property>"
+      And JSON config file "config/config.json" does not have key "<property>"
 
   Examples: Config values to work with
     | property  | value |
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -39,6 +39,7 @@ import (
 	"github.com/minishift/minishift/test/integration/util"
 	"net/http"
 	"path/filepath"
+	"regexp"
 )
 
 var (
@@ -169,10 +170,10 @@ func FeatureContext(s *godog.Suite) {
 		getRoutingUrlAndVerifyHTTPResponse)
 
 	// steps for verifying config file content
-	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value "(.*)"$`,
-		configContains)
-	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)"(.*)$`,
-		configContains)
+	s.Step(`^JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value matching "(.*)"$`,
+		matchConfigValue)
+	s.Step(`^JSON config file "([^"]*)" (has|does not have) key "(.*)"$`,
+		checkConfigKey)
 
 	s.BeforeSuite(func() {
 		testDir = setUp()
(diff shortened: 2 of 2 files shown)


gbraad:
Add to whitelist

test/integration/integration_test.go
```diff
@@ -217,33 +218,61 @@ func ensureTestDirEmpty() {
...
+func jsonConfigContains(configPath string, condition string, keyPath string, expectedValue string) error {
+	expectedRegexp, err := regexp.Compile(expectedValue)
+	if err != nil {
+		return fmt.Errorf("Expected value must be a valid regular expression statement: ", err)
+	}
+	keyValue, err := getConfigValue(configPath, keyPath)
+	if err != nil {
+		return err
+	}
+	if (condition == "contains") && !expectedRegexp.MatchString(keyValue) {
```
gbraad: whiteline

test/integration/integration_test.go
```diff
@@ -217,33 +218,61 @@ func ensureTestDirEmpty() {
...
-		return fmt.Errorf("For key '%s' config contains unexpected value '%s'", expectedKeyPath, actualValue)
+	return keyValue, nil
+}
+
+func jsonConfigContains(configPath string, condition string, keyPath string, expectedValue string) error {
+	expectedRegexp, err := regexp.Compile(expectedValue)
+	if err != nil {
+		return fmt.Errorf("Expected value must be a valid regular expression statement: ", err)
+	}
+	keyValue, err := getConfigValue(configPath, keyPath)
```
gbraad: whiteline

test/integration/integration_test.go
```diff
@@ -217,33 +218,61 @@ func ensureTestDirEmpty() {
...
 
 	return nil
 }
 
+func jsonConfigHasKey(configPath string, condition string, keyPath string) error {
+	keyValue, err := getConfigValue(configPath, keyPath)
+	if err != nil {
+		return err
+	}
+	if (condition == "has") && (keyValue == "<nil>") {
```
gbraad: whiteline

test/integration/integration_test.go
```diff
@@ -217,33 +218,61 @@ func ensureTestDirEmpty() {
...
+		return fmt.Errorf("Config does not contain any value for key %s", keyPath)
+	} else if (condition == "does not have") && (keyValue != "<nil>") {
+		return fmt.Errorf("Config contains key %s with assigned value: %s", keyPath, keyValue)
+	}
+	return nil
```
gbraad: whiteline

test/integration/features/basic.feature
```diff
@@ -54,7 +54,7 @@ Feature: Basic
       And stdout of command "minishift config view --format {{.ConfigKey}}:{{.ConfigValue}}" contains "<property>:<value>"
      When executing "minishift config unset <property>" succeeds
      Then stdout of command "minishift config get <property>" is equal to "<nil>"
-      And JSON config file "config/config.json" does not contain key "<property>"
+      And JSON config file "config/config.json" does not have key "<property>" with value assigned
```
gbraad: this sentence reads weird...

`does not have a value assigned to key "<property>"`

test/integration/integration_test.go
```diff
@@ -169,10 +170,10 @@ func FeatureContext(s *godog.Suite) {
 		getRoutingUrlAndVerifyHTTPResponse)
 
 	// steps for verifying config file content
-	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value "(.*)"$`,
-		configContains)
-	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)"(.*)$`,
-		configContains)
+	s.Step(`^JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value "(.*)"$`,
+		jsonConfigContains)
+	s.Step(`^JSON config file "([^"]*)" (has|does not have) key "(.*)" with value assigned$`,
```
gbraad: confused what this would have to do:

`has key` or `or does not have key`... the value would have to be irrelevant here...
even `""` is a value...

agajdosi: @gbraad Actually I had it without `with value assigned` and changed it just before the PR, if your suggestion is to remove the `with value assigned` I am +1 and will change it on next update.

Thanks for review, Gerard.

gbraad: It does not really make it clearer, so I am OK with removing that part.

LalatenduMohanty: Any specific reason we want to add `^` i.e.  the sentence should start with  `JSON config file`?

gbraad: @LalatenduMohanty there is an issue with godog, when a line partially matches, and is specified before the other cases, it will take precedence. This way, this can be prevented

hferentschik: > has key or or does not have key... the value would have to be irrelevant here...
even "" is a value...

whether an empty value matters would be implementation and Viper dependent. So there might be and probably is a difference here. However, in this context I think we are talking about whether a given key exists or not (assuming that you mange keys properly via config set|unset. So +1 for removing "with value assigned".

test/integration/integration_test.go
```diff
@@ -169,10 +170,10 @@ func FeatureContext(s *godog.Suite) {
 		getRoutingUrlAndVerifyHTTPResponse)
 
 	// steps for verifying config file content
-	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value "(.*)"$`,
-		configContains)
-	s.Step(`JSON config file "([^"]*)" (contains|does not contain) key "(.*)"(.*)$`,
-		configContains)
+	s.Step(`^JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value "(.*)"$`,
```
hferentschik: "s.Step(`^JSON config file "([^"]*)" (contains|does not contain) key "(.*)" with value matching "(.*)"$`"

might be more explicit and making clear in the feature file that we do regular expression matching.

test/integration/integration_test.go
```diff
@@ -217,28 +218,60 @@ func ensureTestDirEmpty() {
...
-			return errors.New("Unexpected type in JSON config, not supported by testsuite")
+			return "", errors.New("Unexpected type in JSON config, not supported by testsuite")
 		}
 	}
-	if (expectingResult == "contains") && (actualValue != expectedValue) {
-		return fmt.Errorf("For key '%s' config contains unexpected value '%s'", expectedKeyPath, actualValue)
+	return keyValue, nil
+}
+
+func jsonConfigContains(configPath string, condition string, keyPath string, expectedValue string) error {
```
hferentschik: jsonConfigContains == configValueMatches?

jsonConfigContains as name creates the image in my mind that we do a regexp over the whole file. However, the regexp is applied to the the value of a specific key.

test/integration/integration_test.go
```diff
@@ -217,28 +218,60 @@ func ensureTestDirEmpty() {
...
+	if (condition == "contains") && !expectedRegexp.MatchString(keyValue) {
+		return fmt.Errorf("For key '%s' config contains unexpected value '%s'", keyPath, keyValue)
+	} else if (condition == "does not contain") && expectedRegexp.MatchString(keyValue) {
+		return fmt.Errorf("For key '%s' config contains value '%s', which it should not contain", keyPath, keyValue)
+	}
+
+	return nil
+}
+
+func jsonConfigHasKey(configPath string, condition string, keyPath string) error {
```
hferentschik: The name is actually misleading. It implies that it checks whether the config defines a given key. However, the logic of the method depends on _contdition_. The same applies for _ jsonConfigContains_. Maybe more generic function names would be better, _checkConfigKey_ and _matchConfigValue_?

test/integration/integration_test.go
agajdosi: @hferentschik ok, +1, going to change it

test/integration/integration_test.go
agajdosi: @hferentschik good point, changing it now

test/integration/integration_test.go
gbraad: precedence

hferentschik:
Merged, thanks.

agajdosi:
Thanks @hferentschik.

PR #1146, opened 2017-07-19
Issue #1148 Reference the latest OpenShift doc

liangxia:
Reference oc usage and authentication concept to the latest OpenShift doc.

### DIFF
--- a/docs/source/openshift/openshift-client-binary.adoc
+++ b/docs/source/openshift/openshift-client-binary.adoc
@@ -38,7 +38,7 @@ Minishift activates this context automatically, but if you need to switch back t
 $ oc config use-context minishift
 ----
 
-For an introduction to `oc` usage, see the link:https://docs.openshift.com/enterprise/3.2/cli_reference/get_started_cli.html[Get Started with the CLI] topic in the OpenShift documentation.
+For an introduction to `oc` usage, see the link:https://docs.openshift.org/latest/cli_reference/get_started_cli.html[Get Started with the CLI] topic in the OpenShift documentation.
 
 [[log-into-cluster]]
 == Logging Into the Cluster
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
@liangxia Thanks for the PR. I have created an issue for the PR i.e. #1148 as per our suggested work flow for a PR. Please update the commit message as `Issue #1148 Reference the latest OpenShift doc `

gbraad:
Isn't there a fixed version of the documentation? Linking to the latest might result in inconsistencies in supported functionality.

praveenkumar:
@gbraad https://docs.openshift.org/ does have drop-down to select a specific version but again using minishift user can deploy openshift v1.4 to 3.6.rc so I think putting latest might make sense. On second thought we can only use same versioned doc which we support by default (like currently we have by default 1.5.1). @minishift/minishift-dev wdyt?

gbraad:
Actually my thought. We currently point to v1.5.1 as the default, and therefore this should be reflected in your documentation.

This, however, means that each change of the default version would also mean a change in the documentation. This is bound to go wrong... but referring to the latest feels even worse, as this might support newer options we do not have/support in the deployed default version. This is why I suggested a fixed version as preferred option

liangxia:
$ pwd
/opt/github/minishift
$ grep -rE 'docs.openshift.com/' * | wc -l
1
$ grep -rE 'docs.openshift.org/' * | wc -l
27
$ grep -rE 'docs.openshift.org/latest' * | wc -l
27

You can see most part are referencing the latest.

gbraad:
Which we might have to reconsider? @ThatDocsLady, WDYT? Point to latest or
have a inclusion handle the links to a fixed version? If not, we can go
ahead with this...

-- 

   [ Doing Open Source Matters ]

thatdocslady:
@gbraad we usually reference the latest docs as they reflect the most updated information, and since we only publish *our* latest docs, they should be in sync. OSO usually publishes the latest or only one version back for their upstream anyway, so if we reference a specific version then it might break in a few months anyway. I'm happy with this link format :)

hferentschik:
ok to test

hferentschik:
Merged. As @thatdocslady, we usually reference the latest Origin docs.

PR #1154, opened 2017-07-20
Issue #1153 Proxy not set for Docker engine

nimatt:
Using dockerEnv variable when collection machine information upon creation. This is done to fix issue with proxy not being used by the docker engine when no --docker-env parameter is given as well as correctly handling the case where both --docker-env and --http-proxy are used.


hferentschik:
ok to test

praveenkumar:
@nimatt Thanks for your finding this critical bug and pushing PR but we have also need to consider config set options which will not work with it. I spent some time and come up with #1165 which now works for config set and individual options. It would be really great if you can test this PR and let us know what you think.

nimatt:
@praveenkumar absolutely, I missed to check the scenario with config set so feel free to close it. I'll have a look at the PR you created to see that it works as expected for my use case.

praveenkumar:
@nimatt Thanks for confirmation, closing it now.

PR #1162, opened 2017-07-21
Issue #485 Improve startup feedback

### LINKED ISSUE #485: Improve feedback to the user during startup

gbraad:
This fixes #485 

Several checks are performed to improve the feedback to the user. Such as basic test if the driver exists, if storage disk is mounted, warn when issues occur with connectivity, etc...

However, one of the major issues remains; checking if the host is reachable before provisioning. This happens lower down the stack (docker-machine) and therefore we are at the moment unable to detect issues which are caused by VPN usage.

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -52,54 +52,78 @@ const (
 
 var (
 	// minishift
-	ISOUrl           = createFlag("iso-url", SetString, []setFn{IsValidUrl}, []setFn{RequiresRestartMsg}, true)
-	CPUs             = createFlag("cpus", SetInt, []setFn{IsPositive}, []setFn{RequiresRestartMsg}, true)
-	Memory           = createFlag("memory", SetInt, []setFn{IsPositive}, []setFn{RequiresRestartMsg}, true)
-	DiskSize         = createFlag("disk-size", SetString, []setFn{IsValidDiskSize}, []setFn{RequiresRestartMsg}, true)
-	VmDriver         = createFlag("vm-driver", SetString, []setFn{IsValidDriver}, []setFn{RequiresRestartMsg}, true)
-	OpenshiftVersion = createFlag("openshift-version", SetString, nil, nil, true)
-	HostOnlyCIDR     = createFlag("host-only-cidr", SetString, []setFn{IsValidCIDR}, nil, true)
-	DockerEnv        = createFlag("docker-env", SetSlice, nil, nil, true)
-	DockerEngineOpt  = createFlag("docker-opt", SetSlice, nil, nil, true)
-	InsecureRegistry = createFlag("insecure-registry", SetSlice, nil, nil, true)
-	RegistryMirror   = createFlag("registry-mirror", SetSlice, nil, nil, true)
-	AddonEnv         = createFlag("addon-env", SetSlice, nil, nil, true)
+	ISOUrl           = createConfigSetting("iso-url", SetString, []setFn{IsValidUrl}, []setFn{RequiresRestartMsg}, true)
+	CPUs             = createConfigSetting("cpus", SetInt, []setFn{IsPositive}, []setFn{RequiresRestartMsg}, true)
+	Memory           = createConfigSetting("memory", SetInt, []setFn{IsPositive}, []setFn{RequiresRestartMsg}, true)
+	DiskSize         = createConfigSetting("disk-size", SetString, []setFn{IsValidDiskSize}, []setFn{RequiresRestartMsg}, true)
+	VmDriver         = createConfigSetting("vm-driver", SetString, []setFn{IsValidDriver}, []setFn{RequiresRestartMsg}, true)
+	OpenshiftVersion = createConfigSetting("openshift-version", SetString, nil, nil, true)
+	HostOnlyCIDR     = createConfigSetting("host-only-cidr", SetString, []setFn{IsValidCIDR}, nil, true)
+	DockerEnv        = createConfigSetting("docker-env", SetSlice, nil, nil, true)
+	DockerEngineOpt  = createConfigSetting("docker-opt", SetSlice, nil, nil, true)
+	InsecureRegistry = createConfigSetting("insecure-registry", SetSlice, nil, nil, true)
+	RegistryMirror   = createConfigSetting("registry-mirror", SetSlice, nil, nil, true)
+	AddonEnv         = createConfigSetting("addon-env", SetSlice, nil, nil, true)
 
 	// cluster up
-	SkipRegistryCheck = createFlag("skip-registry-check", SetBool, nil, nil, true)
-	PublicHostname    = createFlag("public-hostname", SetString, nil, nil, true)
-	RoutingSuffix     = createFlag("routing-suffix", SetString, nil, nil, true)
-	HostConfigDir     = createFlag("host-config-dir", SetString, []setFn{IsValidPath}, nil, true)
-	HostVolumeDir     = createFlag("host-volumes-dir", SetString, []setFn{IsValidPath}, nil, true)
-	HostDataDir       = createFlag("host-data-dir", SetString, []setFn{IsValidPath}, nil, true)
-	HostPvDir         = createFlag("host-pv-dir", SetString, []setFn{IsValidPath}, nil, true)
-	ServerLogLevel    = createFlag("server-loglevel", SetInt, []setFn{IsPositive}, nil, true)
-	OpenshiftEnv      = createFlag("openshift-env", nil, nil, nil, false)
-	Metrics           = createFlag("metrics", SetBool, nil, nil, true)
-	Logging           = createFlag("logging", SetBool, nil, nil, true)
+	SkipRegistryCheck = createConfigSetting("skip-registry-check", SetBool, nil, nil, true)
+	PublicHostname    = createConfigSetting("public-hostname", SetString, nil, nil, true)
+	RoutingSuffix     = createConfigSetting("routing-suffix", SetString, nil, nil, true)
+	HostConfigDir     = createConfigSetting("host-config-dir", SetString, []setFn{IsValidPath}, nil, true)
+	HostVolumeDir     = createConfigSetting("host-volumes-dir", SetString, []setFn{IsValidPath}, nil, true)
+	HostDataDir       = createConfigSetting("host-data-dir", SetString, []setFn{IsValidPath}, nil, true)
+	HostPvDir         = createConfigSetting("host-pv-dir", SetString, []setFn{IsValidPath}, nil, true)
+	ServerLogLevel    = createConfigSetting("server-loglevel", SetInt, []setFn{IsPositive}, nil, true)
+	OpenshiftEnv      = createConfigSetting("openshift-env", nil, nil, nil, false)
+	Metrics           = createConfigSetting("metrics", SetBool, nil, nil, true)
+	Logging           = createConfigSetting("logging", SetBool, nil, nil, true)
 
 	// Setting proxy
-	NoProxyList = createFlag("no-proxy", SetString, nil, nil, true)
-	HttpProxy   = createFlag("http-proxy", SetString, []setFn{IsValidProxy}, nil, true)
-	HttpsProxy  = createFlag("https-proxy", SetString, []setFn{IsValidProxy}, nil, true)
+	NoProxyList = createConfigSetting("no-proxy", SetString, nil, nil, true)
+	HttpProxy   = createConfigSetting("http-proxy", SetString, []setFn{IsValidProxy}, nil, true)
+	HttpsProxy  = createConfigSetting("https-proxy", SetString, []setFn{IsValidProxy}, nil, true)
 
 	// Subscription Manager
-	Username         = createFlag("username", SetString, nil, nil, true)
-	Password         = createFlag("password", SetString, nil, nil, true)
-	SkipRegistration = createFlag("skip-registration", SetBool, nil, nil, true)
+	Username         = createConfigSetting("username", SetString, nil, nil, true)
+	Password         = createConfigSetting("password", SetString, nil, nil, true)
+	SkipRegistration = createConfigSetting("skip-registration", SetBool, nil, nil, true)
 
 	// Global flags
-	LogDir             = createFlag("log_dir", SetString, []setFn{IsValidPath}, nil, true)
-	ShowLibmachineLogs = createFlag("show-libmachine-logs", SetBool, nil, nil, true)
+	LogDir             = createConfigSetting("log_dir", SetString, []setFn{IsValidPath}, nil, true)
+	ShowLibmachineLogs = createConfigSetting("show-libmachine-logs", SetBool, nil, nil, true)
 
 	// Host Folders
-	HostFoldersMountPath = createFlag("hostfolders-mountpath", SetString, nil, nil, true)
-	HostFoldersAutoMount = createFlag("hostfolders-automount", SetBool, nil, nil, true)
+	HostFoldersMountPath = createConfigSetting("hostfolders-mountpath", SetString, nil, nil, true)
+	HostFoldersAutoMount = createConfigSetting("hostfolders-automount", SetBool, nil, nil, true)
 
-	ImageCaching = createFlag("image-caching", SetBool, nil, nil, true)
+	ImageCaching = createConfigSetting("image-caching", SetBool, nil, nil, true)
+
+	// Preflight checks (before start)
+	SkipCheckKVMDriver    = createConfigSetting("skip-check-kvm-driver", SetBool, nil, nil, true)
+	WarnCheckKVMDriver    = createConfigSetting("warn-check-kvm-driver", SetBool, nil, nil, true)
+	SkipCheckXHyveDriver  = createConfigSetting("skip-check-xhyve-driver", SetBool, nil, nil, true)
+	WarnCheckXHyveDriver  = createConfigSetting("warn-check-xhyve-driver", SetBool, nil, nil, true)
+	SkipCheckHyperVDriver = createConfigSetting("skip-check-hyperv-driver", SetBool, nil, nil, true)
+	WarnCheckHyperVDriver = createConfigSetting("warn-check-hyperv-driver", SetBool, nil, nil, true)
+	// Preflight checks (after start)
+	SkipInstanceIP        = createConfigSetting("skip-check-instance-ip", SetBool, nil, nil, true)
+	WarnInstanceIP        = createConfigSetting("warn-check-instance-ip", SetBool, nil, nil, true)
+	SkipCheckNetworkHost  = createConfigSetting("skip-check-network-host", SetBool, nil, nil, true)
+	WarnCheckNetworkHost  = createConfigSetting("warn-check-network-host", SetBool, nil, nil, true)
+	SkipCheckNetworkPing  = createConfigSetting("skip-check-network-ping", SetBool, nil, nil, true)
+	WarnCheckNetworkPing  = createConfigSetting("warn-check-network-ping", SetBool, nil, nil, true)
+	SkipCheckNetworkHTTP  = createConfigSetting("skip-check-network-http", SetBool, nil, nil, true)
+	WarnCheckNetworkHTTP  = createConfigSetting("warn-check-network-http", SetBool, nil, nil, true)
+	SkipCheckStorageMount = createConfigSetting("skip-check-storage-mount", SetBool, nil, nil, true)
+	WarnCheckStorageMount = createConfigSetting("warn-check-storage-mount", SetBool, nil, nil, true)
+	SkipCheckStorageUsage = createConfigSetting("skip-check-storage-usage", SetBool, nil, nil, true)
+	WarnCheckStorageUsage = createConfigSetting("warn-check-storage-usage", SetBool, nil, nil, true)
+	// Preflight values
+	CheckNetworkHttpHost = createConfigSetting("check-network-http-host", SetString, nil, nil, true)
+	CheckNetworkPingHost = createConfigSetting("check-network-ping-host", SetString, nil, nil, true)
 )
 
-func createFlag(name string, set func(MinishiftConfig, string, string) error, validations []setFn, callbacks []setFn, isApply bool) *Setting {
+func createConfigSetting(name string, set func(MinishiftConfig, string, string) error, validations []setFn, callbacks []setFn, isApply bool) *Setting {
 	flag := Setting{
 		Name:        name,
 		set:         set,
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -22,8 +22,10 @@ import (
 	"strings"
 
 	"os"
+	"time"
 
 	"github.com/asaskevich/govalidator"
+	"github.com/briandowns/spinner"
 	units "github.com/docker/go-units"
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/drivers"
@@ -44,6 +46,7 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/hostfolder"
 	"github.com/minishift/minishift/pkg/minishift/openshift"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
+
 	"github.com/minishift/minishift/pkg/util"
 	inputUtils "github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
@@ -62,6 +65,8 @@ const (
 	defaultInsecureRegistry = "172.30.0.0/16"
 
 	unsupportedIsoUrlFormat = "Unsupported value for iso-url. It can be an URL, file URI or one of the following short names: [b2d centos]."
+
+	SPINNER_SPEED = 200 // In milliseconds
 )
 
 var (
@@ -126,6 +131,10 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	ensureNotRunning(libMachineClient, constants.MachineName)
 	validateOpenshiftVersion()
+
+	// preflight check (before start)
+	preflightChecksBeforeStartingHost()
+
 	setSubscriptionManagerParameters()
 
 	proxyConfig := handleProxies()
@@ -134,8 +143,13 @@ func runStart(cmd *cobra.Command, args []string) {
 	// we need to determine whether this is a restart prior to potentially creating a new VM
 	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
 
-	hostVm, ip := startHost(libMachineClient)
+	fmt.Printf("-- Starting local OpenShift cluster")
+	hostVm := startHost(libMachineClient)
 	registerHost(libMachineClient)
+
+	// preflight checks (after start)
+	preflightChecksAfterStartingHost(hostVm.Driver)
+	ip, _ := hostVm.Driver.GetIP()
 
 	if proxyConfig.IsEnabled() {
 		// once we know the IP, we need to make sure it is not proxied in a proxy environment
@@ -173,6 +187,8 @@ func runStart(cmd *cobra.Command, args []string) {
 	}
 
 	clusterUpParams := determineClusterUpParameters(clusterUpConfig)
+	fmt.Println("-- OpenShift cluster will be configured with ...")
+	fmt.Println("   Version:", requestedOpenShiftVersion)
 	err = clusterup.ClusterUp(clusterUpConfig, clusterUpParams, &util.RealRunner{})
 
 	if err != nil {
@@ -259,7 +275,10 @@ func determineInsecureRegistry(key string) []string {
 	return append(s, defaultInsecureRegistry)
 }
 
-func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
+func startHost(libMachineClient *libmachine.Client) *host.Host {
+	spinnerView := spinner.New(spinner.CharSets[9], SPINNER_SPEED*time.Millisecond)
+
+	// Configuration used for creation/setup of the Virtual Machine
 	machineConfig := &cluster.MachineConfig{
 		MinikubeISO:      determineIsoUrl(viper.GetString(startFlags.ISOUrl.Name)),
 		Memory:           viper.GetInt(startFlags.Memory.Name),
@@ -274,11 +293,25 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
 		ShellProxyEnv:    shellProxyEnv,
 	}
 
-	fmt.Printf("Starting local OpenShift cluster using '%s' hypervisor...\n", machineConfig.VMDriver)
+	fmt.Printf(" using '%s' hypervisor ...\n", machineConfig.VMDriver)
 	var hostVm *host.Host
+
+	// configuration with these settings only happen on create
+	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
+	if !isRestart {
+		fmt.Println("-- Minishift VM will be configured with ...")
+		fmt.Println("   Memory:   ", units.HumanSize(float64((machineConfig.Memory/units.KiB)*units.GB)))
+		fmt.Println("   vCPUs :   ", machineConfig.CPUs)
+		fmt.Println("   Disk size:", units.HumanSize(float64(machineConfig.DiskSize*units.MB)))
+		// Should handle cached images
+		//fmt.Println("   Boot ISO:  ", machineConfig.MinikubeISO)
+	}
+	fmt.Printf("-- Starting Minishift VM ... ")
+	spinnerView.Start()
 	start := func() (err error) {
 		hostVm, err = cluster.StartHost(libMachineClient, *machineConfig)
 		if err != nil {
+			fmt.Printf("FAIL ")
 			glog.Errorf("Error starting the VM: %s. Retrying.\n", err)
 		}
 		return err
@@ -287,13 +320,10 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error starting the VM: %v", err))
 	}
+	spinnerView.Stop()
+	fmt.Println("OK")
 
-	ip, err := hostVm.Driver.GetIP()
-	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Error determining host ip: %v ", err))
-	}
-
-	return hostVm, ip
+	return hostVm
 }
 
 func autoMountHostFolders(driver drivers.Driver) {
--- a/cmd/minishift/cmd/start_preflight.go
+++ b/cmd/minishift/cmd/start_preflight.go
@@ -1,0 +1,331 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"fmt"
+	"net"
+	"os"
+	"os/exec"
+	"strconv"
+	"strings"
+
+	"github.com/docker/machine/libmachine/drivers"
+	startFlags "github.com/minishift/minishift/cmd/minishift/cmd/config"
+	miniutil "github.com/minishift/minishift/pkg/minishift/util"
+
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/viper"
+)
+
+const (
+	StorageDisk = "/mnt/sda1"
+)
+
+// preflightChecksAfterStartingHost is executed before the startHost function.
+func preflightChecksBeforeStartingHost() {
+	switch viper.GetString(startFlags.VmDriver.Name) {
+	case "xhyve":
+		preflightCheckSucceedsOrFails(
+			startFlags.SkipCheckXHyveDriver.Name,
+			checkXhyveDriver,
+			"Checking if xhyve driver is installed",
+			false, startFlags.WarnCheckXHyveDriver.Name,
+			"See the 'Setting Up the Driver Plug-in' topic for more information")
+	case "kvm":
+		preflightCheckSucceedsOrFails(
+			startFlags.SkipCheckKVMDriver.Name,
+			checkKvmDriver,
+			"Checking if KVM driver is installed",
+			false, startFlags.WarnCheckXHyveDriver.Name,
+			"See the 'Setting Up the Driver Plug-in' topic for more information")
+	case "hyperv":
+		preflightCheckSucceedsOrFails(
+			startFlags.SkipCheckHyperVDriver.Name,
+			checkHypervDriver,
+			"Checking if Hyper-V driver is configured",
+			false, startFlags.WarnCheckHyperVDriver.Name,
+			"Hyper-V virtual switch is not set")
+	}
+}
+
+// preflightChecksAfterStartingHost is executed after the startHost function.
+func preflightChecksAfterStartingHost(driver drivers.Driver) {
+	preflightCheckSucceedsOrFailsWithDriver(
+		startFlags.SkipInstanceIP.Name,
+		checkInstanceIP, driver,
+		"Checking for IP address",
+		false, startFlags.WarnInstanceIP.Name,
+		"Error determining IP address")
+	/*
+		// This happens too late in the preflight, as provisioning needs an IP already
+			preflightCheckSucceedsOrFailsWithDriver(
+				startFlags.SkipCheckNetworkHost.Name,
+				checkVMConnectivity, driver,
+				"Checking if VM is reachable from host",
+				startFlags.WarnCheckNetworkHost.Name,
+				"Please check our troubleshooting guide")
+	*/
+	preflightCheckSucceedsOrFailsWithDriver(
+		startFlags.SkipCheckNetworkPing.Name,
+		checkIPConnectivity, driver,
+		"Checking if external host is reachable from the Minishift VM",
+		true, startFlags.WarnCheckNetworkPing.Name,
+		"VM is unable to ping external host")
+	preflightCheckSucceedsOrFailsWithDriver(
+		startFlags.SkipCheckNetworkHTTP.Name,
+		checkHttpConnectivity, driver,
+		"Checking HTTP connectivity from the VM",
+		true, startFlags.WarnCheckNetworkHTTP.Name,
+		"VM cannot connect to external URL with HTTP")
+	preflightCheckSucceedsOrFailsWithDriver(
+		startFlags.SkipCheckStorageMount.Name,
+		checkStorageMounted, driver,
+		"Checking if persistent storage volume is mounted",
+		false, startFlags.WarnCheckStorageMount.Name,
+		"Persistent volume storage is not mounted")
+	preflightCheckSucceedsOrFailsWithDriver(
+		startFlags.SkipCheckStorageUsage.Name,
+		checkStorageUsage, driver,
+		"Checking available disk space",
+		false, startFlags.WarnCheckStorageUsage.Name,
+		"Insufficient disk space on the persistent storage volume")
+}
+
+// preflightCheckFunc returns true when check passed
+type preflightCheckFunc func() bool
+
+// preflightCheckFunc used driver to interact with the VM instance and returns
+// true when check passed
+type preflightCheckWithDriverFunc func(driver drivers.Driver) bool
+
+// preflightCheckSucceedsOrFails executes a pre-flight test function and prints
+// the returned status in a standardized way. If the test fails and returns a
+// false, the application will exit with errorMessage to describe what the
+// cause is. It takes configNameOverrideIfSkipped to allow skipping the test.
+// While treatAsWarning and configNameOverrideIfWarning can be used to make the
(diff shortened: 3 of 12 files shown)


gbraad:
If comments about the used addresses, like `8.8.8.8` and `minishift.io`, please also consider good alternatives.

gbraad:
If comments about the used addresses, like `8.8.8.8` and `minishift.io`, please also consider good alternatives.

gbraad:
Seems integration fails... and something happened with rebasing?!?

gbraad:
Command exits with exit `0`, but it has the output in `stderr` ?

```
16:34:55     Command 'start' did not execute successfully. cmdExit: 0, cmdErr: -- Checking if 'kvm' driver installed ... OK
16:34:55 -- Starting local OpenShift cluster-- Instance will be configured with ...
16:34:55    Memory:    2.048 GB
16:34:55    vCPUs :    2
16:34:55    Disksize:  20 GB
16:34:55 -- Starting Minishift VM ... OK
16:34:55 -- Checking for IP ... OK
16:34:55 -- Checking host reachable from VM ... 
16:34:55    8.8.8.8 ... OK
16:34:55 -- Checking host reachable from VM ... 
16:34:55    minishift.io ... OK
16:34:55 -- Checking storage disk mounted ... OK
16:34:55 -- Checking storage disk usage ... 0% OK
```

gbraad:
In `start.go`, when using:

```go
	print("-- Starting Minishift VM ... ")
```

the messages were logged as `stderr`, while using:

```go
	fmt.Printf("-- Starting Minishift VM ... ")
```

seems to be OK. 😕

gbraad:
@hferentschik problem solved... removed WIP status

hferentschik:
@gbraad cool. I'll look into it ...

cmd/minishift/cmd/config/config.go
```diff
@@ -97,6 +97,30 @@ var (
 	HostFoldersAutoMount = createFlag("hostfolders-automount", SetBool, nil, nil, true)
 
 	ImageCaching = createFlag("image-caching", SetBool, nil, nil, true)
+
+	// Preflight checks (before start)
+	SkipCheckKVMDriver    = createFlag("skip-check-kvm-driver", SetBool, nil, nil, true)
```
hferentschik: so each check is configurable?

cmd/minishift/cmd/config/config.go
```diff
@@ -97,6 +97,30 @@ var (
...
+	WarnCheckKVMDriver    = createFlag("warn-check-kvm-driver", SetBool, nil, nil, true)
```
hferentschik: We really should rename _createFlag_ to _createConfigSetting_. The name implies that it creates a CLI flag somewhere, but this is not the case anymore. This is a list of configurable settings some of which are exposed as flags.

hferentschik: I am missing some doc updates which covers the pre-flight checks and the fact they are configurable and how to do it.

cmd/minishift/cmd/start.go
```diff
@@ -62,6 +65,8 @@ const (
 	defaultInsecureRegistry = "172.30.0.0/16"
 
 	unsupportedIsoUrlFormat = "Unsupported value for iso-url. It can be an URL, file URI or one of the following short names: [b2d centos]."
+
+	SPINNER_SPEED = 200 // In milliseconds
```
hferentschik: :+1

cmd/minishift/cmd/start.go
```diff
@@ -126,6 +131,11 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	ensureNotRunning(libMachineClient, constants.MachineName)
 	validateOpenshiftVersion()
+
+	// preflight check (before start)
+	vmDriver := viper.GetString(startFlags.VmDriver.Name)
+	preflightChecksBeforeStartHost(vmDriver)
```
hferentschik: _preflightChecksPreStartHost_? _preflightChecksBeforeStartingHost_?

cmd/minishift/cmd/start.go
```diff
@@ -255,13 +270,15 @@ func determineInsecureRegistry(key string) []string {
 	return append(s, defaultInsecureRegistry)
 }
 
-func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
+func startHost(libMachineClient *libmachine.Client, vmDriver string) *host.Host {
```
hferentschik: Not quite sure whether I'd add the driver parameter here. Why not retrieve it again via Viper. Just wondering?

cmd/minishift/cmd/start.go
```diff
@@ -270,11 +287,23 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
 		ShellProxyEnv:    shellProxyEnv,
 	}
 
-	fmt.Printf("Starting local OpenShift cluster using '%s' hypervisor...\n", machineConfig.VMDriver)
+	fmt.Printf(" using '%s' hypervisor ...\n", machineConfig.VMDriver)
 	var hostVm *host.Host
+
+	// configuration with these settings only happen on create
+	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
+	if !isRestart {
```
hferentschik: 👍 This was long time overdue

cmd/minishift/cmd/start.go
```diff
@@ -270,11 +287,23 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
...
+		fmt.Println("-- Instance will be configured with ...")
+		fmt.Println("   Memory:   ", units.HumanSize(float64(machineConfig.Memory*units.MB)))
+		fmt.Println("   vCPUs :   ", machineConfig.CPUs)
```
hferentschik: vCPU? I guess you mean _virtual_, but not sure this is common usage. Why not just CPUs

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,278 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
```
hferentschik: 2017 (P.S. I hate it as well)

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,278 @@
...
+
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/viper"
+)
+
+const (
+	StorageDisk = "/mnt/sda1"
+)
+
+type preflightCheckRunner func() bool
```
hferentschik: Why 'Runner' instead of 'Func'? I think the latter would make things clearer with not seeing the definition of the type.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,278 @@
...
+type preflightCheckWithDriverRunner func(driver drivers.Driver) bool
+
+func preflightCheckSucceedsOrFails(configNameOverrideIfSkipped string, execute preflightCheckRunner, message string, treatAsWarning bool, configNameOverrideIfWarning string, errorMessage string) {
```
hferentschik: some Godocs would be nice here to understand the parameters. In particular I am wondering about _ configNameOverrideIfSkipped_ and _configNameOverrideIfWarning_ which are not directly clear to me.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,278 @@
...
+			startFlags.SkipCheckHyperVDriver.Name,
+			checkHypervDriver,
+			"Checking if 'hyperv' driver configured",
+			false, startFlags.WarnCheckHyperVDriver.Name,
+			"Hyper-V virtual switch not set")
+	}
+}
+
+func checkXhyveDriver() bool {
+	_, err := exec.LookPath("docker-machine-driver-xhyve")
```
hferentschik: Here we would have to check the _sid_ flag as well. Just checking whether the executable exists might generate wrong checks if the binary is there and executable, but the _sid_ is not set.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,278 @@
...
+	if err != nil {
+		return false
+	}
+	return true
+}
+
+func checkKvmDriver() bool {
+	_, err := exec.LookPath("docker-machine-driver-kvm")
```
hferentschik: not sure whether the _sid_ comment applies here as well. I don't think so, but I am not sure.

cmd/minishift/cmd/start_preflight.go
hferentschik: I think it would be good for testing purposes as well as eventually maybe directing these kind of output to a log file to pass a Writer to these function. Per default it can be os.Stdout, but the nice thing is that it can also be something different. We are applying this approach already at some of the other new functionality (groups).

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,278 @@
...
+	ipToPing := viper.GetString(startFlags.CheckNetworkIPHost.Name)
+	if ipToPing == "" {
+		ipToPing = "8.8.8.8"
+	}
+
+	fmt.Printf("\n   %s ... ", ipToPing)
+	return miniutil.IsIPReachable(driver, ipToPing, false)
+}
+
+func checkDNSConnectivity(driver drivers.Driver) bool {
```
hferentschik: not sure about this. I get the _ checkIPConnectivity_, but _checkDNSConnectivity_ is at best an indirect test. To deduct from the ability to ping _minishift.io_ that DNS works is imo a bit dubious. I would be happy with just an IP connectivity test or "external network connectivity test"

pkg/minikube/cluster/cluster.go
```diff
@@ -386,9 +386,12 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 	}
 
 	if config.ShellProxyEnv != "" {
+		print("-- Setting proxy information ... ")
```
hferentschik: here is a print left. Given our latest experiences and for consistency sake, I recommend to use fmt

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -111,8 +111,12 @@ func (provisioner *MinishiftProvisioner) Provision(swarmOptions swarm.Options, a
 	}
 	provisioner.EngineOptions.StorageDriver = storageDriver
 
+	print("-- Setting hostname ... ")
```
hferentschik: fmt?

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -111,8 +111,12 @@ func (provisioner *MinishiftProvisioner) Provision(swarmOptions swarm.Options, a
...
 	if err := provisioner.SetHostname(provisioner.Driver.GetMachineName()); err != nil {
+		println("FAIL")
```
hferentschik: fmt?

cmd/minishift/cmd/start_preflight.go
gbraad: We are unable to do this with the standard libraries... you can't check for sticky bit, owner, etc... and I have also tried an alternative library, but it already costed too much time so I moved on.

cmd/minishift/cmd/start_preflight.go
gbraad: It needs to do a resolve first, and since the machine has been restarted any cache, is likely to be empty. I likely want can change it to a `curl` check, as this can probably also deal with proxy.

pkg/minikube/cluster/cluster.go
gbraad: It will even result in a writing to `stderr`; missed during the integration fix.

pkg/minishift/provisioner/minishift_provisioner.go
gbraad: It will even result in a writing to `stderr`; missed during the integration fix.

pkg/minishift/provisioner/minishift_provisioner.go
gbraad: It will even result in a writing to `stderr`; missed during the integration fix.

cmd/minishift/cmd/start_preflight.go
gbraad: yeps...

actually, some of the checks, like disk usage, can be useful in general from a command.

cmd/minishift/cmd/start.go
gbraad: was previously suggested to remove the Viper call and pass it to the function...

actually, it would be nicer if the whole config struct gets extracted, as some of the information in this can be used for the precheck or start check improvements.

cmd/minishift/cmd/config/config.go
gbraad: "skippable" and set to "treat as a warning" instead of "exit on error"

cmd/minishift/cmd/config/config.go
gbraad: I am wondering how much we need to describe this...

gbraad: config should also take a default option... this is a shortcoming I encountered.

hferentschik:
+1, how the output aligns with 'oc cluster up' :-)

cmd/minishift/cmd/start.go
gbraad: Quite a general term: https://www.google.com/search?q=vcpu as it does not imply affinity.

cmd/minishift/cmd/start_preflight.go
gbraad: used `start.go` as boilerplate. should have used something else...

changed now

.travis.yml
```diff
@@ -4,8 +4,8 @@ go:
 - 1.7.1
 
 install:
-- mkdir /home/travis/gopath/bin
-- mkdir /home/travis/gopath/pkg
+- mkdir -p /home/travis/gopath/bin
```
hferentschik: I believe this can be dropped now after merging the changes for #1172

cmd/minishift/cmd/config/config.go
hferentschik: > I am wondering how much we need to describe this...

If it is configurable, it need to be described, at least in a general term backed up with an example.

cmd/minishift/cmd/start.go
hferentschik: > was previously suggested to remove the Viper call and pass it to the function...

Really. You are still in the cmd structure. In there I see no problem using Viper. What we have as a goal is to not let Viper/Cobra "leak" into anything under _pkg_. Within the _cmd_ I generally see no issues to access Viper.

> actually, it would be nicer if the whole config struct gets extracted,

Not sure what you mean.

cmd/minishift/cmd/start.go

cmd/minishift/cmd/start_preflight.go
hferentschik: >  I have also tried an alternative library

Which one? It did not work? I feel this really falls a bit short here, especially since it might make things even worse for the user in the case the driver exists and our check "passes", but the boot still fails.

I see two options.

* Add some additional message informing the user that the binary is there, but we are unable to verify the suid. If problems occur check the suid flag. (exact wording needs to be worked out)
* Use os.Exec to execute the driver, then use the returned Process to get the uid of the process and ensure it is '0'. Executing the binary might be a good thing regardless. since it also eliminates issues of the binary being corrupt. 

> but it already cost too much time so I moved on

I think we should get this right, especially since we put so much time into it.

cmd/minishift/cmd/start.go
```diff
@@ -270,11 +287,24 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
...
+		fmt.Println("   Disksize: ", units.HumanSize(float64(machineConfig.DiskSize*units.MB)))
+		fmt.Println("   Boot ISO: ", machineConfig.MinikubeISO)
+	}
+	fmt.Printf("-- Starting Minishift VM ... ")
+	spinnerView.Start()
 	start := func() (err error) {
 		hostVm, err = cluster.StartHost(libMachineClient, *machineConfig)
 		if err != nil {
+			print("FAIL ")
```
gbraad: that should be fmt.Printf(). likely missed during the change to fmt.

cmd/minishift/cmd/start_preflight.go
gbraad: I used
```
info,_ := os.Stat(file)
mode := info.Mode()
```
and then used: https://github.com/phayes/permbits, but had issues with the sticky bit.

/cc: @praveenkumar do you have an idea?

praveenkumar: @gbraad @hferentschik I think below example make sure to check if sticky bit set or not.

```golang
package main

import (
	"fmt"
	"os"
)

func main() {
	info, _ := os.Stat("/tmp/testdir")
	mode := info.Mode()
	if mode&os.ModeSticky != 0 {
		fmt.Println("Sticky bit set")
	}
}
```

gbraad: that is the same as I did, but failed on macOS + Go 1.7.5

cmd/minishift/cmd/start_preflight.go
gbraad: libvirt needs group membership

cmd/minishift/cmd/start_preflight.go
praveenkumar: @gbraad Interesting for me it worked on Mac also, I am not sure if because of version of MacOS I am using.

```
➜  /tmp ls -l | grep testdir 
drwxr-xr-t  2 prkumar  wheel       68 Jul 27 18:09 testdir

➜  test sw_vers 
ProductName:	Mac OS X
ProductVersion:	10.12.5
BuildVersion:	16F73

➜  test go run test.go
Sticky bit set
```

gbraad: Ah, see... and this is not like our test:

```
func checkXhyveDriver() bool {
	path, err := exec.LookPath("docker-machine-driver-xhyve")

	if err != nil {
		return false
	}
	
	fmt.Printf("   Driver is available at %s\n", path)
	fi, _ := os.Stat(path)
	if fi.Mode()&os.ModeSticky == 0 {
		return false
	}
	
	return true
}
```

```
$ ls -al /usr/local/bin/docker-machine-driver-xhyve
-rwsr-sr-x  1 root wheel  7912720  Jun 28 19:28 /usr/local/bin/docker-machine-driver-xhyve
```

For me it fails... as the mode value is 12583405 and when masking with the sticky bit I always get 0.

gbraad: The suggestion of the UID I also tried, but Golang itself only returns the process ID for a process
```
type Process struct {
        Pid int
        // contains filtered or unexported fields
}
```
I haven't found a way to get the user id. So likely a syscall is needed... Other libraries didn't help much: https://github.com/mitchellh/go-ps/blob/master/process_darwin.go

.travis.yml
gbraad: will do during a later rebase

cmd/minishift/cmd/start_preflight.go
gbraad: I guess you meant to check for the SUID (`chmod u+s`). But this does not seem to be conclusive either: @LalatenduMohanty has installed from brew and has no SetUID bit...

```
lrwxr-xr-x  1 lmohanty  admin  75 Dec 12  2016 /usr/local/bin/docker-machine-driver-xhyve -> ../Cellar/docker-machine-driver-xhyve/0.3.1/bin/docker-machine-driver-xhyve
```

gbraad:
retest this please

gbraad:
retest this please

gbraad:
@LalatenduMohanty are you able to test this? Especially as you have installed the driver plugin from `brew`.

gbraad:
@thatdocslady can you have a look at what I added to the `troubleshooting.adoc` and the comments for `quickstart.adoc`?

praveenkumar:
@gbraad below is the output I am getting from kvm.
```
$ ./minishift start
-- Checking if 'kvm' driver installed ... OK ==> Don't we tell user which binary we are using like we are doing for xhyve and setuid bit?
-- Instance will be configured with ...
   Memory:    2 GB
   vCPUs :    2
   Disksize:  20 GB
   Boot ISO:  https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso
-- Checking for IP ... OK
-- Checking host reachable from VM ... 
   Pinging 8.8.8.8 ... OK
-- Checking HTTP connectivity from VM ... 
-- Checking storage disk mounted ... OK
-- Checking storage disk usage ... 0% OK
   host-config-dir ... OK
   host-data-dir ... OK
   host-pv-dir ... OK
   host-volumes-dir ... OK
   routing-suffix ... OK
[...]
```

gbraad:
I can print the path, but no setuid-bit for my driver.
Member of a group...

praveenkumar:
> I can print the path, but no setuid-bit for my driver.

@gbraad path would be good enough I think

gbraad:
Changed... The output is a little bit different, but at least we would know if the user has installed the driver or not.

gbraad:
retest this please

gbraad:
retest this please

cmd/minishift/cmd/config/config.go
```diff
@@ -52,54 +52,78 @@ const (
...
-	DiskSize         = createFlag("disk-size", SetString, []setFn{IsValidDiskSize}, []setFn{RequiresRestartMsg}, true)
-	VmDriver         = createFlag("vm-driver", SetString, []setFn{IsValidDriver}, []setFn{RequiresRestartMsg}, true)
-	OpenshiftVersion = createFlag("openshift-version", SetString, nil, nil, true)
-	HostOnlyCIDR     = createFlag("host-only-cidr", SetString, []setFn{IsValidCIDR}, nil, true)
-	DockerEnv        = createFlag("docker-env", SetSlice, nil, nil, true)
-	DockerEngineOpt  = createFlag("docker-opt", SetSlice, nil, nil, true)
-	InsecureRegistry = createFlag("insecure-registry", SetSlice, nil, nil, true)
-	RegistryMirror   = createFlag("registry-mirror", SetSlice, nil, nil, true)
-	AddonEnv         = createFlag("addon-env", SetSlice, nil, nil, true)
+	ISOUrl           = createConfigSetting("iso-url", SetString, []setFn{IsValidUrl}, []setFn{RequiresRestartMsg}, true)
```
hferentschik: :-)

cmd/minishift/cmd/start.go
```diff
@@ -274,11 +291,24 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
...
 	var hostVm *host.Host
+
+	// configuration with these settings only happen on create
+	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
+	if !isRestart {
+		fmt.Println("-- Instance will be configured with ...")
+		fmt.Println("   Memory:   ", units.HumanSize(float64((machineConfig.Memory/units.KiB)*units.GB)))
+		fmt.Println("   vCPUs :   ", machineConfig.CPUs)
+		fmt.Println("   Disksize: ", units.HumanSize(float64(machineConfig.DiskSize*units.MB)))
+		fmt.Println("   Boot ISO: ", machineConfig.MinikubeISO)
```
hferentschik: This might be  bit confusing, since the output is for example:

    Boot ISO:  https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso

This does not take into affect that the ISO is cached. Really I would expect something like:

    ~/.minishift/cache/iso/minishift-b2d.iso

where the ISO is cached. Given that we don't checksum the ISO and neither properly version, I am pretty sure we can even end up in cases where the cached ISO does not matched the displayed URL. 

I would suggest to remove this for now until we handle the ISO cache a bit better.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,304 @@
...
+	// follow symlinks
+	if fi.Mode()&os.ModeSymlink != 0 {
+		path, err = os.Readlink(path)
+		if err != nil {
+			return false
+		}
+	}
+	fmt.Println("\n   Driver is available at", path)
+
+	fmt.Printf("   Checking for setuid bit ... ")
```
hferentschik: :+1 Nice, works like a charm. So it is possible to check after all.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,304 @@
...
+
+func checkHypervDriver() bool {
+	switchEnv := os.Getenv("HYPERV_VIRTUAL_SWITCH")
+	if switchEnv == "" {
+		return false
+	}
+	return true
+}
+
+// Check to make sure the instance has an IPv4 address.
```
hferentschik: OOI, why don't you not use proper Godoc syntax?

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,304 @@
...
+	cmd := exec.Command("ping", "-n 1", ip)
+	stdoutStderr, err := cmd.CombinedOutput()
+	if err != nil {
+		return false
+	}
+	fmt.Printf("%s\n", stdoutStderr)
+	return false
+}
+
+// Checks if the VM has connectivity to the outside network
```
hferentschik: "checkIPConnectivity checks if the VM has connectivity to the outside network"

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,304 @@
...
+	ipToPing := viper.GetString(startFlags.CheckNetworkPingHost.Name)
+	if ipToPing == "" {
+		ipToPing = "8.8.8.8"
+	}
+
+	fmt.Printf("\n   Pinging %s ... ", ipToPing)
+	return miniutil.IsIPReachable(driver, ipToPing, false)
+}
+
+// Allows to test outside connectivity and possible proxy support
```
hferentschik: comment format, see above

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,51 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
+
+[[minshift-start-failed-due-to-startup-check]]
+== Minishift startup check failed
+
+During the start of Minishift we perform several pre-flight checks to make sure the Minishift VM and OpenShift Cluster are able to start without any issues. Therefore, if something is not correctly configured, we fail a test and the startup will stop.
+
+If the startup failed for one of the driver plugins, please make sure they are correctly configured by checking the  xref:../getting-started/setting-up-driver-plugin.adoc#[setting up the driver plugin] topic. If for some reason you want to force the startup, you can configure Minishift to treat these errors as warnings as follows:
+
+. For KVM/Libvirt on Linux, run the following command:
+
```
hferentschik: You need a '+' character here to make the listing part of the enumeration. Numbering as well as indentation are off now. Please review the docs in rendered form as well.

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,51 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
+----
+$ minishift config set warn-check-kvm-driver true
+----
+
+. For xhyve on macOS, run the following command:
+
```
hferentschik: '+'

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,51 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
+----
+$ minishift config set warn-check-xhyve-driver true
+----
+
+. For Hyper-V on Windows, run the following command:
+
```
hferentschik: '+'

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,51 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
+----
+# minishift config set warn-check-hyperv-driver true
```
hferentschik: '$' instead of '#'

docs/source/using/troubleshooting.adoc
gbraad: No idea about this...  Asked for review by @thatdocslady for this as I am not familiar enough to adoc

docs/source/using/troubleshooting.adoc
gbraad: was part of a previous write-up. yeps, missed.

gbraad: it should actually be `PS >`

cmd/minishift/cmd/start.go
LalatenduMohanty: Agree with @hferentschik . It should show the local cached version.  However the other displayed information are nice.

cmd/minishift/cmd/start.go
```diff
@@ -126,6 +131,10 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	ensureNotRunning(libMachineClient, constants.MachineName)
 	validateOpenshiftVersion()
+
+	// preflight check (before start)
+	preflightChecksBeforeStartingHost()
```
LalatenduMohanty: 👍

cmd/minishift/cmd/start.go
```diff
@@ -134,9 +143,14 @@ func runStart(cmd *cobra.Command, args []string) {
 	// we need to determine whether this is a restart prior to potentially creating a new VM
 	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
 
-	hostVm, ip := startHost(libMachineClient)
+	fmt.Printf("-- Starting local OpenShift cluster")
+	hostVm := startHost(libMachineClient)
 	registerHost(libMachineClient)
 
+	// preflight checks (after start)
+	preflightChecksAfterStartingHost(hostVm.Driver)
```
LalatenduMohanty: I think these tests are making sure OpenShift works fine after `clusterup`. So why not changing it to preflightChecksBeforeClusterUp()?

LalatenduMohanty: I think these tests are making sure OpenShift works fine after clusterup. So why not changing it to preflightChecksBeforeClusterUp()?

cmd/minishift/cmd/start.go
hferentschik: > Agree with @hferentschik . It should show the local cached version

Right, since in the end it is always an ISO somewhere on disk. That said, I think it would take quite some doing to get this right. There are several things which needs addressing. I'd prefer to remove this and create a follow up issue where we can take care of some ISO caching related things (for example also adding the ISO version into the cache path). There are a lot of nice things in this pull request and I'd prefer to get them merged asap.

docs/source/using/troubleshooting.adoc
hferentschik: hmm, but you see how it rendered, right? I would assume it does not match your expectation. I understand that not everyone is familiar with asciidoc formatting, but what's the difference to trying to solve a coding/API issue?

docs/source/using/troubleshooting.adoc
hferentschik: Or that

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,304 @@
...
+func checkHttpConnectivity(driver drivers.Driver) bool {
+	urlToRetrieve := viper.GetString(startFlags.CheckNetworkHttpHost.Name)
+	if urlToRetrieve == "" {
+		urlToRetrieve = "http://minishift.io/index.html"
```
LalatenduMohanty: I am still concerned about URL `http://minishift.io/index.html`. But it is ok as this test if failed will show a warning.

cmd/minishift/cmd/start.go
gbraad: They are less about `oc cluster up`, and more related to the startHost function (closer in proximity) and setting up the VM.

LalatenduMohanty:
@gbraad I tried the PR with OCP 3.6 and here is the output and it looks fine to me. 

```
$ minishift start --openshift-version 3.6.0
-- Instance will be configured with ...
   Memory:    2 GB
   vCPUs :    2
   Disksize:  20 GB
   Pinging 8.8.8.8 ... OK
-- Checking if persisten storage volume is mounted ... OK
-- Checking available disk space ... 0% OK
   host-config-dir ... OK
   host-data-dir ... OK
   host-pv-dir ... OK
   host-volumes-dir ... OK
   version ... OK
   routing-suffix ... OK
Pulled 1/4 layers, 40% complete
Pulled 1/4 layers, 46% complete
Pulled 1/4 layers, 57% complete
Pulled 2/4 layers, 84% complete
Pulled 3/4 layers, 100% complete
Extracting
Image pull complete
OpenShift server started.

    https://192.168.42.245:8443

You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin
```

thatdocslady:
@gbraad I just noticed by reading Lala's output that there's a doc string that I missed (not sure if you added it or if it was there before?): "-- Instance will be configured with ..." > we should change to "Minishift VM" or "OpenShift cluster" depending on what the string is talking about, "instance" is ambiguous. Thnx :)

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,70 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
 If you want to perform development tasks inside the Minishift VM, it is recommended that you use containers, which are stored in persistent storage volumes, and reuse the xref:../using/docker-daemon.adoc#[Minishift Docker daemon].
+
+
+[[minshift-startup-check-failed]]
+== Minishift startup check failed
+
+While Minishift starts, it runs several startup checks to make sure that the Minishift VM and the OpenShift Cluster are able to start without any issues.
+If any configuration is incorrect or missing, the startup checks fail and Minishift does not start.
+
+If you want to force Minishift to start despite any failed checks, you can instruct Minishift to treat these errors as warnings as follows:
```
hferentschik: Are these meant as examples, aka:

"If you want to force Minishift to start despite a failed check, you can instruct Minishift to treat these errors as warnings. For example to treat the KVM driver check as warning: ..."

If it is not an example, I am a bit confused, since it reads to me that depending on the driver, I just have to set say `minishift config set warn-check-xhyvek-driver true` and Minishift will start. 

I guess I am trying to figure out why the drivers checks are described in so much detail, before you start with: "The following sections describe the different startup checks.:" Would they not fall under there?

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,70 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
+- For Hyper-V on Windows, run the following command:
++
+----
+$ minishift config set warn-check-hyperv-driver true
+----
+
+The following sections describe the different startup checks.
+
+Driver plug-in configuration::
+One of the startup checks verifies that the relevant driver plug-in is configured correctly.
```
hferentschik: Should the driver checks from above not live here?

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,70 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
+
+- For Hyper-V on Windows, run the following command:
++
+----
+$ minishift config set warn-check-hyperv-driver true
+----
+
+The following sections describe the different startup checks.
+
+Driver plug-in configuration::
```
hferentschik: Any reason you prefer a labeled list here? Would a sub-heading not be nicer (===)? In this case one has also the option to increase the TOC level and get the headings visible in the TOC

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,70 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
+One of the startup checks verifies that the relevant driver plug-in is configured correctly.
+If this startup check fails, review the xref:../getting-started/setting-up-driver-plugin.adoc#[setting up the driver plug-in] topic and configure the driver.
+
+Persistent storage volume configuration and usage::
```
hferentschik: I'd be careful with the term "Persistent storage volume", it might be confusing with OpenShift's persistent volumes. Also we don't use this term anywhere else, so it would need some explanation. I'd try something like:

"Persistent user data partition" and then in the description, something like: "To keep state between restarts of the VM, Minishift creates a persistent partition on which configuration and user data is stored."

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,70 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
+To skip this test, run run the following command:
++
+----
+$ minishift config set skip-check-storage-usage true
+----
+
+External network connectivity::
+After the Minishift VM starts, it runs several network checks to verify whether external connectivity is possible from within the Minishift VM.
++
+By default, network checks are configured to treat the errors as warnings, because of the diversity of the development environments.
```
hferentschik: error == failures?

hferentschik: By default, network checks are configured to treat the detected failures as warnings, due to the diversity of network and development environments.

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,70 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
++
+For example, one of the network checks pings an external host. You can change the host by running the following command:
++
+----
+$ minishift config set check-network-ping-host <host-IP-address>
+----
++
+You can replace the `host IP address` with the address of your internal DNS server, proxy host, or an external host that you can reach from your machine.
++
+Because proxy connectivity might be problematic, you can run a check that tries to retrieve an external resource.
```
hferentschik: resource == URL? Or can I hit something else as well?

budhrg:
> we should change to "Minishift VM" or "OpenShift cluster" 

Since we have `OpenShift cluster` in line `Starting local OpenShift cluster using 'kvm' hypervisor ...` we can omit `OpenShift` and just show `Cluster will be configured with ...` 

One more thing, what does `v` indicate in `vCPUs` ?

gbraad:
@budhrg https://www.google.com/search?q=vcpu

I'd rather stay consistent with the rest of the terms used
```go
	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
	if !isRestart {
		fmt.Println("-- Minishift VM will be configured with ...")
		...
	}
	fmt.Printf("-- Starting Minishift VM ... ")
```

docs/source/using/troubleshooting.adoc
gbraad: Anything that cURL accepts as file URI scheme, but in our case we would would like it to be a HTTP resource, else it would not really test possible proxy configurations.

docs/source/using/troubleshooting.adoc
gbraad: Persistent Storage Volume has been used in other parts of the documentation as referring to `/mnt/sda1/`, such as the recent troubleshooting entry for the overlay.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,331 @@
...
+			checkHypervDriver,
+			"Checking if Hyper-V driver is configured",
+			false, startFlags.WarnCheckHyperVDriver.Name,
+			"Hyper-V virtual switch is not set")
+	}
+}
+
+// preflightChecksAfterStartingHost is executed after the startHost function.
+func preflightChecksAfterStartingHost(driver drivers.Driver) {
+	preflightCheckSucceedsOrFailsWithDriver(
```
gbraad: It is the "pre-flight check" that "succeeds or fails"... not the "pre-flight", as that refers to a grouping, and "check" is here used as a noun, not the "verb" form, else I would have preferred to use the "gerund" form.

Kept this way to indicate in the `start_preflight.go` that two groupings of test exists, before and after the start of the VM. It is called StartingHost to be indicative to how this is named in the `start.go` file.

docs/source/using/troubleshooting.adoc
gbraad: @thatdocslady I think they can be moved down... Any suggestions about 'flow'?

docs/source/using/troubleshooting.adoc
gbraad: Yep, http://www.thesaurus.com/browse/error

It is not grammatical nice to say: "to treat failure as warnings", as it gives it a personal connotation. @thatdocslady WDYT?

gbraad:
@LalatenduMohanty if no problems, please approve. You can comment that you are OK with the technical side. As at the moment, some documentation issues can be addressed.

docs/source/using/troubleshooting.adoc
hferentschik: > Persistent Storage Volume has been used in other parts of the documentation as referring to /mnt/sda1/, such as the recent troubleshooting entry for the overlay.

Hmm, anywhere else? I don't think so, right? We might want to change it there as well.

docs/source/using/troubleshooting.adoc
hferentschik: I guess we should be explicit then and just say "URL"

cmd/minishift/cmd/start.go
```diff
@@ -274,11 +291,25 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
...
+	fmt.Printf(" using '%s' hypervisor ...\n", machineConfig.VMDriver)
 	var hostVm *host.Host
+
+	// configuration with these settings only happen on create
+	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
+	if !isRestart {
+		fmt.Println("-- Minishift VM will be configured with ...")
+		fmt.Println("   Memory:    ", units.HumanSize(float64((machineConfig.Memory/units.KiB)*units.GB)))
+		fmt.Println("   vCPUs :    ", machineConfig.CPUs)
+		fmt.Println("   Disk size: ", units.HumanSize(float64(machineConfig.DiskSize*units.MB)))
```
hferentschik: You know what would be nice to show, the OpenShift version which gets provisioned. I just had the case that I after an upgrade ended up with OpenShift 1.5.1, even though the default is 3.6.0. I had 1.5.1 in the persistent config. Knowing the OpenShift version which is getting provisioned would be nice.

cmd/minishift/cmd/start_preflight.go
hferentschik: Not quite sure I follow, but ok.

cmd/minishift/cmd/start.go
gbraad: Possible, but... and this is my personal opinion: Lingering in a quest for perfection is not ideal. This PR has been open for already 12 days now... although mostly waiting for comments. But now, most of the current suggestions are nitpicking about function names, which have not much to do with the actual functionality and adding new requests (although small). This is mostly slowing integration down and allows less time to expose it to general testing/usage.

I believe these can be handled as part of https://en.wikipedia.org/wiki/Code_refactoring. Note: Note: we unnecessarily made refactoring harder as we want each commit referring to an issue number?

My 2 BCC cents

cmd/minishift/cmd/start_preflight.go
```diff
@@ -0,0 +1,331 @@
...
+		startFlags.SkipCheckNetworkHTTP.Name,
+		checkHttpConnectivity, driver,
+		"Checking HTTP connectivity from the VM",
+		true, startFlags.WarnCheckNetworkHTTP.Name,
+		"VM cannot connect to external URL with HTTP")
+	preflightCheckSucceedsOrFailsWithDriver(
+		startFlags.SkipCheckStorageMount.Name,
+		checkStorageMounted, driver,
+		"Checking if persistent storage volume is mounted",
```
praveenkumar: looks good now 👍

docs/source/using/troubleshooting.adoc
thatdocslady: I agree with @gbraad, if we leverage known terminology from the logging world, the terms are usually something like INFO/WARNING/ERROR, where ERROR level means that some action is taken, but WARNING level is more FYI.

docs/source/using/troubleshooting.adoc
thatdocslady: This can go either way but we should be consistent, so if we decide to use URL or URI, we should standardize it in other occurances of this term.

docs/source/using/troubleshooting.adoc
```diff
@@ -243,3 +243,71 @@ Workaround: Avoid installing packages or storing large files in the root filesys
...
+
+While Minishift starts, it runs several startup checks to make sure that the Minishift VM and the OpenShift Cluster are able to start without any issues.
+If any configuration is incorrect or missing, the startup checks fail and Minishift does not start.
+
+The following sections describe the different startup checks.
+
+Driver plug-in configuration::
+One of the startup checks verifies that the relevant driver plug-in is configured correctly.
+If this startup check fails, review the xref:../getting-started/setting-up-driver-plugin.adoc#[setting up the driver plug-in] topic and configure the driver.
+
```
thatdocslady: Need to add a + sign here to bing the following paragraph to the sub-section (if you click the View button in the diff tab of the PR you'll see that the alignment is off).

docs/source/using/troubleshooting.adoc
thatdocslady: I agree with @gbraad , if we leverage known terms from the logging world, there are usually 3 levels, INFO/WARNING/ERROR. The Error itself isn't the failure, but it's what triggers an action or stop an existing process. Warning level events don't trigger anything or stop anything from happening but they are informational to indicate there is a problem.

docs/source/using/troubleshooting.adoc
thatdocslady: I have no personal preference but if what @gbraad is saying is true then we should use the most global term. Could it be that some users will change the URL to something else? If we use URL or URI it means we're limiting the usage. It's up to you but of course we should be consistent and use the same term everywhere :)

hferentschik:
Rebased and merged (with some minor changes to the docs).

PR #1165, opened 2017-07-24
Issue #1153 Proxy set not working with docker env

praveenkumar:
This will close #1154 because what we need a solution which works for CLI arguments and config set both. if we only apply @nimatt patch part of his PR then it will only going to work with CLI arguments and then if I set anything in the config set that would fail. 

```
# PR from @nimatt
$ ./minishift start --http-proxy http://fedora:fedora123@10.70.49.109:3128 --docker-env abc=foo
DockerEnv:[]string{"abc=foo", "HTTP_PROXY=http://fedora:fedora123@10.70.49.109:3128", "NO_PROXY=localhost,127.0.0.1,172.30.1.1"},

$ ./minishift config view
- docker-env           : [abc=foo]

$ ./minishift start --http-proxy http://fedora:fedora123@10.70.49.109:3128   ==> not take from config
DockerEnv:[]string{"HTTP_PROXY=http://fedora:fedora123@10.70.49.109:3128", "NO_PROXY=localhost,127.0.0.1,172.30.1.1"},

# Current PR
$ ./minishift start --http-proxy http://fedora:fedora123@10.70.49.109:3128   ==> not take from config
DockerEnv:[]string{"abc=foo", "HTTP_PROXY=http://fedora:fedora123@10.70.49.109:3128", "NO_PROXY=localhost,127.0.0.1,172.30.1.1"},
```

Addresses issue #1153

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -236,9 +236,13 @@ func handleProxies() *util.ProxyConfig {
 	return proxyConfig
 }
 
+// getSlice return slice for provided key if value type is []interface{} otherwise nil
 func getSlice(key string) []string {
 	if viper.IsSet(key) {
-		return viper.GetStringSlice(key)
+		value := viper.Get(key)
+		if _, ok := value.([]interface{}); ok {
+			return viper.GetStringSlice(key)
+		}
 	}
 	return nil
 }
@@ -262,10 +266,10 @@ func startHost(libMachineClient *libmachine.Client) (*host.Host, string) {
 		CPUs:             viper.GetInt(startFlags.CPUs.Name),
 		DiskSize:         calculateDiskSizeInMB(viper.GetString(startFlags.DiskSize.Name)),
 		VMDriver:         viper.GetString(startFlags.VmDriver.Name),
-		DockerEnv:        getSlice(startFlags.DockerEnv.Name),
-		DockerEngineOpt:  getSlice(startFlags.DockerEngineOpt.Name),
+		DockerEnv:        append(dockerEnv, getSlice(startFlags.DockerEnv.Name)...),
+		DockerEngineOpt:  append(dockerEngineOpt, getSlice(startFlags.DockerEngineOpt.Name)...),
 		InsecureRegistry: determineInsecureRegistry(startFlags.InsecureRegistry.Name),
-		RegistryMirror:   getSlice(startFlags.RegistryMirror.Name),
+		RegistryMirror:   append(registryMirror, getSlice(startFlags.RegistryMirror.Name)...),
 		HostOnlyCIDR:     viper.GetString(startFlags.HostOnlyCIDR.Name),
 		ShellProxyEnv:    shellProxyEnv,
 	}
--- a/cmd/minishift/cmd/start_test.go
+++ b/cmd/minishift/cmd/start_test.go
@@ -29,6 +29,11 @@ import (
 	"github.com/minishift/minishift/cmd/minishift/cmd/config"
 
 	"fmt"
+	"sort"
+	"strings"
+
+	"bytes"
+
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minikube/tests"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
@@ -36,8 +41,6 @@ import (
 	"github.com/minishift/minishift/pkg/testing/cli"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/viper"
-	"sort"
-	"strings"
 )
 
 var (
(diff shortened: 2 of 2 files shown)


nimatt:
@praveenkumar I ran a couple of problematic scenarios described in #1153 and this PR seems to solve them nicely (I've only tested with b2d).

praveenkumar:
@nimatt thanks for testing out 👍

cmd/minishift/cmd/start.go
```diff
@@ -238,6 +238,10 @@ func handleProxies() *util.ProxyConfig {
 
 func getSlice(key string) []string {
 	if viper.IsSet(key) {
+		value := viper.Get(key)
+		if _, ok := value.(string); ok {
+			return nil
+		}
```
LalatenduMohanty: @praveenkumar I did some testing around this and it seems, when we pass --docker-env to `minishift start` getSlice is returning nil i.e. viper is not considering it as a string slice. So this method is useful when we read from config.json. Can you please add comments communicating the same for the function?

cmd/minishift/cmd/start.go
```diff
@@ -236,9 +236,13 @@ func handleProxies() *util.ProxyConfig {
 	return proxyConfig
 }
 
+// getSlice return slice for provided key if value type is []interface{} otherwise nil
 func getSlice(key string) []string {
 	if viper.IsSet(key) {
-		return viper.GetStringSlice(key)
+		value := viper.Get(key)
+		if _, ok := value.([]interface{}); ok {
```
hferentschik: Can you give me a quick run down on what's going on here and why? We did verify this, didn't we? Are we not able to write some tests for this? Some generic test which are centred around getting string slices asserting that no matter of whether we get the slice from the command line or from the config, we end up with the right data. The test would also serve as "documentation" on what's actually expected.

hferentschik:
So are we saying that all the options which take slices are broken atm and this fix is needed? If so, we maybe should cut a 1.3.1 release with the fix which we then can base the CDK build upon.

cmd/minishift/cmd/start_test.go
```diff
@@ -384,6 +387,32 @@ func TestDetermineIsoUrlWithInvalidName(t *testing.T) {
 	determineIsoUrl("foo")
 }
 
+func Test_getslice(t *testing.T) {
```
hferentschik: I feels like we are testing two things here, right. One retrieving a slice from the config and the other retrieving a slice from the command lines. Let's put this into two separate tests making the intention clearer

praveenkumar: @hferentschik sure

hferentschik:
After discussion with @praveenkumar we decided to merge this pull request as is, since it is already confirmed to work. This will allow us to cut a 1.3.1 release.

Moving forward, we create a follow up issue for this, since the code is very confusing and basically working around some underlying Viper issues. Viper handles string slices differently when retrieved from the config file vs from the command file. We had the same issue with the _addon-env_ flag. For this we introduced a custom stringSliceValue - https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/util/shared_flags.go#L76 which does the handling of slices properly, regardless of whether they are retrieved from the config or from the command line. We should align approaches. Using custom value types is actually using a proper public API, so it is not just a hack. It might initially not be clear why a custom type is needed here, but at least the implementation is correct and consistent. 

@agajdosi, @amitkrout - how come this was never caught in the QE testing? Using proxies for examples should fail.

cmd/minishift/cmd/start_test.go
```diff
@@ -384,6 +387,37 @@ func TestDetermineIsoUrlWithInvalidName(t *testing.T) {
...
+
+	expectedSlice := []string{"hello", "world"}
+	for i, v := range getSlice("bar") {
+		if expectedSlice[i] != v {
+			t.Errorf("expected %+v. Got %+v", expectedSlice[i], v)
+		}
+	}
+}
+
+func Test_getslice_withCommandLine(t *testing.T) {
```
hferentschik: :+1 nice two tests

praveenkumar:
@hferentschik   #1176 follow up issue.

hferentschik:
Rebased and merged, under the condition that we revisit this after the 1.3.1 release.

@praveenkumar, please add a reference to the follow up issue.

agajdosi:
@hferentschik I did proper and intense checks when this feature was new, there it worked - tested it today with 1.1.0 and it is fine. For later minishift tests I did checks if request for .iso, oc to github.com are present in log and if there are calls to docker registry from image pull - they were not here, which was missed by me due to the fact that there are also calls to docker registry after the pull, so overall it looked like log contained traces of every part of startup.

Now for CDK I am checking it in real time, but blocking proxy environment is needed.

hferentschik:
@agajdosi 

> I did proper and intense checks when this feature was new and version later, there it worked

Sure.  I guess here is partly the rub. Without fully automated tests, regressions won't be found. We cannot rely on this to be manually checked. I mean this not at a personal critique, but rather pointing out that we once more have been bitten by a proxy issue. Without proper automated tests we won't ever be able to get this solved for good. 

Good thing is that we make some progress on the proxy testing front. See pull request #1095.

> For later minishift tests I did checks if request for .iso, oc to github.com are present in log and if there are calls to docker registry from image pull - they were not here, which was missed by me due to the fact that there are also calls to docker registry after the pull, so overall it looked like log contained traces of every part of startup.

Which shows again the problem we discussed before. Just verifying that there is "some" output in the proxy logs will always fall short. We need to also ensure/verify that nothing goes out via the default HTTP(S) ports.

agajdosi:
@hferentschik Not taking it personally, just wanted to explain how it went to be here.
> Without fully automated tests, regressions won't be found. We cannot rely on this to be manually checked.

Completely agree, I would love it to be able just to run `make integration` and see the results. It is coming however and hopefully the progress on this will accelerate as it will save our time and also as less code and more cucumber "only" code will be needed.

> Just verifying that there is "some" output in the proxy logs will always fall short.

For #1095 I am not going the really proper way - it will be just checking the log, but at least it is something on this topic, it covers some cases. However planning to not end just here.

hferentschik:
> Completely agree, I would love it to be able just to run make integration and see the results. It is coming however and hopefully the progress on this will accelerate as it will save our time and also as less code and more cucumber "only" code will be needed.

+1 Seems we agree on what we need to do :-)

> For #1095 I am not going the really proper way - it will be just checking the log, but at least it is something on this topic, it covers some cases. However planning to not end just here

Sure. It is a first step into the right direction.

PR #1166, opened 2017-07-24
Fixes #1164 Correct duplicate IDs in AsciiDoc sources

### LINKED ISSUE #1164: Duplicate IDs in docs hamper downstreaming
The `homebrew-install` and `overview` IDs are duplicated in the AsciiDoc docs sources, which prevents seamless downstreaming of the docs -- downstream docs cannot be built with dupl. IDs, thus necessitating additional modifications.

I propose to submit a PR that would make all the IDs unique.

rkratky:
Fixes #1164 Correct duplicate IDs in AsciiDoc sources

### DIFF
--- a/docs/source/contributing/ci.adoc
+++ b/docs/source/contributing/ci.adoc
@@ -8,7 +8,7 @@
 
 toc::[]
 
-[[overview]]
+[[minishift-ci-overview]]
 == Overview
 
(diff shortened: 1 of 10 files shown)


gbraad:
add to whitelist

gbraad:
... besides that, it would all be OK.

PR #1168, opened 2017-07-24
[Do not merge] Fixes #1137 Add option to run experimental Service Catalog

### LINKED ISSUE #1137: Expose service catalog option for OpenShift 3.6
It is very important that we put the service broker into the hands of users asap.  To begin experimentation of mapping through it (provisioning, binding) into various public cloud services.

gbraad:
Addresses issue #1137

TODO:
  * ~~Does not check if `oc cluster up` supports installing with --service-catalog=true` (See: #1169)~~
  * ~~option to specify extra flags (see #1170)~~

### DIFF
--- a/pkg/minishift/clusterup/clusterup.go
+++ b/pkg/minishift/clusterup/clusterup.go
@@ -66,7 +66,7 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 
 	for key, value := range clusterUpParams {
 		if !oc.SupportFlag(key, config.OcPath, runner) {
-			errors.New(fmt.Sprintf("Flag %s is not supported for oc version %s. Use 'openshift-version' flag to select a different version of OpenShift.", key, config.OpenShiftVersion))
+			return errors.New(fmt.Sprintf("Flag %s is not supported for oc version %s. Use 'openshift-version' flag to select a different version of OpenShift.", key, config.OpenShiftVersion))
 		}
 		cmdArgs = append(cmdArgs, "--"+key)
 		cmdArgs = append(cmdArgs, value)
@@ -74,7 +74,7 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 
 	exitCode := runner.Run(os.Stdout, os.Stderr, config.OcPath, cmdArgs...)
 	if exitCode != 0 {
-		errors.New("Error starting the cluster.")
+		return errors.New("Error starting the cluster.")
 	}
 	return nil
 }
--- a/pkg/minishift/clusterup/clusterup.go
+++ b/pkg/minishift/clusterup/clusterup.go
@@ -68,6 +68,7 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 		if !oc.SupportFlag(key, config.OcPath, runner) {
 			return errors.New(fmt.Sprintf("Flag %s is not supported for oc version %s. Use 'openshift-version' flag to select a different version of OpenShift.", key, config.OpenShiftVersion))
 		}
+
 		cmdArgs = append(cmdArgs, "--"+key)
 		cmdArgs = append(cmdArgs, value)
 	}
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -77,6 +77,7 @@ var (
 	OpenshiftEnv      = createFlag("openshift-env", nil, nil, nil, false)
 	Metrics           = createFlag("metrics", SetBool, nil, nil, true)
 	Logging           = createFlag("logging", SetBool, nil, nil, true)
+	ServiceCatalog    = createFlag("service-catalog", SetBool, nil, nil, true)
 
 	// Setting proxy
 	NoProxyList = createFlag("no-proxy", SetString, nil, nil, true)
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -410,6 +410,7 @@ func initClusterUpFlags() *flag.FlagSet {
 	clusterUpFlagSet.StringSliceVarP(&openShiftEnv, startFlags.OpenshiftEnv.Name, "e", []string{}, "Specify key-value pairs of environment variables to set on the OpenShift container.")
 	clusterUpFlagSet.Bool(startFlags.Metrics.Name, false, "Install metrics (experimental)")
 	clusterUpFlagSet.Bool(startFlags.Logging.Name, false, "Install logging (experimental)")
+	clusterUpFlagSet.Bool(startFlags.ServiceCatalog.Name, false, "Install service catalog (experimental)")
 	clusterUpFlagSet.String(startFlags.OpenshiftVersion.Name, version.GetOpenShiftVersion(), fmt.Sprintf("The OpenShift version to run, eg. %s", version.GetOpenShiftVersion()))
 	clusterUpFlagSet.String(startFlags.NoProxyList.Name, "", "List of hosts or subnets for which no proxy should be used.")
 	clusterUpFlagSet.AddFlag(cmdutil.HttpProxyFlag)


gbraad:
`SupportFlag` is checked, but it doesn't show the error... but rather dumps the `oc cluster -h` text and relies on `oc` giving the error.

This means:
```go
errors.New(fmt.Sprintf("Flag %s is not supported for oc version %s. Use 'openshift-version' flag to select a different version of OpenShift.", key, config.OpenShiftVersion))
```
is not shown.

LalatenduMohanty:
Not working for me at this moment. Trying to find out the root cause.
```
$ minishift start --openshift-version v3.6.0-rc.0 --service-catalog=true
Starting OpenShift using openshift/origin:v3.6.0-rc.0 ...
Pulling image openshift/origin:v3.6.0-rc.0
Pulled 2/4 layers, 66% complete
Pulled 3/4 layers, 86% complete
Extracting
Image pull complete
-- Checking for openshift/origin:v3.6.0-rc.0 image ... 
   Pulling image openshift/origin:v3.6.0-rc.0
   Pulled 2/4 layers, 66% complete
   Pulled 3/4 layers, 86% complete
   Extracting
   Image pull complete
-- Checking Docker daemon configuration ... OK
-- Checking for available ports ... OK
-- Checking type of volume mount ... 
   Using Docker shared volumes for OpenShift volumes
-- Creating host directories ... OK
-- Finding server IP ... 
   Using 192.168.42.59 as the server IP
-- Checking service catalog version requirements ... OK
-- Starting OpenShift container ... 
   Creating initial OpenShift configuration
FAIL
   Error: could not update OpenShift configuration
   Caused By:
     Error: failed generating signer certificate, command output: 
     error: exec error: exec: "oc": executable file not found in $PATH
Error during 'cluster up' execution: Error starting the cluster.
```

gbraad:
@LalatenduMohanty 
notice it pulls the image twice for you
```
Pulling image openshift/origin:v3.6.0-rc.0
Pulled 2/4 layers, 66% complete
Pulled 3/4 layers, 86% complete
Extracting
Image pull complete
```

followed by
```
-- Checking for openshift/origin:v3.6.0-rc.0 image ... 
   Pulling image openshift/origin:v3.6.0-rc.0
   Pulled 2/4 layers, 66% complete
   Pulled 3/4 layers, 86% complete
   Extracting
   Image pull complete
```

LalatenduMohanty:
@gbraad yes, it is odd. I have even deleted ~/.minishift and ~/.kube and I still get the same error. I have tried around 5 times but I still get the same error .@csrwng any suggestion. cc @jorgemoralespou

gbraad:
it service-catalog it feels like it hangs: http://pastebin.centos.org/131791/
```
$ minishift start --openshift-version v3.6.0-rc.0 --service-catalog=true
Starting OpenShift using openshift/origin:v3.6.0-rc.0 ...
Pulling image openshift/origin:v3.6.0-rc.0
Pulled 1/4 layers, 64% complete
Pulled 2/4 layers, 73% complete
Pulled 3/4 layers, 89% complete
Extracting
Image pull complete
^C
```
Does it hang in "Image pull complete" ? Too little feedback... after 3+ mins I broke it off...


And without the service catalog it takes about 2+ mins on the "Image pull complete" step. I saw no progress... and then finally?
```
$ minishift start --openshift-version v3.6.0-rc.0 
Starting OpenShift using openshift/origin:v3.6.0-rc.0 ...
Pulling image openshift/origin:v3.6.0-rc.0
Pulled 1/4 layers, 27% complete
Pulled 1/4 layers, 47% complete
Pulled 2/4 layers, 58% complete
Pulled 2/4 layers, 77% complete
Pulled 3/4 layers, 88% complete
Extracting
Image pull complete
OpenShift server started.

    https://192.168.42.237:8443

You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin

-- Applying addon 'xpaas':........................................................................................
XPaaS OpenShift imagestream and templates installed
See https://github.com/openshift/openshift-ansible/tree/master/roles/openshift_examples/files/examples/v3.6
```

jorgemoralespou:
I think this is an issue with OC cluster output not coming out on real time
as things happen. I think I've seen the same outside minishift with latest
OC.

Cc/ @csrwng

gbraad:
so this output reduction is not by design... ? *phew*

gbraad:
```
minishift start --openshift-version v3.6.0-rc.0 --service-catalog=true
Starting OpenShift using openshift/origin:v3.6.0-rc.0 ...
Pulling image openshift/origin:v3.6.0-rc.0
Pulled 1/4 layers, 47% complete
Pulled 1/4 layers, 70% complete
Pulled 2/4 layers, 78% complete
Pulled 3/4 layers, 89% complete
Extracting
Image pull complete
-- Checking for openshift/origin:v3.6.0-rc.0 image ... 
   Pulling image openshift/origin:v3.6.0-rc.0
   Pulled 1/4 layers, 47% complete
   Pulled 1/4 layers, 70% complete
   Pulled 2/4 layers, 78% complete
   Pulled 3/4 layers, 89% complete
   Extracting
   Image pull complete
-- Checking Docker daemon configuration ... OK
-- Checking for available ports ... OK
-- Checking type of volume mount ... 
   Using Docker shared volumes for OpenShift volumes
-- Creating host directories ... OK
-- Finding server IP ... 
   Using 192.168.42.10 as the server IP
-- Checking service catalog version requirements ... OK
-- Starting OpenShift container ... 
   Creating initial OpenShift configuration
   Starting OpenShift using container 'origin'
   Waiting for API server to start listening
   OpenShift server started
-- Adding default OAuthClient redirect URIs ... OK
-- Installing registry ... OK
-- Installing router ... OK
-- Importing image streams ... OK
-- Importing templates ... OK
-- Installing service catalog ... FAIL
   Error: failed to start the service catalog apiserver: timed out waiting for the condition
Error during 'cluster up' execution: Error starting the cluster.
```

hferentschik:
So what is the relation here to pull request #1171? AFIU this pull request will expose the service catalog as explicit option, whereas adds a generic way to add cluster up flags? Should we not generally use the approach in #1171? What is the plan moving forward?

pkg/minishift/clusterup/clusterup.go
```diff
@@ -66,15 +66,16 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 
 	for key, value := range clusterUpParams {
 		if !oc.SupportFlag(key, config.OcPath, runner) {
-			errors.New(fmt.Sprintf("Flag %s is not supported for oc version %s. Use 'openshift-version' flag to select a different version of OpenShift.", key, config.OpenShiftVersion))
+			return errors.New(fmt.Sprintf("Flag %s is not supported for oc version %s. Use 'openshift-version' flag to select a different version of OpenShift.", key, config.OpenShiftVersion))
```
hferentschik: Since #1169 got addresses separately, I guess we need at least a rebase here.

hferentschik:
For what it's worth, this fails consistently to provision for me. Is this really ready for prime time? If we include ##1171, should we really add the explicit flag? It feels not quite ready yet.

jorgemoralespou:
@hferentschik it will work in 3.6 GA. There was a problem in rc-0 that should be fixed.

@gbraad, there's reduction in output when the cluster already exists, but not if it's newly created

hferentschik:
> it will work in 3.6 GA. There was a problem in rc-0 that should be fixed.

Maybe. However, I would argue that one needs to do releases based on things working and not things going to work. It feels wrong to me to enable and release something which does not work yet.

hferentschik:
@jorgemoralespou, after some discussion within the @minishift/minishift-dev we think adding this flag at this stage would be wrong. It makes more sense to add it, once we know it actually works. From a community perspective we release so often, that this would not be an issue. 

We were wondering whether it makes not more sense to support something like the service catalog via an add-on. We don't have to reply on cluster up flags for these (experimental) features. We can expose them in another way (add-ons as well). 

I recall having similar discussions when enabling the metrics flags where we in hindsight also were not so sure whether it was a good idea to expose the 'cluster up' flag. 

@jorgemoralespou  @csrwng, anything stopping us from provisioning the service catalog using an add-on (at least initially). Nothing stops us to further down the road expose the 'cluster up' service-catalog flag. However, it would be at a time were it would be stable and not considered experimental.

gbraad:
The other issue with adding a flag, is that these are actually feature-toggles, and they are not sure to become actual functionality that will be included. We believe pursuing this as add-ons makes more sense.

hferentschik:
Closing this pull request.

csrwng:
/cc @bparees

Sorry for the delay in responding. The service catalog code relies on having an 'oc' binary present to generate certs for the service catalog. This is what's causing things to break. 

There is nothing stopping you from turning it into an add-on that I can think, but I'll let @bparees comment on that, since he owns the feature.

hferentschik:
> The service catalog code relies on having an 'oc' binary present to generate certs for the service catalog. This is what's causing things to break.

Do you have some more information around this? A issue link for example?  I would like to understand better what you are saying here.

Issue #1170, opened 2017-07-25
Add option to specify extra flags for cluster up


gbraad:
Allow to specify extra options/flags for the startup with `oc cluster up` by adding eg. `--oc-cluster-extra-flags="--service-catalog"`

See also discussion on issue #1137

LalatenduMohanty:
As @jorgemoralespou mentioned in https://github.com/minishift/minishift/issues/1137#issuecomment-315601681
```
For the future, maybe it would be useful to be able to provide additional flags that will get passed "as is" to the oc cluster instantiation in minishift, to be able to early test these functionalities without needing to provide additional capabilities in minishift.

e.g.:

minishift start --openshift-version=v3.6.0-rc.0 --openshift-extra-flags='--service-catalog'
```

gbraad:
suggested by @praveenkumar and me it is better to call this option `--oc-cluster-extra-flags=""`

LalatenduMohanty:
@gbraad IMO it should be `--clusterup-extra-flags` as the command name is `cluster up` and it communicates the purpose well to me.

hferentschik:
>  it should be --clusterup-extra-flags as the command name is cluster up and it communicates the purpose well to me

I agree that it should contain 'clusterup' and I might also change the order a bit _--extra-clusterup-flags_.

hferentschik:
I am not sure whether this is such a good idea though. Would this flag try to filter out flags which can also be specified via the CLI? Or would the idea be to pass on flags as is. In the latter case, what applies the command explicitly specified or the one via --extra-clusterup-flags? The behaviour would need to be deterministic at least. 

Also on the community side, we are able to add flags quite quickly, so I wonder whether it makes sense to  have this additional flag. The situation might be different if we were to say that moving forward we would only use this flag to enable features like service catalog (or metrics, even though too late for it now). In this case, however, we would most likely not add the explicit --service-catalog flag. 

Awe should also think about whether this kind of features are not best supported via the add-on mechanism.

hferentschik:
Another approach is to add extra-clusterup-flags as environment variable (MINISHIFT_EXTRA_CLUSTERUP_FLAGS) and/or in combination with a hidden Viper flag. In this case the features becomes more of a power user feature and gets out of harms way for normal users. It also does not continue to clutter the standard minishift start flags.

I think of all the options mentioned so far, I prefer this option best atm. It is somewhere a middle ground.

jorgemoralespou:
[...]

PR #1171, opened 2017-07-25
[Do not merge] Issue #1170 Add option to specify extra flags for cluster up

gbraad:
Addresses issue #1170

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -77,6 +77,7 @@ var (
 	OpenshiftEnv      = createFlag("openshift-env", nil, nil, nil, false)
 	Metrics           = createFlag("metrics", SetBool, nil, nil, true)
 	Logging           = createFlag("logging", SetBool, nil, nil, true)
+	ServiceCatalog    = createFlag("service-catalog", SetBool, nil, nil, true)
 
 	// Setting proxy
 	NoProxyList = createFlag("no-proxy", SetString, nil, nil, true)
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -78,6 +78,8 @@ var (
 	Metrics           = createFlag("metrics", SetBool, nil, nil, true)
 	Logging           = createFlag("logging", SetBool, nil, nil, true)
 	ServiceCatalog    = createFlag("service-catalog", SetBool, nil, nil, true)
+	// ability to specify additional cluster up flags
+	OcClusterExtraFlags = createFlag("oc-cluster-extra-flags", SetString, nil, nil, true)
 
 	// Setting proxy
 	NoProxyList = createFlag("no-proxy", SetString, nil, nil, true)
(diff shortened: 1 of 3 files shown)


gbraad:
Travis fails due to #1172

cmd/minishift/cmd/config/config.go
```diff
@@ -77,6 +77,9 @@ var (
 	OpenshiftEnv      = createFlag("openshift-env", nil, nil, nil, false)
 	Metrics           = createFlag("metrics", SetBool, nil, nil, true)
 	Logging           = createFlag("logging", SetBool, nil, nil, true)
+	ServiceCatalog    = createFlag("service-catalog", SetBool, nil, nil, true)
+	// ability to specify additional cluster up flags
+	OpenshiftExtraFlags = createFlag("openshift-extra-flags", SetString, nil, nil, true)
```
praveenkumar: I would request it to name as `OcClusterExtraFlags` because at first look `OpenshiftExtraFlags` is something we pass to openshift container.

gbraad: I had the same thought, but I followed Jorge's suggestion on #1137 (and it seemed Hardy had no comments at the time)

gbraad: @praveenkumar changed it to `--oc-cluster-extra-flags`...

cmd/minishift/cmd/config/config.go
```diff
@@ -78,6 +78,8 @@ var (
 	Metrics           = createFlag("metrics", SetBool, nil, nil, true)
 	Logging           = createFlag("logging", SetBool, nil, nil, true)
 	ServiceCatalog    = createFlag("service-catalog", SetBool, nil, nil, true)
+	// ability to specify additional cluster up flags
+	OcClusterExtraFlags = createFlag("oc-cluster-extra-flags", SetString, nil, nil, true)
```
LalatenduMohanty: I think the flag name should be `--clusterup-extra-flags`. @praveenkumar @hferentschik ^

hferentschik: Or _--extra-clusterup-flags_

hferentschik:
I wonder whether we really want to do this, right now or whether the curated approach we have is not enough. This introduces now two ways of doing things and potential misuses. We also loose control over being able to test/verify cluster up options first. 

If we go ahead, it also needs to be documented in the docs.

LalatenduMohanty:
@hferentschik  I mostly agree with you.  Also the overall user experience reduces little bit I think i.e. between `--extra-clusterup-flags= --service-catalog` VS `--service-catalog=true` as an user I will prefer --service-catalog=true.  However --extra-clusterup-flags seems tempting for obvious reasons.

LalatenduMohanty:
Also `-extra-clusterup-flags= --service-catalog ` should be `--extra-clusterup-flags=service-catalog`. It should handle multiple flags i.e. --extra-clusterup-flags=service-catalog, metrics` etc.

gbraad:
@LalatenduMohanty I handles multiple... I just forced an `isFlag` check to make sure it is preceded with `--`, but this can be removed, as the `trim` handles the flags correctly with or without the prefix.

LalatenduMohanty:
As discussed this will create more ways for users to screw up minishift start process, deferring it for the time being.

hferentschik:
Closing pull request

gbraad:
this is also blocked by #1175. Since the extra-flags could contain almost any content, checking and limiting the content of this would be important and needs more attention than what is in this PR.

PR #1174, opened 2017-07-25
Issue #1169 Return error when oc cluster up flag is not supported

### LINKED ISSUE #1169: Unsupported cluster up flags are not returned as error
See: https://github.com/minishift/minishift/blob/5c02dee1f8814d07dace81a11e594bdfbc79a1db/pkg/minishift/clusterup/clusterup.go#L69

The flag is checked for support, but it is not returned as error.

gbraad:
Fixes #1169

### DIFF
--- a/pkg/minishift/clusterup/clusterup.go
+++ b/pkg/minishift/clusterup/clusterup.go
@@ -66,7 +66,7 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 
 	for key, value := range clusterUpParams {
 		if !oc.SupportFlag(key, config.OcPath, runner) {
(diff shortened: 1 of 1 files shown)


pkg/minishift/clusterup/clusterup.go
```diff
@@ -66,15 +66,15 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 
 	for key, value := range clusterUpParams {
 		if !oc.SupportFlag(key, config.OcPath, runner) {
-			errors.New(fmt.Sprintf("Flag %s is not supported for oc version %s. Use 'openshift-version' flag to select a different version of OpenShift.", key, config.OpenShiftVersion))
+			return errors.New(fmt.Sprintf("Flag %s is not supported for oc version %s. Use 'openshift-version' flag to select a different version of OpenShift.", key, config.OpenShiftVersion))
```
praveenkumar: 👍 Good catch

Issue #1152, opened 2017-07-20
Describe the issue of filling up the live root filesystem


gbraad:
Due to the way we run the Minishift instance, it can happen that the `/` (root) feilsystem on the VM gets exhausted. This has been resolved for OpenShift is part of #355, but since people still install software #1076, this issue can still lead to problems.

On the ISO side, we have now implemented something that will show a banner when software is installed using `yum`: https://github.com/minishift/minishift-centos-iso/issues/116
However, as part of the message we show, we want to refer to the documentation that addresses this problem and explains why this is currently unsupported (without having to point to an issue).

Since we do not have a wiki, this has to be done as part of the Minishhift documentation?

hferentschik:
Troubleshooting guide?

gbraad:
I had the same idea of including it in the troubleshooting guide... as long as the can directly link to an anchor within that document as part of the URL in the banner:

```
Installing packages using yum is not supported. For more information, please check: http://docs.openshift.org/latest/minishift/troubleshooting#livecd-root
Continue [y/n]
```

gbraad:
The documentation/troubleshoooting entry will need to contain:

  * why this is unsupported
    * it exhausts the root filesystem of the LiveCD and can lock up the system
    * information provided by tools like `df` are unreliable and misleading about available space
  * what are the alternatives
    * use the persistent disk instead (/mnt/sda1)  
      task: maybe at a relocation of /opt -> /mnt/sda1/opt ?
    * use containers to perform development tasks

LalatenduMohanty:
As I mentioned in the other issue i.e. https://github.com/minishift/minishift-centos-iso/issues/116#issuecomment-316975519 

We should add little more context e.g. `Installing additional packages is not supported because it will exhaust the root file system (/) and will lock the virtual machine. For more information, see:` 

As this communicates the consequences of the action and hopefully create the curiosity of the user to read more about it.

LalatenduMohanty:
@gbraad I am fine with adding the content in troubleshooting guide.

LalatenduMohanty:
Fixed by https://github.com/minishift/minishift/commit/1b1a21a8b069bde499a1d4e48ce7d8d283302794

PR #1182, opened 2017-07-26
Issue #1152 troubleshooting entry about exceeding root filesystem overlay size

thatdocslady:
Fixes issue #1152 

This PR covers a troubleshooting entry where we warn users not to install packages or store large files in the Minishift VM root filesystems, and suggest workarounds.

Approved by @gbraad before submitting.

### DIFF
--- a/docs/source/using/troubleshooting.adoc
+++ b/docs/source/using/troubleshooting.adoc
@@ -230,3 +230,16 @@ If you try to use Minishift with Hyper-V using an external virtual switch while 
 Cause: Hyper-V networking might not route the network traffic in both directions properly when connected to a VPN.
 
 Workaround: Disconnect from the VPN and try again after stopping the VM from the Hyper-V manager.
+
+[[root-filesystem-exceeds-overlay-size]]
+== The root filesystem of the Minishift VM exceeds overlay size
+
+Installing additional packages or copying large files to the root filesystem of the Minishift VM might exceed the allocated overlay size and lock the Minishift VM.
+
+Cause: The Minishift VM root filesystem contains core packages that are configured to optimize running the Minishift VM and containers.
+The available storage on the root filesystem is determined by the overlay size, which is smaller than the total available storage.
+
+Workaround: Avoid installing packages or storing large files in the root filesystem of the Minishift VM.
+Instead, you can create a sub-directory in the *_mnt/sda1/_* persistent storage volume or define and mount xref:../using/host-folders.adoc#[host folders] that can share storage space between the host and the Minishift VM.
+
+If you want to perform development tasks inside the Minishift VM, it is recommended that you use containers, which are stored in persistent storage volumes, and reuse the xref:../using/docker-daemon.adoc#[Minishift Docker daemon].


docs/source/using/troubleshooting.adoc
```diff
@@ -230,3 +230,16 @@ If you try to use Minishift with Hyper-V using an external virtual switch while
 Cause: Hyper-V networking might not route the network traffic in both directions properly when connected to a VPN.
 
 Workaround: Disconnect from the VPN and try again after stopping the VM from the Hyper-V manager.
+
+[[root-filesystem-exceeds-overlay-size]]
+== The root filesystem of the Minishift VM exceeds overlay size
+
+Installing additional packages or copying large files to the root filesystem of the Minishift VM might exceed the allocated overlay size and lock the Minishift VM.
+
+Cause: The Minishift VM root filesystem contains core packages that are configured to optimize running the Minishift VM and the application containers.
```
LalatenduMohanty: It should be just containers as user might run diagnostics tools in containers. s/the application containers/containers/

docs/source/using/troubleshooting.adoc
```diff
@@ -230,3 +230,16 @@ If you try to use Minishift with Hyper-V using an external virtual switch while
...
+The available storage on the root filesystem is determined by the overlay size, which is smaller smaller than the total available storage.
+
+Workaround: Avoid installing packages or storing large files in the root filesystem of the Minishift VM.
+Instead, you can create a sub-directory in the *_mnt/sda1/_* persistent storage volume or define and mount xref:../using/host-folders.adoc#[host folders] that can share storage space between the host and the Minishift VM.
+
+If you want to perform development tasks inside the Minishift VM, it is recommended that you use containers, which are stored in persistent storage volumes, and reuse the xref:../using/docker-daemon.adoc#[Minishift Docker daemon].
```
gbraad: it might be slightly nicer to say "You can do this by re-using", instead of ", and".

docs/source/using/troubleshooting.adoc
```diff
@@ -230,3 +230,16 @@ If you try to use Minishift with Hyper-V using an external virtual switch while
...
 Workaround: Disconnect from the VPN and try again after stopping the VM from the Hyper-V manager.
+
+[[root-filesystem-exceeds-overlay-size]]
+== The root filesystem of the Minishift VM exceeds overlay size
+
+Installing additional packages or copying large files to the root filesystem of the Minishift VM might exceed the allocated overlay size and lock the Minishift VM.
+
+Cause: The Minishift VM root filesystem contains core packages that are configured to optimize running the Minishift VM and the application containers.
+The available storage on the root filesystem is determined by the overlay size, which is smaller smaller than the total available storage.
+
```
LalatenduMohanty: I find the use of overlay confusing. I checked dictionary but did not help much

docs/source/using/troubleshooting.adoc
thatdocslady: @gbraad I think that it changes the meaning, because storing files and installing packages might not necessarily directly link to development tasks. In general also we try to avoid saying "you can do *this*" or "for *this*" because it is ambiguous and not always clear what "this" is referring to.

docs/source/using/troubleshooting.adoc
```diff
@@ -230,3 +230,16 @@ If you try to use Minishift with Hyper-V using an external virtual switch while
...
 
 Workaround: Disconnect from the VPN and try again after stopping the VM from the Hyper-V manager.
+
+[[root-filesystem-exceeds-overlay-size]]
+== The root filesystem of the Minishift VM exceeds overlay size
+
+Installing additional packages or copying large files to the root filesystem of the Minishift VM might exceed the allocated overlay size and lock the Minishift VM.
+
+Cause: The Minishift VM root filesystem contains core packages that are configured to optimize running the Minishift VM and the application containers.
+The available storage on the root filesystem is determined by the overlay size, which is smaller smaller than the total available storage.
```
gbraad: twice smaller?

docs/source/using/troubleshooting.adoc
gbraad: ah, check... now I notice. There is a by-sentence due to the use of the comma.

thatdocslady: Haha reading your comment again @gbraad I understand which part of the sentence you're referring to, which actually reinforces my theory that it's dangerous to use *this* coz it's hard to tell what it refers to ;)

gbraad: Oh yeah, right. Read too quickly and missed a comma. focussed too much on the xref structure I guess. :-P

Issue #292, opened 2017-01-13
The latest compatible default ISO should be downloaded automatically


hferentschik:
ATM a fixed version of the default b2d ISO is downloaded at startup. It would make more sense to determine the latest compatible version using semantic versioning. This way the default ISO can be released independently by Minishift and directly consumed.

gbraad:
We could provide a version tag by publishing a JSON document on github-pages for instance, which can be retrieved as http://minishift.github.io/version Several projects use something similar, such as glide: https://glide.sh/version which is used from the http://glide.sh/get script to retrieve the latest stable version.

budhrg:
We have used a python script to achieve similar thing in [b2d-iso](https://github.com/minishift/minishift-b2d-iso/blob/master/Makefile#L10).

> We could provide a version tag by publishing a JSON document on github-pages for instance, which can be retrieved as http://minishift.github.io/version

I like the this idea too 👍

hferentschik:
The way it is done so far is via the GitHub API accessing the releases and using semantic versioning. Once we reach Minishift 1.0.0, we will anyways drop the alpha,beta,etc part (unless needed in some cases) and do "pure" semantic versioning.

praveenkumar:
Since now we reached the 1.0.0 milestone I think now we can think to fetch latest available iso for b2d ISO. let we adjust the milestone if we think 1.2.0 timeline is not good enough.

praveenkumar:
I am going to change milestone for this issue since PR closed and need more discussion around it.

gbraad:
We will also need to check from the provisioner based on the `os-release` as we need to be backwards compatible and warn if a version of a previously cached ISO could cause problems.

praveenkumar:
@minishift/minishift-dev If you go through the PR discussion which got closed because the way we want to implement this feature then you will find we are better off with this feature then putting effort. Please let me know if you think same or still agree to implement this feature as it supposes to be.

LalatenduMohanty:
@praveenkumar yes, lets go ahead with the plan.

LalatenduMohanty:
Not planning to fix this as we are going for a new solution for OpenShift 4.0

PR #1187, opened 2017-07-27
[WIP] Issue #292 Latest compatible default ISO should be downloaded automatically

praveenkumar:


pkg/minishift/util/fetch_iso_version.go
```diff
@@ -0,0 +1,59 @@
...
+)
+
+type VersionSpec struct {
+	Assets []struct {
+		Name               string `json:"name"`
+		BrowserDownloadURL string `json:"browser_download_url"`
+	} `json:"assets"`
+}
+
+func GetLatestISOVersion(isoName string) string {
```
hferentschik: That's too simplistic really. The idea is not just to get the latest. We should also inspect the version and treat it as semantic version. Then we should determine whether this version is compatible with current Minishift.

What if we release a new non compatible change to the ISO? A change which required changes in Minishift as well? Users which have a older version of Minishift should then not automatically get this new version. 

The latest version of the ISO which is compatible in semantic versioning terms should be downloaded.

Makefile
```diff
@@ -15,8 +15,6 @@
 # Various versions - Minishift, default OpenShift, default B2D ISO
 MINISHIFT_VERSION = 1.3.1
 OPENSHIFT_VERSION = v1.5.1
-B2D_ISO_VERSION = v1.0.2
-CENTOS_ISO_VERSION = v1.1.0
```
hferentschik: I am not sure we can/should completely remove the versions. We might need to some base version to be able to do semantic versioning.

cmd/minishift/cmd/start_test.go
```diff
@@ -361,11 +362,11 @@ func TestDetermineIsoUrl(t *testing.T) {
 		in  string
 		out string
 	}{
-		{"", constants.DefaultB2dIsoUrl},
-		{"b2d", constants.DefaultB2dIsoUrl},
-		{"B2D", constants.DefaultB2dIsoUrl},
-		{"centos", constants.DefaultCentOsIsoUrl},
-		{"CentOs", constants.DefaultCentOsIsoUrl},
+		{"", minishiftUtil.GetLatestISOVersion("b2d")},
```
hferentschik: Hmmm, not sure whether this a good idea.

pkg/minishift/util/fetch_iso_version.go
LalatenduMohanty: As described in the issue, we should check for compatibilty in semantic version.  The question should we check all releases starting from the latest to see which one is compatible and then download that?

Makefile
LalatenduMohanty: @hferentschik  Trying to understand your suggestion. Is it because we need a version to fall back in-case semantic version  check fails? Also what about the versions we put in documentation?

LalatenduMohanty: @hferentschik I get it know.  We need the base version i.e. 1.0.0 for CentOS ISO for semantic version comparison. But I am still not sure how we want to handle the versions of ISOs we put in docs.

praveenkumar:
I am putting back it as WIP and we will discuss when @hferentschik back.

cmd/minishift/cmd/start_test.go
gbraad: can you motivate why this is not a good idea?
you mean you rather want a fixed version?

cmd/minishift/cmd/start_test.go
```diff
@@ -361,11 +362,11 @@ func TestDetermineIsoUrl(t *testing.T) {
...
+		{"b2d", minishiftUtil.GetLatestISOVersion("b2d")},
+		{"B2D", minishiftUtil.GetLatestISOVersion("b2d")},
+		{"centos", minishiftUtil.GetLatestISOVersion("centos")},
+		{"CentOs", minishiftUtil.GetLatestISOVersion("centos")},
```
gbraad: `CentOs` is `CentOS` (as OS is an abbreviation for Operating System)

Note: just a test...

pkg/minishift/util/fetch_iso_version.go
gbraad: In that case we will always have an issue in a way... cached images, in that case, should also be ignored and force a download of a compatible image. 

But what if I redownload an older version of Minishift, eg. v1.6.0, while the latest and the greatest is Minishift v2.0. How can I be sure the image that is compatible with v1.6.0 is used? this should in that case JUST be a single version... ? (Because I can just download a point release which was marked as compatible... or can I?)

I think we need to be more precise about how to version the ISOs in the first place, if that is the case.

But, I am of the opinion that nothing in the image should REALLY be specific to minishift. Minishift should do the downgrade or feature check.

Example. If I use Minishift which can support hostfolders using NFS, I need to check first if the image is compatible with this functionality (see if `mount.nfs` is available). I should not have to rely on the image...

and why? because if we target a VM (or cloud image) that is NOT provided or provisioned by us, we should also be able to start and work.

I can image that the pre-flight checks in the future can be extended to perform feature checks, and record this for use of Minishift

```
-- Check target for support features ...
   Check if VM support CIFS mounts ... OK
   Check if VM support SSHFS mounts ... OK
   Check if VM support NFS mounts ... FAIL
```

cmd/minishift/cmd/start_test.go
hferentschik: I meant the change as a whole and in particular just retrieving the latest version based on the latest tag on the release page. 

We need more than this in order to keep control over compatibility. As mentioned in the issue, the idea is to be using semantic versioning.  

What we want to achieve is to introduce some sort of independency between Minishift release and ISO release. Aka, if we build Minishift against 1.1.1 and then we release we screed something up in the ISO and we release 1.1.2, Minishift should transparently download this version, without having to release a new Minishift version.

On the other hand, if we release a ISO 2.0.0 with some backward incompatible changes (potentially even requiring changes in the Minishift side), this ISO should not be downloaded automatically.

Makefile
hferentschik: So we would need something like

    B2D_ISO_BASE_VERSION = ~1		
    CENTOS_ISO_BASE_VERSION = ~1

This would instruct Miinishift to automatically download ISO >= 1, < 2

hferentschik: > But I am still not sure how we want to handle the versions of ISOs we put in docs.

I don't think we really need the concrete ISO version in the docs. ATM, it is only used once as an example. We don't need this dynamic really.

hferentschik:
I am going to close this pull request for now, since it for now falls short of the intend of #292. The idea is to have some sort of loose coupling between Minishift and ISO release, whilst at the same time still being able to control potential incompatibilities. 

Even though I think this would be a nice feature, it might not be the most important one. Given that implementing it would require considerably more effort than this pull request, we could also decide to defer this issue and move it into a later sprint.

Issue #1186, opened 2017-07-27
Rename of b2d image fails after completing download on Windows


abbr:
vbox shows minishift vm is running but cli hung.
![image](https://user-images.githubusercontent.com/6158277/28654700-2b069326-724b-11e7-90dc-23f00878cc07.png)

Edit by gbraad:
```
C:\Users\x>minishift start --vm-driver virtualbox
-- Installing default add-ons ... OOK
Starting local OpenShift cluster using 'virtualbox' hypervisor...
Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.0.2/minishift-b2d.iso
 40.00 MiB / 40.00 MiB {=============================================] 100.00% 0s
Error starting the VM: rename C:\Users\x\.minishift\cache\iso\minishift-b2d.iso.part C:\Users\x\.minishift\cache\iso\minishift-b2d.iso: The process cannot access the file because it is being used by another process.. Retrying.
Error starting the VM. Error creating machine: Error in driver during machine creation: C:\Program Files\Oracle\VirtualBox\VBoxManage.exe storageattach minishift --storagectl SATA --port 0 --device 0 --type dvddrive --medium C:\Users\x\.minishift\boot2docker.iso failed:
...
```

gbraad:
@praveenkumar seems the rename action on Windows needs to be checked for inconsistencies/race-conditions :-s

Similarly, someone mentioned seeing `.part` on #1180

praveenkumar:
@gbraad ah, I have win10 box and that works as expected. now have to find someone who have win7 to reproduce or check if rename logic works.

gbraad:
This is on Windows 10 as I can see from the version in the prompt.

gbraad:
The error likely happened at: https://github.com/minishift/minishift/commit/854d63c5d069cddb54b0ae9662938f0ada063b4f#diff-221fdad1b39f75fb2d0214c25944dcb3R284

praveenkumar:
@abbr  I think we know what causing this issue looks like os.Rename  is platform dependent and not working with windows. There are already  lot of discussion around it on github. We need to find a fix for it  apply asap.  -

- golang/go#8914 
- golang/go#14527

gbraad:
How about using a mutex around the action, to prevent us from renaming or continuing, before the file download or rename has finished...

praveenkumar:
[...]

PR #1188, opened 2017-07-27
Issue #1186 Close file descriptor to make rename work on Windows

praveenkumar:


pkg/minikube/cluster/cluster.go
```diff
@@ -282,7 +283,19 @@ func (m *MachineConfig) CacheMinikubeISOFromURL() error {
...
-	if err = os.Rename(tmpISOFile.Name(), out.Name()); err != nil {
+	if err = renameFile(tmpISOFile, out); err != nil {
+		return nil
+	}
+
+	return nil
+}
+
+func renameFile(oldFile, newFile *os.File) error {
+	// File descriptor need to be closed otherwise it will throw error
```
LalatenduMohanty: I am not sure why we need another function to close the file descriptor. Whether we close it in the same function or in another function it is the same thing.

praveenkumar: @LalatenduMohanty This function can be used by other functions if required in future. Putting everything in  a single function should be avoided where it make sense.

gbraad: I also think it is on the safer side, as the function forces the closes on function entry. +1

gbraad:
CentOS CI still did not pick this build up... :-s

pkg/minikube/cluster/cluster.go
LalatenduMohanty: @praveenkumar At this moment the function is a private function, so limited to current package.  Also we are trying not to add more code to minikube package.  I am not sure what should be the right approach here. 
I think you are thinking that we should reuse the function in Minishift as it is easy to do the mistake what we we did last time.

gbraad: refactor from private to public is a minor refactor. It should not block this from being integrated...

Issue #1193, opened 2017-07-28
Configuration flag --insecure-registry is not working with minishift start


LalatenduMohanty:
Refer: https://github.com/minishift/minishift/issues/873#issuecomment-318543025

```
$ minishift version
minishift v1.3.1+f4900b07

✔ ~
11:32 $ minishift delete
Deleting the Minishift VM...
Minishift VM deleted.

✔ ~
11:32 $ minishift start --insecure-registry test-registry:5000 --show-libmachine-logs
Starting local OpenShift cluster using 'virtualbox' hypervisor...
(minishift) Trying to access option engine-install-url which does not exist
(minishift) THIS ***WILL*** CAUSE UNEXPECTED BEHAVIOR
(minishift) Type assertion did not go smoothly to string for key engine-install-url
(minishift) Trying to access option swarm-master which does not exist
(minishift) THIS ***WILL*** CAUSE UNEXPECTED BEHAVIOR
(minishift) Type assertion did not go smoothly to bool for key swarm-master
(minishift) Trying to access option swarm-host which does not exist
(minishift) THIS ***WILL*** CAUSE UNEXPECTED BEHAVIOR
Running pre-create checks...
(minishift) Type assertion did not go smoothly to string for key swarm-host
(minishift) Trying to access option swarm-discovery which does not exist
(minishift) THIS ***WILL*** CAUSE UNEXPECTED BEHAVIOR
(minishift) Type assertion did not go smoothly to string for key swarm-discovery
(minishift) Type assertion did not go smoothly to bool for key virtualbox-host-dns-resolver
(minishift) Type assertion did not go smoothly to bool for key virtualbox-hostonly-no-dhcp
(minishift) Type assertion did not go smoothly to bool for key virtualbox-no-share
(minishift) Type assertion did not go smoothly to bool for key virtualbox-no-dns-proxy
(minishift) Type assertion did not go smoothly to bool for key virtualbox-no-vtx-check
Creating machine...
(minishift) Downloading /Users/chad/.minishift/cache/boot2docker.iso from file:///Users/chad/.minishift/cache/iso/minishift-b2d.iso...
(minishift) Creating VirtualBox VM...
(minishift) Creating SSH key...
(minishift) Starting the VM...
(minishift) Check network to re-create if needed...
(minishift) Waiting for an IP...
Waiting for machine to be running, this may take a few minutes...
Detecting operating system of created instance...
Waiting for SSH to be available...
Detecting the provisioner...
[...]

PR #1196, opened 2017-07-28
[WIP] Issue #1195 Add integration test for flags

### LINKED ISSUE #1195: Add missing integration tests for docker related flags
To check for:

  * insecure regsitry #1193 
  * setting optional features
    * docker env variables

gbraad:
Fixes #1195, tests for #1193

Test with
```
$ make integration GODOG_OPTS="-tags flags -format pretty"
```

### DIFF
--- a/test/integration/features/coolstore.feature
+++ b/test/integration/features/coolstore.feature
@@ -1,4 +1,4 @@
-@coolstore
+@coolstore @centos-only
 Feature: Cool Store
   In order to test Minishift under load
   I need to setup a test environment
--- a/test/integration/features/flags.feature
+++ b/test/integration/features/flags.feature
@@ -44,7 +44,7 @@ Feature: Flags
       export "BAZ=BAT"
       """
 
-  @rhel-only
+  @centos-only
   Scenario: User is able to get value of the docker environment variables
     Given Minishift has state "Running"
      When executing "minishift ssh cat /etc/systemd/system/docker.service.d/10-machine.conf" succeeds
@@ -66,7 +66,7 @@ Feature: Flags
       --dns=8.8.8.8
       """
 
-  @rhel-only
+  @centos-only
     Scenario: User is able to get the value of docker optional parameters
     Given Minishift has state "Running"
      When executing "minishift ssh cat /etc/systemd/system/docker.service.d/10-machine.conf" succeeds
--- a/test/integration/features/flags.feature
+++ b/test/integration/features/flags.feature
@@ -1,0 +1,82 @@
+@flags
+Feature: Flags
+  As a user I can perform advanced operations
+  such as starting Minishift with optional features
+
+  Scenario: Set configuration options for Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift config set docker-opt dns=8.8.8.8" succeeds
+     Then executing "minishift config get docker-opt" succeeds
+      And stdout should contain
+      """
+      [dns=8.8.8.8]
+      """
+     When executing "minishift config set docker-env BAZ=BAT" succeeds
+     Then executing "minishift config get docker-env" succeeds
+      And stdout should contain
+      """
+      [BAZ=BAT]
+      """
+
+  Scenario: Starting Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start --insecure-registry test-registry:5000 --docker-env=FOO=BAR" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: User is able to get value of the insecure registry
+    Given Minishift has state "Running"
+     When executing "minishift ssh docker info" succeeds
+     Then stdout should contain
+      """
+      test-registry:5000
+      """
+
+  @b2d-only
+  Scenario: User is able to get value of the docker environment variables
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
+     Then stdout should contain
+      """
+      export "FOO=BAR"
+      """
+      And stdout should not contain
+      """
+      export "BAZ=BAT"
+      """
+
+  @rhel-only
+  Scenario: User is able to get value of the docker environment variables
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /etc/systemd/system/docker.service.d/10-machine.conf" succeeds
+     Then stdout should contain
+      """
+      Environment=FOO=BAR
+      """
+      And stdout should not contain
+      """
+      Environment=BAZ=BAT
+      """
+
+  @b2d-only
+  Scenario: User is able to get the value of docker optional parameters
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
+     Then stdout should contain
+      """
+      --dns=8.8.8.8
+      """
+
+  @rhel-only
+    Scenario: User is able to get the value of docker optional parameters
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /etc/systemd/system/docker.service.d/10-machine.conf" succeeds
+     Then stdout should contain
+      """
+      --dns=8.8.8.8
+      """
+
+  Scenario: Deleting Minishift
+    Given Minishift has state "Running"
+     When executing "minishift delete" succeeds
+     Then Minishift should have state "Does Not Exist"
+
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -66,9 +66,23 @@ func TestMain(m *testing.M) {
 	}
 	runner := util.MinishiftRunner{CommandPath: minishiftBinary}
 	if runner.IsCDK() {
-		godogTags += "~minishift-only ~b2d-only"
+		godogTags += "~minishift-only&&~b2d-only&&~centos-only"
 	} else {
-		godogTags += "~cdk-only ~rhel-only"
+		switch os.Getenv("MINISHIFT_ISO_URL") {
+		case "b2d":
+			godogTags += "~cdk-only&&~centos-only&&~rhel-only"
+			fmt.Println("Test run using b2d iso image.")
+		case "centos":
+			godogTags += "~cdk-only&&~b2d-only&&~rhel-only"
+			fmt.Println("Test run using centos iso image.")
+		case "":
+			godogTags += "~cdk-only&&~centos-only&&~rhel-only"
+			os.Setenv("MINISHIFT_ISO_URL", "b2d")
+			fmt.Println("Test run using b2d iso image.")
+		default:
+			godogTags += "~cdk-only"
+			fmt.Println("Test run using custom path for iso image. WARNING: Only default image names 'b2d' and 'centos' are automatically supported by Minishift test suite. When using images with custom ISO_URL value, please add '~centos-only' or '~b2d-only' tags appropriately.")
+		}
 	}
 
 	status := godog.RunWithOptions("minishift", func(s *godog.Suite) {
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -66,9 +66,9 @@ func TestMain(m *testing.M) {
 	}
 	runner := util.MinishiftRunner{CommandPath: minishiftBinary}
 	if runner.IsCDK() {
-		godogTags += "~minishift-only"
+		godogTags += "~minishift-only ~b2d-only"
 	} else {
-		godogTags += "~cdk-only"
+		godogTags += "~cdk-only ~rhel-only"
 	}
 
 	status := godog.RunWithOptions("minishift", func(s *godog.Suite) {
@@ -180,6 +180,7 @@ func FeatureContext(s *godog.Suite) {
 		if runner.IsCDK() {
 			runner.CDKSetup()
 		}
+
 		fmt.Println("Running Integration test in:", testDir)
 		fmt.Println("Using binary:", minishiftBinary)
 	})


test/integration/features/flags.feature
```diff
@@ -0,0 +1,35 @@
+@flags
+Feature: Flags
+  As a user I can perform advanced operations
+  such as starting Minishift with optional features
+
+  Scenario: Starting Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start --insecure-registry test-registry:5000 --docker-env=FOO=BAR --docker-env=BAZ=BAT" succeeds
```
praveenkumar: @gbraad let's add `--docker-opt` flag also in the same scenario.

praveenkumar:
@gbraad also we should add test when those flags are set using `minishift config set` so we have both side of things tested properly.

gbraad:
@praveenkumar added
@agajdosi @amitkrout is this something you could appreciate?
/cc: @LalatenduMohanty

test/integration/features/flags.feature
```diff
@@ -0,0 +1,58 @@
...
+    Given Minishift has state "Running"
+     When executing "minishift ssh docker info" succeeds
+     Then stdout should contain
+      """
+      test-registry:5000
+      """
+
+  Scenario: User is able to get value of the docker environment variables
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
```
agajdosi: This will probably fail on CentOS iso.

test/integration/features/flags.feature
```diff
@@ -0,0 +1,58 @@
...
+      export "FOO=BAR"
+      """
+      And stdout should not contain
+      """
+      export "BAZ=BAT"
+      """
+
+  Scenario: User is able to get the value of docker optional paramters
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
```
agajdosi: I am wondering if we could implement something like cdk-only and minishift-only tags - just for b2d and centos. We could then have this scenario tagged `@b2d-only` and add another scenario focused on centos with `@centos-only`.

test/integration/features/flags.feature
gbraad: yep, it will... but there is no general way to do this, AFAICR. for CentOS you likely need to cat the systemd unit file.

test/integration/features/flags.feature
gbraad: Possible...

note: typo spotted... parameters

gbraad: There is an issue with the image based tag, as we set this in the feature itself when starting minishift.

However, you can set a scenario variable from the code... I made it specifically for such a purpose. In the `BeforeScenario` we can do:

```go
minishift.SetVariable("ISO-URL", "b2d")`
godogTags += "~centos-only ~rhel-only"
```

And in the scenario we write the step as:

```cucumber
   Scenario: Starting Minishift
      When executing "minishift start --insecure-registry test-registry:5000 --docker-env=FOO=BAR --iso-url $(ISO-URL)"
```

agajdosi: If I understand it correctly, this would force the feature to be run with b2d image as coolstore is using centos.
> There is an issue with the image based tag, as we set this in the feature itself when starting minishift.

Somewhere we are setting it, somewhere we are not - like in basic.feature I think. For basic.feature and flag.feature we could detect the iso and set tags properly, for features like coolstore which requires specific image, detection will fail and for example b2d-only tag will be selected wrongly - but as this feature can and will be run only with centos, then there will not be any b2d-only tags included which will cause exclusion of some scenarios. (Not sure if I am understandable here.)

agajdosi: I am thinking, if we have a lot of feature files and I want to test complete test suite with b2d and upstream minishift, it should run only those features which are valid for b2d and minishift. So coolstore should be excluded and all other centos only tests. Then after b2d testing I want to test centos, in this case coolstore will run.

WDYT?

gbraad:
https://github.com/minishift/minishift/pull/1196/files/e8f3932f1619533fe5b0a0619ac08bc7fca5f2c7#diff-3571a3e1a7c1419570519496d69bfbe9

test/integration/features/flags.feature
```diff
@@ -0,0 +1,60 @@
...
+      """
+
+  @b2d-only
+  Scenario: User is able to get the value of docker optional parameters
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
+     Then stdout should contain
+      """
+      --dns=8.8.8.8
+      """
```
praveenkumar: @gbraad we are also adding that for RHEL right because it is not available at this location.

test/integration/features/flags.feature
```diff
@@ -0,0 +1,60 @@
...
+      And stdout should contain
+      """
+      [dns=8.8.8.8]
+      """
+     When executing "minishift config set docker-env BAZ=BAT" succeeds
+     Then executing "minishift config get docker-env" succeeds
+      And stdout should contain
+      """
+      [BAZ=BAT]
+      """
```
praveenkumar: We also need to do start minishift after set config and then check if flags are part of instance like we are doing for command options. Only verifying with `config get` might not necessary tell us if instance also have same options.

test/integration/features/flags.feature
gbraad: @agajdosi will look into the issue with using `b2d-only`, `centos-only` and maybe `rhel-only` for the images.

test/integration/features/flags.feature
gbraad: this is why in the test we are querying this from the actual instance.

doing the `config get` is to emphasize the settings we have made and information got persisted by our code. (this is to find issues with WriteConfig early)

gbraad:
To make tests sepcific to an image, we probably need to do something like the following. In `BeforeScenario` we can do:

```go
//switch on os.GetEnv("MINISHIFT_ISO_URL"), and when b2d:
minishift.SetVariable("ISO-URL", "b2d")  
godogTags += "~centos-only ~rhel-only"
```

And in the scenario we write the step as:

```cucumber
   Scenario: Starting Minishift
      When executing "minishift start --iso-url $(ISO-URL) --other-params"
```

test/integration/features/flags.feature
gbraad: @praveenkumar for CentOS we need to examine the systemd unit file, right... Or is there any other way to get this information?

gbraad: This seems to be stored as: `/etc/systemd/system/docker.service.d/10-machine.conf`

gbraad:
Added `rhel-only` and `b2d-only` to tests

test/integration/features/flags.feature
```diff
@@ -0,0 +1,78 @@
...
+     When executing "minishift ssh docker info" succeeds
+     Then stdout should contain
+      """
+      test-registry:5000
+      """
+
+  @rhel-only
+  Scenario: User is able to get value of the docker environment variables
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
```
praveenkumar: @gbraad there is no `/var/lib/boot2docker/profile` for rhel based ISO.

gbraad: The wrong file got pushed as it dropped a b2d test, and had a centos-only test (save issue).

test/integration/features/flags.feature
praveenkumar: > this is why in the test we are querying this from the actual instance.

@gbraad where we are doing it? In below scenario I can only see explicit options are tested.

agajdosi:
@gbraad @praveenkumar Will start working on it now, I think we need 3 tags as we have 3 images - even though CentOS and RHEL are quite the same, I feel that mention all 3 will be much clear and explicit.

I am thinking about something like this:
```go
if runner.IsCDK() {
		//RHEL is the only supported image for CDK
		godogTags += "~minishift-only&&~b2d-only&&~centos-only"
	} else {
		switch os.Getenv("MINISHIFT_ISO_URL") {
		case "b2d":
			godogTags += "~cdk-only&&~centos-only&&~rhel-only"
		case "centos":
			godogTags += "~cdk-only&&~b2d-only&&~rhel-only"
		case "":
			godogTags += "~cdk-only&&~centos-only&&~rhel-only"
			os.Setenv("MINISHIFT_ISO_URL", "b2d")
		default:
			fmt.Println("Only default image names 'b2d' and 'centos' are supported by Minishift test suite. When using images with custom ISO_URL value, please add '~centos-only' or '~b2d-only' tags appropriately.)
		}
	}
```

Going to test it on your branch.

test/integration/integration_test.go
```diff
@@ -66,9 +66,9 @@ func TestMain(m *testing.M) {
 	}
 	runner := util.MinishiftRunner{CommandPath: minishiftBinary}
 	if runner.IsCDK() {
-		godogTags += "~minishift-only"
+		godogTags += "~minishift-only ~b2d-only"
 	} else {
-		godogTags += "~cdk-only"
+		godogTags += "~cdk-only ~rhel-only"
```
LalatenduMohanty: What about making it centos-only as for upstream users it will be easy to test with CentOS ISO and RHEL ISO is not available for upstream.

gbraad:
How about downstream interpretation?

test/integration/integration_test.go
praveenkumar: @gbraad you forgot to add same `centos-only` here :)

agajdosi: there is also need for logical and: `&&`, however I will make a PR to this branch soon fixing this

agajdosi:
@gbraad I have created a PR into your branch: https://github.com/gbraad/minishift/pull/1.

About behaviour: When running on minishift binary without MINISHIFT_ISO_URL specified or when it is equal to "b2d", then if you run all tests, basic and flags will run fine, but coolstore won't run at all as it is completely meant only for centos. However when you run all tests with MINISHIFT_ISO_URL="centos" then coolstore will run just fine. If one needs to specify custo path to iso, then integration tests will warn her that there is a need to handle centos-only and b2d-only tags manually.

gbraad:
Thanks. I think with the change provided by @agajdosi we are able to merge...

gbraad:
retest this please

test/integration/integration_test.go
```diff
@@ -66,9 +66,23 @@ func TestMain(m *testing.M) {
 	}
 	runner := util.MinishiftRunner{CommandPath: minishiftBinary}
 	if runner.IsCDK() {
-		godogTags += "~minishift-only"
+		godogTags += "~minishift-only&&~b2d-only&&~centos-only"
 	} else {
-		godogTags += "~cdk-only"
+		switch os.Getenv("MINISHIFT_ISO_URL") {
+		case "b2d":
+			godogTags += "~cdk-only&&~centos-only&&~rhel-only"
```
praveenkumar: @agajdosi, @gbraad  I am curious which test cases are only for rhel, my understanding is test cases are only for b2d, centos and cdk. centos and cdk is only there to make sure we have upstream/downstream test cases differently. If we have CentOS it will auto cover all rhel specific test cases.

gbraad:
this comes from the fact we mix terms and names:

We have tests based on the application (can we register), and based on the
image (where are files located)
It is Minishift or CDK: minishift-only (don't have) or cdk-only.
So the images we have are b2d-only, centos-only and rhel-only.

Calling something cdk-only when we are testing images doesn't make much
sense.
But we need to be precise, we shouldn't use centos-only tests when it
concerns an RHEL-based image
I'd rather see them tagged with both, ...

gbraad:
retest this please

gbraad:
retest this please

test/integration/features/coolstore.feature
```diff
@@ -1,4 +1,4 @@
-@coolstore
+@coolstore @centos-only
```
hferentschik: Why 'cantos-only'? What's this about?

agajdosi:
@praveenkumar We agreed that we will add only whole feature files for CDK in downstream and that we won't add downstream tests in github, so for upstream, there is probably no need to take care of ~cdk-only and ~rhel-only. But this can change.

Bad thing is that currently RHEL will not run centos-only. One possibility can be to mention centos-only and rhel-only tags on those scenarios, which are the same for both (could be nearly 99%?) as @gbraad suggests. When scenario will be tagged just centos-only without rhel-only, then it will not be run with RHEL image. But I need to take a look into Gherkin/Cucumber and Godog, which logical operations are available and if this is possible, did not realized before.

test/integration/integration_test.go
hferentschik: I am also getting a bit confused about how we are going with these tags/annotations. There is the CDK non CDK split and the centos-only now. Also why would we exclude 'centos-only' from the CDK tests?

Is the distinction between centos and rhel really needed?

gbraad:
Issues with AppVeyor continue

test/integration/features/flags.feature
```diff
@@ -0,0 +1,82 @@
...
+     When executing "minishift ssh docker info" succeeds
+     Then stdout should contain
+      """
+      test-registry:5000
+      """
+
+  @b2d-only
+  Scenario: User is able to get value of the docker environment variables
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
```
hferentschik: Instead splitting this into two scenarios and introducing all this confusion around b2d-only and cantos-only, could we not add a new custom step? 

    Then Docker environment contains FOO=BAR
    And  Docker environment contains BAZ=BAT

The details on where and how to look for these settings we deal with under the hood. In this case one only needs one scenario.

gbraad: This means you would hide the actual ISO image information because somewhere we would still need to record the information about which image is used.

How often do we need this that it really needs a new function?
Saying `cdk-only` is not enough... as mentioned earlier, we are mixing terms to begin with: `b2d` and `cdk`...

we do need to find something to deal with testing between CentOS and B2D in a proper way, but I do not think hiding this in functions is a good way either.

BTW, the outcome should have been:

```
Then Docker environment contains FOO=BAR
And  Docker environment does not contain BAZ=BAT
```

test/integration/features/coolstore.feature
agajdosi: @hferentschik Idea was to not drive the the information about image by `--iso-url centos` but instead use environmental variable and prevent it from running when boot2docker is used.

test/integration/integration_test.go
agajdosi: @hferentschik @praveenkumar If need to differentiate between centos and rhel is not most likely happen, then we can remove the rhel-only tag and handle it as centos. It is true that we can exclude some centos scenarios by adding minishift-only, this will prevent it from running in downstream and in fact will behave as rhel-only.

agajdosi:
```go
	if runner.IsCDK() {
		godogTags += "~minishift-only&&~b2d-only"
	} else {
		switch os.Getenv("MINISHIFT_ISO_URL") {
		case "b2d":
			godogTags += "~cdk-only&&~centos-only"
			fmt.Println("Test run using b2d iso image.")
		case "centos":
			godogTags += "~cdk-only&&~b2d-only"
			fmt.Println("Test run using centos iso image.")
		case "":
			godogTags += "~cdk-only&&~centos-only"
			os.Setenv("MINISHIFT_ISO_URL", "b2d")
			fmt.Println("Test run using b2d iso image.")
		default:
			godogTags += "~cdk-only"
			fmt.Println("Test run using custom path for iso image. WARNING: Only default image names 'b2d' and 'centos' are automatically supported by Minishift test suite. When using images with custom ISO_URL value, please add '~centos-only' or '~b2d-only' tags appropriately.")
		}
```

test/integration/features/flags.feature
agajdosi: +1 for not hiding it. I think Cucumber also can serve as some kind of documentation, which describes how the product behaves. If one is used to Cucumber, then she can look into our features and see how Minishift should work. I already used it this way few times. If we have centos and b2d as two basic images which differ in some cases, then these cases are part of behaviour of Minishift and should be "documented" visibly.

hferentschik: You are not hiding things. You are hiding implementation details. What is it you want to assert, that a given environment setting is for example applied to the Docker daemon. You really don't care how this is handled specifically by the ISO, right?

hferentschik: > This means you would hide the actual ISO image information because somewhere we would still need to record the information about which image is used.

why? You would not have to. One way would be to get the release version of the ISO as part of the step implementation. This allows to differentiate between various ISOs. Depending on the release version, I can then make a decision on where to look.

hferentschik:
> About behaviour: When running on minishift binary without MINISHIFT_ISO_URL specified or when it is equal to "b2d", then if you run all tests, basic and flags will run fine, but coolstore won't run at all as it is completely meant only for centos.

Is it? Why? AFAIR, it worked with B2D ISO as well. What's the issue?

> I will later add and take a follow up issue to update documentation for integration tests.

Required doc updates should be part of the pull request.

hferentschik:
I always was working under the assumption that we want to keep the integration tests as much ISO agnostic as possible. If we need to rely on ISO specific things, it should be handled in the step implementation. 

The CDK / Minishift distinction was introduced to make it easier for adding CDK only tests. My hope was that we don't need further distinctions.

agajdosi:
> Required doc updates should be part of the pull request.

@hferentschik ack

> Is it? Why? AFAIR, it worked with B2D ISO as well. What's the issue?

There is `When executing "minishift start --memory 4096 --iso-url centos" succeeds` in coolstore.feature on line 16, so I believed it is here because it won't run properly with b2d.

hferentschik:
> "minishift start --memory 4096 --iso-url centos"

Ahh, right. @gbraad why is this again? Why centos? Is it literally not working for b2d image?

gbraad:
It can run on b2d. actually with less memory

gbraad:
We used CentOS based on the performance concerns we received

gbraad:
The other reason was to test the shorturl form, as basic did not use any image reference

hferentschik:
> It can run on b2d. actually with less memory

Right, that's what I thought. So really, the test should not specify the ISO at all, but let the ISO be controlled by the environment variable. It also allows to run the test against CDK.

gbraad:
@agajdosi will you continue working on this?

agajdosi:
@gbraad Yes. As result of discussion I will remove the centos-only from coolstore for sure. The question of using tags vs custom step for search of Docker env is still open. Can we make a resolution on which way we want to get? @hferentschik

hferentschik:
> As result of discussion I will remove the centos-only from coolstore for sure

+1

> The question of using tags vs custom step for search of Docker env is still open. Can we make a resolution on which way we want to get

I still think this is a implementation detail. We always said the ISO should be pluggable and the intrinsics a black box for the users. From a feature description I am just interested in saying "then the Docker daemon has env variable 'foo' set with value 'bar'". I don't want to leak the specifics on how to assert this into the feature description. 

I am also concerned that introducing these tags adds another level of complexity on how to drive the tests and which combinations to use where and when. Also adding these tags might lead to more tests than necessary using ISO specifics.

gbraad:
> The question of using tags vs custom step for search of Docker env is still open.

Let's perform these checks in code: but keep the logic as short as possible by using small re-usable checks:

```
func CheckForEnvironmentVariable(envName) {
  if isTestingOnCentOS() || isTestingOnRHEL{
    checkEnvCentOSRHEL(envName)
  }
  if isTestingOnB2D() {
    checkEnvB2D(envName)
  }
}
```
Note: purely psuedo to get my point across

hferentschik:
> Let's perform these checks in code: but keep the logic as short as possible by using small re-usable checks:

+1 small and re-usable is music in my ears :-)

hferentschik:
Superseded by pull request #1272

Issue #660, opened 2017-04-04
Exit from "minishift ssh" console returns error


agajdosi:
**Environment:**
OS: RHEL 7.3 Workstation+Server, Windows 7, Fedora 24
Minishift: _1.0.0-beta.5
CDK: 3.0.0-beta.3

**Description:**
I connect to console of minishift machine with `minishift ssh` command. When I exit this ssh instance and last command executed here returned error, then also this exit will throw an error:
```
[hudson@localhost ~]$ minishift ssh
Last login: Tue Apr  4 09:24:43 2017 from 192.168.42.1
[docker@minishift ~]$ mistake
-bash: mistake: command not found
[docker@minishift ~]$ exit
logout
E0404 09:31:11.723209    5180 ssh.go:38] Cannot establish SSH connection to the VM:  exit status 127
```
This is just a cosmetic as I did not find any problem after - for example I am able to connect again with `minishift ssh`.

jorgemoralespou:
Same here on mac and minishift 1.0.0.rc.1

hferentschik:
I think this a case of bad error message and in fact bad error handling. The error message originates here -https://github.com/minishift/minishift/blob/7974f5edc9697c6b878794aa9f98d9d0fe668588/cmd/minishift/cmd/ssh.go#L39

Obviously the message is misleading, since we actually had a connection. The other problem is that per default the ssh session returns with the exit command of the last command executed prior exiting the the ssh session. In case of typing a unknown command the exit code ($?) will be 127 and hence the error we are seeing here. 

If possible we need to distinguish between not being able to make the ssh connection (in which case the current message would make sense) and having an actual session which ends with a non 0 exit code. I would think having an interactive session one should not report any error based on the commands run on the VM.

jhcook:
I have no idea if this is related, but I am having the same issue, and it appears it is an erroneous use of a volume. The / is filling up and I can no longer connect. 

https://github.com/minishift/minishift/issues/355

gbraad:
@jhcook your message is unrelated, as this happens because of the SSH daemon stopping to work.

gbraad:
```
$ minishift ssh "mount.nfs -v"
usage: mount.nfs remotetarget dir [-rvVwfnsh] [-o nfsoptions]
options:
        -r              Mount file system readonly
        -v              Verbose
        -V              Print version
        -w              Mount file system read-write
        -f              Fake mount, do not actually mount
        -n              Do not update /etc/mtab
[...]

PR #1198, opened 2017-07-28
Issue #660 Do not return error when closing interactive SSH session

gbraad:
Do not return an error when closing an interactive session.

See discussion on #660

### DIFF
--- a/cmd/minishift/cmd/ssh.go
+++ b/cmd/minishift/cmd/ssh.go
@@ -26,6 +26,10 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	SshCommunicationError = "exit status 255"
+)
+
 // sshCmd represents the docker-machine ssh command
 var sshCmd = &cobra.Command{
 	Use:   "ssh [-- COMMAND]",
@@ -34,9 +38,13 @@ var sshCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
+
 		err := cluster.CreateSSHShell(api, args)
 		if err != nil {
-			atexit.ExitWithMessage(1, fmt.Sprintf("Cannot establish SSH connection to the VM: %s", err.Error()))
+			if err.Error() == SshCommunicationError {
+				atexit.ExitWithMessage(1, fmt.Sprintf("Cannot establish SSH connection to the VM: %s", err.Error()))
+			}
+			atexit.ExitWithMessage(1, fmt.Sprintf("Command failed: %s", err.Error()))
 		}
 	},
 }
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -57,7 +57,8 @@ var (
 )
 
 const (
-	fileScheme = "file"
+	fileScheme            = "file"
+	SshCommunicationError = "exit status 255"
 )
 
 //This init function is used to set the logtostderr variable to false so that INFO level log info does not clutter the CLI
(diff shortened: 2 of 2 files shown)


gbraad:
Currently working on an improvement ...

cmd/minishift/cmd/ssh.go
```diff
@@ -34,9 +34,10 @@ var sshCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
+
 		err := cluster.CreateSSHShell(api, args)
 		if err != nil {
-			atexit.ExitWithMessage(1, fmt.Sprintf("Cannot establish SSH connection to the VM: %s", err.Error()))
+			atexit.ExitWithMessage(1, fmt.Sprintf("Command failed or cannot establish SSH connection to the VM: %s", err.Error()))
```
LalatenduMohanty: @gbraad There is not way we can separate the error messages i.e. `command failed `and `Cannot establish SSH connection to the VM:` ?

pkg/minikube/cluster/cluster.go
```diff
@@ -595,7 +595,12 @@ func CreateSSHShell(api libmachine.API, args []string) error {
 		return err
 	}
 
-	return client.Shell(args...)
+	err = client.Shell(args...)
+	// do not fail if interactive and able to connect
+	if err != nil && len(args) == 0 && err.Error() != "exit status 255" {
+		return nil
+	}
```
LalatenduMohanty: Works for me.

cmd/minishift/cmd/ssh.go
LalatenduMohanty: I mean when args == 0 that means it a non interactive session, otherwise it is an interactive session. So if args != 0 we can say `ommand failed`. WDYT?

gbraad: length of args == 0 => interactive, yes... but we do not know until inspection of the error if the connection failed, or the command failed.

gbraad: well, if you can except inspecting the error code, I can split it... as "exit status 255" seems to be a connection failed.

cmd/minishift/cmd/ssh.go
```diff
@@ -34,9 +38,13 @@ var sshCmd = &cobra.Command{
 	Run: func(cmd *cobra.Command, args []string) {
 		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 		defer api.Close()
+
 		err := cluster.CreateSSHShell(api, args)
 		if err != nil {
-			atexit.ExitWithMessage(1, fmt.Sprintf("Cannot establish SSH connection to the VM: %s", err.Error()))
+			if err.Error() == SshCommunicationError {
```
LalatenduMohanty: I had a feeling that return code 255 is not enough for finding out error SSH connection. But did some testing around it and @gbraad pointed me to https://man.openbsd.org/ssh#EXIT_STATUS . Usecases where VM's SSH service is down or VM is shutdown the error is proper (The error when VM is shutdown is coming from cluster.go). Looks good to me.

hferentschik: Are we not handling the issue twice now? Here and in _ CreateSSHShell_? Why do we need to introduce and handle _SshCommunicationError_ in the command as well, given the change in _ CreateSSHShell_?

gbraad: Here we check for the actual outcome to decide the message.
`255` is an ssh error, while anything else is a command error.

pkg/minikube/cluster/cluster.go
```diff
@@ -595,7 +596,12 @@ func CreateSSHShell(api libmachine.API, args []string) error {
 		return err
 	}
 
-	return client.Shell(args...)
+	err = client.Shell(args...)
+	// do not fail if interactive and able to connect
+	if err != nil && len(args) == 0 && err.Error() != SshCommunicationError {
```
gbraad: @hferentschik 
If we do not do this:

```
$ minishift ssh
[docker@minishift ~]$ fsd
-bash: fsd: command not found
[docker@minishift ~]$ exit
logout
Command failed: exit status 127
```

hferentschik: I get this part of the change.

cmd/minishift/cmd/ssh.go
hferentschik: Get it now, thanks.

hferentschik:
Rebased and merged. Thanks.

PR #1201, opened 2017-07-31
Issue #1200 Refactor registration code

### LINKED ISSUE #1200: Possible error in stop/unregistration code
Related to #1121. When using a CentOS-based image, I have to use `--skip-unregistration` when SSH is not able to connect. In the case of a CentOS image, registration and deregistration should not checked for.

```
$ minishift ssh "sudo systemctl stop sshd"
$ minishift stop
Error unregistering the VM: ssh command error:
command : sudo cat /etc/os-release
err     : exit status 255
output  : 
$ minishift start
The 'minishift' VM is already running.
$ minishift stop --skip-unregistration
Cluster stopped.
```

gbraad:
Fixes #1200 

### Before
```
$ minishift ssh "sudo systemctl stop sshd"
$ minishift stop
Error unregistering the VM: ssh command error:
command : sudo cat /etc/os-release
err     : exit status 255
output  : 
```

### After
```
$ minishift ssh "sudo systemctl stop sshd"
$ minishift delete
Deleting the Minishift VM...
Minishift VM deleted.
```

### DIFF
--- a/cmd/minishift/cmd/registration.go
+++ b/cmd/minishift/cmd/registration.go
@@ -1,0 +1,81 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/docker/machine/libmachine"
+	"github.com/docker/machine/libmachine/log"
+	"github.com/minishift/minishift/pkg/minishift/cluster"
+	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
+	pkgUtil "github.com/minishift/minishift/pkg/util"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+)
+
+var (
+	skipRegistration   bool
+	skipUnRegistration bool
+)
+
+func isRegistered() bool {
+	return minishiftConfig.InstanceConfig.IsRegistered
+}
+
+func setRegistered(state bool) {
+	minishiftConfig.InstanceConfig.IsRegistered = state
+	minishiftConfig.InstanceConfig.Write()
+}
+
+func registerHost(libMachineClient *libmachine.Client) {
+	if skipRegistration {
+		log.Debug("Skipping registration due to enabled '--skip-registration' flag")
+		return
+	}
+
+	if wasSuccessful, err := cluster.Register(libMachineClient); err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error to register VM: %v", err))
+	} else {
+		// else, we set the IsRegistered state according to result
+		setRegistered(wasSuccessful)
+	}
+}
+
+func unregisterHost(api libmachine.API, allowToSkipUnregistration bool) {
+	if allowToSkipUnregistration && skipUnRegistration {
+		log.Debug("Skipping unregistration due to enabled '--skip-unregistration' flag")
+		return
+	}
+
+	if !isRegistered() {
+		return
+	}
+
+	if wasSuccessful, err := cluster.UnRegister(api); err != nil || !wasSuccessful {
+		handleFailedRegistrationAction()
+	}
+
+	// if continued, we remove the IsRegistered state
+	setRegistered(false)
+}
+
+func handleFailedRegistrationAction() {
+	hasConfirmed := pkgUtil.AskForConfirmation("Current Minishift VM is registered, but unregistration failed.")
+	if !hasConfirmed {
+		atexit.ExitWithMessage(0, fmt.Sprintln("Aborted."))
+	}
+}
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -30,7 +30,6 @@ import (
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/host"
-	"github.com/docker/machine/libmachine/log"
 	"github.com/docker/machine/libmachine/provision"
 	"github.com/golang/glog"
 	"github.com/minishift/minishift/cmd/minishift/cmd/addon"
@@ -39,6 +38,7 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/cache"
+	minishiftCluster "github.com/minishift/minishift/pkg/minishift/cluster"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
 	"github.com/minishift/minishift/pkg/minishift/docker"
@@ -484,10 +484,9 @@ func initClusterUpFlags() *flag.FlagSet {
 // initSubscriptionManagerFlags create the CLI flags which are needed for VM registration
 func initSubscriptionManagerFlags() *flag.FlagSet {
 	subscriptionManagerFlagSet := flag.NewFlagSet(commandName, flag.ContinueOnError)
-
 	subscriptionManagerFlagSet.String(configCmd.Username.Name, "", "Username for the virtual machine registration.")
 	subscriptionManagerFlagSet.String(configCmd.Password.Name, "", "Password for the virtual machine registration.")
-	subscriptionManagerFlagSet.Bool(configCmd.SkipRegistration.Name, false, "Skip the virtual machine registration.")
+	subscriptionManagerFlagSet.BoolVar(&skipRegistration, configCmd.SkipRegistration.Name, false, "Skip the virtual machine registration.")
 
 	return subscriptionManagerFlagSet
 }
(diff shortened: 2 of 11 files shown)


budhrg:
@gbraad commit message is using `deregistration` instead of `unregistration` :smile:

cmd/minishift/cmd/stop.go
```diff
@@ -62,11 +62,15 @@ func runStop(cmd *cobra.Command, args []string) {
 	fmt.Println("Stopping local OpenShift cluster...")
 
 	if !skipUnRegistration {
-		if _, err := cluster.UnRegister(api); err != nil {
+		supportRegistration, err := cluster.UnRegister(api)
```
budhrg: How about `registrationSupport(ed)` ?

pkg/minikube/cluster/cluster.go
```diff
@@ -152,15 +152,15 @@ func UnRegister(api libmachine.API) (bool, error) {
...
-		if err != nil {
-			return false, err
-		}
-	} else {
-		return true, nil
+	if !drivers.MachineInState(host.Driver, state.Running)() {
+		return true, fmt.Errorf("Unable to verify registration. VM not in running state")
+	}
+
+	supportUnregistration, err := registration.UnregisterHostVM(host, RegistrationParameters)
```
budhrg: Same here. `unregistrationSupport(ed)`.

gbraad:
retest this please

cmd/minishift/cmd/stop.go
gbraad: this is consistent with the previous function we modified

gbraad:
rebased on #660 to make sure SSH code is usable

gbraad:
```
$ minishift ssh "sudo systemctl stop sshd"
$ minishift delete
Deleting the Minishift VM...
Minishift VM deleted.
```

cmd/minishift/cmd/stop.go
```diff
@@ -61,12 +61,16 @@ func runStop(cmd *cobra.Command, args []string) {
 
 	fmt.Println("Stopping local OpenShift cluster...")
 
-	if !skipUnRegistration {
-		if _, err := cluster.UnRegister(api); err != nil {
+	if !skipUnRegistration && minishiftConfig.InstanceConfig.IsRegistered {
+		supportUnregistration, err := cluster.UnRegister(api)
```
hferentschik: _supportUnregistration_  is an unfortunate name imo. It implies to me that whatever you call "checks" whether unregistration is supported. Judging by name, I would not expect that unregistration actually occurs. However, this is the case. Why not 'unregistered' or 'unregistrationSuccessful'. This also makes clear why we then clean up the instance config state.

hferentschik: I think the problem is that we try to do two things here. Determine whether the VM supports un-registration as well as doing the actual un-registration.

pkg/minikube/cluster/cluster.go
```diff
@@ -152,15 +152,15 @@ func UnRegister(api libmachine.API) (bool, error) {
...
 
-	if drivers.MachineInState(host.Driver, state.Running)() {
-		err := registration.UnregisterHostVM(host, RegistrationParameters)
-		if err != nil {
-			return false, err
-		}
-	} else {
-		return true, nil
+	if !drivers.MachineInState(host.Driver, state.Running)() {
+		return true, fmt.Errorf("Unable to verify registration. VM not in running state")
```
hferentschik: So why _true_? The VM is not running, so how do you know that it "supports" unregistration.

cmd/minishift/cmd/stop.go
hferentschik: Also, the whole point of 'minishiftConfig.InstanceConfig.IsRegistered' was to be able to handle the un-regristration case, for the case that the VM cannot be accessed. So assuming all data is in a consistent state the check for _minishiftConfig.InstanceConfig.IsRegistered_ already tells us that the VM supports (un)registration. So I think all that's needed is to handle the error a bit differently. 

We might want to align with the _delete_ command where in the case of a failing un-registration the user gets prompted on whether to processed or not. See https://github.com/minishift/minishift/blob/2335e6ac4aa252e7046070f2f8a5ca76675f12c3/cmd/minishift/cmd/delete.go#L54

pkg/minikube/cluster/cluster.go
gbraad: We can also not return false. It is a tristate (true, false, unknown) since it does not allow only an error to be returned, the true is the least error-prone as we prevent to continue when the information is unclear.

cmd/minishift/cmd/stop.go
gbraad: If `minishiftConfig.InstanceConfig.IsRegistered` was the whole point, it seems we neglected to actually check the state of this variable. This is why this change is suggested.

Without checking this, it means that images like B2D and CentOS will always be requested about `supportRegistration`. If for some reason, deployment did not succeed in a working VM (no IP for instance), it becomes problematic for the user to stop (or delete) the VM. This should never happen...

hferentschik: > If minishiftConfig.InstanceConfig.IsRegistered was the whole point, it seems we neglected to actually check the state of this variable.

That's correct. It should have been checked. Similar to the delete _command_.

> This is why this change is suggested/issue got filed.

Sure. I am not saying there is no issue here. I am just saying that if _minishiftConfig.InstanceConfig.IsRegistered_ is _true_, we already know that _supportUnregistration_ is true.

gbraad:
Will consider to refactor some of this code, as I am not happy how it lives inside `cluster.go` and the possible duplication between `delete` and `stop` command.

cmd/minishift/cmd/stop.go
praveenkumar: >If minishiftConfig.InstanceConfig.IsRegistered was the whole point, it seems we neglected to actually check the state of this variable.

@gbraad right it was missed from my sight when implementing `minishiftConfig.InstanceConfig.IsRegistered` flag and should be handled same as delete command which only look for that variable.

hferentschik:
> as I am not happy how it lives inside cluster.go and the possible duplication between delete and stop command.

+1

gbraad:
@praveenkumar @hferentschik open to comments...

gbraad:
```
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
Red Hat Developers or Red Hat Subscription Management (RHSM) username: gbraad
Red Hat Developers or Red Hat Subscription Management (RHSM) password: [HIDDEN]
Registration successful [43.9s]
   Pinging 8.8.8.8 ... OK
-- Checking available disk space ... 1% OK
```

Do I need to change the lines:

```
Red Hat Developers or Red Hat Subscription Management (RHSM) username: gbraad
Red Hat Developers or Red Hat Subscription Management (RHSM) password: [HIDDEN]
Registration successful [43.9s]
```
to represent the same output as the rest?

cmd/minishift/cmd/registration.go
```diff
@@ -0,0 +1,81 @@
...
+	// if continued, we remove the IsRegistered state
+	setRegistered(false)
+}
+
+func handleFailedRegistrationAction() {
+	hasConfirmed := pkgUtil.AskForConfirmation("Current Minishift VM is registered, but unregistration failed.")
+	if !hasConfirmed {
+		atexit.ExitWithMessage(0, fmt.Sprintln("Aborted."))
+	}
+}
```
praveenkumar: much better 👍

cmd/minishift/cmd/start.go
```diff
@@ -588,10 +587,10 @@ func validateOpenshiftVersion() {
...
 
 func setSubscriptionManagerParameters() {
-	cluster.RegistrationParameters.Username = viper.GetString(configCmd.Username.Name)
-	cluster.RegistrationParameters.Password = viper.GetString(configCmd.Password.Name)
-	cluster.RegistrationParameters.GetUsernameInteractive = getUsernameInteractive
-	cluster.RegistrationParameters.GetPasswordInteractive = getPasswordInteractive
+	minishiftCluster.RegistrationParameters.Username = viper.GetString(configCmd.Username.Name)
+	minishiftCluster.RegistrationParameters.Password = viper.GetString(configCmd.Password.Name)
+	minishiftCluster.RegistrationParameters.GetUsernameInteractive = getUsernameInteractive
+	minishiftCluster.RegistrationParameters.GetPasswordInteractive = getPasswordInteractive
```
praveenkumar: Nice no more depend on minikube/cluster 👍

pkg/minishift/cluster/registration.go
```diff
@@ -0,0 +1,56 @@
...
+
+// Register returns true if successfully registered
+func Register(api libmachine.API) (bool, error) {
+	return doRegistrationAction(api, registration.RegisterHostVM)
+}
+
+// UnRegister returns true if successfully unregistered
+func UnRegister(api libmachine.API) (bool, error) {
+	return doRegistrationAction(api, registration.UnregisterHostVM)
+}
```
praveenkumar: 👍

cmd/minishift/cmd/start.go
gbraad: Pretty sure we can now even reduce the code slightly further... But some other time maybe?

cmd/minishift/cmd/registration.go
```diff
@@ -0,0 +1,81 @@
...
+	pkgUtil "github.com/minishift/minishift/pkg/util"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+)
+
+var (
+	skipRegistration   bool
+	skipUnRegistration bool
+)
+
+func isRegistered() bool {
```
hferentschik: 👍

cmd/minishift/cmd/registration.go
```diff
@@ -0,0 +1,81 @@
...
+	return minishiftConfig.InstanceConfig.IsRegistered
+}
+
+func setRegistered(state bool) {
```
hferentschik: 👍

cmd/minishift/cmd/registration.go
```diff
@@ -0,0 +1,81 @@
...
+func registerHost(libMachineClient *libmachine.Client) {
+	if skipRegistration {
+		log.Debug("Skipping registration due to enabled '--skip-registration' flag")
+		return
+	}
+
+	if wasSuccessful, err := cluster.Register(libMachineClient); err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error to register VM: %v", err))
+	} else {
+		// else, we set the IsRegistered state according to result
```
hferentschik: :+1

Issue #1192, opened 2017-07-28
Add check for minishift version into integration tests


agajdosi:
Recently we had an issue in downstream with incorrect version format. To cover this we need to add scenario for minishift version, which will need to implement regexp support into code.

Syntax in Cucumber for these new regexp steps could be for example:
```
Then stdout should match "___"
And stderr should not match "___"
```

hferentschik:
For what it's worth, we cover this in unit test as well. Of course it does to hurt to add it on an integration test level as well.

agajdosi:
@hferentschik Not sure if this will catch something in upstream really. Main plan here is to introduce regexp support and add version check later in downstream, there we had problem with version string in RC2 last week.

hferentschik:
Merged via pull request #1205

Issue #1204, opened 2017-07-31
Minishift version output contains extra new line


agajdosi:
Problem is in `minishift/cmd/minishift/cmd/version.go - line 38`.

Actual behaviour:
```
[agajdosi@dhcp-24-156 minishift]$ minishift version
minishift v1.3.1+f4900b07

[agajdosi@dhcp-24-156 minishift]$
```

Expected behaviour:
```
[agajdosi@dhcp-24-156 minishift]$ minishift version
minishift v1.3.1+f4900b07
[agajdosi@dhcp-24-156 minishift]$
```

Line 38 is now:
```
fmt.Println(fmt.Sprintf("minishift v%s+%s\n", version.GetMinishiftVersion(), version.GetCommitSha()))
```
should be:
```
fmt.Println(fmt.Sprintf("minishift v%s+%s", version.GetMinishiftVersion(), version.GetCommitSha()))
```
or maybe just
```
fmt.Printf("minishift v%s+%s\n", version.GetMinishiftVersion(), version.GetCommitSha())
```

agajdosi:
Will fix this tomorrow if it is not taken until then - blocks #1192.

hferentschik:
Any particular reason you want to remove the newline? Why do you think is should not be there. I'd be careful changing anything here w/o double checking with the JBIDE integration.

hferentschik:
> blocks #1192.

I am intrigued, how can this block a test? Assuming for a moment that the new line is indeed intentional, are you saying you could not write a test which verifies this format? If the only reason for this change is driven by a test, I think it is not worth doing.

agajdosi:
@hferentschik

> Any particular reason you want to remove the newline?

It is not consistent, any other command in Minishift is not using extra new line.

> are you saying you could not write a test which verifies this format?

I have no problem with writing it, it is already done. I just thought that it would be better to fix it as I saw it just as a bug and then create PR for the test, so we do not have to change the test day after it is originally pulled.

If the only reason for it is driven by a test on JBIDE side I am ok with closing this as it is actually a feature. I can then send the PR for test of version which will count with current format.

hferentschik:
Merged via pull request #1207

PR #1205, opened 2017-07-31
Issue #1192 Add test for version command

agajdosi:
Also adds support for regexp matching of stdout, stderr etc in format:
```
Then stdout should match "^minishift v\d\.\d\.\d\+[\w\d]{8}\n$"
And stderr should not match
"""
^minishift v\d\.\d\.\d\+[\w\d]{8}\n$
"""
```

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -142,7 +142,7 @@ func FeatureContext(s *godog.Suite) {
 		minishift.rolloutServicesSuccessfully)
 
 	// steps to verify `stdout`, `stderr` and `exitcode` of commands executed
-	s.Step(`([^"]*) should contain ([^"]*)$`,
+	s.Step(`([^"]*) should contain "([^"]*)"$`,
 		commandReturnShouldContain)
 	s.Step(`([^"]*) should not contain "([^"]*)"$`,
 		commandReturnShouldNotContain)
@@ -160,6 +160,15 @@ func FeatureContext(s *godog.Suite) {
 		commandReturnShouldNotBeEmpty)
 	s.Step(`([^"]*) should be valid ([^"]*)$`,
 		shouldBeInValidFormat)
+	// steps for matching stdout, stderr or exitcode with regular expression
+	s.Step(`([^"]*) should match "([^"]*)"$`,
+		commandReturnShouldMatchRegex)
+	s.Step(`([^"]*) should not match "([^"]*)"$`,
+		commandReturnShouldNotMatchRegex)
+	s.Step(`([^"]*) should match$`,
+		commandReturnShouldMatchRegexContent)
+	s.Step(`([^"]*) should not match$`,
+		commandReturnShouldNotMatchRegexContent)
 
 	// step for HTTP requests for minishift web console
 	s.Step(`(body|status code) of HTTP request to "([^"]*)" (?:|at "([^"]*)" )(contains|is equal to) "([^"]*)"$`,
@@ -243,19 +252,17 @@ func getConfigValue(configPath string, keyPath string) (string, error) {
 }
 
 func matchConfigValue(configPath string, condition string, keyPath string, expectedValue string) error {
-	expectedRegexp, err := regexp.Compile(expectedValue)
-	if err != nil {
-		return fmt.Errorf("Expected value must be a valid regular expression statement: ", err)
-	}
-
 	keyValue, err := getConfigValue(configPath, keyPath)
 	if err != nil {
 		return err
 	}
 
-	if (condition == "contains") && !expectedRegexp.MatchString(keyValue) {
+	matches, err := performRegexMatch(expectedValue, keyValue)
+	if err != nil {
+		return err
+	} else if (condition == "contains") && !matches {
 		return fmt.Errorf("For key '%s' config contains unexpected value '%s'", keyPath, keyValue)
-	} else if (condition == "does not contain") && expectedRegexp.MatchString(keyValue) {
+	} else if (condition == "does not contain") && matches {
 		return fmt.Errorf("For key '%s' config contains value '%s', which it should not contain", keyPath, keyValue)
 	}
 
@@ -279,7 +286,7 @@ func checkConfigKey(configPath string, condition string, keyPath string) error {
 
 func compareExpectedWithActualContains(expected string, actual string) error {
 	if !strings.Contains(actual, expected) {
-		return fmt.Errorf("Output did not match. Expected: %s, Actual: %s", expected, actual)
+		return fmt.Errorf("Output did not match. Expected: '%s', Actual: '%s'", expected, actual)
 	}
 
 	return nil
@@ -287,7 +294,7 @@ func compareExpectedWithActualContains(expected string, actual string) error {
 
 func compareExpectedWithActualNotContains(notexpected string, actual string) error {
 	if strings.Contains(actual, notexpected) {
-		return fmt.Errorf("Output did match. Not expected: %s, Actual: %s", notexpected, actual)
+		return fmt.Errorf("Output did match. Not expected: '%s', Actual: '%s'", notexpected, actual)
 	}
 
 	return nil
@@ -295,7 +302,7 @@ func compareExpectedWithActualNotContains(notexpected string, actual string) err
 
 func compareExpectedWithActualEquals(expected string, actual string) error {
 	if actual != expected {
-		return fmt.Errorf("Output did not match. Expected: %s, Actual: %s", expected, actual)
+		return fmt.Errorf("Output did not match. Expected: '%s', Actual: '%s'", expected, actual)
 	}
 
 	return nil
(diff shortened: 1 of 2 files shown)


agajdosi:
Tagged WIP as there is currently an issue #1204 with `minishift version`. After fix I will edit the matching step and this will be ready.

I am adding this test as separate feature file as I would like to cover all commands in future. This is quite short and could be part of basic.feature, but as part of bigger plan I think it could make sense.

agajdosi:
Or maybe we can go for behaviour - for example behaviour of minishift / its feature that it can show version information about itself, oc, docker etc.

test/integration/features/cmd/version.feature
```diff
@@ -0,0 +1,7 @@
+@minishift-only @command
+Feature: Version command
+Command "minishift version" shows version of minishift binary.
+
+  Scenario: Setup-cdk installs and enables specific addons
```
hferentschik: I guess this is a cut&paste error. Let's have a correct scenario description here.

test/integration/features/cmd/version.feature
```diff
@@ -0,0 +1,7 @@
...
+      When executing "minishift version" succeeds
+      Then stdout should match "^minishift v\d\.\d\.\d\+[\w\d]{7}\n$"
```
hferentschik: FYI - https://github.com/minishift/minishift/blob/d3c8dd05c51357651337cb7011a909b1c3452b2f/cmd/minishift/cmd/version_test.go#L34-L34

test/integration/features/cmd/version.feature
```diff
@@ -0,0 +1,7 @@
+@minishift-only @command
```
hferentschik: why @command and why in a sub-directory _cmd_? _basic_ and _coolstore_ are also running a "bunch" of commands. Is the plan to create a feature for each command individually? The problem I see here is that a lot of commands will require a running Minishift instance. Provisioning a Minishift instance per command will take probably too much time. The version command is a bit special here since it requires no running VM. 

Either way, unless we have a concrete plan with the _cmd_ sub-directory I would argue that there is no need yet to create new directories.

test/integration/integration_test.go
```diff
@@ -279,31 +288,57 @@ func checkConfigKey(configPath string, condition string, keyPath string) error {
 
 func compareExpectedWithActualContains(expected string, actual string) error {
 	if !strings.Contains(actual, expected) {
-		return fmt.Errorf("Output did not match. Expected: %s, Actual: %s", expected, actual)
+		return fmt.Errorf("Output did not match. Expected: '%s', Actual: '%s'", expected, actual)
```
hferentschik: Personally I find _errors.New_ more concise than _fmt.Errorf_, but the end result is probably the same.

+1 for the quotes around '%s' though. I had more than one occasion where having quotes helps to distinguish cases where I was not sure whether something was not working in general or whether the actual string was empty.

test/integration/integration_test.go
```diff
@@ -394,6 +429,22 @@ func variableShouldNotBeEmpty(variableName string) error {
...
+func commandReturnShouldMatch(commandField string, expected string) error {
+	return compareExpectedWithActualMatches(expected, selectFieldFromLastOutput(commandField))
+}
+
+func commandReturnShouldNotMatch(commandField string, notexpected string) error {
+	return compareExpectedWithActualNotMatches(notexpected, selectFieldFromLastOutput(commandField))
+}
+
+func commandReturnShouldMatchContent(commandField string, expected *gherkin.DocString) error {
+	return compareExpectedWithActualMatches(expected.Content, selectFieldFromLastOutput(commandField))
```
hferentschik: Is the content version using regexp confirmed to work? Does creating a regular expression across multiple new lines work? Aka, does something like this work:

     Then stdout should contain
     """
     - anyuid         :  .*   P(0)
     - admin-user     :  .*  P(0)
     """

In some languages this would not work out of the box. Not sure atm how Go's regexp support handles this.

test/integration/features/cmd/version.feature
gbraad: I suggest to make it part of the naming, instead of a subfolder. Folders can always be introduced at a later stage

test/integration/integration_test.go
```diff
@@ -161,6 +161,15 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`([^"]*) should be valid ([^"]*)$`,
 		shouldBeInValidFormat)
 
+	s.Step(`([^"]*) should match "([^"]*)"$`,
```
gbraad: add comment to indicate these are used with regex, as term 'match' is very generic and can be mistaken as equals.
or include 'expression' in the string or function name (I prefer function name)

test/integration/integration_test.go
```diff
@@ -279,31 +288,57 @@ func checkConfigKey(configPath string, condition string, keyPath string) error {
...
+	}
+
+	if !expectedRegexp.MatchString(actual) {
+		return fmt.Errorf("Output did not match. Expected: '%s', Actual: '%s'", expected, actual)
+	}
+
+	return nil
+}
+
+func compareExpectedWithActualNotMatches(notexpected string, actual string) error {
```
gbraad: Duplication of code... the difference between `!expectedRegexp` and `notexpectedRegexp` is so minimal, so please create a function that can be reused:

  * it compiles the regex
  * it compares

In that case `compareExpectedWithActualMatches` and `compareExpectedWithActualNotMatches` re-use functionality and become much shorter.

test/integration/features/cmd/version.feature
agajdosi: > Is the plan to create a feature for each command individually?

@hferentschik Plan is to create features which will focus more on commands and edge cases of their usage than on some kind of story as coolstore and basic do. However for version command there is not much to investigate.

> The problem I see here is that a lot of commands will require a running Minishift instance.

Yes, this bothers me, version is specific and can be alone in small feature file, for other commands there will be need to group them I think, take two or three commands which are not interfering and group them. Or maybe we could go for one command per feature to bring clarity - overall it would take a lot of time but still it would be faster than manual testing we are doing now. Would like to have 3.1.0 as last release where majority of testing did persons.

@gbraad +1 going to do the renaming.

test/integration/features/cmd/version.feature
agajdosi: @hferentschik Yes, thanks for pointing out. So small file and even though I missed that...

test/integration/integration_test.go
agajdosi: @hferentschik Yes, tried it and fortunately it seems that it works in golang just fine, have not find a problem yet, tested with:

```gherkin
  Scenario: Test
    When executing "minishift addons list"
    Then stdout should match "^- .* : .* P\(0\)\n- .* : .* P\(0\)\n- .* : .* P\(0\)\n$"
    Then stdout should match
    """
    ^- .* : .* P\(0\)
    - .* : .* P\(0\)
    - .* : .* P\(0\)\n$
    """
``` 

Test passes.

test/integration/integration_test.go
agajdosi: @gbraad Thank for feedback, going to implement it. Also thinking if I can take code from regexp matching for config files, will try it.

test/integration/integration_test.go
agajdosi: @hferentschik Glad you like '%s'. I looked onto errors.New vs fmt.Errorf topic, if I understand it correctly errors.New does not support formatting, but if you do not need it then it saves one extra call of fmt.Sprintf(), fmt.Errorf supports formatting, which is needed here. Went through integration_test.go, util.go and minishift.go in test folder and found one fmt.Errorf() call, which do not use formatting and can be turned into more effective errors.New(). Going to fix it.

test/integration/features/cmd/version.feature
agajdosi: @hferentschik Useful, thanks. However I am not sure why there is `{7-8}`, should not --short flag give seven characters as sha result?

test/integration/integration_test.go
gbraad: I do not think it is necessary to prematurely optimize. this one extra call internally is not going to kill us, and especially not in a test situation. consistency and readability are more important.

Besides, most of the optimizations are done by the compiler... never be the compiler yourself unless performance is really an issue.

agajdosi: @gbraad Consistency, we have circa 7 errors now which do not need formatting and these are using errors.New(), just one has fmt.Errorf().

gbraad: Original tests had 3x errors.New(), I do not see a need to modify them.

test/integration/minishift.go
```diff
@@ -214,7 +214,7 @@ func (m *Minishift) rolloutServicesSuccessfully(servicesToCheck string) error {
 	}
 
 	if !success {
-		return fmt.Errorf("Not all successfully rolled out")
+		return errors.New("Not all successfully rolled out")
```
gbraad: Eh, why?

gbraad: This should actually be changed to keep records of the returned value to determine which service did NOT successfully deploy.

agajdosi: Ack

test/integration/integration_test.go
hferentschik: I agree that there is no need for optimisation. I also agree that we don't really have to change it. I just said, _errors.New_ communicates better the intend imo. Even w/o knowing anything about Go, I can understand what _errors.New_ does. _fmt.Errorf_ is less obvious imo. Personally I had to look up the implementation/docs the first time I saw this function called. 

Also, in the production code we definitely use more _errors.New_. I have not seen many (if any)  _fmt.Errorf_.

Either way, happy to keep it.

gbraad:
I think I started to use for.Errorf as the godog uses this AFAIR. ;-P

-- 

   [ Doing Open Source Matters ]

test/integration/integration_test.go
agajdosi: > I just said, errors.New communicates better the intend imo.

Agree, but if I need to use `%s`, then it would look like this:
```
errors.New(fmt.Sprintf("This happened: %s, just terrible!", err))
```

hferentschik:
Rebased and merged. Added a commit to make sure this test is run on CentOS CI.

Issue #126, opened 2016-10-19
Ability to create multiple Minishift  clusters


jorgemoralespou:
Would it be possible to have different named clusters, so I can start/stop the cluster I want?

It would be nice to have a workflow that could enable this:

```
minishift start my-example-cluster
minishift stop

minishift start my-minimal-cluster
minishift stop

minishift start my-example-cluster
minishift stop

minishift delete my-example-cluster
```

LalatenduMohanty:
Good idea. But as per my understanding it wont be easy or trivial. So the question is what is the priority of this feature.

jorgemoralespou:
Why can't be easy? Don't have to be running at the same time, but as long as you give the machine a name you can create multiple machines.

jimmidyson:
If it's so easy then I wait for your contribution :-b

It's definitely doable, just need to refactor stuff a bit. I'm sure this would be useful upstream as well btw.

LalatenduMohanty:
@jorgemoralespou I think thats good news. If you can give the steps, I can try to put in minishift.

budhrg:
+1 for the idea @jorgemoralespou

jorgemoralespou:
@jimmidyson I would contribute if I knew go. 
The only thing I can do is blog about my thoughts and how I would like this to work. I have it working with oc cluster up although not the tool. Just a workaround for me for now. 

Since I know you've already read it. I leave it here for others:
http://jorgemoral.es/2016/10/developing-locally-with-openshift/

jimmidyson:
I really love that post - you've explained the use cases for each of your requirements really clearly, thank you. One thing we've talked about is trying to keep minishift ux as close to minikube's as possible. With that in mind, would you be able to raise an issue in https://github.com/kubernetes/minikube/issues with a similar discussion?

jimmidyson:
Next step will be to break down your post into separate issues & then we'll be able to start working towards them. Help to raise those issues, discuss the details & review the implementations would be awesome contributions - & there were you saying you couldn't contribute because it's written in go 😜

jorgemoralespou:
For reference: https://github.com/kubernetes/minikube/issues/886

jorgemoralespou:
Also, I've seen minikube supports addons, which is similar to the plugins we do support in [oc-cluster -wrapper](https://github.com/openshift-evangelists/oc-cluster-wrapper/tree/plugins). Does minishift support addons?

hferentschik:
[...]

PR #1206, opened 2017-08-01
[WIP] Issue #126 Introducing profile feature to manage multiple instances of Minishift

LalatenduMohanty:
Create multiple instances of Minishift. Have not added tests yet.  The PR still does not address the issues where we wanted the ~/.minishift/config/config.json  isolated for each instance. 

### DIFF
--- a/cmd/minishift/cmd/delete.go
+++ b/cmd/minishift/cmd/delete.go
@@ -21,10 +21,12 @@ import (
 	"os"
 
 	"github.com/docker/machine/libmachine"
+	cmdProfile "github.com/minishift/minishift/cmd/minishift/cmd/profile"
 	"github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
+	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
 	"github.com/minishift/minishift/pkg/util/filehelper"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
@@ -46,7 +48,6 @@ var (
 func runDelete(cmd *cobra.Command, args []string) {
 	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
 	defer api.Close()
-
 	util.ExitIfUndefined(api, constants.MachineName)
 
 	// Unregistration, do not allow to be skipped
@@ -68,6 +69,7 @@ func runDelete(cmd *cobra.Command, args []string) {
 	}
 
 	removeInstanceConfigs()
+	removeProfileFromConfig()
 
 	fmt.Println("Minishift VM deleted.")
 }
@@ -94,6 +96,16 @@ func removeInstanceConfigs() {
 	if exists {
 		if err := os.Remove(constants.KubeConfigPath); err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error deleting '%s'", constants.KubeConfigPath))
+		}
+	}
+}
+
+func removeProfileFromConfig() {
+	name := cmdProfile.GetProfileName()
+	if name != "" {
+		err := profileActions.RemoveProfile(name)
+		if err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprintf("%s", err))
 		}
 	}
 }
--- a/cmd/minishift/cmd/oc_env.go
+++ b/cmd/minishift/cmd/oc_env.go
@@ -18,17 +18,20 @@ package cmd
 
 import (
 	"fmt"
+
 	"github.com/minishift/minishift/pkg/minikube/constants"
 
+	"os"
+	"path/filepath"
+	"text/template"
+
 	"github.com/docker/machine/libmachine"
+	cmdProfile "github.com/minishift/minishift/cmd/minishift/cmd/profile"
 	"github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/minishift/config"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/minishift/minishift/pkg/util/shell"
 	"github.com/spf13/cobra"
-	"os"
-	"path/filepath"
-	"text/template"
 )
 
 const (
@@ -42,12 +45,19 @@ type OcShellConfig struct {
 }
 
 func getOcShellConfig(ocPath, forcedShell string) (*OcShellConfig, error) {
+	var cmdLine string
+
 	userShell, err := shell.GetShell(forcedShell)
 	if err != nil {
 		return nil, err
 	}
 
-	cmdLine := "minishift oc-env"
+	if cmdProfile.ProfileExists() {
+		cmdLine = "minishift oc-env --profile " + cmdProfile.GetProfileName()
+	} else {
+		cmdLine = "minishift oc-env"
+	}
+
 	shellCfg := &OcShellConfig{
 		OcDirPath: filepath.Dir(ocPath),
 		UsageHint: shell.GenerateUsageHint(userShell, cmdLine),
@@ -87,7 +97,6 @@ var ocEnvCmd = &cobra.Command{
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error running the oc-env command: %s", err.Error()))
 		}
-
 		executeOcTemplateStdout(shellCfg)
 	},
 }
--- a/cmd/minishift/cmd/profile/profile.go
+++ b/cmd/minishift/cmd/profile/profile.go
@@ -1,0 +1,59 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package profile
+
+import (
+	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
+	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var ProfileCmd = &cobra.Command{
+	Use:     "profile SUBCOMMAND [flags]",
+	Aliases: []string{"instance"},
+	Short:   "Manages Minishift profiles.",
+	Long:    "Manages Minishift profile. Use the sub-commands to list, set profile",
+	Run: func(cmd *cobra.Command, args []string) {
+		cmd.Help()
+	},
+}
+
+//Checks viper or allinstanceconfg to say if profile is defined.
+func ProfileExists() bool {
+	if viper.GetString("profile") != "" {
+		return true
+	} else if minishiftConfig.AllInstancesConfig != nil {
+		if profileActions.GetActiveProfile() != "" {
+			return true
+		}
+	}
+	return false
+}
+
+//Returns the machine name after taking in to account if --profile or
+//profile information available in allinstance config
+func GetProfileName() string {
+	activeProfile := profileActions.GetActiveProfile()
+	if viper.GetString("profile") != "" {
+		return viper.GetString("profile")
+	} else if activeProfile != "" {
+		return activeProfile
+	} else {
+		return ""
+	}
+}
--- a/cmd/minishift/cmd/profile/profile_list.go
+++ b/cmd/minishift/cmd/profile/profile_list.go
@@ -1,0 +1,50 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package profile
+
+import (
+	"fmt"
+
+	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
+	"github.com/spf13/cobra"
+)
+
+var profileListCmd = &cobra.Command{
+	Use:   "list",
+	Short: "Lists profiles",
+	Long:  "Lists existing profiles for Minishift",
+	Run: func(cmd *cobra.Command, args []string) {
+		profiles := profileActions.GetProfileNameList()
+		if len(profiles) == 0 {
+			fmt.Println("There are no profiles defined")
+		} else {
+			displayProfiles(profiles)
+		}
+	},
+}
+
+//List existing profile information
+//TODO: Name, Status (running,stopped etc)
+func displayProfiles(profiles []string) {
+	for i := range profiles {
+		fmt.Println(fmt.Sprintf("- %s", profiles[i]))
+	}
+}
+
+func init() {
+	ProfileCmd.AddCommand(profileListCmd)
+}
--- a/cmd/minishift/cmd/profile/profile_reset.go
+++ b/cmd/minishift/cmd/profile/profile_reset.go
@@ -1,0 +1,47 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package profile
+
+import (
+	"fmt"
+
+	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+)
+
+var profileResetCmd = &cobra.Command{
+	Use:   "reset",
+	Short: "resets the active profile",
+	Long:  "resets the active profile. After running this command there will not be any active profile",
+	Run: func(cmd *cobra.Command, args []string) {
+		if len(args) != 0 {
+			atexit.ExitWithMessage(1, extraArgumentError)
+		}
+
+		err := profileActions.ResetActiveProfile()
+		if err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		} else {
+			fmt.Println("Profile is reset successfully")
+		}
+	},
+}
+
+func init() {
+	ProfileCmd.AddCommand(profileResetCmd)
+}
--- a/cmd/minishift/cmd/profile/profile_set.go
+++ b/cmd/minishift/cmd/profile/profile_set.go
@@ -1,0 +1,73 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package profile
+
+import (
+	"fmt"
+
+	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+)
+
+const (
+	emptyProfileError  = "You must provide the profile name. Run `minishift profile list` to view profiles"
+	extraArgumentError = "You have provided more arguments than required. You must provide a single profile name"
+)
+
+var profileSetCmd = &cobra.Command{
+	Use:   "set",
+	Short: "Sets the default profile for Minishift",
+	Long:  "Sets the default profile for Minishift. After you set the profile, all commands will use the profile by default",
+	Run: func(cmd *cobra.Command, args []string) {
+		var doesProfileExist = false
+		if len(args) == 0 {
+			atexit.ExitWithMessage(1, emptyProfileError)
+		} else if len(args) > 1 {
+			atexit.ExitWithMessage(1, extraArgumentError)
+		}
+
+		profileName := args[0]
+
+		//check if the profile is present in the AllInstancesConfig
+		profileList := profileActions.GetProfileNameList()
+		for i := range profileList {
+			if profileList[i] == profileName {
+				doesProfileExist = true
+			}
+		}
+		if !doesProfileExist {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error: '%s' is not a valid profile", profileName))
+		}
+
+		//Check if the requested profile is already active
+		if profileName == profileActions.GetActiveProfile() {
+			atexit.ExitWithMessage(1, fmt.Sprintf("'%s' is already set as active profile", profileName))
+		}
+
+		err := profileActions.SetActiveProfile(profileName)
+		if err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		} else {
+			fmt.Println(fmt.Sprintf("Profile set to '%s' successfully", profileName))
+		}
+	},
+}
+
+func init() {
+	ProfileCmd.AddCommand(profileSetCmd)
+}
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"regexp"
 	"strings"
 
 	"path/filepath"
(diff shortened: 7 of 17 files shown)


LalatenduMohanty:
How to test
==========

* minishift start --profile test

* minishift status --profile test

* minishift profile set test

* minishift status

gbraad:
```
$ make
...
cmd/minishift/cmd/start.go:25: os redeclared as imported package name
	previous declaration at cmd/minishift/cmd/start.go:21
```

cmd/minishift/cmd/status.go
```diff
@@ -34,8 +34,9 @@ var statusCmd = &cobra.Command{
 	Short: "Gets the status of the local OpenShift cluster.",
 	Long:  `Gets the status of the local OpenShift cluster.`,
 	Run: func(cmd *cobra.Command, args []string) {
-		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+		api := libmachine.NewClient(constants.GetMinipath(), constants.MakeMiniPath("certs"))
```
gbraad: Should print chosen profile, as the `set` command hides this information for future use

LalatenduMohanty:
> $ make ... cmd/minishift/cmd/start.go:25: os redeclared as imported package name previous declaration at cmd/minishift/cmd/start.go:21

Fixed now,  it was result of a recent rebase I did.

gbraad:
There is a possible risk when upgrading

```
$ minishift start   # v1.3.1
$ minishift status
Running
$ minishift update  # let's assume this version is with the multiple profiles
$ minishift status
/home/gbraad/.minishift
Does Not Exist
```

hferentschik:
For what it's worth, at least during reviewing it would have been beneficial to have the change for _constants.GetMinipath()_ in a separate commit. Might still be worth doing imo

budhrg:
I would also like to suggest to add profile name in the console log messages as: 

- `Starting local OpenShift cluster using 'kvm' hypervisor for profile 'test'...`
- `Stopping local OpenShift cluster for profile 'test'...`

and wherever it is appropriate.

cmd/minishift/cmd/status.go
LalatenduMohanty: @gbraad You mean we should print profile name also in `minishift status` output? It means we have to change the put to stay consistent with and without profile.

LalatenduMohanty:
> I would also like to suggest to add profile name in the console log messages as:
> 
>     Starting local OpenShift cluster using 'kvm' hypervisor for profile 'test'...
>     Stopping local OpenShift cluster for profile 'test'...

Sure, then we need to keep the output consistent with or without profile. The issue is what should I print inplace of profile name when no profile is chosen.

LalatenduMohanty:
> Might still be worth doing imo

@hferentschik Not sure how I can separate the changes now. If you still want it, I will do it somehow.

LalatenduMohanty:
At this point only following unit tests are failing. Will fix it soon. 

```
--- FAIL: Test_install_with_enable_flag_works (0.00s)
	addon_install_test.go:62: There should be no add-ons installed. Got got : [&addon.DefaultAddOn{AddOn:addon.AddOn(nil), metaData:(*addon.DefaultAddOnMeta)(0xc4202d28a8), commands:[]command.Command{(*command.OcCommand)(0xc4202d28b0), (*command.OcCommand)(0xc4202d28b8)}, enabled:false, path:"/home/lmohanty/.minishift/addons/admin-user", priority:0} &addon.DefaultAddOn{AddOn:addon.AddOn(nil), metaData:(*addon.DefaultAddOnMeta)(0xc4202d28e0), commands:[]command.Command{(*command.OcCommand)(0xc4202d28e8), (*command.EchoCommand)(0xc4202d28f0), (*command.EchoCommand)(0xc4202d28f8), (*command.EchoCommand)(0xc4202d2900), (*command.EchoCommand)(0xc4202d2908)}, enabled:false, path:"/home/lmohanty/.minishift/addons/anyuid", priority:0} &addon.DefaultAddOn{AddOn:addon.AddOn(nil), metaData:(*addon.DefaultAddOnMeta)(0xc4202d2930), commands:[]command.Command{(*command.OcCommand)(0xc4202d2938), (*command.OcCommand)(0xc4202d2940), (*command.OcCommand)(0xc4202d2948)}, enabled:false, path:"/home/lmohanty/.minishift/addons/eap-templates", priority:0} &addon.DefaultAddOn{AddOn:addon.AddOn(nil), metaData:(*addon.DefaultAddOnMeta)(0xc4202d2970), commands:[]command.Command{(*command.OcCommand)(0xc4202d2978), (*command.OcCommand)(0xc4202d2980), (*command.OcCommand)(0xc4202d2988), (*command.OcCommand)(0xc4202d2990), (*command.OcCommand)(0xc4202d2998), (*command.OcCommand)(0xc4202d29a0), (*command.OcCommand)(0xc4202d29a8), (*command.OcCommand)(0xc4202d29b0), (*command.OcCommand)(0xc4202d29b8), (*command.OcCommand)(0xc4202d29c0), (*command.OcCommand)(0xc4202d29c8), (*command.OcCommand)(0xc4202d29d0), (*command.OcCommand)(0xc4202d29d8), (*command.OcCommand)(0xc4202d29e0), (*command.OcCommand)(0xc4202d29e8)}, enabled:false, path:"/home/lmohanty/.minishift/addons/fuse-templates", priority:0}]
FAIL
FAIL	github.com/minishift/minishift/cmd/minishift/cmd/addon	0.045s
```

cmd/minishift/cmd/config/validations.go
```diff
@@ -41,7 +41,7 @@ func IsValidDriver(string, driver string) error {
 }
 
 func RequiresRestartMsg(name string, value string) error {
-	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+	api := libmachine.NewClient(constants.GetMinipath(), constants.MakeMiniPath("certs"))
```
hferentschik: why can we not keep the _Minipath_ var? I know there are bootstrap issues which need to be worked around, but it seems odd to call a methods from the constant file - _constants.GetMinipath()_. It would also reduce the changeset considerably.

If this is not possible, we might want to consider moving _GetMinipath_ out of _constants.go_.

cmd/minishift/cmd/delete.go
```diff
@@ -113,6 +114,16 @@ func removeInstanceConfigs() {
 	}
 }
 
+func removeProfileFromConfig() {
+	name := constants.GetMachineName()
```
hferentschik: _constants.GetMachineName()_, I don't think _constant.go_ is a good place to add this. After all it is not constant.

cmd/minishift/cmd/oc_env.go
```diff
@@ -42,12 +44,19 @@ type OcShellConfig struct {
...
 func getOcShellConfig(ocPath, forcedShell string) (*OcShellConfig, error) {
+	var cmdLine string
+
 	userShell, err := shell.GetShell(forcedShell)
 	if err != nil {
 		return nil, err
 	}
 
-	cmdLine := "minishift oc-env"
+	if constants.IsProfileExists() {
```
hferentschik: _constants.IsProfileExists()_ - a couple of things. contents.go is again not a good place to add this functionality. Maybe profile related functions can be collected in a new package/file. Also, imo the function should be called _ProfileExists_ or _DoesProfileExists_. IMO common practice is the former.

cmd/minishift/cmd/oc_env.go
```diff
@@ -42,12 +44,19 @@ type OcShellConfig struct {
...
+		cmdLine = "minishift oc-env --profile " + constants.GetMachineName()
```
hferentschik: why do we need the switch here? Is having just a single profile not equivalent with having a profile with the name 'default'?

cmd/minishift/cmd/oc_env.go
```diff
@@ -87,7 +96,7 @@ var ocEnvCmd = &cobra.Command{
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error running the oc-env command: %s", err.Error()))
 		}
-
+		//fmt.Println("shellCFG : ", shellCfg) //temp
```
hferentschik: should be removed

cmd/minishift/cmd/profile/profile_list.go
```diff
@@ -0,0 +1,50 @@
...
+			fmt.Println("There are no profiles defined")
+		} else {
+			displayProfiles(profiles)
+		}
+	},
+}
+
+//List existing profile information
+//TODO: Name, Status (running,stopped etc)
+func displayProfiles(profiles []string) {
```
hferentschik: So this is still work in progress, right?

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -0,0 +1,73 @@
...
+
+const (
+	emptyProfileError  = "You must provide the profile name. Run `minishift profile list` to view profiles"
+	extraArgumentError = "You have provided more arguments than required. You must provide a single profile name"
+)
+
+var profileSetCmd = &cobra.Command{
+	Use:   "set",
+	Short: "Sets the default profile for Minishift",
+	Long:  "Sets the default profile for Minishift. After you set the profile, all commands will use the profile bydefault",
```
hferentschik: missing space between by and default

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -0,0 +1,73 @@
...
+	Run: func(cmd *cobra.Command, args []string) {
+		var isProfileExist = false
```
hferentschik: profileExists

cmd/minishift/cmd/root.go
```diff
@@ -80,27 +83,62 @@ var RootCmd = &cobra.Command{
 			isFreshStart bool
 		)
 
-		if !filehelper.Exists(constants.Minipath) || filehelper.IsEmptyDir(constants.Minipath) {
+		//We need to initialize allinstance config as the profile information would be there
```
hferentschik: I think this method needs really to be split up. This will help to make clear what's actually happening as well.

cmd/minishift/cmd/root.go
```diff
@@ -80,27 +83,62 @@ var RootCmd = &cobra.Command{
...
+		//fmt.Println("Minipath = ", minishiftPath)
+
+		if constants.IsProfileExists() {
+
+			//Update MachineName
+			constants.MachineName = constants.GetMachineName()
+
+			constants.KubeConfigPath = filepath.Join(minishiftPath, "machines", constants.MachineName+"_kubeconfig")
+
+			//we have to recalculate the paths to Minishift directories except the config
```
hferentschik: Recalculate, that seems fishy? Something seems not right here.

cmd/minishift/cmd/root.go
```diff
@@ -116,14 +154,15 @@ var RootCmd = &cobra.Command{
 			fmt.Println("OK")
 		}
 
-		// Check marker file created by update command and perform post update execution steps
-		if filehelper.Exists(filepath.Join(constants.Minipath, constants.UpdateMarkerFileName)) {
-			if err := performPostUpdateExecution(filepath.Join(constants.Minipath, constants.UpdateMarkerFileName)); err != nil {
+		// Check marker file created by update command and perform post update execu	tion steps
```
hferentschik: "execs	ion"

pkg/minikube/constants/constants.go
```diff
@@ -62,6 +67,45 @@ func getMinishiftHomeDir() string {
...
+	} else if IsProfileExists() {
+		return filepath.Join(util.HomeDir(), ".minishift", GetMachineName())
+	} else {
+		return filepath.Join(util.HomeDir(), ".minishift")
+	}
+}
+
+//Checks viper or allinstanceconfg to say if profile is defined.
+func IsProfileExists() bool {
+	if viper.GetString("profile") != "" {
```
hferentschik: -1, no calls to viper outside of the cmd structure

pkg/minikube/constants/constants.go
```diff
@@ -62,6 +67,45 @@ func getMinishiftHomeDir() string {
...
+		}
+	}
+	return false
+}
+
+//Returns the machines after taking in to account if --profile or
+//profile information available in allinstance config
+func GetMachineName() string {
+	activeProfile := profileActions.GetActiveProfile()
+	if viper.GetString("profile") != "" {
```
hferentschik: -1, same as above

cmd/minishift/cmd/delete.go
LalatenduMohanty: Agree on this. I need to find a better place for it.

hferentschik: we probably want a _profile.go_ or something

cmd/minishift/cmd/root.go
```diff
@@ -46,7 +48,7 @@ import (
 )
 
 var dirs = [...]string{
```
hferentschik: I don't think we can keep the initialisation like this. The vars are getting initialised before we determine the profile to use. This is probably also the reason why you later have to reset the directories. Really, we need to do the following 

* Initialise Viper
* Determine the profile
* Initialise the required profile specific data structures/directories

I recommend to put the values of _dirs_ into a struct. This way the required directories can be referred to in a type-safe manner. Using _dirs[0] = "bla"_, etc is very flaky.

gbraad:
As per https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/config/profile.go

"profile MINIKUBE_PROFILE_NAME.  You can return to the default minikube profile by running `minikube profile default`"

hferentschik:
BTW, there seems to be no unit test around profiles and this change will require doc updates as well. Given the importance of this feature we probably will end up with a new topic even.

cmd/minishift/cmd/root.go
LalatenduMohanty: I had also thought about in the same direction but wanted to make it work first. Will fix it.

LalatenduMohanty: @hferentschik Regarding initializing Viper and then checking profile does not seem to work for me. We can discuss about this.

cmd/minishift/cmd/config/validations.go
LalatenduMohanty: Will look in to this and get back. I would be interested to reduce the code change.

cmd/minishift/cmd/root.go
hferentschik: > Regarding initializing Viper and then checking profile does not seem to work for me. We can discuss about this.

what I am asking is why you initialize the variables at all. In the form you are using it, the _dirs_ will be initialized before any method executes. In our case this makes no sense anymore. We need to defer this until we know the profile. So in the case you want to keep the _dirs_ var it would be for example, _var dirs = []string{}_. You initialize it as empty slice/array. Then again, we are probably better of with a struct here. 

Note, you will need do be careful here as well in _initConfig_:

```go
    func initConfig() {
    	configPath := constants.ConfigFile
    	viper.SetConfigFile(configPath)
    	viper.SetConfigType("json")
    	err := viper.ReadInConfig()
    	if err != nil {
    		glog.Warningf("Error reading config file at %s: %s",     configPath, err)
    	}
    	setupViper()
    }
```

You need to determine wether the profile parameter is passed or which profile is selected in all instance config. Only once you know this, you can call to _viper.SetConfigFile(configPath)_

cmd/minishift/cmd/config/validations.go
hferentschik: Well, I am pretty sure that if we wanted to we could keep constants.Minipath. There is no reason why this should not work. If we, however, move this out of constant.go we will have the same amount of code change again, since in this case at least _constant._ would change. 

Personally, I probably would try with constants.Minipath to start with, keeping the change small initially. Once everything would be working, one can apply the name refactoring as a separate step.

hferentschik: Note, constants.Minipath is a global variable. Provided you set it to the right value somewhere during startup, there is no difference to calling constants.Minipath and your constants.GetMinipath(). In fact, provided you make use of the minipath multiple times (as we do), the former is even better, since it evaluates the directory only once, instead of running the same code to determine the directory over and over again.

pkg/minikube/constants/constants.go
```diff
@@ -62,6 +67,45 @@ func getMinishiftHomeDir() string {
...
+	}
+}
+
+//Returns Minishift instance path. Its considers the profile if defined and used
+func GetMinipath() string {
+	homeEnv, ok := os.LookupEnv(MiniShiftHomeEnv)
+	if ok {
+		return homeEnv
+	} else if IsProfileExists() {
+		return filepath.Join(util.HomeDir(), ".minishift", GetMachineName())
```
hferentschik: I don't think this is correct. MINISHIFT_HOME determines the home directory for all the machines. It effectively changes the location of the default ~/.minishift. This means, if it is set we need to take it into accound when building the path to the profile. In your case you are returning from this method if it is set and if not you are just assuming ~/.minishift.

I really think we need unit tests here. They will help thinking this all trough and they also serve as living documentation of how we expect this to work.

hferentschik: BTW, you might be better off making this func GetMinipath(profile string) string. This decouples GetMiniPath from getMachineName. As a side effect it becomes easier to test. 

Developing with testing in mind often leads to a better, less coupled design.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -0,0 +1,73 @@
...
+		if err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		} else {
+			fmt.Println(fmt.Sprintf("Profile set to '%s' successfully", profileName))
+		}
+	},
+}
+
+func init() {
+	ProfileCmd.AddCommand(profileSetCmd)
```
gbraad: Minikube does:

```
$ minikube profile default
minikube profile was successfully set to minikube
$ minikube status
minikube: Running
localkube: Running
kubectl: Correctly Configured: pointing to minikube-vm at 192.168.42.123
```

Why do we break consistency with the Minikube command?
Ref: https://github.com/kubernetes/minikube/blob/09709ccf9b2a39ef81507185e126a3966be77a0f/cmd/minikube/cmd/config/profile.go

cmd/minishift/cmd/root.go
```diff
@@ -83,25 +86,61 @@ var RootCmd = &cobra.Command{
...
+			dirs[0] = constants.Minipath
+			dirs[1] = constants.MakeMiniPath("certs")
+			dirs[2] = constants.MakeMiniPath("machines")
+			dirs[3] = constants.MakeMiniPath("cache")
+			dirs[4] = constants.MakeMiniPath("cache", "iso")
+			dirs[5] = constants.MakeMiniPath("cache", "oc")
+			dirs[6] = constants.MakeMiniPath("cache", "images")
+			dirs[8] = constants.MakeMiniPath("addons")
+			dirs[9] = constants.MakeMiniPath("logs")
+			dirs[10] = constants.MakeMiniPath("tmp")
```
LalatenduMohanty: @hferentschik I have not fixed it yet but working on it.

cmd/minishift/cmd/profile/profile_list.go
LalatenduMohanty: Yes, but I am looking for feedback about what should `minishift profile list` display.

cmd/minishift/cmd/config/validations.go
LalatenduMohanty: This is fixed now.

cmd/minishift/cmd/delete.go
LalatenduMohanty: Renamed the fuction to GetProfileName() and moved to cmd/profile/profile.go

cmd/minishift/cmd/oc_env.go
LalatenduMohanty: Fixed it.

cmd/minishift/cmd/oc_env.go
LalatenduMohanty: @gbraad Also brought up the `default` profile point. I think it depends on how we want to shape the user experience and changing this would change the directory structure when no profile is used as at this point by default the instance name is `minishift`.

cmd/minishift/cmd/profile/profile_set.go
LalatenduMohanty: Fixed.

cmd/minishift/cmd/profile/profile_set.go
LalatenduMohanty: Fixed.

cmd/minishift/cmd/root.go
LalatenduMohanty: Working on it.

pkg/minikube/constants/constants.go
LalatenduMohanty: This is fixed now.

pkg/minikube/constants/constants.go
LalatenduMohanty: Fixed now.

cmd/minishift/cmd/root.go
LalatenduMohanty: I still working on this.

cmd/minishift/cmd/root.go
hferentschik: ok

cmd/minishift/cmd/oc_env.go
```diff
@@ -42,12 +45,19 @@ type OcShellConfig struct {
...
+	var cmdLine string
+
 	userShell, err := shell.GetShell(forcedShell)
 	if err != nil {
 		return nil, err
 	}
 
-	cmdLine := "minishift oc-env"
+	if cmdProfile.ProfileExists() {
+		cmdLine = "minishift oc-env --profile " + cmdProfile.GetProfileName()
```
hferentschik: I don't get what's going here? Did we not say that profile is a global option? Also if the profile is set/activated I would expect that I can just run `minishift oc-env`, right? Only if I explicitly run `minishift --profile foo oc-env`, I would expect the need to pass through the profile, but in this case it should become `minishift --profile foo oc-env`.

docs/source/using/profiles.adoc
```diff
@@ -0,0 +1,31 @@
...
+With the help of profile feature you can create multiple instances of Minishift with xref:../command-ref/minishift_start.adoc#[`minishift start --profile`] command.
+
+[[manage-multi-instances-of-minishift]]
+== Manage multiple instances of Minishift
+
+When you have multiple profiles you can run Minishift commands  against a specific profile using `--profile` flag.
+
+If you do not want to pass `--profile` flag everytime with commands you can set a profile as the default with `minishift profile set` command.
+
+To remove a profile from default profile you need to use `minishift profile reset` command.
```
hferentschik: _profile reset_? I don't get this.

docs/source/using/profiles.adoc
```diff
@@ -0,0 +1,31 @@
...
+
++
+----
+$ minishift profile list
+
+$ minishift profile set <profile_name>
+
+$ minishift profile reset
```
hferentschik: What are you really resetting? Once I would execute this command what is the active profile? _default_ I assume? Why not _minishift profile set default_?

cmd/minishift/cmd/root.go
```diff
@@ -174,18 +214,21 @@ func init() {
...
 	RootCmd.AddCommand(addon.AddonsCmd)
 	RootCmd.AddCommand(image.ImageCmd)
+	RootCmd.AddCommand(cmdProfile.ProfileCmd)
 	pflag.CommandLine.AddGoFlagSet(goflag.CommandLine)
 	logDir := pflag.Lookup("log_dir")
 	if !logDir.Changed {
 		logDir.Value.Set(constants.MakeMiniPath("logs"))
 	}
 	viper.BindPFlags(RootCmd.PersistentFlags())
 	cobra.OnInitialize(initConfig)
```
hferentschik: @LalatenduMohanty have you tested the use of the persistent config? For all I can tell from looking at this code, the code initialises against the default location instead of taking the profile into account. _init_ is called after _vars_ are initialised and before main. This means this method is called long before anything else happens. It then registered a callback for cobra.OnInitialize which is most likely also called before _PersistentPreRun_. So when you finally get around changing the various directories Viper is already initialised with the wrong/default _config.json_. Correct me from wrong.

root.go needs to change in various forms.

hferentschik:
> minishift start --profile test
> minishift status --profile test

How does this relate to a global --profile option? Did we not say any command can be explicitly reference a profile, e.g. `minishift --profile foo logs`? Or are you planning to add the profile parameters to each single command? If it is on the global level, do you then not end up with something like `minishift --profile foo --start --profile bar`. Should it hence not be `minishift --profile foo start`?

cmd/minishift/cmd/oc_env.go
hferentschik: @LalatenduMohanty This relates to https://github.com/minishift/minishift/pull/1206#issuecomment-321843743. Can you explain what the plan is?

cmd/minishift/cmd/profile/profile.go
```diff
@@ -0,0 +1,59 @@
...
+	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
+	"github.com/spf13/cobra"
+	"github.com/spf13/viper"
+)
+
+var ProfileCmd = &cobra.Command{
+	Use:     "profile SUBCOMMAND [flags]",
+	Aliases: []string{"instance"},
+	Short:   "Manages Minishift profiles.",
+	Long:    "Manages Minishift profile. Use the sub-commands to list, set profile",
```
hferentschik: This sentence does not seem complete.

cmd/minishift/cmd/profile/profile_reset.go
```diff
@@ -0,0 +1,47 @@
...
+import (
+	"fmt"
+
+	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+)
+
+var profileResetCmd = &cobra.Command{
+	Use:   "reset",
```
hferentschik: I don't think there is a need for reset. Set along should be enough.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -0,0 +1,73 @@
...
+
+		//check if the profile is present in the AllInstancesConfig
+		profileList := profileActions.GetProfileNameList()
+		for i := range profileList {
+			if profileList[i] == profileName {
+				doesProfileExist = true
+			}
+		}
+		if !doesProfileExist {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error: '%s' is not a valid profile", profileName))
```
hferentschik: 'is not a valid profile' instead of "not valid" it should say "does not exist". Not valid implies to me that the name is for reasons not allowed

cmd/minishift/cmd/profile/profile_set.go
hferentschik: I agree. Not only would this align with Minikube  behaviour it also makes sense imo

cmd/minishift/cmd/start.go
```diff
@@ -359,6 +364,16 @@ func autoMountHostFolders(driver drivers.Driver) {
 	}
 }
 
+func addProfileToConfig() {
```
hferentschik: Is this handling the case where one tries to start a new instance using an existing profile name? Some test would be nice here.

hferentschik:
@LalatenduMohanty what's about closing this PR and opening a new one the next time you push an update?

LalatenduMohanty:
@hferentschik agree.

PR #1207, opened 2017-08-01
Issue #1204 Version command contains extra line

agajdosi:


LalatenduMohanty:
Lets discuss why we need to change as mentioned in https://github.com/minishift/minishift/issues/1204#issuecomment-319308468  as it can break integration stuff with JBIDE https://github.com/minishift/minishift/issues/1204#issuecomment-319308468

hferentschik:
Right now, I think I prefer to keep things as they are.

agajdosi:
@LalatenduMohanty @hferentschik Only need for change is that current format is not consistent with other commands in minishift which do not use extra line. I am fine with closing this, will then just remove [WIP] from #1205.

hferentschik:
+1 for closing this pull request and issue and keeping things as they are. @minishift/minishift-dev thoughts?

praveenkumar:
@hferentschik @LalatenduMohanty this PR is only making sure our output should consistent. Right now if you even do version check on downstream then there is no new line and this is not consistent so I think we should take this patch in. I don't think it will break anything from server adapter side.

```
$ ./minishift version  => downstream
minishift v1.3.1+a2d37991
CDK v3.x.x-x
$ ./minishift version => upstream
minishift v1.3.1+d8d4317

$ 
```

hferentschik:
>  this PR is only making sure our output should consistent. Right now if you even do version check on downstream then there is no new line and this is not consistent so I think we should take this patch in.

Ahh, I thought the CDK version string has the empty line as well. Ok then, but let's make sure to contact QE of JBIDE to make sure this gets tested.

hferentschik:
Ok after from further discussion, rebased and merged :-)

Issue #1158, opened 2017-07-20
Update ROADMAP


hferentschik:

hferentschik:
Merged via pull request #1208

PR #1208, opened 2017-08-01
Issue #1158 Updating ROADMAP

hferentschik:


ROADMAP.adoc
```diff
@@ -63,27 +63,35 @@ version name of a given release.
...
 
-- [ ] Ability to provide and use custom a host- and domain-name https://github.com/minishift/minishift/issues/201[#201]
+- [ ] Ability to manage users https://github.com/minishift/minishift/issues/390[#390]
+- [ ] Add a profile parameter https://github.com/minishift/minishift/issues/549[#549]
+- [ ] Provide caching of docker images https://github.com/minishift/minishift/issues/952[#952]
+
+[[minishift-160]]
+=== Minishift 1.6.0
+
+- [ ] Bundle custom libmachine drivers into Minishift https://github.com/minishift/minishift/issues/217[#217]
```
LalatenduMohanty: I think we can take this for later releases and add using Minikube as a library here.

hferentschik: Fair enough. Let me make the adjustments

hferentschik: Minikube as dependency is not really a user facing feature. It is really more of an impl detail. But maybe it is a big enough change to be added to the ROADMAP. Anyways, done

Issue #1159, opened 2017-07-20
Add mailing list info to README and docs


hferentschik:

LalatenduMohanty:
Fixed by https://github.com/minishift/minishift/pull/1210

PR #1210, opened 2017-08-01
Issue #1159 Adding mailing list registration link to docs

hferentschik:


LalatenduMohanty:
@hferentschik Are the issues with respect to mailing list has been resolved yet?

hferentschik:
> Are the issues with respect to mailing list has been resolved yet?

I think there are still some things which are not so nice, but they seem out of our control for now. For example, the wording in the signup message. Apparently these kind of things are stored in some sort of translations resource files of mailman itself. 

Other concerns we had was around navigation, however, the link I added to the docs, points to a page which is least ambiguous. On this page it is quite clear that you can either sign up or just register via email. I think this is fair enough. We also did not get any more feedback from the other @minishift/minishift-dev members regarding issues with the signup.

It might not be perfect, but I think we need to make a move on this now.

hferentschik:
@LalatenduMohanty - good to merge ?

budhrg:
LGTM :+1:

README.adoc
```diff
@@ -62,7 +62,7 @@ Minishift is an open-source project dedicated to developing and supporting Minis
 The code base is forked from the https://github.com/kubernetes/minikube[Minikube] project.
 
 Contributions, questions, and comments are all welcomed and encouraged!
-Minishift developers hang out on IRC in the #minishift channel on Freenode.
+Join the Minishift team on the #minishift IRC channel on  https://freenode.net/[Freenode] or sign up for the https://lists.minishift.io/admin/lists/minishift.lists.minishift.io[Minishift Mailing List].
```
gbraad: "on the [sth] channel on freenode" is not nice to read. Prevent to use the same preposition closely following each other.

Also, this is not an "or" situation. It is better to say, maybe in an amicable way, "You can also signup"

docs/source/contributing/index.adoc
```diff
@@ -8,4 +8,4 @@ Join the Minishift community! Learn about the link:https://github.com/minishift/
 - xref:../contributing/writing-docs.adoc#[Writing Documentation]
 - xref:../contributing/releasing.adoc#[Releasing Minishift]
 
-If you have more questions join us on Freenode IRC #minishift.
+If you have more questions join us on the #minishift IRC channel on link:https://freenode.net/[Freenode] or sign up for the link:https://lists.minishift.io/admin/lists/minishift.lists.minishift.io[Minishift Mailing List].
```
gbraad: Ditto

README.adoc
gbraad: /cc: @thatdocslady WDYT?

hferentschik: I agree the repeated preposition is not so nice. I had "Join the Minishift team on the #minishift Freenode IRC channel" before, but was not quite happy about this either. Open for suggestions.

hferentschik: > Also, this is not an "or" situation. It is better to say, maybe in an amicable way, "You can also signup"

not sure, I don't see this or in the sense of "either ... or". I would not read in a way where I need to make a choice for one of them.

LalatenduMohanty: We should make the information about IRC channel and mailing list more prominent in the readme. e.g.

**Join the Minishift community at**

*  #minishift IRC channel on [Freenode](https://freenode.net/)
* [Minishift Mailing List](https://lists.minishift.io/admin/lists/minishift.lists.minishift.io)

budhrg: Like the idea by @LalatenduMohanty  :+1:

thatdocslady: I also like Lala's suggestion but I would rephrase it a little bit to avoid using "join" as the over-arching verb for both actions. The reason for this is that the word "join" sometimes tends to scare potential community members away since it could be implied as "join us a work on stuff", where some people are just users for now who want to ask a question or follow updates.

This is especially true since in the README for example we clearly say that we welcome questions and comments, and not just contributions :)

I would use something similar to this (both in README and INDEX files, we should be consistent):

===

You can reach the Minishift community by:

* Signing up to our [mailing list](https://lists.minishift.io/admin/lists/minishift.lists.minishift.io)
* Joining the #minishift channel on [Freenode IRC](https://freenode.net/)

===

cc @LalatenduMohanty @budhrg @hferentschik @gbraad  - thoughts?

hferentschik:
@budhrg, @LalatenduMohanty updated using the list approach. This is better indeed.

README.adoc
LalatenduMohanty: @thatdocslady sounds good.

README.adoc
```diff
@@ -62,7 +62,10 @@ Minishift is an open-source project dedicated to developing and supporting Minis
 The code base is forked from the https://github.com/kubernetes/minikube[Minikube] project.
 
 Contributions, questions, and comments are all welcomed and encouraged!
-Minishift developers hang out on IRC in the #minishift channel on Freenode.
+Join the Minishift team on:
+
+- #minishift on https://freenode.net/[Freenode]
+- the https://lists.minishift.io/admin/lists/minishift.lists.minishift.io[Minishift Mailing List]
```
budhrg: I think `the` is not required.

docs/source/contributing/index.adoc
```diff
@@ -8,4 +8,7 @@ Join the Minishift community! Learn about the link:https://github.com/minishift/
 - xref:../contributing/writing-docs.adoc#[Writing Documentation]
 - xref:../contributing/releasing.adoc#[Releasing Minishift]
 
-If you have more questions join us on Freenode IRC #minishift.
+If you have more questions join us on:
+
+- #minishift on https://freenode.net/[Freenode]
+- the https://lists.minishift.io/admin/lists/minishift.lists.minishift.io[Minishift Mailing List]
```
budhrg: I think `the` is not required.

README.adoc
thatdocslady: I think "join" and "team" are not appropriate to use here, because of the same reasoning I gave in https://github.com/minishift/minishift/pull/1210#pullrequestreview-53867569.

docs/source/contributing/index.adoc
gbraad: No need for 'the'

gbraad:
APPROVE with comment to remove 'the'

README.adoc
LalatenduMohanty: I also like @thatdocslady 's suggestion i.e. 

You can reach the Minishift community by:

* Signing up to our [mailing list](https://lists.minishift.io/admin/lists/minishift.lists.minishift.io)
* Joining the #minishift channel on [Freenode IRC](https://freenode.net/)

hferentschik: Like it as well. Pull request updated

thatdocslady:
LTGM :)

hferentschik:
Thanks

Issue #680, opened 2017-04-05
.tar.gz contents should be in a subdirectory


bobmcwhirter:
It's extraordinarily bad-mannered to expand a .tar.gz directly into $PWD without a subdirectory of some sort containing the contents.

    melvin:launchpad-addon bob$ tar ztf ~/Downloads/minishift-1.0.0-rc.1-darwin-amd64.tgz
    LICENSE
    README.md
    minishift

/me goes to clean up his $HOME.

hferentschik:
Depends on #1217

LalatenduMohanty:
Resolved by #1418

PR #1212, opened 2017-08-01
[WIP] Issue #657 Using tar instead of gnutar to build release bundles

hferentschik:
- In order to use tar we need to assemble the content of the archive first in a temp directory

Addresses #657 and #680

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,9 +126,19 @@ synopsis_docs: $(DOCS_SYNOPISIS_DIR)/*.md
 .PHONY: release
 release: clean fmtcheck test prerelease $(GOPATH)/bin/gh-release cross
 	mkdir -p release
-	gnutar -zcf release/minishift-$(MINISHIFT_VERSION)-darwin-amd64.tgz LICENSE README.adoc -C $(BUILD_DIR)/darwin-amd64 minishift
-	gnutar -zcf release/minishift-$(MINISHIFT_VERSION)-linux-amd64.tgz LICENSE README.adoc -C $(BUILD_DIR)/linux-amd64 minishift
-	zip -j release/minishift-$(MINISHIFT_VERSION)-windows-amd64.zip LICENSE README.adoc $(BUILD_DIR)/windows-amd64/minishift.exe
+
+	@mkdir -p $(BUILD_DIR)/minishift-$(MINISHIFT_VERSION)-darwin-amd64
+	@cp LICENSE README.adoc $(BUILD_DIR)/darwin-amd64/minishift $(BUILD_DIR)/minishift-$(MINISHIFT_VERSION)-darwin-amd64
+	tar -zcf release/minishift-$(MINISHIFT_VERSION)-darwin-amd64.tgz -C $(BUILD_DIR) minishift-$(MINISHIFT_VERSION)-darwin-amd64
+
+	@mkdir -p $(BUILD_DIR)/minishift-$(MINISHIFT_VERSION)-linux-amd64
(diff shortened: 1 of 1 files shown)


hferentschik:
> Need to check if update command will fail

@LalatenduMohanty Very good catch. It will fail indeed. I was hoping that _extractBinary_ would be directory agnostic, but in it's current form it is not :-( This can be rectified as part of this pull request, but it still means that one needs a manually upgrade minishift, since the versions won't be able to handle the new structure.

There are several options I can see:

* Ignore #680 and as part of this issue just replace _gnutar_ with _tar_ and keep existing artefact layout
* Amend this pull request with changes to the update code and merge. This means auto update will be broken and a manual install will be needed
* Take a more long term approach and change the update mechanism in a way which can handle both layouts. Get this change merged and released before changing the artefact layout. In combination to this the update command probably needs a 'update --to-version x.y.z' option. With these things in place we can change the layout and everyone using the update mechanism will be able to migrate.

Makefile
```diff
@@ -126,9 +126,19 @@ synopsis_docs: $(DOCS_SYNOPISIS_DIR)/*.md
...
-	gnutar -zcf release/minishift-$(MINISHIFT_VERSION)-linux-amd64.tgz LICENSE README.adoc -C $(BUILD_DIR)/linux-amd64 minishift
-	zip -j release/minishift-$(MINISHIFT_VERSION)-windows-amd64.zip LICENSE README.adoc $(BUILD_DIR)/windows-amd64/minishift.exe
+
+	@mkdir -p $(BUILD_DIR)/minishift-$(MINISHIFT_VERSION)-darwin-amd64
+	@cp LICENSE README.adoc $(BUILD_DIR)/darwin-amd64/minishift $(BUILD_DIR)/minishift-$(MINISHIFT_VERSION)-darwin-amd64
+	tar -zcf release/minishift-$(MINISHIFT_VERSION)-darwin-amd64.tgz -C $(BUILD_DIR) minishift-$(MINISHIFT_VERSION)-darwin-amd64
+
+	@mkdir -p $(BUILD_DIR)/minishift-$(MINISHIFT_VERSION)-linux-amd64
+	@cp LICENSE README.adoc $(BUILD_DIR)/linux-amd64/minishift $(BUILD_DIR)/minishift-$(MINISHIFT_VERSION)-linux-amd64
+	tar -zcf release/minishift-$(MINISHIFT_VERSION)-linux-amd64.tgz -C $(BUILD_DIR) minishift-$(MINISHIFT_VERSION)-linux-amd64
```
gbraad: Is the folder actually IN the archive... ? No no no... we are not going to do this...
I missed seeing the missing `/`

LalatenduMohanty:
> Ignore #680 and as part of this issue just replace gnutar with tar and keep existing artefact layout

For this PR I would suggest we do that. 

> Take a more long term approach and change the update mechanism in a way which can handle both layouts. Get this change merged and released before changing the artefact layout. In combination to this the update command probably needs a 'update --to-version x.y.z' option. With these things in place we can change the layout and everyone using the update mechanism will be able to migrate.

+1 for the idea of supporting both layouts. I would suggest we take it as part of #680 (may be next sprint)

hferentschik:
>> Ignore #680 and as part of this issue just replace gnutar with tar and keep existing artefact layout

> For this PR I would suggest we do that.

+1, I'll think that is indeed the best option. I'll update the pull request.

> +1 for the idea of supporting both layouts. I would suggest we take it as part of #680 (may be next sprint)

I think this is the best approach as well. However, #680 won't do. We need an intermediate issue which changes the code prior to changing the artefact layout. We need to rollout this change before changing the layout. Once it is rolled out we address #680.

hferentschik:
Closing this pull request. I will create a new one just replacing _gnutar_ with _tar_. I created issue #1217 as a pre-requisite for #680 and the ability to change the artefact layout.

Issue #1202, opened 2017-07-31
Update OpenShift version to v3.6.0


LalatenduMohanty:
Refer: https://github.com/openshift/origin/releases/tag/v3.6.0

hferentschik:
@LalatenduMohanty are you creating a pull request for this?

gbraad:
Be aware that we might have to include `--loglevel=1` to get the output we are used to.
And the printed output is not the same as before (as it misses all OK confirmations ATM.
See https://github.com/minishift/minishift/issues/1137#issuecomment-319248918 and onward.

LalatenduMohanty:
@gbraad feel the path with --loglevel=1 might not be good for us. If we do that than we have to change it everytime `cluster up` changes it and we need to maintain backward compatibility with older Origin versions with respect to the output of `minishift start`. 

> And the printed output is not the same as before (as it misses all OK confirmations ATM.
> See #1137 (comment) and onward.

You mean the ok confirmations from cluster up right.

hferentschik:
Merged via pull request #1213

PR #1213, opened 2017-08-02
Issue #1202 Changing the default version of OpenShift to v3.6.0

LalatenduMohanty:

### DIFF
--- a/cmd/minishift/cmd/start_test.go
+++ b/cmd/minishift/cmd/start_test.go
@@ -131,9 +131,9 @@ func (r *RecordingRunner) Output(command string, args ...string) ([]byte, error)
 
 var (
 	testConfig = &clusterup.ClusterUpConfig{
-		OpenShiftVersion: "v1.5.1",
+		OpenShiftVersion: "v3.6.0",
 		Ip:               "192.168.99.42",
-		OcPath:           "/home/john/.minishift/cache/oc/1.5.1/oc",
+		OcPath:           "/home/john/.minishift/cache/oc/3.6.0/oc",
 		RoutingSuffix:    "192.168.99.42.nip.io",
 	}
 	testDir    string
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 
 # Various versions - Minishift, default OpenShift, default B2D ISO
 MINISHIFT_VERSION = 1.3.1
-OPENSHIFT_VERSION = v1.5.1
+OPENSHIFT_VERSION = v3.6.0
 B2D_ISO_VERSION = v1.0.2
 CENTOS_ISO_VERSION = v1.1.0
 COMMIT_SHA=$(shell git rev-parse --short HEAD)


cmd/minishift/cmd/start_test.go
```diff
@@ -131,9 +131,9 @@ func (r *RecordingRunner) Output(command string, args ...string) ([]byte, error)
 
 var (
 	testConfig = &clusterup.ClusterUpConfig{
-		OpenShiftVersion: "v1.5.1",
+		OpenShiftVersion: "v3.6.0",
```
LalatenduMohanty: Changed the version here as it is drastic change in naming pattern from Origin.

LalatenduMohanty:
Looks like integration tests are failing for coolstore

```
09:11:28 
09:11:28 --- Failed scenarios:
09:11:28 
09:11:28     features/basic.feature:117
09:11:28 
09:11:28 26 scenarios (25 passed, 1 failed)
09:11:28 109 steps (108 passed, 1 failed)
```

hferentschik:
> Looks like integration tests are failing for coolstore

coolstrore is or should not be tested per default. There must be something else going on. Message seems indeed a bit weird.

hferentschik:
retest this please

hferentschik:
Hmm, coolstore might be running indeed per default. I don't think this is intentional. AFAIR the plan was to run coolstore only on a nightly basis, since it takes such a long time to run. @gbraad, wdyt?

hferentschik:
Looking at error message more closely, it seems the persistent volume setup/check failed:

    features/basic.feature:117

hferentschik:
There is a name change. It is now 'persistentvolumes/pv0001', our assertion is for 'persistentvolume/pv0001'. Hence the test failure.

gbraad:
Coolstore does not run... Likely what you are seeing is the following:

```
Feature: Coolstore
   blah bla

Feature: Basic
   blah bla

  Scenario: blah bla
     Given ...
```

It will always show the featured that are available, but it only runs the tags we request for.

agajdosi:
> It will always show the featured that are available, but it only runs the tags we request for.

We can get over this by specifying tests not by tags but by filenames.

gbraad:
I think we should be able to run multiple tests, but the confusioon came from how godog reports this back.

I did not find a scenario that matched the criteria. It does however showo the features it has tried to loook for this... nothing wrong ;-)

hferentschik:
@gbraad 

> Coolstore does not run... Likely what you are seeing is the following:

It does. As mentioned in another comment, this is not the issue and it is also not failing on coolstore test. The problem is with the persistent volume check. I am testing an update as we speak.

Regarding coolstore:

    $ make integration 

does run it. 

Look at the target:

    go test -timeout 3600s $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) $(GODOG_OPTS)

What would exclude it from being run? 

So I think we should open a new issue to actually make sure that coolstore is not run per default and maybe while on it setup the CI job to run the full integration test suite as part of a nightly build.

gbraad:
CentOS CI runs the command as mentioned above:

```
$ make integration GODOG_OPTS="-tags basic -format pretty"
```

hferentschik:
@gbraad I was about to take it back myself ;-) I've seen this as well now. I would argue though we should change the Makefile as well.

gbraad:
I think we need to be able to specify the tags as follows:

```
$ make integration basic
```

hferentschik:
Updated Cucumber test and merged.

agajdosi:
@gbraad +1 for changing the tag to `@nightly` and adding it into basic.feature also. We can later go for several tiers describing time cost of tests: basic aka quick, nightly and complete/release.

Please create an issue for `make integration basic` will be happy to implement it. We had a discussion around this already I think, but back then we agreed that we will try it with GODOG_OPTS, however `make integration basic` and probably also `make integration nightly` are just friendlier.

Issue #1176, opened 2017-07-25
Handle slice variables explicitly in the start command


praveenkumar:
This is follow up for #1153 we found out there is problem is in the viper and the way it handle config option and command line option is not consistent and that's kind of back us using Var at all and instead we should create our own Flag with custom implementation of slice so it work consistently with config and command line options. @hferentschik already did all hard work to implement that custom string slice we need to consume and test our all start slice option and then use it.

- https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/util/shared_flags.go#L56

hferentschik:
Merged via pull request #1220

PR #1220, opened 2017-08-03
Issue #1176 Handle slice variables explicitly in start command

praveenkumar:
@hferentschik have a look.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -70,12 +70,39 @@ const (
 )
 
 var (
-	dockerEnv        []string
-	dockerEngineOpt  []string
-	insecureRegistry []string
-	registryMirror   []string
-	openShiftEnv     []string
-	shellProxyEnv    string
+	dockerEnv     []string
+	openShiftEnv  []string
+	shellProxyEnv string
+)
+
+var (
+	dockerEnvFlag = &flag.Flag{
+		Name:      configCmd.DockerEnv.Name,
+		Shorthand: "",
+		Usage:     "Environment variables to pass to the Docker daemon. Use the format <key>=<value>.",
+		Value:     cmdutil.NewStringSliceValue([]string{}, &[]string{}),
+	}
+
+	insecureRegistryFlag = &flag.Flag{
+		Name:      configCmd.InsecureRegistry.Name,
+		Shorthand: "",
+		Usage:     "Non-secure Docker registries to pass to the Docker daemon.",
+		Value:     cmdutil.NewStringSliceValue([]string{defaultInsecureRegistry}, &[]string{}),
+	}
+
+	dockerEngineOptFlag = &flag.Flag{
+		Name:      configCmd.DockerEngineOpt.Name,
+		Shorthand: "",
+		Usage:     "Specify arbitrary flags to pass to the Docker daemon in the form <flag>=<value>.",
+		Value:     cmdutil.NewStringSliceValue([]string{}, &[]string{}),
+	}
+
+	registryMirrorFlag = &flag.Flag{
+		Name:      configCmd.RegistryMirror.Name,
+		Shorthand: "",
+		Usage:     "Registry mirrors to pass to the Docker daemon.",
+		Value:     cmdutil.NewStringSliceValue([]string{}, &[]string{}),
+	}
 )
 
 // startCmd represents the start command
@@ -252,13 +279,10 @@ func handleProxies() *util.ProxyConfig {
 	return proxyConfig
 }
 
-// getSlice return slice for provided key if value type is []interface{} otherwise nil
+// getSlice return slice for provided key otherwise nil
 func getSlice(key string) []string {
 	if viper.IsSet(key) {
-		value := viper.Get(key)
-		if _, ok := value.([]interface{}); ok {
-			return viper.GetStringSlice(key)
-		}
+		return viper.GetStringSlice(key)
 	}
 	return nil
 }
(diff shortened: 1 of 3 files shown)


hferentschik:
> have a look.

Will do. Did you verify that the settings are passed through regardless whether they come from the config or form the command line? I will also try ...

praveenkumar:
> Did you verify that the settings are passed through regardless whether they come from the config or form the command line?

yes it does work, even test case pass which have config option and command line argument.

hferentschik:
> yes it does work, even test case pass which have config option and command line argument.

:-)

cmd/minishift/cmd/util/shared_flags.go
```diff
@@ -37,26 +38,26 @@ var HttpProxyFlag = &flag.Flag{
 	Name:      HttpProxy,
 	Shorthand: "",
 	Usage:     "HTTP proxy in the format http://<username>:<password>@<proxy_host>:<proxy_port>. Overrides potential HTTP_PROXY setting in the environment.",
-	Value:     newStringValue("", new(string)),
+	Value:     NewStringValue("", new(string)),
```
hferentschik: +1

cmd/minishift/cmd/start.go
```diff
@@ -409,23 +433,23 @@ func determineIsoUrl(iso string) string {
...
-	startFlagSet.StringArrayVar(&dockerEnv, configCmd.DockerEnv.Name, nil, "Environment variables to pass to the Docker daemon. Use the format <key>=<value>.")
-	startFlagSet.StringSliceVar(&dockerEngineOpt, configCmd.DockerEngineOpt.Name, nil, "Specify arbitrary flags to pass to the Docker daemon in the form <flag>=<value>.")
-	startFlagSet.StringSliceVar(&insecureRegistry, configCmd.InsecureRegistry.Name, []string{defaultInsecureRegistry}, "Non-secure Docker registries to pass to the Docker daemon.")
-	startFlagSet.StringSliceVar(&registryMirror, configCmd.RegistryMirror.Name, nil, "Registry mirrors to pass to the Docker daemon.")
-	startFlagSet.AddFlag(cmdutil.AddOnEnvFlag)
-
-	return startFlagSet
+// initconfigCmd creates the CLI flags which needs to be passed on to 'libmachine'
+func initconfigCmd() *flag.FlagSet {
+	configCmdet := flag.NewFlagSet(commandName, flag.ContinueOnError)
```
hferentschik: Is this related to the merge conflict or what does _configCmdet_ mean. It feels like an over eager find and replace

cmd/minishift/cmd/start.go
```diff
@@ -252,13 +279,10 @@ func handleProxies() *util.ProxyConfig {
 	return proxyConfig
 }
 
-// getSlice return slice for provided key if value type is []interface{} otherwise nil
+// getSlice return slice for provided key otherwise nil
 func getSlice(key string) []string {
 	if viper.IsSet(key) {
```
hferentschik: :+1

cmd/minishift/cmd/start.go
praveenkumar: :), you are right.

cmd/minishift/cmd/start.go
```diff
@@ -286,9 +310,9 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 		DiskSize:         calculateDiskSizeInMB(viper.GetString(configCmd.DiskSize.Name)),
 		VMDriver:         viper.GetString(configCmd.VmDriver.Name),
 		DockerEnv:        append(dockerEnv, getSlice(configCmd.DockerEnv.Name)...),
-		DockerEngineOpt:  append(dockerEngineOpt, getSlice(configCmd.DockerEngineOpt.Name)...),
-		InsecureRegistry: append(insecureRegistry, determineInsecureRegistry(configCmd.InsecureRegistry.Name)...),
-		RegistryMirror:   append(registryMirror, getSlice(configCmd.RegistryMirror.Name)...),
+		DockerEngineOpt:  getSlice(configCmd.DockerEngineOpt.Name),
```
hferentschik: +1

cmd/minishift/cmd/start.go
```diff
@@ -115,7 +142,7 @@ For the latter see 'minishift config -h'.`,
 
 	clusterUpFlagSet = initClusterUpFlags()
 	startCmd.Flags().AddFlagSet(clusterUpFlagSet)
-	startCmd.Flags().AddFlagSet(initStartFlags())
+	startCmd.Flags().AddFlagSet(initconfigCmd())
```
hferentschik: here is another one

cmd/minishift/cmd/start_test.go
```diff
@@ -413,8 +413,11 @@ func Test_getslice_withCommandLine(t *testing.T) {
 	viper.Set("foo", []string{"hello", "world"})
 	defer viper.Reset()
 
-	if getSlice("foo") != nil {
-		t.Errorf("expected %+v. Got %+v", nil, getSlice("foo"))
+	expectedSlice := []string{"hello", "world"}
+	for i, v := range getSlice("bar") {
```
hferentschik: There is a bug here. You are not testing anything here. You call _getSlice_ for _bar_. This will be empty and hence it will not go into the loop and pass the test. You need to add an asserts that you get a non empty slice. And you need of course to request 'foo', not 'bar'

hferentschik: Same btw for _Test_getslice_withConfig_. You need to assert the size, otherwise it will be easy to break the test unnoticed.

hferentschik:
I verified this change using _docker-env_ from the CLI as well as from the config. In both cases it worked (verified in the running VM). This will hopefully be properly tested as part of #1196 and its integration tests.

hferentschik:
Merged, thanks.

Issue #613, opened 2017-03-27
Add integration test for service command


budhrg:
This is a follow issue of https://github.com/minishift/minishift/issues/342 on adding integration test for service command.

hferentschik:
Merged via pull request #1228

Issue #1227, opened 2017-08-04
Minishift openshift config view fails when node is targeted


agajdosi:
I have a problem seeing OpenShift configuration for node with command`minishift openshift config view --target node`. I am using Minishift v1.3.1+7c97e62 from master, the file might have been relocated in oc - tried with GA and there it is fine.

```
[agajdosi@dhcp-24-156 bin]$ ./minishift openshift config view --target node
Cannot display the OpenShift configuration: ssh command error:
command : docker exec -t origin cat /var/lib/origin/openshift.local.config/node-192.168.42.5/node-config.yaml
err     : exit status 1
output  : cat: /var/lib/origin/openshift.local.config/node-192.168.42.5/node-config.yaml: No such file or directory
```

hferentschik:
Ahh, this is due to a change in the node name between OpenShift 1.5 and 3.6. I believe the node config is now always node-127.0.0.1.

hferentschik:
Merged via pull request #1284

hferentschik:
Cucumber integration test will be added as part of pull request #1228

PR #1228, opened 2017-08-04
Issue #613 Add tests for 'minishift openshift' sub command

agajdosi:
This adds integration test for `minishift openshift service` and other subcommands of `minishift openshift` command. This is first bigger command I am trying to cover, so I will be thankful for any feedback. Can this be the way we could go in parallel to user story based tests?

To remove WIP there is need to at least add `minishift openshift config set` subcommand for which I need to do changes in "executing xyz succeeds" as I need to insert for example this: `minishift openshift config set --patch '{"assetConfig": {"logoutURL": "http://www.minishift.io"}}'` and right now it is not really quotes welcoming.

### DIFF
--- a/test/integration/features/cmd-openshift.feature
+++ b/test/integration/features/cmd-openshift.feature
@@ -1,0 +1,119 @@
+@cmd-openshift
+Feature: Openshift commands
+Commands "minishift openshift [sub-command]" are used for interaction with Openshift
+cluster in VM provided by Minishift.
+
+  Scenario: Trying service command when Minishift is not running
+     Given Minishift has state "Does Not Exist"
+      When executing "minishift openshift service list" succeeds
+      Then stdout should contain
+      """
+      Running this command requires an existing 'minishift' VM, but no VM is defined.
+      """
+
+  Scenario: Minishift start
+  Minishift must be started in order to interact with OpenShift via "minishift openshift" command
+      When executing "minishift start" succeeds
+
+  Scenario: Service list sub-command
+     Given Minishift has state "Running"
+      When executing "minishift openshift service list" succeeds
+      Then stdout should contain "docker-registry"
+       And stdout should contain "kubernetes"
+       And stdout should contain "router"
+
+  Scenario: Restarting the OpenShift cluster
+  Note: This step is based on observation and might be unstable in some environments. It checks for the time when container
+        finished last time. When container is new and had never finished then this time value is set to 0001-01-01T00:00:00Z.
+        On restart of OpenShift cluster containers are terminated, which sets FinishedAt to actual time. This value persist
+        after next start of container.
+     Given stdout of command "minishift ssh -- docker inspect --format={{.State.FinishedAt}} origin" is equal to "0001-01-01T00:00:00Z"
+      When executing "minishift openshift restart" succeeds
+      Then stdout should contain "Restarting OpenShift"
+       And stdout of command "minishift ssh -- docker inspect --format={{.State.FinishedAt}} origin" is not equal to "0001-01-01T00:00:00Z"
+
+  Scenario: User deploys nodejs example application from OpenShift repository
+      When executing "oc new-app https://github.com/openshift/nodejs-ex -l name=myapp" succeeds
+      Then stdout should contain
+      """
+      Run 'oc status' to view your app.
+      """
+
+  @minishift-only
+  Scenario: Getting information about OpenShift and kubernetes versions
+  Prints the current running OpenShift version to the standard output.
+     Given Minishift has state "Running"
+      When executing "minishift openshift version" succeeds
+      Then stdout should match
+      """
+      ^openshift v[0-9]+\.[0-9]+\.[0-9]+\+[0-9a-z]{7}
+      kubernetes v[0-9]+\.[0-9]+\.[0-9]+\+[0-9a-z]{10}
+      etcd [0-9]+\.[0-9]+\.[0-9]+
+      """
+
+  Scenario: Getting address of internal docker registry
+  Prints the host name and port number of the OpenShift registry to the standard output.
+     Given Minishift has state "Running"
+      When executing "minishift openshift registry" succeeds
+      Then stdout should be valid IP with port number
+
+  Scenario: Getting existing service without route
+      When executing "minishift openshift service nodejs-ex" succeeds
+      Then stdout should contain "nodejs-ex"
+       And stdout should not match
+       """
+       ^http:\/\/nodejs-ex-myproject\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.nip\.io
+       """
+
+  Scenario: Getting non-existing service
+  If service does not exist, user gets an empty table.
+      When executing "minishift openshift service not-present" succeeds
+      Then stdout should not contain "not-present"
+
+  Scenario: Getting service from non-existing namespace
+      When executing "minishift openshift service nodejs-ex --namespace does-not-exist" fails
+      Then stderr should contain "Namespace does-not-exist doesn't exist"
+
+  Scenario: Forgotten service name
+      When executing "minishift openshift service --namespace myapp" fails
+      Then stderr should contain "You must specify the name of the service."
+
+  Scenario: User creates route to the service
+      When executing "oc expose svc/nodejs-ex" succeeds
+      Then stdout should contain
+      """
+      route "nodejs-ex" exposed
+      """
+
+  Scenario: Getting existing service with route
+      When executing "minishift openshift service nodejs-ex" succeeds
+      Then stdout should contain "nodejs-ex"
+       And stdout should match
+       """
+       http:\/\/nodejs-ex-myproject\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.nip\.io
+       """
+
+  Scenario: Getting URL of service using --url flag
+      When executing "minishift openshift service nodejs-ex --url" succeeds
+      Then stdout should be valid URL
+
+  Scenario: Seeing configuration of OpenShift master
+  Minishift openshift config view prints YAML configuration of OpenShift cluster.
+  Note: --target=master is default value for minishift openshift config command
+      When executing "minishift openshift config view" succeeds
+      Then stdout should be valid YAML
+
+  Scenario: Seeing configuration of OpenShift node
+      When executing "minishift openshift config view --target node" succeeds
+      Then stdout should be valid YAML
+
+  Scenario: Setting configuration on OpenShift master
+      When executing »minishift openshift config set --patch '{"assetConfig": {"logoutURL": "http://www.minishift.io"}}'« succeeds
+      Then stdout should contain "Patching OpenShift configuration"
+      When executing "minishift openshift config view" succeeds
+      Then stdout is YAML which contains key "assetConfig.logoutURL" with value matching "http://www\.minishift\.io"
+
+  Scenario: Deleting the Minishift instance
+     Given Minishift has state "Running"
+      When executing "minishift delete" succeeds
+      Then Minishift should have state "Does Not Exist"
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -37,6 +37,7 @@ import (
 
 	"github.com/DATA-DOG/godog"
 	"github.com/DATA-DOG/godog/gherkin"
+	"gopkg.in/yaml.v2"
 
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	testProxy "github.com/minishift/minishift/test/integration/proxy"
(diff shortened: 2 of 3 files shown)


hferentschik:
ok to test

hferentschik:
add to whitelist

agajdosi:
Failure on Centos CI is expected - only one scenario failed, which is: "When executing "minishift openshift config view --target node" succeeds". And this is expected as there is #1227.

hferentschik:
> Failure on Centos CI is expected - only one scenario failed, which is: When executing "minishift openshift config view --target node" succeeds. And this is expected as there is #1227.

Can we in this case exclude this test for now?

agajdosi:
@hferentschik Sure, will do that in next update of this PR.

hferentschik:
+1

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,114 @@
...
+      Then stdout should contain "nodejs-ex"
+       And stdout should not match
+       """
+       ^http:\/\/nodejs-ex-myproject\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.nip\.io
+       """
+
+  Scenario: Getting non-existing service
+  If service does not exist, user gets an empty table.
+      When executing "minishift openshift service not-present" succeeds
+      Then stdout should not contain "not-present"
```
hferentschik: Interesting, what does it contain? Just empty?

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,114 @@
...
+
+  Scenario: Getting service from non-existing namespace
+      When executing "minishift openshift service nodejs-ex --namespace does-not-exist" fails
+      Then stderr should contain "Namespace does-not-exist doesn't exist"
+
+  Scenario: Forgotten service name
```
hferentschik: 👍 nice

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,114 @@
...
+      When executing "minishift openshift config view" succeeds
+      Then stdout should be valid YAML
+
+#  Scenario: Seeing configuration of OpenShift node
+#      When executing "minishift openshift config view --target node" succeeds
+#      Then stdout should be valid YAML#
+
+  Scenario: Setting configuration on OpenShift master
+      When executing »minishift openshift config set --patch '{"assetConfig": {"logoutURL": "http://www.minishift.io"}}'« succeeds
+      Then stdout should contain "Patching OpenShift configuration"
```
hferentschik: I guess here is somehow missing that the patch got applied. One could run an assert prior to the command that 'minishift openshift config view' does not contain 'logoutURL": "http://www.minishift.io"', but afterwards it does. Also the YAML should still be valid

test/integration/features/cmd-openshift.feature
agajdosi: @hferentschik Yes, it is just empty table.

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,114 @@
...
+      
+
+  Scenario: Restarting the OpenShift cluster
+     Given Minishift has state "Running"
+      When executing "minishift openshift restart" succeeds
+      Then stdout should contain "Restarting OpenShift"
```
hferentschik: Again, something is missing here. You need a verification that OpenShift really restarted. One can do this by comparing the container ids. They need to change. You would need to run a docker command prior to restart to determine the id of the container. Then you run the restart and get the id again. They need to differ. I am not sure whether we have all the pieces to do this. I guess you can use @gbraad ability to store the outcome of the command as variable. Not quite sure how you go then about comparing.

test/integration/features/cmd-openshift.feature
agajdosi:
```
[agajdosi@dhcp-24-156 ~]$ minishift openshift service not-present
|-----------|------|----------|-----------|--------|
| NAMSEPACE | NAME | NODEPORT | ROUTE-URL | WEIGHT |
|-----------|------|----------|-----------|--------|
|-----------|------|----------|-----------|--------|

```

test/integration/integration_test.go
```diff
@@ -380,6 +385,23 @@ func shouldBeInValidFormat(commandField, format string) error {
 		if net.ParseIP(result) == nil {
 			return fmt.Errorf("%s of previous command is not a valid IP address: %s", commandField, result)
 		}
+	case "IP with port number":
+		split := strings.Split(result, ":")
```
hferentschik: I split the content of each case statement into its own function. Small functions is the aim. Makes understanding and testing the code easier. 

A propos testing, we maybe should start writing unit test for these functions ;-)

test/integration/integration_test.go
```diff
@@ -455,6 +477,7 @@ func executingOcCommandSucceedsOrFails(command string, expectedResult string) er
 }
 
 func executingMinishiftCommandSucceedsOrFails(command string, expectedResult string) error {
+	fmt.Println(command)
```
hferentschik: Do you want to have this here permanently? If so, maybe we should format it a bit more

test/integration/util/util.go
```diff
@@ -42,7 +43,7 @@ type OcRunner struct {
 }
 
 func runCommand(command string, commandPath string) (stdOut string, stdErr string, exitCode int) {
-	commandArr := strings.Split(command, " ")
+	commandArr := utilCmd.SplitCmdString(command)
```
hferentschik: +1 :-)

test/integration/integration_test.go
agajdosi: will be removed

test/integration/features/cmd-openshift.feature
hferentschik: Right, right, I remember now

test/integration/features/cmd-openshift.feature
hferentschik: this should give you the container id

    minishift ssh -- docker ps -q --filter "name=origin"

test/integration/features/cmd-openshift.feature
agajdosi: +1 Actually I was thinking about making code around JSON config file to be YAML compatible also. And to refactor it a little bit so it could also extract values from stdout, stderr etc.

test/integration/features/cmd-openshift.feature
agajdosi: @hferentschik Agree, thanks for suggestion, will try to implement it.

test/integration/integration_test.go
agajdosi: @hferentschik Forgotten line from debugging the problem with executing command step, will be removed.

agajdosi:
Implemented YAML support (ufff, took me some time), also it is now possible to check stdout and stderr for values in JSON/YAML.

Still needs check for restart of OpenShift + possibly a split of func shouldBeInValidFormat().

agajdosi:
Forgot to delete test.feature file.

test/integration/features/cmd-openshift.feature
agajdosi: @hferentschik So I have tested it and the ID of container is not changing:
```
[agajdosi@dhcp-24-156 ~]$ minishift ssh -- docker ps -q --filter "name=origin"
ab78dbc622f0
[agajdosi@dhcp-24-156 ~]$ minishift openshift restart
Restarting OpenShift
[agajdosi@dhcp-24-156 ~]$ minishift ssh -- docker ps -q --filter "name=origin"
ab78dbc622f0
```

agajdosi: So I have found the solution :)
```gherkin
  Scenario: Restarting the OpenShift cluster
     Given stdout of command "minishift ssh -- docker inspect --format={{.State.FinishedAt}} origin" is equal to "0001-01-01T00:00:00Z"
      When executing "minishift openshift restart" succeeds
      Then stdout should contain "Restarting OpenShift"
       And stdout of command "minishift ssh -- docker inspect --format={{.State.FinishedAt}} origin" is not equal to "0001-01-01T00:00:00Z"
```

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,115 @@
+@cmd-openshift
+Feature: Openshift commands
+Commands "minishift openshift [sub-command]" are used for interaction with Openshift
+cluster in VM provided by Minishift.
+
+  Scenario: Trying service command when Minishift is not running
```
hferentschik: nice 👍

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,115 @@
...
+     Given Minishift has state "Does Not Exist"
+      When executing "minishift openshift service list" succeeds
+      Then stdout should contain
+      """
+      Running this command requires an existing 'minishift' VM, but no VM is defined.
+      """
+
+  Scenario: Minishift start
```
hferentschik: :+1

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,115 @@
...
+
+  Scenario: Service list sub-command
+     Given Minishift has state "Running"
+      When executing "minishift openshift service list" succeeds
+      Then stdout should contain "docker-registry"
+       And stdout should contain "kubernetes"
+       And stdout should contain "router"
+
+  Scenario: Restarting the OpenShift cluster
+     Given stdout of command "minishift ssh -- docker inspect --format={{.State.FinishedAt}} origin" is equal to "0001-01-01T00:00:00Z"
```
hferentschik: Nice to use docker inside the VM +1

hferentschik: I am intrigued by '0001-01-01T00:00:00Z' is that a given? Is that the default value for a running container?

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,115 @@
...
+      When executing "minishift openshift restart" succeeds
+      Then stdout should contain "Restarting OpenShift"
+       And stdout of command "minishift ssh -- docker inspect --format={{.State.FinishedAt}} origin" is not equal to "0001-01-01T00:00:00Z"
```
hferentschik: so is the '.State.FinishedAt' field then referencing the time when this container had finished the last time?

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,115 @@
...
+       And stdout should match
+       """
+       http:\/\/nodejs-ex-myproject\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.nip\.io
+       """
+
+  Scenario: Getting URL of service using --url flag
+      When executing "minishift openshift service nodejs-ex --url" succeeds
+      Then stdout should be valid URL
+
+  Scenario: Seeing configuration of OpenShift master
```
hferentschik: +1 :-)

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,115 @@
...
+  Minishift openshift config view prints YAML configuration of OpenShift cluster.
+  Note: --target=master is default value for minishift openshift config command
+      When executing "minishift openshift config view" succeeds
+      Then stdout should be valid YAML
+
+  Scenario: Seeing configuration of OpenShift node
+      When executing "minishift openshift config view --target node" succeeds
+      Then stdout should be valid YAML
```
hferentschik: that should do. nice follow up to @praveenkumar pull request.

test/integration/features/cmd-openshift.feature
agajdosi: It seems to be default null value.

agajdosi:
@hferentschik Added comment on scenarios using FinishedAt time value. Also this test passed with CentOS image used.

test/integration/features/cmd-openshift.feature
```diff
@@ -0,0 +1,119 @@
...
+
+  Scenario: Service list sub-command
+     Given Minishift has state "Running"
+      When executing "minishift openshift service list" succeeds
+      Then stdout should contain "docker-registry"
+       And stdout should contain "kubernetes"
+       And stdout should contain "router"
+
+  Scenario: Restarting the OpenShift cluster
+  Note: This step is based on observation and might be unstable in some environments. It checks for the time when container
```
hferentschik: thanks :-)

test/integration/integration_test.go
```diff
@@ -150,11 +151,15 @@ func FeatureContext(s *godog.Suite) {
 		minishift.shouldHaveState)
 	s.Step(`^executing "minishift ([^"]*)"$`,
 		minishift.executingMinishiftCommand)
+	s.Step(`^executing »minishift ([^«]*)«$`,
```
hferentschik: I think the problem is in this regexp '^executing "minishift ([^"]*)"$'. Matches are greedy, but '[^"]' forces it to stop after the first occurnace of a quote.

I think it should be just '^executing "minishift (.*)"$'. That should make the need to add another step obsolete.

hferentschik:
Merged, but I took the liberty to change the 'executing "minishift' step and remove the step using the french quotes. The issue imo is not that regexp are not greedy, but the fact that we explicitly excluded any " to appear in the command. I changed this for the '^executing "minishift (.*)"$' and '^executing "minishift (.*)" (succeeds|fails)$'. 

@agajdosi as you said, it might be worth changing other step definitions as well. Feel free to create a an issue and pull request to change them as well.

PR #1236, opened 2017-08-07
Issue #1235 Fix issues with CI release job

### LINKED ISSUE #1235: Wrong version info in v1.4.0 binaries
This is following up on #786 which resulted in wrong version in binary.

* _'curl -s https://api.github.com/repos/minishift/minishift/milestones'_ should specify 'state=all' when requesting the milestones - https://developer.github.com/v3/issues/milestones/
* _gen_adoc_tar_ should be called after making the release commit 
* _release_ target should depend on _clean_ and _cross_

budhrg:
Fix #1235

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,7 +137,7 @@ prerelease: clean fmtcheck test cross
 	fi
 
 .PHONY: release
-release: $(GOPATH)/bin/gh-release
+release: clean $(GOPATH)/bin/gh-release cross
 	mkdir -p release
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -128,7 +128,7 @@ $(DOCS_SYNOPISIS_DIR)/*.md: vendor $(ADDON_ASSET_FILE)
 synopsis_docs: $(DOCS_SYNOPISIS_DIR)/*.md
 
 .PHONY: prerelease
-prerelease: clean fmtcheck test cross
```
hferentschik: We should keep cross here as well. As you say, we want to make sure that cross compilation works beforehand

budhrg: Agree :+1:

hferentschik:
Merged via pull request #1236

Issue #460, opened 2017-02-21
Add GitLab CI runner build defintion


gbraad:
Although not essential, it is something I use myself, and runs on my own instance of GitLab. It however, can also run public as can be seen on: https://gitlab.com/gbraad/minishift-build and an example build: https://gitlab.com/gbraad/minishift-build/builds/10847637

However, the current setup relies on webhooks to have a build triggered, as the definition lives outside the archive. If a slightly modified .gitlab-ci.yml could live inside the minishift archive, the builds get triggered on each commit. It will so allow the community to easily take the archive and produce a completely automated pipeline using GitLab and the CI runners for themselves.

LalatenduMohanty:
@gbraad I think it would be useful for community developers who want to contribute to minishift project, right?

gbraad:
yes... as they easily replicate an environment that can build, and eventually even test.

praveenkumar:
@gbraad you still want to look into it or we can close it right now and reopen when have some time?

budhrg:
I am happy to keep it if it provide consistent results, easy configuration over Travis CI. 
If the infrastructure provides nested virt, then its worth to try.

gbraad:
Alreayd have this working for myself... can contribute this. just didn't take the time to adjust it for consistency with the current `centos_ci.sh` to prevent duplication.

hferentschik:
What would it give us? We have quite a few CI systems already? Is there any feature we would get which we don't have. If you say it supported nested virt, I am in. If not, I am not sure whether it really adds value at this stage.

LalatenduMohanty:
Fixed via https://github.com/minishift/minishift/pull/1239

PR #1239, opened 2017-08-08
Issue #460 Add GitLab CI runner

gbraad:
Added GitLab CI runner definition to build on a GitLab instance. Example:
  * https://gitlab.com/gbraad/minishift/-/jobs/28627751
 * https://gitlab.com/minishift/minishift/-/jobs/28634280

`oc_env_test.go` needed to be given a forced shell, else the test fails with:
```
=== RUN   Test_unix_oc_path
The default lines below are for a sh/bash shell, you can specify the shell you're using, with the --shell flag.

--- FAIL: Test_unix_oc_path (0.00s)
	oc_env_test.go:28: Unexepcted error: Error: Unknown shell
```
Note: running with `bash -c "make test"` did not make a difference

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,18 @@
+build:
+  image: fedora:26
+  script:
+    - dnf install -y golang glide make git python
+    - export CURRENT_BUILD_PATH=$PWD
+    - mkdir -p $HOME/gopath
+    - export GOPATH=$HOME/gopath
+    - export PATH=$PATH:$GOPATH/bin
+    - mkdir -p $HOME/gopath/src/github.com/minishift
+    - ln -s $CURRENT_BUILD_PATH $HOME/gopath/src/github.com/minishift/minishift
+    - cd $HOME/gopath/src/github.com/minishift/minishift/
+    - make prerelease
+    - rm -rf $HOME/gopath/src/github.com/minishift/minishift/out/bindata
+    - mv $HOME/gopath/src/github.com/minishift/minishift/out/ $CURRENT_BUILD_PATH/public
+  artifacts:
+    paths:
+      - public
+
--- a/cmd/minishift/cmd/oc_env_test.go
+++ b/cmd/minishift/cmd/oc_env_test.go
@@ -23,7 +23,7 @@ import (
 )
 
 func Test_unix_oc_path(t *testing.T) {
-	shellConfig, err := getOcShellConfig("/Users/john/.minishift/cache/oc/v1.5.0/oc", "")
+	shellConfig, err := getOcShellConfig("/Users/john/.minishift/cache/oc/v1.5.0/oc", "bash")
 	if err != nil {
 		t.Fatalf("Unexepcted error: %s", err)
 	}


.gitlab-ci.yml
```diff
@@ -0,0 +1,19 @@
...
+    - ln -s $CURRENT_BUILD_PATH $HOME/gopath/src/github.com/minishift/minishift
+    - cd $HOME/gopath/src/github.com/minishift/minishift/
+    - make prerelease
+    - rm -rf $HOME/gopath/src/github.com/minishift/minishift/out/bindata
+    - mv $HOME/gopath/src/github.com/minishift/minishift/out/ $CURRENT_BUILD_PATH/public
+  artifacts:
+    paths:
+      - public
+
+# git is needed for glide
```
budhrg: This looks like random comment and you forgot to remove it.

cmd/minishift/cmd/oc_env_test.go
```diff
@@ -23,7 +23,7 @@ import (
 )
 
 func Test_unix_oc_path(t *testing.T) {
-	shellConfig, err := getOcShellConfig("/Users/john/.minishift/cache/oc/v1.5.0/oc", "")
+	shellConfig, err := getOcShellConfig("/Users/john/.minishift/cache/oc/v1.5.0/oc", "bash")
```
budhrg: Fair enough.

.gitlab-ci.yml
gbraad: It is not random comment. It seems git is needed for glide to work. but it can be removed.

praveenkumar:
@gbraad Can you explain what value we are going to add with this PR? Are we going to add another CI on gitlab side?

gbraad:
I use GitLab myself a lot. For me, this means I do not have to use a build-wrapper which I manually trigger.

For community. this would mean they can easily set up a complete environment that can perform the same builds as we do, without using 3rd party services.

gbraad:
@praveenkumar none of the builds from yesterday mentioned this: https://travis-ci.org/minishift/minishift/builds/261835519

praveenkumar:
> None of the builds from yesterday mentioned

https://github.com/minishift/minishift/commit/9d36b5c261daec20dcdda3539b9d391e56bf225a this is where it happen. Yesterday @budhrg and @hferentschik discussing it not sure if that merged yestereday or on Friday.

gbraad:
I mean, they do not report it in their errors/output when building. While `make` on Fedora reported:
```
make: Nothing to be done for 'fmtcheck'.
make: Nothing to be done for 'cross'.
make: Nothing to be done for 'test'.
```

hferentschik:
> For community. this would mean they can easily set up a complete environment that can perform the same builds as we do, without using 3rd party services.

I am confused. What do you mean? Why are you saying no 3rd party services? They would be using hosted GitLab, right? Or are you expecting them to install GitLab locally? 

Regarding same environment and build, if you fork the minishift repo you automatically get for example Travis builds. What is the benefit for your of GitLab over any of the configured CI systems. Each of then allows you to build Minishift from your own fork? (just trying to understand)

gbraad:
Hosted or on-premise makes no difference in how you can use GitLab. For
hosted you can use shared or personal runners.

gbraad:
What is the harm of adding this definition file?

Note: maintenance is not an issue as I will run it daily.

hferentschik:
> What is the harm of adding this definition file?

No harm directly, just adding more "code" which needs to be "understood". I am neutral on this one. I just wanted to understand where you are coming from and why the existing options are not "working" for you.

gbraad:
the other options are working, but not a preference. I use a GitLab instance at home and the hosted.

At the moment I have to resolve to a build-wrapper. This would not be a problem if GitLab would have resolved: https://gitlab.com/gitlab-org/gitlab-ce/issues/15041, but until now they do not allow external build definitions.

Builds here runs with the provided RPMs from Fedora's repository for `golang` and `glide`.

gbraad:
Note: the understanding of the code could be simplified if Travis and GitLab share the same build script,

hferentschik:
@minishift/minishift-dev WDYT?

.gitlab-ci.yml
```diff
@@ -0,0 +1,18 @@
+build:
+  image: fedora:26
```
LalatenduMohanty: 👍

praveenkumar:
@hferentschik I don't see any issue for merging this PR if that helps any way to community.

LalatenduMohanty:
I was going through https://about.gitlab.com/features/gitlab-ci-cd/ and talked to @gbraad about it.  The USP of gitlab is that it can be used on the gitlab cloud as well with local gitlab instance. For example we can run the CI on a local machine through gitlab. I think that can come handy in future. Also as per PR it runs on fedora and thats a added advantage. May be we can try running it on CentOS too. 

I understand and share the concern that we will have 5 CI instances and that would be a PITA to maintain in long run.  The question is should we have travis or gitlab? At this point I am not sure how stable is gitlab but Travis has this limitation that we can only run on top of Ubuntu. So lets try both and after 1 month lets discuss if we want to remove one of them.

praveenkumar:
> I understand and share the concern that we will have 5 CI instances and that would be a PITA to maintain in long run. The question is should we have travis or gitlab? At this point I am not sure how stable is gitlab but Travis has this limitation that we can only run on top of Ubuntu. So lets try both and after 1 month lets discuss if we want to remove one of them.

I don't think we can have gitlab-ci running for a github repo https://forum.gitlab.com/t/is-there-a-way-to-trigger-a-gitlab-ci-build-from-github/2276/3 so having it 5 CI is out of question right now. What I understand is if a user using own private gitlab instance then for him it is easy to use gitlab ci which this PR make use of. 

@gbraad Did I understand it correct?

gbraad:
It is possible to trigger a build using a webhook from GitHub. The build definition for this would also have to change a little and would copy the build-wrapper idea I used before. I would not recommend going down this road. At the moment I do not say it is one of the main CI pipelines yet, and for now not going to replace Travis, but maybe it is a noble goal to attempt this over time. And I am more than happy to take this as a side project

hferentschik:
> I don't think we can have gitlab-ci running for a github repo https://forum.gitlab.com/t/is-there-a-way-to-trigger-a-gitlab-ci-build-from-github/2276/3 so having it 5 CI is out of question right now. 

+1, that's also my understanding. This is just about adding the config that it can be used, not about enabling another CI build integrated into GitHub.

> What I understand is if a user using own private gitlab instance then for him it is easy to use gitlab ci which this PR make use of.

+1

hferentschik:
> and for now not going to replace Travis

+1 Travis is a de facto standard of course. I would not like to replace it.

PR #1240, opened 2017-08-08
Issue #1241 Use 'make prerelease' target

### LINKED ISSUE #1241: Use 'make prerelease' target
When setting up the build definition for GitLab, I noticed from the error reporting that we actually use an unnecessary target for make in most of the CI:

  `make test cross fmtcheck prerelease`, while `make prerelease` (#1240) would have been enough.

Seems that this output is shallowed by all of the CI environments

gbraad:
Fixes #1241

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
 - export PATH=$PATH:/tmp/glide/${GLIDE_OS_ARCH}
 
 script:
-- make cross test fmtcheck prerelease
+- make prerelease
 
 notifications:
   webhooks: http://minibot.34e99f76.svc.dockerapp.io:9009/hubot/travis-ci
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -19,7 +19,7 @@ install:
 - export PATH=$PATH:/tmp/glide/${GLIDE_OS_ARCH}
 
 script:
-- make cross test fmtcheck prerelease
+- make prerelease
```
praveenkumar: LGTM 👍

gbraad:
if @budhrg is OK with this, we can merge

budhrg:
Yes, it is Ok. We updated the dependencies of `make prerelease` to have most of the checks to unburden `release` target. It looks more cleaner now. 

LGTM as well.

budhrg:
Merged and thanks

PR #1243, opened 2017-08-08
Issue #1147 Add registry add-on as part of default

### LINKED ISSUE #1147: Include registry-route addon in defaults
When you install default addons, this one is not installed, so it needs to be installed through directory install, which requires you to clone the minishift repo.

Test if it works fine and add it by default.

praveenkumar:
Fix #1147

### DIFF
--- a/cmd/minishift/cmd/openshift/registry.go
+++ b/cmd/minishift/cmd/openshift/registry.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 
 	"github.com/docker/machine/libmachine"
+	"github.com/minishift/minishift/cmd/minishift/cmd/addon"
 	"github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/openshift"
@@ -45,8 +46,12 @@ var registryCmd = &cobra.Command{
 		}
 
 		util.ExitIfNotRunning(host.Driver, constants.MachineName)
-
-		registryInfo, err := openshift.GetDockerRegistryInfo()
+		registryRoute := addon.GetAddOnManager().Get("registry-route")
+		registryAddonEnabled := true
+		if registryRoute == nil || !registryRoute.IsEnabled() {
+			registryAddonEnabled = false
+		}
+		registryInfo, err := openshift.GetDockerRegistryInfo(registryAddonEnabled)
 		if err != nil {
 			atexit.ExitWithMessage(1, err.Error())
 		}
--- a/pkg/minishift/addon/command/openshift_command.go
+++ b/pkg/minishift/addon/command/openshift_command.go
@@ -19,8 +19,9 @@ package command
 import (
 	"errors"
 	"fmt"
-	"github.com/minishift/minishift/pkg/minishift/openshift"
 	"strings"
+
+	"github.com/minishift/minishift/pkg/minishift/constants"
 )
 
 type OpenShiftCommand struct {
@@ -41,7 +42,7 @@ func (c *OpenShiftCommand) doExecute(ec *ExecutionContext) error {
 	fmt.Print(".")
 
 	commander := ec.GetDockerCommander()
-	_, err := commander.Exec("-t", openshift.OPENSHIFT_CONTAINER_NAME, openshift.OPENSHIFT_EXEC, ec.Interpolate(cmd))
+	_, err := commander.Exec("-t", constants.OpenshiftContainerName, constants.OpenshiftExec, ec.Interpolate(cmd))
 	if err != nil {
 		return errors.New(fmt.Sprintf("Error executing command '%s':", err.Error()))
 	}
--- a/pkg/minishift/constants/constants.go
+++ b/pkg/minishift/constants/constants.go
@@ -17,6 +17,8 @@ limitations under the License.
 package constants
 
 const (
-	B2dIsoAlias    = "b2d"
-	CentOsIsoAlias = "centos"
+	B2dIsoAlias            = "b2d"
+	CentOsIsoAlias         = "centos"
+	OpenshiftContainerName = "origin"
+	OpenshiftExec          = "/usr/bin/openshift"
 )
--- a/pkg/minishift/openshift/openshift.go
+++ b/pkg/minishift/openshift/openshift.go
@@ -19,14 +19,13 @@ package openshift
 import (
 	"errors"
 	"fmt"
+	"strings"
+
 	"github.com/golang/glog"
+	"github.com/minishift/minishift/pkg/minishift/constants"
 	"github.com/minishift/minishift/pkg/minishift/docker"
 	"github.com/pborman/uuid"
-	"strings"
 )
-
-const OPENSHIFT_CONTAINER_NAME = "origin"
-const OPENSHIFT_EXEC = "/usr/bin/openshift"
 
 type OpenShiftPatchTarget struct {
 	target              string
@@ -86,7 +85,7 @@ func (t *OpenShiftPatchTarget) localConfigFilePath() (string, error) {
 
 func RestartOpenShift(commander docker.DockerCommander) (bool, error) {
 	fmt.Println("Restarting OpenShift")
-	ok, err := commander.Restart(OPENSHIFT_CONTAINER_NAME)
+	ok, err := commander.Restart(constants.OpenshiftContainerName)
 	if err != nil {
 		return false, err
 	}
@@ -107,7 +106,7 @@ func Patch(target OpenShiftPatchTarget, patch string, commander docker.DockerCom
 	}
 
 	patchCommand := fmt.Sprintf("ex config patch %s --patch='%s'", containerConfigPath, patch)
-	result, err := commander.Exec("-t", OPENSHIFT_CONTAINER_NAME, OPENSHIFT_EXEC, patchCommand)
+	result, err := commander.Exec("-t", constants.OpenshiftContainerName, constants.OpenshiftExec, patchCommand)
 	if err != nil {
 		glog.Error("Creating patched configuration failed. Not applying the changes.", err)
 		return false, nil
@@ -136,7 +135,7 @@ func Patch(target OpenShiftPatchTarget, patch string, commander docker.DockerCom
 // IsRunning checks whether the origin container is in running state.
 // This method returns true if the origin container is running, false otherwise
 func IsRunning(commander docker.DockerCommander) bool {
-	status, err := commander.Status(OPENSHIFT_CONTAINER_NAME)
+	status, err := commander.Status(constants.OpenshiftContainerName)
 	if err != nil || status != "running" {
 		return false
 	}
@@ -150,7 +149,7 @@ func ViewConfig(target OpenShiftPatchTarget, commander docker.DockerCommander) (
 		return "", err
 	}
 
-	result, err := commander.Exec("-t", OPENSHIFT_CONTAINER_NAME, "cat", path)
+	result, err := commander.Exec("-t", constants.OpenshiftContainerName, "cat", path)
 	if err != nil {
 		return "", err
 	}
@@ -160,7 +159,7 @@ func ViewConfig(target OpenShiftPatchTarget, commander docker.DockerCommander) (
 func rollback(target OpenShiftPatchTarget, commander docker.DockerCommander, patchId string) {
 	fmt.Println("Unable to restart OpenShift after patchting it. Rolling back.")
 	restoreConfig(target, patchId, commander)
-	commander.Restart(OPENSHIFT_CONTAINER_NAME)
+	commander.Restart(constants.OpenshiftContainerName)
 	deleteBackup(target, patchId, commander)
 }
 
--- a/pkg/minishift/openshift/registry.go
+++ b/pkg/minishift/openshift/registry.go
@@ -24,7 +24,48 @@ import (
 	instanceState "github.com/minishift/minishift/pkg/minishift/config"
 )
 
-func GetDockerRegistryInfo() (string, error) {
+func GetDockerRegistryInfo(registryAddonEnabled bool) (string, error) {
+	var registryInfo string
+	var err error
+	if isRegistryRouteEnabled(registryAddonEnabled) {
+		registryInfo, err = fetchRegistryRoute()
+	} else {
+		registryInfo, err = fetchRegistryService()
+	}
+	return registryInfo, err
+}
+
+func isRegistryRouteEnabled(registryAddonEnabled bool) bool {
+	if !registryAddonEnabled {
+		return false
+	}
+	namespace := "default"
+	cmdArgText := fmt.Sprintf("get route docker-registry -n %s --config=%s", namespace, systemKubeConfigPath)
+	tokens := strings.Split(cmdArgText, " ")
+	cmdName := instanceState.InstanceConfig.OcPath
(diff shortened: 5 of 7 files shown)


hferentschik:
@praveenkumar I am confused, are the issues we have been experiencing with exposing the route resolved?

praveenkumar:
@hferentschik yes with v3.6.0 those issues are now resolved and we can expose registry as route using this addon. @jorgemoralespou is using it quite some time now and didn't face any issue. I also did try and it's working.

hferentschik:
> yes with v3.6.0 those issues are now resolved and we can expose registry as route using this addon. @jorgemoralespou is using it quite some time now and didn't face any issue. I also did try and it's working.

Without the issues around having to specify the port? I need to refresh my memory around this.

LalatenduMohanty:
> Without the issues around having to specify the port? I need to refresh my memory around this.

I think the port issue is still there (I have not tested yet) but the issues we reported when creating the add-on are resolved with 3.6.0 release.

praveenkumar:
> Without the issues around having to specify the port?

If you see the PR and merged registry addons we are still using 443 as port and with that combination it works.

hferentschik:
hmm, so the issue we originally had is still there then, right? Our aim was to make this work without specifying the port, right? AFAIR there was an issue in the registry. Can we programmatically determine the port? AFAICS you are hard coding it.

praveenkumar:
> hmm, so the issue we originally had is still there then, right?

AFAIR during 1.5.0 and earlier release we were not even able to use edge route along with port also but now we can use it.

> Our aim was to make this work without specifying the port, right? 

Yes, that's what we actually want. I need to dig some of edge route stuff.

> AFAIR there was an issue in the registry. Can we programmatically determine the port? 

I tried to get info around route and pod but didn't able to find that port details. Let me again dig a bit if something I can found or @csrwng can suggest something.

hferentschik:
Ok thanks.

hferentschik:
Let me try this before merging ...

pkg/minishift/openshift/registry.go
```diff
@@ -25,6 +25,44 @@ import (
...
+	var err error
+	if isRegistryRouteEnable() {
+		registryInfo, err = fetchRegistryRoute()
+	} else {
+		registryInfo, err = fetchRegistryService()
+	}
+	return registryInfo, err
+}
+
+func isRegistryRouteEnable() bool {
```
hferentschik: _ isRegistryRouteEnabled_

hferentschik: This func is imo problematic. There are wrong assumptions about cause and effect. You use _oc_ to see whether there is a route _docker-registry_ and then you basically conclude that this route is created by the add-on and hence later on you append the '443' port. What if the user exposed the route in a different way by manual configuration? 

I think I would feel better if we really check the add-on here, similar to what we do for persistent volumes and anyuid - https://github.com/minishift/minishift/blob/e65bd066c9ee5012abbf0cf162e297967bd45342/pkg/minishift/clusterup/clusterup.go#L211-L211. WDYT?

hferentschik:
@agajdosi, @amitkrout - here we have another candidate for a integration test. Being able to push directly to the OpenShift registry is a core feature which we don't properly cover by testing yet.

I think testing it via Cucumber is quite straight forward. Here is what one would do https://github.com/minishift/minishift/pull/704#issuecomment-293568580 (image to push and run can be discussed).

There is only one rub with this test. It need to use and run _docker_ on the host. So to run and pass this test, you need to have _docker_ on your PATH. I guess it would be fair to say, that we can safely assume that that's the case. On the various CI environment we can ensure anyways that the Docker client gets installed. For dev, I would think that most have _docker_ installed as well. Provided we make this test not run per default it would anyways not be an issue. Thoughts?

pkg/minishift/openshift/registry.go
hferentschik: @praveenkumar ^^

praveenkumar: > There are wrong assumptions about cause and effect. You use oc to see whether there is a route docker-registry and then you basically conclude that this route is created by the add-on and hence later on you append the '443' port. What if the user exposed the route in a different way by manual configuration?

@hferentschik , yes in that case this is problematic but what if a user use `addon apply` does that recorded by `AddOnManager` I didn't test it if that is true then will modify.

praveenkumar: @hferentschik I made that change but as I suspected if a user use `addon apply` than it is not going to work :(,

praveenkumar:
@hferentschik I made that change but as I suspected if a user use addon apply than it is not going to work :( but do a review.

hferentschik:
> I made that change but as I suspected if a user use addon apply than it is not going to work

Good point. There is always the alternative to add the add-on and not to change the registry command. Using the known IP is not going to stop working just because you install the add-on. The issue itself is also just asking for the add-on to be there per default, not that the registry command picks it up.

I really believe correctness goes first. If there is no definitive way to do this, I'd rather skip the code modifying the registry command.

pkg/minishift/openshift/registry.go
```diff
@@ -21,10 +21,53 @@ import (
...
+	var err error
+	if isRegistryRouteEnable(addOnManager) {
+		registryInfo, err = fetchRegistryRoute()
+	} else {
+		registryInfo, err = fetchRegistryService()
+	}
+	return registryInfo, err
+}
+
+func isRegistryRouteEnable(addOnManager *manager.AddOnManager) bool {
```
hferentschik: _isRegistryRouteEnabled_ (d in the end)

pkg/minikube/constants/constants.go
```diff
@@ -35,6 +35,8 @@ const (
 	DefaultCPUS                  = 2
 	DefaultDiskSize              = "20g"
 	UpdateMarkerFileName         = "updated"
+	OPENSHIFT_CONTAINER_NAME     = "origin"
```
hferentschik: -1, let's  not move this here. This is Minikube territory.

pkg/minishift/addon/command/openshift_command.go
```diff
@@ -41,7 +41,7 @@ func (c *OpenShiftCommand) doExecute(ec *ExecutionContext) error {
 	fmt.Print(".")
 
 	commander := ec.GetDockerCommander()
-	_, err := commander.Exec("-t", openshift.OPENSHIFT_CONTAINER_NAME, openshift.OPENSHIFT_EXEC, ec.Interpolate(cmd))
+	_, err := commander.Exec("-t", constants.OPENSHIFT_CONTAINER_NAME, constants.OPENSHIFT_EXEC, ec.Interpolate(cmd))
```
hferentschik: Why do you want to move this?

pkg/minishift/openshift/openshift.go
```diff
@@ -160,7 +158,7 @@ func ViewConfig(target OpenShiftPatchTarget, commander docker.DockerCommander) (
 func rollback(target OpenShiftPatchTarget, commander docker.DockerCommander, patchId string) {
 	fmt.Println("Unable to restart OpenShift after patchting it. Rolling back.")
 	restoreConfig(target, patchId, commander)
-	commander.Restart(OPENSHIFT_CONTAINER_NAME)
+	commander.Restart(constants.OPENSHIFT_CONTAINER_NAME)
```
hferentschik: -1

pkg/minishift/openshift/registry.go
```diff
@@ -21,10 +21,53 @@ import (
...
 
+	"github.com/minishift/minishift/pkg/minishift/addon/manager"
 	instanceState "github.com/minishift/minishift/pkg/minishift/config"
 )
 
-func GetDockerRegistryInfo() (string, error) {
+func GetDockerRegistryInfo(addOnManager *manager.AddOnManager) (string, error) {
+	var registryInfo string
+	var err error
+	if isRegistryRouteEnable(addOnManager) {
```
hferentschik: I find this more correct. It won't work in the case where the route add-on is applied (opposed to enabling it and applying it via start), but for the case where we get here as part of the enabled route add-on it is correct and in this case we can assume and hard code the port (443).

I could live with this. The alternative imo is to just include the add-on and make no changes to the registry command. I am somewhat 50/50 between these two alternatives.

@minishift/minishift-dev Thoughts?

pkg/minikube/constants/constants.go
budhrg: Looks like we need `constants` folder under `pkg/minishift` directory.

budhrg:
I think we need to add `$ eval $(minishift oc-env)` instruction to `minishift/registry-route.addon` file here https://github.com/minishift/minishift/blob/master/addons/registry-route/registry-route.addon#L17.

I got the error while testing.

```
$ minishift addons list

$ minishift addons enable registry-route
Add-on 'registry-route' enabled

$ minishift start
...

-- Applying addon 'registry-route':........
Add-on 'registry-route' created docker-registry route. Please run following commands to login to the OpenShift docker registry:
$ docker login -u developer -p `oc whoami -t` docker-registry-default.192.168.42.175.nip.io:443

$ eval $(minishift oc-env)

$ docker login -u developer -p `oc whoami -t` docker-registry-default.192.168.42.175.nip.io:443
could not read CA certificate "/etc/docker/ca.pem": open /etc/docker/ca.pem: no such file or directory
```

gbraad:
> $ docker login -u developer -p `oc whoami -t`
docker-registry-default.192.168.42.175.nip.io:443
> could not read CA certificate "/etc/docker/ca.pem": open
/etc/docker/ca.pem: no such file or directory

This looks like me a familiar issue from fedora 25. Check the issue tracker
for Fedora and re-using docker.

-- 

   [ Doing Open Source Matters ]

pkg/minikube/constants/constants.go
hferentschik: Not really or better not necessarily.

hferentschik:
> I think we need to add $ eval $(minishift oc-env)

I don't think this would help in the case of your error.

agajdosi:
@hferentschik I am a bit late with reply about integration test. I will take a look on the comment you provided later this day and will create an issue for creation of this feature / set of scenarios.

> There is only one rub with this test. It need to use and run docker on the host.

I also think it should not be a problem, probably we could do a check if docker is on PATH and if not, we can skip the test.

hferentschik:
> I also think it should not be a problem, probably we could do a check if docker is on PATH and if not, we can skip the test.

I don't think we even need to add this check. I think we can take this as a prerequisite. Assuming this pre-requisite we can add a whole bunch of test we have not looked into yet, in particular the access to the OpenShift registry.

agajdosi:
Ack, sounds reasonable. Will also help with docker-env command. +1

pkg/minikube/constants/constants.go
praveenkumar: @hferentschik We do need to move it somewhere otherwise we can't import addon package to openshift package level because of cyclic dependencies not allowed. I do agree let's not use Minikube directory but we can have a constants for openshift pkg level.

budhrg:
> I don't think this would help in the case of your error.

@hferentschik, Yes it will not help in the certificate error but will help to prevent having error `bash: oc: command not found...`.

budhrg:
For record purpose, the fedora issue for certificate on reusing docker daemon https://github.com/minishift/minishift/issues/552

budhrg:
I passed the certificate error by adding following in `/etc/sysconfig/docker` : 
```
if [ -z "${DOCKER_CERT_PATH}" ]; then
    DOCKER_CERT_PATH=/etc/docker
fi
```

hferentschik:
>  Yes it will not help in the certificate error but will help to prevent having error bash: oc: command not found....

Ok, got you. Fair enough.

hferentschik:
> Anyone able to get it work?

The exact steps provided work for me. Not quite sure what the issue is in your case. It might be related to the other issue you have had.

The question which no one seems to take position to though is whether we should just add the add-on and be done with it or whether we should do what @praveenkumar did and try to make adjustments to `minishift openshift registry`. And if the latter, in which form.

cmd/minishift/cmd/openshift/registry.go
```diff
@@ -46,7 +47,7 @@ var registryCmd = &cobra.Command{
 
 		util.ExitIfNotRunning(host.Driver, constants.MachineName)
 
-		registryInfo, err := openshift.GetDockerRegistryInfo()
+		registryInfo, err := openshift.GetDockerRegistryInfo(addon.GetAddOnManager())
```
hferentschik: I think the problem lies here (regarding circular dependencies and all). Why do you pass the AddOnManager? This creates an odd dependency in the openshift package to add-ons. All you want to know is whether the route add-on is enabled, right? So why don't just pass a bool and do the _isRegistryRouteEnabled_ here in the command.

pkg/minishift/openshift/registry.go
```diff
@@ -21,10 +21,53 @@ import (
...
+		registryInfo, err = fetchRegistryRoute()
+	} else {
+		registryInfo, err = fetchRegistryService()
+	}
+	return registryInfo, err
+}
+
+func isRegistryRouteEnable(addOnManager *manager.AddOnManager) bool {
+	registryRoute := addOnManager.Get("registry-route")
+	if registryRoute == nil || !registryRoute.IsEnabled() {
```
hferentschik: Is this not enough?

pkg/minishift/openshift/registry.go
```diff
@@ -21,10 +21,53 @@ import (
...
+		return false
+	}
+	namespace := "default"
+	cmdArgText := fmt.Sprintf("get route docker-registry -n %s --config=%s", namespace, systemKubeConfigPath)
```
hferentschik: What's this really doing? You call whether 'get route docker-registry' succeeds. Is this as an additional check to ensure the add-on application worked?

cmd/minishift/cmd/openshift/registry.go
hferentschik: Just pass 'addOnManager.Get("registry-route")' to openshift.GetDockerRegistryInfo. This is better encapsulation and it most likely will fix your dependency issues.

pkg/minishift/openshift/registry.go
praveenkumar: @hferentschik yes this is just an additional check if addon works and route is enabled.

pkg/minishift/constants/constants.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package constants
 
 const (
-	B2dIsoAlias    = "b2d"
-	CentOsIsoAlias = "centos"
+	B2dIsoAlias              = "b2d"
+	CentOsIsoAlias           = "centos"
+	OPENSHIFT_CONTAINER_NAME = "origin"
```
hferentschik: :-)

hferentschik: > Why are the constants inconsistent, by using all caps?

fair point

budhrg:
> The exact steps provided work for me. Not quite sure what the issue is in your case. It might be related to the other issue you have had.

@hferentschik , might be issue with my machine.

cben:
> /etc/docker/ca.pem: no such file or directory

For others googling this, this still happens on Fedora 30 docker, which is apparently too old to understand DOCKER_CERT_PATH :shrug:
You have to use `docker --tlscacert=$DOCKER_CERT_PATH/ca.pem --tlscert=$DOCKER_CERT_PATH/cert.pem --tlskey=$DOCKER_CERT_PATH/key.pem ...`

Issue #1242, opened 2017-08-08
Minishift unable to add impersonation 'sudoer' role when try to provision OpenShift =< v1.5.1


praveenkumar:
We have now a bug because we are using v3.6 oc version to provision openshift < 3.6 and when we try to provision v1.5.1 it failed to add sudoer role.

Code which we introduced for check openshift version and then according to it select which part of kubeconfig should be part of minishift_kubeconfig is broken. https://github.com/minishift/minishift/blob/master/pkg/minishift/clusterup/clusterup.go#L118 we might remove this whole addition code-base since now 3.6 is default.

```
The server is accessible via web console at:
    https://192.168.42.117:8443

You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin

Error during post cluster up configuration: Unable to add sudoer role
```

hferentschik:
Good catch. 

@agajdosi here we have something we need dearly some integration tests as well. They will be long running, but we should try to make sure that we can provision earlier (major) versions of OpenShift.

agajdosi:
@hferentschik +1. I was thinking about this also, this "minishift + OSO of nearly whatever version" seems like a problematic zone. If you have some ideas on this topic already, how test should like, how many versions etc, please create an issue. If not, I will create the issue later today or tomorrow.

hferentschik:
Merged via pull request #1245

PR #1245, opened 2017-08-08
Issue #1242 Fix sudoer role to provision openshift <= 1.5.1

praveenkumar:


pkg/minishift/clusterup/clusterup.go
```diff
@@ -106,14 +107,8 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
...
 func PostClusterUp(clusterUpConfig *ClusterUpConfig, sshCommander provision.SSHCommander, addOnManager *manager.AddOnManager) error {
-	isOpenshift3_6 := util.VersionOrdinal(clusterUpConfig.OpenShiftVersion) >= util.VersionOrdinal("v3.6.0")
-
-	clusterName := getConfigClusterName(clusterUpConfig.Ip, clusterUpConfig.Port)
-	userName := fmt.Sprintf("system:admin/%s", clusterName)
-	// See issue https://github.com/minishift/minishift/issues/1011. Needs to be revisted whether needed with final OpensShift 3.6 (HF)
-	if isOpenshift3_6 {
-		userName = fmt.Sprintf("system:admin/127-0-0-1:%d", clusterUpConfig.Port)
-	}
+	// See issue https://github.com/minishift/minishift/issues/1011. Needs to be revisted if anything change for oc cluster up side
```
budhrg: typo on `revisited`

praveenkumar: @budhrg ack, will fix it to next iteration.

hferentschik: @praveenkumar can you explain to me what's going on here? I think there are a few assumptions here which might be worth adding a comment about. 

How does this work? Pre 3.6 the username used contained the clustername/ip. In 3.6 it became 127-0-0-1. This is why we explicitly checked for >= 3.6 and if it was true we used 127-0-0-1. You removed this part now. The only way I can see this work is that if that if you are using oc >= 3.6 to provision a <= 1.5 version is not retrospectively uses  127-0-0-1 there as well. Is that correct? 

It also assumes that we use the default / baseline OpenShift version to provision versions of OpenShift which are <= this baseline. I recall that we discussed this in this form once.

I think we need to comment this somewhere. Not sure whether referencing #1011 provides the right context here.

hferentschik: Why not add an integration test to provision 1.5.1? That should be easy to do.

hferentschik: off, just noticed:

> Need to put integration test, have to co-ordinate with @agajdosi, would be good to have another features like provision_earlier_major_verion.feature?

+1 for integration test and dedicated feature

test/integration/features/provision_earier_version.feature
```diff
@@ -0,0 +1,42 @@
+@minishift-only
```
praveenkumar: @agajdosi @hferentschik Have a look if that looks as expected?

test/integration/features/provision_earier_version.feature
```diff
@@ -0,0 +1,42 @@
...
+     """
+
+  Scenario: User is able to do ssh into Minishift VM
+    Given Minishift has state "Running"
+     When executing "minishift ssh echo hello" succeeds
+     Then stdout should contain
+      """
+      hello
+      """
+
```
agajdosi: Maybe it could be useful to add deployment of an application.

test/integration/features/provision_earier_version.feature
agajdosi: @praveenkumar There is a typo in filename. Deployment of an application is an open question as it adds time to execution of this feature, but I would suggest to go for it.

Rest looks fine to me, big +1 for adding this test as part of PR.

praveenkumar: @agajdosi yes deployment take fair amount of time that's why I didn't add it here. I think we are mostly concern here if we can start MiniShift with a older version cleanly. @hferentschik @gbraad  wdyt?

test/integration/features/provision_earlier_version.feature
```diff
@@ -0,0 +1,42 @@
+@minishift-only
+Feature: Provision an older major release
```
hferentschik: +1 nice

test/integration/features/provision_earlier_version.feature
```diff
@@ -0,0 +1,42 @@
...
+  As a user I can provision an older major version of minishift
+  Scenario: Starting Minishift with v1.5.1
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start --openshift-version v1.5.1" succeeds
+     Then Minishift should have state "Running"
```
hferentschik: We should assert the version as well. That's in fact the most important here. Nothing in your tests proves that 1.5.1 is provisioned. It might be the flag is actually ignored and the default gets provisioned. None of the following tests would show that.

You should use:

    $ minishift openshift version

It is important to actually verify/test the core assertion we are trying to make. I

pkg/minishift/clusterup/clusterup.go
```diff
@@ -106,14 +107,10 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
...
 func PostClusterUp(clusterUpConfig *ClusterUpConfig, sshCommander provision.SSHCommander, addOnManager *manager.AddOnManager) error {
-	isOpenshift3_6 := util.VersionOrdinal(clusterUpConfig.OpenShiftVersion) >= util.VersionOrdinal("v3.6.0")
-
-	clusterName := getConfigClusterName(clusterUpConfig.Ip, clusterUpConfig.Port)
-	userName := fmt.Sprintf("system:admin/%s", clusterName)
-	// See issue https://github.com/minishift/minishift/issues/1011. Needs to be revisted whether needed with final OpensShift 3.6 (HF)
-	if isOpenshift3_6 {
-		userName = fmt.Sprintf("system:admin/127-0-0-1:%d", clusterUpConfig.Port)
-	}
+	// With oc 3.6 client, clusterName entry to kubeconfig file become 127.0.0.1:<port> and
```
hferentschik: +1

test/integration/features/provision_earlier_version.feature
praveenkumar: @hferentschik Agree, Added a scenario to verify it.

test/integration/features/provision_earlier_version.feature
```diff
@@ -0,0 +1,51 @@
...
+
+  Scenario: Provisioned Minishift have openshift v1.5.1
+     When executing "minishift openshift version" succeeds
+     Then stderr should be empty
+     And exitcode should equal "0"
+     And stdout should contain
+     """
+     openshift v1.5.1
+     """
```
praveenkumar: @hferentschik here you go

hferentschik: :-) nice

hferentschik:
retest this please

test/integration/features/provision_earlier_version.feature
gbraad: Is the Then-And block needed? I thought succeeds deals with this already

test/integration/features/provision_earlier_version.feature
```diff
@@ -0,0 +1,51 @@
...
+    and OpenShift web console should be accessible.
+    Given Minishift has state "Running"
+     Then status code of HTTP request to "OpenShift" at "/healthz" is equal to "200"
+      And body of HTTP request to "OpenShift" at "/healthz" contains "ok"
+      And status code of HTTP request to "OpenShift" at "/healthz/ready" is equal to "200"
+      And body of HTTP request to "OpenShift" at "/healthz/ready" contains "ok"
+      And status code of HTTP request to "OpenShift" at "/console" is equal to "200"
+      And body of HTTP request to "OpenShift" at "/console" contains "<title>OpenShift Web Console</title>"
+
+  Scenario: User has a pre-configured set of persistent volumes
```
hferentschik: We need to remove this scenario or enable anyuid. There is no guarantee that this tests passed otherwise (have failures on locally). See https://github.com/minishift/minishift/blob/master/pkg/minishift/clusterup/clusterup.go#L214

hferentschik:
Merged with some amendments to the test.

Issue #1217, opened 2017-08-02
Make Minishift update code resilient towards exact location of minishift binary 


hferentschik:
As part of issue #680, we wanted to change the structure of the release artefacts. However, [update.extractBinary](https://github.com/minishift/minishift/blob/88843dbfcdba668256fbf273b6bbc0b87cf5c277/pkg/minishift/update/update.go#L245-L245) does make explicit assumptions on where to find the _minishift_ binary. 

This prevents us from merging pull request #1212. 

To mitigate this, we need to change the implementation of _extractBinary_ to dynamically locate the _minishift_ binary. This change should then be released one release prior to a change for #680. Assuming the change will occur in 1.5.0, people can update to that version and then when 1.6.0 with a fix for #680 is released, they can seamlessly upgrade to 1.6.0 with the changed artefact layout.

As part of this issue we should also add a flag to the _update_ command which allows to "update" to a specific version. This will allow users who for example stayed on 1.4.0 to upgrade to 1.6.0 as well. For these users a 'minishift update' would fail, since their version of Minishift would not be able to handle the new layout. For them it would be:

    $ minishift update --to-version 1.5.0
    $ minishift update

Having the _--to-version_ might be handy feature to have regardless.

gbraad:
>  one release prior

safe bet... at least two releases prior.

hferentschik:
> safe bet... at least two releases prior.

sure. Good idea.

hferentschik:
Merged via pull request #1217. Note there is no '--to-version' flag. Really it would not help as described anyways.

PR #1253, opened 2017-08-09
Issue #1217 Enable update code resilient towards exact location of minishift binary

praveenkumar:
Todo
-------

- Need to update unit test to accommodate all those changes.


pkg/minishift/update/update.go
```diff
@@ -264,12 +267,34 @@ func extractBinary(downloadedArchivePath, archiveDir string) (string, error) {
...
+func minishiftBinaryPath(extractPath, binaryName string) (string, error) {
+	var binaryPath string
+	var err error
+	fileList := []string{}
+	err = filepath.Walk(extractPath, func(path string, f os.FileInfo, err error) error {
+		fileList = append(fileList, path)
+		return nil
+	})
+	for _, fileName := range fileList {
+		if strings.Contains(fileName, binaryName) {
```
budhrg: I doubt we can use this login in `filepath.Walk...` invocation above and avoid this loop. 

What are the content of `fileList` ?

hferentschik: I am not quite sure what @budhrg means with _login_, but afaict you walk the tree and append everything to a slice. Then you iterate the slice to see whether you got a match. Why not do everything in one go in the anonymous function passed to file path.Walk? Just check each entry whether it is a match. If so record the entry. After file path.Walk you just check whether you found something. 

Up to you what you want to do.

budhrg: Typo here :) It is `lines`. Corrected above.

praveenkumar:
>  would expect that we try at least with a current archive structure, the one we are aiming for (one more intermediate directory) and one with non existent binary.

I agree so currently test case contain first 2 and I will add non-existent binary tarball which should fail to provide binary path.

> Checking in all these binaries can blows up the repo quite quickly.

@hferentschik these tarballs are dummy if you check the size it's just same as a empty file.

hferentschik:
> these tarballs are dummy if you check the size it's just same as a empty file.

ahh, perfect. That's fine then. I did not check the size.

praveenkumar:
@hferentschik, @budhrg  Please do the review, I am done adding respective test cases.

hferentschik:
ok

hferentschik:
@praveenkumar Seems some of your tests are not working on Windows.

pkg/util/archive/archive.go
```diff
@@ -58,27 +58,50 @@ func Untar(tarball, targetDir string) error {
 
 	for {
 		header, err := tarReader.Next()
-		if err == io.EOF {
-			break
-		} else if err != nil {
+		switch {
```
hferentschik: what drives these changes? A bug in the untar?

pkg/minishift/update/update_test.go
```diff
@@ -84,15 +89,73 @@ func TestExtractBinary(t *testing.T) {
...
+		osName = "windows"
+	}
+	archiveName := fmt.Sprintf("minishift-%s-%s-%s.%s", "1.2.0", osName, "amd64", extName)
 	testArchivePath := filepath.Join(testDataDir, archiveName)
 	tmpArchivePath := filepath.Join(testDir, archiveName)
 	// Copy test archive to tmpDir for testing purpose
 	copyArchive(t, testArchivePath, tmpArchivePath)
-	extractedBinaryPath, err := extractBinary(tmpArchivePath, testDir)
+	fileName, err := extractBinary(tmpArchivePath, testDir)
+	t.Log(fileName)
```
hferentschik: t.Log?

pkg/minishift/update/update_test.go
```diff
@@ -84,15 +89,73 @@ func TestExtractBinary(t *testing.T) {
...
+	if !strings.HasSuffix(fileName, "minishift") {
+		t.Fatalf("Expected minishift binary path, Got : %s\n", fileName)
```
hferentschik: On Windows we have minishift.exe

pkg/minishift/update/update_test.go
```diff
@@ -84,15 +89,73 @@ func TestExtractBinary(t *testing.T) {
 		extName = "zip"
 		osName = "windows"
 	}
+	archiveName := fmt.Sprintf("minishift-%s-%s-%s.%s", "1.0.0", osName, "amd64", extName)
+	testArchivePath := filepath.Join(testDataDir, archiveName)
+	tmpArchivePath := filepath.Join(testDir, archiveName)
+	// Copy test archive to tmpDir for testing purpose
+	copyArchive(t, testArchivePath, tmpArchivePath)
+	fileName, err := extractBinary(tmpArchivePath, testDir)
+	t.Log(fileName)
```
hferentschik: I am not sure whether we should add more noise to the test output. Generally we should aim to have as little was possible get mixed into the test output.

pkg/minishift/update/update_test.go
```diff
@@ -84,15 +89,73 @@ func TestExtractBinary(t *testing.T) {
...
+	if !strings.HasSuffix(fileName, "minishift") {
```
hferentschik: minishift.exe on Windows

pkg/util/archive/archive.go
praveenkumar: @hferentschik yes the way we implemented it first time only thinking about our tarball structure which had files in first level but when I tried same implementation it break when files are underneath of a directory structure.

hferentschik: got you. makes sense

hferentschik:
Rebased and merged, thanks.

PR #1255, opened 2017-08-09
Issue #1244 Fix openshift version list

### LINKED ISSUE #1244: Minishift openshift version list should show supported list
Currently Minishift version list only show version >= our default set version (as below) but what we want to show openshift versions which can be provisioned (at least v1.4.0 onwards).

```
$ ./minishift openshift version list
	- v3.6.0
	- v3.6.0-alpha.0
	- v3.6.0-alpha.1
	- v3.6.0-alpha.2
	- v3.6.0-rc.0
	- v3.7.0-alpha.0
```

LalatenduMohanty:
Fixes #1244

### DIFF
--- a/pkg/minishift/openshift/version/openshift_versions.go
+++ b/pkg/minishift/openshift/version/openshift_versions.go
@@ -18,33 +18,20 @@ package version
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
-	"github.com/minishift/minishift/pkg/util"
 	"io"
 	"net/http"
+	"regexp"
 	"sort"
 	"strings"
-	"time"
+
+	"github.com/minishift/minishift/pkg/minishift/clusterup"
+	"github.com/minishift/minishift/pkg/util"
 )
 
 type ImageTags struct {
-	Count    int         `json:"count"`
-	Next     string      `json:"next"`
-	Previous interface{} `json:"previous"`
-	Results  []Results   `json:"results"`
-}
-
-type Results struct {
-	Name        string      `json:"name"`
-	FullSize    int         `json:"full_size"`
-	Images      []ImageInfo `json:"images"`
-	ID          int         `json:"id"`
-	Repository  int         `json:"repository"`
-	Creator     int         `json:"creator"`
-	LastUpdater int         `json:"last_updater"`
-	LastUpdated time.Time   `json:"last_updated"`
-	ImageID     interface{} `json:"image_id"`
-	V2          bool        `json:"v2"`
+	Name string `json:"name"`
 }
 
 type ImageInfo struct {
@@ -57,19 +44,17 @@ type ImageInfo struct {
 	OsFeatures   interface{} `json:"os_features"`
 }
 
-func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) {
+func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) error {
 	resp, err := getResponseBody("https://registry.access.redhat.com/v1/repositories/openshift3/ose/tags")
 	if err != nil {
-		fmt.Println("Error Occured", err)
-		return
+		return err
 	}
 	defer resp.Body.Close()
 	decoder := json.NewDecoder(resp.Body)
 	var data map[string]string
 	err = decoder.Decode(&data)
 	if err != nil {
-		fmt.Printf("%T\n%s\n%#v\n", err, err, err)
-		return
+		return errors.New(fmt.Sprintf("%T\n%s\n%#v\n", err, err, err))
 	}
 	fmt.Fprint(output, "The following OpenShift versions are available: \n")
 	var tagsList []string
@@ -88,36 +73,43 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) {
 	for _, tag := range tagsList {
 		fmt.Fprintf(output, "\t- %s\n", tag)
 	}
+	return nil
 }
 
-func PrintUpStreamVersions(output io.Writer, minSupportedVersion string) {
-	resp, err := getResponseBody("https://registry.hub.docker.com/v2/repositories/openshift/origin/tags/")
+func PrintUpStreamVersions(output io.Writer, minSupportedVersion string, defaultVersion string) error {
+	dockerRegistryUrl := "https://registry.hub.docker.com/v1/repositories/openshift/origin/tags"
+	resp, err := getResponseBody(dockerRegistryUrl)
 	if err != nil {
-		fmt.Fprintf(output, "Error Occured %s", err)
-		return
+		return err
 	}
 	defer resp.Body.Close()
 	decoder := json.NewDecoder(resp.Body)
-	var data ImageTags
+	var data []ImageTags
 	err = decoder.Decode(&data)
 	if err != nil {
-		fmt.Printf("%T\n%s\n%#v\n", err, err, err)
-		return
+		return errors.New(fmt.Sprintf("%T\n%s\n%#v\n", err, err, err))
 	}
 	fmt.Fprint(output, "The following OpenShift versions are available: \n")
 	var tagsList []string
-	for _, imageinfo := range data.Results {
+	for _, imageinfo := range data {
 		if strings.Contains(imageinfo.Name, "latest") {
 			continue
 		}
-		if util.VersionOrdinal(imageinfo.Name) >= util.VersionOrdinal(minSupportedVersion) {
-			tagsList = append(tagsList, imageinfo.Name)
+		if valid, _ := clusterup.ValidateOpenshiftMinVersion(imageinfo.Name, minSupportedVersion); valid {
+			if valid, _ := clusterup.ValidateOpenshiftMinVersion(imageinfo.Name, defaultVersion); valid {
+				tagsList = append(tagsList, imageinfo.Name)
+			} else {
+				if !isPrerelease(imageinfo.Name) {
+					tagsList = append(tagsList, imageinfo.Name)
+				}
+			}
 		}
 	}
 	sort.Strings(tagsList)
 	for _, tag := range tagsList {
 		fmt.Fprintf(output, "\t- %s\n", tag)
 	}
+	return nil
 }
 
 func getResponseBody(url string) (resp *http.Response, err error) {
@@ -126,4 +118,11 @@ func getResponseBody(url string) (resp *http.Response, err error) {
 		return nil, err
 	}
 	return resp, nil
+}
+
+func isPrerelease(tag string) bool {
+	if match, _ := regexp.MatchString("alpha|beta|rc", tag); match {
+		return true
+	}
+	return false
 }
--- a/cmd/minishift/cmd/openshift/list_versions.go
+++ b/cmd/minishift/cmd/openshift/list_versions.go
@@ -17,10 +17,15 @@ limitations under the License.
 package openshift
 
 import (
+	"os"
+
+	"fmt"
+
+	"github.com/minishift/minishift/pkg/minikube/constants"
 	openshiftVersions "github.com/minishift/minishift/pkg/minishift/openshift/version"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/minishift/minishift/pkg/version"
 	"github.com/spf13/cobra"
-	"os"
 )
 
 // getVersionsCmd represents the ip command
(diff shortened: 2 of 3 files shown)


pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -91,19 +92,23 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) {
 }
 
 func PrintUpStreamVersions(output io.Writer, minSupportedVersion string) {
-	resp, err := getResponseBody("https://registry.hub.docker.com/v2/repositories/openshift/origin/tags/")
+	dockerRegistryUrl := "https://registry.hub.docker.com/v2/repositories/openshift/origin/tags/?page_size=1024"
```
LalatenduMohanty: The URL got changed because the older URL did not contain all the tags for OpenShift origin images. Hence v1.4.1 did not show up.

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -91,19 +92,23 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) {
...
 		return
 	}
 	defer resp.Body.Close()
 	decoder := json.NewDecoder(resp.Body)
 	var data ImageTags
-	err = decoder.Decode(&data)
-	if err != nil {
-		fmt.Printf("%T\n%s\n%#v\n", err, err, err)
-		return
-	}
```
LalatenduMohanty: The New URL JSON has `last_updated` field empty, hence it was returning following error. I am not sure how should we handle the error as we do not care about the `last_updated` field. 
```
*time.ParseError
parsing time "null" as ""2006-01-02T15:04:05Z07:00"": cannot parse "null" as """
&time.ParseError{Layout:"\"2006-01-02T15:04:05Z07:00\"", Value:"null", LayoutElem:"\"", ValueElem:"null", Message:""}
```

LalatenduMohanty:
Here is the output of `openshift version list` with the PR
```
$ minishift openshift version list
	- v1.4.1
	- v1.5.0
	- v1.5.0-alpha.0
	- v1.5.0-alpha.1
	- v1.5.0-alpha.2
	- v1.5.0-alpha.3
	- v1.5.0-rc.0
	- v1.5.1
	- v3.6.0
	- v3.6.0-alpha.0
	- v3.6.0-alpha.1
	- v3.6.0-alpha.2
	- v3.6.0-rc.0
	- v3.7.0-alpha.0
```

pkg/minishift/openshift/version/openshift_versions.go
hferentschik: I am wondering whether we should v1 of the API - https://registry.hub.docker.com/v1/repositories/openshift/origin/tags. This seems to return in a more condensed format what we want and it is proper v1 API syntax. According to the docs there should be a _tags/list_ - https://docs.docker.com/registry/spec/api/#tags - which does not seem to work. _page_size_ does not seem to be documented anywhere. I am a bit worried that this will change/break.

cmd/minishift/cmd/openshift/list_version_test.go
```diff
@@ -0,0 +1,39 @@
+/*
+Copyright 2017 The Kubernetes Authors All rights reserved.
```
budhrg: Forgot to have Red Hat license here.

pkg/minishift/openshift/version/openshift_versions.go
budhrg: > https://registry.hub.docker.com/v1/repositories/openshift/origin/tags

All the tags seem to be there and less noisy. Each record only contains `name` and `layer` and would be fast to fetch.

> I am a bit worried that this will change/break.

Same here.

pkg/minishift/openshift/version/openshift_versions.go
budhrg: Will not be the case if use v1 url.

pkg/minishift/openshift/version/openshift_versions.go
LalatenduMohanty: > I am wondering whether we should v1 of the API - https://registry.hub.docker.com/v1/repositories/openshift/origin/tags. This seems to return in a more condensed format what we want and it is proper v1 API syntax.

V1 url has the same issue of some empty fields and gave below error, I un-comment the error checking code. 

```
$ minishift openshift version list
*json.UnmarshalTypeError
json: cannot unmarshal array into Go value of type version.ImageTags
&json.UnmarshalTypeError{Value:"array", Type:(*reflect.rtype)(0xa8b980), Offset:1}
```
Also V1 URL is not working at all with the current code as it is getting the above error when we try to parse it.

pkg/minishift/openshift/version/openshift_versions.go
LalatenduMohanty: For v1 I am getting following error
```
$ minishift openshift version list
*json.UnmarshalTypeError
json: cannot unmarshal array into Go value of type version.ImageTags
&json.UnmarshalTypeError{Value:"array", Type:(*reflect.rtype)(0xa8b980), Offset:1}
```

LalatenduMohanty:
> page_size does not seem to be documented anywhere. I am a bit worried that this will change/break.

@hferentschik Agree. This a risk. But I do not see a better alternative yet.

budhrg:
@LalatenduMohanty , With V1 Api, I need to make following changes
```
type ImageTags struct {
	Name  string `json:"name"`
	Layer string `json:"layer"`
}
# Deleted Results struct

# PrintUpstreamVersions
...
        var data []ImageTags

	err = decoder.Decode(&data)
	if err != nil {
		fmt.Printf("%T\n%s\n%#v\n", err, err, err)
		return
	}

	fmt.Fprint(output, "The following OpenShift versions are available: \n")
	var tagsList []string
	for _, imageinfo := range data {
		if strings.Contains(imageinfo.Name, "latest") {
			continue
		}
		if util.VersionOrdinal(imageinfo.Name) >= util.VersionOrdinal(minSupportedVersion) {
			tagsList = append(tagsList, imageinfo.Name)
		}
	}
....
```
And then working as expected. Please confirm.

pkg/minishift/openshift/version/openshift_versions.go
hferentschik: @LalatenduMohanty Sure, you will have to change the types into which you want to un-marshall accordingly. It will be much shorter in fact, since v1 layer and name. The former is irrelevant for us and afaict all tags are there.

pkg/minishift/openshift/version/openshift_versions.go
hferentschik: @LalatenduMohanty You are un-marshalling into a provided data structure - https://github.com/minishift/minishift/blob/29c6b76418c372882f00b29461b7d75195aa1464/pkg/minishift/openshift/version/openshift_versions.go#L101-L101 (ImageTags). You will need to change this to match the format of the v1 output.

LalatenduMohanty:
The current output looks like
```
$ minishift openshift version list
	- v1.4.1
	- v1.5.0
	- v1.5.1
	- v3.6.0
	- v3.6.0-alpha.0
	- v3.6.0-alpha.1
	- v3.6.0-alpha.2
	- v3.6.0-rc.0
	- v3.7.0-alpha.0
```

LalatenduMohanty:
I wanted to remove the alpha, rc builds for 3.6.0 also , but `util.VersionOrdinal(imageinfo.Name) > util.VersionOrdinal(defaultVersion)` returns true for alpha and rc tags of v3.6.0. So we need to fix `func VersionOrdinal` https://github.com/minishift/minishift/blob/master/pkg/util/utils.go#L91-L122 and if we want to fix that we should track that in a separate issue.

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -90,28 +76,35 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) {
...
 	fmt.Fprint(output, "The following OpenShift versions are available: \n")
 	var tagsList []string
-	for _, imageinfo := range data.Results {
+	for _, imageinfo := range data {
 		if strings.Contains(imageinfo.Name, "latest") {
 			continue
 		}
 		if util.VersionOrdinal(imageinfo.Name) >= util.VersionOrdinal(minSupportedVersion) {
-			tagsList = append(tagsList, imageinfo.Name)
+			if util.VersionOrdinal(imageinfo.Name) > util.VersionOrdinal(defaultVersion) {
```
praveenkumar: @LalatenduMohanty We should use semver logic for upstream which is part of clusterup code https://github.com/minishift/minishift/blob/master/pkg/minishift/clusterup/clusterup.go#L282 `VersionOrdinal` only for downstream because we don't follow there semver logic.

praveenkumar: Some thing like

```
if ok, _ := clusterup.ValidateOpenshiftMinVersion(imageinfo.Name, minSupportedVersion); ok {
			tagsList = append(tagsList, imageinfo.Name)
		}
```

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -19,32 +19,18 @@ package version
...
-	FullSize    int         `json:"full_size"`
-	Images      []ImageInfo `json:"images"`
-	ID          int         `json:"id"`
-	Repository  int         `json:"repository"`
-	Creator     int         `json:"creator"`
-	LastUpdater int         `json:"last_updater"`
-	LastUpdated time.Time   `json:"last_updated"`
-	ImageID     interface{} `json:"image_id"`
-	V2          bool        `json:"v2"`
+	Name  string `json:"name"`
```
hferentschik: technically this would not even be needed, since we are not interested in it

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -19,32 +19,18 @@ package version
...
+	Layer string `json:"layer"`
```
hferentschik: technically this would not even be needed, since we are not interested in it

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -90,28 +76,35 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) {
 	}
 }
 
-func PrintUpStreamVersions(output io.Writer, minSupportedVersion string) {
-	resp, err := getResponseBody("https://registry.hub.docker.com/v2/repositories/openshift/origin/tags/")
+func PrintUpStreamVersions(output io.Writer, minSupportedVersion string, defaultVersion string) {
+	dockerRegistryUrl := "https://registry.hub.docker.com/v1/repositories/openshift/origin/tags"
+	resp, err := getResponseBody(dockerRegistryUrl)
 	if err != nil {
```
hferentschik: This is actually not quite right here. _PrintUpStreamVersions_ should really propagate the error and let the caller handle it. The caller can then decide whether to print nothing, to abort or whatever is appropriate. Right now the method just returns without any indication on whether it worked or not. Well, it prints a message, but this is not really its task. This applies to all errors in this function.

pkg/minishift/openshift/version/openshift_versions.go
LalatenduMohanty: @hferentschik you are right. I forgot to remove it.

hferentschik: it does not matter really. Just wanted to point it out after all the discussion we had.

budhrg: @hferentschik I think you are talking about `ImageInfo` struct.

hferentschik: No ImageTags. One does not have to map all fields of the JSON into the struct. You can map what you need. Other fields will be ignored.

hferentschik:
merged, thanks.

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -88,36 +73,43 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) {
...
 	fmt.Fprint(output, "The following OpenShift versions are available: \n")
 	var tagsList []string
-	for _, imageinfo := range data.Results {
+	for _, imageinfo := range data {
 		if strings.Contains(imageinfo.Name, "latest") {
 			continue
 		}
-		if util.VersionOrdinal(imageinfo.Name) >= util.VersionOrdinal(minSupportedVersion) {
-			tagsList = append(tagsList, imageinfo.Name)
+		if valid, _ := clusterup.ValidateOpenshiftMinVersion(imageinfo.Name, minSupportedVersion); valid {
```
hferentschik: imo, ValidateOpenshiftMinVersion should have been moved into another package. It feels wrong to have a dependency to the clusterup package here. Also, ValidateOpenshiftMinVersion does not really tell what this method is doing. But anyways, ...

LalatenduMohanty: @hferentschik I agree. Specially with the name `ValidateOpenshiftMinVersion`.

LalatenduMohanty: @hferentschik #1261

gbraad:
Please Upstream and Downstream are used as a single term, do not break them
up, as Print Up and Print Down sound like phrasal verbs

-- 

   [ Doing Open Source Matters ]

PR #1265, opened 2017-08-14
Issue #1203 Install default add-ons if addons directory is not present

### LINKED ISSUE #1203: Behaviour of automatic installation of addons
I have tried to run coolstore.feature test, which failed to me. The problem is that between runs of integration tests, the integration-test folder is not being deleted. Due to this fact the addons folder is not being populated by default addons - aka automatic installation of default addons. 

Question is: **In which cases we want minishift to automatically install addons?**

Currently it is:

- minishift home folder does not exist
- minishift home folder is completely empty

Should it also trigger automatic installation when "addons" folder is not present in minishift home? Now it just creates empty folder called "addons".

LalatenduMohanty:
Fixes #1203 

### DIFF
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -79,12 +79,12 @@ var RootCmd = &cobra.Command{
 	Long:  `Minishift is a command-line tool that provisions and manages single-node OpenShift clusters optimized for development workflows.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		var (
-			err          error
-			isFreshStart bool
+			err                    error
+			isAddonInstallRequired bool
 		)
 
-		if !filehelper.Exists(constants.Minipath) || filehelper.IsEmptyDir(constants.Minipath) {
-			isFreshStart = true
+		if !filehelper.Exists(constants.MakeMiniPath("addons")) {
+			isAddonInstallRequired = true
 		}
 
 		for _, path := range dirs {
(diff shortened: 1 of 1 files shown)


cmd/minishift/cmd/root.go
```diff
@@ -79,12 +79,12 @@ var RootCmd = &cobra.Command{
...
 		var (
-			err          error
-			isFreshStart bool
+			err                    error
+			isAddonInstallRequired bool
 		)
 
-		if !filehelper.Exists(constants.Minipath) || filehelper.IsEmptyDir(constants.Minipath) {
-			isFreshStart = true
+		if !filehelper.Exists(constants.MakeMiniPath("addons")) {
```
budhrg: @hferentschik , are we taking this path now instead of what we decided as part of PR https://github.com/minishift/minishift/pull/990.

hferentschik: Works for me. The big issue I am having with this code, is the use of constants.MakeMiniPath("addons"). This is really a misuse of the constants and very error prone. However we are not only doing this here, but for other paths as well. I am hoping to be able to address this once we have some sort of MinishiftPaths struct which gives typed access to the various paths in $MINISHIFT_HOME. AFAIK @LalatenduMohanty is creating this struct as part of the profile feature. 

I already see this code breaking due to this. Really we should have named these variables a long time ago in the constant files, e.g. constants.AddonPath or something. Same for the other paths. Using constants.MakeMiniPath("addons") is just wrong.

hferentschik:
Merged, thanks.

cmd/minishift/cmd/root.go
LalatenduMohanty: @hferentschik Yes, I will fix this as part of #126.

Issue #1269, opened 2017-08-16
Che fails on version 1.4.1


gorkem:
Che has  a service named che-host that has a route configured. When the service is trying to reach itself through the internal names it fails, public route works fine. When I ssh into the pod and try to curl to http://che-host it fails after a long wait where as http://localhost and with internal ip address works fine.

Also when I enable the logs I see lines like follow for every pod.
`1610 docker_sandbox.go:263] Couldn't find network status for eclipse-che/che-3-deploy through plugin: invalid network status for`

We also did a `minishift ssh` and compared `/etc/resolv.conf` with 1.3.1 and 1.4.1 only has a single `nameserver` entry where as 1.3.1 also includes and entry for search.

@amisevsk  Anything else I am missing?

gbraad:
Interesting, I'll have a look. BTW, did you use the B2D or CentOS image?

gorkem:
I seem to have b2d

gbraad:
I did a quick run of the version you mentioned, using B2D, but I never get an entry for 'search' in `/etc/resolv.conf`

* Minishift v.1.4.1
  * OpenShift v3.6 - `nameserver 192.168.122.1`
  * OpenShift v1.5.1 -  `nameserver 192.168.122.1`
* Minishift v1.3.1
  * OpenShift v1.5.1 - `nameserver 192.168.122.1`

gorkem:
Hmmm. I do not have those entries anymore on my 1.3.1 either

amisevsk:
I've managed to narrow this down somewhat.

Running minishift 1.4.1 and Openshift version
```
OpenShift Master: v3.6.0+c4dd4cf
Kubernetes Master: v1.6.1+5115d708d7
```
I am able to reproduce the issue. Running minishift 1.4.1 and OpenShift v1.5.1 it does not occur.

However, the issue is actually that **pods cannot resolve their own service or service's clusterIP**. Starting a second pod, I can curl che-host without issue, but from within che host I cannot. There are no real networking issues except that pods cannot access their own services -- localhost and external work fine.

After a bit of digging, I came across this section of [kubernetes documentation](https://kubernetes.io/docs/tasks/debug-application-cluster/debug-service/#a-pod-cannot-reach-itself-via-service-ip) that seems related. 

@gbraad Is there a setting that has changed between 1.5.1 and 3.6? I don't see this issue on OpenShift Online, running
```
OpenShift Master: v3.6.173.0.5 (online version 3.5.0.20)
[...]

PR #1272, opened 2017-08-16
Issue #1195 - Add integration test for flags

agajdosi:
Addresses issue #1195

### DIFF
--- a/test/integration/features/coolstore.feature
+++ b/test/integration/features/coolstore.feature
@@ -13,7 +13,7 @@ Feature: Cool Store
       """
       Add-on 'xpaas' enabled
       """
-     When executing "minishift start --memory 4096 --iso-url centos" succeeds
+     When executing "minishift start --memory 4096" succeeds
      Then Minishift should have state "Running"
 
   Scenario: User creates a project and deploys CoolStore
--- a/test/integration/features/flags.feature
+++ b/test/integration/features/flags.feature
@@ -1,0 +1,82 @@
+@flags
+Feature: Flags
+  As a user I can perform advanced operations
+  such as starting Minishift with optional features
+
+  Scenario: Set configuration options for Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift config set docker-opt dns=8.8.8.8" succeeds
+     Then executing "minishift config get docker-opt" succeeds
+      And stdout should contain
+      """
+      [dns=8.8.8.8]
+      """
+     When executing "minishift config set docker-env BAZ=BAT" succeeds
+     Then executing "minishift config get docker-env" succeeds
+      And stdout should contain
+      """
+      [BAZ=BAT]
+      """
+
+  Scenario: Starting Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start --insecure-registry test-registry:5000 --docker-env=FOO=BAR" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: User is able to get value of the insecure registry
+    Given Minishift has state "Running"
+     When executing "minishift ssh docker info" succeeds
+     Then stdout should contain
+      """
+      test-registry:5000
+      """
+
+  @b2d-only
+  Scenario: User is able to get value of the docker environment variables
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
+     Then stdout should contain
+      """
+      export "FOO=BAR"
+      """
+      And stdout should not contain
+      """
+      export "BAZ=BAT"
+      """
+
+  @rhel-only
+  Scenario: User is able to get value of the docker environment variables
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /etc/systemd/system/docker.service.d/10-machine.conf" succeeds
+     Then stdout should contain
+      """
+      Environment=FOO=BAR
+      """
+      And stdout should not contain
+      """
+      Environment=BAZ=BAT
+      """
+
+  @b2d-only
+  Scenario: User is able to get the value of docker optional parameters
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /var/lib/boot2docker/profile" succeeds
+     Then stdout should contain
+      """
+      --dns=8.8.8.8
+      """
+
+  @rhel-only
+    Scenario: User is able to get the value of docker optional parameters
+    Given Minishift has state "Running"
+     When executing "minishift ssh cat /etc/systemd/system/docker.service.d/10-machine.conf" succeeds
+     Then stdout should contain
+      """
+      --dns=8.8.8.8
+      """
+
+  Scenario: Deleting Minishift
+    Given Minishift has state "Running"
+     When executing "minishift delete" succeeds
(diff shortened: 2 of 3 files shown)


budhrg:
add to whitelist

budhrg:
And what about this PR https://github.com/minishift/minishift/pull/1196 ?

agajdosi:
@budhrg: Today @gbraad suggested on IRC to create a new PR, it will make the progress faster. Old one could be probably closed later.

centos_ci.sh
```diff
@@ -305,7 +305,7 @@ function build_and_test() {
   cd $GOPATH/src/github.com/minishift/minishift
   make prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags ~coolstore -format pretty"
+  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags flags,basic -format pretty"
```
praveenkumar: This mean we are not going to execute coolstore test case at all? It might be good for PR but I think at least for nightly we should have it.

test/integration/features/coolstore.feature
```diff
@@ -13,7 +13,7 @@ Feature: Cool Store
       """
       Add-on 'xpaas' enabled
       """
-     When executing "minishift start --memory 4096 --iso-url centos" succeeds
+     When executing "minishift start --memory 4096" succeeds
```
praveenkumar: 👍

test/integration/integration_test.go
```diff
@@ -555,3 +583,18 @@ func proxyLogShouldContain(expected string) error {
 func proxyLogShouldContainContent(expected *gherkin.DocString) error {
 	return compareExpectedWithActualContains(expected.Content, testProxy.GetLog())
 }
+
+func CatDockerConfigFile() error {
```
praveenkumar: 👍 If this function is not needed to be access outside of this scope then better to name it as `catDockerConfigFile()`

centos_ci.sh
gbraad: CoolStore should be enabled. It might have been able to detect #1269 earlier. Although haven't confirmed this.

test/integration/integration_test.go
agajdosi: true, will fix it

centos_ci.sh
hferentschik: I am also intrigued about this. So far only coolstore was excluded, everything else ran. Now it is just basic and flags? What's about cmd-version, proxy and provision_earlier_version? Can we not stick to ~coolstore for now? Or have we reached a limit where the current timeout limit is reached?

test/integration/features/flags.feature
```diff
@@ -0,0 +1,51 @@
+@flags
+Feature: Flags
+  As a user I can perform advanced operations
+  such as starting Minishift with optional features
+
+  Scenario: Set configuration options for Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift config set docker-opt dns=8.8.8.8" succeeds
+     Then executing "minishift config get docker-opt" succeeds
```
hferentschik: I think this is more of a 'And' - When I can set the config and and I get retrieve then, then ...

test/integration/features/flags.feature
```diff
@@ -0,0 +1,51 @@
...
+      And stdout should contain
+      """
+      [dns=8.8.8.8]
+      """
+     When executing "minishift config set docker-env BAZ=BAT" succeeds
+     Then executing "minishift config get docker-env" succeeds
```
hferentschik: 'And'?

test/integration/features/flags.feature
```diff
@@ -0,0 +1,51 @@
...
+      And stdout should contain
```
hferentschik: 'Then'

centos_ci.sh
agajdosi: @praveenkumar @gbraad This was to make sure that it runs fine, not sure now what we agreed to run on PRs, is it basic+coolstore+version or even with flags?

test/integration/integration_test.go
```diff
@@ -68,8 +69,30 @@ func TestMain(m *testing.M) {
...
 	if runner.IsCDK() {
 		godogTags += "~minishift-only"
+		isoName = "rhel"
+		fmt.Println("Test run using CDK binary with RHEL iso.")
 	} else {
 		godogTags += "~cdk-only"
+		switch os.Getenv("MINISHIFT_ISO_URL") {
+		case "b2d":
+			fmt.Println("Test run using Boot2Docker iso image.")
+			isoName = "b2d"
```
hferentschik: ahh, that's nice. We use this to determine the ISO type. Works I guess while we use short names. Should be good enough. I was initially thinking of determining the ISO dynamically using the release info content of the VM.

centos_ci.sh
gbraad: On PRs it is not advised to run CoolStore. This would slow everything down considerably

test/integration/integration_test.go
```diff
@@ -68,8 +69,30 @@ func TestMain(m *testing.M) {
...
+		case "centos":
+			fmt.Println("Test run using CentOS iso image.")
+			isoName = "centos"
+		case "":
+			fmt.Println("Test run using Boot2Docker iso image.")
+			isoName = "b2d"
+		default:
+			envIso := os.Getenv("INTEGRATION_ISO")
```
hferentschik: I guess here is where things fall apart using the MINISHIFT_ISO_URL. Is this because you are using file://... when running integration tests. Are you requiring in this case to set another variable? I guess here the dynamic approach would be better. Would it not also be possible to do a regexp check of the iso name and decide based on that. Seems "nicer" than introducing a new variable.

test/integration/integration_test.go
```diff
@@ -555,3 +583,18 @@ func proxyLogShouldContain(expected string) error {
 func proxyLogShouldContainContent(expected *gherkin.DocString) error {
 	return compareExpectedWithActualContains(expected.Content, testProxy.GetLog())
 }
+
+func CatDockerConfigFile() error {
+	var err error
+	if isoName == "b2d" {
```
hferentschik: :+1

test/integration/features/flags.feature
```diff
@@ -0,0 +1,51 @@
...
+    Given Minishift has state "Running"
+     When executing "minishift ssh -- docker info"
+     Then stdout should contain
+      """
+      test-registry:5000
+      """
+
+  Scenario: User is able to get value of the docker environment variables
+    Given Minishift has state "Running"
+     When user prints content of docker daemon configuration into stdout
```
hferentschik: "When printing Docker daemon configuration to stdout"

I avoid "user" in this case. And one prints to stdout

test/integration/integration_test.go
agajdosi: @hferentschik +1 for regexp check, did not realized that. It seems like enough coverage, one can use her own iso to test, however it is required at least include b2d/boot2docker or centos in filename. Seems fair to me.

hferentschik:
Merged, thanks.

Issue #1136, opened 2017-07-14
Provide an easy way to have multiline descriptions in add-ons


jorgemoralespou:
When having an addon with a long description, you need to write it as a single line:

e.g.:

````
# Name: prometheus
# Description: This template creates a Prometheus instance preconfigured to gather OpenShift and Kubernetes platform and node metrics and report them to admins. It is protected by an OAuth proxy that only allows access for users who have view access to the prometheus namespace. You may customize where the images (built from openshift/prometheus and openshift/oauth-proxy) are pulled from via template parameters.
````

It would be nice to allow for better readability with something that allowed multiline:

`````
# Name: prometheus
# Description: This template creates a Prometheus instance preconfigured to gather OpenShift and \
                       Kubernetes platform and node metrics and report them to admins. It is protected by an \
                       OAuth proxy that only allows access for users who have view access to the prometheus \
                       namespace. You may customize where the images (built from openshift/prometheus \
                       and openshift/oauth-proxy) are pulled from via template parameters.
````

NOTE: Notation is not even a recommendation ;-)

hferentschik:
I had this on my todo list as well.

hferentschik:
@praveenkumar The code is here https://github.com/minishift/minishift/blob/29b2959185e716056cf85ae5c6b81c6d354ad8c4/pkg/minishift/addon/parser/addon_parser.go#L182-L182

As far as I can tell, all header lines prefixed with '#' gets passed to _createMetaData_. It then created a map of meta tag to meta value. This is using a regexp for now which works on a line by line basis. This would need to change. How depends on how we want to support multi lines. Do we require a '\' at the end to indicate continuation? Or do we want to have additional indentation on the following lines? Or nothing of this and all lines following a "Foo: bla bla" will be "merged" into the "Foo" meta until either a new tag appears of the end of header is reached. 

Up to us how to change this.

gbraad:
You suggest to implement this as a state machine? Where state changes when
a different recognizable instruction has been detected.

-- 

   Gerard Braad | http://gbraad.nl
   [ Doing Open Source Matters ]

hferentschik:
Pretty much. Start iterating the lines. If you detect a meta tag start collecting all following lines within this metatag until you reach a new meta tag or hit another condition which wold indicate that the current meta tag data is complete (for example the line is not indented anymore)

hferentschik:
Merged via pull request #1280

PR #1280, opened 2017-08-18
Issue #1136 Enable multiline description

praveenkumar:


pkg/minishift/addon/parser/addon_parser_test.go
```diff
@@ -51,6 +52,15 @@ var noDescription string = `# Name: One command of each type
 oc foo
 `
 
+var addOnWithMultilineDescription string = `# Name: foo
+# Description: This is one line
+#   This is second line
```
hferentschik: Do we need the indentation? You are suggesting that for multiline descriptionm the successive lines needs to start with 3 spaces (or is it a tab)? 

I was wondering whether it would be easier (for the user) and also nicer to read, if no indentation would be required. 

     # Description: This is one line
     # This is second line
     # This is the third line
     # Required-Vars: This is a new meta header

During parsing of the header you start a new meta tag when you encounter the first 'XYZ:'. Then you aggregate all following lines into the same header until you hit the next tag 'XYA'.

The potential catch is that without indentation you might have issues with lines containing the ':' itself. But this problem exists already. It would probably be a good idea to harden the regexp '(`^# ?(.*):(.*)`)' a bit.

hferentschik: Maybe we can add some more tests as well. Some tests with other meta tags following the multi lines description and if we decide to use indentation, I would like to see a test for something like:

    # Name: foo
    # Description: line one
    #   line two
    # No indentation - what happens with this line
    #  Required-Vars: ...

hferentschik: Either way this needs to be documented in the docs. If nothing else than a simple TIP in the add-on topic.

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -182,10 +182,17 @@ func (parser *AddOnParser) parseCommands(scanner *bufio.Scanner) ([]command.Comm
 func createMetaData(header []string) (addon.AddOnMeta, error) {
 	regex, _ := regexp.Compile(`^# ?(.*):(.*)`)
 	metaMap := make(map[string]string)
+	var key string
 	for _, line := range header {
 		matches := regex.FindAllStringSubmatch(line, -1)
+		if len(matches) == 0 {
```
hferentschik: So this is when there is no match, right? In this case you append to 'metaMap[key]', right? What if metaMap[key] is not set yet? What if someone does:

    # Hello 
    # Name:foo

Would you not get a panic there. Personally, I probably would aggregate multiple lines in a separate variable and only add things to the metaMap once I progress to the next tag or each the end of the header. Might also be worth to extract this into a function with a name describing what this is about, e.g. _handleMultileLine_. Took me a while to understand the code,

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -182,10 +182,17 @@ func (parser *AddOnParser) parseCommands(scanner *bufio.Scanner) ([]command.Comm
...
+			line = strings.TrimPrefix(line, commentChar)
+			metaMap[key] = metaMap[key] + " " + strings.TrimSpace(line)
+			continue
+		}
 		for _, match := range matches {
-			metaMap[strings.Trim(match[1], " ")] = strings.Trim(match[2], " ")
+			key = strings.Trim(match[1], " ")
```
hferentschik: As mentioned in the other command, we might want to harden the regexp a bit, maybe:

    regexp.Compile(`^# ?([a-zA-Z-]*):(.*)`)

This way something like:

    # Description: This add-on
    # prints the host:port of the OpenShift cluster

docs/source/using/addons.adoc
```diff
@@ -273,5 +273,19 @@ After you define the add-on, you can install it by running:
...
 ----
 
+TIP: You can also write metadata in multiple line.
+[[example-multiline-metadata]]
+.Example: Add-on definition which contain multiline description
+
+----
+# Name: prometheus
+# Description: This template creates a Prometheus instance preconfigured to gather OpenShift and
+#  Kubernetes platform and node metrics and report them to admins. It is protected by an
```
gbraad:
```
# Description: This template creates a Prometheus instance preconfigured to gather OpenShift and
#  Kubernetes platform and node metrics and report them to admins. It is protected by an
```

is the extra space following the hash `#` important for the multi-line description?

praveenkumar: @gbraad no it's not you can have any number of space after `#`

gbraad: So it can also be as follows?

```
# Description: This template creates a Prometheus instance preconfigured to gather OpenShift and
# Kubernetes platform and node metrics and report them to admins. It is protected by an
```

praveenkumar: @gbraad yes

hferentschik: +1 to that. Might have been worth to write the test this way to document the default expected case. The test might imply that you need extra spaces and hence @gbraad comments.

docs/source/using/addons.adoc
```diff
@@ -273,5 +273,19 @@ After you define the add-on, you can install it by running:
 $ minishift addons install <ADDON_DIR_PATH>
 ----
 
+TIP: You can also write metadata in multiple line.
+[[example-multiline-metadata]]
+.Example: Add-on definition which contain multiline description
+
```
hferentschik: the indentation is not right, when rendered. The actual example is not part of the TIP anymore, but rendered in the surrounding context. One needs a '+' to connect the content within the TIP in this case. 

There is really no alternative to actually rendering the Asciidoc to see whether the syntax is correct and as expected. 

Simple preview of some editors make it sometimes look like correct, but it is not.

I recommend to install the Ruby tool chain. 'rake serve' takes no time at all to spin up. The Docker setup is mainly there to provide a Happy Path. The better experience is to have the Ruby toolchain installed to run the Rake tasks directly. @minishift/minishift-dev please try. The editing experience when using 'rake serve' in conjunction with a live reload enabled browser is unbeatable. You edit, you save, you look at your bowser and you see the rendered output fully integrated into the site.

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -180,12 +180,19 @@ func (parser *AddOnParser) parseCommands(scanner *bufio.Scanner) ([]command.Comm
 }
 
 func createMetaData(header []string) (addon.AddOnMeta, error) {
-	regex, _ := regexp.Compile(`^# ?(.*):(.*)`)
+	regex, _ := regexp.Compile(`^# ?([a-zA-Z-]*):(.*)`)
```
hferentschik: :+1

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -180,12 +180,19 @@ func (parser *AddOnParser) parseCommands(scanner *bufio.Scanner) ([]command.Comm
...
 	metaMap := make(map[string]string)
+	var key string
 	for _, line := range header {
 		matches := regex.FindAllStringSubmatch(line, -1)
+		if len(matches) == 0 && key != "" {
+			line = strings.TrimPrefix(line, commentChar)
+			metaMap[key] = metaMap[key] + " " + strings.TrimSpace(line)
```
hferentschik: This really removes the new line. I think we should keep it. I added for example two lines to the anyuid add-on:

        # Name: anyuid
        # Description: Changes the default security context constraints to allow pods to run as any
        # foo
        # bar

When I then run `minishift addon list --verbose` I get:

    Name       : anyuid
    Description: Changes the default security context constraints to allow pods to run as any user. foo bar
    Enabled    : enabled
    Priority   : 0

I would expect the newlines to be preserved. I always had in mind to add:

    $ minishift add-on describe <name> 

which just dumps the description to the console. An add-on would become self describing of sorts and I would expect that one would less accompanying READMEs for add-ons. One would put the essential parts into the add-on description. 

@jorgemoralespou, I would expect that you also expected the description to preserve the new lines, right?

hferentschik: Maybe we should even store the description as a slice of strings. That might be more flexible for various use cases. For example in the case of `minishift addon list --verbose` you would just display the first line, whereas an explicit `minishift add-on describe <name>` would write the full description content. Having a slice makes things a bit more flexible. Otherwise you might have to split yet again at newline.

hferentschik: Required-Vars for example is also a slice.

pkg/minishift/addon/parser/addon_parser.go
gbraad: I would prefer regular expressions are defined as a constant and get a name, as this makes their intention of use much clearer.

docs/source/using/addons.adoc
gbraad: As also mentioned below, it felt like the whitespace was intentional

pkg/minishift/addon/parser/addon_parser.go
praveenkumar: > Maybe we should even store the description as a slice of strings. That might be more flexible for various use cases. For example in the case of minishift addon list --verbose you would just display the first line, whereas an explicit minishift add-on describe <name> would write the full description content. Having a slice makes things a bit more flexible. Otherwise you might have to split yet again at newline.

We can store in the slice of strings but I didn't see any use until we break `createMetaData` function and have another function which return slice to different metadata key. Right now we are not enabling multi-line feature only for description but all sort of metadata tags.

docs/source/using/addons.adoc
```diff
@@ -273,5 +273,21 @@ After you define the add-on, you can install it by running:
 $ minishift addons install <ADDON_DIR_PATH>
 ----
 
+[TIP]
+====
+You can also write metadata in multiple line.
+[[example-multiline-metadata]]
+.Example: Add-on definition which contain multiline description
+----
```
hferentschik: +1

pkg/minishift/addon/addon_meta.go
```diff
@@ -31,19 +31,28 @@ const (
...
+		case string:
+			if v == "" {
+				return nil, errors.New(fmt.Sprintf("Metadata does not contain an mandatory entry for '%s'", tag))
+			}
+		case []string:
+			if len(v) == 0 {
+				return nil, errors.New(fmt.Sprintf("Metadata does not contain an mandatory entry for '%s'", tag))
+			}
+		default:
+			return nil, nil
```
praveenkumar: @hferentschik Does that what you wanted as part of having multiline string for a specific key?

cmd/minishift/cmd/addon/addons_list.go
```diff
@@ -88,7 +89,8 @@ func printAddOnList(manager *manager.AddOnManager, writer io.Writer, template *t
 	addOns := manager.List()
 	sort.Sort(addon.ByStatusThenPriorityThenName(addOns))
 	for _, addon := range addOns {
-		addonTemplate := DisplayAddOn{addon.MetaData().Name(), addon.MetaData().Description(), stringFromStatus(addon.IsEnabled()), addon.GetPriority()}
+		description := strings.Join(addon.MetaData().Description(), fmt.Sprintf("\n%13s", " "))
```
hferentschik: +1

pkg/minishift/addon/addon_meta.go
```diff
@@ -31,19 +31,28 @@ const (
 // AddOnMeta defines a set of meta data for an AddOn. Name and Description are required. Others are optional.
 type AddOnMeta interface {
 	Name() string
-	Description() string
+	Description() []string
```

pkg/minishift/addon/addon_meta.go
```diff
@@ -31,19 +31,28 @@ const (
...
 	RequiredVars() []string
 	GetValue(key string) string
 }
 
 type DefaultAddOnMeta struct {
-	headers map[string]string
+	headers map[string]interface{}
```
hferentschik: one way of doing it ;-)

pkg/minishift/addon/addon_meta.go
```diff
@@ -31,19 +31,28 @@ const (
...
 }
 
-func NewAddOnMeta(headers map[string]string) (AddOnMeta, error) {
+func NewAddOnMeta(headers map[string]interface{}) (AddOnMeta, error) {
 	for _, tag := range requiredMetaTags {
-		if headers[tag] == "" {
-			return nil, errors.New(fmt.Sprintf("Metadata does not contain an mandatory entry for '%s'", tag))
+		switch v := headers[tag].(type) {
+		case string:
```
hferentschik: nice with the switch

pkg/minishift/addon/addon_meta.go
hferentschik: that will do for now

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -180,12 +180,25 @@ func (parser *AddOnParser) parseCommands(scanner *bufio.Scanner) ([]command.Comm
...
 func createMetaData(header []string) (addon.AddOnMeta, error) {
-	regex, _ := regexp.Compile(`^# ?(.*):(.*)`)
-	metaMap := make(map[string]string)
+	regex, _ := regexp.Compile(`^# ?([a-zA-Z-]*):(.*)`)
+	metaMap := make(map[string]interface{})
+	var key string
+	var value []string
 	for _, line := range header {
 		matches := regex.FindAllStringSubmatch(line, -1)
+		if len(matches) == 0 && key == "Description" {
```
hferentschik: that's a bit fragile. You are using twice the hard coded string "Description". At the very least it should be a constant in this file, or why not adding two constants in _addon_meta.go_ (DescriptionMetaTagName and NameMetaTagName). You can then reference this constant here and you can add these constants to _requiredMetaTags_ in _addon_meta.go_.

The more generic solution is to introduce as mentioned a MetaTagType, but maybe let's not go there.

gbraad: As mentioned earlier, I 'd rather also see the regexps as a contant, as this makes them named and more understandable in their purpose. If this is extracted as a constant, consider the regex patterns too, please.

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -180,12 +180,25 @@ func (parser *AddOnParser) parseCommands(scanner *bufio.Scanner) ([]command.Comm
 }
 
 func createMetaData(header []string) (addon.AddOnMeta, error) {
-	regex, _ := regexp.Compile(`^# ?(.*):(.*)`)
-	metaMap := make(map[string]string)
+	regex, _ := regexp.Compile(`^# ?([a-zA-Z-]*):(.*)`)
```
gbraad: extract as constant

hferentschik:
Just tested `minishift addons list --verbose` and the output looks good now.

hferentschik:
Rebased and merged. Thanks.

PR #1284, opened 2017-08-18
Issue #1227 Update node config path

praveenkumar:


pkg/minishift/openshift/openshift.go
```diff
@@ -44,8 +43,8 @@ var (
 
 	NODE = OpenShiftPatchTarget{
 		"node",
-		"/var/lib/origin/openshift.local.config/node-%s/node-config.yaml",
-		"/var/lib/minishift/openshift.local.config/node-%s/node-config.yaml",
+		"/var/lib/origin/openshift.local.config/node-localhost/node-config.yaml",
```
hferentschik: ahh, and that will again work backwards compatible, right? Since we are using 3.6 now to provision say 1.5.

pkg/minishift/openshift/openshift.go
```diff
@@ -63,21 +62,15 @@ func (t *OpenShiftPatchTarget) SetIp(ip string) {
 
 func (t *OpenShiftPatchTarget) containerConfigFilePath() (string, error) {
 	if t.target == "node" {
-		if t.ip == "" {
```
hferentschik: ahh, that we can rid of indeed then

pkg/minishift/openshift/openshift.go
praveenkumar: Right :)

hferentschik: Got you.

hferentschik:
Merged, thanks. I actually also removed the _ip_ member from the _OpenShiftPatchTarget_ struct, since it was not used anymore.

hferentschik:
Cucumber integration test will be added as part of pull request #1228

Issue #1289, opened 2017-08-21
Update "executing" steps in integration tests to accept commands with quotes


agajdosi:
Currently nearly all steps which execute a command have problem with commands containing quotes.

```golang
s.Step(`([^"]*) of command "minishift ([^"]*)" (is equal|is not equal) to "([^"]*)"$`,
commandReturnEquals)
```
Most commands for minishift do not use quotes, but there are some exceptions, for example `minishift openshift config set` where following argument must be a JSON, which uses quotes. And this will not be matched properly by adequate steps in test suite.

This can be probably fixed by using matching group and its surroundings similar to this: ```"minishift (.*)"```, which is already added in one place into tests by @hferentschik.

hferentschik:
Merged via pull request #1290

PR #1290, opened 2017-08-21
Issue #1289 Update regex matching in test suite

agajdosi:
This PR on several places changes matching groups from `([^"]*)"` to `(.*)"` as there is potential need to be able to match inputs which contain quotes.

### DIFF
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -147,29 +147,29 @@ func FeatureContext(s *godog.Suite) {
 	minishift = &Minishift{runner: runner}
 
 	// steps to execute `minishift` commands
-	s.Step(`Minishift (?:has|should have) state "([^"]*)"$`,
+	s.Step(`^Minishift (?:has|should have) state "(Does Not Exist|Running|Stopped)"$`,
 		minishift.shouldHaveState)
 	s.Step(`^executing "minishift (.*)"$`,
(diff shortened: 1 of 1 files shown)


gbraad:
Appveyor and network timeouts, again

test/integration/integration_test.go
```diff
@@ -147,29 +147,29 @@ func FeatureContext(s *godog.Suite) {
 	minishift = &Minishift{runner: runner}
 
 	// steps to execute `minishift` commands
-	s.Step(`Minishift (?:has|should have) state "([^"]*)"$`,
+	s.Step(`^Minishift (?:has|should have) state "(Does Not Exist|Running|Stopped)"$`,
```
hferentschik: +1 much better to be explicit

budhrg: :+1: explicit is better :smile:

gbraad: but wrong, as it covers only the wanted states:

https://github.com/docker/machine/blob/master/libmachine/state/state.go

`Paused`, `Error` and `Timeout` are states I have also encountered during testing.

hferentschik:
Merged, thanks.

test/integration/integration_test.go
agajdosi: @gbraad Thanks for pointing this out, saving the link for future.

Issue #1249, opened 2017-08-09
Can't mount /Users on macos x using xhyve drive


cmoulliard:
Minishift version : 1.4.1
Driver : xhyve
MacOS Sierra : 10.12.5

We can't mount /Users on Macos X - using this command

```
minishift hostfolder add public
UNC path: //192.168.1.80/public
Mountpoint [/mnt/sda1/public]:
Username: dabou
Password: [HIDDEN]
Domain:
Added: public

minishift hostfolder mount public
   Mounting 'public': '//192.168.1.80/public' as '/mnt/sda1/public' ... FAIL
```

even if the folders is shared

![screenshot 2017-08-09 09 39 26](https://user-images.githubusercontent.com/463790/29110340-a76969a2-7ce6-11e7-9543-5ea54360e51e.png)

gbraad:
is this using the B2D or CentOS image?

cmoulliard:
I'm using B2D

cmoulliard:
Here is the stack trace

```
minishift hostfolder mount users --show-libmachine-logs -v 5
Found binary path at /usr/local/bin/docker-machine-driver-xhyve
Launching plugin server for driver xhyve
Plugin server listening at address 127.0.0.1:60970
() DBG | operation not supported by device
() Calling .GetVersion
Using API Version  1
() Calling .SetConfigRaw
() Calling .GetMachineName
(minishift) Calling .GetState
(minishift) Calling .GetState
(minishift) Calling .GetSSHHostname
(minishift) Calling .GetSSHPort
(minishift) Calling .GetSSHKeyPath
(minishift) Calling .GetSSHKeyPath
(minishift) Calling .GetSSHUsername
Using SSH client type: external
Using SSH private key: /Users/dabou/.minishift/machines/minishift/id_rsa (-rw-------)
&{[-F /dev/null -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=none docker@192.168.64.2 -o IdentitiesOnly=yes -i /Users/dabou/.minishift/machines/minishift/id_rsa -p 22] /usr/bin/ssh <nil>}
About to run SSH command:
sudo mkdir -p /mnt/sda1/users
SSH cmd err, output: <nil>:
[...]

PR #1297, opened 2017-08-23
[WIP] Issue #1249 Force NTLM authentication

gbraad:
Fix for issue #1249 ?

### DIFF
--- a/pkg/minishift/hostfolder/hostfolder.go
+++ b/pkg/minishift/hostfolder/hostfolder.go
@@ -321,7 +321,7 @@ func mountCifsHostfolder(driver drivers.Driver, hostfolder *config.HostFolder) e
 	}
 
 	cmd := fmt.Sprintf(
-		"sudo mount -t cifs %s %s -o username=%s,password=%s",
+		"sudo mount -t cifs %s %s -o username=%s,password=%s,sec=ntlm",
 		hostfolder.Options["uncpath"],
 		hostfolder.Mountpoint(),
 		hostfolder.Options["username"],


hferentschik:
retest this please

hferentschik:
@gbraad, so is this something needed you think or just an experiment? Should this be added regardless of whether the reported issue is "user error"?

gbraad:
I am not sure what the issue with the user is, but this doesn't hurt to test...

hferentschik:
What exactly does it do?

gbraad:
force authentication to happen with NTLM v1


```
$ man mount.cifs
...
sec=
           Security mode. Allowed values are:
           ·   none - attempt to connection as a null user (no name)
           ·   krb5 - Use Kerberos version 5 authentication
           ·   krb5i - Use Kerberos authentication and forcibly enable packet signing
           ·   ntlm - Use NTLM password hashing
           ·   ntlmi - Use NTLM password hashing and force packet signing
           ·   ntlmv2 - Use NTLMv2 password hashing
           ·   ntlmv2i - Use NTLMv2 password hashing and force packet signing
           ·   ntlmssp - Use NTLMv2 password hashing encapsulated in Raw NTLMSSP message
           ·   ntlmsspi - Use NTLMv2 password hashing encapsulated in Raw NTLMSSP message, and force packet signing

           The default in mainline kernel versions prior to v3.8 was sec=ntlm. In v3.8, the default was changed to sec=ntlmssp.

           If the server requires signing during protocol negotiation, then it may be enabled automatically. Packet signing may also be enabled automatically if it's enabled in /proc/fs/cifs/SecurityFlags.
```

the default is now `ntlmssp` and it might fail if trying to connect to an older default.

hferentschik:
So you want to add  'sec=ntlm' just to see whether it makes a difference for issue #1249 specifically? I guess what I am concerned about is to merge this into the production code and breaking it for people for whom things actually work. In case it would make a difference for #1249, would be the plan to make this a config option in some form?

gbraad:
Other issues related to WannaCry have been reported, so we might want to be able to force a `sec` method. Configurable would be a preferred way. I have the executable and tested this... so closing.

hferentschik:
ok, thanks.

Issue #1225, opened 2017-08-04
Doc review for experimental features


thatdocslady:
Following PR https://github.com/minishift/minishift/pull/1185 that was merged before a doc review was completed, the new content and relevant doc strings need to be reviewed.

gbraad:
Please go ahead. You can use the same issue ID as #1185 referenced

hferentschik:
Merged via pull request #1298

PR #1298, opened 2017-08-23
Issue #1225 Doc review for experimental features

thatdocslady:
Fixes issue #1225 

This PR includes edits to the experimental features content. Note that the doc strings were reviewed and approved (no need for edits there).

@gbraad do you mind giving this an initial glance before I remove the WIP flag?

### DIFF
--- a/docs/source/using/managing-minishift.adoc
+++ b/docs/source/using/managing-minishift.adoc
@@ -293,12 +293,10 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
 71fe8ff16548 openshift/origin:v1.5.1 "/usr/bin/openshift s" 4 minutes ago  Up 4 minutes  origin
 ----
 
-[[enabling-experimental-openshift-features]]
-== Enabling Experimental OpenShift Features
+[[experimental-features]]
+== Experimental Features
 
-By default, Minishift does not expose all `oc cluster up` flags in the CLI, either because certain features are not considered stable or because Minishift does not yet support the latest options in `oc cluster up`.
-
-However, if you want to get early access to some upcoming features and experiment, you can enable some of those in Minishift.
+If you want to get early access to some upcoming features and experiment, you can enable some of those in Minishift.
 
 [IMPORTANT]
 ====
@@ -312,10 +310,12 @@ You do this by setting the environment variable `MINISHIFT_ENABLE_EXPERIMENTAL`,
 $ export MINISHIFT_ENABLE_EXPERIMENTAL=y
 ----
 
-Currently this environment variable controls additional options only for the xref:../command-ref/minishift_start.adoc#[`minishift start`] command.
-The additional options are related to  xref:../openshift/openshift-client-binary.adoc#[`oc cluster up`].
+[[enabling-experimental-oc-flags]]
+=== Enabling experimental `oc cluster up` flags in `minishift start`
 
-When you set the environment variable, you get access to the following flags:
+By default, Minishift does not expose all xref:../openshift/openshift-client-binary.adoc#[`oc cluster up`] flags in the CLI, either because certain features are not considered stable or because Minishift does not yet support the latest options in `oc cluster up`.
+
+You can set the environment variable to enable the following options for the xref:../command-ref/minishift_start.adoc#[`minishift start`] command:
 
 `service-catalog`::
 Enables provisioning the OpenShift link:https://docs.openshift.org/latest/architecture/service_catalog/index.html[service catalog].
--- a/docs/source/using/managing-minishift.adoc
+++ b/docs/source/using/managing-minishift.adoc
@@ -298,27 +298,27 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
 
 If you want to get early access to some upcoming features and experiment, you can enable some of those in Minishift.
 
-[IMPORTANT]
-====
-Experimental features are not officially supported, and might break or result in unexpected behavior.
-To share your feedback on these features, you are welcome to link:https://github.com/minishift/minishift#community[contact the Minishift community].
-====
-
 You do this by setting the environment variable `MINISHIFT_ENABLE_EXPERIMENTAL`, which makes additional flags available:
 
 ----
 $ export MINISHIFT_ENABLE_EXPERIMENTAL=y
 ----
 
+[IMPORTANT]
+====
+Experimental features are not officially supported, and might break or result in unexpected behavior.
+To share your feedback on these features, you are welcome to link:https://github.com/minishift/minishift#community[contact the Minishift community].
+====
+
 [[enabling-experimental-oc-flags]]
 === Enabling experimental `oc cluster up` flags in `minishift start`
 
-By default, Minishift does not expose all xref:../openshift/openshift-client-binary.adoc#[`oc cluster up`] flags in the CLI, either because certain features are not considered stable or because Minishift does not yet support the latest options in `oc cluster up`.
+By default, Minishift does not expose all link:https://github.com/openshift/origin/blob/master/docs/cluster_up_down.md[`oc cluster up`] flags in the Minishift CLI.
 
-You can set the environment variable to enable the following options for the xref:../command-ref/minishift_start.adoc#[`minishift start`] command:
+You can set the `MINISHIFT_ENABLE_EXPERIMENTAL` environment variable to enable the following options for the xref:../command-ref/minishift_start.adoc#[`minishift start`] command:
 
 `service-catalog`::
 Enables provisioning the OpenShift link:https://docs.openshift.org/latest/architecture/service_catalog/index.html[service catalog].
 
 `extra-clusterup-flags`::
-Enables passing any extra flags to `oc cluster up`.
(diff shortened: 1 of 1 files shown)


docs/source/using/managing-minishift.adoc
```diff
@@ -293,23 +293,32 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
 71fe8ff16548 openshift/origin:v1.5.1 "/usr/bin/openshift s" 4 minutes ago  Up 4 minutes  origin
 ----
 
-[[experimental-functionality]]
-== Experimental functionality
+[[enabling-experimental-openshift-features]]
+== Enabling Experimental OpenShift Features
```
hferentschik: It might not be OpenShift. Right now we use it to expose some OpenShift/cluster up functionality, but I would imagine that we could use it for Minishift experimental features as well.

docs/source/using/managing-minishift.adoc
```diff
@@ -293,23 +293,32 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
...
 
-To allow users and developers to get early access to some upcoming features, we offer a way to enable experimental functionality.
+By default, Minishift does not expose all `oc cluster up` flags in the CLI, either because certain features are not considered stable or because Minishift does not yet support the latest options in `oc cluster up`.
```
hferentschik: I recall that we had this before, but decided it was to specific. This description is geared towards _oc_ and _cluster up_ flags. This is what we use it for now, but experimental features could and probably will be used for other things as well. For example @gbraad is looking into adding options for provisioning Hyper-V against a fixed IP. This might go behind the experimental flag as well, but has nothing to do with _cluster up_.

docs/source/using/managing-minishift.adoc
```diff
@@ -293,23 +293,32 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
...
 
-You also get access to the `extra-clusterup-flags` flag, allowing you to pass any extra flags to `oc cluster up`.
-Minishift does per default not expose all `oc cluster up` flags in its CLI, either because certain features are not considered stable or simply because Minishift has not caught up yet with the latest options in `oc cluster up`.
-In these cases `extra-clusterup-flags` can be used to run experiments with these options.
+`service-catalog`::
+Enables provisioning the OpenShift link:https://docs.openshift.org/latest/architecture/service_catalog/index.html[service catalog].
 
-Be aware that these features can break or give a result you weren't expecting.
-We welcome your feedback when using them.
+`extra-clusterup-flags`::
```
hferentschik: This sounds now imo a bit like _extra-clusterup-flags_ being a flag of 'cluster up' which it is not.

docs/source/using/managing-minishift.adoc
```diff
@@ -293,23 +293,32 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
...
 
-To allow users and developers to get early access to some upcoming features, we offer a way to enable experimental functionality.
+By default, Minishift does not expose all `oc cluster up` flags in the CLI, either because certain features are not considered stable or because Minishift does not yet support the latest options in `oc cluster up`.
 
-You do this by setting the environment variable `MINISHIFT_ENABLE_EXPERIMENTAL` which will make additional flags available:
+However, if you want to get early access to some upcoming features and experiment, you can enable some of those in Minishift.
+
+[IMPORTANT]
+====
+Experimental features are not officially supported, and might break or result in unexpected behavior.
```
hferentschik: I thought we try to avoid the term "supported" as much as possible.

thatdocslady:
@hferentschik thanks for the clarifications. Based on your feedback, I did the following:

- Separated the general description of the env var and its purpose and put it at the top of the section.
- Created a sub-heading for the oc cluster up implementation and cleaned up the text a bit to make it more use-case oriented.

By doing this, we now have a framework inside the section to add more sub-headings for each type of experimental feature that you might add in the future (so that the content can scale without losing structure or flow).

Hope this is better now.

docs/source/using/managing-minishift.adoc
```diff
@@ -293,23 +293,32 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
...
-[[experimental-functionality]]
-== Experimental functionality
+[[experimental-features]]
+== Experimental Features
 
-To allow users and developers to get early access to some upcoming features, we offer a way to enable experimental functionality.
+If you want to get early access to some upcoming features and experiment, you can enable some of those in Minishift.
 
-You do this by setting the environment variable `MINISHIFT_ENABLE_EXPERIMENTAL` which will make additional flags available:
+[IMPORTANT]
```
hferentschik: Could we move this under the code listing 'export ...' It really breaks my flow in reading and understanding this, when it sits between these two sentences.

docs/source/using/managing-minishift.adoc
```diff
@@ -293,23 +293,32 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
...
 ----
 $ export MINISHIFT_ENABLE_EXPERIMENTAL=y
 ----
 
-At the moment this is only relevant for the xref:../command-ref/minishift_start.adoc#[`minishift start`] command where you get additional options related to  xref:../openshift/openshift-client-binary.adoc#[`oc cluster up`].
-By setting the environment variable, you get access to the `service-catalog` flag which allows you to provision OpenShift's link:https://docs.openshift.org/latest/architecture/service_catalog/index.html[service catalog].
+[[enabling-experimental-oc-flags]]
+=== Enabling experimental `oc cluster up` flags in `minishift start`
+
+By default, Minishift does not expose all xref:../openshift/openshift-client-binary.adoc#[`oc cluster up`] flags in the CLI, either because certain features are not considered stable or because Minishift does not yet support the latest options in `oc cluster up`.
```
hferentschik: I don't think this is such a good link target. The target does not say much about _cluster up_ and in fact in itself refers to this link https://github.com/openshift/origin/blob/master/docs/cluster_up_down.md

hferentschik: "flags in the CLI" -> "flags its CLI"

hferentschik: " either because certain features are not considered stable or because Minishift does not yet support the latest options in `oc cluster up`." -> This sounds odd, especially the second half. I don't think we need to explain anything here. Just "By default, Minishift does not expose all xref:../openshift/openshift-client-binary.adoc#[`oc cluster up`] flags in its CLI."

docs/source/using/managing-minishift.adoc
```diff
@@ -293,23 +293,32 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
...
+
+You can set the environment variable to enable the following options for the xref:../command-ref/minishift_start.adoc#[`minishift start`] command:
```
hferentschik: What's "the environment variable"? I think you need to say explicitly "You can set the MINISHIFT_ENABLE_EXPERIMENTAL environment variable "

docs/source/using/managing-minishift.adoc
```diff
@@ -293,23 +293,32 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
...
-You also get access to the `extra-clusterup-flags` flag, allowing you to pass any extra flags to `oc cluster up`.
-Minishift does per default not expose all `oc cluster up` flags in its CLI, either because certain features are not considered stable or simply because Minishift has not caught up yet with the latest options in `oc cluster up`.
-In these cases `extra-clusterup-flags` can be used to run experiments with these options.
+`service-catalog`::
+Enables provisioning the OpenShift link:https://docs.openshift.org/latest/architecture/service_catalog/index.html[service catalog].
 
-Be aware that these features can break or give a result you weren't expecting.
-We welcome your feedback when using them.
+`extra-clusterup-flags`::
+Enables passing any extra flags to `oc cluster up`.
```
hferentschik: Enabling the ability to pass along any `oc cluster up` flag which is otherwise not directly exposed by the Minishift flags.

docs/source/using/managing-minishift.adoc
```diff
@@ -293,23 +293,32 @@ CONTAINER    IMAGE                   COMMAND                CREATED        STATU
...
+You do this by setting the environment variable `MINISHIFT_ENABLE_EXPERIMENTAL`, which makes additional flags available:
 
 ----
 $ export MINISHIFT_ENABLE_EXPERIMENTAL=y
 ----
 
-At the moment this is only relevant for the xref:../command-ref/minishift_start.adoc#[`minishift start`] command where you get additional options related to  xref:../openshift/openshift-client-binary.adoc#[`oc cluster up`].
-By setting the environment variable, you get access to the `service-catalog` flag which allows you to provision OpenShift's link:https://docs.openshift.org/latest/architecture/service_catalog/index.html[service catalog].
+[[enabling-experimental-oc-flags]]
+=== Enabling experimental `oc cluster up` flags in `minishift start`
```
hferentschik: +1 This makes sense. I like this new structure.

hferentschik:
Thanks @thatdocslady. I squashed your commits and merged.

Issue #1129, opened 2017-07-13
Create uninstall add-on command


jorgemoralespou:
I want to be able to use addon mechanism to remove an addon.

As I'm currently developing some addons, I find myself having to do manual removal of things through oc commands, and eventually outputing the command needed as part of the addon installation. That's useful but I will not have a way to access that information some time later.

e.g.
````
-- Applying addon 'cockpit':....
  You have installed cockpit
  To access cockpit go to http://cockpit.192.168.64.2.nip.io
  or
  Use minishift:
  minishift openshift service cockpit -n cockpit

  To delete:
  oc delete oauthclients/cockpit-oauth-client project/cockpit --as=system:admin
````

I want a way to uninstall an addon. Having in the addon folder two files:

````
myaddon.addon
myaddon.addon.uninstall
````

So that when I do "minishift addon remove" will execute that other script.
If I do "minishift addon uninstall" will also uninstall the addon, hence will be marked in the configuration as not installed.

hferentschik:
+1 for the idea. Let's try to think that trough. I guess some uninstall file is a good idea and quite intuitive.

praveenkumar:
Here is my thought on this issue.

- Right now we have `addon apply` which even doesn't update configuration so when we run uninstall script we only change it to disable if it enabled or we don't do anything so in next run it again come up?

- Does uninstall also need to have mandatory metadata field like (name, description) etc,?

hferentschik:
> Right now we have addon apply which even doesn't update configuration so when we run uninstall script we only change it to disable if it enabled or we don't do anything so in next run it again come up?

Not quite sure what you mean. I have problems to "parse" this sentence. Can you please clarify?

> Does uninstall also need to have mandatory metadata field like (name, description) etc,?

I like what @jorgemoralespou suggested. So there would be a new command 'minishfift add-on uninstall foo'. This command would look at the directory of the foo addon for foo.addon.uninstall and if found execute it. If the file does not exist we report "No uninstall action defined for add-on foo". If it exists we execute the commands in there. The DSL would really be the same, so we would just re-use the same parser. We might want to relax the required meta data. For example 'Description' might not be needed. Whether we enforce a name is discussable. From a technical point of view it would not be necessary. What for example if they don't match? The name in the add-on 'foo' and in the un-install 'bar'. My gut feeling is that for the uninstall we don't require any meta-data and just ignore all lines with '#'.

[...]

PR #1305, opened 2017-08-24
Issue #1129 Implement remove command for add-ons

praveenkumar:
Fix for #1129

Todo
====

- Unit tests
- Documentations

### DIFF
--- a/addons/admin-user/admin-user.addon.remove
+++ b/addons/admin-user/admin-user.addon.remove
@@ -1,0 +1,5 @@
+# Name: admin-user
+# Description: Remove admin user addon
+
+oc delete user admin
+echo admin user deleted
--- a/pkg/minishift/addon/manager/addon_manager.go
+++ b/pkg/minishift/addon/manager/addon_manager.go
@@ -23,13 +23,13 @@ import (
 	"path/filepath"
 	"sort"
 
-	"github.com/golang/glog"
+	"strings"
+
 	"github.com/minishift/minishift/pkg/minishift/addon"
 	"github.com/minishift/minishift/pkg/minishift/addon/command"
 	"github.com/minishift/minishift/pkg/minishift/addon/parser"
 	"github.com/minishift/minishift/pkg/util/filehelper"
 	"github.com/pkg/errors"
-	"strings"
 )
 
 // AddOnManager is the central point for all operations around managing addons. An addon
@@ -63,7 +63,7 @@ func NewAddOnManager(baseDir string, configMap map[string]*addon.AddOnConfig) (*
 		if err != nil {
 			_, ok := err.(parser.ParseError)
 			if ok {
-				glog.Warning(fmt.Sprintf("Skipping addon '%s' in '%s' due to parse error: %s", f.Name(), fullPath, err.Error()))
+				fmt.Println(fmt.Sprintf("Skipping addon '%s' in '%s' due to parse error: %s", f.Name(), fullPath, err.Error()))
 				continue
 			} else {
 				return nil, errors.Wrapf(err, "Unable to create addon manager for %s. ", baseDir)
@@ -187,6 +187,33 @@ func (m *AddOnManager) ApplyAddOn(addOn addon.AddOn, context *command.ExecutionC
 
 	os.Chdir(addOn.InstallPath())
 	for _, c := range addOn.Commands() {
+		err := c.Execute(context)
+		if err != nil {
+			return err
+		}
+	}
+	fmt.Print("\n\n")
+	return nil
+}
+
+func (m *AddOnManager) RemoveAddOn(addOn addon.AddOn, context *command.ExecutionContext) error {
+	fmt.Print(fmt.Sprintf("-- Removing addon '%s':", addOn.MetaData().Name()))
+	context.AddToContext("addon-name", addOn.MetaData().Name())
+	defer context.RemoveFromContext("addon-name")
+
+	err := m.verifyRequiredVariablesInContext(context, addOn.MetaData())
+	if err != nil {
+		return err
+	}
+
+	oldDir, err := os.Getwd()
+	if err != nil {
+		return errors.Wrap(err, "Unable to remove addon due to failing IO operation")
+	}
+	defer os.Chdir(oldDir)
+
+	os.Chdir(addOn.InstallPath())
+	for _, c := range addOn.RemoveCommands() {
 		err := c.Execute(context)
 		if err != nil {
 			return err
--- a/pkg/minishift/addon/parser/addon_parser.go
+++ b/pkg/minishift/addon/parser/addon_parser.go
@@ -34,9 +34,10 @@ import (
 const (
 	commentChar = "#"
 
-	noAddOnDefinitionFoundError   = "There needs to be an addon file per addon directory. Found none in %s"
-	multipleAddOnDefinitionsError = "There can only be one addon file per addon directory. Found %s"
-	regexToGetMetaTagInfo         = `^# ?([a-zA-Z-]*):(.*)`
+	noAddOnDefinitionFoundError         = "There needs to be an addon file per addon directory. Found none in %s"
+	multipleAddOnDefinitionsError       = "There can only be one addon file per addon directory. Found %s"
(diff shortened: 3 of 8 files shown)


pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -59,21 +61,28 @@ func NewAddOnManager(baseDir string, configMap map[string]*addon.AddOnConfig) (*
 			continue
 		}
 
-		addOn, err := p.Parse(fullPath)
+		addOns, err := p.Parse(fullPath)
 		if err != nil {
 			_, ok := err.(parser.ParseError)
 			if ok {
-				glog.Warning(fmt.Sprintf("Skipping addon '%s' in '%s' due to parse error: %s", f.Name(), fullPath, err.Error()))
+				fmt.Println(fmt.Sprintf("Skipping addon '%s' in '%s' due to parse error: %s", f.Name(), fullPath, err.Error()))
```
praveenkumar: This is right now only for test purpose, I will remove it before final PR update.

hferentschik: ok

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -59,21 +61,28 @@ func NewAddOnManager(baseDir string, configMap map[string]*addon.AddOnConfig) (*
...
+				removeAddon := addOns[1]
+				detectedRemoveAddons[removeAddon.MetaData().Name()] = removeAddon
+			}
+		}
 		setStateAndPriority(addOn, configMap)
 		detectedAddOns[addOn.MetaData().Name()] = addOn
 	}
 
-	return &AddOnManager{baseDir: baseDir, addOns: detectedAddOns}, nil
+	return &AddOnManager{baseDir: baseDir, addOns: detectedAddOns, removeAddon: detectedRemoveAddons}, nil
```
hferentschik: IMO this is unfortunate. You basically don't tie the remove commands to the actual addon. Instead you say here is a set of remove and apply add-ons with no tie between them. From a model perspective I would expect this to be part of the AddOn interface. There is for example the Commands function - https://github.com/minishift/minishift/blob/4d6d0d30e8f37982e32c1b2b849ad5d17954e917/pkg/minishift/addon/addon.go#L26-L26. There could for example be a RemoveCommands function. I think from a design pov this is better. In this case one would handle the two separate set of commands in AddOnParser.Parse. This method gets the base directory for the add-on. It is a question of checking whether there is a .addon.remove and parsing its commands.

BTW, I am using '.addon.remove' since I think this is what we should use sticking to the terminology we want to use. These are the commands we want to run when we call `minishift addon remove`. Uninstall is about removing the actual files from the add-on dir.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -165,6 +181,11 @@ func (m *AddOnManager) IsInstalled(name string) bool {
 	return installed
 }
 
+func (m *AddOnManager) CanRemoved(name string) bool {
```
hferentschik: CanRemoved? Not sure what this means and why you would need it

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -196,6 +217,33 @@ func (m *AddOnManager) ApplyAddOn(addOn addon.AddOn, context *command.ExecutionC
...
+	}
+
+	oldDir, err := os.Getwd()
+	if err != nil {
+		return errors.Wrap(err, "Unable to remove addon due to failing IO operation")
+	}
+	defer os.Chdir(oldDir)
+
+	os.Chdir(addOn.InstallPath())
+	for _, c := range addOn.Commands() {
```
hferentschik: This ties into the design discussion. You pass the add-on and then iterate the add-on commands. IMO this is confusing. It would be nicer here to pass an add-on, but then since it is the remove function, iterate the RemoveCommands.

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -72,26 +73,30 @@ func NewAddOnParser() *AddOnParser {
 
 // Parse takes as parameter a reader containing an addon definition and returns an AddOn instance.
 // If an error occurs nil and the error are returned.
-func (parser *AddOnParser) Parse(addOnDir string) (addon.AddOn, error) {
-	reader, err := parser.getAddOnContentReader(addOnDir)
+func (parser *AddOnParser) Parse(addOnDir string) ([]addon.AddOn, error) {
+	readers, err := parser.getAddOnContentReader(addOnDir)
```
hferentschik: this should not returns multiple add-ons, but a single one

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -102,10 +107,14 @@ func (parser *AddOnParser) getAddOnContentReader(addOnDir string) (io.Reader, er
 	}
 
 	var addOnFiles []string
+	var addOnRemoveFiles []string
```
hferentschik: ahh, so here is where you check the file name.

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -102,10 +107,14 @@ func (parser *AddOnParser) getAddOnContentReader(addOnDir string) (io.Reader, er
...
 	for _, fileInfo := range files {
 		if strings.HasSuffix(fileInfo.Name(), ".addon") {
```
hferentschik: 'addon.remove' is better I think

addons/admin-user/admin-user.addon.remove
```diff
@@ -0,0 +1,5 @@
+# Name: admin-user
```
hferentschik: +1, good addition.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -196,6 +196,33 @@ func (m *AddOnManager) ApplyAddOn(addOn addon.AddOn, context *command.ExecutionC
...
+	}
+
+	oldDir, err := os.Getwd()
+	if err != nil {
+		return errors.Wrap(err, "Unable to remove addon due to failing IO operation")
+	}
+	defer os.Chdir(oldDir)
+
+	os.Chdir(addOn.InstallPath())
+	for _, c := range addOn.RemoveCommands() {
```
hferentschik: +1 Much nicer now imo. I guess in the context of remove, there might be an even greater need to have commands which will not abort the overall execution in case an error occurs. But that's a different issue.

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -73,12 +74,15 @@ func NewAddOnParser() *AddOnParser {
 // Parse takes as parameter a reader containing an addon definition and returns an AddOn instance.
 // If an error occurs nil and the error are returned.
 func (parser *AddOnParser) Parse(addOnDir string) (addon.AddOn, error) {
-	reader, err := parser.getAddOnContentReader(addOnDir)
+	addonInstallReader, err := parser.getAddOnContentReader(addOnDir, ".addon")
```
hferentschik: +1

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -73,12 +74,15 @@ func NewAddOnParser() *AddOnParser {
...
 	if err != nil {
 		return nil, err
 	}
-
-	meta, commands, err := parser.parseAddOnContent(reader)
+	addonRemoveReader, err := parser.getAddOnContentReader(addOnDir, ".addon.remove")
```
hferentschik: 👍

praveenkumar:
> Are you planning to add the 'uninstall' command as well? Would be nice.

I will create a separate issue for adding `uninstall` flag for addons. Lets have this PR only for `remove`.

hferentschik:
> I will create a separate issue for adding uninstall flag for addons or create a new PR addressing same issue whatever make sense. Lets have this PR only for remove.

Ok, np.

hferentschik:
Merged, thanks. Added another commit to extend on the docs.

Issue #607, opened 2017-03-27
Cucumber tests should log to file


hferentschik:
To make it easier to debug Cucumber test failures, we should add a logging framework which logs to file (in order to not interfere with the default output of GoDog). In this file we should then for example log the commands executed and their results, etc.

agajdosi:
Took a brief look onto this today. I am successful with using StdoutPipe - `myout, _ := cmd.StdoutPipe()` for stdout of command. This gives a possibility to:

+ log the stdout into file realtime, having each line with exact time
+ could be also handy to implement an option of live preview of executed commands

Current code is quite "hacky", will clean it tomorrow and probably give some new info or WIP PR.

agajdosi:
Hi @hferentschik, if you have time, could you please take a look on [these lines](https://github.com/agajdosi/minishift/blob/issue607/test/integration/util/util.go#L46-L96), if my way is completely off or is fine?

I took a look on Tee struct and its methods which you mentioned on IRC, but I got no idea where/how it should be used? If better way is to implement Tee, please let me know, I will try to look better or will need to call for bjn call.

gbraad:
this makes the runCommand function much harder to understand and reason
about.
Is it not possible to wrap the actual call to runCommand and just use the
stdOut, stdErr, etc afrom return and store these?

agajdosi:
So I got logging into file working. One last thing is: `panic: interface conversion: interface {} is *gherkin.ScenarioOutline, not *gherkin.Scenario`, which I will fix tomorrow, forgot that these are two different types.

This is how log currently looks:

```
16:10:27.610979    info: Log initiated
16:10:27.611804    info: ----- scenario Version -----
16:10:27.611920 command: /home/agajdosi/golang/bin/minishift version
16:10:27.674226  stdout: -- Installing default add-ons ... OK
minishift v1.4.1+e5cc0e8
16:10:27.674660    info: ----- scenario Config -----
16:10:27.674719 command: /home/agajdosi/golang/bin/minishift config view
16:10:27.688485  stdout: 
16:10:27.689006    info: ----- scenario Fail -----
16:10:27.689077 command: /home/agajdosi/golang/bin/minishift configaro
16:10:27.699849  stdout: 
16:10:27.699869  stderr: Error: unknown command "configaro" for "minishift"
Run 'minishift --help' for usage.
unknown command "configaro" for "minishift"
16:10:27.700063 command: /home/agajdosi/golang/bin/minishift delete
[...]

PR #1306, opened 2017-08-24
Issue #607 Integration tests should log to file

agajdosi:
Adds possibility to log into file with package`test/integration/util`. One can use util.IntegrationTest.Print() or other functions to log, but recommended is to use LogMessage(info string, message string) which will take care of formatting and also adds info about the message, like if it is warning, stdout, stderr, info or anything else. With this function the should be hopefully nicely readable.

This PR also implemented the logging possibilities into actual tests.

### DIFF
--- a/docs/source/contributing/developing.adoc
+++ b/docs/source/contributing/developing.adoc
@@ -258,6 +258,14 @@ $ make integration GODOG_OPTS="-paths ~/tests/custom.feature,~/my.feature -tags 
 NOTE: When multiple values are used for options in `GODOG_OPTS`, then they have to be separated by a comma without whitespace.
 While `-tags basic,openshift` will be parsed properly by make, `-tags basic, openshift` will result in only `@basic` being used.
 
+==== Viewing Results
+
+Integration test logs its progress directly into a console with accent of providing additional useful information when failures of individual Gherkin steps happen.
+This information is often enough to find and debug the reason of failure.
+
+However for cases which needs further investigation the integration test also logs more detailed progress into a log file. This file is located at `$GOPATH/github.com/minishift/minishift/out/integration-test/integration.log`.
+Please note that this file is being deleted on start of each subsequent test run.
+
 [[format-source-code]]
 === Formatting the Source Code
 
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -244,12 +244,27 @@ func FeatureContext(s *godog.Suite) {
 			runner.CDKSetup()
 		}
 
+		util.StartLog(testDir)
+
 		fmt.Println("Running Integration test in:", testDir)
 		fmt.Println("Using binary:", minishiftBinary)
 	})
 
 	s.AfterSuite(func() {
+		util.LogMessage("info", "----- Cleaning Up -----")
 		minishift.runner.EnsureDeleted()
+		util.CloseLog()
+	})
+
+	s.BeforeScenario(func(this interface{}) {
+		switch this.(type) {
+		case *gherkin.Scenario:
+			scenario := *this.(*gherkin.Scenario)
+			util.LogMessage("info", fmt.Sprintf("----- Scenario: %s -----", scenario.ScenarioDefinition.Name))
+		case *gherkin.ScenarioOutline:
+			scenario := *this.(*gherkin.ScenarioOutline)
+			util.LogMessage("info", fmt.Sprintf("----- Scenario Outline: %s -----", scenario.ScenarioDefinition.Name))
+		}
 	})
 
 	s.AfterScenario(func(interface{}, error) {
--- a/test/integration/util/integration_logger.go
+++ b/test/integration/util/integration_logger.go
@@ -1,0 +1,103 @@
+// +build integration
+
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path"
+	"strings"
+)
+
+const (
+	messageInfoMaxLength = 7
+	timeHeaderLength     = 16
+)
+
+var (
+	IntegrationLog *log.Logger
+	logFile        *os.File
+)
+
+func init() {
+	IntegrationLog = log.New(ioutil.Discard, "", 0)
+}
+
+func StartLog(logPath string) error {
+	logPath = path.Join(logPath, "integration.log")
+	logFile, err := os.OpenFile(logPath, os.O_WRONLY|os.O_CREATE|os.O_APPEND, 0644)
+	if err != nil {
+		return err
+	}
+
+	IntegrationLog = log.New(logFile, "", log.Ltime|log.Lmicroseconds)
+	LogMessage("info", "Log Initiated")
+	fmt.Println("Log successfully started, logging into:", logPath)
+
+	return nil
+}
+
+func CloseLog() error {
+	err := logFile.Close()
+	return err
+}
+
+func LogMessage(messageInfo, message string) error {
+	messageInfo = formatMessageInfo(messageInfo)
+	message = formatMessage(message)
+	IntegrationLog.Print(messageInfo + message)
+
+	return nil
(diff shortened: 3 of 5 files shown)


agajdosi:
Example output from `/out/integration-test/integration.log`:
```
11:37:13.204340    info: Log Initiated
11:37:13.204429    info: ----- Scenario: User can install default add-ons -----
11:37:13.204476 command: /home/agajdosi/golang/bin/minishift status
11:37:13.287289  stdout: -- Installing default add-ons ... OK
                         Does Not Exist
11:37:13.287636 command: /home/agajdosi/golang/bin/minishift addons install --defaults
11:37:13.361391  stdout: Default add-ons anyuid, admin-user, xpaas, registry-route installed
11:37:13.361855    info: ----- Scenario: User can enable the anyuid add-on -----
11:37:13.361893 command: /home/agajdosi/golang/bin/minishift status
11:37:13.373158  stdout: Does Not Exist
11:37:13.373313 command: /home/agajdosi/golang/bin/minishift addons enable anyuid
11:37:13.388444  stdout: Add-on 'anyuid' enabled
11:37:13.388771    info: ----- Scenario: User can list enabled plugins -----
11:37:13.388861 command: /home/agajdosi/golang/bin/minishift status
11:37:13.410501  stdout: Does Not Exist
11:37:13.410842 command: /home/agajdosi/golang/bin/minishift addons list
11:37:13.425528  stdout: - anyuid         : enabled    P(0)
11:37:13.425870    info: ----- Scenario: Starting Minishift -----
11:37:13.425920 command: /home/agajdosi/golang/bin/minishift status
11:37:13.440508  stdout: Does Not Exist
11:37:13.440660 command: /home/agajdosi/golang/bin/minishift start
11:40:31.347381  stdout: -- Checking if KVM driver is installed ... 
                            Driver is available at /home/agajdosi/minishift/docker-machine-driver-kvm ... OK
                            Memory:    2 GB
                            vCPUs :    2
                            Disk size: 20 GB
                         -- Starting Minishift VM ... |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   /   -   \   |   //HERE IT GETS UGLY A BIT
[skipping few lines]
11:43:09.243984    info: ----- Scenario: Deleting Minishift -----
11:43:09.244021 command: /home/agajdosi/golang/bin/minishift status
11:43:09.304528  stdout: Stopped
11:43:09.304668 command: /home/agajdosi/golang/bin/minishift delete
11:43:09.797570  stdout: Deleting the Minishift VM...
                         Minishift VM deleted.
11:43:09.797972 command: /home/agajdosi/golang/bin/minishift status
11:43:09.807179  stdout: Does Not Exist
11:43:09.807381 command: /home/agajdosi/golang/bin/minishift ip
11:43:09.818459  stdout: 
11:43:09.818478  stderr: Error getting IP: Host does not exist: "minishift"
11:43:09.818706    info: ----- Cleaning Up -----
11:43:09.818720 command: /home/agajdosi/golang/bin/minishift delete
11:43:09.830964  stdout: Running this command requires an existing 'minishift' VM, but no VM is defined.
11:43:09.831054 command: /home/agajdosi/golang/bin/minishift status
11:43:09.841947  stdout: Does Not Exist
```

Overall 508 lines for basic scenario.

docs/source/contributing/developing.adoc
```diff
@@ -258,6 +258,14 @@ $ make integration GODOG_OPTS="-paths ~/tests/custom.feature,~/my.feature -tags
 NOTE: When multiple values are used for options in `GODOG_OPTS`, then they have to be separated by a comma without whitespace.
 While `-tags basic,openshift` will be parsed properly by make, `-tags basic, openshift` will result in only `@basic` being used.
 
+==== Viewing Results
+
+Integration test logs its progress directly into a console with accent of providing additional useful information when failures of individual Gherkin steps happen.
```
hferentschik: "The integration tests log their progress directly to the console"

hferentschik: "with accent" - not sure what you mean?

docs/source/contributing/developing.adoc
```diff
@@ -258,6 +258,14 @@ $ make integration GODOG_OPTS="-paths ~/tests/custom.feature,~/my.feature -tags
...
+This information is often enough to find and debug the reason of failure.
+
+However for cases which needs further investigation the integration test also logs more detailed progress into a log file. This file is located at `$GOPATH/github.com/minishift/minishift/out/integration-test/integration.log`.
```
hferentschik: "However, for cases needing further investigation, the integration tests also log more detailed progress into a log file"

test/integration/integration_test.go
```diff
@@ -244,12 +244,27 @@ func FeatureContext(s *godog.Suite) {
 			runner.CDKSetup()
 		}
 
+		util.StartLog(testDir)
```
hferentschik: +1 to puts these logging methods into util

test/integration/integration_test.go
```diff
@@ -244,12 +244,27 @@ func FeatureContext(s *godog.Suite) {
...
+		util.LogMessage("info", "----- Cleaning Up -----")
 		minishift.runner.EnsureDeleted()
+		util.CloseLog()
+	})
+
+	s.BeforeScenario(func(this interface{}) {
+		switch this.(type) {
+		case *gherkin.Scenario:
+			scenario := *this.(*gherkin.Scenario)
+			util.LogMessage("info", fmt.Sprintf("----- Scenario: %s -----", scenario.ScenarioDefinition.Name))
```
hferentschik: nice

test/integration/util/integration_logger.go
```diff
@@ -0,0 +1,103 @@
...
+	"path"
+	"strings"
+)
+
+const (
+	messageInfoMaxLength = 7
+	timeHeaderLength     = 16
+)
+
+var (
```
hferentschik: +1 nice to get all logging related things into a separate file

test/integration/util/integration_logger.go
```diff
@@ -0,0 +1,103 @@
...
+	IntegrationLog *log.Logger
+	logFile        *os.File
+)
+
+func init() {
+	IntegrationLog = log.New(ioutil.Discard, "", 0)
```
hferentschik: Why would you need this. This means that an instance is created even before main is called. The actual log instance is created in StartLog, right - 'IntegrationLog = log.New(logFile, "", log.Ltime|log.Lmicroseconds)'. What's the purpose of this?

test/integration/util/util.go
```diff
@@ -51,7 +52,9 @@ func runCommand(command string, commandPath string) (stdOut string, stdErr strin
 	cmd.Stdout = &outbuf
 	cmd.Stderr = &errbuf
 
+	LogMessage("command", fmt.Sprintf("%s %s", commandPath, command))
```
hferentschik: here we go :-)

docs/source/contributing/developing.adoc
agajdosi: @hferentschik Wanted to say that tests are designed to provide information, not just fail.

docs/source/contributing/developing.adoc
agajdosi: I need to take English comma course really, thanks.

test/integration/util/integration_logger.go
agajdosi: The reason is that before the beforeScenario(), where testDir is set, there is a call for isCDK() in testMain(). And there it would fai without init(), because isCDK() runs `minishift -h` where it searches for setup-cdk command and it would log into non-existing Logger.

Second option I had in my head was to add a check into LogMessage(), but this would need manual duplicate setup of value which would be equal to testDir which is not good for maintance and also it would not prevent the use of util.ntegrationLog.Print(). So I went for init(). (However I am not sure if there are any costs for doing init(). Seemed fine to me.)

hferentschik: I don't think cost is an issue, especially since this is a test harness. It is more about letting other people understand the code. So at the very least one would need a comment. 

That said, we are controlling the calls to the logger, right? We call it via the util.LogMessage function. Why don't you add a nil check for the logger there. If the logger is not configured return directly. For good measurement, you can print a message to the console.

docs/source/contributing/developing.adoc
hferentschik: "Integration tests log per default to the console, providing basic test progress information. This information is often enough to find and debug failure reasons. However, for cases needing further investigation, the integration tests also log more detailed information to  the log file `$GOPATH/github.com/minishift/minishift/out/integration-test/integration.log`."

Maybe something like that.

docs/source/contributing/developing.adoc
hferentschik: I am not sure about the second one either. The first one is needed for sure.

hferentschik:
Merged, thanks.

Issue #1328, opened 2017-08-30
minishift start with --show-libmachine-logs does not show oc cluster up command 


thomasmckay:
It was suggested that --show-libmachine-logs would show the exact 'oc cluster up' command being run. This is the output, though:
```
      MINISHIFT_ENABLE_EXPERIMENTAL=on minishift start --iso-url centos \
            --v 5 --show-libmachine-logs \
            --extra-clusterup-flags=--version=latest \
            --insecure-registry 172.30.0.0/16 --insecure-registry minishift --insecure-registry 192.168.0.24:53170
...snip...
About to run SSH command:
sudo mkdir -p /var/lib/minishift/openshift.local.config /var/lib/minishift/hostdata /var/lib/minishift/openshift.local.volumes /var/lib/minishift/openshift.local.pv
SSH cmd err, output: <nil>: 
-- Downloading OpenShift binary 'oc' version 'v3.6.0'
 34.72 MiB / 34.72 MiB [==================================================================] 100.00% 0s
-- OpenShift cluster will be configured with ...
   Version: v3.6.0
-- Checking `oc` support for startup flags ... 
   host-config-dir ... OK
   host-data-dir ... OK
   host-pv-dir ... OK
   host-volumes-dir ... OK
   routing-suffix ... OK
-- Extra `oc` cluster up flags (experimental) ... 
   '--version=latest'
Starting OpenShift using openshift/origin:latest ...
Pulling image openshift/origin:latest
Pulled 1/4 layers, 26% complete
Pulled 2/4 layers, 67% complete
Pulled 3/4 layers, 76% complete
Pulled 4/4 layers, 100% complete
Extracting
Image pull complete
-- Checking OpenShift client ... OK
-- Checking Docker client ... OK
-- Checking Docker version ... OK
-- Checking for existing OpenShift container ... OK
-- Checking for openshift/origin:latest image ... 
   Pulling image openshift/origin:latest
   Pulled 1/4 layers, 26% complete
   Pulled 2/4 layers, 67% complete
   Pulled 3/4 layers, 76% complete
   Pulled 4/4 layers, 100% complete
   Extracting
   Image pull complete
-- Checking Docker daemon configuration ... OK
-- Checking for available ports ... OK
-- Checking type of volume mount ... 
   Using nsenter mounter for OpenShift volumes
-- Creating host directories ... OK
-- Finding server IP ... 
   Using 192.168.42.115 as the server IP
-- Starting OpenShift container ... 
[...]

PR #1338, opened 2017-09-01
Issue #1328 Print 'oc cluster up' options

thomasmckay:


gbraad:
add to whitelist

thomasmckay:
@gbraad - This seemed least intrusive while still being useful. Shall I keep my changes and add glog to this file or to cmd/start.go?

gbraad:
Adding `glog` sounds good. After this I will try a test run and see if additional change might be needed.

Note: I do not want to really show TOO much information to the user (clutters). Maybe the values can be printed only when the verbosity level is increased?

praveenkumar:
@thomasmckay Do you need any help with this regard,  PR is not updated for quite some time now.

thomasmckay:
My suggestion is to add this line before runner.Run:
```
log.Debug(fmt.Sprintf("-- Running `oc` with: '%s'", strings.Join(cmdArgs, " ")))
```
I could not get glog.V(5).Info() to work w/ args of --v 5 --show-libmachine-logs. 

Thoughts?

pkg/minishift/clusterup/clusterup.go
```diff
@@ -99,6 +100,7 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 		}
 	}
 
+        log.Debug(fmt.Sprintf("-- Running `oc` with: '%s'", strings.Join(cmdArgs, " ")))
```
praveenkumar: @thomasmckay extra space begin of this log.Debug.

@LalatenduMohanty @gbraad I can see the command args for oc cluster up when using below which I think what we anticipating here.

```
$ minishift start --show-libmachine-logs --v=5
[...]
About to run SSH command:
sudo mkdir -p /var/lib/minishift/openshift.local.config /var/lib/minishift/hostdata /var/lib/minishift/openshift.local.volumes /var/lib/minishift/openshift.local.pv
SSH cmd err, output: <nil>: 
   Version: v3.6.0
   host-volumes-dir ... OK
   routing-suffix ... OK
   host-config-dir ... OK
   host-data-dir ... OK
   host-pv-dir ... OK
-- Running `oc` with: 'cluster up --use-existing-config --host-volumes-dir /var/lib/minishift/openshift.local.volumes --routing-suffix 192.168.42.79.nip.io --host-config-dir /var/lib/minishift/openshift.local.config --host-data-dir /var/lib/minishift/hostdata --host-pv-dir /var/lib/minishift/openshift.local.pv'
[...]
```

LalatenduMohanty: @praveenkumar @thomasmckay Tried the PR and it looks good to me.

LalatenduMohanty:
We need to fix the go fmt of the PR (just run `make fmt` and it will fix it for you) and rebase it with current master.

LalatenduMohanty:
@thomasmckay if you use glog as below it works fine. I recently found out. Need to check why `glog.V(2).Info` does not work for me. 

```
if glog.V(2) {
    anything
}

```

LalatenduMohanty:
The advantage of using glog is that we do not have to depend on --show-libmachine-logs for this log i.e. if user uses --v 2 or anything >2 user would be able to see the log.

pkg/minishift/clusterup/clusterup.go
```diff
@@ -99,6 +100,9 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 		}
 	}
 
+	if glog.V(5) {
+		fmt.Println("-- Running `oc` with: '%s'", strings.Join(cmdArgs, " "))
```
LalatenduMohanty: @thomasmckay fmt.Println can not print formatted strings. So you need to use fmt.Sprintf i.e. `fmt.Println(fmt.Sprintf("-- Running `oc` with: '%s'", strings.Join(cmdArgs, " ")))`

LalatenduMohanty: Also I think `fmt.Println(fmt.Sprintf("-- Running oc '%s'", strings.Join(cmdArgs, " ")))` would be better.

praveenkumar: > Also I think fmt.Println(fmt.Sprintf("-- Running oc '%s'", strings.Join(cmdArgs, " "))) would be better

I would say use `fmt.Printf("-- Running `oc` with: '%s'\n", strings.Join(cmdArgs, " "))` instead calling 2 different functions.

gbraad: `fmt.Printf("%s\n",sth)` is the preferred approach

budhrg: > I would say use fmt.Printf("-- Runningocwith: '%s'\n", strings.Join(cmdArgs, " ")) instead calling 2 different functions.

:+1:  This is fine @thomasmckay

thomasmckay: switched to fmt.Printf()

gbraad:
Ready to merge, but seems CentOS-CI is not progressing...

Issue #1339, opened 2017-09-01
Fix spinning wheel during startup for non b2d ISO


budhrg:
The spinning wheel is not properly showing when used with RHEL 7 or CentOS ISO.

```
$ minishift start --iso-url file:///home/budhram/cache/iso/minishift-rhel7.iso --username $USER --password $PASS
-- Checking if KVM driver is installed ... 
   Driver is available at /usr/local/bin/docker-machine-driver-kvm ... OK
-- Starting local OpenShift cluster using 'kvm' hypervisor ...
-- Minishift VM will be configured with ...
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ... | 
   Setting hostname ..| OK
OK
Registering machine using subscription-manager
Registration successful 45.1s
[...]
```

LalatenduMohanty:
@budhrg I never noticed the issue when I used CentOS or RHEL ISO. Can you please explain the issue?

budhrg:
@LalatenduMohanty , when did you tried?

Try with latest master, 
```
$ minishift version
minishift v1.5.0+0e75c4ec

$ minishift start --username bgurung@redhat.com --password '****' --iso-url file:///home/budhram/cache/iso/minishift-rhel7.iso
-- Checking if KVM driver is installed ... 
   Driver is available at /usr/local/bin/docker-machine-driver-kvm ... 
   Checking driver binary is executable ... OK
-- Starting local OpenShift cluster using 'kvm' hypervisor ...
-- Minishift VM will be configured with ...
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ... \ 
   Setting hostname ..| OK
OK
Registering machine using subscription-manager
[...]
```

LalatenduMohanty:
Fixed by https://github.com/minishift/minishift/issues/1340

PR #1340, opened 2017-09-01
Issue #1277 and #1339 Add no progress bar flag and fix spinning wheel

### LINKED ISSUE #1277: Add a flag for non-interactive shells or detect them automatically
When you launch minishift/cdk from Eclipse, it uses the Console view to show the output.
At the beginning there is this spinning thing showing progress. Unfortunately it doesn't behave well in the non-interactive console view:
```
--
[37m\|[0m   [37m/[0m   [37m-[0m   [37m\[0m   [37m\|[0m   [37m/[0m   [37m-[0m   [37m\[0m
```

See more details at https://issues.jboss.org/browse/JBIDE-24676 .

Rob suggested that we need a flag to turn this off. Another option would be to detect the shell automatically and output something non-changing in a non-interactive shell.

budhrg:
Fix #1277  and #1339 .

```
$ minishift start
-- Installing default add-ons ... OK
[...]

   Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso'
 40.00 MiB / 40.00 MiB [==========================================================================================================================================================] 100.00% 0s
[...]
-- Downloading OpenShift binary 'oc' version 'v3.6.0'
 12.04 MiB / 34.72 MiB [=====================================================>---------------------------------------------------------------------------------------------------]  34.69% 24s^C



$ MINISHIFT_NO_PROGRESSBAR=y minishift start
-- Installing default add-ons ... OK
[...]
   Downloading ISO 'https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso'
[...]
-- Downloading OpenShift binary 'oc' version 'v3.6.0'



$ minishift start --iso-url file:///home/budhram/cache/iso/minishift-rhel7.iso --username $USER --password $PASS
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
[...]
-- Downloading OpenShift binary 'oc' version 'v3.6.0'
 798.53 KiB / 34.72 MiB [===>---------------------------------------------------------------------------------------------------------------------------------------------------]   2.25% 7m7s


$ MINISHIFT_NO_PROGRESSBAR=y minishift start --iso-url file:///home/budhram/cache/iso/minishift-rhel7.iso --username $USER --password $PASS
[...]
Registration successful 40.6s
   Pinging 8.8.8.8 ... OK
-- Checking available disk space ... 1% OK
-- Downloading OpenShift binary 'oc' version 'v3.6.0'
...
```

### DIFF
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -70,7 +70,7 @@ var viperWhiteList = []string{
 	"log_dir",
 }
 
-var hasEnabledExperimental bool = false
+var hasEnabledExperimental bool
 
 // RootCmd represents the base command when called without any subcommands
 var RootCmd = &cobra.Command{
@@ -166,13 +166,16 @@ func setFlagsUsingViper() {
 	}
 }
 
-func init() {
+func processEnvVariables() {
 	enableExperimental, err := util.GetBoolEnv(enableExperimentalEnv)
 	if err == util.BooleanFormatError {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error enabling experimental features: %s", err))
 	}
 	hasEnabledExperimental = enableExperimental
+}
 
+func init() {
+	processEnvVariables()
 	RootCmd.PersistentFlags().Bool(showLibmachineLogs, false, "Show logs from libmachine.")
 	RootCmd.AddCommand(configCmd.ConfigCmd)
 	RootCmd.AddCommand(cmdOpenshift.OpenShiftCmd)
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -22,10 +22,8 @@ import (
 	"strings"
 
 	"os"
-	"time"
 
 	"github.com/asaskevich/govalidator"
-	"github.com/briandowns/spinner"
 	units "github.com/docker/go-units"
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/drivers"
@@ -34,7 +32,7 @@ import (
 	"github.com/golang/glog"
 	"github.com/minishift/minishift/cmd/minishift/cmd/addon"
 	configCmd "github.com/minishift/minishift/cmd/minishift/cmd/config"
-	cmdutil "github.com/minishift/minishift/cmd/minishift/cmd/util"
+	cmdUtil "github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/cache"
@@ -49,7 +47,6 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
 
 	"github.com/minishift/minishift/pkg/util"
-	inputUtils "github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	stringUtils "github.com/minishift/minishift/pkg/util/strings"
 	"github.com/minishift/minishift/pkg/version"
@@ -67,8 +64,6 @@ const (
 	defaultInsecureRegistry = "172.30.0.0/16"
 
 	unsupportedIsoUrlFormat = "Unsupported value for iso-url. It can be an URL, file URI or one of the following short names: [b2d centos]."
-
-	SPINNER_SPEED = 200 // In milliseconds
 )
 
 var (
@@ -82,28 +77,28 @@ var (
 		Name:      configCmd.DockerEnv.Name,
 		Shorthand: "",
 		Usage:     "Environment variables to pass to the Docker daemon. Use the format <key>=<value>.",
-		Value:     cmdutil.NewStringSliceValue([]string{}, &[]string{}),
+		Value:     cmdUtil.NewStringSliceValue([]string{}, &[]string{}),
 	}
 
 	insecureRegistryFlag = &flag.Flag{
 		Name:      configCmd.InsecureRegistry.Name,
 		Shorthand: "",
 		Usage:     "Non-secure Docker registries to pass to the Docker daemon.",
-		Value:     cmdutil.NewStringSliceValue([]string{defaultInsecureRegistry}, &[]string{}),
+		Value:     cmdUtil.NewStringSliceValue([]string{defaultInsecureRegistry}, &[]string{}),
 	}
 
 	dockerEngineOptFlag = &flag.Flag{
 		Name:      configCmd.DockerEngineOpt.Name,
 		Shorthand: "",
 		Usage:     "Specify arbitrary flags to pass to the Docker daemon in the form <flag>=<value>.",
-		Value:     cmdutil.NewStringSliceValue([]string{}, &[]string{}),
+		Value:     cmdUtil.NewStringSliceValue([]string{}, &[]string{}),
 	}
 
 	registryMirrorFlag = &flag.Flag{
 		Name:      configCmd.RegistryMirror.Name,
 		Shorthand: "",
 		Usage:     "Registry mirrors to pass to the Docker daemon.",
-		Value:     cmdutil.NewStringSliceValue([]string{}, &[]string{}),
+		Value:     cmdUtil.NewStringSliceValue([]string{}, &[]string{}),
 	}
 )
 
@@ -170,7 +165,7 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	// to determine whether we need to run post cluster up actions,
 	// we need to determine whether this is a restart prior to potentially creating a new VM
-	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
+	isRestart := cmdUtil.VMExists(libMachineClient, constants.MachineName)
 
 	fmt.Printf("-- Starting local OpenShift cluster")
 	hostVm := startHost(libMachineClient)
@@ -212,7 +207,7 @@ func runStart(cmd *cobra.Command, args []string) {
 		Project:          defaultProject,
 		KubeConfigPath:   constants.KubeConfigPath,
 		OcPath:           ocPath,
-		AddonEnv:         viper.GetStringSlice(cmdutil.AddOnEnv),
+		AddonEnv:         viper.GetStringSlice(cmdUtil.AddOnEnv),
 	}
 
 	clusterUpParams := determineClusterUpParameters(clusterUpConfig)
@@ -255,7 +250,7 @@ func getRequiredHostDirectories() []string {
 }
 
 func handleProxies() *util.ProxyConfig {
-	proxyConfig, err := util.NewProxyConfig(viper.GetString(cmdutil.HttpProxy), viper.GetString(cmdutil.HttpsProxy), viper.GetString(configCmd.NoProxyList.Name))
+	proxyConfig, err := util.NewProxyConfig(viper.GetString(cmdUtil.HttpProxy), viper.GetString(cmdUtil.HttpsProxy), viper.GetString(configCmd.NoProxyList.Name))
 
 	if err != nil {
 		atexit.ExitWithMessage(1, err.Error())
@@ -269,11 +264,11 @@ func handleProxies() *util.ProxyConfig {
 		// It could be that the proxy config is retrieved from the environment. To make sure that
 		// proxy settings are properly passed to cluster up we need to explicitly set the values.
 		if proxyConfig.HttpProxy() != "" {
-			viper.Set(cmdutil.HttpProxy, proxyConfig.HttpProxy())
+			viper.Set(cmdUtil.HttpProxy, proxyConfig.HttpProxy())
 		}
 
 		if proxyConfig.HttpsProxy() != "" {
-			viper.Set(cmdutil.HttpsProxy, proxyConfig.HttpsProxy())
+			viper.Set(cmdUtil.HttpsProxy, proxyConfig.HttpsProxy())
 		}
 		viper.Set(configCmd.NoProxyList.Name, proxyConfig.NoProxy())
 	}
@@ -302,7 +297,7 @@ func determineInsecureRegistry(key string) []string {
 }
 
 func startHost(libMachineClient *libmachine.Client) *host.Host {
-	spinnerView := spinner.New(spinner.CharSets[9], SPINNER_SPEED*time.Millisecond)
+	progressDots := make(chan bool)
 
 	// Configuration used for creation/setup of the Virtual Machine
 	machineConfig := &cluster.MachineConfig{
@@ -323,7 +318,7 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 	var hostVm *host.Host
 
 	// configuration with these settings only happen on create
-	isRestart := cmdutil.VMExists(libMachineClient, constants.MachineName)
+	isRestart := cmdUtil.VMExists(libMachineClient, constants.MachineName)
 	if !isRestart {
 		fmt.Println("-- Minishift VM will be configured with ...")
 		fmt.Println("   Memory:   ", units.HumanSize(float64((machineConfig.Memory/units.KiB)*units.GB)))
@@ -335,12 +330,12 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 
 	cacheMinishiftISO(machineConfig)
 
-	fmt.Print("-- Starting Minishift VM ... ")
-	spinnerView.Start()
+	fmt.Print("-- Starting Minishift VM ...")
+	util.StartProgressDots(progressDots)
 	start := func() (err error) {
 		hostVm, err = cluster.StartHost(libMachineClient, *machineConfig)
 		if err != nil {
-			fmt.Printf("FAIL ")
+			fmt.Print(" FAIL ")
 			glog.Errorf("Error starting the VM: %s. Retrying.\n", err)
 		}
 		return err
@@ -349,9 +344,9 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error starting the VM: %v", err))
 	}
-	spinnerView.Stop()
-	fmt.Println("OK")
+	util.StopProgressDots(progressDots)
 
+	fmt.Println(" OK")
 	return hostVm
 }
 
@@ -474,7 +469,7 @@ func initStartFlags() *flag.FlagSet {
 	startFlagSet.AddFlag(dockerEngineOptFlag)
 	startFlagSet.AddFlag(insecureRegistryFlag)
 	startFlagSet.AddFlag(registryMirrorFlag)
-	startFlagSet.AddFlag(cmdutil.AddOnEnvFlag)
+	startFlagSet.AddFlag(cmdUtil.AddOnEnvFlag)
 
 	return startFlagSet
 }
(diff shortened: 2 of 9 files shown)


pkg/minishift/registration/registration.go
```diff
@@ -36,10 +36,6 @@ type Detector interface {
 
 type StandardRegistrator struct{}
 
-func SetDetector(newDetector Detector) {
```
budhrg: Unused code.

pkg/minishift/registration/utils.go
```diff
@@ -31,13 +31,6 @@ type RegistrationParameters struct {
 	GetPasswordInteractive func(message string) string
 }
 
-// Detect supported Registrator
-func NeedRegistration(host *host.Host) (bool, error) {
```
budhrg: Unused code.

gbraad:
I will hold off for a short moment to merge this.

Want feedback on the spinner library and see if this can be fixed in a different way.
Adding a flag or feature is easy, but it will be difficult to remove this.

Also, has the OP responded to the suggestion of using an environment variable... will this work for them?

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -111,12 +111,12 @@ func (provisioner *MinishiftProvisioner) Provision(swarmOptions swarm.Options, a
 	}
 	provisioner.EngineOptions.StorageDriver = storageDriver
 
-	fmt.Printf("\n   Setting hostname ... ")
+	log.Info("\n   Setting hostname ... ")
```
gbraad: This is not part of the same issue

pkg/minishift/registration/redhat.go
```diff
@@ -63,8 +64,12 @@ func (registrator *RedHatRegistrator) CompatibleWithDistribution(osReleaseInfo *
 
 // Register attempts to register the system with RHSM
 func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) error {
+	var spinnerView *spinner.Spinner
+
 	if isRegistered, err := registrator.isRegistered(); !isRegistered && err == nil {
-		spinnerView := spinner.New(spinner.CharSets[9], SPINNER_SPEED*time.Millisecond)
+		if !cmdUtil.HasEnabledNoProgressbar {
```
gbraad: I believe we should use `ShowProgressbars` (default `true`) and allow the `ProcessEnvironmentVariables` function to toggle this off. This sentence has a double negation... :-s

pkg/minishift/registration/redhat.go
```diff
@@ -87,11 +92,15 @@ func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) er
 				param.Username,
 				minishiftStrings.EscapeSingleQuote(param.Password))
 
-			// start timed SSH command to register
-			spinnerView.Start()
+			if spinnerView != nil {
```
gbraad: and if done so, we do not check for `if spinnerView != nil`, but only need to concern about `if ShowProgressBars { ... }`

pkg/minishift/registration/utils.go
```diff
@@ -47,7 +40,7 @@ func doRegistrationHostAction(actionMessage string, registrationAction Registrat
 		return false, err
 	}
 
-	// was succesfully
+	// was successfully
```
gbraad: if you change it, please correct it to `was successful` (no adjective form, as I actually forgot to add `completed`)

pkg/util/github/github.go
```diff
@@ -133,7 +134,7 @@ func DownloadOpenShiftReleaseBinary(binaryType OpenShiftBinaryType, osType minis
 		defer func() { _ = httpResp.Body.Close() }()
 
 		asset = httpResp.Body
-		if httpResp.ContentLength > 0 {
+		if !util.HasEnabledNoProgressbar && httpResp.ContentLength > 0 {
```
gbraad: if `ShowProgressBars &&`

pkg/minishift/registration/registration.go
gbraad: Create an issue for removing the code and create a separate commit for this...

pkg/minishift/registration/utils.go
gbraad: ditto... not part of this commit

cmd/minishift/cmd/root.go
```diff
@@ -166,13 +166,22 @@ func setFlagsUsingViper() {
...
 	if err == util.BooleanFormatError {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error enabling experimental features: %s", err))
 	}
 	hasEnabledExperimental = enableExperimental
 
+	noProgressbar, err := util.GetBoolEnv(util.EnableNoProgressbarEnv)
+	if err == util.BooleanFormatError {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error enabling no progress bar action: %s", err))
+	}
+	util.HasEnabledNoProgressbar = noProgressbar
```
gbraad: `ShowProgressBars = !enableNoProgressBars`

Not sure if `util` is the correct place for this boolean. Sounds more like `config`

gbraad:
I will hold off for a short moment to merge this.

Want feedback on the spinner library and see if this can be fixed in a different way.
Adding a flag or feature is easy, but it will be difficult to remove this.

Also, has the OP responded to the suggestion of using an environment variable... will this work for them?

pkg/minishift/registration/redhat.go
budhrg: Ack

budhrg:
> Also, has the OP responded to the suggestion of using an environment variable... will this work for them?

gbraad:
In the library colorization is used (forced to use white), and this is written to the output with https://github.com/briandowns/spinner/blob/master/spinner.go#L108. I believe Eclipse just does not handle ANSI codes (as used for colorization) correctly.

Can you test if turning off the colorization will work? If so, this can be a fix to the upstream library in which this can be turned off. Maybe by modifying this line first?

gbraad:
We can use an alternative strategy.

We use our own Charset: https://github.com/briandowns/spinner/blob/master/character_sets.go with only `{".", " "}` which then will print a blinking dot. but on the Eclipse console becomes a spaced dot indicating the progress? it is worth a shot. WDYT?

gbraad:
```
package main

import (
        "github.com/briandowns/spinner"
        "time"
)

func main() {
        s := spinner.New([]string{".", " "}, time.Second)
        s.Start() 
        time.Sleep(30 * time.Second)
        s.Stop()
}
```

budhrg:
> We use our own Charset: https://github.com/briandowns/spinner/blob/master/character_sets.go with only {".", " "} which then will print a blinking dot. but on the Eclipse console becomes a spaced dot indicating the progress? it is worth a shot. WDYT?

@gbraad checking this behavior eclipse/devstudio.

gbraad:
On both Linux and Windows. At the moment I can not test it... but it looked
'OK-ish' here...

pkg/minishift/provisioner/minishift_provisioner.go
budhrg: @gbraad , getting some spinning wheel issue on print it here. See https://github.com/minishift/minishift/issues/1339

budhrg:
Now, the logs are shown as : 
```
$ minishift start --username bgurung@redhat.com --password '*****' --iso-url file:///home/budhram/cache/iso/minishift-rhel7.iso
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ............ OK
-- Registering machine using subscription-manager ............... OK [1m0s]
   Pinging 8.8.8.8 ... OK
-- Checking available disk space ... 1% OK
   Version: v3.6.0
   routing-suffix ... OK
   host-config-dir ... OK
   host-data-dir ... OK
   host-pv-dir ... OK
   host-volumes-dir ... OK
[...]
```

gbraad:
This is VERY acceptable...

cmd/minishift/cmd/util/util.go
```diff
@@ -71,3 +72,17 @@ func UnpackAddons(addonsDir string) error {
 
 	return nil
 }
+
+func ShowProgressDots(progressDots chan bool) {
```
gbraad: +1

pkg/minishift/provisioner/minishift_provisioner.go
gbraad: This is not how it was originally printed. The OK was not on the informative line about setting the hostname.

cmd/minishift/cmd/start.go
```diff
@@ -22,10 +22,8 @@ import (
 	"strings"
 
 	"os"
-	"time"
 
 	"github.com/asaskevich/govalidator"
-	"github.com/briandowns/spinner"
```
gbraad: remove from glide.yaml and glide.lock

glide.lock
```diff
@@ -1,13 +1,10 @@
-hash: 8a468283a9a4675c087c7d25a00ce4a72c4fc0f5f05abe60eec93bfce22be14b
-updated: 2017-07-10T11:56:37.047777606+02:00
+hash: 675036c36eaf7c64def2072063154d6e2f3d670a2c09c4396db9b2359b83bdcc
+updated: 2017-09-04T20:16:55.757240937+05:30
 imports:
 - name: github.com/asaskevich/govalidator
   version: 7b3beb6df3c42abd3509abfc3bcacc0fbfb7c877
 - name: github.com/blang/semver
   version: b38d23b8782a487059e8fc8773e9a5b228a77cb6
-- name: github.com/briandowns/spinner
```
gbraad: OK

pkg/minishift/registration/utils.go
```diff
@@ -93,7 +93,7 @@ func UnregisterHostVM(host *host.Host, param *RegistrationParameters) (bool, err
 	}
 
 	return doRegistrationHostAction(
-		"Unregistering machine",
+		"Unregistering machine\n",
```
gbraad: isn't this needed for all the registration host actions? if soo, make that a println

pkg/minishift/registration/utils.go
```diff
@@ -41,13 +41,13 @@ func NeedRegistration(host *host.Host) (bool, error) {
 type RegistrationHostActionFunc func(param *RegistrationParameters) error
 
 func doRegistrationHostAction(actionMessage string, registrationAction RegistrationHostActionFunc, param *RegistrationParameters) (bool, error) {
-	fmt.Println(actionMessage)
+	fmt.Print(actionMessage)
```
gbraad: ????? and later you add a `\n` ?

pkg/minishift/registration/utils.go
```diff
@@ -71,7 +71,7 @@ func RegisterHostVM(host *host.Host, param *RegistrationParameters) (bool, error
 	}
 
 	return doRegistrationHostAction(
-		"Registering machine using subscription-manager",
+		"-- Registering machine using subscription-manager ...",
```
gbraad: because of the progress dots?

pkg/minishift/registration/utils.go
gbraad: I see...

pkg/minishift/registration/utils.go
gbraad: Maybe this should be on a new line anyways, with an indicative message.

because when I need to specify the username and password interactively, how is it shown now?

cmd/minishift/cmd/util/util.go
```diff
@@ -71,3 +72,17 @@ func UnpackAddons(addonsDir string) error {
...
+	go func() {
+		for {
+			select {
+			case <-progressDots:
+				return
+			default:
+				fmt.Print(".")
+				time.Sleep(5 * time.Second)
```
gbraad: Maybe configurable... 5 seconds between a dot is quite long. 

The general human response time to think something is acted on is about 300ms. So, printing needs to happen fast before it starts a long process... but it is possible to take about 3 seconds before the brain needs another stimulus. 5 stretches it... so in the case of the registration it might need more ? but for starting the VM it can be faster...

WDYT? what gave the best results?

pkg/util/utils.go
```diff
@@ -137,7 +137,7 @@ func TimeElapsed(start time.Time, friendly bool) string {
 		elapsed = FriendlyDuration(elapsed)
 	}
 
-	return elapsed.String()
+	return fmt.Sprintf("[%v]", elapsed.String())
```
gbraad: Why the brackets?

pkg/minishift/registration/redhat.go
```diff
@@ -87,15 +86,16 @@ func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) er
...
+			// stops the progress of dots
+			progressDots <- true
 			if err == nil {
-				fmt.Print("Registration successful ")
+				fmt.Print(" OK ")
 			} else {
-				fmt.Print("Registration unsuccessful ")
+				fmt.Print(" FAIL ")
 			}
 			fmt.Println(util.TimeElapsed(startTime, true))
```
gbraad: If you want brackets, you would have to put them here...

The function https://github.com/budhrg/minishift/blob/03606c244dcf5030a2e0c2d357d7f0f888fb2c59/pkg/util/utils.go#L133 is a util function, and would need to produce a result that can be processed easily without side-effects or markup.

pkg/util/utils.go
gbraad: Brackets should not be added here... but where the output is used:
https://github.com/budhrg/minishift/blob/03606c244dcf5030a2e0c2d357d7f0f888fb2c59/pkg/minishift/registration/redhat.go#L100

pkg/minishift/registration/redhat.go
budhrg: Sure

cmd/minishift/cmd/util/util.go
budhrg: 3 seconds looks good. Configurable make sense. For registration, it takes long so will keep three and for starting VM 2 second probably.

budhrg:
@gbraad , @LalatenduMohanty , I updated the PR make interactive and non-active registration consistent

```
#----------- Non-Interactive Registration
-- Starting Minishift VM ........................ OK
Red Hat Developers or Red Hat Subscription Management (RHSM) username: bgurung@redhat.com
Red Hat Developers or Red Hat Subscription Management (RHSM) password: [HIDDEN]
...............
Registration successful [30s]


#----------- Interactive Registration

-- Starting Minishift VM ........................ OK
................
Registration successful [32s]

```

pkg/minishift/registration/utils.go
budhrg: @gbraad , Updated the PR now. It was not looking intutive :)

```
-- Registering machine using subscription-manager ...Red Hat Developers or Red Hat Subscription Management (RHSM) username: bgurung@redhat.com
Red Hat Developers or Red Hat Subscription Management (RHSM) password: [HIDDEN]
........... OK [33s]
```

cmd/minishift/cmd/root.go
```diff
@@ -70,7 +70,7 @@ var viperWhiteList = []string{
 	"log_dir",
 }
 
-var hasEnabledExperimental bool = false
+var hasEnabledExperimental bool
```
LalatenduMohanty: 👍

gbraad:
@budhrg maybe indicating what is happening as follows?
```
Awaiting response ................ OK [32s]
```

cmd/minishift/cmd/util/util.go
```diff
@@ -71,3 +72,17 @@ func UnpackAddons(addonsDir string) error {
...
+
+func ShowProgressDots(progressDots chan bool) {
+	go func() {
+		for {
+			select {
+			case <-progressDots:
+				return
+			default:
+				fmt.Print(".")
+				time.Sleep(2 * time.Second)
```
LalatenduMohanty: I think we should use 1 second instead of 2 seconds. While testing it I felt the dots were little slow.

cmd/minishift/cmd/util/util.go
LalatenduMohanty: I think we should name it `StartProgressDots`.

LalatenduMohanty: Also I think we should move this function to minishift pkg/Util.

cmd/minishift/cmd/start.go
```diff
@@ -349,9 +345,10 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error starting the VM: %v", err))
 	}
-	spinnerView.Stop()
-	fmt.Println("OK")
+	// Stops the progress of dots
+	progressDots <- true
```
LalatenduMohanty: `progressDots <- true` should be in a function. Even if it is just one line a function would help to encapsulate it and if we want to change or add more to it in the future it would be easy.

cmd/minishift/cmd/util/util.go
LalatenduMohanty: Tested the updated code and 2 seconds seems fine now.

cmd/minishift/cmd/start.go
gbraad: Agree... abstract it.

Maybe even as a package for just the progress dots, so it reads as
```
progress.Start()
...
proogress.Stop()
```

budhrg:
@LalatenduMohanty , @gbraad , Updated as per our discussion

```
#------------------- Non-Interactive Registration -----------
$ minishift start --username bgurung@redhat.com --password '****' --iso-url file:///home/budhram/cache/iso/minishift-rhel7.iso
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ....................... OK
Registration in progress ............... OK [30s]
[...]


#------------------- Interactive Registration -----------

$ minishift start --iso-url file:///home/budhram/cache/iso/minishift-rhel7.iso
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ....................... OK
Red Hat Developers or Red Hat Subscription Management (RHSM) username: bgurung@redhat.com
Red Hat Developers or Red Hat Subscription Management (RHSM) password: [HIDDEN]
Registration in progress ................. OK [34s]
[...]
```

pkg/minishift/provisioner/minishift_provisioner.go
budhrg: @LalatenduMohanty , WDYT about this change? Right now if we print this, it is altering the expected progress dots behavior. See above comment.

budhrg:
@LalatenduMohanty , 

> We should fix the alignment of "Invalid username or password. Retry:

Fixed now

```
-- Starting Minishift VM ......................... OK
-- Registering machine using subscription-manager 
   Red Hat Developers or Red Hat Subscription Management (RHSM) username: basfd
   Red Hat Developers or Red Hat Subscription Management (RHSM) password: [HIDDEN]
   Registration in progress .. FAIL [4s]
   Invalid username or password. Retry: 1
   Red Hat Developers or Red Hat Subscription Management (RHSM) username: bgurung@redhat.com
   Red Hat Developers or Red Hat Subscription Management (RHSM) password: [HIDDEN]
   Registration in progress ............... OK [30s]
....
```

pkg/minishift/provisioner/minishift_provisioner.go
LalatenduMohanty: I am ok with using `glog.V(2).Infof` or log.Info as we need to fix it anyway with #169 .

budhrg: With `log.Info`, this message will appear in `--show-libmachine-logs` flag enabled as well. Hence, I am in favor of it.

pkg/minishift/registration/redhat.go
```diff
@@ -63,9 +61,9 @@ func (registrator *RedHatRegistrator) CompatibleWithDistribution(osReleaseInfo *
 
 // Register attempts to register the system with RHSM
 func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) error {
-	if isRegistered, err := registrator.isRegistered(); !isRegistered && err == nil {
-		spinnerView := spinner.New(spinner.CharSets[9], SPINNER_SPEED*time.Millisecond)
+	progressDots := make(chan bool)
```
gbraad: This should have been in a function too...

cmd/minishift/cmd/start.go
```diff
@@ -302,7 +297,7 @@ func determineInsecureRegistry(key string) []string {
 }
 
 func startHost(libMachineClient *libmachine.Client) *host.Host {
-	spinnerView := spinner.New(spinner.CharSets[9], SPINNER_SPEED*time.Millisecond)
+	progressDots := make(chan bool)
```
gbraad: should be abstracted

PR #1349, opened 2017-09-05
Issue #1348 Abstract the code to create progress dots

### LINKED ISSUE #1348: Abstract the code to create progress dots
As per the fix of #1277 in the PR #1340 , we missed extracting the code to create progress dots here https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/start.go#L300.

Probably, good to extract this into `pkg/util/utils.go` along with functions like `StartProgressDots`.

budhrg:
Fix #1348

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -48,6 +48,7 @@ import (
 
 	"github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/minishift/minishift/pkg/util/progress"
 	stringUtils "github.com/minishift/minishift/pkg/util/strings"
 	"github.com/minishift/minishift/pkg/version"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


gbraad:
Had placed the code on the wrong pull request, so here goes again:

Move to subdir of pkg/util

```go
package progress

// Returns a new instance of the ProgressDots with a given interval
func New(interval int) *ProgressDots{
	return &ProgressDots{
            ...
	}
}

type ProgressDots struct {
    interval int
}

func (s *ProgressDots) Start() {
}

func (s *ProgressDots) Stop() {
}
```

cmd/minishift/cmd/start.go
```diff
@@ -48,6 +48,7 @@ import (
 
 	"github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/minishift/minishift/pkg/util/progress"
```
LalatenduMohanty: The package name `progress` should be changed to `progressDots` or something better.

gbraad: package names do not have casing progressdots is possible, but not sure if this really adds value.

gbraad: If progressDots is used as a package name, it clashes with:
```
progressDots := progress.New()
```

and therefore would need an override from the import to use
```
    progressdots"github.com/minishift/minishift/pkg/util/progressDots"
```

`progressdots` it is... changing and merging...

gbraad:
Merged as https://github.com/minishift/minishift/commit/9aa240de4dff47619b3823a96291553fd46b5a21

Issue #1361, opened 2017-09-11
minishift status returns "Does Not Exist" when launched from a terminal without admin right and minishift start was launched from a terminal with admin rights


apupier:
'minishift setup-cdk' requires admin privilege
if using 'minishift start' in the same terminal with admin privileges,
then if launched in a terminal without admin rights, 'minishift status' will return "Does Not Exist"

apupier:
@budhrg I think that it was the issue that we encountered during our call

LalatenduMohanty:
@apupier you should not need admin privilege for running setup-cdk. Are you getting any error when you run it as an user?  The issue you are facing i.e. ` 'minishift status' will return "Does Not Exist"` should be gone if you run setup-cdk without admin previlage. 

Also it is an CDK issue so we will move it to CDK Jira https://issues.jboss.org/projects/CDK/issues/.

apupier:
my bad, it is 'minishift start' which requires admin right

otherwise there is:

> C:\Users\Aurelien Pupier>minishift start
> Starting local OpenShift cluster using 'hyperv' hypervisor...
> E0911 15:28:20.827670   14712 start.go:282] Error starting the VM: Error creating the VM. Error with pre-create check: "Hyper-v commands have to be run as an Administrator". Retrying.
> E0911 15:28:23.485364   14712 start.go:282] Error starting the VM: Error creating the VM. Error with pre-create check: "Hyper-v commands have to be run as an Administrator". Retrying.
> E0911 15:28:26.142630   14712 start.go:282] Error starting the VM: Error creating the VM. Error with pre-create check: "Hyper-v commands have to be run as an Administrator". Retrying.
> Error starting the VM: Error creating the VM. Error with pre-create check: "Hyper-v commands have to be run as an Administrator"
> Error creating the VM. Error with pre-create check: "Hyper-v commands have to be run as an Administrator"
> Error creating the VM. Error with pre-create check: "Hyper-v commands have to be run as an Administrator"

gbraad:
Are you sure thr user is allowed to use HyperV? The user needs to be part of the HyperV adminstrators group

budhrg:
@apupier , I doubt you ran `minishift setup-cdk` with admin rights.

When we did last time in the call, it was fine AFAIR. 
Could you retry again?

apupier:
> Are you sure thr user is allowed to use HyperV? The user needs to be part of the HyperV adminstrators group

[...]

PR #1367, opened 2017-09-12
Add PowerShell wrapper and check if Hyper-V installed

### LINKED ISSUE #1368: Add PowerShell wrapper and check if Hyper-V installed

gbraad:
Fixes #1368 

Needed for Hyper-V preflight check and static IP address...

### DIFF
--- a/cmd/minishift/cmd/start_preflight.go
+++ b/cmd/minishift/cmd/start_preflight.go
@@ -26,7 +26,8 @@ import (
 
 	"github.com/docker/machine/libmachine/drivers"
 	configCmd "github.com/minishift/minishift/cmd/minishift/cmd/config"
-	miniutil "github.com/minishift/minishift/pkg/minishift/util"
+	"github.com/minishift/minishift/pkg/minishift/shell/powershell"
+	miniUtil "github.com/minishift/minishift/pkg/minishift/util"
 
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/viper"
@@ -38,6 +39,8 @@ const (
 
 // preflightChecksAfterStartingHost is executed before the startHost function.
 func preflightChecksBeforeStartingHost() {
+	driverErrorMessage := "See the 'Setting Up the Driver Plug-in' topic for more information"
+
 	switch viper.GetString(configCmd.VmDriver.Name) {
 	case "xhyve":
 		preflightCheckSucceedsOrFails(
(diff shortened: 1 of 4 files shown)


gbraad:
retest this please

gbraad:
failing due to #1369

glide.yaml
```diff
@@ -61,6 +61,10 @@ import:
 - package: github.com/DATA-DOG/godog
   version: v0.6.2
 - package: github.com/elazarl/goproxy
+- package: github.com/gorillalabs/go-powershell
+  version: 3bc7a60b1df80a5766820c4a53d59668f4553f75
+- package: github.com/juju/errors
```
budhrg: @gbraad , are you using this package anywhere?

cmd/minishift/cmd/start_preflight.go
```diff
@@ -38,28 +39,30 @@ const (
 
 // preflightChecksAfterStartingHost is executed before the startHost function.
 func preflightChecksBeforeStartingHost() {
+	driverErrorMessage := "See the 'Setting Up the Driver Plug-in' topic for more information"
```
budhrg: IMO, documentation link should be given here. https://docs.openshift.org/latest/minishift/getting-started/setting-up-driver-plugin.html

budhrg: Also, since `getting-started/setting-up-driver-plugin` doesn't have mention of being member of HyperV group either we need to mention there that one need to part of hyper-v group or point to https://docs.openshift.org/latest/minishift/troubleshooting/troubleshooting-driver-plugins.html#insufficient-privileges documentation.

glide.yaml
gbraad: compilation fails if not included, since go-powershell does not do proper dependency management I guess. Glide can not resolve this.

budhrg: Ack

cmd/minishift/cmd/start_preflight.go
gbraad: Well, none of the drivers actually specify a link online.
And maybe, the permissions should be restructured in the documentation? I do not think we should have to link to two possible locations. However, I believe the application should never have to point to a location at all.

budhrg: > And maybe, the permissions should be restructured in the documentation?

Works. We need to document this member role as part of https://docs.openshift.org/latest/minishift/getting-started/setting-up-driver-plugin.html#hyper-v-driver

gbraad: Can we do this as a separate issue... Maybe use #1361 for this?

gbraad: A simple note in the driver plugin topic might do for the moment... ?

budhrg: > Maybe use #1361 for this?

Sure

gbraad: #1373

gbraad:
Will merge when #1373 is also approved

cmd/minishift/cmd/start_preflight.go
LalatenduMohanty: S/"See the 'Setting Up the Driver Plug-in' topic for more information"/ Refer the documentation for ''Setting Up the Driver Plug-in' topic for more information/

LalatenduMohanty: Or `Refer ''Setting Up the Driver Plug-in' topic in documentation for more information.`

LalatenduMohanty: I still think the message is not descriptive enough. But I  can not come up with anything better. So we are fine at the moment.

Issue #1351, opened 2017-09-06
minishift doesn't start with an alternate public-hostname/routing-suffix


chmouel:
A new issue opened following comment on this bug https://github.com/minishift/minishift/issues/1011#issuecomment-327459567

When trying to use the switch : 

    minishift start --public-hostname localhost --routing-suffix 127.0.0.1.nip.io --vm-driver=kvm --memory=6144 --cpus=4 --disk-size=50g

minishift will error at the end with : 

    Error during post cluster up configuration: Unable to add sudoer role

It seems that the command line switch cause the issue : 

    --public-hostname localhost --routing-suffix 127.0.0.1.nip.io

This used to work but I don't really know until which version

praveenkumar:
@chmouel what is the use case of using `localhost` as public-hostname for minishift? We might need to put an documentation around it.

chmouel:
I get a fairly limited laptop so I deploy minishift on another laptop and use 

```
ssh -L 8443:localhost:8443  remotelaptop
```

to forward the connection. 

And connect with https://localhost:8443 from my locallaptop

praveenkumar:
> I get a fairly limited laptop so I deploy minishift on another laptop and use

@chmouel Ack, I will try to get a patch in for it. Thanks for clarification.

praveenkumar:
@csrwng I need some insight about this. Below is explanation about what we do currently, let me know if we are missing something here.

1. We try to extract configuration from the `~/.kube/config` file and make it part of local to instance.

 - https://github.com/minishift/minishift/blob/master/pkg/minishift/clusterup/clusterup.go#L115
  - https://github.com/minishift/minishift/blob/master/pkg/minikube/kubeconfig/config.go

Generated file not works when we use public-hostname with minishift but it works without it.

csrwng:
The code in cluster up that appends to ~/.kube/config (and logs in) is here:
https://github.com/openshift/origin/blob/master/pkg/oc/bootstrap/docker/openshift/login.go

Does it not work with any public hostname or just 'localhost' ?

praveenkumar:
@chmouel  Can you please try out http://artifacts.ci.centos.org/minishift/minishift/pr/1372/ binary and see if your issue #1351 resolved?

chmouel:
This works perfecty thanks : 

after downloading it, i ran it with the public-hostname=localhost : 

[...]

PR #1372, opened 2017-09-13
Issue #1351 Start fails when use public-hostname

praveenkumar:


pkg/minishift/clusterup/clusterup.go
```diff
@@ -275,7 +276,10 @@ func applyAddOns(addOnManager *manager.AddOnManager, ip string, routingSuffix st
 	return nil
 }
 
-func getConfigClusterName(ip string, port int) string {
+func getConfigClusterName(hostname string, ip string, port int) string {
+	if hostname != "" {
+		return "local-cluster"
+	}
 	return fmt.Sprintf("%s:%d", strings.Replace(ip, ".", "-", -1), port)
```
gbraad: but if set, then what?

praveenkumar: @gbraad if set then it will return "local-cluter" because that something part of cluster up code : https://github.com/openshift/origin/blob/master/pkg/oc/bootstrap/docker/openshift/login.go#L56

LalatenduMohanty: @praveenkumar We should have added a comment that `cluster up` 's context it is local-cluster and we depend on `cluster up` for the kubeconfig.

Issue #1374, opened 2017-09-13
Progress dot should have timeout and slow down over period of time.


praveenkumar:
Currently we don't have any timeout for progress dot and also when a operation gets longer than it should slow down over period of time. 

```
Disk size: 20 GB
-- Starting Minishift VM ..........................................................
............................................................................................
............................................................................................
............................................................................................
............................................................................................
```

gbraad:
if `slowDown`, we use a `easeOffFactor` or `slowDownFactor`? Which does it based on a modulo or similar, so if 10 doys have been written we add an additional second to the interval.

gbraad:
But I believe the biggest issue is that we can not ctrl+c out of the process. I am pretty sure it responded to a break before...

praveenkumar:
@gbraad `ctrl+c` works in Linux need to check in windows.

```
$ ./minishift start 
-- Checking if KVM driver is installed ... 
   Driver is available at /usr/sbin/docker-machine-driver-kvm ... 
   Checking driver binary is executable ... OK
-- Starting local OpenShift cluster using 'kvm' hypervisor ...
-- Minishift VM will be configured with ...
   Memory:    4 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM .....^C
[prkumar@dhcp193-13 bin]$ 
```

budhrg:
@gbraad , `ctrl+c` works in Windows too for me
```
PS C:\gowork\src\github.com\minishift\minishift> minishift start --iso-url file://C:/ms_cache/iso/1.1.0/minishift-b2d.iso
-- Checking if Hyper-V driver is configured ... OK
-- Starting local OpenShift cluster using 'hyperv' hypervisor ...
-- Minishift VM will be configured with ...
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ................
```

gbraad:
```
PS C:\work\src\github.com\minishift\minishift\out\windows-amd64> .\minishift.exe start
-- Checking if Hyper-V driver is configured ... OK
-- Starting local OpenShift cluster using 'hyperv' hypervisor ...
-- Starting Minishift VM ...........................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
....................................................................................................
[...]

PR #1383, opened 2017-09-14
Issue #1374 Append extra second to progress dot

praveenkumar:
- Replaced bool with type struct{} signature which require no memory allocation ( https://dave.cheney.net/2014/03/25/the-empty-struct )
    - Close the channel when Stop

### DIFF
--- a/pkg/util/progressdots/progressdots.go
+++ b/pkg/util/progressdots/progressdots.go
@@ -21,18 +21,30 @@ import (
 	"time"
 )
 
-const DefaultInterval = 2 * time.Second
+const (
+	DefaultInterval = 2 * time.Second
+	DefaultEasing   = 10
+)
 
 type ProgressDots struct {
-	interval time.Duration
-	handler  chan bool
+	interval   time.Duration
+	easing     int
+	dotCounter int
+	handler    chan bool
 }
 
-// NewProgressDots creates the channel to handle progress dots
-func New() *ProgressDots {
+// New creates the channel to handle progress dots
+// It takes easingOptional as optional parameter to control dots interval
+func New(easingOptional ...int) *ProgressDots {
+	easing := DefaultEasing
+	if len(easingOptional) > 0 {
+		easing = easingOptional[0]
+	}
 	return &ProgressDots{
-		interval: DefaultInterval,
-		handler:  make(chan bool),
+		interval:   DefaultInterval,
+		easing:     easing,
+		dotCounter: 0,
+		handler:    make(chan bool),
 	}
 }
 
@@ -45,7 +57,12 @@ func (s *ProgressDots) Start() {
 				return
 			default:
 				fmt.Print(".")
+				s.dotCounter++
 				time.Sleep(s.interval)
+				if s.easing != 0 && s.dotCounter%s.easing == 0 {
+					s.dotCounter = 0
+					s.interval = s.interval + 1*time.Second
+				}
 			}
 		}
 	}()
@@ -53,6 +70,7 @@ func (s *ProgressDots) Start() {
 
 // Stop stops the dots
 func (s *ProgressDots) Stop() {
+	defer close(s.handler)
 	s.handler <- true
 }
 


pkg/util/progressdots/progressdots.go
```diff
@@ -25,14 +25,14 @@ const DefaultInterval = 2 * time.Second
 
 type ProgressDots struct {
 	interval time.Duration
-	handler  chan bool
+	handler  chan struct{}
```
budhrg: Any benefit in making it struct ?

praveenkumar: @budhrg https://dave.cheney.net/2014/03/25/the-empty-struct

pkg/util/progressdots/progressdots.go
```diff
@@ -46,14 +46,16 @@ func (s *ProgressDots) Start() {
 			default:
 				fmt.Print(".")
 				time.Sleep(s.interval)
+				s.interval = s.interval + 1*time.Second
```
gbraad: Doesn't this slowdown too quickly?

I would make this optional and over a modulo of time...
This is more important for starting with HyperV than the registration.

budhrg: Yes, with every second it is increasing ...

I think the idea was after some dots(say 10), it should increase per second

pkg/util/progressdots/progressdots.go
gbraad: I see no real benefit. It is not like we consume hundreds of booleans, for different channels. It is just a bool.

This obscures code...

pkg/util/progressdots/progressdots.go
gbraad: A certain amount of dots if set... 10 is an arbitrary number.

Else the meaning of interval is obsolete too. In this case it is a startInterval.

LalatenduMohanty:
@praveenkumar I am not sure how your PR fixes https://github.com/minishift/minishift/issues/1374#issuecomment-329125849 ?

pkg/util/progressdots/progressdots.go
```diff
@@ -45,15 +54,20 @@ func (s *ProgressDots) Start() {
 				return
 			default:
 				fmt.Print(".")
+				s.dotCounter++
 				time.Sleep(s.interval)
+				if s.dotCounter%10 == 0 {
```
gbraad: Not sure yet how this works. In my opinion the terms are not easy to understand, so the code does not describe itself well. In electronics and motion we have something called 'easing' which can slow down a motion or value according to a curve. Although we are not describing not a real curve here, the effect is similar.

To me, `dotCounter` only has an internal respresentation to keep a record of the dots I have written since the last easing. 

```go
  	return &ProgressDots{
 		interval:   DefaultInterval,
  		easing:     DefaultEasing,
 		handler:    make(chan struct{}),
  	}

...

s.dotCounter++
if easing != 0 && s.dotCounter % easing == 0 {
   s.dotCounter = 0
   s.interval = s.interval + 1 * time.Second
}
```

pkg/util/progressdots/progressdots.go
```diff
@@ -21,18 +21,27 @@ import (
...
-	interval time.Duration
-	handler  chan bool
+	interval   time.Duration
+	dotCounter int
+	handler    chan struct{}
 }
 
 // NewProgressDots creates the channel to handle progress dots
-func New() *ProgressDots {
+func New(dotCounterOptional ...int) *ProgressDots {
```
gbraad: Without documentation, I am not sure how this should be used...

pkg/util/progressdots/progressdots.go
```diff
@@ -21,18 +21,27 @@ import (
...
+	dotCounter := DefaultDotCounter
+	if len(dotCounterOptional) > 0 {
+		dotCounter = dotCounterOptional[0]
+	}
 	return &ProgressDots{
-		interval: DefaultInterval,
-		handler:  make(chan bool),
+		interval:   DefaultInterval,
+		dotCounter: dotCounter,
+		handler:    make(chan struct{}),
```
gbraad: `handler: make(chan bool);` makes ProgressDots easier to re-use with a function `SetHandler` to allow for syncing with an external handler

```go
func main() {
    done := make(chan bool)
    
    p := progressDots.New()
    p.SetHandler(done)
    p.Start()
    
    <- done
```

pkg/util/progressdots/progressdots.go
```diff
@@ -45,15 +54,20 @@ func (s *ProgressDots) Start() {
 				return
 			default:
 				fmt.Print(".")
+				s.dotCounter++
 				time.Sleep(s.interval)
+				if s.dotCounter%10 == 0 {
+					s.interval = s.interval + 1*time.Second
```
budhrg: This means after 10 seconds, we are doubling the frequency of dots. so even after 30 seconds we have 2 dots in one seconds.

pkg/util/progressdots/progressdots.go
praveenkumar: @gbraad Currently we don't expose a sethandler functionality for progressDots but in future we require then I do agree on this change.

pkg/util/progressdots/progressdots.go
praveenkumar: @gbraad Somehow I completely missed to need a different variable to store the variable :(

Issue #1322, opened 2017-08-29
Add integration test for add-on command


agajdosi:
We should test addons command in more depth, we are missing some subcommands and edge cases. Also we should focus on failures, if minishift is able handle or recover from installation of totally wrong addon - for example just a random text file or broken addon file etc.

Also we should keep an eye on #1321 and test for `minishift addons uninstall` which is incoming into master.

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/pull/1385/

Issue #1377, opened 2017-09-13
Command addons remove tries to run when Minishift is not running


agajdosi:
Command `minishift addons remove` should work as other commands which need Minishift to be in `Running state`, expected:

```gherkin
  Scenario: Removing addon when Minishift is not running
    Given Minishift has state "Does Not Exist"
     When executing "minishift addons remove xpaas" succeeds
     Then stdout should contain
      """
      Running this command requires an existing 'minishift' VM, but no VM is defined.
      """
```
But instead it gives:
```
No xpaas.addon.remove file found for xpaas add-on.
```

gbraad:
Isn't this related to #1378

agajdosi:
@gbraad Good point, is related. However this is about fact that Minishift shoukd first check if the command can be run in current `state`. If it can't be run, then it will tell me. Now for this command it first checks if it can uninstall the addon. The #1378 is a story when state is running but .remove was not installed so we still can not uninstall.

gbraad:
Is this still reproducible?

agajdosi:
Thanks for noticing, @gbraad. Fortunately it is no longer reproducible! I am adding integration-test tag and will take this issue as task to implement needed changes into integration test suite. There are some not failing but also not truly correct lines around https://github.com/minishift/minishift/blob/master/test/integration/features/cmd-addons.feature#L40.

Issue #1375, opened 2017-09-13
Improve isolation of features by using before- and afterFeature() hooks


agajdosi:
The beforeFeature() and afterFeature() hooks have been added into Godog recently. We could update to newest Godog and improve the isolation of individual features. Right now we are doing cleaning and checks just before the whole suite, but we should do it on individual features to make sure there is nothing which leaks into following features - for example when we manipulate the config file.

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/pull/1385/

Issue #1337, opened 2017-09-01
Document VT-x/AMD-v requirement


budhrg:
With issues like this https://github.com/minishift/minishift/issues/1335, I feel we need to document enabling of `VT-x/AMD-v` prior to starting minishift. 

Even minikube has documented in its [requirement list](https://github.com/kubernetes/minikube#requirements).

I feel this is required for users who are trying first time any virtualization stuff.

gbraad:
For some reason, it got rewritten and lost. I agree this needs to be documented again.

chrisfalter:
You should be aware that when Hyper-V is enabled, it prevents any other hypervisor (such as VirtualBox) from accessing VT-x., per [this Technet discussion thread](https://social.technet.microsoft.com/Forums/windows/en-US/118561b9-7155-46e3-a874-6a38b35c67fd/hyperv-disables-vtx-for-other-hypervisors?forum=w8itprogeneral).

Some users may wish to install both Hyper-V and Virtual-Box, enabling only one at a time. Such users evidently can only use minishift with Virtualbox, as the presence of an installed (and disabled) Virtualbox means that minishift will try to use it, even when the user passes in the command-line argument "vm-driver=hyperv". That was certainly my experience, as mentioned in #1335

LalatenduMohanty:
@budhrg agree lets add it to the documentation.

gbraad:
@chrisfalter And this is expected. When VTx is disabled, it means that VirtualBox does not run optimally, as it can not use the Virtualization Extensions. Hyper-V requires VTx: https://technet.microsoft.com/en-us/library/cc731898(v=ws.11).aspx, https://technet.microsoft.com/en-us/library/hh831531(v=ws.11).aspx

You can NOT install both the same time. This is a known limitation: https://en.wikipedia.org/wiki/Hyper-V#VT-x_handling I was unable to find the related TechNet articles, but if Wikipedia states it, it must be true ;-)

The behaviour as you mentioned is due to libmachine. The same behaviour exists on macOS with xhyve, which can not work when VirtualBox is installed alongside the requested 'other' hypervisor. Therefore libmachine will not even consider the passed vm-driver.

gbraad:
@robin-owen old, but very relevant issue

gbraad:
BTW, support for VT-x can be tested for with `msr-tools`, but that would be an entry in the troubleshooting section.

PR #1385, opened 2017-09-14
Issue #1322 - Add int. test for add-on command

agajdosi:


budhrg:
retest this please

agajdosi:
Ok, so from logs it is clear that non-default state is leaking from feature to feature. In this case it is addons configuration. The addons.feature as it is now is fine, however when it is run after basic.feature, it has problems due to the leakage. I will implement #1375 as part of this PR to prevent this behaviour.

test/integration/features/cmd-addons.feature
```diff
@@ -0,0 +1,158 @@
...
+  Scenario: Applying disabled add-on which was not applied during the startup
+    Given Minishift has state "Running"
+     Then executing "minishift addons apply anyuid" succeeds
+      And stdout should contain
+      """
+      Add-on 'anyuid' changed the default security context constraints to allow pods to run as any user.
+      """
+
+  Scenario: Removing applied addon
+  To be done.
```
gbraad: OOh...

test/integration/integration_test.go
```diff
@@ -753,3 +776,33 @@ func catDockerConfigFile() error {
 
 	return err
 }
+
+func downloadFileIntoLocation(downloadURL string, destinationFolder string) error {
```
gbraad: 👍

test/integration/util/util.go
```diff
@@ -98,8 +98,8 @@ func (m *MinishiftRunner) CDKSetup() {
 }
 
 func (m *MinishiftRunner) IsCDK() bool {
-	cmdOut, _, _ := m.RunCommand("-h")
-	return strings.Contains(cmdOut, "setup-cdk")
+	cmdOut, _, _ := m.RunCommand("setup-cdk -h")
```
budhrg: :+1:

test/integration/features/cmd-addons.feature
budhrg: @agajdosi ,  still in progress?

glide.yaml
```diff
@@ -59,7 +59,7 @@ import:
 - package: github.com/asaskevich/govalidator/
   version: 5
 - package: github.com/DATA-DOG/godog
-  version: v0.6.2
+  version: v0.7.4
```
LalatenduMohanty: Any specific reason we are going to the newer version? Or it is just to update the version of godog.

test/integration/features/cmd-addons.feature
```diff
@@ -0,0 +1,158 @@
...
+     When executing "minishift addons uninstall anyuid" succeeds
+     Then executing "minishift addons list" succeeds
+      And stdout should not contain "anyuid"
+
+  Scenario: Installing add-on from a folder
+  Note: working directory when executing Minishift commands is: /test/integration.
+     When file from "https://raw.githubusercontent.com/minishift/minishift/master/addons/anyuid/anyuid.addon" is downloaded into location "download/anyuid"
+      And executing "minishift addons install ../../out/integration-test/download/anyuid" succeeds
+     Then executing "minishift addons list" succeeds
+      And stdout should contain "anyuid"
```
LalatenduMohanty: 👍

test/integration/features/cmd-addons.feature
```diff
@@ -0,0 +1,158 @@
...
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift addons apply anyuid" succeeds
+     Then stdout should contain
+      """
+      Running this command requires an existing 'minishift' VM, but no VM is defined.
+      """
+
+  Scenario: Removing add-on when Minishift is not running
+  After issue no. 1337 is resolved, please change the expected stdout to:
+  Minishift should return: "Running this command requires an existing 'minishift' VM, but no VM is defined."
```
LalatenduMohanty: You mean https://github.com/minishift/minishift/issues/1337? I guess you are talking about a different issue.

test/integration/features/cmd-addons.feature
```diff
@@ -0,0 +1,158 @@
...
+      """
+     When executing "minishift addons list" succeeds
+     Then stdout should match "xpaas\s*: disabled"
+
+  Scenario: Disabling disabled add-on
+     When executing "minishift addons disable xpaas" succeeds
+     Then stdout should contain
+      """
+      Add-on 'xpaas' disabled
+      """
```
LalatenduMohanty: Nitpick : I feel the output from minishift should be different e.g. `Add-on 'xpaas' is already in disabled state`

test/integration/features/cmd-addons.feature
agajdosi: @gbraad @budhrg Oh, I have missed this, will fix it asap, sorry for that.

glide.yaml
agajdosi: @LalatenduMohanty Yes, in 0.7.4 there were beforeFeature() and afterFeature() hooks introduced, which I use to secure more isolation between individual features - until now if change in config for example was made in basic it went through all other features after the basic feature. Issue of mine for the hooks: https://github.com/DATA-DOG/godog/issues/97

test/integration/features/cmd-addons.feature
agajdosi: @LalatenduMohanty Should have been #1377. Going to change it.

test/integration/features/cmd-addons.feature
agajdosi: @LalatenduMohanty +1 That would be nicer, seems like easy-fix, will create an issue for that.

agajdosi:
Before merge please remove the change on L:308 in centos_ci.sh - this change was just for testing proposed feature in Centos-CI.

agajdosi:
retest this please

test/integration/integration_test.go
```diff
@@ -301,6 +317,13 @@ func ensureTestDirEmpty() {
 	}
 }
 
+func cleanTestDirConfiguration() {
+	configPath := filepath.Join(testDir, "config")
+	addonsPath := filepath.Join(testDir, "addons")
+	os.RemoveAll(configPath)
+	os.RemoveAll(addonsPath)
```
LalatenduMohanty: @agajdosi We should check if `os.RemoveAll` returns any error. Otherwise it will fail saliently and it might effect the tests.

centos_ci.sh
```diff
@@ -305,7 +305,7 @@ function build_and_test() {
   cd $GOPATH/src/github.com/minishift/minishift
   make prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags=basic,openshift,version,cmd-config -format pretty"
+  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags basic,cmd-addons -format pretty"
```
LalatenduMohanty: @agajdosi I think you should remove this temporary change and update the PR as it will allow it run the complete tests before we merge it.

agajdosi: @LalatenduMohanty ack

test/integration/integration_test.go
agajdosi: @LalatenduMohanty Good point, will add the check.

agajdosi:
@LalatenduMohanty Requested changes implemented.

test/integration/features/cmd-addons.feature
gbraad: It would be an additional check that does not really solve a problem.

  * If set, we do not take action
  * we just do as requested

Either way, we do not break the user interface design rule of "providing the user feedback". We perform the expected action. Indicating it was already in that state, does not change the outcome.

PR #1391, opened 2017-09-15
Issue #1316 Set IP address (Hyper-V only)

### LINKED ISSUE #1316: Assign Static IP address for Hyper-V
See #1293 for background information
PR #1391 

#### Todo

  * [x] Modify hvkvp to read a specified key ,eg. `hvkvp -k ipaddress` (see https://github.com/gbraad/go-hvkvp/issues/2)
    * [x] setup proper build environment https://github.com/gbraad/go-hvkvp/issues/3
  * [x] Modify ISO to use `hvkvp` to set IP address
    * [x] CentOS https://github.com/minishift/minishift-centos-iso/issues/137
      * [x] add RPM https://github.com/gbraad/go-hvkvp/issues/6
      * [x] network settings configuration file
        * [x] disable NetworkManager, start when needed
        * [x] assigns IP using `ip a add`
    * [x] ~~B2D https://github.com/minishift/minishift-b2d-iso/issues/59~~  
           **blocker**: pool gets cleared on start https://github.com/boot2docker/boot2docker/issues/1274  
             workaround does not provide a good enough experience (needs to disable IPv6)
      * [x] add TCZ extension https://github.com/gbraad/go-hvkvp/issues/8
      * [x] automount starts before network (add to `dhcp.sh`?)
  * [x] Refactor feature-toggles (see #1288)
    * [x] ~~Add `EnableHyperV` option~~
    * [x] Implement experimental flag `--ip-address` for HyperV-only
      * [x] `if EnableHyperV && EnableExperimental`
  * [x] Create a minimal PowerShell wrapper #1367
    * [x] Allow creating a KVP using the WMIObject https://github.com/gbraad/go-hvkvp/issues/4  
           See: https://github.com/gbraad/go-cmdbasher/blob/master/cmd/basher/example.go
    * [x] allow settings multiple KeyValue-pairs, eg `IpAddress`, `Gateway`, `NameServer`, etc
       * [x] Network script (filled as template)
  * [ ] find a hosting location for the binaries as used by Minishift for the ISOs
       * [ ] copr? https://github.com/gbraad/go-hvkvp/issues/9
  * [x] documentation

##### Needed updates for IP assignment script

* [x] Allow reading config from file on disk (this solves static assignment for other hypervisors [no change on reboot])
* [x] check assigned IpAddress for netmask bits
  * if none specified, + "/24"
* [x] use Gateway KVP
  * else assign gateway based on IpAddress
* [x] use DnsServer KVP
  * if none specified, use "8.8.8.8"
  * check connectivity of DNS
  * allow specifying multiple entries

#### Related pull requests

  * [ ] ~~https://github.com/minishift/minishift-centos-iso/pull/138~~  
        ~~https://github.com/minishift/minishift-centos-iso/pull/154~~
        https://github.com/minishift/minishift-centos-iso/pull/164
  * [x] ~~https://github.com/minishift/minishift-b2d-iso/pull/61~~

#### Ideas and possible improvements

  * #1457 assign fixed IP address for other hypervisors
  * #1473 only send hvkvp values when VM is in state "Running"
  * #1492 automagically assign default gateway
  * #1493 automagically assign nameservers
  * #1496 feature detection during provisioning to determine if ISO supports requested feature
  * #1497 assign multiple nameservers
  * #1499 add support to the minikube ISO
  * #1500 support IPv6 possible?
  * #1503 improve check if HyperV is used

gbraad:
Fixes #1316

How to test:
  * CentOS ISO needed: http://artifacts.ci.centos.org/minishift/minishift-centos-iso/pr/164/

```powershell
PS> $env:MINISHIFT_ENABLE_EXPERIMENTAL="y"
PS> $env:HYPERV_VIRTUAL_SWITCH="Wireless"
PS> minishift.exe start --network-ipaddress 10.0.21.40 --network-gateway 10.0.21.1
```

```powershell
PS> New-NetNat -Name SharedNAT -InternalIPInterfaceAddressPrefix 10.0.75.1/24
PS> $env:MINISHIFT_ENABLE_EXPERIMENTAL="y"
PS> $env:HYPERV_VIRTUAL_SWITCH="DockerNAT"
PS> minishift.exe start --network-ipaddress 10.0.75.124 --network-gateway 10.0.75.1
```

Note 1: the `New-NetNat` preparation step is necessary when using DockerNAT.
Note 2: `--network-gateway` is needed for outside communication, else SSH will fail to respond to connection requests, and downloads would never start. A `--network-nameserver` is needed to resolve locations to pull containers, imagestreams, etc. (currently defaults to `8.8.8.8`)

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -120,6 +120,13 @@ var (
 	// Preflight values
 	CheckNetworkHttpHost = createConfigSetting("check-network-http-host", SetString, nil, nil, true)
 	CheckNetworkPingHost = createConfigSetting("check-network-ping-host", SetString, nil, nil, true)
+
+	// Network settings (Hyper-V only)
+	NetworkDevice = createConfigSetting("network-device", SetString, nil, nil, true)
+	IPAddress     = createConfigSetting("network-ipaddress", SetString, []setFn{validations.IsValidIPv4Address}, nil, true)
+	Netmask       = createConfigSetting("network-netmask", SetString, []setFn{validations.IsValidNetmask}, nil, true)
+	Gateway       = createConfigSetting("network-gateway", SetString, []setFn{validations.IsValidIPv4Address}, nil, true)
+	Nameserver    = createConfigSetting("network-nameserver", SetString, []setFn{validations.IsValidIPv4Address}, nil, true)
 )
 
 func createConfigSetting(name string, set func(MinishiftConfig, string, string) error, validations []setFn, callbacks []setFn, isApply bool) *Setting {
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -336,7 +336,7 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 	}
 
 	// Experimental features
-	if hasEnabledExperimental {
+	if minishiftConfig.EnableExperimental {
 		networkSettings := minishiftNetwork.NetworkSettings{
 			Device:    viper.GetString(configCmd.NetworkDevice.Name),
 			IPAddress: viper.GetString(configCmd.IPAddress.Name),
@@ -505,7 +505,7 @@ func initStartFlags() *flag.FlagSet {
 	startFlagSet.AddFlag(registryMirrorFlag)
 	startFlagSet.AddFlag(cmdUtil.AddOnEnvFlag)
 
-	if hasEnabledExperimental && runtime.GOOS == "windows" {
+	if minishiftConfig.EnableExperimental && runtime.GOOS == "windows" {
 		startFlagSet.String(configCmd.NetworkDevice.Name, "eth0", "Specify the network device to use for the IP address. Ignored if no IP address specified (experimental - Hyper-V only)")
 		startFlagSet.String(configCmd.IPAddress.Name, "", "Specify IP address to assign to the instance (experimental - Hyper-V only)")
 		startFlagSet.String(configCmd.Netmask.Name, "24", "Specify netmask to use for the IP address. Ignored if no IP address specified (experimental - Hyper-V only)")
@@ -537,7 +537,7 @@ func initClusterUpFlags() *flag.FlagSet {
 	clusterUpFlagSet.AddFlag(cmdUtil.HttpProxyFlag)
 	clusterUpFlagSet.AddFlag(cmdUtil.HttpsProxyFlag)
 
-	if hasEnabledExperimental {
+	if minishiftConfig.EnableExperimental {
 		clusterUpFlagSet.Bool(configCmd.ServiceCatalog.Name, false, "Install service catalog (experimental)")
 		clusterUpFlagSet.String(configCmd.ExtraClusterUpFlags.Name, "", "Specify optional flags for use with 'cluster up' (unsupported)")
 	}
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -19,6 +19,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"runtime"
 	"strings"
 
 	"github.com/asaskevich/govalidator"
@@ -41,6 +42,7 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/docker"
 	"github.com/minishift/minishift/pkg/minishift/docker/image"
 	"github.com/minishift/minishift/pkg/minishift/hostfolder"
+	minishiftNetwork "github.com/minishift/minishift/pkg/minishift/network"
 	"github.com/minishift/minishift/pkg/minishift/openshift"
 	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
@@ -305,6 +307,7 @@ func determineInsecureRegistry(key string) []string {
 
 func startHost(libMachineClient *libmachine.Client) *host.Host {
 	progressDots := progressdots.New()
+
 	// Configuration used for creation/setup of the Virtual Machine
 	machineConfig := &cluster.MachineConfig{
 		MinikubeISO:      determineIsoUrl(viper.GetString(configCmd.ISOUrl.Name)),
@@ -330,6 +333,22 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 		fmt.Println("   Memory:   ", units.HumanSize(float64((machineConfig.Memory/units.KiB)*units.GB)))
 		fmt.Println("   vCPUs :   ", machineConfig.CPUs)
 		fmt.Println("   Disk size:", units.HumanSize(float64(machineConfig.DiskSize*units.MB)))
+	}
+
+	// Experimental features
+	if hasEnabledExperimental {
+		networkSettings := minishiftNetwork.NetworkSettings{
+			Device:    viper.GetString(configCmd.NetworkDevice.Name),
+			IPAddress: viper.GetString(configCmd.IPAddress.Name),
+			Netmask:   viper.GetString(configCmd.Netmask.Name),
+			Gateway:   viper.GetString(configCmd.Gateway.Name),
+			DNS1:      viper.GetString(configCmd.Nameserver.Name),
+		}
+
+		// Configure networking on startup only works on Hyper-V
+		if networkSettings.IPAddress != "" {
+			minishiftNetwork.ConfigureNetworking(constants.MachineName, networkSettings)
+		}
 	}
 
 	cacheMinishiftISO(machineConfig)
@@ -485,6 +504,14 @@ func initStartFlags() *flag.FlagSet {
 	startFlagSet.AddFlag(insecureRegistryFlag)
 	startFlagSet.AddFlag(registryMirrorFlag)
 	startFlagSet.AddFlag(cmdUtil.AddOnEnvFlag)
+
+	if hasEnabledExperimental && runtime.GOOS == "windows" {
+		startFlagSet.String(configCmd.NetworkDevice.Name, "eth0", "Specify the network device to use for the IP address. Ignored if no IP address specified (experimental - Hyper-V only)")
+		startFlagSet.String(configCmd.IPAddress.Name, "", "Specify IP address to assign to the instance (experimental - Hyper-V only)")
+		startFlagSet.String(configCmd.Netmask.Name, "24", "Specify netmask to use for the IP address. Ignored if no IP address specified (experimental - Hyper-V only)")
+		startFlagSet.String(configCmd.Gateway.Name, "", "Specify gateway to use for the instance. Ignored if no IP address specified (experimental - Hyper-V only)")
+		startFlagSet.String(configCmd.Nameserver.Name, "8.8.8.8", "Specify nameserver to use for the instance. Ignored if no IP address specified (experimental - Hyper-V only)")
+	}
 
 	return startFlagSet
 }
--- a/docs/source/using/managing-minishift.adoc
+++ b/docs/source/using/managing-minishift.adoc
@@ -322,3 +322,59 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
 
 `extra-clusterup-flags`::
 Enables passing flags that are not directly exposed in the Minishift CLI directly to `oc cluster up`.
+
+
+[[hyperv-static-ip]]
+=== Assign IP address to Hyper-V (experimental)
+
+Since the Internal Virtual Switch for Hyper-V does not provide a DHCP offer option, an IP address needs to be provided in a different way.
+For Hyper-V a functionality is provided to assign an IP address on startup using the Data Exchange Service.
+
+[IMPORTANT]
+====
+-  While the default image is B2D, this only works with the CentOS/RHEL based image in combination with Hyper-V. The B2D image experiences
+   a problem when the values are being send to the Minishift instance and consumed by the B2D iso. We are looking into the issue and hope
+   to provide a solution in the coming future.
+====
+
+To make this work you need to create a Virtual Switch using NAT
+
+- link:https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/setup-nat-network[Set up a NAT network]
+
+
+[NOTE]
+====
+WinNAT is limited to one NAT network per host. For more details about capabilities, and limitations, please see the link:https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/setup-nat-network[WinNAT capabilities and limitations blog]
+====
+
+The following command will attempt to assign an IP address for use on the Internal Virtual Switch 'MyInternal'.
+
+----
+PS> $env:MINISHIFT_ENABLE_EXPERIMENTAL="y"
+PS> $env:HYPERV_VIRTUAL_SWITCH="MyInternal"
+PS> minishift.exe start `
+  --iso-url centos `
+  --network-ipaddress 192.168.1.10 `
+  --network-gateway 192.168.1.1 `
+  --network-nameserver 8.8.8.8
+----
+
+
+If you want to use the 'DockerNAT' network, the following commands are needed to setup the correct NAT networking and assigning an IP in the range expected.
+
+----
+PS> New-NetNat -Name SharedNAT -InternalIPInterfaceAddressPrefix 10.0.75.1/24
+PS> $env:MINISHIFT_ENABLE_EXPERIMENTAL="y"
+PS> $env:HYPERV_VIRTUAL_SWITCH="DockerNAT"
+PS> minishift.exe start `
+  --iso-url centos `
+  --network-ipaddress 10.0.75.128 `
+  --network-gateway 10.0.75.1 `
+  --network-nameserver 8.8.8.8
+----
+
+[NOTE]
+====
+- Be sure to specify a valid gateway and nameserver. Failing to do so will result in connectivity issuesi
+====
+
--- a/glide.lock
+++ b/glide.lock
@@ -104,6 +104,8 @@ imports:
   version: bbd0c6e271208dce66d8fda4bc536453cd27fc4a
 - name: github.com/juju/errors
   version: c7d06af17c68cd34c835053720b21f6549d9b0ee
+- name: github.com/gbraad/go-hvkvp
+  version: a692e030379b6de28a10b7697decdf8109899f6e
 - name: github.com/kardianos/osext
   version: ae77be60afb1dcacde03767a8c37337fad28ac14
 - name: github.com/magiconair/properties
--- a/glide.yaml
+++ b/glide.yaml
@@ -65,6 +65,8 @@ import:
   version: 3bc7a60b1df80a5766820c4a53d59668f4553f75
 - package: github.com/juju/errors
   version: c7d06af17c68cd34c835053720b21f6549d9b0ee
+- package: github.com/gbraad/go-hvkvp
+  version: 0.3
 # The following repositories need to be specified explicitly, since the versions referenced by the openshift/origin dependency
 # refer to forks of these projects within the openshift organizaton
 # This needs to be manually updated after upgrading the openshift/origin dependnecy
--- a/pkg/minishift/clusterup/clusterup.go
+++ b/pkg/minishift/clusterup/clusterup.go
@@ -35,6 +35,7 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/kubeconfig"
 	"github.com/minishift/minishift/pkg/minishift/addon/command"
 	"github.com/minishift/minishift/pkg/minishift/addon/manager"
+	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
 	"github.com/minishift/minishift/pkg/minishift/oc"
 
 	"regexp"
@@ -90,13 +91,15 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
 		cmdArgs = append(cmdArgs, value)
 	}
 
-	// Deal with extra flags (add to command arguments)
-	if len(extraFlags) > 0 {
-		fmt.Println("-- Extra `oc` cluster up flags (experimental) ... ")
-		fmt.Printf("   '%s'\n", extraFlags)
-		extraFlagFields := strings.Fields(extraFlags)
-		for _, extraFlag := range extraFlagFields {
-			cmdArgs = append(cmdArgs, extraFlag)
+	if minishiftConfig.EnableExperimental {
+		// Deal with extra flags (add to command arguments)
+		if len(extraFlags) > 0 {
+			fmt.Println("-- Extra `oc` cluster up flags (experimental) ... ")
+			fmt.Printf("   '%s'\n", extraFlags)
+			extraFlagFields := strings.Fields(extraFlags)
+			for _, extraFlag := range extraFlagFields {
+				cmdArgs = append(cmdArgs, extraFlag)
+			}
 		}
 	}
 
--- a/pkg/minishift/config/validations.go
+++ b/pkg/minishift/config/validations.go
@@ -121,3 +121,36 @@ func IsValidUrl(_ string, isoURL string) error {
 	}
 	return nil
 }
+
+func IsValidIPv4Address(name string, address string) error {
+	ip := net.ParseIP(address).To4()
+	if ip == nil {
+		return fmt.Errorf("%s IPv4 address is not valid: %s", name, address)
+	}
+
+	return nil
+}
+
+func IsValidNetmask(name string, mask string) error {
+	err := fmt.Errorf("%s netmask is not valid: %s", name, mask)
+
+	if stringUtils.HasOnlyNumbers(mask) {
+		prefixSize, _ := strconv.Atoi(mask)
+		if prefixSize == 0 || prefixSize > 25 {
+			return err
+		}
+		return nil
+	}
+
+	// got something like '255.255.255.0' instead
+	ip := net.ParseIP(mask).To4()
+	if ip == nil {
+		return err
+	}
+	prefixSize, _ := net.IPMask(ip).Size()
+	if prefixSize == 0 {
+		return err
+	}
+
+	return nil
+}
--- a/pkg/minishift/config/validations_test.go
+++ b/pkg/minishift/config/validations_test.go
@@ -188,3 +188,81 @@ func TestValidProxyURL(t *testing.T) {
 	}
 	runValidations(t, tests, "iso-url", IsValidProxy)
 }
+
+func TestValidIPv4Address(t *testing.T) {
+
+	var tests = []validationTest{
+		{
+			value:     "",
+			shouldErr: true,
+		},
+		{
+			value:     "10.0.75.128",
+			shouldErr: false,
+		},
+		{
+			value:     "localhost",
+			shouldErr: true,
+		},
+		{
+			value:     "::1",
+			shouldErr: true,
+		},
+		{
+			value:     "fe80::200:5aee:feaa:20a2",
+			shouldErr: true,
+		},
+	}
+	runValidations(t, tests, "ipaddress", IsValidIPv4Address)
+}
+
+func TestValidNetmask(t *testing.T) {
+
+	var tests = []validationTest{
+		{
+			value:     "",
+			shouldErr: true,
+		},
+		{
+			value:     "0",
+			shouldErr: true,
+		},
+		{
+			value:     "1",
+			shouldErr: false,
+		},
+		{
+			value:     "24",
+			shouldErr: false,
+		},
+		{
+			value:     "25",
+			shouldErr: false,
+		},
+		{
+			value:     "26",
+			shouldErr: true,
+		},
+		{
+			value:     "255.255.255.0",
+			shouldErr: false,
+		},
+		{
+			value:     "255.255.255.128",
+			shouldErr: false,
+		},
+		{
+			value:     "128.0.0.0",
+			shouldErr: false,
+		},
+		{
+			value:     "255.255.225.0",
+			shouldErr: true,
+		},
+		{
+			value:     "127.0.0.0",
+			shouldErr: true,
+		},
+	}
+	runValidations(t, tests, "netmask", IsValidNetmask)
+}
--- a/pkg/minishift/network/settings.go
+++ b/pkg/minishift/network/settings.go
@@ -1,0 +1,73 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package network
+
+import (
+	"bytes"
+	"fmt"
+	"text/template"
+
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+)
+
+const (
+	configureIPAddressMessage      = "-- Attempting to set network settings ..."
+	configureIPAddressFailure      = "FAIL\n   not supported on this platform or hypervisor"
+	configureNetworkScriptTemplate = `DEVICE={{.Device}}
+IPADDR={{.IPAddress}}
+NETMASK={{.Netmask}}
+GATEWAY={{.Gateway}}
+DNS1={{.DNS1}}
+DNS2={{.DNS2}}
+`
+)
+
+type NetworkSettings struct {
+	Device    string
+	IPAddress string
+	Netmask   string
+	Gateway   string
+	DNS1      string
+	DNS2      string
+}
+
+func printNetworkSettings(networkSettings NetworkSettings) {
+	fmt.Println(configureIPAddressMessage)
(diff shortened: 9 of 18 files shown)


gbraad:
#### Result using DockerNAT
Note: be sure to use the preparation steps as mentioned in the first message

```powershell
PS C:\work\src\github.com\minishift\minishift\out\windows-amd64> .\minishift.exe start --networkipaddress 10.0.75.124 --network-gateway 10.0.75.1 --network-nameserver 8.8.8.8 --iso-url file://c:/Users/gbraad/.minishift/cache/iso/minishift-centos7.iso
-- Checking if Hyper-V driver is configured ... OK
-- Starting local OpenShift cluster using 'hyperv' hypervisor ...
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Attempting to set network settings ...
   Device:      eth0
   IP Address:  10.0.75.124/24
   Gateway:     10.0.75.1
   Nameservers: 8.8.8.8
-- Starting Minishift VM ............................................................. OK
   Pinging 10.0.75.1 ... OK
   Retrieving http://10.0.75.1:5000/v2/ ... OK
   Version: v3.6.0
   host-config-dir ... OK
   host-data-dir ... OK
   host-pv-dir ... OK
   host-volumes-dir ... OK
   routing-suffix ... OK
Pulled 2/4 layers, 73% complete
Pulled 3/4 layers, 96% complete
Extracting
Image pull complete
OpenShift server started.

    https://10.0.75.124:8443

You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin

File: pkg/minishift/network/settings.go [hunk 1]
```diff
@@ -0,0 +1,73 @@
...
+type NetworkSettings struct {
+	Device    string
+	IPAddress string
+	Netmask   string
+	Gateway   string
+	DNS1      string
+	DNS2      string
+}
+
+func PrintNetworkSetting(networkSettings NetworkSettings) {
```
gbraad: `PrintNetworkSettings`

cmd/minishift/cmd/config/config.go
```diff
@@ -120,6 +120,13 @@ var (
 	// Preflight values
 	CheckNetworkHttpHost = createConfigSetting("check-network-http-host", SetString, nil, nil, true)
 	CheckNetworkPingHost = createConfigSetting("check-network-ping-host", SetString, nil, nil, true)
+
+	// Network settings (Hyper-V only)
+	NetworkDevice = createConfigSetting("network-device", SetString, nil, nil, true)
+	IPAddress     = createConfigSetting("network-ipaddress", SetString, []setFn{validations.IsValidIPv4Address}, nil, true)
+	Netmask       = createConfigSetting("network-netmask", SetString, []setFn{validations.IsValidNetmask}, nil, true)
+	Gateway       = createConfigSetting("network-gateway", SetString, []setFn{validations.IsValidIPv4Address}, nil, true)
+	Nameserver    = createConfigSetting("network-nameserver", SetString, []setFn{validations.IsValidIPv4Address}, nil, true)
```
gbraad: Would like a slice to allow two nameservers, but this would need to be enforced in several places. So at the moment I think using a single option is OK, as this can be modified/improved in later commits

cmd/minishift/cmd/start.go
```diff
@@ -171,6 +172,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	//Adding active profile information to all instance config
 	addActiveProfileInformation()
 
+	// TODO: force static IP?
```
gbraad: At the moment the API does not return the other possible IP addresses, so I can not check for this in the `preflightCheckAfterStartinghost`

pkg/minishift/config/validations_test.go
```diff
@@ -188,3 +188,81 @@ func TestValidProxyURL(t *testing.T) {
...
+		{
+			value:     "0",
+			shouldErr: true,
+		},
+		{
+			value:     "1",
+			shouldErr: false,
+		},
+		{
+			value:     "24",
```
gbraad: default

pkg/minishift/config/validations_test.go
```diff
@@ -188,3 +188,81 @@ func TestValidProxyURL(t *testing.T) {
...
+			shouldErr: false,
+		},
+		{
+			value:     "25",
```
gbraad: last possible range

pkg/minishift/network/settings_linux.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package network
+
+import (
+	"fmt"
+)
+
+func ConfigureNetworking(machineName string, networkSettings NetworkSettings) {
+	fmt.Println(configureIPAddressMessage, configureIPAddressFailure)
```
gbraad: should never be shown, but allows to build an override function.

pkg/minishift/network/settings_darwin.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+
+package network
+
+import (
+	"fmt"
+)
+
+func ConfigureNetworking(machineName string, networkSettings NetworkSettings) {
+	fmt.Println(configureIPAddressMessage, configureIPAddressFailure)
```
gbraad: should never be shown, but allows to build an override function.

pkg/minishift/network/settings_windows.go
```diff
@@ -0,0 +1,95 @@
...
+	hvkvp "github.com/gbraad/go-hvkvp"
+	"github.com/golang/glog"
+	"github.com/minishift/minishift/pkg/minishift/shell/powershell"
+)
+
+const (
+	resendInterval = time.Second
+)
+
+func ConfigureNetworking(machineName string, networkSettings NetworkSettings) {
```
gbraad: enforced to be Windows only

cmd/minishift/cmd/start.go
```diff
@@ -324,6 +327,22 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
...
+			Device:    viper.GetString(configCmd.NetworkDevice.Name),
+			IPAddress: viper.GetString(configCmd.IPAddress.Name),
+			Netmask:   viper.GetString(configCmd.Netmask.Name),
+			Gateway:   viper.GetString(configCmd.Gateway.Name),
+			DNS1:      viper.GetString(configCmd.Nameserver.Name),
+		}
+
+		// Configure networking on startup only works on Hyper-V
+		if machineConfig.VMDriver == "hyperv" && networkSettings.IPAddress != "" {
+			minishiftNetwork.ConfigureNetworking(constants.MachineName, networkSettings)
```
gbraad: Although driver is checked, only the `_windows.go` file will have an implementation.

pkg/minishift/network/settings_windows.go
```diff
@@ -0,0 +1,95 @@
...
+	PrintNetworkSetting(networkSettings)
+
+	networkScript := FillNetworkSettingsScript(networkSettings)
+	record := hvkvp.NewMachineKVPRecord(machineName,
+		"MINISHIFT_NETWORKING",
+		// to allow sending multiple lines in the value we encode the script
+		base64.StdEncoding.EncodeToString([]byte(networkScript)))
+
+	command := hvkvp.NewMachineKVPCommand(record)
+	b := newConfigBasher(doConfigure, command)
```
gbraad: This exist to resend the command over-and-over, until a `4096` is returned, which means the command succeeded.

cmd/minishift/cmd/start.go
gbraad: Should we however abstract the `Driver.GetIP()`-call and return the static assigned IP address?

gbraad:
## Note

On an External Switch you will always receive a local DHCP address. And due to how libmachine does `GetIP` you will get `NetworkAdapters[0].IpAddress[0]`, which is likely the DHCP assigned IP.

I have considered a `--network-disable/no-dhcp` flag, but this means to also customize our image even further by modifying the NetworkManager systemd unit file

budhrg:
@gbraad , I am getting following error twice using `DockerNAT`. 
```
PS C:\gowork\src\github.com\minishift\minishift> minishift.exe start --network-ipaddress 10.0.75.124 --network-gateway 10.0.75.1 --network-nameserver 8.8.8.8 --iso-url file://C:/minishift-cen
tos7.iso
-- Installing default add-ons ... OK
-- Checking if Hyper-V driver is installed ... OK
-- Checking if Hyper-V driver is configured to use a Virtual Switch ... OK
-- Starting local OpenShift cluster using 'hyperv' hypervisor ...
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Attempting to set network settings ...
   Device:      eth0
   IP Address:  10.0.75.124/24
   Gateway:     10.0.75.1
   Nameservers: 8.8.8.8
-- Starting Minishift VM ................................................................................................................................................................... OK
   Pinging 8.8.8.8 ... OK
   Retrieving http://minishift.io/index.html ... FAIL
   VM cannot connect to external URL with HTTP
-- Downloading OpenShift binary 'oc' version 'v3.6.0'
 33.92 MiB / 33.92 MiB [==========================================================================================================================================================] 100.00% 0s-
- Downloading OpenShift v3.6.0 checksums ... OK
   Version: v3.6.0
   host-pv-dir ... OK
   host-volumes-dir ... OK
   routing-suffix ... OK
   host-config-dir ... OK
   host-data-dir ... OK
-- Checking for openshift/origin:v3.6.0 image ...
FAIL
   Error: error pulling Docker image openshift/origin:v3.6.0
   Caused By:
     Error: Network timed out while trying to connect to https://index.docker.io/v1/repositories/openshift/origin/images. You may want to check your internet connection or if you are behind a
 proxy.
Error during 'cluster up' execution: Error starting the cluster.
```

Any idea  why network timed out here. Internet connection is there as you can see oc got downloaded.

gbraad:
@budhrg try `minishift ssh -- ping www.redhat.com` to make sure DNS resolves. It should as the preflight  `CheckHttPConnectivity` already confirmed. I have this also at times to be honest... (Note: pinging index.docker.io is of no use, as this is hosted on aws... curling might help to indicate what happens: `minishift ssh -- curl https://index.docker.io/` )

gbraad:
@budhrg `oc` downloads locally (not on the VM), so can you confirm NAT is properly set? `Get-NetNAT` ?

budhrg:
@gbraad , Ya DNS resolution issue. 

```
PS C:\gowork\src\github.com\minishift\minishift> minishift.exe ssh -- curl https://index.docker.io/
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:19 --:--:--     0curl: (6) Could not resolve host: index.docker.io; Unknown error
Command failed: exit status 6
```

gbraad:
@budhrg `minishift.exe ssh -- cat /etc/resolv.conf` ?

docs/source/using/managing-minishift.adoc
```diff
@@ -322,3 +322,52 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
...
+On Hyper-V a functionality is provided to assign an IP address on startup using the Data Exchange Service.
+
+[IMPORTANT]
+====
+-  This only works with the CentOS/RHEL based image in combination with Hyper-V
+-  DHCP is still enabled for fallback. Since this will timeout on an internal virtual switch it slightly increases the startup time.
+====
+
+To make this work you need to create a Virtual Switch using NAT
+
```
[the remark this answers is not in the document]
gbraad: This is an experimental feature, so likely only for advanced users. I am not really inclined to open up the floodgates and explain 'why' NAT is necessary.

docs/source/using/managing-minishift.adoc
```diff
@@ -322,3 +322,52 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
...
+[[hyperv-static-ip]]
+=== Assign IP address to Hyper-V (experimental)
+
+Since the Internal Virtual Switch for Hyper-V does not provide a DHCP offer option, an IP address needs to be provided in a different way.
+On Hyper-V a functionality is provided to assign an IP address on startup using the Data Exchange Service.
+
+[IMPORTANT]
+====
+-  This only works with the CentOS/RHEL based image in combination with Hyper-V
+-  DHCP is still enabled for fallback. Since this will timeout on an internal virtual switch it slightly increases the startup time.
```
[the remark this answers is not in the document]
gbraad: `DHCPRequest` will be sent, but no `DHCPOffer` will follow. The unit for NetworkManager will timeout per config... so hence the timeout. remark.

budhrg:
Since, I was connected to VPN, using the DNS nameserver my network is using resolved the issue.

gbraad:
@budhrg for your information:

```
   Pinging 8.8.8.8 ... OK
   Retrieving http://minishift.io/index.html ... FAIL
```

the first test checks if the VM can reach IP addresses outside of the VM, this means that the route and gateway is properly set up

the second test checks if DNS resolves, and if HTTP is possible to traverse the network. This last therefore can also confirm if the proxy is correctly set up.

pkg/minishift/network/settings_darwin.go
budhrg: :+1:

pkg/minishift/network/settings_windows.go
budhrg: Yes, since file is appended with `_windows.go`

pkg/minishift/network/settings.go
```diff
@@ -0,0 +1,73 @@
...
+	fmt.Println("   IP Address: ", fmt.Sprintf("%s/%s", networkSettings.IPAddress, networkSettings.Netmask))
+	if networkSettings.Gateway != "" {
+		fmt.Println("   Gateway:    ", networkSettings.Gateway)
+	}
+	if networkSettings.DNS1 != "" {
+		fmt.Println("   Nameservers:", fmt.Sprintf("%s %s", networkSettings.DNS1, networkSettings.DNS2))
+	}
+}
+
+func FillNetworkSettingsScript(networkSettings NetworkSettings) string {
```
budhrg: The naming is sounding confusing to me. It seems like this function fill the network settings into script(file).

Probably, `FillNetworkSettingsInfo` or similar?

pkg/minishift/network/settings_windows.go
```diff
@@ -0,0 +1,95 @@
...
+	b.start()
+}
+
+func doConfigure(success chan bool, command string) {
+	posh := powershell.New()
+	result, _ := posh.Execute(command)
+
+	if strings.Contains(result, "4096") {
```
budhrg: What is `4096` here?

cmd/minishift/cmd/start.go
gbraad: https://github.com/minishift/minishift/issues/1457#issuecomment-332811221

pkg/minishift/network/settings_windows.go
```diff
@@ -0,0 +1,95 @@
...
+		if glog.V(5) {
+			fmt.Printf("*")
+		}
+		success <- true
+	}
+}
+
+type bashingFunc func(handler chan bool, command string)
+
+type configbasher struct {
```
budhrg: Isn't this and related methods along with constructor function `newConfigBasher` should be extracted out into another file?

docs/source/using/managing-minishift.adoc
```diff
@@ -322,3 +322,52 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
 
 `extra-clusterup-flags`::
 Enables passing flags that are not directly exposed in the Minishift CLI directly to `oc cluster up`.
+
+
+[[hyperv-static-ip]]
+=== Assign IP address to Hyper-V (experimental)
+
+Since the Internal Virtual Switch for Hyper-V does not provide a DHCP offer option, an IP address needs to be provided in a different way.
+On Hyper-V a functionality is provided to assign an IP address on startup using the Data Exchange Service.
```
budhrg: please add comma after `On Hyper-V` to make it more readable.

docs/source/using/managing-minishift.adoc
```diff
@@ -322,3 +322,52 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
...
+
+[IMPORTANT]
```
budhrg: Could you replace it with `NOTE` and here as well https://github.com/gbraad/minishift/blob/e2e6dab3af183b7660cb877d59a5342f23503751/docs/source/using/managing-minishift.adoc#experimental-features. It is breaking consistency IMO as we have `NOTE` everywhere.

docs/source/using/managing-minishift.adoc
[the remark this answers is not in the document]
budhrg: This line need to be elaborated more to make user understand more.

docs/source/using/managing-minishift.adoc
```diff
@@ -322,3 +322,52 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
...
+
+- link:https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/setup-nat-network[Set up a NAT network]
+
+
+[NOTE]
+====
+WinNAT is limited to one NAT network per host. For more details about capabilities, and limitations, please see the link:https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/setup-nat-network[WinNAT capabilities and limitations blog]
+====
+
+The following command will attempt to assign an IP address for use on the Virtual Switch 'Internal'.
```
budhrg: if `Internal` is internal switch then please mentioned it as well `... address for use on the internal Virtual Switch 'Internal'.`

docs/source/using/managing-minishift.adoc
```diff
@@ -322,3 +322,52 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
...
+
+----
+PS> $env:MINISHIFT_ENABLE_EXPERIMENTAL="y"
+PS> $env:HYPERV_VIRTUAL_SWITCH="Internal"
+PS> minishift.exe start `
```
budhrg: Extra back-tick at the end.

docs/source/using/managing-minishift.adoc
budhrg: It seems, `Internal` is external switch here right?

docs/source/using/managing-minishift.adoc
```diff
@@ -322,3 +322,52 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
...
+  --iso-url centos `
+  --network-ipaddress 192.168.1.10 `
+  --network-gateway 192.168.1.1 `
+  --network-nameserver 8.8.8.8
+----
+
+
+If you want to use the 'DockerNAT' network, the following commands are needed to setup the correct NAT networking and assigning an IP in the range expected.
```
budhrg: This is the case for using internal virtual switch or DockerNAT which is internal right?

pkg/minishift/network/settings_windows.go
gbraad: I didn't as they served no purpose other than this usecase.

pkg/minishift/network/settings.go
gbraad: It does become a script, but the script is unrolled in the iso.

docs/source/using/managing-minishift.adoc
gbraad: Above in the same file IMPORTANT is used. This is not a note, but a block with requirements.

pkg/minishift/network/settings.go
gbraad: The idea is to re-use this in a follow-up task.
The naming was done as it is using the old networj-script structure and info fields

pkg/minishift/network/settings_windows.go
gbraad: It is success result code for WMI objects. Can't help the weird numbering... But this is the implementation.

pkg/minishift/network/settings_windows.go
gbraad: @budhrg this is where I explained the resultcode

pkg/minishift/network/settings_windows.go
gbraad: Since it is followed by
```
  success <- true
```
I guess it is/was clear enough

glide.yaml
```diff
@@ -65,6 +65,8 @@ import:
   version: 3bc7a60b1df80a5766820c4a53d59668f4553f75
 - package: github.com/juju/errors
   version: c7d06af17c68cd34c835053720b21f6549d9b0ee
+- package: github.com/gbraad/go-hvkvp
+  version: a692e030379b6de28a10b7697decdf8109899f6e
```
praveenkumar: It would be better to use tag/release-version instead of sha.

pkg/minishift/clusterup/clusterup.go
```diff
@@ -90,13 +91,15 @@ func ClusterUp(config *ClusterUpConfig, clusterUpParams map[string]string, runne
...
-	// Deal with extra flags (add to command arguments)
-	if len(extraFlags) > 0 {
-		fmt.Println("-- Extra `oc` cluster up flags (experimental) ... ")
-		fmt.Printf("   '%s'\n", extraFlags)
-		extraFlagFields := strings.Fields(extraFlags)
-		for _, extraFlag := range extraFlagFields {
-			cmdArgs = append(cmdArgs, extraFlag)
+	if minishiftConfig.EnableExperimental {
+		// Deal with extra flags (add to command arguments)
+		if len(extraFlags) > 0 {
```
praveenkumar: 👍

glide.lock
```diff
@@ -104,6 +104,8 @@ imports:
   version: bbd0c6e271208dce66d8fda4bc536453cd27fc4a
 - name: github.com/juju/errors
   version: c7d06af17c68cd34c835053720b21f6549d9b0ee
+- name: github.com/gbraad/go-hvkvp
+  version: a692e030379b6de28a10b7697decdf8109899f6e
```
gbraad: https://github.com/gbraad/go-hvkvp/releases/tag/0.3

glide.yaml
gbraad: https://github.com/gbraad/go-hvkvp/releases/tag/0.3

gbraad:
Updated ISO is needed from: http://artifacts.ci.centos.org/minishift/minishift-centos-iso/pr/164/

cmd/minishift/cmd/start.go
```diff
@@ -476,6 +496,14 @@ func initStartFlags() *flag.FlagSet {
 	startFlagSet.AddFlag(registryMirrorFlag)
 	startFlagSet.AddFlag(cmdUtil.AddOnEnvFlag)
 
+	if minishiftConfig.EnableExperimental {
+		startFlagSet.String(configCmd.NetworkDevice.Name, "eth0", "Specify the network device to use for the IP address. Ignored if no IP address specified (experimental - Hyper-V only)")
+		startFlagSet.String(configCmd.IPAddress.Name, "", "Specify IP address to assign to the instance (experimental - Hyper-V only)")
+		startFlagSet.String(configCmd.Netmask.Name, "24", "Specify netmask to use for the IP address. Ignored if no IP address specified (experimental - Hyper-V only)")
+		startFlagSet.String(configCmd.Gateway.Name, "", "Specify gateway to use for the instance. Ignored if no IP address specified (experimental - Hyper-V only)")
+		startFlagSet.String(configCmd.Nameserver.Name, "8.8.8.8", "Specify nameserver to use for the instance. Ignored if no IP address specified (experimental - Hyper-V only)")
```
gbraad: remove default!

pkg/minishift/network/settings_windows.go
gbraad: This will likely be refactored anyway, as it will be re-used by minikube. Extracting this logic would add additional complexity.

docs/source/using/managing-minishift.adoc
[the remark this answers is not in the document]
gbraad: Removed as https://github.com/minishift/minishift-centos-iso/pull/164 makes this irrelevant

glide.yaml
gbraad: https://github.com/gbraad/go-hvkvp/releases/tag/0.3

gbraad:
@LalatenduMohanty @praveenkumar @budhrg please review

gbraad:
ping?

glide.yaml
```diff
@@ -65,6 +65,8 @@ import:
   version: 3bc7a60b1df80a5766820c4a53d59668f4553f75
 - package: github.com/juju/errors
   version: c7d06af17c68cd34c835053720b21f6549d9b0ee
+- package: github.com/gbraad/go-hvkvp
+  version: 0.3
```
praveenkumar: 👍

pkg/minishift/config/validations.go
```diff
@@ -121,3 +121,36 @@ func IsValidUrl(_ string, isoURL string) error {
...
+func IsValidIPv4Address(name string, address string) error {
+	ip := net.ParseIP(address).To4()
+	if ip == nil {
+		return fmt.Errorf("'%s' is not a valid IPv4 address", address)
+	}
+
+	return nil
+}
+
+func IsValidNetmask(name string, mask string) error {
```
praveenkumar: @gbraad where is the `name` variable used for this method?

gbraad: fixed... although I think we should look critically at the text, as it now the correct use of English.

pkg/minishift/network/settings.go
LalatenduMohanty: @gbraad Both FillNetworkSettingsScript and PrintNetworkSetting are currently used in the same package.So I do not see any value in making them public functions. Is there any specific reason you wanted them as public functions?

cmd/minishift/cmd/config/config.go
LalatenduMohanty: @gbraad Do you already have a follow up issue for this?

cmd/minishift/cmd/start.go
LalatenduMohanty: @gbraad Please remove the comment as it is not required for this PR.

cmd/minishift/cmd/start.go
LalatenduMohanty: @gbraad I am concerned about the defaults. This can create issues if user does not provide proper values as per user particular network, isn't it?

cmd/minishift/cmd/start.go
```diff
@@ -332,6 +335,22 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
...
+			Netmask:   viper.GetString(configCmd.Netmask.Name),
+			Gateway:   viper.GetString(configCmd.Gateway.Name),
+			DNS1:      viper.GetString(configCmd.Nameserver.Name),
+		}
+
+		// Configure networking on startup only works on Hyper-V
+		if machineConfig.VMDriver == "hyperv" && networkSettings.IPAddress != "" {
+			minishiftNetwork.ConfigureNetworking(constants.MachineName, networkSettings)
+		}
+	}
```
LalatenduMohanty: @gbraad We should check for CentOS or RHEL and communicate to user about if the options will not for users ISO else if user uses B2D it will fail and user wont have any clue unless user has read the documentation.

LalatenduMohanty: I think there should be a validation function or something which would make sure to print right message for users if it is not a CentOS or RHEL ISO and on Windows HyperV

gbraad:
The functions are public for initial use by `start.go` but left them public for the follow-up issue that deals with this for other hypervisors, as in that case we will use a file that gets inserted in the filesystem of the VM.

cmd/minishift/cmd/start.go
gbraad: windows only already... and here unsured to be hyperv only...

RHEL will support this...

gbraad: we can not check IsCDK/IsCentOS image, as this is done AFTER the the vm already started and starts provisioning... this is too late.

therefore the message is now: "Attempting to set network settings"... and the preflight check after host start can check for the final IP... but this is an improvement task. Remember this is experimental.

docs/source/using/managing-minishift.adoc
```diff
@@ -322,3 +322,57 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
...
+[[hyperv-static-ip]]
+=== Assign IP address to Hyper-V (experimental)
+
+Since the Internal Virtual Switch for Hyper-V does not provide a DHCP offer option, an IP address needs to be provided in a different way.
+For Hyper-V a functionality is provided to assign an IP address on startup using the Data Exchange Service.
+
+[IMPORTANT]
+====
+-  This only works with the CentOS/RHEL based image in combination with Hyper-V
+====
```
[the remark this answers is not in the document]
gbraad: at this stage? it is experimental... not a GA

cmd/minishift/cmd/config/config.go
gbraad: not specifically... but will do now...

docs/source/using/managing-minishift.adoc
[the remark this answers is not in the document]
LalatenduMohanty: @gbraad I still think we should add some explanation about why it does not work with b2d atm.

docs/source/using/managing-minishift.adoc
```diff
@@ -322,3 +322,57 @@ Enables provisioning the OpenShift link:https://docs.openshift.org/latest/archit
...
+
+The following command will attempt to assign an IP address for use on the Internal Virtual Switch 'MyInternal'.
+
+----
+PS> $env:MINISHIFT_ENABLE_EXPERIMENTAL="y"
+PS> $env:HYPERV_VIRTUAL_SWITCH="MyInternal"
+PS> minishift.exe start `
+  --iso-url centos `
+  --network-ipaddress 192.168.1.10 `
```
hferentschik: So these options appear in the start command, even though they only apply for Hyper-V? Would it not be nicer to have a more generic approach.  I am concerend about flag proliferation and also people trying to use these flags with the wrong driver?

If there is no generic mechanism, we could while it is experimental go with env settings only. This is how driver specific options are set. So why could an initial version of this not be based on some HYPERV_XYZ options? 

We already had some bad experience having a VirtuaBox specific flag once upon a time.

hferentschik:
>  minishift.exe start --network-ipaddress 10.0.21.40 --network-gateway 10.0.21.1 --network-nameserver 8.8.8.8

Do we really have to expose this as explicit flags to the start command? I think this is opening us up for all sorts of misunderstanding. I'd like to see a more generic approach.

docs/source/using/managing-minishift.adoc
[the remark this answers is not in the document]
hferentschik: It does not work with B2D? Why not? If this is the case it definitely needs mentioning, experimental or not. After all it is our default ISO.

docs/source/using/managing-minishift.adoc
gbraad: Why prefix, as they will also work for other hypervisors in the follow up issue... just not on startup.

they are currently hidden behind ENABLE_EXPERIMENTAL, but can be enforced with `runtime.GOOS == "windows"`

gbraad: On startup, you will likely only use `--network-ipaddress` as the remaining values are network-specific, so you will likely store them as

```
m config set network-gateway 10.0.75.1
m config set network-netmask 24
m config set network-nameserver 208.67.220.220
```
anyways

gbraad:
> misunderstanding

such as?

> Do we really have to expose this

After
```
PS> minishift config set network-gateway 10.0.75.1
PS> minishift config set network-nameserver 208.67.220.220
```
all that remains is

```
PS> minishift.exe start --network-ipaddress 10.0.21.40
```

which in improvements will be automagically resolved: #1492, #1493...

docs/source/using/managing-minishift.adoc
gbraad: If hypervisor or platform does not support this, the following is shown:

```
PS C:\work\src\github.com\minishift\minishift\out\windows-amd64> .\minishift.exe start --vm-driver virtualbox --network-ipaddress 10.0.21.128
-- Starting local OpenShift cluster using 'virtualbox' hypervisor ...
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Attempting to set network settings ... FAIL
   not supported on this platform or hypervisor
-- Starting Minishift VM ... 
```

cmd/minishift/cmd/start.go
gbraad: Addressed in #1493

gbraad: The solution is not as straightforward... See previously mentioned follow-up.

LalatenduMohanty:
@gbraad PR looks fine to me. Unfortunately I am not able to test it because of my HyperV setup issue. Please rebase the PR and  we are good to merge IMO.

pkg/minishift/config/validations.go
```diff
@@ -121,3 +121,36 @@ func IsValidUrl(_ string, isoURL string) error {
...
+	ip := net.ParseIP(address).To4()
+	if ip == nil {
+		return fmt.Errorf("%s IPv4 address is not valid: %s", name, address)
+	}
+
+	return nil
+}
+
+func IsValidNetmask(name string, mask string) error {
+	err := fmt.Errorf("%s netmask is not valid: %s", name, mask)
```
praveenkumar: 👍

LalatenduMohanty:
> Do we really have to expose this as explicit flags to the start command? I think this is opening us up for all sorts of misunderstanding. I'd like to see a more generic approach.

IMO this is a generic approach and can be extended to other hypervisors if it can support this. The error message is clear enough with respect to if it does not supported in an OS or hypervisor. As this is an experimental feature I am going to merge it and then we can discuss in the follow up issues (which @gbraad has already created) if we want to change anything.

PR #1410, opened 2017-09-19
Issue #1407 Making profile tests part of integration tests

### LINKED ISSUE #1407: Enable profile integration tests in CentOS CI
Currently https://github.com/minishift/minishift/blob/master/centos_ci.sh only runs below tests. 
```
MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags=basic,openshift,version,cmd-config -format pretty"
```
So we should add basic tag to profile tests [1]  i.e. `s/@profile/@profile @basic/` as this should be part of sanity testing.

[1] https://github.com/minishift/minishift/blob/master/test/integration/features/profile.feature#L1

LalatenduMohanty:
Fixes #1407 

### DIFF
--- a/test/integration/features/profile.feature
+++ b/test/integration/features/profile.feature
@@ -1,4 +1,4 @@
-@profile
+@profile @command
 Feature: Profile
   As a user I can perform basic operations of Minishift with profile feature
 
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -305,7 +305,7 @@ function build_and_test() {
   cd $GOPATH/src/github.com/minishift/minishift
   make prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags=basic,openshift,version,cmd-config -format pretty"
+  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="-tags=basic,openshift,version,cmd-config,profile -format pretty"
   echo "CICO: Tests ran successfully"
 
   artifacts_upload_on_pr_and_master_trigger $1;


gbraad:
I do not think profiles is a basic functionality, but rather advanced usage.
I think the majority of people will not have the need for this.

It is in my opinion rather part of the commands we test.
Our basic test is what we also call a 'smoke' test...

test/integration/features/profile.feature
```diff
@@ -1,4 +1,4 @@
-@profile
+@profile @basic
```
gbraad: The basic test only covers a 'smoke' test scenario

LalatenduMohanty:
> I do not think profiles is a basic functionality, but rather advanced usage.
> I think the majority of people will not have the need for this.
> 
> It is in my opinion rather part of the commands we test.
> Our basic test is what we also call a 'smoke' test...

@gbraad I was not sure it as user always the default profile even if user does not know about it. For example `minishift start` will start the minishift profile and `minishift profile list` lists the default profile.  WDYT?
@minishift/minishift-dev thoughts?

LalatenduMohanty:
I think the ideal way to fix this is to add a tests in basic tests to check `minishift profile list` and make the detailed profile tests  as part of commands tag.

gbraad:
I disagree. A smoke test does a basic verification test and therefore I do not believe profile suffices for this. Else, we should also add environment variables, overriding the paths, etc. For me, basic should only cover the OOTB (Out of the Box) experience, in which a user will do `m start` and have fun with Minishift, and cleans up. If this does not work, we can not release at all...

`m profiles` and `m --profile` to me is an advanced use-case, that goes beyond the basic verification test.  This is certainly not something you will do in the first 5 minutes of using Minishift (as a new user).

LalatenduMohanty:
> m profiles and m --profile to me is an advanced use-case, that goes beyond the basic verification test. This is certainly not something you will do in the first 5 minutes of using Minishift (as a new user).

@gbraad Do you disagree about https://github.com/minishift/minishift/pull/1410#issuecomment-330453421 ?

gbraad:
Also, `m profile list` not returning the expected information, does not prevent a basic integration test to run, or even being the gated check to start other integration tests.

LalatenduMohanty:
@gbraad smoke / basic tests are not 100% equal to what features user uses. Sure it definitely have a strong influence. But it also should cover what we consider as basic functionality of Minishift project. For example we run basic tests for each PR to check if we have broken any basic functionality. I strongly feel there should be some basic tests of profile in `basic` tag, just to make sure new changes does not break it.

PR #1416, opened 2017-09-20
Issue #1364 Provide correct selinux label

### LINKED ISSUE #1364: hostfolders need to add extra flags to cifs mount for selinux labels
Using RHEL as a minishift iso (CDK), when mounting a hostfolder, the following command is executed (example):
```
sudo mount -t cifs //192.168.99.1/rgordill/Clients/minishift/persistent /var/lib/persistent-on-hosts -o username=rgordill,password=minishift
```
With that, selinux labels are as follows:

`system_u:object_r:cifs_t:s0      persistent-on-hosts/`

If we want to use this share to persist the configuration, we get the following error:
```
   Creating initial OpenShift configuration
FAIL
   Error: could not copy OpenShift configuration
   Caused By:
     Error: Error response from daemon: {"message":"SELinux relabeling of /var/lib/persistent-on-hosts/openshift.local.config is not allowed: \"operation not supported\""}
```
That is because docker cannot relabel the selinux policy. But if we mount with:
```
sudo mount -t cifs //192.168.99.1/rgordill/Clients/minishift/persistent /var/lib/persistent-on-hosts -o username=rgordill,password=minishift,context=system_u:object_r:svirt_sandbox_file_t:s0
```
Then, the context is correct:

`system_u:object_r:svirt_sandbox_file_t:s0 persistent-on-hosts/`

Is it possible to add some extra flags, like "context" to hostfolders, to pass them to cifs mount?

praveenkumar:
Fix #1364 

- Provide correct selinux labels for mounting cifs
- Add flags for ISO type in the config

### DIFF
--- a/pkg/minishift/hostfolder/hostfolder.go
+++ b/pkg/minishift/hostfolder/hostfolder.go
@@ -29,6 +29,7 @@ import (
 	"github.com/spf13/viper"
 
 	config "github.com/minishift/minishift/pkg/minishift/config"
+	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
 	miniutil "github.com/minishift/minishift/pkg/minishift/util"
 	"github.com/minishift/minishift/pkg/util"
 )
@@ -327,7 +328,9 @@ func mountCifsHostfolder(driver drivers.Driver, hostfolder *config.HostFolder) e
 		hostfolder.Options["username"],
 		password)
 
-	cmd = fmt.Sprintf("%s,context=system_u:object_r:svirt_sandbox_file_t:s0", cmd)
+	if minishiftConfig.InstanceConfig.IsRHELBased {
+		cmd = fmt.Sprintf("%s,context=system_u:object_r:svirt_sandbox_file_t:s0", cmd)
+	}
 
 	if len(hostfolder.Options["domain"]) > 0 { // != ""
 		cmd = fmt.Sprintf("%s,domain=%s", cmd, hostfolder.Options["domain"])
--- a/pkg/minishift/hostfolder/hostfolder.go
+++ b/pkg/minishift/hostfolder/hostfolder.go
@@ -327,6 +327,8 @@ func mountCifsHostfolder(driver drivers.Driver, hostfolder *config.HostFolder) e
 		hostfolder.Options["username"],
 		password)
 
+	cmd = fmt.Sprintf("%s,context=system_u:object_r:svirt_sandbox_file_t:s0", cmd)
+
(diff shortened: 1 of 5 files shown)


pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -147,11 +152,9 @@ func (provisioner *MinishiftProvisioner) GenerateDockerOptions(dockerPort int) (
 
 	// systemd / redhat will not load options if they are on newlines
 	// instead, it just continues with a different set of options; yeah...
-	engineConfigTemplate := ""
+	engineConfigTemplate := engineConfigTemplateCentOS
 	if provisioner.GetRedhatRelease() {
 		engineConfigTemplate = engineConfigTemplateRHEL
-	} else {
-		engineConfigTemplate = engineConfigTemplateCentOS
```
praveenkumar: Just a simple best practise change which doesn't have any relation for this PR.

pkg/minishift/registration/redhat.go
```diff
@@ -55,9 +55,8 @@ func (registrator *RedHatRegistrator) CompatibleWithDistribution(osReleaseInfo *
 	}
 	if _, err := registrator.SSHCommand("sudo -E subscription-manager"); err != nil {
 		return false
-	} else {
-		return true
 	}
+	return true
```
praveenkumar: Just a simple best practise change which doesn't have any relation for this PR.

budhrg:
@praveenkumar , Unit tests failures.

pkg/minishift/hostfolder/hostfolder.go
```diff
@@ -327,6 +328,10 @@ func mountCifsHostfolder(driver drivers.Driver, hostfolder *config.HostFolder) e
 		hostfolder.Options["username"],
 		password)
 
+	if minishiftConfig.InstanceConfig.IsRHELBased {
```
gbraad: RHELBased 👍

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -147,11 +152,9 @@ func (provisioner *MinishiftProvisioner) GenerateDockerOptions(dockerPort int) (
 
 	// systemd / redhat will not load options if they are on newlines
 	// instead, it just continues with a different set of options; yeah...
-	engineConfigTemplate := ""
+	engineConfigTemplate := engineConfigTemplateCentOS
```
gbraad: I had actually more improvements for this code, but postponed them for now. Think I can share them and let's discuss...

budhrg:
Merged thanks.

PR #1418, opened 2017-09-20
Issue #680 .tar.gz contents should be in a subdirectory

praveenkumar:
Since this will go with 1.7.0 release that means we waited around 2 releases since we had our extraction feature in.

How to test
=========

Using this PR I created a test release https://github.com/praveenkumar/minishift/releases/tag/v1.7.0 which can use to test out minishift version 1.5 or 1.6 to upgrade to 1.7 

you need to change https://github.com/minishift/minishift/blob/master/pkg/minishift/update/update.go#L44 and add my githubowner `praveenkumar` build locally and then try to update.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -139,9 +139,19 @@ prerelease: clean fmtcheck test cross
 .PHONY: release
 release: clean $(GOPATH)/bin/gh-release cross
 	mkdir -p release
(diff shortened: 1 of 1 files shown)


anjannath:
tried extracting the tar ball. works for me.

PR #1419, opened 2017-09-20
Issue #1342 Moving the default add-on install to start

### LINKED ISSUE #1342: Default add-ons install message should not come for all commands
```
$ make integration GODOG_OPTS="-tags command -format pretty"
go install -pkgdir=/home/amit/go/src/github.com/minishift/minishift/out/bindata -ldflags="-X github.com/minishift/minishift/pkg/version.minishiftVersion=1.5.0 -X github.com/minishift/minishift/pkg/version.b2dIsoVersion=v1.1.0 -X github.com/minishift/minishift/pkg/version.centOsIsoVersion=v1.1.0 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v3.6.0 -X github.com/minishift/minishift/pkg/version.commitSha=0e75c4ec" ./cmd/minishift
mkdir -p /home/amit/go/src/github.com/minishift/minishift/out/integration-test
go test -timeout 3600s github.com/minishift/minishift/test/integration --tags=integration -v -args --test-dir /home/amit/go/src/github.com/minishift/minishift/out/integration-test --binary /home/amit/go/bin/minishift -tags command -format pretty
Keeping Minishift cache directory '/home/amit/go/src/github.com/minishift/minishift/out/integration-test/cache' for test run.
Log successfully started, logging into: /home/amit/go/src/github.com/minishift/minishift/out/integration-test/integration.log
Running Integration test in: /home/amit/go/src/github.com/minishift/minishift/out/integration-test
Using binary: /home/amit/go/bin/minishift
Feature: Basic
  As a user I can perform basic operations of Minishift and OpenShift

Feature: Openshift commands
  Commands "minishift openshift [sub-command]" are used for interaction with Openshift
  cluster in VM provided by Minishift.

Feature: Version command
  Command "minishift version" shows version of minishift binary.

  Scenario: User can get information about version of Minishift                  # features/cmd-version.feature:5
    When executing "minishift version" succeeds                                  # integration_test.go:652 -> github.com/minishift/minishift/test/integration.executingMinishiftCommandSucceedsOrFails
    Then stdout should match "^minishift v[0-9]\.[0-9]\.[0-9]\+[a-z0-9]{7,8}\n$" # integration_test.go:630 -> github.com/minishift/minishift/test/integration.commandReturnShouldMatchRegex
    Output did not match. Expected: '^minishift v[0-9]\.[0-9]\.[0-9]\+[a-z0-9]{7,8}\n$', Actual: '-- Installing default add-ons ... OK
minishift v1.5.0+0e75c4ec
'

Feature: Cool Store
  In order to test Minishift under load
  I need to setup a test environment
  Using CentOS and 4G of memory
  After which I will import image streams
  Finally I will deploy the Cool Store

Feature: Flags
  As a user I can perform advanced operations
  such as starting Minishift with optional features

Feature: Provision an older major release
  As a user I can provision an older major version of minishift

Feature: Minishift can run behind proxy
  As a user I can use Minishift behind a proxy.
  INFO: Code behind this feature will start a proxy server in background
  on port 8181 and will try to get IP of host machine automatically
  to set MINISHIFT_HTTP_PROXY for this feature only.
  However one can override port number or set IP manually if needed.
  To use custom port number run this test with environmental variable
  INTEGRATION_PROXY_CUSTOM_PORT.
  To use custom IP address please set INTEGRATION_PROXY_CUSTOM_IP
  environmental variable.

--- Failed scenarios:

    features/cmd-version.feature:7

1 scenarios (1 failed)
2 steps (1 passed, 1 failed)
67.708103ms
PASS
exit status 1
FAIL    github.com/minishift/minishift/test/integration 0.085s
make: *** [Makefile:176: integration] Error 1
12:46 $
```

LalatenduMohanty:
Fixes #1342 

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -36,6 +36,7 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	minishiftCluster "github.com/minishift/minishift/pkg/minishift/cluster"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
+	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
 	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
 	"github.com/minishift/minishift/pkg/minishift/docker"
 	"github.com/minishift/minishift/pkg/minishift/docker/image"
@@ -45,6 +46,7 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
 
 	"github.com/minishift/minishift/pkg/util"
+	"github.com/minishift/minishift/pkg/util/filehelper"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/minishift/minishift/pkg/util/progressdots"
 	stringUtils "github.com/minishift/minishift/pkg/util/strings"
@@ -150,16 +152,22 @@ func runStart(cmd *cobra.Command, args []string) {
 	ensureNotRunning(libMachineClient, constants.MachineName)
 	validateOpenshiftVersion()
 
+	// to determine whether we need to run post cluster up actions,
+	// we need to determine whether this is a restart prior to potentially creating a new VM
+	isRestart := cmdUtil.VMExists(libMachineClient, constants.MachineName)
+
+	// Install default add-ons when it is a fresh start. For a restart
+	// we should not install as user might have deleted them intentionally.
+	if !isRestart {
+		installDefaultAddons()
+	}
+
 	// preflight check (before start)
 	preflightChecksBeforeStartingHost()
 
 	setSubscriptionManagerParameters()
 
 	proxyConfig := handleProxies()
-
-	// to determine whether we need to run post cluster up actions,
-	// we need to determine whether this is a restart prior to potentially creating a new VM
-	isRestart := cmdUtil.VMExists(libMachineClient, constants.MachineName)
 
 	fmt.Printf("-- Starting local OpenShift cluster")
 	hostVm := startHost(libMachineClient)
@@ -632,5 +640,16 @@ func cacheMinishiftISO(config *cluster.MachineConfig) {
 		if err := config.CacheMinikubeISOFromURL(); err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error caching the ISO: %s", err.Error()))
 		}
+	}
+}
+
+// Run the default addons on fresh minishift start
+func installDefaultAddons() {
+	if filehelper.IsEmptyDir(minishiftConfig.InstanceDirs.Addons) {
+		fmt.Print("-- Installing default add-ons ... ")
+		if err := cmdUtil.UnpackAddons(minishiftConfig.InstanceDirs.Addons); err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error installing default add-ons : %s", err))
+		}
+		fmt.Println("OK")
 	}
 }
--- a/test/integration/features/cmd-addons.feature
+++ b/test/integration/features/cmd-addons.feature
@@ -8,9 +8,12 @@ Feature: Addons command and its subcommands
       And executing "minishift add-ons" fails
       And executing "minishift add-on" fails
 
-  @minishift-only
-  Scenario: Default add-ons are installed but not enabled by default
-  Note: default addons were installed during previous commands automatically.
+ Scenario: Installing default add-ons manually
+     When executing "minishift addons install --defaults" succeeds
+     Then stdout should contain
+      """
+      Default add-ons anyuid, admin-user, xpaas, registry-route installed
+      """
      When executing "minishift addons list"
      Then stdout should match "admin-user\s*: disabled\s*P\(0\)"
       And stdout should match "anyuid\s*: disabled\s*P\(0\)"
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -36,7 +36,6 @@ import (
 	"github.com/minishift/minishift/cmd/minishift/cmd/image"
 	cmdOpenshift "github.com/minishift/minishift/cmd/minishift/cmd/openshift"
 	cmdProfile "github.com/minishift/minishift/cmd/minishift/cmd/profile"
-	"github.com/minishift/minishift/cmd/minishift/cmd/util"
 	cmdUtil "github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
@@ -70,8 +69,7 @@ var RootCmd = &cobra.Command{
 	Long:  `Minishift is a command-line tool that provisions and manages single-node OpenShift clusters optimized for development workflows.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		var (
-			err                    error
-			isAddonInstallRequired bool
+			err error
 		)
 
 		constants.MachineName = constants.ProfileName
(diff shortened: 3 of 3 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -161,6 +163,12 @@ func runStart(cmd *cobra.Command, args []string) {
 	// we need to determine whether this is a restart prior to potentially creating a new VM
 	isRestart := cmdUtil.VMExists(libMachineClient, constants.MachineName)
```
gbraad: Can we move this one as much as possible to the top of the function?

This file now contains three `isRestart`, but we use them all as `!isRestart`... so maybe we should invert the meaning of the variable to something like `isFirstStart`? WDYT

LalatenduMohanty: @gbraad I think isRestart is fine and communicates the right intent.

LalatenduMohanty: On second thought I guess you meant `isFreshStart` I am fine with both isRestart and isFreshStart.

gbraad: The most important is to move it to the top of the function, as this is a conditional statement which is important for the rest of the execution.

gbraad: not a huge fan of inverted logic... especially when this is the only way we use it. inversion can lead to trouble in semantics (reading it and reasoning about it)

LalatenduMohanty:
Need to fix the integration tests as the add-ons will not be installed before `minishift start`.

test/integration/features/cmd-addons.feature
```diff
@@ -8,9 +8,12 @@ Feature: Addons command and its subcommands
...
 
-  @minishift-only
-  Scenario: Default add-ons are installed but not enabled by default
-  Note: default addons were installed during previous commands automatically.
+ Scenario: Installing default add-ons manually
+     When executing "minishift addons install --defaults" succeeds
+     Then stdout should contain
+      """
+      Default add-ons anyuid, admin-user, xpaas, registry-route installed
+      """
```
LalatenduMohanty: @agajdosi @amitkrout FYI The add-on list default behavior is changing with this PR.

hferentschik:
I need to catch up on the context and I am currently still in transit, but there is a good reason why the default add-on install is not occurring during start. There are several use cases where one want them installed prior to calling start. For example I might want to enable a given add-on or set it's priority prior to starting. There is also the interaction with the _update_ command. This has been discussed before afair. Not sure what the problem is here, but it seems wrong to me.

LalatenduMohanty:
> There is also the interaction with the update command. This has been discussed before afair

The update context is different and the code is still present in root.go and there is not change in behavior.

LalatenduMohanty:
> I need to catch up on the context and I am currently still in transit, but there is a good reason why the default add-on install is not occurring during start. There are several use cases where one want them installed prior to calling start. For example I might want to enable a given add-on or set it's priority prior to starting.

Sure lets discuss when you are here. But the behavior around when to execute is not changing. Also `minishift addons install --defaults` behavior does not change.

gbraad:
An explicit command for this use case 'before start' seems more fitting than having side-effects occur on commands like `m version`, `m status`, `m ip`, `m profile`, `m config`, etc, etc.

gbraad:
Logic should installs the addons 'before the start' of the VM, pretty much right before the preflightChecks. Nothing of libmachine is triggered at that time, so no VM exists

gbraad:
@budhrg @praveenkumar do you remember why `root.go` was needed?
@budhrg can you check if this in any way would influence the update scenario?

@LalatenduMohanty can you repush (so the merge disappears? "Update branch" seems broken these days)

gbraad:
@budhrg unchanged

however, `Default add-ons will be updated the next time you run any 'minishift' command.` will likely have to change going forward. WDYT? `Any` => `m addons` and `m start` would be understandable.

gbraad:
Any update or comments? if not, I will merge today...

test/integration/features/cmd-addons.feature
agajdosi: @LalatenduMohanty Sorry I am replying late. This change is correct, in other features we use addons only in basic and there is seems that this PR won't affect the test.

gbraad:
Merged with 349f4084b706e55e96cc536e441b161a7c6c2f87

hferentschik:
@gbraad why did you merge? I was still having a discussion with @LalatenduMohanty? The behaviour did change for the first time install. In this case the default add-ons don't get installed in time. I suggest to reopen this issue.

hferentschik:
> do you remember why root.go was needed?

Because you might want to interact with add-ons prior (enable, disable, change priorities) priror to starting the VM. _root.go_ is the place where we ensure the whole Minishift environment is in a consistent state. Add-ons are not tied to a 'minishift start'. If the whole issues is related to some (unwanted) output, let's talk about this and try to find a solution for that.

gbraad:
Can you provide a testcase which indicates this situation? As at the moment, none of the errors you have mentioned have been triggered in a test or broke integration. This is undocumented behaviour.

> Add-ons are not tied to a 'minishift start'.

If this the case, they should have a place inside the `m addon` command and not inside root.  Default addons, however, need to be installed as part of the profile, prior to the actual start of the VM. This is happening...

hferentschik:
> Can you provide a testcase which indicates this situation? As at the moment, none of the errors you have mentioned have been triggered in a test or broke integration. 

That does not mean that the behaviour has not changed. Take this (clean install, no prior Minishift)

    $ minishift add-on enable anyuid
    $ minishift start

This used to work, it does not anymore, since now the default add-ons get installed during start which is too late.

> If this the case, they should have a place inside the addon command and not inside root.

If you place it there you need to handle the default add-on install twice. Once in start and once in the addon command. PersistentPreRun is the one place where we can make sure that the Minishift install/config is complete and consistent. Hence the update functionality is there as well. 

> Default addons, however, need to be installed as part of the profile, prior to the actual start of the VM and this seems more logical to happen from the start itself. 

That's what I keep saying, it is not bound to the start command. If it were, I would not have a problem placing it there. However, there are use cases and workflows which need them installed earlier. ATM, a user on a clean system cannot configure the add-ons he wants prior to the first start, even if he wanted to.

gbraad:
> This used to work, it does not anymore, since now the default add-ons get installed during start which is too late.

I am not following you with 'too late'. What do you expect happens?

`minishift add-on enable anyuid` only enables the install, while `minishift start` now triggers the install of the add-on still right before the VM actually starts. I am not seeing the point of it being 'too late'

> ATM, a user on a clean system cannot configure the add-ons he wants prior to the first start, even if he wanted to.

Then I would suggest to move this to `m addon` and the Prerun, but not as part of `root.go`... or at least well documented and commented.

hferentschik:
> minishift add-on enable anyuid only enables the install, while minishift start now triggers the install of the add-on still right before the VM actually starts. I am not seeing the point of it being 'too late'

But now you cannot even enable it. `minishift add-on enable anyuid` will fail, because the default add-ons are not installed. They are now only installed once 'start' is called which is too late. 

> Then I would suggest to move this to m addon and the Prerun, but not as part of root.go

Not quite sure what you are saying. But the actual functionality to install/unpack the default add-ons is already in a shared utility package -  util.UnpackAddons. It's about where this is called. _PersistentPreRun_ is part of root.go and belongs to the root command. We discussed many times that this is the place where one needs to ensure the "correctness" of the Minishift setup. I guess we should have also added a comment, but even more that this I think we need to document the bigger architectural decisions. I propose to start using "Lightweight Architecture Decision Records" - https://www.thoughtworks.com/radar/techniques/lightweight-architecture-decision-records. Happy to get one started, but I guess this is a different issue.

gbraad:
> because the default add-ons are not installed. 

installed => unpacked... ah, check
missing test

gbraad:
>  I guess we should have also added a comment, but even more that this I think we need to document the bigger architectural decisions

+1

hferentschik:
> installed => unpacked... ah, check

Right.

Also as it stands now, you still have the unwanted message in case of a Minishift _update_ command. The issue is the message not that we unpacked the add-ons in the persistent pre-run.

Issue #1406, opened 2017-09-19
Line "Checking available disk space ... 0% OK" in minishift start log looks weird


amitkrout:
```
minishift version : 1.6.0
```
```
$ ./minishift start
-- Installing default add-ons ... OK
.
.
-- Checking available disk space ... 0% OK
-- Downloading OpenShift binary 'oc' version 'v3.6.0'
.
.
$
```

praveenkumar:
@amitkrout it's not weird, here we check the space of mounted drive so that we also know if disk is mounted as expected. cc @gbraad

gbraad:
It actually says: checking available space...
0% is used... so OK follows.

... as expected.


Suggestion on how to improve the message?

amitkrout:
@praveenkumar The line is ok, however the ```0%``` confuses me. As a user i would consider this as ```no disk space is available``` cc @gbraad

gbraad:
The whole message is
```
-- Checking if persistent storage volume is mounted ... OK
-- Checking available disk space ... 1% OK
```

Maybe we could include:
```
-- Checking available disk space ... 1% used OK
```

budhrg:
> -- Checking available disk space ... 0% OK

Agree with @amitkrout , as the wording is little confusing. One might think here that disk space is 0% available. 

+1 for changing the wording here. 

> Checking available disk space ... 1% used OK

Looks fine. WDYT @amitkrout ?

gbraad:
add easyfix tag

amitkrout:
> Checking available disk space ... 1% used OK

+1

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/pull/1422

PR #1422, opened 2017-09-21
Issue #1406 added clear wording for the disk usage check.

anjannath:


anjannath:
@gbraad Did the change, i was wandering though why the change? is it for more readable code?

cmd/minishift/cmd/start_preflight.go
```diff
@@ -311,8 +313,8 @@ func checkStorageMounted(driver drivers.Driver) bool {
 // space available.
 func checkStorageUsage(driver drivers.Driver) bool {
 	usedPercentage := getDiskUsage(driver, StorageDisk)
-	fmt.Printf("%s ", usedPercentage)
-	usedPercentage = strings.TrimRight(usedPercentage, "%")
+	fmt.Printf("%s used ", usedPercentage)
+	usedPercentage = stringUtils.GetOnlyNumbers(usedPercentage)
 	usage, err := strconv.ParseInt(usedPercentage, 10, 8)
```
gbraad: I think this can also be written as `strconv.ParseInt(stringUtils.GetOnlyNumbers(usedPercentage), 10, 0)`. because in that case we do not re-use the `userPercentage` variable.

gbraad:
>  It will not change the 0% to 1% as 0% comes from the df -h output.

Nort sure what you mean with this. Why does it need to change from `0%` into `1%` ???

cmd/minishift/cmd/start_preflight.go
anjannath: I too was thinking of that, but in the interest of not making that line too long (i am using a x230 thinkpad) used the variable again. will change it now.

LalatenduMohanty:
@gbraad I thought you wanted to show that https://github.com/minishift/minishift/issues/1406#issuecomment-330442212

gbraad:
@anjannath The intention is to make the code more robust, and promote re-use of earlier used functions.

The `GetOnlyNUmbers` functionality is tested as part of a unittest, while the `TrimRight("%")` is not.

gbraad:
@LalatenduMohanty you misread the example ;-) Anyways. just the inclusion of "used" or if inverted "free" would be needed.

gbraad:
add to whitelist

gbraad:
@anjannath you can always use a newline if needed

```go
usage, err := strconv.ParseInt(stringUtils.GetOnlyNumbers(usedPercentage), 10, 8)
```

or
```go
usage, err := strconv.ParseInt(
                 stringUtils.GetOnlyNumbers(usedPercentage), 10, 8)
```

are functionally the same.

In some cases it would be best to use extra newlines to make the code readable.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -311,9 +313,8 @@ func checkStorageMounted(driver drivers.Driver) bool {
 // space available.
 func checkStorageUsage(driver drivers.Driver) bool {
 	usedPercentage := getDiskUsage(driver, StorageDisk)
-	fmt.Printf("%s ", usedPercentage)
-	usedPercentage = strings.TrimRight(usedPercentage, "%")
-	usage, err := strconv.ParseInt(usedPercentage, 10, 8)
+	fmt.Printf("%s used ", usedPercentage)
+	usage, err := strconv.ParseInt(stringUtils.GetOnlyNumbers(usedPercentage), 10, 8)
```
LalatenduMohanty: Not related to this PR, but commenting because we are at it. 
@gbraad AFAICT we are not gaining anything by using int bit size as `8` in strconv.ParseInt("string", 10, `8`) because it returns int64 (for a 64bit system) anyway. So we are not saving any memory here.  So I will suggest we replace `ParseInt` with `Atoi` and it would make code simple by a tiny bit. WDYT?

gbraad: I think `strconv.Atoi(stringUtils.GetOnlyNumbers(usedPercentage))` doesn't read very nicely, but we can change...

Issue #1247, opened 2017-08-08
Extend status command to provide details about the VM


gbraad:
We should extend the `minishift status` command to show some basic information about the VM, besides the state. Some of this could be the IP address, to align with `minikube status`, but also `minishift status --check` could be introduced to do some of the `preflightChecksAfterStartingHost`, like IP reachability check, disk usage, or provide information about the RHSM subscription status for RHEL images.

gbraad:
Ref: https://github.com/kubernetes/minikube/blob/master/cmd/minikube/cmd/status.go#L81

hferentschik:
Duplicate of #373

hferentschik:
I'd also like to see a differentiation between VM and OpenShift state:

    $ minishift status
    VM: Running 
    OpenShift: [Running|Stopped|Unavailable]

gbraad:
+1 Good point

Can you explain the status Unavailable?

hferentschik:
I meant something like "Error" or "Unkown". But this was just a example. On second thought "Unavailable" could also mean "Not deployed". The VM is running, but no Origin container around.

gbraad:
Exactly, that is what I thought. There are several states possible and
definitely would like to explore them more.

Is there a way to extract the version info from the origin/OCP container?

hferentschik:
> Exactly, that is what I thought. There are several states possible and definitely would like to explore them more.

Right, the same way that there are several other states than 'Running' and 'Stopped' for the libmachine VM. We just tend to ignore them. 

> Is there a way to extract the version info from the origin/OCP container?

Yes, we already do that

gbraad:
> Yes, we already do that

haven‘t noticed.

hferentschik:
https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/openshift/version.go#L50

gbraad:
Ah, just a simple:

```
$ docker exec origin openshift version
```

We definitely need to use this...

gbraad:
```
$ minishift status
VM: Running
OpenShift: Running... v3.6.1
Storage: 10% of 20GB
```

LalatenduMohanty:
Moving to the next milestone because I think we should do it after #126 is fixed.

praveenkumar:
Resolved by #1425

PR #1425, opened 2017-09-21
Issue #1247 Extend status command to provide more details

praveenkumar:
Have to include storage piece also.

Current output
```
$ ./minishift status
VM: Running
OpenShift: Running (openshift v3.6.0+c4dd4cf)
```

### DIFF
--- a/cmd/minishift/cmd/status.go
+++ b/cmd/minishift/cmd/status.go
@@ -19,29 +19,82 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
+	"text/template"
 
 	"github.com/docker/machine/libmachine"
-	"github.com/spf13/cobra"
-
+	"github.com/docker/machine/libmachine/state"
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"
+	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
 )
+
+var statusFormat = `Minishift:  {{.MinishiftStatus}}
+Profile:    {{.ProfileName}}
+OpenShift:  {{.ClusterStatus}}
+DiskUsage:  {{.DiskUsage}}
+`
+
+type Status struct {
+	MinishiftStatus string
+	ProfileName     string
+	ClusterStatus   string
+	DiskUsage       string
+}
 
 // statusCmd represents the status command
 var statusCmd = &cobra.Command{
 	Use:   "status",
 	Short: "Gets the status of the local OpenShift cluster.",
 	Long:  `Gets the status of the local OpenShift cluster.`,
-	Run: func(cmd *cobra.Command, args []string) {
-		api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
-		defer api.Close()
+	Run:   runStatus,
+}
+
+func runStatus(cmd *cobra.Command, args []string) {
+	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+	defer api.Close()
+
+	host, err := api.Load(constants.MachineName)
+	if err != nil {
 		s, err := cluster.GetHostStatus(api)
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error getting cluster status: %s", err.Error()))
 		}
-		fmt.Fprintln(os.Stdout, s)
-	},
+		atexit.ExitWithMessage(0, s)
+	}
+
+	openshiftStatus := "Stopped"
+	diskUsage := "Unknown"
+	profileName := profileActions.GetActiveProfile()
+
+	vmStatus, err := cluster.GetHostStatus(api)
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting cluster status: %s", err.Error()))
+	}
+
+	if vmStatus == state.Running.String() {
+		openshiftVersion, err := openshiftVersion.GetOpenshiftVersion(host)
+		if err == nil {
+			openshiftStatus = fmt.Sprintf("Running (%s)", strings.Split(openshiftVersion, "\n")[0])
+		}
+
+		diskSize, diskUse := getDiskUsage(host.Driver, StorageDisk)
+		diskUsage = fmt.Sprintf("%s of %s", diskUse, diskSize)
+	}
+
+	status := Status{vmStatus, profileName, openshiftStatus, diskUsage}
+
+	tmpl, err := template.New("status").Parse(statusFormat)
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error creating status template: %s", err.Error()))
+	}
+	err = tmpl.Execute(os.Stdout, status)
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error executing status template:: %s", err.Error()))
+	}
 }
 
 func init() {
--- a/test/integration/util/util.go
+++ b/test/integration/util/util.go
@@ -102,19 +102,16 @@ func (m *MinishiftRunner) IsCDK() bool {
 	return strings.Contains(cmdOut, "minishift setup-cdk [flags]")
 }
 
-func (m *MinishiftRunner) EnsureRunning() {
-	if m.GetStatus() != "Running" {
-		m.Start()
-	}
-	m.CheckStatus("Running")
+func (m *MinishiftRunner) IsMinishiftRunning() bool {
+	return strings.Contains(m.GetStatus(), "Minishift:  Running")
 }
 
-func (m *MinishiftRunner) IsRunning() bool {
-	return m.GetStatus() == "Running"
+func (m *MinishiftRunner) IsOpenshiftRunning() bool {
+	return strings.Contains(m.GetStatus(), "OpenShift:  Running")
 }
 
 func (m *MinishiftRunner) GetOcRunner() *OcRunner {
-	if m.IsRunning() {
+	if m.IsMinishiftRunning() {
 		return NewOcRunner()
 	}
 	return nil
@@ -152,7 +149,7 @@ func (m *MinishiftRunner) GetStatus() string {
 }
 
 func (m *MinishiftRunner) CheckStatus(desired string) bool {
-	return m.GetStatus() == desired
+	return strings.Contains(m.GetStatus(), desired)
 }
 
 func NewOcRunner() *OcRunner {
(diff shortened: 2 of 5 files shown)


gbraad:
Somehow, it doesn't look pleasing. But...

```
$ ./minishift status
VM:        Running
OpenShift: Running (openshift v3.6.0+c4dd4cf)
```

still not. It is the wording VM. Also, what about the machinename? Now with the profiles, this becomes more relevant.

praveenkumar:
@gbraad Now it looks like below

```
Minishift: Running
OpenShift: Running (openshift v3.6.0+c4dd4cf)
Profile: minishift
DiskUsage: 7% of 27.3G
```

praveenkumar:
retest this please.

gbraad:
Can you use something like `w := tabwriter.NewWriter(os.Stdout, 4, 8, 3, ' ', 0)` to write the values? This way they align nicely

cmd/minishift/cmd/status.go
```diff
@@ -19,29 +19,84 @@ package cmd
...
+	util.ExitIfUndefined(api, constants.MachineName)
+
+	host, err := api.Load(constants.MachineName)
+	if err != nil {
+		atexit.ExitWithMessage(1, err.Error())
+	}
+
+	openshiftStatus := state.None.String()
+	profileName := state.None.String()
+	diskUsage := state.None.String()
```
gbraad: 👍

praveenkumar:
@gbraad does below what you think looks much better?

```
   MiniShift| Running
   OpenShift| Running (openshift v3.6.0+c4dd4cf)
     Profile| minishift
   DiskUsage| 7% of 27.3G
```

gbraad:
```
MiniShift:  Running
OpenShift:  Running (openshift v3.6.0+c4dd4cf)
Profile:    minishift
DiskUsage:  7% of 27.3G
```

praveenkumar:
retest this please

gbraad:
Test cases involving status need to be changed. The current form checks for a matching string

praveenkumar:
@gbraad As per logs looks like something else going on. Below test case should work right?

```   
 Scenario: User can install default add-ons                      # features/basic.feature:5
Given Minishift has state "Does Not Exist"                    # integration_test.go:152 -> *Minishift.shouldHaveState
08:17:51     Minishift state did not match. Expected: Does Not Exist, Actual: Running this command requires an existing 'minishift' VM, but no VM is defined.
```

gbraad:
Maybe something else is happening too (no VM defined feels like an issue with start or profile), but surely the status check will not work now.

test/integration/util/util.go
```diff
@@ -103,14 +103,14 @@ func (m *MinishiftRunner) IsCDK() bool {
 }
 
 func (m *MinishiftRunner) EnsureRunning() {
-	if m.GetStatus() != "Running" {
+	if !strings.Contains(m.GetStatus(), "Running") {
```
praveenkumar: This was the reason we had integration test cases were failing.

gbraad: What do we want to ensure running here? In this case it might be the VM, but OpenShift could be stopped or unavailable.

test/integration/util/util.go
```diff
@@ -102,15 +102,8 @@ func (m *MinishiftRunner) IsCDK() bool {
...
-func (m *MinishiftRunner) EnsureRunning() {
-	if m.GetStatus() != "Running" {
-		m.Start()
-	}
-	m.CheckStatus("Running")
-}
-
 func (m *MinishiftRunner) IsRunning() bool {
-	return m.GetStatus() == "Running"
+	return strings.Contains(m.GetStatus(), "Running")
```
gbraad: In this case the status can be checked as `contains/matches "Minishift: Running" AND contains/matches "OpenShift: Running"`, but I think our test is more about Minishft...

So maybe we should have two different functions for this... WDYT @agajdosi ?
  * `IsOpenShiftRunning`
  * `IsMinishiftRunning`

test/integration/util/util.go
```diff
@@ -152,7 +145,7 @@ func (m *MinishiftRunner) GetStatus() string {
 }
 
 func (m *MinishiftRunner) CheckStatus(desired string) bool {
-	return m.GetStatus() == desired
+	return strings.Contains(m.GetStatus(), desired)
```
gbraad: This is only about Minishift in the current tests... so name should be `CheckMinishiftStatus` and check for `contains/matches "Minishift: %s"`

test/integration/util/util.go
```diff
@@ -102,19 +102,12 @@ func (m *MinishiftRunner) IsCDK() bool {
...
-	if m.GetStatus() != "Running" {
-		m.Start()
-	}
-	m.CheckStatus("Running")
-}
-
-func (m *MinishiftRunner) IsRunning() bool {
-	return m.GetStatus() == "Running"
+func (m *MinishiftRunner) IsMinishiftRunning() bool {
+	return strings.Contains(m.GetStatus(), "Minishift:  Running")
```
praveenkumar: @gbraad @agajdosi Since right now we are only consuming  Minishift status so only add this function in future if we need `IsOpenshiftRunning` then it can be added.

test/integration/util/util.go
praveenkumar: @gbraad If you see this function is only used for `Does not Exist` when `EnsureDeleted` called and here "Minishift: %s"  is not comparable. If in future we need it some other place then we should pass `desired` state variable as `'minishift:  Running|stopped`

cmd/minishift/cmd/status.go
```diff
@@ -19,29 +19,84 @@ package cmd
...
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting cluster status: %s", err.Error()))
+	}
+
+	if vmStatus == state.Running.String() {
+		openshiftVersion, err := openshiftVersion.GetOpenshiftVersion(host)
+		openshiftStatus = "Stopped"
+		if err == nil {
+			openshiftStatus = fmt.Sprintf("Running (%s)", strings.Split(openshiftVersion, "\n")[0])
+		}
```
LalatenduMohanty: @praveenkumar Seems like we have taken a short cut to find out if the OpenShift is running or not. I would suggest we implement a function IsOpenshiftRunning.

test/integration/util/util.go
LalatenduMohanty: +1 for adding `IsOpenshiftRunning`.

cmd/minishift/cmd/status.go
budhrg: Agree :+1:

LalatenduMohanty:
I got following output when Minishift was stopped. 

```
$ minishift status
Minishift:  Stopped
OpenShift:  
Profile:    
DiskUsage:  
```
I would expect this as below.

```
$ minishift status
Minishift:  Stopped
OpenShift:  Stopped
Profile:  minishift
```

test/integration/util/util.go
gbraad: If also believe, as we already show the status, that `isOpenShiftRunning` should be implemented.

cmd/minishift/cmd/status.go
praveenkumar: @LalatenduMohanty It is not a sortcut if you check the implementation of this method it actually run the command on the origin container name. If think we should have a different method which tells openshift status then what docker command you have in mind? I still thinks this is clean.

praveenkumar:
> I would expect this as below.

@LalatenduMohanty @budhrg   Current implementation is same as minikube we are using statusFormat which is same for a stopped or running cluster I don't want to make it more complicated to add 2 different format for different things but what I can do is something below.

```
$ ./minishift status
Minishift:  Stopped
OpenShift:  minishift
Profile:    
DiskUsage:  
```

cmd/minishift/cmd/status.go
gbraad: @LalatenduMohanty 
```
docker@minishift:~$ docker exec origin openshift version
openshift v3.6.0+c4dd4cf
kubernetes v1.6.1+5115d708d7
etcd 3.2.1
docker@minishift:~$ docker stop origin
origin
docker@minishift:~$ docker exec origin openshift version
Error response from daemon: Container e31d14e98b38202279fdbd04f1211859674c3a7210bada2d6924ae216518746c is not running
```

test/integration/util/util.go
gbraad: ONLY if tests are available for this...

test/integration/util/util.go
gbraad: maybe a candidate to be removed...

test/integration/util/util.go
agajdosi: @LalatenduMohanty @gbraad isOpenShiftRunning() could have some use, but to apply it to features could take some time, so I would suggest to save @praveenkumar's time and create an issue for this topic and we can add them later. For this PR the isOpenShiftRunning() can be added, however I have no problem with adding them later by myself.

cmd/minishift/cmd/status.go
LalatenduMohanty: @praveenkumar @gbraad Either we should use `oc cluster status` to see if Openshift instance is running or not. Here is the cluster up implementation https://github.com/openshift/origin/blob/master/pkg/oc/bootstrap/docker/status.go

LalatenduMohanty: We can do it using `ocRunner` .  IMO we should be able to use the function to get the status without worrying about the VM state as the function will handle the same.

cmd/minishift/cmd/status.go
```diff
@@ -19,29 +19,84 @@ package cmd
...
+		profileName = profileActions.GetActiveProfile()
```
LalatenduMohanty: This should be outside of the if condition i.e. `if vmStatus == state.Running.String()`

LalatenduMohanty:
@praveenkumar The output you suggested looks good to me.

cmd/minishift/cmd/status.go
praveenkumar: @LalatenduMohanty There is a reason why we are using version because if I use `ocRunner` there is no way I can get running version of the cluster using `oc` commands. if you look the output of status command when running is look below.

```
MiniShift:  Running
OpenShift:  Running (openshift v3.6.0+c4dd4cf)  => here running openshift version coming from version fuctions.
Profile:    minishift
Storage:    used 7% of 27.3G
```

LalatenduMohanty: As discussed with @praveenkumar , I am ok with current implemented.

cmd/minishift/cmd/status.go
```diff
@@ -19,29 +19,83 @@ package cmd
...
+	profileName := profileActions.GetActiveProfile()
+
+	vmStatus, err := cluster.GetHostStatus(api)
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting cluster status: %s", err.Error()))
+	}
+
+	if vmStatus == state.Running.String() {
+		openshiftVersion, err := openshiftVersion.GetOpenshiftVersion(host)
+		openshiftStatus = "Stopped"
```
LalatenduMohanty: We have already initialized `openshiftStatus` as `stopped` before. We should remove this line.

budhrg: Agree :+1:

Issue #1432, opened 2017-09-22
Troubleshooting: Network not found: no network with matching name 'default'


kishansag:
While using minishift you might get errors like `Network not found: no network with matching name 'default'` or `Network not found: no network with matching name 'docker-machines'`. It would be great if we can add a fix to this to the troubleshooting document.

anjannath:
@gbraad I think we can close this as the preflight check for the `default` network is implemented now.

gbraad:
Documentation to add default network is not inc;luded on our end. I think this is necessary

PR #1403, opened 2017-09-18
Fix for error: Network not found: no network with matching name 'default'

kishansag:
Fix for error: Network not found: no network with matching name 'default'

Related to: https://github.com/kubernetes/minikube/issues/828

PR #1437, opened 2017-09-22
[WIP] Issue #1432 Network not found: no network with matching name 'default'

kishansagathiya:
Addresses #1432 :Troubleshooting: Network not found: no network with matching name 'default'
Referred https://github.com/minishift/minishift/pull/1403, https://github.com/kubernetes/minikube/issues/828, https://github.com/dhiltgen/docker-machine-kvm/issues/24#issuecomment-249050583

### DIFF
--- a/docs/source/troubleshooting/troubleshooting-driver-plugins.adoc
+++ b/docs/source/troubleshooting/troubleshooting-driver-plugins.adoc
@@ -134,6 +134,28 @@ $ sudo virsh undefine minishift
 
 In case all of this fails, you might want to xref:../getting-started/uninstalling.adoc#[uninstall Minishift] and do a fresh install of Minishift.
 
+[[Network-not-found-no-network-with-matching-name-default]]
+=== Network not found: no network with matching name 'default'
+
+At present, the KVM driver doesn't attempt to create the default network - it assumes it's already present. If you create a VM with another tool like virt-manager first it will create it (or you can create it from the command line.) Try the following to create the network
+
+[[using-command-line]]
+==== Using Commnad Line
+----
+virsh net-undefine default
+virsh net-destroy default
+virsh net-list # This will list all the networks
+virsh net-define /usr/share/libvirt/networks/default.xml
+virsh net-start default
+virsh net-autostart default
+----
+
+[[using-virt-manager]]
+==== Using virt-manager
+----
+virt-manager > edit > connection details > virtual networks > default
+----
+
 [[troubleshooting-driver-xhyve]]
 == xhyve
 


kishansag:
@praveenkumar

gbraad:
Actually, it should be written in a way devoid of the issue. We do not really care the driver does not implement this behaviour, but we care that the user would have to take steps when the default network is not found

We could mention something along the lines of "Since the driver plugin does not handle the default network creation, starting Minishift with KVM might result in a default network not found. To solve ..." ... perhaps you can use "When starting Minishift using the KVM driver plugin for the first time you might get a message .[[]]].. To solve ... [[]]. This happens due to the fact that the driver plugin does not handle the network creation" However I doubt the reason really needs to be included.

I'd rather refer to https://wiki.libvirt.org/page/Networking if needed than an issue number. Note: this is documentation that will be published on http://docs.openshift.org/latest/minishift

gbraad:
Changed description as this does not fix #1432. A possible solution is given in #1454.

docs/source/troubleshooting/troubleshooting-driver-plugins.adoc
```diff
@@ -134,6 +134,28 @@ $ sudo virsh undefine minishift
 
 In case all of this fails, you might want to xref:../getting-started/uninstalling.adoc#[uninstall Minishift] and do a fresh install of Minishift.
 
+[[Network-not-found-no-network-with-matching-name-default]]
+=== Network not found: no network with matching name 'default'
+
+At present, the KVM driver doesn't attempt to create the default network - it assumes it's already present. If you create a VM with another tool like virt-manager first it will create it (or you can create it from the command line.) Try the following to create the network
+
```
LalatenduMohanty: To run Minishift successfully on KVM, it needs the `default virtual network` provided by libvirt.The default virtual network usually gets created by automatically for a standard installation. However in case it is not present in the host you will see the discussed error. You can use the following commands to create the default network.

docs/source/troubleshooting/troubleshooting-driver-plugins.adoc
```diff
@@ -134,6 +134,28 @@ $ sudo virsh undefine minishift
...
+virsh net-start default
+virsh net-autostart default
+----
+
+[[using-virt-manager]]
+==== Using virt-manager
+----
+virt-manager > edit > connection details > virtual networks > default
+----
+
```
LalatenduMohanty: We might want to add that after that user can verify by running following command if the default network is created successfully or not. 
```
$ sudo virsh net-list --all
Name                 State      Autostart 
-----------------------------------------
default              active     yes
```

docs/source/troubleshooting/troubleshooting-driver-plugins.adoc
gbraad: "gets created by automatically for a standard installation." I do not think this is needed, as this makes people starting to guess WHY it didn't set it up. most likely as they didn't use virtmanager, etc. It creates ambiguity, so I would leave it out.

docs/source/troubleshooting/troubleshooting-driver-plugins.adoc
gbraad: Please, do not use `sudo` as this makes the command dependent of administrator/root privileges.

The usage should be:
```
$ virsh --connect qemu:///system net-list 
```

gbraad: We should be able to verify if the network exists without using `sudo`, unless we need to create it:

```
$ virsh --connect qemu:///system net-list 
```

gbraad:
add to whitelist

gbraad:
The fix #1454 to check for this network has been merged.

LalatenduMohanty:
@kishansagathiya Are you planning to update the PR or you want us to do the required modification and take it forward. Kindly let us know.

kishansagathiya:
@gbraad You can close this, no problem

LalatenduMohanty:
Closing the PR as we have the pre-flight check to see if default network is present or not.

Issue #1332, opened 2017-08-31
Add integration test for docker-env and oc-env commands


agajdosi:
We had some troubles with docker-env and oc-env commands few months ago. These are quite specific commands as they are bound to different shells on different platforms which can cause problems sometimes. To prevent this it would be nice to create an integration test for these commands.

I did a spike on possibilities for testing these commands. Right now I have an early prototype for Linux which serves as a proof of a concept. In short:

- I spawn a process with bash: `terminal = exec.Command("bash")` and I start it `terminal.Start()`
- I can send a stdin into this process with a pipeline: `inPipe, _ = terminal.StdinPipe()` - `inPipe.Write([]byte(command + "\n"))`
- to be able to check for an exit code on next line I add: `inPipe.Write([]byte("echo exitCode=$?\n"))`
- then I wait for exit code: `exitCode := <- exitCodeChannel`, this channel is from a separate goroutine which checks stdout pipe of "terminal *Cmd"

So one can do:
```gherkin
    Scenario: Executing output from docker-env
       Given user starts terminal instance
        When executing command "minishift $dockerenvstdout"
        When executing command "docker ps"
        When executing command "sleep 10"
```

where dockerenvstdout is stdout from `minishift docker-env`.

TODO:

- [x] implement `When setting scenario variable "dockerenvstdout" to the stdout from executing "minishift docker-env"`, currently it supports only oc commands

- [x] implement shell type recognition - bash, cmd, powershell, macos terminal...

- [x] add support for different shells - the commands for checking the exit code `echo $?`, `echo %errorvalue%` etc

LalatenduMohanty:
@agajdosi Good idea. 

> implement When setting scenario variable "dockerenvstdout" to the stdout from executing "minishift docker-env", currently it supports only oc commands

What do you mean by  `currently it supports only oc commands`?



> implement shell type recognition - bash, cmd, powershell, macos terminal...

I think to start with lets do it for bash. `bash` should also work fine for OS X

agajdosi:
@LalatenduMohanty 
> What do you mean by currently it supports only oc commands?

Currently it is implemented this way:

```go
s.Step(`^setting scenario variable "(.*)" to the stdout from executing "oc (.*)"$`,
		minishift.setVariableExecutingOcCommand)
```
[...]

Issue #1318, opened 2017-08-29
Integration test coverage for registry add-on


amitkrout:
This is a followup integration test coverage issue on pr #1243. Reference for creating the test scenario [#704 (comment)](https://github.com/minishift/minishift/pull/704#issuecomment-293568580)

agajdosi:
Depending on the size of this test we could also think of covering addons in general as part of this issue - focusing on all aspects of addons command + focusing on individual addons.

agajdosi:
Based on our conversation I am going to create a separate issue for adding a specialized test for addon command. It seems that with 4 default addons one feature file would get quite long and thus it will be better to not merge them together with broader addons command test. Also it would give us a good level of separation and better understanding of these feature files.

amitkrout:
For exposing docker registry, shell need to be configured as per the instructed output of ```minishift docker-env```. This scenario is implemented in the ongoing pr #1453. Some part of the code can be reused for this issue once the pr #1453 merge is done.

praveenkumar:
@amitkrout @agajdosi Does this still valid?

amitkrout:
@praveenkumar yes, will take this on next sprint.

amitkrout:
I have added a working registry-route feature file. I have commented some steps there which yet to be implemented/factored. To make this work i used the step ```And user waits "40" seconds``` (Not recommended). Later ```wait``` step will be removed once container status is implemented.

PR #1453, opened 2017-09-27
Issue #1332 Add integration tests for docker-env and oc-env

agajdosi:
Changes:
- adds ability to run commands in shell instance - for now just bash
- adds feature for docker-env command
- moves some check functions to util package

TBD:
- add support for other shells (shouldn't be hard)
- add feature for oc-env command (shouldn't be hard)

I will appreciate any feedback on this PR.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,6 +36,9 @@ TIMEOUT ?= 3600s
 PACKAGES := go list ./... | grep -v /vendor | grep -v /out
 SOURCE_DIRS = cmd pkg test
 
+# defines non-default shell name to be used in godog steps which use instance of shell
+TEST_WITH_SPECIFIED_SHELL ?=
+
 # Linker flags
 VERSION_VARIABLES := -X $(REPOPATH)/pkg/version.minishiftVersion=$(MINISHIFT_VERSION) \
 	-X $(REPOPATH)/pkg/version.b2dIsoVersion=$(B2D_ISO_VERSION) \
@@ -202,21 +205,21 @@ integration: GODOG_OPTS = --tags=basic
 integration: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
 	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
-	--run-before-feature="$(RUN_BEFORE_FEATURE)" $(GODOG_OPTS)
+	--run-before-feature="$(RUN_BEFORE_FEATURE)" --test-with-specified-shell="$(TEST_WITH_SPECIFIED_SHELL)" $(GODOG_OPTS)
 
 .PHONY: integration_all
 integration_all: GODOG_OPTS = --tags=~coolstore
 integration_all: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
 	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
-	--run-before-feature="$(RUN_BEFORE_FEATURE)" $(GODOG_OPTS)
+	--run-before-feature="$(RUN_BEFORE_FEATURE)" --test-with-specified-shell="$(TEST_WITH_SPECIFIED_SHELL)" $(GODOG_OPTS)
 
 .PHONY: integration_pr
 integration_pr: GODOG_OPTS = --tags=~coolstore\&\&~addon-xpaas
 integration_pr: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
 	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
-	--run-before-feature="$(RUN_BEFORE_FEATURE)" $(GODOG_OPTS)
+	--run-before-feature="$(RUN_BEFORE_FEATURE)" --test-with-specified-shell="$(TEST_WITH_SPECIFIED_SHELL)" $(GODOG_OPTS)
 
 .PHONY: fmt
 fmt:
--- a/docs/source/contributing/developing.adoc
+++ b/docs/source/contributing/developing.adoc
@@ -274,11 +274,28 @@ RUN_BEFORE_FEATURE::
 The `RUN_BEFORE_FEATURE` parameter specifies {project} commands to be run before each feature.
 This provides the ability to run integration tests against a {project} instance which is not in its default state.
 When multiple commands are specified, they must be delimited by a semicolon.
-For example, tests can be run against a stopped {project} instance with the image caching option turned on by running the following:
+For example, tests can be run against a stopped {project} instance with the image caching option turned on by running the following: 
 
 ----
 $ make integration_all RUN_BEFORE_FEATURE="start; stop; config set image-caching true"
 ----
+
+TEST_WITH_SHELL::
+
+About:::
+The majority of {project} integration tests pass command line arguments directly to the `minishift` or `oc` binaries via Go's `os/exec` package.
+However, some integration tests first start persistent instance of the shell to pass commands into later, for example: `When user starts shell instance on host machine`.
+Due to persistency of the shell instance all changes done in it will be present until its closure, for example, changed environmental variables.
+This allows testing of commands which are dependent on the shell they are being executed in.
+The steps using this approach have the `shell` keyword in their definition, for example: `When executing "minishift oc-env" in host shell succeeds`.
+
+Usage:::
+`TEST_WITH_SPECIFIED_SHELL` is an optional parameter.
+If nothing is passed with `TEST_WITH_SPECIFIED_SHELL`, then the shell type is automatically determined.
+By default, the shell will be PowerShell on Windows, Bash on macOS and Linux.
+The `TEST_WITH_SPECIFIED_SHELL` parameter can be used to run the tests with different shells.
+To run the tests using the Windows Command Prompt, for example, run `make integration_all TEST_WITH_SPECIFIED_SHELL=cmd`.
+The `TEST_WITH_SPECIFIED_SHELL` parameter supports the `powershell`, `cmd`, `bash`, `tcsh` and `zsh` options.
 
 [[godog-options]]
 ==== Using the GODOG_OPTS Parameter
--- a/test/integration/features/cmd-docker-env.feature
+++ b/test/integration/features/cmd-docker-env.feature
@@ -1,0 +1,38 @@
+@cmd-docker-env @command
+Feature: Command docker-env
+Command docker-env sets docker environment variables for supported shells.
+INFO: This feature runs against a shell instance. To use a non-default shell, please select
+one from: bash, cmd, powershell, tcsh, zsh with TEST_WITH_SPECIFIED_SHELL parameter of make integration.
+
+  Scenario: Docker client is not set up
+  INFO: This scenario starts interactive shell instance, which will be closed in the end of this feature.
+    Given user starts shell instance on host machine
+     When executing "docker info" in host shell
+     Then stdout of host shell should not contain "Operating System: Minishift Boot2Docker ISO Version"
+      And stdout of host shell should not contain "Name: minishift"
+
+  Scenario: Starting minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start" succeeds
+     Then Minishift has state "Running"
+
+  Scenario: Setting Docker client using docker-env command
+     When executing "minishift docker-env" in host shell succeeds
+      And evaluating stdout of the previous command in host shell
+     Then executing "docker info" in host shell succeeds
+      And stdout of host shell should contain "OSType: linux"
+      And stdout of host shell should contain "Name: minishift"
+
+  Scenario: Unsetting Docker client using --unset flag of docker-env command
+     When executing "minishift docker-env --unset" in host shell succeeds
+      And evaluating stdout of the previous command in host shell
+      And executing "docker info" in host shell
+     Then stdout of host shell should not contain "Operating System: Minishift Boot2Docker ISO Version"
+      And stdout of host shell should not contain "Name: minishift"
+
+  Scenario: Deleting Minishift
+  INFO: Removes the interactive shell instance.
+    Given user closes shell instance on host machine
+      And Minishift has state "Running"
+     When executing "minishift delete --force" succeeds
+     Then Minishift has state "Does Not Exist"
--- a/test/integration/features/cmd-oc-env.feature
+++ b/test/integration/features/cmd-oc-env.feature
@@ -1,0 +1,29 @@
+@cmd-oc-env @command
+Feature: Command oc-env
+Command oc-env sets the path to oc binary.
+INFO: This feature runs against a shell instance. To use a non-default shell, please select
+one from: bash, cmd, powershell, tcsh, zsh with TEST_WITH_SPECIFIED_SHELL parameter of make integration.
+
+  Scenario: User starts shell instance without oc in PATH
+  INFO: This scenario starts interactive shell instance, which will be closed in the end of this feature.
+    Given user starts shell instance on host machine
+     When executing "unset PATH" in host shell
+     Then executing "oc status" in host shell fails
+
+  Scenario: Starting minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start" succeeds
+     Then Minishift has state "Running"
+
+  Scenario: Setting oc binary to PATH with oc-env command
+     When executing "minishift oc-env" in host shell succeeds
+      And evaluating stdout of the previous command in host shell
+     Then executing "oc status" in host shell succeeds
+      And stdout of host shell contains "In project My Project (myproject)"
+
+  Scenario: Deleting Minishift
+  INFO: Removes the interactive shell instance.
+    Given user closes shell instance on host machine
+      And Minishift has state "Running"
+     When executing "minishift delete --force" succeeds
+     Then Minishift has state "Does Not Exist"
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -55,6 +55,7 @@ var (
 	isoName       string
 
 	runBeforeFeature string
+	testWithShell    string
 
 	// Godog options
 	godogFormat              string
@@ -120,6 +121,7 @@ func parseFlags() {
 	flag.StringVar(&minishiftArgs, "minishift-args", "", "Arguments to pass to minishift")
 	flag.StringVar(&minishiftBinary, "binary", "", "Path to minishift binary")
 	flag.StringVar(&runBeforeFeature, "run-before-feature", "", "Set of minishift commands to be executed before every feature. Individual commands must be delimited by a semicolon.")
+	flag.StringVar(&testWithShell, "test-with-specified-shell", "", "Name of shell to be used for steps which executes commands directly in persistent shell instance.")
 	flag.StringVar(&testDir, "test-dir", "", "Path to the directory in which to execute the tests")
 
 	flag.StringVar(&godogFormat, "format", "pretty", "Sets which format godog will use")
@@ -180,6 +182,8 @@ func FeatureContext(s *godog.Suite) {
 	// steps for scenario variables
 	s.Step(`^setting scenario variable "(.*)" to the stdout from executing "oc (.*)"$`,
 		minishift.setVariableExecutingOcCommand)
+	s.Step(`^setting scenario variable "(.*)" to the stdout from executing "minishift (.*)"$`,
+		minishift.setVariableExecutingMinishiftCommand)
 	s.Step(`^scenario variable "(.*)" should not be empty$`,
 		variableShouldNotBeEmpty)
 
@@ -252,6 +256,34 @@ func FeatureContext(s *godog.Suite) {
 	// steps for download of minishift-addons repository
 	s.Step(`^file from "(.*)" is downloaded into location "(.*)"$`,
 		downloadFileIntoLocation)
+
+	// steps for executing commands in shell
+	s.Step(`^user starts shell instance on host machine$`,
+		startHostShellInstance)
+	s.Step(`^user closes shell instance on host machine$`,
+		util.CloseHostShellInstance)
+	s.Step(`^executing "minishift (.*)" in host shell$`,
+		util.ExecuteMinishiftInHostShell)
+	s.Step(`^executing "minishift (.*)" in host shell (succeeds|fails)$`,
+		util.ExecuteMinishiftInHostShellSucceedsOrFails)
+	s.Step(`^executing "(.*)" in host shell$`,
+		util.ExecuteInHostShell)
+	s.Step(`^executing "(.*)" in host shell (succeeds|fails)$`,
+		util.ExecuteInHostShellSucceedsOrFails)
+	s.Step(`^(stdout|stderr) of host shell (?:should contain|contains) "(.*)"$`,
+		util.HostShellCommandReturnShouldContain)
+	s.Step(`^(stdout|stderr) of host shell (?:should not contain|does not contain) "(.*)"$`,
+		util.HostShellCommandReturnShouldNotContain)
+	s.Step(`^(stdout|stderr) of host shell (?:should contain|contains)$`,
+		util.HostShellCommandReturnShouldContainContent)
+	s.Step(`^(stdout|stderr) of host shell (?:should not contain|does not contain)$`,
+		util.HostShellCommandReturnShouldNotContainContent)
+	s.Step(`^(stdout|stderr) of host shell (?:should equal|equals) "(.*)"$`,
+		util.HostShellCommandReturnShouldEqual)
+	s.Step(`^(stdout|stderr) of host shell (?:should equal|equals)$`,
+		util.HostShellCommandReturnShouldEqualContent)
+	s.Step(`^evaluating stdout of the previous command in host shell$`,
+		util.ExecuteInHostShellLineByLine)
 
 	s.BeforeSuite(func() {
 		testDir = setUp()
@@ -435,7 +467,7 @@ func configFileContainsKeyMatchingValue(format string, configPath string, condit
 		return err
 	}
 
-	matches, err := performRegexMatch(expectedValue, keyValue)
+	matches, err := util.PerformRegexMatch(expectedValue, keyValue)
 	if err != nil {
 		return err
 	} else if (condition == "contains") && !matches {
@@ -475,7 +507,7 @@ func stdoutContainsKeyMatchingValue(commandField string, format string, conditio
 		return err
 	}
 
-	matches, err := performRegexMatch(expectedValue, keyValue)
+	matches, err := util.PerformRegexMatch(expectedValue, keyValue)
 	if err != nil {
 		return err
 	} else if (condition == "contains") && !matches {
@@ -499,69 +531,6 @@ func stdoutContainsKey(commandField string, format string, condition string, key
 		return fmt.Errorf("%s does not contain any value for key %s", commandField, keyPath)
 	} else if (condition == "does not have") && (keyValue != "<nil>") {
 		return fmt.Errorf("%s contains key %s with assigned value: %s", commandField, keyPath, keyValue)
-	}
-
-	return nil
-}
-
-func compareExpectedWithActualContains(expected string, actual string) error {
-	if !strings.Contains(actual, expected) {
-		return fmt.Errorf("Output did not match. Expected: '%s', Actual: '%s'", expected, actual)
-	}
-
-	return nil
-}
-
-func compareExpectedWithActualNotContains(notexpected string, actual string) error {
-	if strings.Contains(actual, notexpected) {
-		return fmt.Errorf("Output did match. Not expected: '%s', Actual: '%s'", notexpected, actual)
-	}
-
-	return nil
-}
-
-func compareExpectedWithActualEquals(expected string, actual string) error {
-	if actual != expected {
-		return fmt.Errorf("Output did not match. Expected: '%s', Actual: '%s'", expected, actual)
-	}
-
-	return nil
-}
-
-func compareExpectedWithActualNotEquals(notexpected string, actual string) error {
-	if actual == notexpected {
-		return fmt.Errorf("Output did match. Not expected: '%s', Actual: '%s'", notexpected, actual)
-	}
-
-	return nil
-}
-
-func performRegexMatch(regex string, input string) (bool, error) {
-	compRegex, err := regexp.Compile(regex)
-	if err != nil {
-		return false, fmt.Errorf("Expected value must be a valid regular expression statement: ", err)
-	}
-
-	return compRegex.MatchString(input), nil
-}
-
-func compareExpectedWithActualMatchesRegex(expected string, actual string) error {
-	matches, err := performRegexMatch(expected, actual)
-	if err != nil {
-		return err
-	} else if !matches {
-		return fmt.Errorf("Output did not match. Expected: '%s', Actual: '%s'", expected, actual)
-	}
-
-	return nil
-}
-
-func compareExpectedWithActualNotMatchesRegex(notexpected string, actual string) error {
-	matches, err := performRegexMatch(notexpected, actual)
-	if err != nil {
-		return err
-	} else if matches {
-		return fmt.Errorf("Output did match. Not expected: '%s', Actual: '%s'", notexpected, actual)
 	}
 
 	return nil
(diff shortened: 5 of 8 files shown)


gbraad:
I originally had added a bashrunner, but decided against this later... as this is not cross-platform.

agajdosi:
@gbraad It is not yet able to run on Windows or macOS, but from beginning I planned for this to be supported. Right now I start an instance of bash https://github.com/minishift/minishift/pull/1453/files#diff-760d6c0d4b85a60586b257589346a757R60 and then send commands to it via pipe. The only platform dependent part (if I am not totally missing something) should be the start of shell instance - if bash, cmd, powershell or mac terminal and then the command which checks for exit code of last command.

amitkrout:
#1318

hferentschik:
What's the status on this one?

budhrg:
@agajdosi , an important change is coming in this PR https://github.com/minishift/minishift/pull/1681 to remove docker api version from docker-env.

agajdosi:
@hferentschik The work is stopped for this sprint, will try to do it in next one.

@budhrg Thank you for information, will later update it so it will be fine with new implementation.

praveenkumar:
@hferentschik any idea why on this PR there is no WIP CI which stop for merge? Or it is because this was created long before then WIP hook introduced?

agajdosi:
@praveenkumar Yes, might be due it is old, will try to remove and add the [WIP].

agajdosi:
@praveenkumar Fixed now :)

hferentschik:
> Yes, might be due it is old, will try to remove and add the [WIP].

+1 Or if you know it will take a while, close the pull request and re-open when you have something ready.

agajdosi:
@hferentschik I have started working on this, so will keep it, hopefully I will be able to move it closer to merge soon.

hferentschik:
np

agajdosi:
So current status is that bash, tcsh and zsh shells are working just fine. Also Windows' cmd is fine now after escaping `%` in `%errorvalue%`, making it `%%errorvalue%%`. Before pinging anybody I will also clean the code, currently it contains too much dirt used for debugging.

Powershell had problem - the shell got stuck and did not react on the input pipe from Go, however I have found out that starting it with `powershell -Command -` do the job. Need to implement it tomorrow.

However the fish shell is still not working and I am out of ideas why is this happening.

agajdosi:
Both cmd and powershell now runs fine, the last problem is that in step
```
When executing "minishift docker-env" succeeds
```
the shell is not being selected (it is executed via golang exec package, question is why on Linux this works great...). However the proper way is to transport it into:

```
When executing in shell "minishift docker-env" succeeds
```

But before that I need to set the path to minishift properly in initialization of shell instance, will implement it tomorrow.

agajdosi:
Fixed `errorlevel` and `$lastexitcode` checks - on Windows these will be ignored for the step "executing line by line from last stdout." Added step `When executing in shell command "minishift docker-env"`, this will replace the minishift with absolute path so the tested binary and not something from user's path will be used.

Needed: Executing line by line from last stdout needs to get the stdout from shell, not overall golang exec.

agajdosi:
Code cleaned, now writing the cmd-oc-env.feature, which is last thing needed here.

agajdosi:
Code cleaned, now writing the cmd-oc-env.feature, which is last thing needed here.

agajdosi:
Have tested on cmd, powershell, bash, zsh, tcsh all works (tcsh does not pass one step due to #1743). Fish is not supported, was not able to implement it. Code was cleaned, thus I am removing the WIP.

agajdosi:
@minishift/minishift-dev If somebody has some extra time, please take a look on this PR.

Makefile
```diff
@@ -200,21 +200,21 @@ integration: GODOG_OPTS = --tags=basic
 integration: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
 	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
-	--run-before-feature="$(RUN_BEFORE_FEATURE)" $(GODOG_OPTS)
+	--run-before-feature="$(RUN_BEFORE_FEATURE)" --test-with-shell=$(TEST_WITH_SHELL) $(GODOG_OPTS)
```
LalatenduMohanty: @agajdosi `$(TEST_WITH_SHELL)` where is this coming from? I do not see how we are deciding the shell for the Makefile.

hferentschik: what's the default?

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,41 @@
+@cmd-docker-env @command
+Feature: Command docker-env
+Command docker-env helps the user to set environmental variables which are needed force
+the usage of Minishift with docker command on VM host.
+INFO: This feature runs against persistent shell instance. To use non-default shell, please
```
hferentschik: "persistent shell instance" - persistent?

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,41 @@
...
+select one from: bash, cmd, powershell, tcsh, zsh with TEST_WITH_BASH parameter of make integration.
+
+  Scenario: Docker client is not set up
+  INFO: This scenario starts interactive shell instance, which will be closed in the end of this feature.
+    Given user starts shell instance
+     When executing in shell command "docker info"
+     Then stdout of shell should not contain "Operating System: Minishift Boot2Docker ISO Version"
```
hferentschik: so what should it contain then? Seems to be odd to test what should _not_ be there. Why do you expect this to be there?

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,41 @@
...
+      And stdout of shell should not contain "Name: minishift"
+
+  Scenario: Starting minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start" succeeds
+     Then Minishift has state "Running"
+
+  Scenario: Setting Docker client using docker-env command
+     When executing in shell command "minishift docker-env" succeeds
```
hferentschik: is 'command' not superfluous here and "When executing in shell  "minishift docker-env" succeeds" easier to read?

Makefile
agajdosi: @hferentschik @LalatenduMohanty default is empty string, if empty string is being detected in shell.go, then the test will use bash if test runs on linux or mac, or powershell if the OS is windows, pretty much it imitates what Minishift uses.

When one uses TEST_WITH_SHELL, then the test can be forced to run with: cmd, powershell, bash, tcsh, zsh (fish has problems and the program will inform you about it and will run the bash).

test/integration/util/shell.go
```diff
@@ -0,0 +1,278 @@
...
+	bashExitCodeCheck       = "echo %v$?"
+	fishExitCodeCheck       = "echo %v$status"
+	tcshExitCodeCheck       = "echo %v$?"
+	zshExitCodeCheck        = "echo %v$?"
+	cmdExitCodeCheck        = "echo %v%%errorlevel%%"
+	powershellExitCodeCheck = "echo %v$lastexitcode"
+)
+
+var (
+	shellStartArgument    []string
```
hferentschik: I think it would make sense to introduce a struct for the shell related things here

test/integration/util/shell.go
```diff
@@ -0,0 +1,278 @@
...
+			shellStartArgument = []string{"-Command", "-"}
+			shellCheckExitCodeCmd = fmt.Sprintf(powershellExitCodeCheck, exitCodeIdentifier)
+		}
+	}
+
+	return
+}
+
+func StartShellInstance() error {
+	var err error
```
hferentschik: it would be nicer to encapsulate the shell functionality into to struct

test/integration/features/cmd-docker-env.feature
agajdosi: @hferentschik +1 much cleaner, will implement it, thanks

Makefile
hferentschik: what's about the docs around this?

agajdosi: @hferentschik As this is quite specific I have documented it directly in feature files (https://github.com/minishift/minishift/issues/1332), but I can add it into docs also.

test/integration/features/cmd-docker-env.feature
agajdosi: @hferentschik Wanted to highlight the fact that environmental variables and other modifications done will persist until the moment the shell instance is closed, but sure it can be deleted, wdyt?

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,41 @@
+@cmd-docker-env @command
+Feature: Command docker-env
+Command docker-env helps the user to set environmental variables which are needed force
+the usage of Minishift with docker command on VM host.
```
LalatenduMohanty: This explanation of this feature file is not clear to me. Suggestion:  command `docker-env` sets docker environment variables for supported shells.`

test/integration/features/cmd-docker-env.feature
LalatenduMohanty: @agajdosi use of `persistent shell instance` is confusing to me too.

test/integration/features/cmd-oc-env.feature
```diff
@@ -0,0 +1,31 @@
+@cmd-oc-env @command
+Feature: Command oc-env
+Command oc-env helps the user to set path to oc binary so she can run the oc command within the VM host.
```
LalatenduMohanty: This is not correct. The `oc` command does nor run with the Minishift VM. The command only helps to set the path to the oc binary so that user can use the oc cli against the OpenShift instance from command line.

test/integration/util/shell.go
agajdosi: @hferentschik Will take a look on that and implement it. Thanks.

test/integration/features/cmd-oc-env.feature
```diff
@@ -0,0 +1,31 @@
...
+INFO: This feature runs against persistent shell instance. To use non-default shell, please
+select one from: bash, cmd, powershell, tcsh, zsh with TEST_WITH_BASH parameter of make integration.
+
+  Scenario: User starts shell instance without oc in PATH
+  INFO: This scenario starts interactive shell instance, which
+  will be closed in the end of this feature.
+    Given user starts shell instance
+     When executing in shell command "unset PATH"
+     Then executing in shell command "oc version" fails
```
LalatenduMohanty: I feel we should use `oc status` instead of `oc version` as it actually shows the current OpenShift namespace/project and the cluster IP etc in the output

test/integration/features/cmd-docker-env.feature
agajdosi: ok, will remove, thanks 👍

test/integration/features/cmd-docker-env.feature
agajdosi: @hferentschik this is here to prevent potential situation when the one who runs it has the separate Minishift instance running and has Docker preconfigured via .bashrc, might then the docker info pass? Tried to be defensive here.

test/integration/features/cmd-docker-env.feature
agajdosi: @LalatenduMohanty Thanks, that sentence is confused, missed that, your suggestion seems fine to me.

test/integration/features/cmd-oc-env.feature
agajdosi: @LalatenduMohanty it talks about the usage of oc command withing the VM host, not the Minishift VM aka guest. Is it fine, or should I make it more clear somehow?

test/integration/features/cmd-oc-env.feature
agajdosi: @LalatenduMohanty Will take a look on that, thanks for suggestion.

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,41 @@
...
+
+  Scenario: Unsetting Docker client using --unset flag of docker-env command
+     When executing in shell command "minishift docker-env --unset" succeeds
+      And executing in shell shell's stdout of previous command line by line
+      And executing in shell command "docker info"
+     Then stdout of shell should not contain "Operating System: Minishift Boot2Docker ISO Version"
+      And stdout of shell should not contain "Name: minishift"
+
+  Scenario: Closing shell instance
+  INFO: Removes the interactive shell instance.
```
budhrg: I doubt `Given` or `When` is missing here.

agajdosi:
Moved variables into a struct.

@hferentschik sorry for asking, I am doing the package with struct design for the first time so I better try to clear my intentions... Question is, how to identify which functions should I transform to a method? Any quick tip for how Minishift handles with this?

Currently mine understanding is that all functions which are currently a private/non-exported functions and work with the struct, should be a method. But is it ok to keep the exported/public functions as they are? 

Thank you for the help :)

test/integration/features/cmd-docker-env.feature
agajdosi: @budhrg this step is actually not a test, but a job which needs to be done here, but I agree it looks a little bit strange, I could move it into the scenario where Minishift instance is being deleted. WDYT?

budhrg: Sure

Makefile
hferentschik: Well, given that it is exposed in the Makefile I would expect to get some form of information. Maybe a comment in the Makefile. The rub is that if I read the Makefile I see '$(TEST_WITH_SHELL)'. I then notice that there is no value for that at all, which slowly get my alarm bells ringing. I guess @LalatenduMohanty was in a similar boat. 

For what it's worth, the same issue is with 'RUN_BEFORE_FEATURE'. It feels wrong to always see '--run-before-feature' and '--test-with-shell', effectively knowing that these values are empty. The expectation from using command line switches is that one does not specify them, if there is no value one wants to pass. People react to that when reading the Makefile.

test/integration/util/shell.go
```diff
@@ -0,0 +1,282 @@
...
+			shell.checkExitCodeCmd = fmt.Sprintf(powershellExitCodeCheck, exitCodeIdentifier)
+		}
+	}
+
+	return
+}
+
+func StartShellInstance() error {
+	var err error
+	shell.stdoutChannel = make(chan string)
```
hferentschik: You really should have a constructor which creates the instance. Now you create an empty `shell shellInstance` as a var and keep setting various bits and pieces. That's marginally better. However, the idea behind a struct is that you control its initalization and only make available what needs to be made available. Also you guarantee that if you are dealing with an instance of shellInstance it is in a consistent state. See also http://www.golangpatterns.info/object-oriented/constructors

note constructors do not really exist in Golang. We are talking factory methods really. Check our codebase as well. There you see it at play at several places as well.

hferentschik:
> sorry for asking, I am doing the package with struct design for the first time so I better try to clear my intentions... 

no problem at all :-)

> Question is, how to identify which functions should I transform to a method? Any quick tip for how Minishift handles with this?

Basically, all which operate on the state of the struct. Think of it this way. Right now you have 16! variables which define your state of the shell. One needs to be very careful how to operate on this, since if you forget to set/unset something, the whole thing might blow up, right? The idea of the struct is that you create a single point to create the instance of the struct. In this constructor / factory method you ensure that the overall state of the struct is correct. Then you add methods which operate on the state in a controlled manner. You don't want outside code be able to change any of the state bypassing these methods. Hence all struct members should be kept private. Does that make sense?

agajdosi:
@hferentschik It is full of sense 👍 Thank you for it. Back on track now.

Also going back to the constructor comment of yours, the one problematic part is that via godog step I can't pass anything else than the matched words. I would like to ideally construct the shell struct once the step `user starts shell instance` is executed, but I can't pass shellName and path to minishiftbinary through it. So probably will move this step definition from shell.go to integration_test.go then it all should be fine.

One more thank, will ping you once the WIP is ended.

Makefile
agajdosi: @hferentschik I also feel that this solution we use already for some time is a bit strange. One simple solution would be to make these values empty with quotes, so it would be at least more clear that it is passing nothing:
```
go test -timeout 3600s [...] --run-before-feature="" --test-with-shell="" [] -tags=cmd-docker-env,cmd-oc-env
```

another possibility would be to have just one parameter, for example `TEST_OPTS` to which one would need to pass the all the flags she wants to use, probably also mixed with GODOG_OPTS, so we would just se the TEST_OPTS:

```
make integration TEST_OPTS="--test-with-shell=bash --tags=cmd-docker-env --run-before-feature='minishift config set memory 4GB;minishift config set cpus 4'"
```

WDYT?

agajdosi: However for the second solution I would say it is better to create a follow up issue.

docs/source/contributing/developing.adoc
```diff
@@ -274,12 +274,26 @@ RUN_BEFORE_FEATURE::
...
 This provides the ability to run integration tests against a {project} instance which is not in its default state.
 When multiple commands are specified, they must be delimited by a semicolon.
-For example, tests can be run against a stopped {project} instance with the image caching option turned on by running the following:
+For example, tests can be run against a stopped {project} instance with the image caching option turned on by running the following: 
 
 ----
 $ make integration_all RUN_BEFORE_FEATURE="start; stop; config set image-caching true"
 ----
 
+TEST_WITH_SHELL::
```
agajdosi: Hi, @robin-owen, I am adding in this PR some small documentation changes, could you please take a look and review them? Thanks :)

agajdosi:
@hferentschik I have moved most of the parts into a struct with private methods, all is made accessible via public functions. The `"constructor"` concept was not implemented as smallest function in which I could call it would be whole suite, so I keep it as variable in package, however there is just one moment of setting-and-starting it and one moment for its closing.

docs/source/contributing/developing.adoc
```diff
@@ -274,12 +274,26 @@ RUN_BEFORE_FEATURE::
...
+
+About:::
+Majority of integration tests in Minishift passes command arguments directly to `minishift` or `oc` binaries with usage of Go's `os/exec` package.
+However some integration tests first starts an instance of shell, into which the tests later passes the commands.
+This attitude allows to test `minishift docker-env` or `minishift oc-env` commands, which are dependent on which kind of shell they are being executed in.
+The steps using this approach have the `shell` keyword in their definition, for example: `When executing in shell "minishift oc-env" succeeds`.
+
+Usage:::
+`TEST_WITH_SHELL` is an optional parameter.
+If nothing is passed with `TEST_WITH_SHELL`, then the shell type is being detected automatically: for Windows the `PowerShell`, while on macOS and Linux the `bash` will be used.
```
robin-owen:
```adoc
If nothing is passed with `TEST_WITH_SHELL`, then the shell type is automatically determined.
By default, the shell will be PowerShell on Windows, Bash on macOS and Linux.
```

docs/source/contributing/developing.adoc
```diff
@@ -274,12 +274,26 @@ RUN_BEFORE_FEATURE::
...
+To run the tests with different shells, TEST_WITH_SHELL parameter can be used, for example run: `minishift integration_all TEST_WITH_SHELL=cmd` to execute the tests with Windows' `Command Prompt`.
+The `powershell`, `cmd`, `bash`, `tcsh` and `zsh` are the options currently supported for `TEST_WITH_SHELL` parameter.
```
robin-owen:
```adoc
The `TEST_WITH_SHELL` parameter supports the `powershell`, `cmd`, `bash`, `tcsh` and `zsh` options.
```

docs/source/contributing/developing.adoc
```diff
@@ -274,12 +274,26 @@ RUN_BEFORE_FEATURE::
...
+About:::
+Majority of integration tests in Minishift passes command arguments directly to `minishift` or `oc` binaries with usage of Go's `os/exec` package.
+However some integration tests first starts an instance of shell, into which the tests later passes the commands.
+This attitude allows to test `minishift docker-env` or `minishift oc-env` commands, which are dependent on which kind of shell they are being executed in.
+The steps using this approach have the `shell` keyword in their definition, for example: `When executing in shell "minishift oc-env" succeeds`.
+
+Usage:::
+`TEST_WITH_SHELL` is an optional parameter.
+If nothing is passed with `TEST_WITH_SHELL`, then the shell type is being detected automatically: for Windows the `PowerShell`, while on macOS and Linux the `bash` will be used.
+To run the tests with different shells, TEST_WITH_SHELL parameter can be used, for example run: `minishift integration_all TEST_WITH_SHELL=cmd` to execute the tests with Windows' `Command Prompt`.
```
robin-owen: Note: I assume that `minishift integration_all` was a typo, and `make integration_all` was intended.

```adoc
The `TEST_WITH_SHELL` parameter can be used to run the tests with different shells.
To run the tests using the Windows Command Prompt, for example, run `make integration_all TEST_WITH_SHELL=cmd`.
```

docs/source/contributing/developing.adoc
```diff
@@ -274,12 +274,26 @@ RUN_BEFORE_FEATURE::
...
 
 ----
 $ make integration_all RUN_BEFORE_FEATURE="start; stop; config set image-caching true"
 ----
 
+TEST_WITH_SHELL::
+
+About:::
+Majority of integration tests in Minishift passes command arguments directly to `minishift` or `oc` binaries with usage of Go's `os/exec` package.
+However some integration tests first starts an instance of shell, into which the tests later passes the commands.
```
robin-owen:
```adoc
However, some integration tests first start an instance of the shell to pass commands into later.
```

docs/source/contributing/developing.adoc
```diff
@@ -274,12 +274,26 @@ RUN_BEFORE_FEATURE::
...
+This attitude allows to test `minishift docker-env` or `minishift oc-env` commands, which are dependent on which kind of shell they are being executed in.
```
robin-owen:
```adoc
This allows testing of commands which are dependent on the shell they are being executed in.
```

I don't think we need to explicitly state which commands are dependent on the shell. If that's necessary, though, I would add this information to the end of the sentence in a parenthetical:

```(such as `minishift docker-env` and `minishift oc-env`)```

Comment (agajdosi) [reply docs/source/contributing/developing.adoc]: Will keep the shorter variant without mentioning the exact commands, thank you.

File: docs/source/contributing/developing.adoc [hunk 18]
```diff
@@ -274,12 +274,26 @@ RUN_BEFORE_FEATURE::
...
+For example, tests can be run against a stopped {project} instance with the image caching option turned on by running the following: 
 
 ----
 $ make integration_all RUN_BEFORE_FEATURE="start; stop; config set image-caching true"
 ----
 
+TEST_WITH_SHELL::
+
+About:::
+Majority of integration tests in Minishift passes command arguments directly to `minishift` or `oc` binaries with usage of Go's `os/exec` package.
```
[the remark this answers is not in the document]
agajdosi: Aah, missed the `{project}`. Must get used to it..

agajdosi:
@robin-owen Suggestions were implemented. The text reads much better now, I like the shortened sentences and skipping some unnecessary information, thank you for fast reaction and help.

agajdosi:
getting some attention 😉 @minishift/minishift-dev

gbraad:
still WIP or ready?

agajdosi:
@gbraad Ready. I have just rebased now.

Makefile
```diff
@@ -36,6 +36,9 @@ TIMEOUT ?= 3600s
 PACKAGES := go list ./... | grep -v /vendor | grep -v /out
 SOURCE_DIRS = cmd pkg test
 
+# defines non-default shell name to be used in godog steps which use instance of shell
+TEST_WITH_SHELL ?= default
```
gbraad: Is this value "default" needed? or does an empty string also work?

If so, it would have to be named `TEST_WITH_SPECIFIED_SHELL` or `OVERRIDE`, `OTHER`, `ALTERNATIVE`, etc... to indicate it is not using the default.

agajdosi: @gbraad Both variants works, originally I used an empty string but then I have added default to make it easier to understand. Not sure whether this was a good move.

Please let me know which solution you would like to see and I will change it accordingly.

gbraad: The current way it is coded feels like this variable is a necessity. maybe it is the naming, maybe it is the value...

agajdosi: @gbraad Ok, will use `TEST_WITH_SPECIFIED_SHELL` which will be without specifying anything just empty string.

agajdosi:
@LalatenduMohanty this PR is ready for review, if you have spare time

docs/source/contributing/developing.adoc
```diff
@@ -274,12 +274,28 @@ RUN_BEFORE_FEATURE::
...
 ----
 $ make integration_all RUN_BEFORE_FEATURE="start; stop; config set image-caching true"
 ----
 
+TEST_WITH_SHELL::
+
+About:::
+The majority of {project} integration tests pass command line arguments directly to the `minishift` or `oc` binaries via Go's `os/exec` package.
+However, some integration tests first start an instance of the shell to pass commands into later.
+This allows testing of commands which are dependent on the shell they are being executed in.
```
gbraad: Is a single shell session maintained, or is this started on each command invocation?

If re-invoked, it means the environment variables are being reset. Right?
This would not be preferred, and in this case would need documentation (actual issue could be fixed in a later PR).

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,38 @@
+@cmd-docker-env @command
+Feature: Command docker-env
+Command docker-env sets docker environment variables for supported shells.
+INFO: This feature runs against shell instance. To use non-default shell, please select
```
gbraad: s/against/against a
s/use/use a

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,38 @@
...
+one from: bash, cmd, powershell, tcsh, zsh with TEST_WITH_SHELL parameter of make integration.
+
+  Scenario: Docker client is not set up
+  INFO: This scenario starts interactive shell instance, which will be closed in the end of this feature.
+    Given user starts shell instance
+     When executing in shell "docker info"
```
gbraad: I think we need to be more specific
```cucumber
   When user start local shell instance
```

to prevent the confusion of where the command `docker info` is executed.
Especially because of "Name: minishift" in the same scenario, it can confuse

gbraad: or sth like `host`, as:

```cucumber
  When executing in local shell "docker info"
```

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,38 @@
...
+  Scenario: Setting Docker client using docker-env command
+     When executing in shell "minishift docker-env" succeeds
+      And executing in shell shell's stdout of previous command line by line
+     Then executing in shell "docker info" succeeds
+      And stdout of shell should contain "OSType: linux"
+      And stdout of shell should contain "Name: minishift"
+
+  Scenario: Unsetting Docker client using --unset flag of docker-env command
+     When executing in shell "minishift docker-env --unset" succeeds
+      And executing in shell shell's stdout of previous command line by line
```
gbraad: ?????

test/integration/features/cmd-oc-env.feature
```diff
@@ -0,0 +1,29 @@
+@cmd-oc-env @command
+Feature: Command oc-env
+Command oc-env sets the path to oc binary.
+INFO: This feature runs against shell instance. To use non-default shell, please select
```
gbraad: ditto

test/integration/features/cmd-oc-env.feature
```diff
@@ -0,0 +1,29 @@
...
+one from: bash, cmd, powershell, tcsh, zsh with TEST_WITH_SHELL parameter of make integration.
```
gbraad: TEST_WITH_SHELL has been changed

test/integration/integration_test.go
```diff
@@ -253,6 +257,34 @@ func FeatureContext(s *godog.Suite) {
...
+	s.Step(`^(stdout|stderr) of shell (?:should contain|contains)$`,
+		util.ShellCommandReturnShouldContainContent)
+	s.Step(`^(stdout|stderr) of shell (?:should not contain|does not contain)$`,
+		util.ShellCommandReturnShouldNotContainContent)
+	s.Step(`^(stdout|stderr) of shell (?:should equal|equals) "(.*)"$`,
+		util.ShellCommandReturnShouldEqual)
+	s.Step(`^(stdout|stderr) of shell (?:should equal|equals)$`,
+		util.ShellCommandReturnShouldEqualContent)
+	s.Step(`^executing in shell shell's stdout of previous command line by line$`,
+		util.ExecuteInShellLineByLine)
```
gbraad: curious about this...

test/integration/util/shell.go
```diff
@@ -0,0 +1,298 @@
...
+	bashExitCodeCheck       = "echo %v$?"
+	fishExitCodeCheck       = "echo %v$status"
+	tcshExitCodeCheck       = "echo %v$?"
+	zshExitCodeCheck        = "echo %v$?"
+	cmdExitCodeCheck        = "echo %v%%errorlevel%%"
+	powershellExitCodeCheck = "echo %v$lastexitcode"
+)
+
+var (
+	shell shellInstance
```
gbraad: OK, shell is maintained from start till exit.

docs/source/contributing/developing.adoc
gbraad: Shell session is maintained. This should be clarified in the doucmentation, as this is important as it means setting variables and and oneliners would be possible.

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,38 @@
...
+      And stdout of shell should not contain "Name: minishift"
+
+  Scenario: Starting minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start" succeeds
+     Then Minishift has state "Running"
+
+  Scenario: Setting Docker client using docker-env command
+     When executing in shell "minishift docker-env" succeeds
+      And executing in shell shell's stdout of previous command line by line
```
gbraad: this does the eval of `docker-env`... can we also call it something like `eval` instead?

test/integration/util/shell.go
```diff
@@ -0,0 +1,298 @@
...
+		err = fmt.Errorf("Command '%s', expected to succeed, exited with exit code: %s\n", command, exitCode)
+	}
+	if expectedResult == "fails" && exitCode == "0" {
+		err = fmt.Errorf("Command '%s', expected to fail, exited with exit code: %s\n", command, exitCode)
+	}
+
+	return err
+}
+
+func ExecuteInShellLineByLine() error {
```
gbraad: The function name is clearer than the gherkin step.

test/integration/features/cmd-oc-env.feature
agajdosi: @gbraad good catch, missed this, thanks

test/integration/util/shell.go
agajdosi: @gbraad Unfortunately it is, however the shell instance is being created and removed during that time.

test/integration/features/cmd-docker-env.feature
agajdosi: @gbraad calling it eval is shell-centric I think, I have tried to describe the actual process, however I agree that this sentence is a little bit complicated.

test/integration/features/cmd-docker-env.feature
agajdosi: @gbraad Originally I used "persistent shell" to highlight that in contrary to `executing "minishift/oc <command> <subcommand>"` the changes will remain - for example the exports.

I am fine with mentioning it as `local shell`.

docs/source/contributing/developing.adoc
agajdosi: @gbraad Thought the concept and implementation of `user starts shell instance` -> `executing "docker info" in shell` -> `user closes shell instance` will talk for itself. But I agree that if there is this part, it could mention it explicitly. WDYT about:

```
About:::
The majority of {project} integration tests pass command line arguments directly to the `minishift` or `oc` binaries via Go's `os/exec` package.
However, some integration tests first start persistent instance of the shell to pass commands into later, for example: `When user starts shell instance on host machine`.
Due to persistency of the shell instance all changes done in it will be present until its closure, for example, changed environmental variables.
This allows testing of commands which are dependent on the shell they are being executed in.
The steps using this approach have the `shell` keyword in their definition, for example: `When executing "minishift oc-env" in host shell succeeds`.
```

test/integration/features/cmd-docker-env.feature
```diff
@@ -0,0 +1,38 @@
...
+      And stdout of host shell should not contain "Name: minishift"
+
+  Scenario: Starting minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start" succeeds
+     Then Minishift has state "Running"
+
+  Scenario: Setting Docker client using docker-env command
+     When executing "minishift docker-env" in host shell succeeds
+      And executing in shell shell's stdout of previous command line by line
```
gbraad: `And evaluating stdout of  the previous command in local shell`

test/integration/util/shell.go
gbraad: why unfortunately... this is actually what you want to happen...

Issue #1461, opened 2017-09-29
Move integration.log from integration-test directory


agajdosi:
Having integration.log in integration-test directory causes problems in downstream. It will be better to move it from integration-test dir as this could cause further problems in future. The new location can be probably just `out/integration.log`.

amitkrout:
It would be nice if the time stamp is appended to the log file something like ```integration_TismeStamp.log```. This will be very help full to keep record of all integartion test result rather overwriting the log file on each run.

agajdosi:
This is needed to be a quick fix, so I am not going to add/change any not necessary functionality, but I agree: it would be nice. Please create a separate issue where we can discuss how others see this.

agajdosi:
Merged via #1463.

PR #1463, opened 2017-09-29
Issue #1461 Move integration.log from integration-test directory

agajdosi:


test/integration/integration_test.go
```diff
@@ -245,7 +246,13 @@ func FeatureContext(s *godog.Suite) {
 
 	s.BeforeSuite(func() {
 		testDir = setUp()
-		util.StartLog(testDir)
+		testResultDir = filepath.Join(testDir, "..", "test-results")
```
praveenkumar: That means now it not going to make that `test-results` in the testDir but one more level up.

agajdosi: @praveenkumar Yes, it is one level up. Now content of `out` directory is:
```
bindata
integration-test
test-results
```

test/integration/integration_test.go
```diff
@@ -50,8 +50,9 @@ var (
 	minishiftArgs   string
 	minishiftBinary string
 
-	testDir string
-	isoName string
+	testDir       string
+	testResultDir string
```
praveenkumar: 👍

Issue #1460, opened 2017-09-29
AllInstanceConfig should be created in PersistentPreRun


LalatenduMohanty:
As part of profiles feature we are creating the AllInstanceConfig file in init() of root.go [1]. which is wrong in some way as ideally it should be done in PersistentPreRun to keep it consistent. Hence we are getting https://issues.jboss.org/browse/CDK-180
[1] https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/root.go#L217

gbraad:
Not sure why this would triggers a "delete"...
```
$ ./minishift setup-cdk
Setting up CDK 3 on host using '/home/amit/.minishift' as Minishift's home directory
The MINISHIFT_HOME directory '/home/amit/.minishift' exists. Continuing will delete any existing VM and all other data in this directory. Do you want to continue? [y/N]
```

LalatenduMohanty:
@gbraad It is because the directory is created before setup-cdk is run and if setup-cdk sees .minishift it will ask confirmation to overwrite it.

LalatenduMohanty:
Resolved by #1465

PR #1465, opened 2017-09-29
Issue #1460 Move all instance config creation to PersistentPreRun 

LalatenduMohanty:
…tPreRun

Fix issue #1460

### DIFF
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -82,6 +82,15 @@ var RootCmd = &cobra.Command{
 		// creating all directories for minishift run
 		createMinishiftDirs(minishiftConfig.InstanceDirs)
 
+		// If AllInstanceConfig is not defined we should define it now.
+		if minishiftConfig.AllInstancesConfig == nil {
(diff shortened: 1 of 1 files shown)


cmd/minishift/cmd/root.go
```diff
@@ -213,14 +223,11 @@ func initializeProfile() string {
 		}
 	}
 
-	// We need to initialize allinstance config as the active profile information present in it.
-	ensureAllInstanceConfigPath(constants.AllInstanceConfigPath)
 	minishiftConfig.AllInstancesConfig, err = minishiftConfig.NewAllInstancesConfig(constants.AllInstanceConfigPath)
-	if err != nil {
```
gbraad: what happened to the error?

n that case we just ignore the fact the config could not be created? This could result in issues with hostfolders which relies on the fact that the configuration file is available.

Has this been tested?

LalatenduMohanty: @gbraad I have a better idea to  fix this. Yes, this is tested. But I will update it with a better solution.

Issue #1474, opened 2017-10-03
command "minishift openshift version" generates a blank new line


amitkrout:
```
Enviornment : OS X
minishift master 
```
Current output : 
```
$ ./minishift openshift version
openshift v3.6.0+c4dd4cf
kubernetes v1.6.1+5115d708d7
etcd 3.2.1

$
```
Expected output : 
```
$ ./minishift openshift version
openshift v3.6.0+c4dd4cf
kubernetes v1.6.1+5115d708d7
etcd 3.2.1
$
```

budhrg:
Resolved via https://github.com/minishift/minishift/pull/1478

PR #1478, opened 2017-10-03
Issue #1474 Remove blank line from openshift version

praveenkumar:
version output coming from DockerCommander already have a new line  so used `Fprintf` instead `Fprintln`.

### DIFF
--- a/cmd/minishift/cmd/openshift/version.go
+++ b/cmd/minishift/cmd/openshift/version.go
@@ -48,7 +48,7 @@ func runVersion(cmd *cobra.Command, args []string) {
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting the OpenShift cluster version: %s", err.Error()))
 	}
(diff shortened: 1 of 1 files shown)


cmd/minishift/cmd/openshift/version.go
```diff
@@ -48,7 +48,7 @@ func runVersion(cmd *cobra.Command, args []string) {
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting the OpenShift cluster version: %s", err.Error()))
 	}
-	fmt.Fprintln(os.Stdout, version)
+	fmt.Fprintf(os.Stdout, version)
```
budhrg: Should be `fmt.Fprint` since we are not using any format specifier

praveenkumar: @budhrg right, updated.

budhrg:
Merged and thanks.

Issue #1134, opened 2017-07-14
Add a minimum Openshift version for executing add-ons


jorgemoralespou:
OpenShift 3.6 will introduce a big change on the APIs where now are using APIGroups, that means that the resource version definitions of many things, like templates created for 3.6+ will no longer work in previous releases. All those addons and templates that will be created with this API need to guarantee that the version of OpenShift is at a minimum.

Add an annotation to the addon with similar syntax:

````
# Required-OpenShift-Version 3.6+
````

Example of such an addon can be found [here](https://github.com/jorgemoralespou/minishift-addons/tree/master/prometheus)

hferentschik:
Thanks for the heads up. So much for OpenShift staying backward compatible ;-)

hferentschik:
Merged via pull request #1487

PR #1487, opened 2017-10-04
Issue #1134 Add a minimum Openshift version for executing add-ons

praveenkumar:


praveenkumar:
How to test
========

```
$ minishift addons install --defaults

$ cat ~/.minishift/addons/admin-user/admin-user.addon
# Name: admin-user
# OpenShift-Version: v3.7.0+

oc create user admin --full-name=admin
oc adm policy add-cluster-role-to-user cluster-admin admin

$ minishift apply admin-user
-- Applying addon 'admin-user':Error applying the add-on: Addon does not support OpenShift version v3.6.0. You need to use a version v3.7.0+

$ cat ~/.minishift/addons/admin-user/admin-user.addon
# Name: admin-user
# OpenShift-Version: v3.6.0+

oc create user admin --full-name=admin
oc adm policy add-cluster-role-to-user cluster-admin admin

$ minishift addon apply admin-user
-- Applying addon 'admin-user':..
```

praveenkumar:
@jorgemoralespou can you give it a try now and provide some early feedback.

Artifact location: http://artifacts.ci.centos.org/minishift/minishift/pr/1487/

docs/source/using/addons.adoc
```diff
@@ -26,13 +26,15 @@ The following example shows the contents of an add-on file, including the name a
 # Name: anyuid                                                                          // <1>
 # Description: Allows authenticated users to run images under a non pre-allocated UID   // <2>
 # Required-vars: ACME_TOKEN                                                             // <3>
+# OpenShift-Version: v3.6.0+                                                            // <4>
```
hferentschik: What is the syntax here? Exact version match or supporting something like >v3.6.0 and <v3.6.0 as well? Is the '+' in v3.6.0+ intentional and what does it mean? Is + the same as >?

docs/source/using/addons.adoc
```diff
@@ -26,13 +26,15 @@ The following example shows the contents of an add-on file, including the name a
...
 
-oc adm policy add-scc-to-group anyuid system:authenticated                              // <4>
+oc adm policy add-scc-to-group anyuid system:authenticated                              // <5>
 ----
 <1> (Required) Name of the add-on.
 <2> (Required) Description of the add-on.
 <3> (Optional) Comma separated list of required interpolation variables. See xref:../using/addons.adoc#addon-variable-interpolation[Variable Interpolation]
-<4> Actual add-on command. In this case, the command executes the *oc* binary.
+<4> (Optional) If provided then it will compare with runing OpenShift version in minishift and throw an error in case add-on require an higher version.
```
hferentschik: And what is the syntax? Exact matches only or is there a range semantic as well. How does it look like? I think there is more needed than just this short reference. There is a dedicated section needed for this as well.

pkg/minishift/addon/addon_meta.go
```diff
@@ -25,9 +25,10 @@ import (
 var requiredMetaTags = []string{NameMetaTagName, DescriptionMetaTagName}
 
 const (
-	requiredVars           = "Required-Vars"
-	NameMetaTagName        = "Name"
-	DescriptionMetaTagName = "Description"
+	requiredVars             = "Required-Vars"
+	NameMetaTagName          = "Name"
+	DescriptionMetaTagName   = "Description"
+	RequiredOpenShiftVersion = "OpenShift-Version"
```

pkg/minishift/addon/addon_meta.go
```diff
@@ -89,6 +91,13 @@ func (meta *DefaultAddOnMeta) GetValue(key string) string {
 	return meta.headers[key].(string)
 }
 
+func (meta *DefaultAddOnMeta) OpenShiftVersion() string {
+	if val, contains := meta.headers[RequiredOpenShiftVersion].(string); contains {
+		return val
+	}
+	return ""
```
hferentschik: So empty string is any version? Maybe make this a constant 'any_version = ""'?

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -272,6 +284,32 @@ func (m *AddOnManager) verifyRequiredVariablesInContext(context *command.Executi
...
+	dockerCommander := context.GetDockerCommander()
+	versionInfo, err := dockerCommander.Exec(" ", constants.OpenshiftContainerName, "openshift", "version")
+	if err != nil {
+		return err
+	}
+	openShiftVersionAlongWithCommitSha := strings.Split(strings.Split(versionInfo, "\n")[0], " ")[1]
+	openShiftVersion := strings.Split(openShiftVersionAlongWithCommitSha, "+")[0]
+	requiredOpenshiftVersion := meta.OpenShiftVersion()
+	if requiredOpenshiftVersion != "" {
+		if strings.HasSuffix(requiredOpenshiftVersion, "+") {
```
hferentschik: What is the whole '+' business about? You seem to trim it off here. I am not quite following.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -272,6 +284,32 @@ func (m *AddOnManager) verifyRequiredVariablesInContext(context *command.Executi
...
+			// This will work for both upstream and downstream.
```
hferentschik: Does it, what's about tests?

pkg/minishift/addon/addon_meta.go
gbraad: Should we allow empty sting to mean all? What if I just left it out by accident? Empty should maybe fail? Or ...

Default is all now?

docs/source/using/addons.adoc
LalatenduMohanty: I have same questions as @hferentschik.

hferentschik: I see two options (and it depends really on how we think this is going to be used). 

*  Using this meta tag means always >= the specified version. In this case, we should probably name the tag 'Min-OpenShift-Versoin'. 
* Have some form of version range semantic aka allow the use of <> or maybe even things like ~. But this needs to be documented

docs/source/using/addons.adoc
```diff
@@ -75,6 +77,18 @@ This can be useful in cases where you know that a command such as `oc` might tak
 
 NOTE: Trying to use an undefined command will cause an error when the add-on gets parsed.
 
+[[addon-openshift-version-semantics]]
+== OpenShift Version Semantics
+
+Minishift allows the use of OpenShift version within add-on metadata.
```
hferentschik: "Within an add-on, one can enforce the OpenShift version which needs to be running in order to apply the add-on." To do so one can specify the optional _OpenShift-Version_ metadata field. The semantics are as follow:

docs/source/using/addons.adoc
```diff
@@ -75,6 +77,18 @@ This can be useful in cases where you know that a command such as `oc` might tak
...
+It use below semantic and compare with running OpenShift cluster in Minishift.
+
+----
+- v3.6.0+     : Add-on can run on OpenShift version 3.6.0 or higher
```
hferentschik: OpenShift version 3.6.0 or higher needs to be running in order to apply the add-on.

docs/source/using/addons.adoc
```diff
@@ -75,6 +77,18 @@ This can be useful in cases where you know that a command such as `oc` might tak
...
+- v3.6.0      : Add-on can only run on OpenShift version 3.6.0
```
hferentschik: OpenShift version 3.6.0 needs to be running in order to apply the add-on.

docs/source/using/addons.adoc
hferentschik: Why '+'? On what are you basing this syntax? Is when it comes to version ranges not something like > and >= more common? It is also less confusing when thinking about our actual Minishift versions. We have something like 'minishift v1.6.0+2b005fb', meaning we use the '+' to separate the sha part. I initially thought the '+' was some sort of left over from you cutting of the sha.

hferentschik: What's about things like 3 or 3.6? Is this supported? Implementation wise I think it might be.

praveenkumar:
@hferentschik @LalatenduMohanty I already updated all the bits now as per our discussion.

docs/source/using/addons.adoc
```diff
@@ -75,6 +77,22 @@ This can be useful in cases where you know that a command such as `oc` might tak
...
+== OpenShift Version Semantics
+
+Within an add-on, one can enforce the OpenShift version which needs to be running in order to apply the add-on.
+To do so one can specify the optional OpenShift-Version metadata field. The semantics are as follow:
+
+----
+- >v3.6.0     : OpenShift version higher than 3.6.0 needs to be running in order to apply the add-on.
+- >=v3.6.0    : OpenShift version 3.6.0 or higher needs to be running in order to apply the add-on.
+- v3.6.0      : OpenShift version 3.6.0 needs to be running in order to apply the add-on.
+----
```
LalatenduMohanty: 👍

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -272,6 +285,51 @@ func (m *AddOnManager) verifyRequiredVariablesInContext(context *command.Executi
...
+			return nil
+		}
+		if !checkOpenShiftVersionSemantic(strings.TrimPrefix(requiredOpenshiftVersion, ">")) {
+			return versionErr
+		}
+		if openShiftVersion != requiredOpenshiftVersion {
+			return errors.New(fmt.Sprintf("Addon does not support OpenShift version %s. "+
+				"You need to use a version %s\n", openShiftVersion, requiredOpenshiftVersion))
+		}
+	}
```
LalatenduMohanty: Modified the add-on to use `OpenShift-Version: >v3.6.0`
```
$ minishift addon apply admin-user
-- Applying addon 'admin-user':Error applying the add-on: Addon does not support OpenShift version v3.6.0. You need to use a version >v3.6.0
```
Modified the add-on to use `OpenShift-Version: <v3.6.0`
```
[lmohanty@LalatenduM-laptop minishift]$ minishift addon apply admin-user
-- Applying addon 'admin-user':Error applying the add-on: Add-on only support OpenShift version as <major>.<minor>.<patch> (ex. v3.6.0) Provided <v3.6.0
```
@praveenkumar can we please simplify the error and make it consistent.  For example the last error should be `Addon does not support OpenShift version v3.6.0. You need to use a version <v3.6.0.  Add-on only support OpenShift version as <major>.<minor>.<patch> (ex. v3.6.0)`  or something better.

jorgemoralespou:
What happens with a version like 3.6.0-alpha.1? Would it work with >=3.6.0? 

Also, even though I find the notation clear, it only allows you to apply a minimum version. Sometimes a max version could also be desired. For this, I find very useful the maven range notation, which is mathematical notation and very easy to understand.

[3.6.0,)
[3.6.0,3.7.0)
[3.6.0,3.7.0]
(,3.6.0]

https://dev.to/ebud7/maven-version-ranges

LalatenduMohanty:
> Sometimes a max version could also be desired.

I think with current implementation it should be `=< maximum version`. Somehow I like the current notation (may be because I am not a maven user).  @minishift/minishift-dev WDYT?

hferentschik:
> I think with current implementation it should be =< maximum version.

+1

I don't think we should use the Maven notation here. Even though I also find it intuitive, I don't know any other tool which adopted this type of notation. I'd rather orient ourselves at  tools like Bundler, dip, Glide, etc.  If we want to allow rangers it would be something like:

    OpenShift-Version: >=3.6.0, < 3.7

I find this notation as clear and I think more suitable for the average user who might not be familiar with the mathematical range notation.

That said, I think we should drop the 'v' prefix when specifying versions

jorgemoralespou:
Ok, as long as you can express the same kinf of minor, minor equal, mayor, mayor equal, then it's fine.

Still, what would be the notation to enable 3.6.0-alpha.1? 
>= 3.6.0 
> 3.5.0
> 3.5.*

praveenkumar:
> Still, what would be the notation to enable 3.6.0-alpha.1?

@jorgemoralespou Currently we are not supporting alpha/beta/rc flags because then it will add inconsistency to the downstream side where we don't follow semantic versioning. 

Assume you are running minishift with v3.7.0 alpha and make sure addon should work only for 3.7.0 version then having `>=v3.7.0` should be enough.

hferentschik:
> Currently we are not supporting alpha/beta/rc flags because then it will add inconsistency to the downstream side where we don't follow semantic versioning.

Impl wise we use ordinal comparison, right? There is no distinction between community and OCP when comparing versions, right? 

If we want to support 'alpha/beta/rc', we would have to sue the semver library for comparison and it currently fails when it is used with OCP versions. In this case we would either need a switch where er compare community version via semver and OCP via ordinal comparison. Alternatively we use semver and in the case of OCP versions we consider only major.minor.version.

> Assume you are running minishift with v3.7.0 alpha and make sure addon should work only for 3.7.0 version then having >=v3.7.0 should be enough.

@jorgemoralespou how important do you think is it to also include 'alpha/beta/rc'?

docs/source/using/addons.adoc
```diff
@@ -25,14 +25,16 @@ The following example shows the contents of an add-on file, including the name a
 ----
 # Name: anyuid                                                                          // <1>
 # Description: Allows authenticated users to run images under a non pre-allocated UID   // <2>
-# Required-vars: ACME_TOKEN                                                             // <3>
+# Required-Vars: ACME_TOKEN                                                             // <3>
+# OpenShift-Version: >v3.6.0                                                            // <4>
```
hferentschik: we should drop the 'v' in the example

docs/source/using/addons.adoc
```diff
@@ -75,6 +77,22 @@ This can be useful in cases where you know that a command such as `oc` might tak
 
 NOTE: Trying to use an undefined command will cause an error when the add-on gets parsed.
 
+[[addon-openshift-version-semantics]]
+== OpenShift Version Semantics
+
+Within an add-on, one can enforce the OpenShift version which needs to be running in order to apply the add-on.
```
hferentschik: "As part of the add-on metadata you can specify the OpenShift version which needs to be running in order to apply the add-on."

docs/source/using/addons.adoc
```diff
@@ -75,6 +77,22 @@ This can be useful in cases where you know that a command such as `oc` might tak
...
+To do so one can specify the optional OpenShift-Version metadata field. The semantics are as follow:
```
hferentschik: you can

docs/source/using/addons.adoc
```diff
@@ -75,6 +77,22 @@ This can be useful in cases where you know that a command such as `oc` might tak
...
+
+----
+- >3.6.0     : OpenShift version higher than 3.6.0 needs to be running in order to apply the add-on.
+- >=3.6.0    : OpenShift version 3.6.0 or higher needs to be running in order to apply the add-on.
+- 3.6.0      : OpenShift version 3.6.0 needs to be running in order to apply the add-on.
```
hferentschik: Missing docs for discussed <= and range

praveenkumar:
retest this please.

praveenkumar:
@jorgemoralespou So now we have implemented a way to provide the range of openshift version supported for an add-on. Can you give it a try now?

- https://github.com/praveenkumar/minishift/blob/935fb7525954747b36b09f9ba527cffdc46e8533/docs/source/using/addons.adoc#addon-openshift-version-semantics
- http://artifacts.ci.centos.org/minishift/minishift/pr/1487/

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -266,12 +280,38 @@ func (m *AddOnManager) verifyRequiredVariablesInContext(context *command.Executi
...
 	return nil
 }
 
+func (m *AddOnManager) verifyRequiredOpenshiftVersion(context *command.ExecutionContext, meta addon.AddOnMeta) error {
+	dockerCommander := context.GetDockerCommander()
+	versionInfo, err := dockerCommander.Exec(" ", constants.OpenshiftContainerName, "openshift", "version")
+	if err != nil {
+		return err
+	}
+	openShiftVersionAlongWithCommitSha := strings.Split(strings.Split(versionInfo, "\n")[0], " ")[1]
```
hferentschik: These lines are a bit of black magic to me. Would a regexp not be easier than all the splits? Some comment on what the format is you are going to process would be nice as well. I think I'd extract the lines which extract the OpenShift version into a helper function which can easily be tested and documents the expected inputs as a side effect. I am afraid this code is fragfile and hard to maintain.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -266,12 +280,38 @@ func (m *AddOnManager) verifyRequiredVariablesInContext(context *command.Executi
...
+	openShiftVersion := strings.TrimSpace(strings.TrimPrefix(strings.Split(openShiftVersionAlongWithCommitSha, "+")[0], minikubeConstants.VersionPrefix))
+
+	requiredOpenshiftVersions := strings.TrimSpace(meta.OpenShiftVersion())
+	if requiredOpenshiftVersions != "" {
+		versionErr := fmt.Errorf("\nAdd-on only support listed OpenShift version semantics https://docs.openshift.org/latest/minishift/using/addons.html#addon-openshift-version-semantics "+
```
hferentschik: I don't think adding URLs like this is a good idea. This becomes easily out of data.

hferentschik: why do you create the error here, even though you might not need it?

pkg/minishift/addon/addon_meta.go
```diff
@@ -36,6 +38,7 @@ type AddOnMeta interface {
 	Description() []string
 	RequiredVars() []string
 	GetValue(key string) string
+	OpenShiftVersion() string
```
hferentschik: Instead of a string why don't you create a struct for it? One field containing the version and on a range. This could make things a bit easier down the road. Also, you could detect the wrong usage of the version and range when the add-on is parsed, compared to when it is applied.

pkg/minishift/addon/addon_meta.go
```diff
@@ -89,6 +92,13 @@ func (meta *DefaultAddOnMeta) GetValue(key string) string {
 	return meta.headers[key].(string)
 }
 
+func (meta *DefaultAddOnMeta) OpenShiftVersion() string {
```
hferentschik: Could return a struct instead of a single string. The content of the struct could be the version together with a range

pkg/minishift/addon/manager/addon_manager_test.go
```diff
@@ -131,3 +131,28 @@ func Test_invalid_addons_get_skipped(t *testing.T) {
...
 	}
 }
+
+func Test_check_openshift_version_sematic(t *testing.T) {
+	var versionTestData = []struct {
+		OpenshiftVersion string
+		expectedResult   bool
+	}{
+		{"3.6.0", true},
+		{"3.4", false},
```
hferentschik: Why is this false?

pkg/minishift/addon/manager/addon_manager_test.go
```diff
@@ -131,3 +131,28 @@ func Test_invalid_addons_get_skipped(t *testing.T) {
...
+		{"<3.6.0", true},
+		{"3.7.1.1", false},
+		{"3.5.3=", false},
+		{">=3.5.0, <=3.7.0", true},
+		{">3.5.0, <=3.7.0", true},
+		{"3.5.0, <=3.7.0", true},
+		{">=3.5.0, <=3.7", false},
```
hferentschik: why is this false?

hferentschik:
@jorgemoralespou what do you think about only supporting major.minor.patch? This makes the whole feature a bit more portable between Minishift and CDK, but I am undecided on whether we should go further.

praveenkumar:
@hferentschik made changes as per our discussion yesterday.

pkg/minishift/addon/addon_meta.go
```diff
@@ -60,6 +64,11 @@ func NewAddOnMeta(headers map[string]interface{}) (AddOnMeta, error) {
 			return nil, nil
 		}
 	}
+	if headers[RequiredOpenShiftVersion] != nil {
+		if !checkVersionSemantic(headers[RequiredOpenShiftVersion].(string)) {
+			return nil, errors.New("Add-on only support OpenShift version semantics eg. 3.6.0 or >3.6.0, <3.9.0 or >=3.5 etc.")
```
praveenkumar: @hferentschik Does that error message is good enough for the user instead of providing document link? we can reword it if you think something else should be added.

pkg/minishift/addon/addon_meta.go
```diff
@@ -99,3 +115,10 @@ func (meta *DefaultAddOnMeta) splitAndTrim(s string) []string {
 	}
 	return cleanSplit
 }
+
+func checkVersionSemantic(version string) bool {
+	// Strict match for <major> or <major>.<minor> or <major>.<minor>.<patch>
+	// (>=|>|<|<=)3.6.0, (>=|>|<|<=)3.6.0
+	match, _ := regexp.MatchString("^(|>|>=|<|<=)[0-9](.[0-9]){0,2}(|\\s*,\\s*(|>|>=|<|<=)[0-9](.[0-9]){0,2})$", version)
```
hferentschik: So does this not only work for single digit versions, eg 3.6.0? You only match a digit one [0-9]. What if we start getting 3.10.0 or 3.7.10 or something like this? Should it not be [0-9]+?

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -266,9 +279,37 @@ func (m *AddOnManager) verifyRequiredVariablesInContext(context *command.Executi
...
+}
+
+func (m *AddOnManager) verifyRequiredOpenshiftVersion(context *command.ExecutionContext, meta addon.AddOnMeta) error {
+	dockerCommander := context.GetDockerCommander()
+	versionInfo, err := dockerCommander.Exec(" ", constants.OpenshiftContainerName, "openshift", "version")
+	if err != nil {
+		return err
 	}
 
+	// verionInfo variable have below string as value along with new line
```
hferentschik: ok, works for me. I still argue a regexp using a pattern buffer would be easier: 'openshift v([0-9]+\.[0-9]+\.[0-9]+\.)+.*'

And then you get the version from the first pattern buffer. 
But if you want to go ahead with string splitting, that's fine with me

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -266,9 +279,37 @@ func (m *AddOnManager) verifyRequiredVariablesInContext(context *command.Executi
...
+	// kubernetes v1.6.1+5115d708d7
+	// etcd 3.2.1
+	// openShiftVersionAlongWithCommitSha is contain *v3.6.0+c4dd4cf* (first split on new line and second on space)
+	openShiftVersionAlongWithCommitSha := strings.Split(strings.Split(versionInfo, "\n")[0], " ")[1]
+	// openshiftVersion is contain *3.6.0* (split on *+* string and then trim the *v* as perfix)
+	// TrimSpace is there to make sure no whitespace around version string
+	openShiftVersion := strings.TrimSpace(strings.TrimPrefix(strings.Split(openShiftVersionAlongWithCommitSha, "+")[0], minikubeConstants.VersionPrefix))
+	requiredOpenshiftVersions := strings.TrimSpace(meta.OpenShiftVersion())
+	if requiredOpenshiftVersions != "" {
+		for _, requiredOpenshiftVersion := range strings.Split(requiredOpenshiftVersions, ",") {
```
hferentschik: Maybe we can introduce ',' as constant - VersionRangeSeperator? Just to make the code more explicit

pkg/minishift/addon/manager/addon_manager.go
praveenkumar: @hferentschik I mostly find regex bit hard to understand (my personal opinion) :)

pkg/minishift/addon/manager/addon_manager.go
praveenkumar: @hferentschik Agree, will do 👍

pkg/minishift/addon/addon_meta.go
praveenkumar: @hferentschik right, I somehow considered only one digit :(

hferentschik:
Rebased and merged (with minor change to docs)

PR #1489, opened 2017-10-04
Issue #1342 Moving the default add-on installation to PersistentPreRun

LalatenduMohanty:

### DIFF
--- a/test/integration/features/cmd-addons.feature
+++ b/test/integration/features/cmd-addons.feature
@@ -1,19 +1,9 @@
 @cmd-addons @command
 Feature: Addons command and its subcommands
 
-  Scenario: Command "addons" has "adddon" alias
-    Given Minishift has state "Does Not Exist"
-     When executing "minishift addons" succeeds
-     Then executing "minishift addon" succeeds
-      And executing "minishift add-ons" fails
-      And executing "minishift add-on" fails
-
- Scenario: Installing default add-ons manually
-     When executing "minishift addons install --defaults" succeeds
-     Then stdout should contain
-      """
-      Default add-ons anyuid, admin-user, xpaas, registry-route installed
-      """
+  @minishift-only
+  Scenario: Default add-ons are installed but not enabled by default
+     Note: default addons were installed during previous commands automatically.
      When executing "minishift addons list"
      Then stdout should match "admin-user\s*: disabled\s*P\(0\)"
       And stdout should match "anyuid\s*: disabled\s*P\(0\)"
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -67,7 +67,8 @@ var RootCmd = &cobra.Command{
 	Long:  `Minishift is a command-line tool that provisions and manages single-node OpenShift clusters optimized for development workflows.`,
 	PersistentPreRun: func(cmd *cobra.Command, args []string) {
 		var (
-			err error
+			err                    error
+			isAddonInstallRequired bool
 		)
 
 		constants.MachineName = constants.ProfileName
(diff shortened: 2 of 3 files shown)


LalatenduMohanty:
Related, opened #1505

test/integration/features/cmd-addons.feature
```diff
@@ -8,12 +8,9 @@ Feature: Addons command and its subcommands
...
 
- Scenario: Installing default add-ons manually
-     When executing "minishift addons install --defaults" succeeds
-     Then stdout should contain
-      """
-      Default add-ons anyuid, admin-user, xpaas, registry-route installed
-      """
+  @minishift-only
+  Scenario: Default add-ons are installed but not enabled by default
+  Note: default addons were installed during previous commands automatically.
```
hferentschik: Maybe you could indent this so that it becomes clear that it is part of the scenario description

test/integration/features/cmd-addons.feature
```diff
@@ -8,12 +8,9 @@ Feature: Addons command and its subcommands
       And executing "minishift add-ons" fails
       And executing "minishift add-on" fails
 
- Scenario: Installing default add-ons manually
-     When executing "minishift addons install --defaults" succeeds
-     Then stdout should contain
-      """
-      Default add-ons anyuid, admin-user, xpaas, registry-route installed
-      """
+  @minishift-only
```
hferentschik: I am intrigued, was this working for CDK before?

test/integration/features/cmd-addons.feature
```diff
@@ -8,12 +8,9 @@ Feature: Addons command and its subcommands
       And executing "minishift add-ons" fails
       And executing "minishift add-on" fails
 
- Scenario: Installing default add-ons manually
```
hferentschik: Instead of removing this scenario, you should keep it under the ' Default add-ons are installed but not enabled by default' scenario maybe once w/ and w/o the '--force' option. By removing it you remove the manual install test compleltey.

test/integration/features/cmd-addons.feature
hferentschik: Unrelated, line 8 - 9:

      And executing "minishift add-ons" fails
      And executing "minishift add-on" fails

I think this should be removed. Why would we want to ensure that this fails?

hferentschik: Personally I would try to keep the use of '@minishift-only' only minimal. I would use liberal regular expressions for the default add-ons which are enabled per default in the CDK. This way the whole test can run against Minishift or CDK.

LalatenduMohanty: @hferentschik I just reverting the changes I had done as part of https://github.com/minishift/minishift/pull/1419/. I think we can remove the @minishift-only from here and it would work fine in CDK . @agajdosi @praveenkumar WDYT?

test/integration/features/cmd-addons.feature
LalatenduMohanty: @hferentschik nope. The manual install test is there at the line number 50 of this file. I changed it as part of https://github.com/minishift/minishift/pull/1419/ and now I reverted it back.

test/integration/features/cmd-addons.feature
agajdosi: @LalatenduMohanty If we remove it, then it will fail, as `minishift addons list` will return addons which are enabled by default on CDK.

@hferentschik I am trying to keep the use of `minishift-only` minimal.  Here it makes sense I think as we also want to test that default addons are installed and disabled.

LalatenduMohanty: @hferentschik @praveenkumar @agajdosi I kept the the test @minishift-only as we should also check that add-ons are disabled by default for upstream Minishift.

Issue #1512, opened 2017-10-07
Minikube ISO name does not show in the iso-url usage


LalatenduMohanty:
Currently on b2d and CentOS is shown in the usage and the documentation does not have information about minikube ISO [1].
```
$ minishift start --help | grep iso-url
      --iso-url string                  Location of the minishift ISO. Can be an URL, file URI or one of the following short names: [b2d centos]. (default "b2d")
```
[1] https://docs.openshift.org/latest/minishift/using/managing-minishift.html#choosing-iso-image

gbraad:
At the moment I think this is done on purpose... however, I think some of this also needs to be behind a feature toggle.

LalatenduMohanty:
@gbraad I understand your concerns but unless we do not expose the `minikube` ISO name in the usage I do not it will get any proper testing.  I would suggest that we expose the short name and mention in the documentation [1] that the ISO is newly added to Minishift and should be considered experimental. 
@minishift/minishift-dev  comments?

[1] https://docs.openshift.org/latest/minishift/using/managing-minishift.html#choosing-iso-image

gbraad:
But when EnableExperimental is enabled I would say, as we have not performed any rigorous testing by ourselves.

budhrg:
Resolved via https://github.com/minishift/minishift/pull/1514

PR #1514, opened 2017-10-08
Issue #1512 Adding Minikube ISO to usage and documentation

LalatenduMohanty:

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -486,7 +486,6 @@ func determineIsoUrl(iso string) string {
 func initStartFlags() *flag.FlagSet {
 	startFlagSet := flag.NewFlagSet(commandName, flag.ContinueOnError)
 
-	startFlagSet.String(configCmd.ISOUrl.Name, minishiftConstants.B2dIsoAlias, "Location of the minishift ISO. Can be an URL, file URI or one of the following short names: [b2d centos].")
 	startFlagSet.String(configCmd.VmDriver.Name, constants.DefaultVMDriver, fmt.Sprintf("The driver to use for the Minishift VM. Possible values: %v", constants.SupportedVMDrivers))
 	startFlagSet.Int(configCmd.CPUs.Name, constants.DefaultCPUS, "Number of CPU cores to allocate to the Minishift VM.")
 	startFlagSet.String(configCmd.Memory.Name, constants.DefaultMemory, "Amount of RAM to allocate to the Minishift VM. Use the format <size><unit>, where unit = MB or GB.")
@@ -504,6 +503,12 @@ func initStartFlags() *flag.FlagSet {
 		startFlagSet.String(configCmd.Netmask.Name, "24", "Specify netmask to use for the IP address. Ignored if no IP address specified (experimental - Hyper-V only)")
 		startFlagSet.String(configCmd.Gateway.Name, "", "Specify gateway to use for the instance. Ignored if no IP address specified (experimental - Hyper-V only)")
 		startFlagSet.String(configCmd.Nameserver.Name, "8.8.8.8", "Specify nameserver to use for the instance. Ignored if no IP address specified (experimental - Hyper-V only)")
+	}
+
+	if minishiftConfig.EnableExperimental {
+		startFlagSet.String(configCmd.ISOUrl.Name, minishiftConstants.B2dIsoAlias, "Location of the minishift ISO. Can be an URL, file URI or one of the following short names: [b2d centos minikube].")
+	} else {
+		startFlagSet.String(configCmd.ISOUrl.Name, minishiftConstants.B2dIsoAlias, "Location of the minishift ISO. Can be an URL, file URI or one of the following short names: [b2d centos].")
 	}
 
 	return startFlagSet
--- a/docs/source/using/managing-minishift.adoc
+++ b/docs/source/using/managing-minishift.adoc
@@ -61,6 +61,13 @@ The image size is small and optimized for development but not for production.
 This ISO image is based on link:https://www.centos.org/[CentOS], which is an enterprise-ready Linux distribution that more closely resembles a production environment.
 The image size is larger than the Boot2Docker ISO image.
 
+- link:https://github.com/kubernetes/minikube/tree/master/deploy/iso/minikube-iso[minikube ISO].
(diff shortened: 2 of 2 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -486,7 +486,7 @@ func determineIsoUrl(iso string) string {
 func initStartFlags() *flag.FlagSet {
 	startFlagSet := flag.NewFlagSet(commandName, flag.ContinueOnError)
 
-	startFlagSet.String(configCmd.ISOUrl.Name, minishiftConstants.B2dIsoAlias, "Location of the minishift ISO. Can be an URL, file URI or one of the following short names: [b2d centos].")
+	startFlagSet.String(configCmd.ISOUrl.Name, minishiftConstants.B2dIsoAlias, "Location of the minishift ISO. Can be an URL, file URI or one of the following short names: [b2d centos minikube].")
```
praveenkumar: 👍

docs/source/using/managing-minishift.adoc
```diff
@@ -61,6 +61,13 @@ The image size is small and optimized for development but not for production.
 This ISO image is based on link:https://www.centos.org/[CentOS], which is an enterprise-ready Linux distribution that more closely resembles a production environment.
 The image size is larger than the Boot2Docker ISO image.
 
+- link:https://github.com/kubernetes/minikube/tree/master/deploy/iso/minikube-iso[minikube ISO].
+This ISO image is provided by link:https://github.com/kubernetes/minikube/[Minikube project]. It is a customized Linux and based on link:https://www.kernel.org/[upstream kernel].
+Minikube ISO image provides alternate container runtime e.g. cri-o and rkt along with docker.
```
praveenkumar: `image provides` => `images also provides`

docs/source/using/managing-minishift.adoc
```diff
@@ -61,6 +61,13 @@ The image size is small and optimized for development but not for production.
...
+Please check link:https://github.com/kubernetes/minikube/blob/master/docs/contributors/minikube_iso.md[Minikube documentation] for details.
+
+NOTE: Minikube ISO image use is experimental in Minishift. Work is in progress to make it stable similar to CentOS and Boot2Docker ISO images.
```
praveenkumar: `Minikube ISO image use is experimental` => `Minikube ISO use is experimental`

LalatenduMohanty: @praveenkumar using `ISO image` instead of `ISO` is something we have discussed and agreed in the past with @thatdocslady . The current documentation also uses `ISO image`.

cmd/minishift/cmd/start.go
gbraad: -1

gbraad:
I believe this is premature and needs to be an experimental feature. We have not done any integration tests ourselves. We do jot even know if the image is on feature parity with hostfolders support, etc.

docs/source/using/managing-minishift.adoc
gbraad: Similarly

cmd/minishift/cmd/start.go
gbraad: Although the provisioner is not experimentally enabled, at least this string could be behind the toggle.

docs/source/using/managing-minishift.adoc
gbraad: ... as stable as..

cmd/minishift/cmd/start.go
gbraad: ... and the shorturl functionality that will likely follow this.

cmd/minishift/cmd/start.go
```diff
@@ -60,7 +60,7 @@ import (
 const (
 	commandName             = "start"
 	defaultInsecureRegistry = "172.30.0.0/16"
-	unsupportedIsoUrlFormat = "Unsupported value for iso-url. It can be an URL, file URI or one of the following short names: [b2d centos]."
+	unsupportedIsoUrlFormat = "Unsupported value for iso-url. It can be an URL, file URI or one of the following short names: [b2d centos minikube]."
```
budhrg: `b2d, centos, minikube` should be extracted into constants as `supportedISOs`.

cmd/minishift/cmd/start.go
LalatenduMohanty: @gbraad Even if we only show the usage of minikube short name when experimental is enabled it will still work if a user explicitly mentions it `minishift start --iso-url minikube`. Do you mean I should put a check to stop the execution if experimental features are not enabled and then minikube ISO is used.

cmd/minishift/cmd/start.go
LalatenduMohanty: @budhrg I think we should handle this in an separate issue for the suggested code restructuring as I wanted to merge this asap for v1.7.0 release

cmd/minishift/cmd/start.go
```diff
@@ -506,6 +505,12 @@ func initStartFlags() *flag.FlagSet {
 		startFlagSet.String(configCmd.Nameserver.Name, "8.8.8.8", "Specify nameserver to use for the instance. Ignored if no IP address specified (experimental - Hyper-V only)")
 	}
 
+	if minishiftConfig.EnableExperimental {
+		startFlagSet.String(configCmd.ISOUrl.Name, minishiftConstants.B2dIsoAlias, "Location of the minishift ISO. Can be an URL, file URI or one of the following short names: [b2d centos minikube].")
+	} else {
+		startFlagSet.String(configCmd.ISOUrl.Name, minishiftConstants.B2dIsoAlias, "Location of the minishift ISO. Can be an URL, file URI or one of the following short names: [b2d centos].")
+	}
+
```
LalatenduMohanty: @gbraad @praveenkumar @hferentschik do you think it adds any value? Honestly  I am not sure.

cmd/minishift/cmd/start.go
budhrg: Sure :+1:

cmd/minishift/cmd/start.go
gbraad: For the time being... The actual problem is that the provsioner is not behind a toggle. Also the shorturl function is more effective in this regard.

Issue #1491, opened 2017-10-04
Upgrade to use Go 1.8


hferentschik:
I think it is time to align with a later Go version. 1.9 is out now. In the documentation as well as in the CI systems we align with 1.7. It is time to update. 

Doing so would not also give the opportunity to make some build improvements since 1.9 handles the vendor directory a bit differently now.

praveenkumar:
There is only one problem, currently from CentOS/RHEL side we don't have 1.9 available yet but 1.8 is there quite some time and we should consider that for this iteration till 1.9 get listed in official repository of RHEL/CentOS. 

@hferentschik If that sound good then I will take this issue and try to update all the CI systems.

hferentschik:
To gain some additional benefits in relation to the build environment we need 1.9. I am using 1.8 for quite some time and it works just fine, but it "just" offers a Go version update (including potential Go bug fixes). With 1.9 we can do a bit more. 

We can as a stepping stone go to 1.8 first, but we should then create a new issues for that. This would really just affect the CI configs and the docs. We can then at a later point in time make a jump to 1.9. 

Personally, I think not having a CentOS/RHEL package for Go 1.9 should not be a show stopper. After all you can install it manually - https://golang.org/dl/. In fact this is probably the better solution anyways. This way you can install multiple versions of Go in parallel and switch between them by changing a symbolic link.

praveenkumar:
> We can as a stepping stone go to 1.8 first, but we should then create a new issues for that. This would really just affect the CI configs and the docs. We can then at a later point in time make a jump to 1.9.

Yes agree to that.

>Personally, I think not having a CentOS/RHEL package for Go 1.9 should not be a show stopper. After all you can install it manually - https://golang.org/dl/. In fact this is probably the better solution anyways. This way you can install multiple versions of Go in parallel and switch between them by changing a symbolic link.

@hferentschik I do agree but on downstream side to build our CDK binary, we can only able to use what provided in the official repository so I still think to have same golang version to build and test on upstream side.

gbraad:
>  After all you can install it manually - https://golang.org/dl/. In fact this is probably the better solution anyways

I would prefer to stick with packaged versions, as that will make deployment a lot easier and consistent. These also live in cache on the CentOS CI side.

> We can as a stepping stone go to 1.8 first,

This sounds to me as the better approach... for now.

hferentschik:
[...]

PR #1516, opened 2017-10-09
Issue #1491 Update CI to go-1.8.3

praveenkumar:


hferentschik:
Don't forget the docs ;-)

praveenkumar:
> Don't forget the docs 

@hferentschik I can only see go version reference in the `developing` documents and same updated. Not sure if we have anywhere else doc regarding specific go version tag.

docs/source/contributing/developing.adoc
```diff
@@ -15,7 +15,7 @@ The following sections describe how to build and test Minishift.
 == Prerequisites
 
 - Git
-- A recent Go distribution (>1.7)
+- A recent Go distribution (>1.8)
```
LalatenduMohanty: @praveenkumar @hferentschik I think we should keep it (>1.7) as it still works with 1.7.

hferentschik: I think we should raise bar here as well. I don't want to have to chase issues due to someone using an older Go version. Especially if we make some Make changes uses some 1.8 features.

praveenkumar: I do agree, let's not use an older version on the doc side.

LalatenduMohanty: Ok, then we should mark the version specifically 1.8 because with current change means it will work with 1.9 and future version i.e. Go distribution version 1.8` or at least `Go distribution (>= 1.8)`.

appveyor.yml
```diff
@@ -6,11 +6,16 @@ clone_folder: C:\gopath\src\github.com\minishift\minishift
...
 environment:
   GOPATH: C:\gopath
+  GOVERSION: 1.8.3
 
 platform:
 - x64
 
 install:
+# Install the specific Go version.
+- rmdir c:\go /s /q
```
hferentschik: Interesting to see how this is done on app veyor

centos_ci.sh
```diff
@@ -46,6 +46,7 @@ function install_core_deps() {
   yum -y install epel-release
   # Get all the deps in
   yum -y install gcc \
+                 golang \
```
hferentschik: so where is the version here? How does this ensure that the version does not change? Do we not need to specify one?

centos_ci.sh
```diff
@@ -103,14 +104,7 @@ function setup_kvm_docker_machine_driver() {
   echo 'CICO: Setup KVM docker-machine driver setup successfully'
 }
 
-function install_and_setup_golang() {
-  # Install Go 1.7
-  curl -LO https://storage.googleapis.com/golang/go1.7.linux-amd64.tar.gz
```
hferentschik: Would we not just download the appropriate 1.8 version here?

docs/source/contributing/developing.adoc
hferentschik: +1 for >= 1.8

centos_ci.sh
praveenkumar: @hferentschik Here go version comes from official repo of centos which is 1.8.3 currently. This is the reason we are not moving to 1.9

LalatenduMohanty: The right approach for installing a particular version with YUM is to download the RPM using wget or curl and then use `yum localinstall /path/to/package.rpm`

praveenkumar: @LalatenduMohanty  big -1 we should never download an rpm locally, assume tomorrow if distribution move with 1.9 then you never know package dependencies which a specific package depend on should be same because it will be break with conflict (you can try to download a lower version of vim rpm locally and then install it, you can see all the conflicts). Also if that was the intentions then why not go for 1.9 because packages are available on Fedora side and we can download but it again breaks on conflict and dependencies resolution.

centos_ci.sh
praveenkumar: @hferentschik No now we are consuming version directly from official repository.

centos_ci.sh
LalatenduMohanty: @praveenkumar  It depends on what we want. If we want to hard fix the version at 1.8.3 it is fine as if the distribution moves on and we get conflicts the build will fail eventually and that would be the point when we will have to move to a new version consciously. Personally I think the current code is fine as it is inline with the documentation i.e. >= 1.8.

centos_ci.sh
hferentschik: But here we pin the version and it is clear what is used. If we just use the latest from the repo, we don't really know what's used.

centos_ci.sh
hferentschik: > If we want to hard fix the version at 1.8.3 it is fine using localinstall

If possible we want to pin the version. That's what the existing local install did. I don't understand why we want to open ourself up for potential problems due to the fact the packaged version change.

>  i.e. if the distribution moves on and we get conflicts the build will fail eventually and that would be the point when we will have to move to a new version consciously.

A failing build might be the easy case. There could be much more subtle errors. If we can we should be explicit and imo in this case we can. 

 > Personally I think the current code is fine as it is inline with the documentation i.e. >= 1.8.

For me this change is still a -1. If we can explicitly set the version we should do so.

hferentschik: Also, how do I know which version is installed?

LalatenduMohanty: @hferentschik  We are contradicting our self here because in the documentation we are saying any version >= 1.8 is fine. But we want to pin to a particular version. Lets assume we pinned the version to 1.8.3 and everything is fine. We will never know if there is an issue exists with 1.9 version and then a contributor might use it and get in to issues. Thats why I mentioned that we should decide between = 1.8.3 or >= 1.8.3 in the docs.

hferentschik: > We are contradicting our self here because in the documentation we are saying any version >= 1.8 is fine. 

That's local development. Personally, I even use 1.9, but that's a conscious decision. The CI environment is a different beast. The less moving parts the better. 

The main issue is that there is no indication of what version is used whatsoever. I would be ok with the yum install if it were in the form `sudo yum install golang-<version info>`. But then again, we have this function which does a manual install, why change it? Maybe for some reason, we will use a slightly different version from the packaged version in a short time again. Do you then want to dig this code out again. Most likely it is by the forgotten how it was done and someone needs to re-do it.

> But we want to pin to a particular version. Lets assume we pinned the version to 1.8.3 and everything is fine. We will never know if there is an issue exists with 1.9 version and then a contributor might use it and get in to issues. 

You will know since some people will try with newer versions. Or you schedule a dedicated task for the upgrade. If you want to be proactive and get notified early, you could create an additional matrix job which compiles against multiple versions. But this is again an active choice and explicit. In the pull request builder you want to minimise side effects.

> Thats why I mentioned that we should decide between = 1.8.3 or >= 1.8.3 in the docs.

I think I am in this case more inclined to change the docs. Something like "The recommended Golang version is 1.8 which is used on the CI systems as well as during the release."

Anyways, I made my case. If you guys see no merit in what I am saying, go ahead and merge.

praveenkumar:
@hferentschik put some comments as per our discussion also updated PR.

centos_ci.sh
```diff
@@ -45,7 +45,10 @@ function install_core_deps() {
   # Install EPEL repo
   yum -y install epel-release
   # Get all the deps in
+  # We are installing golang from offical repository to make sure our downstream builds works as expected.
```
hferentschik: 👍

centos_ci.sh
```diff
@@ -103,14 +106,9 @@ function setup_kvm_docker_machine_driver() {
...
-  # Install Go 1.7
-  curl -LO https://storage.googleapis.com/golang/go1.7.linux-amd64.tar.gz
-  tar -xf go1.7.linux-amd64.tar.gz
-  sudo mv go /usr/local
-
-  # Setup GOROOT
-  export GOROOT=/usr/local/go
+function setup_golang() {
+  # Show which version of golang in the offical repo.
+  go version
```
hferentschik: +1 for dumping the Go version

hferentschik:
@praveenkumar if it passes the CI systems, feel free to merge

Issue #1518, opened 2017-10-09
Profile set should create the profile and set it as active profile


LalatenduMohanty:
Currently `profile set` returns error if a profile does not exist. 
```
$ minishift profile set test123
Error: 'test123' is not a valid profile
```
But it will be useful if it creates a empty profile and sets it as an active profile for use cases where users wants to set config values for a profile and then want to start it. Here is an example

```
$ minishift profile set xyz
created an empty profile 'xyz' and set it as an active profile

$ minishift profile list
- minishift	Does Not Exist 
- xyz		Does Not Exist (active)

$ minishift config set memory 8GB

$ minishift start
```

LalatenduMohanty:
CC @jorgemoralespou @hferentschik @praveenkumar

praveenkumar:
@LalatenduMohanty thanks, I am assigning it to myself to get some idea around profile codebase.

budhrg:
There is a discussion going on adding create command here https://github.com/minishift/minishift/issues/179.  

I assumed that `set` command has pre-requisite of existing profile. 
I feel with `profile set` has a different meaning altogether. 

`minishift profile create` would have been ideal command to achieve that.

LalatenduMohanty:
@budhrg From an user experience point of view `profile set xyz` should set zyz as the active profile and we all agree to this. Now the question is if the profile does not exists, what should be the behavior. 

It seems users want to run the config set commands first to set parameters and then run `minishift start`. Also they want to set active profile before hand so that any config set commands  they run it would be against this profile. So I think we can extend profile set to create a profile if it does not exist. 

Regarding `minishift create` we still need to think about it.

gbraad:
I think this is a side-effect which is not expected. This could lead to creating a multiple of profiles which are not intended. Think of the user who wants to set to a demo profile: `m profile set demo2`. But it doesn't exist and creates it... "Eh, that is not what I wanted?", so he actually moved to a config and execution context of something that is empty.

This is not the expected result of the command I performed.

jorgemoralespou:
Agree. I don't like side effects.

I would warn the user that the profile does not exist, and then let him create. I would probably not add a create command, but add a flag to both "m profile set". This could be combined with the flag I describe here (https://github.com/minishift/minishift/issues/1517#issuecomment-335540444) to make template profiles

As an example:

```
$m profile set example
-- example profile does not exist, use --force if you want to create the profile
$m profile set example --force 
-- You're now in example profile
```


```
[...]

PR #1540, opened 2017-10-11
Issue #1518 Profile set should create the profile and set it as active profile

praveenkumar:


LalatenduMohanty:
@praveenkumar Lets reach to a conclusion on #1518 first and then decide what to do next.

Issue #1266, opened 2017-08-15
Refactor: proxy URL validation redundant


gbraad:
https://github.com/minishift/minishift/blob/3b82cbed7eba3ee6890621e5e966ea4515b19fa0/pkg/util/proxy.go#L145-L154

https://github.com/minishift/minishift/blob/377c7db723615794dceabdbfcbf8c4979f5d33dd/cmd/minishift/cmd/config/validations.go#L95-L100

gbraad:
With #1263 most of the validation code is moved from `cmd/config` to `pkg/minishift`.

gbraad:
Needs to be addressed

anjannath:
@gbraad the refactor would involve removing the check from `proxy.go` and making it contained in the `validations.go` file?

gbraad:
likely, as currently it is used in `m config set`, `m start` and maybe in future rather as a pre-flight check (instead of start).

anjannath:
This will be implemented as part of #1569

PR #1543, opened 2017-10-11
Issue #1397 Improve pre-flight check for ISO URL

### LINKED ISSUE #1397: Improve pre-flight checks: add test for iso-url path
While `m config set iso-url ...` allows a file to be specified, we aren't sure if this file exists. When we do `m start` (using config set) or `m start --iso-url file://`, only the existence of this file will be checked by libmachine/driver. This means that horrible output can follow like:

```
.Creating machine...
(minishift) Calling .Create
(minishift) Downloading C:\Users\gbraad\.minishift\cache\boot2docker.iso from file:///c:/users/gbraa
d/.minishift/cache/iso/minishift-hvkvp.iso...
... FAIL E0918 13:05:12.333096   17352 start.go:347] Error starting the VM: Error creating the VM. E
rror creating machine: Error in driver during machine creation: open /c:/users/gbraad/.minishift/cac
he/iso/minishift-hvkvp.iso: The filename, directory name, or volume label syntax is incorrect.. Retr
ying.
```

There is no need in retrying, as the path is invalid: `file:///c:/users/gbraad/.minishift/cache/iso/minishift-hvkvp.iso`

Possible valid:
  * Windows  
     `--iso-url file://X:/` (no backslashes, absolute starting from a drive letter)
  * Linux/macOS  
     `--iso-url file:///` (absolute from `/` [root], hence the three slashes)
  * does file exist?

anjannath:
Fixes #1397

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -60,13 +60,14 @@ import (
 const (
 	commandName             = "start"
 	defaultInsecureRegistry = "172.30.0.0/16"
-	unsupportedIsoUrlFormat = "Unsupported value for iso-url. It can be an URL, file URI or one of the following short names: [b2d centos]."
 )
 
 var (
-	dockerEnv     []string
-	openShiftEnv  []string
-	shellProxyEnv string
+	dockerEnv               []string
+	openShiftEnv            []string
+	shellProxyEnv           string
+	unsupportedIsoUrlFormat = fmt.Sprintf("Unsupported value for iso-url. It can be a URL, file URI or one of the following aliases: [%s].",
+		strings.Join(minishiftConstants.ValidIsoAliases, ","))
 )
 
 var (
--- a/cmd/minishift/cmd/start_preflight.go
+++ b/cmd/minishift/cmd/start_preflight.go
@@ -22,6 +22,7 @@ import (
 	"os"
 	"os/exec"
 	"regexp"
+	"runtime"
 	"strconv"
 	"strings"
 
@@ -33,6 +34,7 @@ import (
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/viper"
 
+	"github.com/minishift/minishift/pkg/minishift/constants"
 	stringUtils "github.com/minishift/minishift/pkg/util/strings"
 )
 
@@ -91,6 +93,13 @@ func preflightChecksBeforeStartingHost() {
 			false, configCmd.WarnCheckHyperVDriver.Name,
 			driverErrorMessage)
 	}
+
+	preflightCheckSucceedsOrFails(
+		configCmd.SkipCheckIsoUrl.Name,
+		checkIsoUrl,
+		"Checking the ISO URL",
+		false, configCmd.WarnCheckIsoUrl.Name,
+		"See the 'Managing Minishift' topic for more information")
 }
 
 // preflightChecksAfterStartingHost is executed after the startHost function.
@@ -416,4 +425,37 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
 	}
 
 	return true, nil
+}
+
+// checkIsoUrl checks the Iso url and returns true if the iso file exists
+func checkIsoUrl() bool {
+	isoUrl := viper.GetString(configCmd.ISOUrl.Name)
+	for _, isoAlias := range constants.ValidIsoAliases {
+		if isoUrl == isoAlias {
+			return true
+		}
+	}
+	if !strings.HasSuffix(isoUrl, ".iso") {
+		return false
+	}
+
+	if runtime.GOOS == "windows" {
+		match, _ := regexp.MatchString("^file://[a-zA-Z]:/.+", isoUrl)
+		if !match {
+			return false
+		}
+	} else {
+		match, _ := regexp.MatchString("^file:///.+", isoUrl)
+		if !match {
+			return false
+		}
+	}
+
+	fmt.Printf("\n   Checking if %s exists ... ", strings.TrimPrefix(isoUrl, "file://"))
+	_, err := os.Stat(strings.TrimPrefix(isoUrl, "file://"))
+	if err != nil {
+		return false
+	}
+
+	return true
 }
--- a/docs/source/using/managing-minishift.adoc
+++ b/docs/source/using/managing-minishift.adoc
@@ -85,9 +85,21 @@ $ minishift start --iso-url https://github.com/minishift/minishift-centos-iso/re
 ----
 
 - Manually download the {project} CentSO ISO image from the link:https://github.com/minishift/minishift-centos-iso/releases[releases page] and enter the file URI to the image:
+
+[NOTE]
+====
+The path given for a file URI must be an absolute path.
+====
+- On GNU/Linux or macOS, the path must begin with *_/_*. For example:
 +
 ----
-$ minishift start --iso-url file:///<path_to_ISO_image>
+$ minishift start --iso-url file:///path/to/image.iso
+----
+
+- On Windows, the path must begin with a drive letter (*_C:_*, *_D:_*). For example:
++
+----
+$ minishift start --iso-url file://c:/path/to/image.iso
 ----
 
 NOTE: You cannot run {project} with both ISO images concurrently.
--- a/pkg/minishift/constants/constants.go
+++ b/pkg/minishift/constants/constants.go
@@ -25,3 +25,7 @@ const (
 	DefaultProject         = "myproject"
 	DefaultUser            = "developer"
 )
+
+var (
+	ValidIsoAliases = []string{"b2d", "centos", "minikube"}
+)
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -104,6 +104,8 @@ var (
 	WarnCheckXHyveDriver  = createConfigSetting("warn-check-xhyve-driver", SetBool, nil, nil, true)
 	SkipCheckHyperVDriver = createConfigSetting("skip-check-hyperv-driver", SetBool, nil, nil, true)
 	WarnCheckHyperVDriver = createConfigSetting("warn-check-hyperv-driver", SetBool, nil, nil, true)
+	SkipCheckIsoUrl       = createConfigSetting("skip-check-iso-url", SetBool, nil, nil, true)
+	WarnCheckIsoUrl       = createConfigSetting("warn-check-iso-url", SetBool, nil, nil, true)
 	// Preflight checks (after start)
 	SkipInstanceIP        = createConfigSetting("skip-check-instance-ip", SetBool, nil, nil, true)
 	WarnInstanceIP        = createConfigSetting("warn-check-instance-ip", SetBool, nil, nil, true)


anjannath:
retest this please

anjannath:
retest this please

anjannath:
cc @gbrrad

cmd/minishift/cmd/start_preflight.go
```diff
@@ -414,3 +422,31 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
 
 	return true, nil
 }
+
+// checkIsoUrl checks the Iso url and returns true if the iso file exists
+func checkIsoUrl() bool {
+	isoUrl := viper.GetString(configCmd.ISOUrl.Name)
+
+	if isoUrl == "b2d" || isoUrl == "centos" || isoUrl == "minikube" {
```
hferentschik: We need the list of short names now at multiple places. We should define it in a single place (eg as slice) and use it where appropriate. We already have with the minikube ISO the case where not all usages where updated.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -414,3 +422,31 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
...
+		return true
+	}
+
+	if runtime.GOOS == "windows" {
```
hferentschik: do we really need a OS check. Can we not check as single regexp?

docs/source/using/managing-minishift.adoc
```diff
@@ -89,6 +89,11 @@ $ minishift start --iso-url https://github.com/minishift/minishift-centos-iso/re
 ----
 $ minishift start --iso-url file:///<path_to_ISO_image>
 ----
+- In a WIndows machine the path always starts with a drive letter.
```
hferentschik: On a Windows machine ...

Also, you say it, but your example does not reflect it.

cmd/minishift/cmd/start_preflight.go
anjannath: But a path like `file:///home/anath/shared` will be valid in a linux host but not in a windows host, so if we don't check for the OS and put it in the same regexp, will it work as intended, because it will match any one of the pattern.

docs/source/using/managing-minishift.adoc
anjannath: Will update the document with `file://c:/<path_to_ISO_image>`

cmd/minishift/cmd/start_preflight.go
anjannath: Thanks, There's this file in `pkg/minishift/constants/constants.go` which has these constants, i'll  use that.

LalatenduMohanty: I agree with @hferentschik. We should have a slice which contains the vallid short names .Recently @budhrg also brought it up.

docs/source/using/managing-minishift.adoc
```diff
@@ -89,6 +89,11 @@ $ minishift start --iso-url https://github.com/minishift/minishift-centos-iso/re
 ----
 $ minishift start --iso-url file:///<path_to_ISO_image>
 ----
```
LalatenduMohanty: Can you also fix that the previous example i.e. `$ minishift start --iso-url file:///<path_to_ISO_image>` and mention it is for macOS and GNU/Linux.

LalatenduMohanty:
@anjannath I guess you have not tested the PR for Windows right as you do not have a Windows machine with you. So we (Me @hferentschik or @budhrg ) need to test it Windows before merging it.

anjannath:
@LalatenduMohanty Yes i have not tested it in windows. and also the short names are already defined in `pkg/minishift/constants/constants.go`, will it be any better if a slice is used and we still have to loop over it.

anjannath:
@LalatenduMohanty Is there a way to check if the file type is iso? Because now `os.Stat` will find anything that is on the file system (directories, etc.) the check will pass but the VM won't start and will get the same error.

budhrg:
Appveyor CI is failing due to make dependency installation failures. Will test in Windows machine.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -414,3 +423,29 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
 
 	return true, nil
 }
+
+// checkIsoUrl checks the Iso url and returns true if the iso file exists
+func checkIsoUrl() bool {
```
budhrg: @anjannath , this function can be make testable with adding argument to it.

```
func checkISOURL(url string) bool {
  
   ...
}

// Call as  checkISOURL(viper.GetString(configCmd.ISOUrl.Name))
```

And then add unit test for it. Probably this could be moved to some util package.

anjannath: is there any other place where this function will be needed? To make it testable we will have to add arguments like? and if we don't call it from `preflightCheckSucceedsOrFails` we will loose the nice error, output like `OK` `FAIL`, i don't know if that's abosolutely required but looks nicer.

budhrg: @anjannath , its only about function `checkISOURL`. 
We should always try to make our code testable.  In its current form, I hardly think of writing unit test for it.

anjannath: @budhrg Okay, i'll look at other unit tests and try to make it tesable. But what did you mean by 
`// Call as  checkISOURL(viper.GetString(configCmd.ISOUrl.Name))` this is for testing or calling in the preflight check?

LalatenduMohanty:
> also the short names are already defined in pkg/minishift/constants/constants.go, will it be any better if a slice is used and we still have to loop over it.

I think so, we should define  a slice for validIsoAlias and we should re use it. It should replace the hard coded names in start.go too i.e. https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/start.go#L63

LalatenduMohanty:
> Is there a way to check if the file type is iso? 

I do not think so. At least I am not aware of it. I think we should be fine by checking of the file names as `blah.iso` exists on the filesystem or not.

cmd/minishift/cmd/start_preflight.go
budhrg: @anjannath ,  We probably need to change the  syntax of `preflightCheckSucceedsOrFails` to pass arguments to `preflightCheckFunc`.

Ignore my comments. I will start new issue to have this discussion.

budhrg:
Right now not working in Windows 10 machine. Regex seems to be issue here. Debugging.

cmd/minishift/cmd/start_preflight.go
anjannath: @budhrg Okay, right now there are no unit tests for preflight checks. we need to add those

cmd/minishift/cmd/start_preflight.go
```diff
@@ -414,3 +423,29 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
...
+	if isoUrl == constants.B2dIsoAlias || isoUrl == constants.CentOsIsoAlias || isoUrl == constants.MinikubeIsoAlias {
+		return true
+	}
+
+	if runtime.GOOS == "windows" {
+		match, _ := regexp.MatchString("^file://[a-z]:/.+", isoUrl)
+		if match != true {
+			return false
+		}
+	}
```
budhrg: Line 441-444 need to be inside else block of above if condition since it checks Unix file path. Hence, I was getting preflight failure error even in passing valid iso url path.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -414,3 +423,29 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
...
+
+// checkIsoUrl checks the Iso url and returns true if the iso file exists
+func checkIsoUrl() bool {
+	isoUrl := viper.GetString(configCmd.ISOUrl.Name)
+
+	if isoUrl == constants.B2dIsoAlias || isoUrl == constants.CentOsIsoAlias || isoUrl == constants.MinikubeIsoAlias {
+		return true
+	}
+
+	if runtime.GOOS == "windows" {
```
budhrg: @anjannath , I believe, we can have one quick check of extension before doing OS based file path check. If iso path doesn't contain `.iso` extention, we even don't need to check OS based path.

cmd/minishift/cmd/start_preflight.go
gbraad: Validations.go

cmd/minishift/cmd/start_preflight.go
gbraad: The check for windows path is file://x:/ and for other platforms is file:///home/users/

I haven't check the code yet, but I guess this is whybhe checks for the OS

cmd/minishift/cmd/start_preflight.go
budhrg: slice should be go to constants.go

docs/source/using/managing-minishift.adoc
gbraad: Correction: `file://` (not three, as the third is part of the path already). Our documentation does not mention this correctly

cmd/minishift/cmd/start_preflight.go
budhrg: @anjannath 

> There's this file in pkg/minishift/constants/constants.go which already has these constants.

Those are independent constant name. We don't have slice.
May be
```
SupportedISOAlias = []string{B2dIsoAlias, CentOsIsoAlias, MinikubeIsoAlias}
```

cmd/minishift/cmd/start_preflight.go
```diff
@@ -414,3 +423,36 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
...
+		}
+	} else {
+		match, _ := regexp.MatchString("^file:///.+", isoUrl)
+		if !match {
+			return false
+		}
+	}
+
+	fmt.Printf("\n   Checking if %s exists... ", strings.TrimPrefix(isoUrl, "file://"))
+	_, err := os.Stat(strings.TrimPrefix(isoUrl, "file://"))
```
gbraad: +1

docs/source/using/managing-minishift.adoc
gbraad: It is better to mention it is an absolute path...

cmd/minishift/cmd/start_preflight.go
gbraad: @budhrg I disagree. You never write code specifically to be testable... You make code reusable, and testable comes eight this as it likely gets parameterized and has a result value.

Generally, preflight checks have been harder to test as they are specific to a situation. We only test on KVM in our integration. Mocking is suggested, but it is harder to mock power shell sessions, etc. You test what is deemed necessary... Because at the moment automation will not compensate for the lack of manual testing.

gbraad: I actually think we should move some of these tests to validations.go, as we have duplicated the proxy test, and I also see the need we actually do a preflight check to test this, which means three tests?

And what if we also implement a connect test to the proxy? The reusable code is than more valuable to be unit-tested

gbraad: s/than/then

pkg/minishift/constants/constants.go
```diff
@@ -25,3 +25,7 @@ const (
 	DefaultProject         = "myproject"
 	DefaultUser            = "developer"
 )
+
+var (
+	ValidIsoAlias = []string{"b2d", "centos", "minikube"}
```
gbraad: Duplication for what noe also exists/used in start.go, etc ?

cmd/minishift/cmd/start.go
```diff
@@ -60,13 +60,14 @@ import (
...
 )
 
 var (
-	dockerEnv     []string
-	openShiftEnv  []string
-	shellProxyEnv string
+	dockerEnv               []string
+	openShiftEnv            []string
+	shellProxyEnv           string
+	unsupportedIsoUrlFormat = fmt.Sprintf("Unsupported value for iso-url. It can be an URL, file URI or one of the following short names: [%s].",
```
gbraad: Where is this used?

cmd/minishift/cmd/start.go
```diff
@@ -60,13 +60,14 @@ import (
...
+		strings.Join(minishiftConstants.ValidIsoAlias, ","))
```
gbraad: Aliases

docs/source/using/managing-minishift.adoc
anjannath: `file:///` is shown as an example, and so is the case for the drive letter as a windows example. `file://c:/`.
 we can mention it as an absolute path and show only one example.

anjannath:
@gbraad does it need to be moved to validation.go right away or its fine here for now.
i have also updated the doc, based on your comments.

gbraad:
there is another issue #1266 to deal with the proxy validation, so maybe create at least a new issue to track it.

anjannath:
@gbraad can we not track it in that issue(#1266) itself?

cmd/minishift/cmd/start_preflight.go
```diff
@@ -414,3 +423,36 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
...
+		if isoUrl == isoAlias {
+			return true
+		}
+	}
+	if !strings.HasSuffix(isoUrl, ".iso") {
+		return false
+	}
+
+	if runtime.GOOS == "windows" {
+		match, _ := regexp.MatchString("^file://[a-z]:/.+", isoUrl)
```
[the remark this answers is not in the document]
gbraad: It should be case-insensitive in all cases, however on Windows it does not really matter as "MYFILE" and "myfile" are the same, as on most of the filesystems used with Windows the casing has no influence.

docs/source/using/managing-minishift.adoc
```diff
@@ -85,9 +85,17 @@ $ minishift start --iso-url https://github.com/minishift/minishift-centos-iso/re
...
+====
+The path given for a file URI must be an absolute path.
+
+On GNU/Linux or macOS, the path must begin with *_/_*.
+
+On Windows, the path must begin with a drive letter (*_C:_*, *_D:_*).
+====
 ----
-$ minishift start --iso-url file:///<path_to_ISO_image>
+$ minishift start --iso-url file://<absolute_path_to_ISO_image>
```
LalatenduMohanty: This is wrong I think (at least on macOS) as the two slashes `//` does not work. However the 3 slashes work fine.
```
$ minishift start --iso-url file://Users/lmohanty/Downloads/minishift-iso/minishift-b2d.iso
-- Checking if xhyve driver is installed ...
   Driver is available at /usr/local/bin/docker-machine-driver-xhyve
   Checking for setuid bit ... OK
-- Checking the ISO URL ... FAIL
   See the 'Managing Minishift' topic for more information
```
```
$ minishift start --iso-url file:///Users/lmohanty/Downloads/minishift-iso/minishift-b2d.iso
-- Checking if xhyve driver is installed ...
   Driver is available at /usr/local/bin/docker-machine-driver-xhyve
   Checking for setuid bit ... OK
-- Starting local OpenShift cluster using 'xhyve' hypervisor ...
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ...................... OK
   Pinging 8.8.8.8 ... OK
   Version: v3.6.0
   host-config-dir ... OK
   host-data-dir ... OK
   host-pv-dir ... OK
   host-volumes-dir ... OK
   routing-suffix ... OK
```

gbraad: @LalatenduMohanty an absolute path is from `/`... which means `file://` + `/my/absolute/path` becomes `file:///my/absolute/path`.

Maybe an example will highlight this fact better.

anjannath: @gbraad Right, but adding two examples would be noisy, I think more elaborate examples, for both OS can go into the FAQ (if this gets asked a lot). and only an example absolute path for *nix systems in the docs.

robin-owen: @anjannath I don't think two examples would be excessively noisy. Having more detailed information about using the `file://` scheme with absolute paths in the FAQ certainly makes sense to me, though.

As it is, the absolute path itself is explained clearly, but we don't explain how this interacts with `file://` -- this leads directly to the confusion regarding two versus three slashes. The two slashes are part of the `file://` scheme, whereas the third slash is the beginning of the absolute path on Linux/macOS.

By the way, if we only have an example of `--iso-url` using a `file://` scheme on Linux/macOS, then we'll encounter confusion from Windows users who are using three slashes because that's what was demonstrated in the documentation (`file:///c:/path/to/iso` and `file://c:/path/to/iso` are not equivalent).

Having an example `file://` URL for both Linux/macOS and Windows makes sense to me given the potential for confusion.

If we can explain this in a clear and concise way, I think that would be the best solution.

@gbraad @LalatenduMohanty Thoughts?

LalatenduMohanty: > Having an example file:// URL for both Linux/macOS and Windows makes sense to me given the potential for confusion.

@robin-owen Agree. I would suggest to go for two example i..e. one for Windows and one for Linux/macOS. 

In general the idea is to give the information in a way which would be easier to read and understand.

anjannath:
```
NOTE: The path given for a file URI must be an absolute path.

- On GNU/Linux or macOS, the path must begin with *_/_*. For example:
+
----
$ minishift start --iso-url file:///path/to/image.iso
----

- On Windows, the path must begin with a drive letter (*_C:_*, *_D:_*). For example:
+
----
$ minishift start --iso-url file://c:/path/to/image.iso
----
```

I think now it is clear enough, with both the examples. @robin-owen @LalatenduMohanty what do you guys think?

robin-owen: That's absolutely fine with me, I'd just continue using a block for the admonition so that we can keep all of the information inside of the note more clearly/easily:

```adoc
[NOTE]
====
// info here
====
```

As written in your comment, only the sentence "The path given for a file URI must be an absolute path" would be placed in the note. :)

gbraad:
Are more changes coming or ready for review again?

anjannath:
@gbraad i think its ready for review

cmd/minishift/cmd/start_preflight.go
```diff
@@ -417,3 +426,36 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
...
+			return false
+		}
+	} else {
+		match, _ := regexp.MatchString("^file:///.+", isoUrl)
+		if !match {
+			return false
+		}
+	}
+
+	fmt.Printf("\n   Checking if %s exists... ", strings.TrimPrefix(isoUrl, "file://"))
```
gbraad: For aesthetic consistency, please add a space
```diff
-	fmt.Printf("\n   Checking if %s exists... ", strings.TrimPrefix(isoUrl, "file://"))
+	fmt.Printf("\n   Checking if %s exists ... ", strings.TrimPrefix(isoUrl, "file://"))
```

gbraad:
Will do the rebase and correct this small issue and merge...

PR #1546, opened 2017-10-11
Issue #1135 Provide notation for default variable values in add-on metadata

### LINKED ISSUE #1135: Provide notation for default value for addon-envs
It would be cool, if one could define a default value for an addon-env so that you don't require to add it for those addons where a default value could be provided.

Example can be found in the [prometheus addon](https://github.com/minishift/minishift-addons/tree/master/add-ons/prometheus) where the template would install in the *kube-system* namespace, but in order to make that configurable so the user can select the namespace, a default will no longer apply. The user will always be required to install it.

Also it would be nice to provide the ability to have a notation in the description for the variables an addon can have, so that you can get details on the addon:

e.g:
````
minishift addons list --verbose
Name       : prometheus
Description: This template creates a Prometheus instance preconfigured to gather OpenShift and Kubernetes platform and node metrics and report them to admins. It is protected by an OAuth proxy that only allows access for users who have view access to the prometheus namespace. You may customize where the images (built from openshift/prometheus and openshift/oauth-proxy) are pulled from via template parameters.
Enabled    : disabled
Priority   : 0
Required-Envs: namespace
Default-Envs: namespace=kube-system
````

Additionally, bringing this out into an addon info command for that single addon would also be handy.

### PR DESCRIPTION
**Pull request by budhrg**
Fix #1135

### Updated

Tested with
```
# ~/.minishift/addons/demo-addon/demo-addon.addon 
# Name: demo-addon
# Description: Demo addon
# Required-Vars: USER,NAMESPACE
# Var-Defaults: NAMESPACE=kube-system

echo User is #{USER} and namespace is #{NAMESPACE} and project is #{PROJECT}
```

#### Run as: 
- minishift addons install demo-addon
- minishift addons apply demo-addon --addon-env USER=foo

#### Expected Output
User is budhram and namespace is kube-system and project is #{PROJECT}

Since, `PROJECT` variable doesn't has any value from `--addon-env` or `minishift config set addon-env` and not required it is displayed as expected. [Was in wrong assumption that it will shown as empty]

### DIFF
--- a/cmd/minishift/cmd/addon/addon_apply.go
+++ b/cmd/minishift/cmd/addon/addon_apply.go
@@ -20,19 +20,15 @@ import (
 	"fmt"
 
 	"github.com/docker/machine/libmachine"
-	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/provision"
 	"github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
-	"github.com/minishift/minishift/pkg/minishift/docker"
 	"github.com/minishift/minishift/pkg/minishift/oc"
-	"github.com/minishift/minishift/pkg/minishift/openshift"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
-	"gopkg.in/yaml.v2"
 )
 
 var (
@@ -98,29 +94,4 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
 			atexit.ExitWithMessage(1, fmt.Sprint("Error applying the add-on: ", err))
 		}
 	}
-}
-
-func determineRoutingSuffix(driver drivers.Driver) string {
-	defer func() {
-		if r := recover(); r != nil {
-			atexit.ExitWithMessage(1, "Cannot determine the routing suffix from the OpenShift master configuration.")
-		}
-	}()
-
-	sshCommander := provision.GenericSSHCommander{Driver: driver}
-	dockerCommander := docker.NewVmDockerCommander(sshCommander)
-
-	raw, err := openshift.ViewConfig(openshift.MASTER, dockerCommander)
-	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot get the OpenShift master configuration: %s", err.Error()))
-	}
-
-	var config map[interface{}]interface{}
-	err = yaml.Unmarshal([]byte(raw), &config)
-	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Cannot parse the OpenShift master configuration: %s", err.Error()))
-	}
-
-	// making assumptions about the master config here. In case the config structure changes, the code might panic here
-	return config["routingConfig"].(map[interface{}]interface{})["subdomain"].(string)
 }
--- a/docs/source/using/addons.adoc
+++ b/docs/source/using/addons.adoc
@@ -171,7 +171,7 @@ $ minishift addons apply acme                                 // <2>
 If the environment variable is not set not, interpolation does not occur.
 <2> When the add-on is applied, each occurrence of `#{PROJECT_USER}` within an add-on command gets replaced with the value of the environment variable `USER`.
 
-As add-on developer, you can enforce that a variable value is provided when the add-on gets applied by adding the varaible name to the _Required-Vars_ metadata header.
+As an add-on developer, you can enforce that a variable value is provided when the add-on gets applied by adding the variable name to the _Required-Vars_ metadata header.
 Multiple variables need to be comma separated.
 
 ----
@@ -179,6 +179,21 @@ Multiple variables need to be comma separated.
 # Description: ACME add-on
 # Required-Vars: PROJECT_USER
 ----
+
+You can set a default value for a variable using the _Var-Defaults_ metadata header.
+_Var-Defaults_ needs to be specified in the format of `<key>=<value>`.
+Multiple default key/value pairs can be separated by comma.
+
+NOTE: `=` and `,` are metacharacters and cannot be used as part of keys or values.
+
+----
+# Name: acme
+# Description: ACME add-on
+# Required-Vars: PROJECT_USER
+# Var-Defaults: PROJECT_USER=john
+----
+
+NOTE: Variable values specified via the command line using the `--addon-env` or set via `minishift config set addon-env` have precedence over _Var-Defaults_.
 
 [[default-addons]]
 == Default Add-ons
--- a/pkg/minishift/addon/addon_meta.go
+++ b/pkg/minishift/addon/addon_meta.go
@@ -19,8 +19,8 @@ package addon
 import (
 	"errors"
 	"fmt"
+	minishiftStrings "github.com/minishift/minishift/pkg/util/strings"
 	"regexp"
-	"strings"
 )
 
 var requiredMetaTags = []string{NameMetaTagName, DescriptionMetaTagName}
@@ -31,13 +31,20 @@ const (
 	DescriptionMetaTagName   = "Description"
 	RequiredOpenShiftVersion = "OpenShift-Version"
 	anyOpenShiftVersion      = ""
+	varDefaults              = "Var-Defaults"
 )
+
+type RequiredVar struct {
+	Key   string
+	Value string
+}
 
 // AddOnMeta defines a set of meta data for an AddOn. Name and Description are required. Others are optional.
 type AddOnMeta interface {
 	Name() string
 	Description() []string
-	RequiredVars() []string
+	RequiredVars() ([]string, error)
+	VarDefaults() ([]RequiredVar, error)
 	GetValue(key string) string
 	OpenShiftVersion() string
 }
@@ -47,27 +54,14 @@ type DefaultAddOnMeta struct {
 }
 
 func NewAddOnMeta(headers map[string]interface{}) (AddOnMeta, error) {
-	for _, tag := range requiredMetaTags {
-		if headers[tag] == nil {
-			return nil, errors.New(fmt.Sprintf("Metadata does not contain an mandatory entry for '%s'", tag))
-		}
-		switch v := headers[tag].(type) {
-		case string:
-			if v == "" {
-				return nil, errors.New(fmt.Sprintf("Metadata does not contain an mandatory entry for '%s'", tag))
-			}
-		case []string:
-			if len(v) == 0 {
-				return nil, errors.New(fmt.Sprintf("Metadata does not contain an mandatory entry for '%s'", tag))
-			}
-		default:
-			return nil, nil
-		}
+	if err := requiredMetaTagsCheck(headers); err != nil {
+		return nil, err
 	}
-	if headers[RequiredOpenShiftVersion] != nil {
-		if !checkVersionSemantic(headers[RequiredOpenShiftVersion].(string)) {
-			return nil, errors.New("Add-on only support OpenShift version semantics eg. 3.6.0 or >3.6.0, <3.9.0 or >=3.5 etc.")
-		}
+	if err := requiredOpenShiftVersionCheck(headers); err != nil {
+		return nil, err
+	}
+	if err := varDefaultsCheck(headers); err != nil {
+		return nil, err
 	}
 
 	metaData := &DefaultAddOnMeta{headers: headers}
@@ -86,12 +80,28 @@ func (meta *DefaultAddOnMeta) Description() []string {
 	return meta.headers[requiredMetaTags[1]].([]string)
 }
 
-func (meta *DefaultAddOnMeta) RequiredVars() []string {
+func (meta *DefaultAddOnMeta) RequiredVars() ([]string, error) {
 	if val, contains := meta.headers[requiredVars].(string); contains {
-		return meta.splitAndTrim(val)
+		return minishiftStrings.SplitAndTrim(val, ",")
 	} else {
-		return []string{}
+		return []string{}, nil
 	}
+}
+
+func (meta *DefaultAddOnMeta) VarDefaults() ([]RequiredVar, error) {
+	// Ignore errors as checking has been done during varDefaultsCheck as part of NewAddOnMeta
+	if val, contains := meta.headers[varDefaults].(string); contains {
+		items, _ := minishiftStrings.SplitAndTrim(val, ",")
+		defaults := make([]RequiredVar, 0, len(items))
+		for _, item := range items {
+			varDefault, _ := minishiftStrings.SplitAndTrim(item, "=")
+			defaults = append(defaults, RequiredVar{Key: varDefault[0], Value: varDefault[1]})
+		}
+
+		return defaults, nil
+	}
+
+	return []RequiredVar{}, nil
 }
 
 func (meta *DefaultAddOnMeta) GetValue(key string) string {
@@ -105,20 +115,66 @@ func (meta *DefaultAddOnMeta) OpenShiftVersion() string {
 	return anyOpenShiftVersion
 }
 
-func (meta *DefaultAddOnMeta) splitAndTrim(s string) []string {
-	// Trims the stream and then splits
-	trimmed := strings.TrimSpace(s)
-	split := strings.Split(trimmed, ",")
-	cleanSplit := make([]string, len(split))
-	for i, val := range split {
-		cleanSplit[i] = strings.TrimSpace(val)
-	}
-	return cleanSplit
-}
-
 func checkVersionSemantic(version string) bool {
 	// Strict match for <major> or <major>.<minor> or <major>.<minor>.<patch>
 	// (>=|>|<|<=)3.6.0, (>=|>|<|<=)3.6.0
 	match, _ := regexp.MatchString("^(|>|>=|<|<=)[0-9]+(.[0-9]+){0,2}(|\\s*,\\s*(|>|>=|<|<=)[0-9]+(.[0-9]+){0,2})$", version)
 	return match
+}
+
+func requiredMetaTagsCheck(headers map[string]interface{}) error {
+	for _, tag := range requiredMetaTags {
+		if headers[tag] == nil {
+			return errors.New(fmt.Sprintf("Metadata does not contain a mandatory entry for '%s'", tag))
+		}
+		switch v := headers[tag].(type) {
+		case string:
+			if v == "" {
+				return errors.New(fmt.Sprintf("Metadata does not contain a mandatory entry for '%s'", tag))
+			}
+		case []string:
+			if len(v) == 0 {
+				return errors.New(fmt.Sprintf("Metadata does not contain a mandatory entry for '%s'", tag))
+			}
+		default:
+			return nil
+		}
+	}
+	return nil
+}
+
+func requiredOpenShiftVersionCheck(headers map[string]interface{}) error {
+	if headers[RequiredOpenShiftVersion] != nil {
+		if !checkVersionSemantic(headers[RequiredOpenShiftVersion].(string)) {
+			return errors.New("Add-on only support OpenShift version semantics eg. 3.6.0 or >3.6.0, <3.9.0 or >=3.5 etc.")
+		}
+	}
+
+	return nil
+}
+
+func varDefaultsCheck(headers map[string]interface{}) error {
+	if val, contains := headers[varDefaults].(string); contains {
+		items, err := minishiftStrings.SplitAndTrim(val, ",")
+		if err != nil {
+			return err
+		}
+
+		for _, item := range items {
+			varDefault, err := minishiftStrings.SplitAndTrim(item, "=")
+			if err != nil {
+				return err
+			}
+			// Expect varDefault to be only having key and value
+			if len(varDefault) != 2 {
+				return errors.New(fmt.Sprintf("'%s' is not well formed", val))
+			}
+
+			if varDefault[0] == "" || varDefault[1] == "" {
+				return errors.New(fmt.Sprintf("'%s' is not well formed", val))
+			}
+		}
+	}
+
+	return nil
 }
--- a/pkg/minishift/addon/addon_meta_test.go
+++ b/pkg/minishift/addon/addon_meta_test.go
@@ -52,7 +52,7 @@ func Test_missing_name_returns_error(t *testing.T) {
 		t.Fatal("Expected an error, got none")
 	}
 
-	expectedError := "Metadata does not contain an mandatory entry for 'Name'"
+	expectedError := "Metadata does not contain a mandatory entry for 'Name'"
 	if err.Error() != expectedError {
 		t.Fatal(fmt.Sprintf("Expected error '%s', but got '%s'", expectedError, err.Error()))
 	}
@@ -70,7 +70,7 @@ func Test_missing_description_returns_error(t *testing.T) {
 		t.Fatal("Expected an error, got none")
 	}
 
-	expectedError := "Metadata does not contain an mandatory entry for 'Description'"
+	expectedError := "Metadata does not contain a mandatory entry for 'Description'"
 	if err.Error() != expectedError {
 		t.Fatal(fmt.Sprintf("Expected error '%s', but got '%s'", expectedError, err.Error()))
 	}
@@ -100,8 +100,8 @@ func Test_required_vars_meta_is_extracted(t *testing.T) {
 	testMap["Required-Vars"] = " USER, access_token "
 
 	addOnMeta := getAddOnMeta(testMap, t)
-
-	minishiftTesting.AssertEqualSlice(addOnMeta.RequiredVars(), []string{"USER", "access_token"}, t)
+	requiredVars, _ := addOnMeta.RequiredVars()
+	minishiftTesting.AssertEqualSlice(requiredVars, []string{"USER", "access_token"}, t)
 }
 
 func Test_required_vars_empty_if_not_specified(t *testing.T) {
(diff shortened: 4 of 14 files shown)

### CODE REVIEW

File: pkg/minishift/addon/command/interpolation_context.go [hunk 1]
```diff
@@ -70,13 +70,13 @@ func (c *defaultInterpolationContext) RemoveFromContext(key string) error {
...
 	for _, v := range c.context {
-		cmd = v.regexp.ReplaceAllLiteralString(cmd, v.subsitution)
+		cmd = v.regexp.ReplaceAllLiteralString(cmd, v.substitution)
 	}
 	return cmd
 }
 
 func (c *defaultInterpolationContext) Vars() []string {
-	keys := make([]string, len(c.context))
+	keys := make([]string, 0, len(c.context))
```
hferentschik: Why starting with a 0 size if you know exactly how many elements you are going to add?

budhrg:
@hferentschik Please check the discussion after this comment https://github.com/minishift/minishift/issues/1135#issuecomment-332777138. 

@jorgemoralespou  also seems to agree.

> I would expect that for specifying a default you would do something like #{USER:foo}.

This also seems nice and another approach.

pkg/minishift/addon/command/interpolation_context.go
budhrg: Seems best practice to minimize memory allocations https://stackoverflow.com/a/21363587/1120530.

budhrg: @hferentschik , looks like if we add length 2, we will not get the expected result https://play.golang.org/p/wFEQUJYUKb

budhrg: We could have achieve, same with `keys := make([]string, 0)` as well. Seems that Golang slice will grow organically as the elements get appended. But to minimize memory allocations, `keys := make([]string, 0, 2)` is the suggested approach.

docs/source/using/addons.adoc
```diff
@@ -157,6 +157,18 @@ Multiple variables need to be comma separated.
...
 ----
 
+In case, you want to have a default value for variable name, you can achieve that with _Default-Envs_ metadata header.
+Multiple default values need to be comma separated.
+
+----
+# Name: acme
+# Description: ACME add-on
+# Required-Vars: PROJECT_USER
+# Default-Envs: PROJECT_USER=john
```
LalatenduMohanty: I somehow do not find the name `Default-Envs` intuitive. What about `Default-values`?

LalatenduMohanty:
Tested the PR and it worked for me fine.

docs/source/using/addons.adoc
hferentschik: Default-Var-Values, Default-Values does not give any hint for what the defaults are. In the context of add-ons we talk about variable interpolation and required variables, so imo we are talking about default variable values here.

budhrg:
@hferentschik , @LalatenduMohanty  Updated with `Var-Defaults` however, not sure about format. So need agreement for format. See https://github.com/minishift/minishift/issues/1135#issuecomment-336424844.

cmd/minishift/cmd/addon/addon_apply.go
```diff
@@ -79,13 +75,13 @@ func runApplyAddon(cmd *cobra.Command, args []string) {
...
 	}
 
-	routingSuffix := determineRoutingSuffix(host.Driver)
 	sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
 	ocRunner, err := oc.NewOcRunner(minishiftConfig.InstanceConfig.OcPath, constants.KubeConfigPath)
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error applying the add-on: %s", err.Error()))
 	}
 
+	routingSuffix := determineRoutingSuffix(host.Driver)
```
hferentschik: why does this move?

pkg/minishift/addon/addon_meta.go
```diff
@@ -50,11 +52,11 @@ func NewAddOnMeta(headers map[string]interface{}) (AddOnMeta, error) {
 		switch v := headers[tag].(type) {
 		case string:
 			if v == "" {
-				return nil, errors.New(fmt.Sprintf("Metadata does not contain an mandatory entry for '%s'", tag))
+				return nil, errors.New(fmt.Sprintf("Metadata does not contain a mandatory entry for '%s'", tag))
```

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -280,3 +285,21 @@ func setStateAndPriority(addOn addon.AddOn, configMap map[string]*addon.AddOnCon
 	addOn.SetEnabled(addOnConfig.Enabled)
 	addOn.SetPriority(int(addOnConfig.Priority))
 }
+
+func addVarDefaultsToContext(context *command.ExecutionContext, varDefaults []string) error {
+	fmt.Println("varDefaults: ====", varDefaults)
+	for _, env := range varDefaults {
+		match, _ := regexp.Match(".*=.*", []byte(env))
```
hferentschik: So how is this going to work if the variable or value has a ':' itself?

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -280,3 +285,21 @@ func setStateAndPriority(addOn addon.AddOn, configMap map[string]*addon.AddOnCon
 	addOn.SetEnabled(addOnConfig.Enabled)
 	addOn.SetPriority(int(addOnConfig.Priority))
 }
+
+func addVarDefaultsToContext(context *command.ExecutionContext, varDefaults []string) error {
+	fmt.Println("varDefaults: ====", varDefaults)
```
hferentschik: I guess this needs to be removed

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -280,3 +285,21 @@ func setStateAndPriority(addOn addon.AddOn, configMap map[string]*addon.AddOnCon
...
+	for _, env := range varDefaults {
+		match, _ := regexp.Match(".*=.*", []byte(env))
+		if !match {
+			return errors.New(fmt.Sprintf("Var-Defaults variable need to be specified in the format <key>=<value>.'%s' is not.", env))
```
hferentschik: it's not '=' anymore

pkg/minishift/util/util.go
```diff
@@ -0,0 +1,26 @@
...
+limitations under the License.
+*/
+
+package util
+
+import (
+	"strings"
+)
+
+func SplitKeyValue(s string, separator string) (string, string) {
```
hferentschik: This is really a string helper function independent of Minshift. There should go into the generic util pkg. I think there is even already a strings.go

docs/source/using/addons.adoc
```diff
@@ -157,6 +157,18 @@ Multiple variables need to be comma separated.
 # Required-Vars: PROJECT_USER
 ----
 
+In case, you want to have a default value for variable name, you can achieve that with _Var-Defaults_ metadata header.
```
hferentschik: "for a variable" (no name)

hferentschik:
@budhrg If you allow mutlipe key/value pairs separated by ',', the comma gets special meaning as well. This needs handling as well. 

An alternative would be one _Var-Default_ line per default variable value one defines. This way you don't have to deal with this and the only border case is the separator between var and default value. Going down this path might need some changes in the meta data extraction though (one needs to aggregate all specified _Var-Default_ into a slice/array.

budhrg:
@hferentschik ,

> If you allow mutlipe key/value pairs separated by ',', the comma gets special meaning as well. This needs handling as well.

Are we handling this in `Required-Vars` as well? Not sure, whether there is any use case where one would put comma or `:` as part of `key`/`value`. I think documenting would help here.

budhrg:
Integration tests is taking 55 mins+ :(

budhrg:
retest this please

hferentschik:
> Not sure, whether there is any use case where one would put comma

what if you want to make the default user for some command 'system:admin' or 'system:authenticated'? I can think of multiple _oc_ commands which have a ':' in them (btw, so '=' is probably better).

>  I think documenting should be enough here.

I think we should do some sanity checking 

> Are we handling this in Required-Vars as well?

We might not, but probably should. Note though, that Required-Vars is a comma separated list of variable names. Unless you are out to break things, using ',' in the variable name is probably quite uncommon. In the case of Var-Defaults we are dealing with variable values where the range of possible values is much greater and harder to foresee.

hferentschik:
@budhrg to clarify, I don't say we necessarily need to invent now a whole escaping syntax as well, but why don't we start with some simple tests:

    # Var-Defaults: BAR=
    # Var-Defaults: ,
    # Var-Defaults: FOO=bar,,
    # Var-Defaults: FOO=BAR=SNAFU

These are just some examples. One can probably come up with a more systematic approach to border cases. I would like to see that the system does for example not panic and were appropriate returns a proper error message. In other cases I am wondering what the var/value pairs will be. In the last case for example, do I end up with variable FOO with a default value of BAR=SNAFU?

budhrg:
> Var-Defaults: FOO=BAR=SNAFU

Yes, for last case, key is `FOO` and value is `BAR=SNAFU`. as seen in the test.

budhrg:
retest this please

budhrg:
@hferentschik Added some sort of checks. 

#### Data                        Error Message
---------------------------------------------------------------------------------------------    
- `Required-Vars: USER,`        : Error applying the add-on: 'USER,' has empty key
- `Var-Defaults: =ABC`          : Error applying the add-on: '=ABC' has empty key.
- `Var-Defaults: USER=`         : Error applying the add-on: 'USER=' has empty value.
- `Var-Defaults: USER=foo,`     : Error applying the add-on: 'USER=foo,' has empty key
- `Var-Defaults: USER=foo,,`    : Error applying the add-on: 'USER=foo,' has empty key
- `Var-Defaults: USER=BAR=SNAFU`: Key is USER, Value is BAR (since [`SplitKeyValue`](https://github.com/minishift/minishift/blob/master/pkg/minishift/clusterup/clusterup.go#L210-L213) return first two elements)

budhrg:
Waiting to get more input here https://github.com/minishift/minishift/issues/1135#issuecomment-337216886

budhrg:
@jorgemoralespou  @minishift/minishift-dev This PR is ready. Please review. Also updated the PR description https://github.com/minishift/minishift/pull/1546#issue-264663235.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -358,5 +371,29 @@ func compareOpenshiftVersions(openShiftVersion, requiredOpenshiftVersion string)
...
 	}
+
+	return nil
+}
+
+func addVarDefaultsToContext(context *command.ExecutionContext, varDefaults []string) error {
+	for _, name := range varDefaults {
+		match, _ := regexp.Match(".*=.*", []byte(name))
+		if !match {
+			return errors.New(fmt.Sprintf("Var-Defaults variable need to be specified in the format of <key>=<value>.'%s' is not.", name))
```
robin-owen: Should be either "needs to be" or "must be". There should also be a space between the period and single quote containing the `%s` replacement.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -358,5 +371,29 @@ func compareOpenshiftVersions(openShiftVersion, requiredOpenshiftVersion string)
...
+		}
+
+		key, value := utilStrings.SplitKeyValue(name, "=")
+		if key == "" {
+			return errors.New(fmt.Sprintf("'%s' has empty key.", name))
+		}
+		if value == "" {
+			return errors.New(fmt.Sprintf("'%s' has empty value.", name))
```
robin-owen: "an empty key" and "an empty value"

pkg/minishift/addon/addon_meta.go
```diff
@@ -105,15 +115,18 @@ func (meta *DefaultAddOnMeta) OpenShiftVersion() string {
...
 
-func (meta *DefaultAddOnMeta) splitAndTrim(s string) []string {
+func (meta *DefaultAddOnMeta) splitAndTrim(s string) ([]string, error) {
 	// Trims the stream and then splits
 	trimmed := strings.TrimSpace(s)
 	split := strings.Split(trimmed, ",")
 	cleanSplit := make([]string, len(split))
 	for i, val := range split {
+		if val == "" {
+			return []string{}, errors.New(fmt.Sprintf("'%s' has empty key", s))
```
robin-owen: "an empty key"

docs/source/using/addons.adoc
```diff
@@ -171,7 +171,7 @@ $ minishift addons apply acme                                 // <2>
 If the environment variable is not set not, interpolation does not occur.
 <2> When the add-on is applied, each occurrence of `#{PROJECT_USER}` within an add-on command gets replaced with the value of the environment variable `USER`.
 
-As add-on developer, you can enforce that a variable value is provided when the add-on gets applied by adding the varaible name to the _Required-Vars_ metadata header.
+As add-on developer, you can enforce that a variable value is provided when the add-on gets applied by adding the variable name to the _Required-Vars_ metadata header.
```
robin-owen: "As an add-on developer", although I actually think we can remove that statement entirely. The sentence reads perfectly well without it:

```adoc
You can enforce that a variable value is provided when....
```

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,19 @@ Multiple variables need to be comma separated.
 # Required-Vars: PROJECT_USER
 ----
 
+In case, you want to have a default value for a variable, you can achieve that with _Var-Defaults_ metadata header.
```
robin-owen:
```adoc
Default values for variables may be achieved using the _Var-Defaults_ metadata header.
```

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,19 @@ Multiple variables need to be comma separated.
...
+The `Var-Default` need to be specified in the format of `<key>=<value>` where `=` is reserved character which determine the split between `key` and `value` and should not be part of either `key` or `value`.
```
robin-owen:
```adoc
`Var-Default` entries must be specified in the format of `<key>=<value>`.

[NOTE]
====
The `=` character is reserved for determining which element is a key or value and cannot be used as part of either.
====
```

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,19 @@ Multiple variables need to be comma separated.
...
+You can specify multiple default values by separating each with comma.
```
robin-owen: "with a comma"

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,19 @@ Multiple variables need to be comma separated.
...
+
+----
+# Name: acme
+# Description: ACME add-on
+# Required-Vars: PROJECT_USER
+# Var-Defaults: PROJECT_USER=john
+----
+
+NOTE: Value specified for variable in _Var-Defaults_ will be overridden by the value specified during `--addon-env` flag or `minishift config set addon-env` command.
```
robin-owen:
```adoc
[NOTE]
====
Values specified for variables in _Var-Defaults_ will be overridden by values specified with the `--addon-env` flag or `minishift config set addon-env` command.
====
```

pkg/minishift/addon/manager/addon_manager_test.go
```diff
@@ -131,3 +133,11 @@ func Test_invalid_addons_get_skipped(t *testing.T) {
 		t.Fatal(fmt.Sprintf("Unexpected number of addons. Expected %d, got %d", expectedNumberOfAddOns, actualNumberOfAddOns))
```
robin-owen: Not introduced in this PR, but this should read "add-ons", not "addons".

jorgemoralespou:
Quick question. What happens if instead of __Var-Default__ an add-on use writes __Var-default__ or even __var-default__? Silent ignore or warning.

budhrg:
> What happens if instead of Var-Default an add-on use writes Var-default or even var-default? Silent ignore or warning.

@jorgemoralespou , this is the behavior from PR. So the header will not be parsed then during apply it is asking for env 

```
$ minishift addons apply demoaddon
-- Applying addon 'demoaddon':Error applying the add-on: The variable(s) USER, NAMESPACE are required by the add-on, but are not defined in the context

# Working if variables value passed via --addon-env
$ minishift addons apply demoaddon --addon-env USER=budhram --addon-env NAMESPACE=pqr
-- Applying addon 'demoaddon':
User is budhram and namespace is pqr and project is #{PROJECT}
```

docs/source/using/addons.adoc
budhrg: Done with "As an add-on developer".

docs/source/using/addons.adoc
budhrg: > Default values for variables may be achieved using the _Var-Defaults_ metadata header.

@robin-owen , `may be` might not be good phrase here since we want to inform user it is the way to add default value.

Fixed with `Default values for variables can be achieved using the _Var-Defaults_ metadata header.`

pkg/minishift/addon/manager/addon_manager.go
budhrg: Done

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,21 @@ Multiple variables need to be comma separated.
 # Required-Vars: PROJECT_USER
 ----
 
+Default values for variables can be achieved using the _Var-Defaults_ metadata header.
```
hferentschik: "achieved" does not really fit in this context. Maybe something like:

You can set a default value for a variable using the _Var-Defaults_ metadata header.

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,21 @@ Multiple variables need to be comma separated.
...
+_Var-Defaults_ need to be specified in the format of `<key>=<value>`.
```
hferentschik: _needs_

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,21 @@ Multiple variables need to be comma separated.
...
+You can specify multiple default values by separating each with a comma.
```
hferentschik: "Multiple default key/value pairs can be separated by comma"

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,21 @@ Multiple variables need to be comma separated.
...
+
+NOTE: The `=` character is reserved for determining which element is a key or value and cannot be used as part of either.
```
hferentschik: Do we need this? Can we even determine whether it is used differently?

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,21 @@ Multiple variables need to be comma separated.
...
+
+----
+# Name: acme
+# Description: ACME add-on
+# Required-Vars: PROJECT_USER
+# Var-Defaults: PROJECT_USER=john
+----
+
+NOTE: Values specified for variables in _Var-Defaults_ will be overridden by values specified with the `--addon-env` flag or `minishift config set addon-env` command.
```
hferentschik: Variable values specified via the command line using the `--addon-env` have precedence over _Var-Defaults_

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -358,5 +371,29 @@ func compareOpenshiftVersions(openShiftVersion, requiredOpenshiftVersion string)
...
+}
+
+func addVarDefaultsToContext(context *command.ExecutionContext, varDefaults []string) error {
+	for _, name := range varDefaults {
+		match, _ := regexp.Match(".*=.*", []byte(name))
+		if !match {
+			return errors.New(fmt.Sprintf("Var-Defaults variable must be specified in the format of <key>=<value>. '%s' is not.", name))
+		}
+
+		key, value := utilStrings.SplitKeyValue(name, "=")
```
hferentschik: Why don't you use capture groups in the regexp and you get the key and value directly? I really don't get the obsession with strings.Split

pkg/minishift/addon/manager/addon_manager_test.go
```diff
@@ -128,6 +130,14 @@ func Test_invalid_addons_get_skipped(t *testing.T) {
 	expectedNumberOfAddOns := 1
 	actualNumberOfAddOns := len(addOns)
 	if actualNumberOfAddOns != expectedNumberOfAddOns {
-		t.Fatal(fmt.Sprintf("Unexpected number of addons. Expected %d, got %d", expectedNumberOfAddOns, actualNumberOfAddOns))
+		t.Fatal(fmt.Sprintf("Unexpected number of add-ons. Expected %d, got %d", expectedNumberOfAddOns, actualNumberOfAddOns))
 	}
 }
+
+func Test_add_default_envs_to_context(t *testing.T) {
```
hferentschik: So why only a test for the happy path? What's about testing the border conditions and the error cases?

pkg/minishift/addon/addon_meta.go
```diff
@@ -31,13 +31,15 @@ const (
...
+	varDefaults              = "Var-Defaults"
 )
 
 // AddOnMeta defines a set of meta data for an AddOn. Name and Description are required. Others are optional.
 type AddOnMeta interface {
 	Name() string
 	Description() []string
-	RequiredVars() []string
+	RequiredVars() ([]string, error)
+	VarDefaults() ([]string, error)
```
hferentschik: A slice of strings? Are we are not talking key/value pairs here? I expect to get a map map[string]string or a slice of a struct RequiredVar (with two elements)

hferentschik: IMO it is the metadata manager who should do the heavy lifting in terms of parsing the giving access to the metadata. This will also allow failing early in case there are syntactical issues in the metadata.

pkg/minishift/clusterup/clusterup.go
```diff
@@ -41,6 +41,7 @@ import (
 	"regexp"
 
 	"github.com/minishift/minishift/pkg/util"
+	utilStrings "github.com/minishift/minishift/pkg/util/strings"
```
hferentschik: imo not a good name. stringUtils might already be better. Or maybe minishiftStrings to make it clear that this is our won string helper

pkg/minishift/addon/manager/addon_manager.go
budhrg: I just reused what already presenthttps://github.com/minishift/minishift/blob/40c37de0cc16651d821cccf826082fff97e6cf9e/pkg/minishift/clusterup/clusterup.go#L210-L213 .

pkg/minishift/addon/addon_meta.go
```diff
@@ -31,13 +32,20 @@ const (
 	DescriptionMetaTagName   = "Description"
 	RequiredOpenShiftVersion = "OpenShift-Version"
 	anyOpenShiftVersion      = ""
+	varDefaults              = "Var-Defaults"
 )
 
+type RequiredVar struct {
```
hferentschik: :+1 Much better. You see where I was coming from?

pkg/minishift/addon/addon_meta.go
```diff
@@ -31,13 +32,20 @@ const (
...
+	Value string
+}
+
 // AddOnMeta defines a set of meta data for an AddOn. Name and Description are required. Others are optional.
 type AddOnMeta interface {
 	Name() string
 	Description() []string
-	RequiredVars() []string
+	RequiredVars() ([]string, error)
+	VarDefaults() ([]RequiredVar, error)
```

docs/source/using/addons.adoc
```diff
@@ -180,6 +180,21 @@ Multiple variables need to be comma separated.
 # Required-Vars: PROJECT_USER
 ----
 
+You can set a default value for a variable using the _Var-Defaults_ metadata header.
+_Var-Defaults_ needs to be specified in the format of `<key>=<value>`.
+Multiple default key/value pairs can be separated by comma.
+
+NOTE: The `=` character is reserved for determining which element is a key or value and cannot be used as part of either.
```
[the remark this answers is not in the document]
robin-owen: I'd avoid negation as much as we can while describing this.

```adoc
[NOTE]
====
`=` and `,` are metacharacters and cannot be used as part of keys or values.
====
```

pkg/minishift/addon/addon_meta.go
```diff
@@ -86,11 +94,36 @@ func (meta *DefaultAddOnMeta) Description() []string {
...
 	if val, contains := meta.headers[requiredVars].(string); contains {
 		return meta.splitAndTrim(val)
 	} else {
-		return []string{}
+		return []string{}, nil
+	}
+}
+
+func (meta *DefaultAddOnMeta) VarDefaults() ([]RequiredVar, error) {
+	if val, contains := meta.headers[varDefaults].(string); contains {
```
hferentschik: -1 Basically ok, but you want to do this as part of _NewAddOnMeta_. There is no need to defer this until _VarDefaults()_ is called. Also, each time you call VarDefault(), you will have to re-create it. If you do it as part of __NewAddOnMeta__ you also get this fail early approach I was talking about. You will get an error already when the add-on is read and the metadata is created. Now I would think you won't get an error until you apply the add-on.

pkg/minishift/addon/addon_meta_test.go
```diff
@@ -110,8 +110,73 @@ func Test_required_vars_empty_if_not_specified(t *testing.T) {
...
+
+type varDefaultTestCase struct {
+	expectedVarDefault string
+	expectedError      string
+}
+
+func Test_invalid_var_defaults_meta(t *testing.T) {
+	testCases := []varDefaultTestCase{
+		{
+			expectedVarDefault: "USER=foo,,",
```
hferentschik: What's about something like 'User=foo=bar'? I'd like to see a test for this as well.

pkg/minishift/addon/addon_meta_test.go
```diff
@@ -110,8 +110,73 @@ func Test_required_vars_empty_if_not_specified(t *testing.T) {
...
+			expectedError:      "'USER=foo,,' has an empty key",
```
hferentschik: The error is confusing. I think it is enough to generally say "The Var-Default value 'USER=foo,,' is not well formed." Applies for the other cases as well.

pkg/minishift/addon/addon_meta_test.go
hferentschik: What's about the empty string?

    Var-Defaults:

pkg/util/strings/strings.go
```diff
@@ -91,3 +91,9 @@ func GetOnlyNumbers(yourString string) string {
 func GetSignedNumbers(yourString string) string {
 	return getOnlyMatch(signedNumbersMatch, yourString)
 }
+
+// Split the string into two trimmed strings based on the passed separator
+func SplitKeyValue(s string, separator string) (string, string) {
```
hferentschik: Do you need this? There is also splitAndTrim in addon_meta.

docs/source/using/addons.adoc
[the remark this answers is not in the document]
budhrg: Done

pkg/util/strings/strings.go
budhrg: `splitAndTrim` doesn't accept another parameter for `separator`. I will move this into strings package as splitKeyAndValue has been used in other places as well.

pkg/minishift/addon/addon_meta_test.go
budhrg: Done

pkg/minishift/addon/addon_meta_test.go
budhrg: Done

pkg/minishift/addon/addon_meta.go
budhrg: Done and looking nice now. Thanks for suggestion :smile:

budhrg:
@hferentschik , Implemented all your suggestions. Looking nice now. Now, user will get error during install itself.
```
$ minishift addons install demoaddon
Add-on installation failed with the error: Unable to parse specified addon: 'NAMESPACE=kube-system,' is not well formed

$ minishift addons install demoaddon
Add-on installation failed with the error: Unable to parse specified addon: 'NAMESPACE=kube-system,,' is not well formed

$ minishift addons install demoaddon
Add-on installation failed with the error: Unable to parse specified addon: 'NAMESPACE=kube-system=fooo' is not well formed

$ minishift addons install demoaddon
Add-on installation failed with the error: Unable to parse specified addon: 'NAMESPACE=' is not well formed

$ minishift addons install demoaddon
Add-on installation failed with the error: Unable to parse specified addon: '=kube-system' is not well formed

$ minishift addons install demoaddon
Add-on installation failed with the error: Unable to parse specified addon: '' is not well formed

$ minishift addons install demoaddon
Add-on installation failed with the error: Unable to parse specified addon: 'NAMESPACE=kube-system,,,' is not well formed

$ minishift addons install demoaddon # For NAMESPACE=kube-system,
Addon 'demoaddon' installed
```

hferentschik:
>  Implemented all your suggestions. Looking nice now. 

Nice. I'll take another look.

> Now, user will get error during install itself.

Right, ;-) +1

hferentschik:
Rebased and applied. Made some minor changes to docs and error message.

Issue #1323, opened 2017-08-29
Use probot stale to automatically close stale issues or PRs


budhrg:
Found something interesting which can help us in cleaning stale issues with certain period like issues with no activity in last 60 days.

https://github.com/probot/stale/

Demo example - https://github.com/Linuxbrew/homebrew-core/pull/3011.

budhrg:
Or notify maintainers if issues have no activity since last 30 days and close automatically on 60 days.

hferentschik:
Sure, why not. Personally, I am more of a believer in manual grooming. Inactivity is imo not always a good indicator of things. One can try.

hferentschik:
Thinking more about it, I think we should give it a try. I think it might aid the work of the archaeologist and partly since I would like to explore these bots for another unrelated integration.

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/pull/1551

PR #1551, opened 2017-10-12
Issue #1323 Enabling and configuring the stale probot

hferentschik:
We can add this and see how it works out...

Addresses issue #1323

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
+# Issues with these labels will never be considered stale
+exemptLabels:
+  - status/pinned
+# Label to use when marking an issue as stale
+staleLabel: status/stale
+# Comment to post when marking an issue as stale. Set to `false` to disable
+markComment: >
+  This issue has been automatically marked as stale because it has not had
+  recent activity. It will be closed if no further activity occurs. Thank you
+  for your contributions.
+# Comment to post when closing a stale issue. Set to `false` to disable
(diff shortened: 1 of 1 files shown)


.github/stale.yml
```diff
@@ -0,0 +1,16 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 14
```
LalatenduMohanty: Looks like it is going to close the issue if no activity happens. But I think we should add a label i.e. `status/state` to the issue and then  we will take manual action.

LalatenduMohanty: As per the stale probot documentation there is no mention of how someone can skip the closing part. So thats an issue as if the issue gets closed then I think non-memebers can not even re-open the issue, isn't it?

hferentschik:
> Looks like it is going to close the issue if no activity happens

That's the goal in the end, right? The grace period is quite high. Also, the issues will be labeled as well, so it is easy to find the issues closed by the bot.

> But I think we should add a label i.e. status/state to the issue and then we will take manual action.

The labelling is happening.

.github/stale.yml
hferentschik: > So thats an issue as if the issue gets closed then I think non-members of the project can not even re-open the issue, isn't it?

As long as someone changes anything on the issue with 14 days, it won't get closed. There is really no different to what a human would do. 

I think we can give it a trial run. I can watch it as part of my archeologist week. If we don't like it, we change the settings or disable it completely. We can even remove all sorts of traces the bot left.

LalatenduMohanty: Sure, lets give it a try.

Issue #1547, opened 2017-10-11
Use AsciiDoc variables to refer to Minishift generically in documentation


robin-owen:
Currently, the name of the project is embedded in the AsciiDoc sources by name. To assist in making the documentation project-agnostic, I suggest adding a `project` variable to the Rake-generated **variables.adoc** files.

This would cause changes across a number of files, most notably by requiring `include::variables.adoc[]` in each file of the documentation and replacing all references to "Minishift" as a project with "{project}".

These changes seem reasonable to me -- any thoughts?

robin-owen:
While investigating this, I also noticed that the Rakefile currently in use creates a new **variables.adoc** file for each section of the documentation during the build process. Would it be more efficient to create a directory for common files in **docs/sources** which can be symbolically linked within each section?

Tossed-together example:

```sh
mkdir docs/sources/common_content && cd docs/sources
for d in contributing getting_started openshift troubleshooting using ; do
  pushd $d && ln -s ../common_content common_content && popd
done
```

From here, we could just replace "variables.adoc" with "common_content/variables.adoc" in each file (while noting in **contributing/writing_docs.adoc** the usage of **common_content**).

This would also make for a few changes in the Rakefile so that the build only creates **variables.adoc** in **common_content**.

Let me know what you think.

hferentschik:
> Currently, the name of the project is embedded in the AsciiDoc sources by name. To assist in making the documentation project-agnostic, I suggest adding a project variable to the Rake-generated variables.adoc files.

+1

> This would cause changes across a number of files, most notably by requiring include::variables.adoc[] in each file of the documentation and replacing all references to "Minishift" as a project with "{project}".

Sure. We already discussed once that we can make the project name a variable

hferentschik:
> While investigating this, I also noticed that the Rakefile currently in use creates a new variables.adoc file for each section of the documentation during the build process. 

Right, this is due to some incompatibilities between how AsciiBinder works (there is a bug report/ feature request in the AsciiBinder project) and Middleman. Really there should be only one single file and this should be referenced. However, for some reason in AsciiBinder includes must be relative to the root of the publication. We had a workaround, but had to remove it, in order to be able to have image includes as sub-directories of topics. 

We could try to go deeper into the Ruby code of Middleman and try to hack something, but the easier approach is to just write the file into each topic directory. After all it is generated and the content is defined in a single place (the Makefile).

[...]

PR #1553, opened 2017-10-12
Issue #1547 Use adoc attributes for project

robin-owen:
Add a static `project` attribute to the generated variables.adoc file and refer to Minishift as a project with this attribute -- this allows for easier project naming downstream.

Ensure variables.adoc is added to the troubleshooting section directory.

Add `minikube-iso-version` attribute to variables.adoc.

### DIFF
--- a/docs/Rakefile
+++ b/docs/Rakefile
@@ -10,6 +10,7 @@ ADOC_VARIABLES     = FileList.new('source/variables.adoc',
  'source/contributing/variables.adoc',
  'source/getting-started/variables.adoc',
  'source/openshift/variables.adoc',
+ 'source/troubleshooting/variables.adoc',
  'source/using/variables.adoc'
 )
 GENERATED_ADOC_DIR = "source/command-ref"
@@ -98,15 +99,18 @@ end
 
 desc 'Create adoc variables'
 task :adoc_variables => :init do
-  minishift_version  = ENV['MINISHIFT_VERSION'] || 'unset'
-  openshift_version  = ENV['OPENSHIFT_VERSION'] || 'unset'
-  centos_iso_version = ENV['CENTOS_ISO_VERSION'] || 'unset'
+  minishift_version    = ENV['MINISHIFT_VERSION']    || 'unset'
+  openshift_version    = ENV['OPENSHIFT_VERSION']    || 'unset'
+  centos_iso_version   = ENV['CENTOS_ISO_VERSION']   || 'unset'
+  minikube_iso_version = ENV['MINIKUBE_ISO_VERSION'] || 'unset'
 
   ADOC_VARIABLES.each do |f|
     File.open(f, "w+") do |f|
+      f.write(":project: Minishift\n")
       f.write(":minishift-version: #{minishift_version}\n")
       f.write(":openshift-version: #{openshift_version}\n")
       f.write(":centos-iso-version: #{centos_iso_version}\n")
+      f.write(":minikube-iso-version: #{minikube_iso_version}\n")
     end
   end
 end
--- a/docs/source/contributing/developing.adoc
+++ b/docs/source/contributing/developing.adoc
@@ -1,4 +1,6 @@
-= Developing Minishift
+include::variables.adoc[]
+
+= Developing {project}
 :icons:
 :toc: macro
 :toc-title:
@@ -9,7 +11,7 @@ toc::[]
 [[developing-overview]]
 == Overview
 
-The following sections describe how to build and test Minishift.
+The following sections describe how to build and test {project}.
 
 [[develop-prerequisites]]
 == Prerequisites
@@ -17,7 +19,7 @@ The following sections describe how to build and test Minishift.
 - Git
 - A recent Go distribution (>1.7)
 
-NOTE: You should be able to develop Minishift on any operating system, such as GNU/Linux, macOS or Windows.
+NOTE: You should be able to develop {project} on any operating system, such as GNU/Linux, macOS or Windows.
 The Windows operating system might require additional steps or have some limitations.
 
 [[set-up-dev-env]]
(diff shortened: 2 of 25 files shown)


budhrg:
add to whitelist

docs/Rakefile
```diff
@@ -98,15 +99,18 @@ end
 
 desc 'Create adoc variables'
 task :adoc_variables => :init do
-  minishift_version  = ENV['MINISHIFT_VERSION'] || 'unset'
-  openshift_version  = ENV['OPENSHIFT_VERSION'] || 'unset'
-  centos_iso_version = ENV['CENTOS_ISO_VERSION'] || 'unset'
+  minishift_version    = ENV['MINISHIFT_VERSION']    || 'unset'
+  openshift_version    = ENV['OPENSHIFT_VERSION']    || 'unset'
+  centos_iso_version   = ENV['CENTOS_ISO_VERSION']   || 'unset'
+  minikube_iso_version = ENV['MINIKUBE_ISO_VERSION'] || 'unset'
```
hferentschik: You are doing this for completeness reasons, right? It is not uses anywhere yet?

docs/Rakefile
```diff
@@ -98,15 +99,18 @@ end
...
 
   ADOC_VARIABLES.each do |f|
     File.open(f, "w+") do |f|
+      f.write(':project: Minishift')
```
hferentschik: +1

docs/source/contributing/developing.adoc
```diff
@@ -9,15 +11,15 @@ toc::[]
 [[developing-overview]]
 == Overview
 
-The following sections describe how to build and test Minishift.
+The following sections describe how to build and test {project}.
```
hferentschik: Do we really need this? The whole contributing topic seems pointless for the CDK. Would you for the CDK docs not ignore this altogether anyway? I guess there might be a case for consistency.

docs/source/contributing/releasing.adoc
```diff
@@ -9,7 +11,7 @@ toc::[]
 [[releasing-overview]]
 == Overview
 
-The following sections describe how to release Minishift either via automated job or manually.
+The following sections describe how to release {project} either via automated job or manually.
```
hferentschik: So here it does just not make sense anymore. This is really just about Minishift release, not CDK. TBH, I don't think we get away with a plain find&replace of all Minishift occurrences.

docs/source/openshift/openshift-client-binary.adoc
```diff
@@ -27,12 +27,12 @@ export PATH="/Users/john/.minishift/cache/oc/v1.5.0:$PATH"
 ----
 
 [[minishift-oc-context]]
-== Minishift CLI Profile
+== {project} CLI Profile
 
-As a part of the `minishift start` command, a Minishift https://docs.openshift.org/latest/cli_reference/manage_cli_profiles.html[CLI profile] is also created.
+As a part of the `minishift start` command, a {project} https://docs.openshift.org/latest/cli_reference/manage_cli_profiles.html[CLI profile] is also created.
```
hferentschik: I did not a thorough review, but here is another case where just replacing Minishift w/ CDK is probably not quite right

docs/Rakefile
robin-owen: The environment variable is already being passed along to `rake` during the build -- I assumed that adding it to the Rakefile was overlooked, and that it was going to be used in the future.

docs/source/contributing/developing.adoc
robin-owen: Yes, the CDK documentation would ignore this completely -- all of the changes made were entirely for the sake of consistency throughout the documentation.

This also means not having to think about whether you should use "Minishift" or "{project}" -- it's always "{project}" unless we're referring to the binary. If that's not desired, I'd be happy to do a more thorough PR which only replaces "Minishift" as-needed for CDK docs.

docs/source/openshift/openshift-client-binary.adoc
robin-owen: Thanks, that actually gives me a lot of information about what does and does not apply here.

I'll look at this much more thoroughly and make some changes.

robin-owen: Interestingly, CDK has this replaced with {cdk} in both the current docs branch and live documentation. Was that inaccurate to begin with?

docs/source/contributing/developing.adoc
robin-owen: To elaborate, the main point here is to provide flexibility and consistency. The `minishift` binary never changes, but the project name does. Having a split between "do I write {project} or Minishift here" when writing the documentation adds some overhead and introduces the potential for errors.

As an example, writing under **contributing/** followed by some edits to a document in **using/** could easily result in accidentally using "Minishift" where {project} is more appropriate. In addition to this, if we're using both at varying times in varying places, it becomes easy to accidentally overlook the mistake.

docs/Rakefile
hferentschik: +1

docs/source/contributing/developing.adoc
hferentschik: > This also means not having to think about whether you should use "Minishift" or "{project}" -- it's always "{project}" unless we're referring to the binary.

Sure, makes sense. I was expecting that this was the idea behind it. Just trying to highlight the pros and cons and potential issues.

As you say, if we are not going to use "{project}" consistently, people will always have a hard time to know whether to use the variable or the literal 'Minishift'

docs/source/openshift/openshift-client-binary.adoc
hferentschik: the profile is called _minishift_ regardless. So in case someone expect a literal CDK profile, he won't find one. I think the problem might be that in this context we should not use the project name at all, but use lowercase _minishift_ representing the actual name of the profile which gets created. Something like: "As a part of the `minishift start` command, a https://docs.openshift.org/latest/cli_reference/manage_cli_profiles.html[CLI profile] with the name _minishift_ is also created.

docs/Rakefile
hferentschik: There is actually a newline missing here.

hferentschik:
Rebased and merged. thanks.

Issue #1559, opened 2017-10-13
Add target to verify commit message format using git-validation 


hferentschik:

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/pull/1560

PR #1560, opened 2017-10-13
Issue #1559 Adding a commit message validation target

hferentschik:
- Integrating validate_commits into Circle CI

Issue #1559

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -13,11 +13,12 @@ Before you submit a new issue, we recommend that you search the list of issues t
 
 Thank you for your contributions! Please follow this process to submit a patch:
 
-.  Create an issue describing your proposed change to the repository.
-.  The repository owners will triage and respond to your issue promptly.
-.  Fork the repository and create a topic branch.
-.  Refer to the link:https://docs.openshift.org/latest/minishift/contributing/developing.html[developer documentation] for guidelines and tips on how to build and test Minishift.
-.  Submit a pull request with the proposed changes.
+. Create an issue describing your proposed change to the repository.
+. The repository owners will triage and respond to your issue promptly.
+. Fork the repository and create a topic branch.
+. Refer to the link:https://docs.openshift.org/latest/minishift/contributing/developing.html[developer documentation] for guidelines and tips on how to build and test Minishift.
+. Submit a pull request with the proposed changes.
+  To make reviewing and integration as smooth as possible, please run the `make` targets link:https://docs.openshift.org/latest/minishift/contributing/developing.html#unit-tests[`test`], https://docs.openshift.org/latest/minishift/contributing/developing.html#format-source-code[`fmtcheck`] and https://docs.openshift.org/latest/minishift/contributing/developing.html#commit-messages[`validate_commits`] prior to submitting the pull request.
 
 [[questions]]
 == Questions?
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,9 @@ IMAGE_UID ?= 1000
 DOCS_SYNOPISIS_DIR = docs/source/_tmp
 DOC_VARIABLES = -e OPENSHIFT_VERSION=$(OPENSHIFT_VERSION) -e MINISHIFT_VERSION=$(MINISHIFT_VERSION) -e CENTOS_ISO_VERSION=$(CENTOS_ISO_VERSION) -e MINIKUBE_ISO_VERSION=$(MINIKUBE_ISO_VERSION)
 
+# MISC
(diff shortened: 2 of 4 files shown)


hferentschik:
retest this please

Makefile
```diff
@@ -199,3 +202,7 @@ fmt:
 .PHONY: fmtcheck
 fmtcheck:
 	@gofmt -l -s $(SOURCE_DIRS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
+
+.PHONY: validate_commits
+validate_commits: $(GOPATH)/bin/git-validation
+	git-validation -q -run short-subject,message_regexp='(Issue #[0-9]+ [A-Z].*|cut v[0-9]+.[0-9]+.[0-9]+)' -range master..HEAD
```
gbraad: Do we really need to be so pedantic to disallow a commit when the message starts with a lower case ? Come on...

gbraad:
I see the same happening for the titles and I wonder how this helps:

@hferentschik changed the title from Issue #1397 preflight check for iso url to Issue #1397 Preflight check for iso url 3 days ago

Ref: https://github.com/minishift/minishift/pull/1543

Note: if you want to change the title, please make URL uppercase too...

LalatenduMohanty:
> The target should work fine for local development and for my part I am happy if at least one CI system creates an error if the commit message is not correct.

I think it is ok as this not something we should spend a lot of time.

LalatenduMohanty:
> Please update relevant documentation, such as https://github.com/minishift/minishift/blob/master/CONTRIBUTING.adoc to reflect this need

I think we should add `CONTRIBUTING.adoc` to Minishift documentation also. At this point of time it is only exists in the github repo. So when a contributor goes through documentation i.e. https://docs.openshift.org/latest/minishift/contributing/index.html does not see this.

Makefile
hferentschik: Can be changed ;-)

gbraad: I see we need the option to specify like: "Issue #xxx #xxx", so "Issue #[0-9]+" would be enough I guess.

hferentschik:
> Note: if you want to change the title, please make URL uppercase too...

Agreed

docs/source/contributing/developing.adoc
```diff
@@ -315,6 +315,20 @@ You can correct the formatting errors yourself or instruct the violations to be
 $ make fmt
 ----
 
+[[commit-messages]]
+=== Commit messages
+
+Each commit message messages needs meet two requirements:
```
robin-owen: Repeated "message messages", should be "message".

Also, "needs meet" is grammatically incorrect (should be "needs to meet") -- I also think that the wording is more concise and clear if we replace "needs to" with "must":

"Each commit message must meet two requirements:"

CONTRIBUTING.adoc
```diff
@@ -13,11 +13,12 @@ Before you submit a new issue, we recommend that you search the list of issues t
...
-.  Fork the repository and create a topic branch.
-.  Refer to the link:https://docs.openshift.org/latest/minishift/contributing/developing.html[developer documentation] for guidelines and tips on how to build and test Minishift.
-.  Submit a pull request with the proposed changes.
+. Create an issue describing your proposed change to the repository.
+. The repository owners will triage and respond to your issue promptly.
+. Fork the repository and create a topic branch.
+. Refer to the link:https://docs.openshift.org/latest/minishift/contributing/developing.html[developer documentation] for guidelines and tips on how to build and test Minishift.
+. Submit a pull request with the proposed changes.
+  To make reviewing and integration as smooth as possible, please run the targets link:https://docs.openshift.org/latest/minishift/contributing/developing.html#unit-tests[`test`], https://docs.openshift.org/latest/minishift/contributing/developing.html#format-source-code[`fmtcheck`] and https://docs.openshift.org/latest/minishift/contributing/developing.html#commit-messages[`validate_commits`] prior to submitting the pull request.
```
robin-owen: Should we clarify that these are `make` targets, or is the context enough to convey this information?

hferentschik: Will clarify

docs/source/contributing/developing.adoc
hferentschik: Thanks, good catch.

hferentschik:
> Please correct the typos in developing.adoc.

done

gbraad:
retest this please

There seems to be issues with x/crypto

gbraad:
There seems to be issues with x/crypto

# cd .; git clone https://go.googlesource.com/crypto /Users/distiller/src/golang.org/x/crypto
Cloning into '/Users/distiller/src/golang.org/x/crypto'...

gbraad:
retest this please

PR #1562, opened 2017-10-13
Issue #1557 Fix openshift registry command for profiles

### LINKED ISSUE #1557: There is no openshift registry defined after running profile different than default
See https://issues.jboss.org/browse/CDK-193.

Interesting part is when there is a default profile 'minishift' exists then  'test' profile use default profile registry.
```
[budhram@dhcp35-83:~ ] $ ./minishift profile list
- minishift Running
- test    Running (active)

[budhram@dhcp35-83:~ ] $ ./minishift status
Minishift:  Running
Profile:    test
OpenShift:  Running (openshift v3.6.0+c4dd4cf)
DiskUsage:  14% of 18G

[budhram@dhcp35-83:~ ] $ ./minishift openshift registry
172.30.1.1:5000

[budhram@dhcp35-83:~ ] $ ./minishift openshift registry --profile test
172.30.1.1:5000

[budhram@dhcp35-83:~ ] $ ./minishift profile set minishift

[budhram@dhcp35-83:~ ] $ ./minishift openshift registry
172.30.1.1:5000

[budhram@dhcp35-83:~ ] $ ./minishift stop
Cluster stopped.

[budhram@dhcp35-83:~ ] $ ./minishift profile set test
Profile 'test' set as active profile

[budhram@dhcp35-83:~ ] $ ./minishift openshift registry
No information found for 'service/docker-registry'
```

budhrg:
Fix #1557 

Also fix `minishift openshift service` command whose root cause is same.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -171,7 +171,7 @@ func runStart(cmd *cobra.Command, args []string) {
 	// preflight checks (after start)
 	preflightChecksAfterStartingHost(hostVm.Driver)
 
-	//Adding active profile information to all instance config
+	// Adding active profile information to all instance config
 	addActiveProfileInformation()
 
 	ip, _ := hostVm.Driver.GetIP()
--- a/cmd/minishift/cmd/util/oc.go
+++ b/cmd/minishift/cmd/util/oc.go
@@ -57,7 +57,7 @@ func SetOcContext(profileName string) error {
 
 	profileActions.UpdateProfileConstants(profileName)
 
-	// Need create the kube config path for the profile for ocrunner to use it.
+	// Need to create the kube config path for the profile for ocrunner to use it.
 	kubeConfigPath := filepath.Join(constants.Minipath, "machines", constants.MachineName+"_kubeconfig")
 
 	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
--- a/pkg/minishift/openshift/registry.go
+++ b/pkg/minishift/openshift/registry.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"strings"
 
+	"github.com/minishift/minishift/pkg/minikube/constants"
 	instanceState "github.com/minishift/minishift/pkg/minishift/config"
 )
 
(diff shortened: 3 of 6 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -171,7 +172,10 @@ func runStart(cmd *cobra.Command, args []string) {
 	// preflight checks (after start)
 	preflightChecksAfterStartingHost(hostVm.Driver)
 
-	//Adding active profile information to all instance config
+	// Persist instance kube config path to be used by other commands
+	minishiftConfig.InstanceConfig.KubeConfigPath = filepath.Join(constants.Minipath, "machines", constants.MachineName+"_kubeconfig")
```
LalatenduMohanty: Shouldn't we do this in PersistentPreRun in root.go. I do not see the value changing for each command and once set it can be reused by other commands.

budhrg: Make sense :+1:

cmd/minishift/cmd/root.go
```diff
@@ -103,6 +103,10 @@ var RootCmd = &cobra.Command{
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating config for VM: %s", err.Error()))
 		}
 
+		// Persist instance kube config path to be used by other commands
+		minishiftConfig.InstanceConfig.KubeConfigPath = constants.KubeConfigPath
```
LalatenduMohanty: After taking a better look, we should us `constants.KubeConfigPath` where ever `minishiftConfig.InstanceConfig.KubeConfigPath` is used currently in the PR. Both of their purpose is the same thing.  IMO `constants.KubeConfigPath` is better solution than InstanceConfig.KubeConfigPath. 
@hferentschik WDYT?

LalatenduMohanty:
@budhrg can you please also the shorten the commit message, currently is very long. I have updated it in the github and you can reuse it.

cmd/minishift/cmd/root.go
budhrg: Agree @LalatenduMohanty , same can be achieved with `constants.KubeConfigPath`. Changes done.

Issue #916, opened 2017-05-18
chcon/restorecon fail for relabeling SELinux context


praveenkumar:
Currently we are waiting for pv creation as part of our start process and during that time we also change the SELinux context and then restore it. This fails for `<pv_dir>/registry` because we don't give it `777` permission and we shouldn't also. We need to exclude this directory from chcon/restorecon section also.  A user might not notice because only visible when you enable logs.

```
# minishift start --iso-url <centos/rhel> --show-libmachine-logs --v=5
[...]

&{[-F /dev/null -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=none docker@192.168.42.213 -o IdentitiesOnly=yes -i /home/prkumar/.minishift/machines/minishift/id_rsa -p 22] /usr/bin/ssh <nil>}
About to run SSH command:
sudo which chcon; if [ $? -eq 0 ]; then chcon -R -t svirt_sandbox_file_t /var/lib/minishift/openshift.local.pv/*; fi
SSH cmd err, output: exit status 1: /bin/chcon
chcon: cannot read directory '/var/lib/minishift/openshift.local.pv/registry': Permission denied

(minishift) Calling .GetSSHHostname
(minishift) DBG | GetIP called for minishift
(minishift) DBG | Failed to retrieve dnsmasq leases from /var/lib/libvirt/dnsmasq/docker-machines.leases
(minishift) DBG | IP address: 192.168.42.213
(minishift) DBG | Unable to locate IP address for MAC 52:54:00:c3:b3:a6
(minishift) Calling .GetSSHPort
(minishift) Calling .GetSSHKeyPath
(minishift) Calling .GetSSHKeyPath
(minishift) Calling .GetSSHUsername
Using SSH client type: external
Using SSH private key: /home/prkumar/.minishift/machines/minishift/id_rsa (-rw-------)
[...]

PR #1563, opened 2017-10-13
Issue #1554 Fix restorecon failure with new persistent volume for registry

### LINKED ISSUE #1554: restorecon fails with new persistent volume for registry
These two lines, the first to chmod _pv*_ dirs and the last to restorecon, fails since there is a new persistent volume dir named _registry_.
https://github.com/minishift/minishift/blob/master/pkg/minishift/clusterup/clusterup.go#L254-L263

```
About to run SSH command:
sudo which restorecon; if [ $? -eq 0 ]; then restorecon -R /var/lib/minishift/openshift.local.pv; fi
SSH cmd err, output: <nil>: /sbin/restorecon
restorecon:  unable to read directory /mnt/sda1/var/lib/minishift/openshift.local.pv/registry
```

praveenkumar:
Fix #1554 

That was the regression, we did fix it as part of #916 but looks like forget again when did refactor.

### DIFF
--- a/pkg/minishift/clusterup/clusterup.go
+++ b/pkg/minishift/clusterup/clusterup.go
@@ -257,10 +257,10 @@ func configurePersistentVolumes(hostPvDir string, addOnManager *manager.AddOnMan
 	// if we have SELinux enabled we need to sort things out there as well
 	// 'cluster up' does this as well, but we do it here as well to have all required actions collected in one
 	// place, instead of relying on some implicit knowledge on what 'cluster up does (HF)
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
@praveenkumar @gbraad @hferentschik @budhrg Do we know how the `wip` bot is added to `Minishift` repo.

praveenkumar:
@LalatenduMohanty no idea but I do like that also 👍

hferentschik:
> Do we know how the wip bot is added to Minishift repo.

Yes :-)

Issue #1579, opened 2017-10-17
Integration log name contains colon


agajdosi:
With change of introducing timestamp into integration log's file name we overlooked that it contains a colon, which is not allowed symbol on Windows 7 and 10. Thus tests on these platforms do not create the log, or fails strictly - from my understandings Win7 just do not create the file, while Win10 fails the tests.

Current format:
```go
logFileName := fmt.Sprintf("integration_%d-%d-%d_%02d:%02d:%02d.log", t.Year(), t.Month(),
t.Day(), t.Hour(), t.Minute(), t.Second())
```

As fix we could probably just change colon for a hyphen resulting it:
```go
logFileName := fmt.Sprintf("integration_%d-%d-%d_%02d-%02d-%02d.log", t.Year(), t.Month(),
t.Day(), t.Hour(), t.Minute(), t.Second())
```

hferentschik:
merged by pull request #1580

PR #1580, opened 2017-10-17
Issue #1579 Remove colons from filename of integration test log

agajdosi:
Quick fix for #1579 which causes problems on Windows machines.


hferentschik:
Merged, thanks.

Issue #1592, opened 2017-10-20
Unclear display of active profile


hferentschik:
Currently a `minishift profile list` displays like:

    $ minishift profile list
     - minishift	Stopped
     - bar		Does Not Exist
     - msa		Stopped	(active)
     - snafu        Does Not Exist

The active marker '(active)' disappears in the VM status message. Instead of appended to the status, it should display in tabular form in its own column.

anjannath:
@hferentschik WDYT about this??
```
+-----------+----------------+--------+
|  PROFILE  |   VM STATUS    | STATE  |
+-----------+----------------+--------+
| minishift | Stopped        |        |
| anjan     | Stopped        | Active |
| foo       | Does Not Exist |        |
+-----------+----------------+--------+
```

gbraad:
Please, let's not start using OpenStack-like table listings...

Just use a tab writer... and always place active at the top, and provide the rest as sorted added to the end?

hferentschik:
> Just use a tab writer... and always place active at the top, and provide the rest as sorted added to the end?

+1. I am not a fan of this either. Tab writer will do just fine.

anjannath:
@gbraad @hferentschik I thought the tabular form looked more pleasing, but using a tabwriter now. Also i have to iterate over the `profiles` slice twice to have a sorted output, you guys can review that in the PR and suggest a nicer solution.

LalatenduMohanty:
@hferentschik so I think we just needs to add more tabs before the `(active)` string.  I think we should sort the profile names before printing but not sure if we want to print the active profile as the first one. cc @gbraad @anjannath

anjannath:
@LalatenduMohanty Printing the active profile first  makes sense to me, because then it'll attract more attention, and that's what we are trying to do.

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/pull/1599

PR #1599, opened 2017-10-23
Issue #1592 Making sure output of profile list appears in columns

anjannath:


anjannath:
it seems the integration tests are failing because of the tab before (Active)??

cmd/minishift/cmd/profile/profile_list.go
```diff
@@ -44,7 +45,15 @@ func displayProfiles(profiles []string) {
...
 		vmStatus := cmdUtil.GetVMStatus(profile)
 		if profile == activeProfile {
-			fmt.Fprintln(display, fmt.Sprintf("- %s\t%s\t(active)", profile, vmStatus))
+			fmt.Fprintln(display, fmt.Sprintf("- %s\t%s\t%s", profile, vmStatus, "(Active)"))
+		}
+	}
+
+	sort.Strings(profiles)
+	for _, profile := range profiles {
+		vmStatus := cmdUtil.GetVMStatus(profile)
```
LalatenduMohanty: -1  two for loops.  We need to add more tab space before `(active)` and sort the profiles . Not sure if we want to add active profile as the first one https://github.com/minishift/minishift/issues/1592#issuecomment-338601379
In case you want to print the active profile first, you can just print the VM status of active profile (of the profile returned by profileActions.GetActiveProfile()) then in the for loop do not print anything for the active profile.

anjannath: @LalatenduMohanty Thanks, i also don't like the two loops, was thinking of how to avoid that.

budhrg:
Integration tests need to be updated here. Plz check the failure @anjannath

gbraad:
Please also consider updating to commit message to include the output is sorted and displayed in columns

hferentschik:
> Please also consider updating to commit message to include the output is sorted and displayed in columns

+1

LalatenduMohanty:
Merged. Thanks.

Issue #1600, opened 2017-10-23
Improve 'Using Minishift' documentation 


hferentschik:
There are a few things I'd like to address in order to then add documentation for #952.

* Rename "Managing Minishift" to "Basic Commands". Naming is ambiguous 
* Extract image caching and experimental features in their own topics
* Adjust cross-links
* Align header styles, eg headers should use present continuous

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/pull/1601

PR #1590, opened 2017-10-19
Issue #403 Adding architecture diagram to Minishift usage topic

hferentschik:
Took a stab at a simple diagram using ditaa which is also used within the Origin docs. It ain't super pretty, but imo does a pretty good job visualizing the high-level components. Thoughts?

minishift-architecture

Issue #1485, opened 2017-10-04
Improve profile documentation 


hferentschik:
The profile documentation is currently very short and basically just list briefly the existing commands. We should explain the feature in more detail, providing background and context as well as some more examples.

LalatenduMohanty:
We need to add that each profile have a separate home directory and what it means i.e. a separate config.jason files, separate cache etc. for each profile. Also if an user runs any command (e.g. config set) it by default runs against the current active profile. An example if how user can set globla config using environment variables at this moment. Also what is a typical work flow user can follow.

robin-owen:
@hferentschik I'd love to help with this. Context is essential, especially when explaining a feature which is largely conceptual: "what is a profile, why would I want to use one, what can I do with it?"

> We need to add that each profile have a separate home directory and what it means i.e. a separate config.jason files, separate cache etc. for each profile.

@LalatenduMohanty Do you think it would also make sense to explain files which exist in **$MINISHIFT_HOME**, but _not_ in the "home" for each respective profile (for example, **_allinstances.json_**)?

> Also what is a typical work flow user can follow.

I would like to have more information about this. If I don't know what a typical workflow for using profiles looks like, I can't accurately write about it.

LalatenduMohanty:
> @LalatenduMohanty Do you think it would also make sense to explain files which exist in $MINISHIFT_HOME, but not in the "home" for each respective profile (for example, allinstances.json)?

@robin-owen I do not think we need to explain the directory structure as it might be overkill for the user. 

Also I am planning to put the PR and you can review it.

robin-owen:
> I do not think we need to explain the directory structure as it might be overkill for the user.

Agreed, a complete breakdown would be overkill -- I simply meant to mention the few files that exist in **$MINISHIFT_HOME** which all profiles make use of (I'm not sure if there are any more files than **_allinstances.json_** which fit this description).

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/pull/1605

PR #1601, opened 2017-10-23
Issue #1600 Restructuring the usage topic

hferentschik:
- Rename "Managing Minishift" to "Basic Commands". Naming is ambiguous
- Extract image caching and experimental features in their own topics
- Adjust cross-links
- Align header styles, eg headers should use present continuous

### DIFF
--- a/docs/source/_topic_map.yml
+++ b/docs/source/_topic_map.yml
@@ -25,8 +25,10 @@ Topics:
     Topics:
       - Name: Overview
         File: index
-      - Name: Managing Minishift
-        File: managing-minishift
+      - Name: Basic Usage
+        File: basic-usage
+      - Name: Image Caching
+        File: image-caching
       - Name: Add-ons
         File: addons
       - Name: Host Folders
@@ -35,6 +37,8 @@ Topics:
         File: docker-daemon
       - Name: Profiles
         File: profiles
+      - Name: Experimental Features
+        File: experimental-features
   - Name: Interacting with OpenShift
     Dir: openshift
     Topics:
--- a/docs/source/getting-started/quickstart.adoc
+++ b/docs/source/getting-started/quickstart.adoc
@@ -13,7 +13,7 @@ toc::[]
 == Overview
 
 This section contains a brief demo of {project} and of the provisioned OpenShift cluster.
-For details on the usage of {project}, see the xref:../using/managing-minishift.adoc#[Managing {project}] section.
+For details on the usage of {project}, see the xref:../using/basic-usage.adoc#[Basic Usage] section.
 
 The interaction with OpenShift is with the command-line tool _oc_ which is copied to your host.
 For more information on how {project} can assist you in interacting with and configuring your local OpenShift instance, see the xref:../openshift/openshift-client-binary.adoc#[OpenShift Client Binary] section.
--- a/docs/source/openshift/index.adoc
+++ b/docs/source/openshift/index.adoc
@@ -6,8 +6,8 @@ include::variables.adoc[]
 {project} creates a virtual machine and provisions a local, single-node OpenShift cluster in this VM.
 The following sections describe how {project} can assist you in interacting and configuring your local OpenShift cluster.
 
-For details about managing the {project} VM, see the xref:../using/managing-minishift.adoc#[Managing {project}] section.
-
 - xref:../openshift/openshift-client-binary.adoc#[Using the OpenShift Client Binary]
 - xref:../openshift/exposing-services.adoc#[Exposing Services]
 - xref:../openshift/openshift-docker-registry.adoc#[Accessing the OpenShift Docker Registry]
+
+For details about the basic commands to interact with the {project} VM, see the xref:../using/basic-usage.adoc#[Basic Usage] section.
--- a/docs/source/using/basic-usage.adoc
+++ b/docs/source/using/basic-usage.adoc
@@ -1,6 +1,6 @@
 include::variables.adoc[]
 
-= Managing {project}
+= Basic Usage
 :icons:
 :toc: macro
 :toc-title:
@@ -8,7 +8,7 @@ include::variables.adoc[]
 
 toc::[]
 
-[[managing-minishift-overview]]
+[[basic-usage-overview]]
 == Overview
 
 When you use {project}, you interact with two components:
@@ -226,31 +226,6 @@ CAUTION: Driver-specific options might overlap with values specified using {proj
 Examples are boot2docker URL, memory size, cpu count, and so on.
 In this case, driver-specific environment variables will override {project}-specific settings.
 
-[[caching-openshift-images]]
-== Caching OpenShift images (experimental)
-
-To speed up provisioning of the OpenShift cluster and to minimize network traffic, the core OpenShift images can be cached on the host.
-This feature is considered experimental and needs to be explicitly enabled using the xref:../command-ref/minishift_config_set#[`minishift config set`] command:
-
-----
-$ minishift config set image-caching true
-----
-
-Once enabled, caching occurs transparently, in a background process, the first time you use the xref:../command-ref/minishift_start#[`minishift start`] command.
-Once the images are cached under _$MINISHIFT_HOME/cache/images_, successive {project} VM creations will use these cached images.
-
-Each time an image exporting background process runs, a log file is generated under _$MINISHIFT_HOME/logs_ which can be used to verify the progress of the export.
-
-You can disable the caching of the OpenShift images by setting `image-caching` to `false` or removing the setting altogether using xref:../command-ref/minishift_config_unset#[`minishift config unset`]:
-
-----
-$ minishift config unset image-caching
-----
-
-NOTE: Image caching is considered experimental and its semantics and API is subject to change.
-The aim is to allow caching of arbitrary images, as well as using a better format for storing the images on the host.
-You can track the progress on this feature on the GitHub issue link:https://github.com/minishift/minishift/issues/952[#952].
-
 [[persistent-volumes]]
 == Persistent Volumes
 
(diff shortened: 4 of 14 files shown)


docs/source/using/docker-daemon.adoc
```diff
@@ -6,11 +6,6 @@ include::variables.adoc[]
 :toc-title:
 :toclevels: 1
 
-toc::[]
```
budhrg: @hferentschik Why to remove toc from here? It is creating inconsistency from other pages

docs/source/_topic_map.yml
```diff
@@ -25,8 +25,10 @@ Topics:
     Topics:
       - Name: Overview
         File: index
-      - Name: Managing Minishift
-        File: managing-minishift
+      - Name: Basic commands
```
robin-owen: Please use title casing: "Basic Commands".

docs/source/getting-started/quickstart.adoc
```diff
@@ -13,7 +13,7 @@ toc::[]
 == Overview
 
 This section contains a brief demo of {project} and of the provisioned OpenShift cluster.
-For details on the usage of {project}, see the xref:../using/managing-minishift.adoc#[Managing {project}] section.
+For details on the usage of {project}, see the xref:../using/basic-commands.adoc#[Managing {project}] section.
```
robin-owen: The section was renamed to "Basic Commands", yet the `xref` is still named "Managing {project}" here.

docs/source/openshift/index.adoc
```diff
@@ -6,7 +6,7 @@ include::variables.adoc[]
 {project} creates a virtual machine and provisions a local, single-node OpenShift cluster in this VM.
 The following sections describe how {project} can assist you in interacting and configuring your local OpenShift cluster.
 
-For details about managing the {project} VM, see the xref:../using/managing-minishift.adoc#[Managing {project}] section.
+For details about managing the {project} VM, see the xref:../using/basic-commands.adoc#[Managing {project}] section.
```
robin-owen: The `xref` here is still named "Managing {project}".

docs/source/using/basic-commands.adoc
```diff
@@ -1,14 +1,14 @@
 include::variables.adoc[]
 
-= Managing {project}
+= Basic commands
```
robin-owen: Title casing, please: "Basic Commands".

docs/source/using/profiles.adoc
```diff
@@ -15,7 +15,7 @@ toc::[]
 Each {project} instance can be created with different configurations and users can switch between different instances.
 
 [[create-profile]]
-== Create profile
+== Creating profiles
```
robin-owen: Title casing for each heading in this document, please:

"Creating Profiles"
"Listing Existing Profiles"
"Switching Profiles"
"Deleting a Profile"

As a side note, the use of title case seems inconsistent across the documentation. I'll open a separate issue for this.

docs/source/using/experimental-features.adoc
```diff
@@ -0,0 +1,89 @@
...
+To make this work you need to create a Virtual Switch using NAT
+
+- link:https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/setup-nat-network[Set up a NAT network]
+
+[NOTE]
+====
+WinNAT is limited to one NAT network per host. For more details about capabilities, and limitations, please see the link:https://docs.microsoft.com/en-us/virtualization/hyper-v-on-windows/user-guide/setup-nat-network[WinNAT capabilities and limitations blog]
+====
+
+The following command will attempt to assign an IP address for use on the Internal Virtual Switch 'MyInternal'.
```
robin-owen: This sentence should end in a colon.

docs/source/using/experimental-features.adoc
```diff
@@ -0,0 +1,89 @@
...
+PS> $env:MINISHIFT_ENABLE_EXPERIMENTAL="y"
+PS> $env:HYPERV_VIRTUAL_SWITCH="MyInternal"
+PS> minishift.exe start `
+  --iso-url centos `
+  --network-ipaddress 192.168.1.10 `
+  --network-gateway 192.168.1.1 `
+  --network-nameserver 8.8.8.8
+----
+
+If you want to use the 'DockerNAT' network, the following commands are needed to setup the correct NAT networking and assigning an IP in the range expected.
```
robin-owen: This sentence should also end in a colon.

docs/source/using/experimental-features.adoc
```diff
@@ -0,0 +1,89 @@
...
+For Hyper-V a functionality is provided to assign an IP address on startup using the Data Exchange Service.
+
+[IMPORTANT]
+====
+-  While the default image is B2D, this only works with the CentOS/RHEL based image in combination with Hyper-V. The B2D image experiences
+   a problem when the values are being send to the {project} instance and consumed by the B2D iso. We are looking into the issue and hope
+   to provide a solution in the coming future.
+====
+
+To make this work you need to create a Virtual Switch using NAT
```
robin-owen: The lack of ending punctuation and use of a single-item list here feels awkward. Can we incorporate the link directly into this sentence, remove the list, and add the appropriate punctuation?

docs/source/using/docker-daemon.adoc
```diff
@@ -6,11 +6,6 @@ include::variables.adoc[]
...
-
-[[reusing-docker-daemon]]
-== Reusing the Docker Daemon
```
robin-owen: I imagine both this heading and the table of contents were removed due to this document only having a single heading/topic. This is inconsistent with the rest of the documentation, though, and also introduces more work later if another heading needs to be added to this document in the future.

Personally, I'd prefer to keep `toc::[]` and the heading even if it seems like overkill for a single topic.

docs/source/using/experimental-features.adoc
```diff
@@ -0,0 +1,89 @@
...
+PS> minishift.exe start `
+  --iso-url centos `
+  --network-ipaddress 10.0.75.128 `
+  --network-gateway 10.0.75.1 `
+  --network-nameserver 8.8.8.8
+----
+
+[NOTE]
+====
+- Be sure to specify a valid gateway and nameserver. Failing to do so will result in connectivity issuesi
```
robin-owen: Typo here: "issuesi".

docs/source/using/docker-daemon.adoc
hferentschik: Because it is not needed. Consistency is one thing, but it must make sense. The topic is too short for a TOC to make sense. As the topic might expand we can re-add it.

docs/source/_topic_map.yml
hferentschik: Ahh, of course. Will fix

docs/source/getting-started/quickstart.adoc
hferentschik: Good catch. Let me fix that.

docs/source/openshift/index.adoc
hferentschik: ok

docs/source/using/profiles.adoc
hferentschik: I noticed this as well. In fact I had this already changed in a different branch as well. Will align in this pull request as well.

docs/source/using/experimental-features.adoc
hferentschik: ok

docs/source/using/experimental-features.adoc
hferentschik: ok

docs/source/using/docker-daemon.adoc
hferentschik: For me it is really about the L&F. IMO it does not look good w/. Adding a [toc] when required is trivial.

docs/source/using/experimental-features.adoc
hferentschik: ok

docs/source/using/docker-daemon.adoc
robin-owen: Understood, and I agree that it feels somewhat awkward.

In this case, I like the consistency and the fact that adding a new header would require no more effort than it would in any other topic within the documentation. Removing the TOC and header means that we'll have to remember that this one particular part of the documentation is different. There's also the `xref` in **_contributing/writing_docs.adoc_** to keep in mind, which would also have to be updated if we add the TOC and header back in at any point.

I just think that it would make more sense to avoid potential errors in the future. I don't feel particularly strongly about this, though, so if anyone else agrees that we should remove the TOC from this particular topic, I don't mind.

docs/source/using/experimental-features.adoc
hferentschik: +1

docs/source/using/docker-daemon.adoc
hferentschik: Let me see what I can do. Seems others feel similar :-)

docs/source/using/docker-daemon.adoc
hferentschik: I see what I can do.

hferentschik:
@robin-owen 

> Overall, I like this a lot. There are a few relatively small changes that I would like to see, 

Cool. I applied the suggestions you made. Let me know whether I missed something.

> though, and I'm not convinced that we should remove the table of contents and header from using/docker-daemon.adoc.

ok, ok - I got them back. Same for image caching. I tried to come up with an additional heading on top of the overview which makes sense. For the caching section, it is really not an issue, since a major update is coming there in the context of issue #952. I guess @budhrg will like this as well :-)

hferentschik:
I'd prefer to apply this pull request first. Then I can rebase #1590 as well. I think overall we get some quite nice improvements in the docs. Then we would just be missing a re-write of the profiles docs (#1485).

docs/source/getting-started/quickstart.adoc
```diff
@@ -13,7 +13,7 @@ toc::[]
 == Overview
 
 This section contains a brief demo of {project} and of the provisioned OpenShift cluster.
-For details on the usage of {project}, see the xref:../using/managing-minishift.adoc#[Managing {project}] section.
+For details on the usage of {project}, see the xref:../using/basic-commands.adoc#[Basic commands] section.
```
robin-owen: Title case for "Basic Commands", please.

docs/source/using/docker-daemon.adoc
```diff
@@ -8,13 +8,16 @@ include::variables.adoc[]
...
+[[docker-daemon-overview]]
+== Overview
 
 When running OpenShift in a single VM, you can reuse the Docker daemon managed by {project} for other Docker use-cases as well.
 By using the same docker daemon as {project}, you can speed up your local development.
 
-In order to configure your console to reuse the {project} Docker dameon, follow these steps:
+[[docker-daemon-configuration]]
+== Console configuration
```
robin-owen: Title case: "Console Configuration"

docs/source/using/basic-commands.adoc
```diff
@@ -1,14 +1,14 @@
...
+= Basic Commands
 :icons:
 :toc: macro
 :toc-title:
 :toclevels: 2
 
 toc::[]
 
-[[managing-minishift-overview]]
+[[basic-commands-overview]]
```
LalatenduMohanty: -1 for naming this as`basic-commands` because topics like 
```
Runtime Options
Persistent Volumes
HTTP/HTTPS Proxies
Networking
```

are not basic commands IMO. Also I do not find it intuitive. If you want to have `basic commands` section we should split the doc in to two parts i.e. 1. Basic commands 2. Managing minishift.

hferentschik: > -1 for naming this asbasic-commands because topics like

I want to get away from 'managing minishift'. I think this is a bad term regardless. I think the major parts of basic commands is exactly that - start/stop/config/ssh. 

Persistence volumes could be split out. Networking can be removed.

robin-owen: Very good point, I happened to overlook this while reviewing the changes. I think it's also worth considering moving Experimental Features into its own topic under Using Minishift at some point.

I think that splitting the documentation between Basic Commands and Managing Minishift makes sense -- ultimately, these are different concepts.

hferentschik: Except "Managing Minishift" is imo a misnomer anyway. Let's try to get away from it.

robin-owen: Yes, it's not an ideal name, I agree. It's particularly vague.

hferentschik: IMO the current change is already a step into the right direction. "Basic usage" might not be perfect, but it is better than "Managing Minishift". I have some more ideas and can do some further refinements as part of #952.  For now, I think there is enough value in this for a merge.

LalatenduMohanty: @hferentschik @robin-owen I do not have a better suggestion for `Managing Minishift` at this moment. But I would suggest we split-up the actual content in to two parts. Changing page names might cause difficulty for users as they are familiar with certain structure or the docs. So we should only change the name when the new change makes the documentation more intuitive to navigate.

hferentschik: > But I would suggest we split-up the actual content in to two parts. 

Into what? If you have start/stop/config/ssh under "basic", what do you place the other ones under? "Managing" does not work really, since the remaining sections would just be some informational tidbits.

I am intrigued why you think "Basic Commands" is so much worse than "Managing Minishift". In both cases you will have to actually at least look at the TOC to get a better understanding of what's covered. For me it sounds natural to start a topic like "Using Minishift" with "Basic Commands". Maybe it is the "Commands" part which you don't like? We could just use "Basics". The following topics would then be covering specific, more advanced features.

robin-owen: For the time being, "Basic Usage" seems like an appropriate name to me.

budhrg: Basic Usage is reasonable

hferentschik: Ok, "Basic Usage" it is then ;-) Let me adjust the PR...

hferentschik:
@robin-owen, @budhrg - I switched to "Basic Usage". I think this is a quite good fit now. IT nicely compares with the other topics which are specific more advanced topics. 

@LalatenduMohanty Does this work for you now?

docs/source/using/basic-usage.adoc
```diff
@@ -1,14 +1,14 @@
...
+= Basic Usage
 :icons:
 :toc: macro
 :toc-title:
 :toclevels: 2
 
 toc::[]
 
-[[managing-minishift-overview]]
+[[basic-commands-overview]]
```
budhrg: Let's use `basic-usage-overview` to match with header name.

docs/source/using/image-caching.adoc
```diff
@@ -0,0 +1,37 @@
+include::variables.adoc[]
+
+= Image Caching
+:icons:
+:toc: macro
+:toc-title:
+:toclevels: 2
+
```
budhrg: `toc::[]` is  missing here to display horizontal bar.

screenshot from 2017-10-24 15-04-23

docs/source/using/basic-usage.adoc
hferentschik: true true

docs/source/using/image-caching.adoc
hferentschik: let me check ...

docs/source/using/basic-usage.adoc
hferentschik: fixed

docs/source/_topic_map.yml
```diff
@@ -25,8 +25,10 @@ Topics:
     Topics:
       - Name: Overview
         File: index
-      - Name: Managing Minishift
-        File: managing-minishift
+      - Name: Basic Usage
+        File: basic-usage
```
LalatenduMohanty: 👍

hferentschik: :-)

hferentschik:
Cheers

PR #1605, opened 2017-10-24
Issue #1485 Improving the profile documentation

LalatenduMohanty:

### DIFF
--- a/docs/source/using/profiles.adoc
+++ b/docs/source/using/profiles.adoc
@@ -8,30 +8,80 @@ include::variables.adoc[]
 
 toc::[]
 
-[[pofile-overview]]
+[[profile-overview]]
 == Overview
 
-{project} allows to create multiple instances of {project}.
-Each {project} instance can be created with different configurations and users can switch between different instances.
+A profile is an instance of the Minishift VM along with all of its configuration and state.
+The profile feature allows you to create and manage these isolated instances of Minishift.
 
-[[creating-profile]]
+Each {project} profile is created with its own configuration (memory, CPU, disk size, add-ons, and so on) and completely independent of other profiles.
+See xref:../using/basic-usage.adoc#environment-variables[environment variables], if you want to apply some common configurations (CPU, memory etc) to all profiles.
+
+The current profile being used is called the active profile.
+Unless the global `--profile` flag is used, {project} commands are executed against the active profile.
+You can determine the active profile by using the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command. 
+You can execute single commands against a non-active profile, by using the `--profile` flag.
+Example:
+
+----
+$ minishift --profile my-profile console --url
+----
+
+On top of the `--profile` flag, there are xref:../command-ref/minishift_profile.adoc#[commands] for listing, deleting and setting the active profile.
+These commands are described in the following sections.
+
+[[creating-profiles]]
 == Creating Profiles
 
-You can create a new instance of {project} with xref:../command-ref/minishift_start.adoc#[`minishift start --profile`] command.
+A profile is created when the configuration files associated with it gets created.
+There are two ways to create a new profile.
 
-NOTE: The default profile is `minishift`. It will be present by-default and user does not need to create it specifically.
+==== Using The `--profile` Flag
 
-To create a new profile, run:
+When you run any {project} command with the `--profile` flag the profile gets created if it does not exist.
+
+Example:
+----
+$ minishift --profile PROFILE_NAME config set PROPERTY_NAME PROPERTY_VALUE
+----
+
+=== Using `profile set` Command
+
+You can run the `profile set` command for a non existing profile to set it as the active profile.
+The profile is automatically created when you subsequently run any `minishift` command.
+See xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`].
+
+Example:
+
+----
+$ minishift profile set PROFILE_NAME
+----
+
+[NOTE]
+====
+The default profile is *minishift*.
+It will be present by default and it does not need to be created.
+====
+
+[[starting-a-profile]]
+== Starting A Profile
+
+You can start the active profile with xref:../command-ref/minishift_start.adoc#[`minishift start`] command.
+To start a non active profile, you need to use the `--profile` flag.
+
+To start a non active profile, run:
 
 ----
 $ minishift --profile PROFILE_NAME start
 ----
 
+NOTE: A profile is automatically set to active profile when the VM instance is started successfully.
+
 [[listing-profiles]]
 == Listing Profiles
 
-You can list all existing profiles with xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command.
-You can also see the active profile in the output.
+You can list all existing profiles with the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command.
+You can also see the active profile is highlighted in the output.
 
 To list profiles, run:
 
@@ -39,20 +89,17 @@ To list profiles, run:
 $ minishift profile list
 ----
 
-
 [[switching-profiles]]
 == Switching Profiles
 
-With the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] you can switch between profiles or set a profile as active.
-When a profile is active, it is not required to pass `--profile` flag to run minishift commands against that profile.
+You can switch between profiles with the xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] command.
+Only one profile can be active at any time.
 
-To switch between profiles or to set a profile as active, run:
+To switch between profiles, run:
 
 ----
 $ minishift profile set PROFILE_NAME
 ----
-
-When you have multiple profiles you can use `--profile` flag with any minishift command to run the command against a specific profile.
 
 [[deleting-profiles]]
 == Deleting Profiles
@@ -63,4 +110,28 @@ To delete a profile, run:
 $ minishift profile delete PROFILE_NAME
 ----
 
-NOTE: The default profile 'minishift'  can not be deleted.
+NOTE: The default profile *minishift* cannot be deleted.
+
+[[example-workflows-profiles]]
+== Example Workflows
+
+To create a new profile with specific configurations, run:
+
+----
+$ minishift profile set foo
+$ minishift config set memory 4GB
+$ minishift config set cpus 2
+$ minishift addon enable anyuid
+$ minishift start
+----
+
+The alternative is to execute a series of commands each specifying the targeted profile explicitly using the `--profile` flag.
+
+Example:
+
+----
+$ minishift --profile foo config set memory 4GB
+$ minishift --profile foo config set cpus 2
+$ minishift --profile foo addon enable anyuid
+$ minishift --profile foo minishift start
+----
--- a/docs/source/using/profiles.adoc
+++ b/docs/source/using/profiles.adoc
@@ -15,46 +15,54 @@ A profile is an instance of the Minishift VM along with all of its configuration
 The profile feature allows you to create and manage these isolated instances of Minishift.
 
 Each {project} profile is created with its own configuration (memory, CPU, disk size, add-ons, and so on) and completely independent of other profiles.
-See xref:../using/basic-usage.adoc#environment-variables[environment variables], if you want to apply some common configurations (CPU, memory etc) to all profiles.
+Refer to the use of xref:../using/basic-usage.adoc#environment-variables[environment variables] if you want to make sure that certain configuration settings, for example `cpus` or `memory`, get applied to all profiles.
 
-The current profile being used is called the active profile.
-Unless the global `--profile` flag is used, {project} commands are executed against the active profile.
-You can determine the active profile by using the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command. 
-You can execute single commands against a non-active profile, by using the `--profile` flag.
-Example:
+The _active_ profile is the profile against which all commands are executed, unless the global `--profile` flag is used.
+You can determine the active profile by using the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command.
+You can execute single commands against a non-active profile by using the `--profile` flag, for example `minishift --profile profile-demo console` to open the OpenShift console for the specified _profile-demo_ profile.
 
-----
-$ minishift --profile my-profile console --url
-----
-
-On top of the `--profile` flag, there are xref:../command-ref/minishift_profile.adoc#[commands] for listing, deleting and setting the active profile.
+On top of the `--profile` flag, there are xref:../command-ref/minishift_profile.adoc#[commands] for xref:../command-ref/minishift_profile_list.adoc#[listing], xref:../command-ref/minishift_profile_delete.adoc#[deleting] and xref:../command-ref/minishift_profile_set.adoc#[setting] the active profile.
 These commands are described in the following sections.
 
 [[creating-profiles]]
 == Creating Profiles
 
-A profile is created when the configuration files associated with it gets created.
 There are two ways to create a new profile.
 
 ==== Using The `--profile` Flag
 
-When you run any {project} command with the `--profile` flag the profile gets created if it does not exist.
-
-Example:
-----
-$ minishift --profile PROFILE_NAME config set PROPERTY_NAME PROPERTY_VALUE
-----
-
-=== Using `profile set` Command
-
-You can run the `profile set` command for a non existing profile to set it as the active profile.
-The profile is automatically created when you subsequently run any `minishift` command.
-See xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`].
-
-Example:
+When you run any {project} command with the `--profile` flag the profile gets created if it does not exist, for example:
 
 ----
-$ minishift profile set PROFILE_NAME
+$ minishift --profile profile-demo start
+-- Checking if requested hypervisor 'xhyve' is supported on this platform ... OK
+-- Checking if xhyve driver is installed ...
+   Driver is available at /usr/local/bin/docker-machine-driver-xhyve
+   Checking for setuid bit ... OK
+-- Checking the ISO URL ... OK
+-- Starting local OpenShift cluster using 'xhyve' hypervisor ...
+-- Minishift VM will be configured with ...
+   Memory:    2 GB
+   vCPUs :    2
+   Disk size: 20 GB
+...
+----
+
+See also xref:../using/profiles.adoc#example-workflows-profiles[workflow for profile configuration].
+
+[NOTE]
+====
+ A profile automatically becomes the active profile when a {project} instance is started successfully via `minishift start`.
+====
+
+=== Using the `profile set` Command
+
+The other option to create a profile is to use the `profile set` command.
+If the specified profile does not exist, it is implicitly created:
+
+----
+$ minishift profile set demo
+Profile 'demo' set as active profile
 ----
 
 [NOTE]
@@ -63,43 +71,29 @@ The default profile is *minishift*.
 It will be present by default and it does not need to be created.
 ====
 
-[[starting-a-profile]]
-== Starting A Profile
-
-You can start the active profile with xref:../command-ref/minishift_start.adoc#[`minishift start`] command.
-To start a non active profile, you need to use the `--profile` flag.
-
-To start a non active profile, run:
-
-----
-$ minishift --profile PROFILE_NAME start
-----
-
-NOTE: A profile is automatically set to active profile when the VM instance is started successfully.
-
 [[listing-profiles]]
 == Listing Profiles
 
 You can list all existing profiles with the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command.
-You can also see the active profile is highlighted in the output.
-
-To list profiles, run:
+You can also see the active profile highlighted in the output.
 
 ----
 $ minishift profile list
+- minishift     Running     	(Active)
+- profile-demo  Does Not Exist
 ----
 
 [[switching-profiles]]
 == Switching Profiles
 
-You can switch between profiles with the xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] command.
-Only one profile can be active at any time.
-
-To switch between profiles, run:
+To switch between profiles use the xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] command:
 
 ----
-$ minishift profile set PROFILE_NAME
+$ minishift profile set profile-demo
+Profile 'profile-demo' set as active profile
 ----
+
+NOTE: Only one profile can be active at any time.
 
 [[deleting-profiles]]
 == Deleting Profiles
@@ -107,31 +101,36 @@ $ minishift profile set PROFILE_NAME
 To delete a profile, run:
 
 ----
-$ minishift profile delete PROFILE_NAME
+$ minishift profile delete profile-demo
+You are deleting the active profile. It will remove the VM and all related artifacts. Do you want to continue [y/N]?: y
+Deleted:  /Users/john/.minishift/profiles/profile-demo
+Profile 'profile-demo' deleted successfully
+Switching to default profile 'minishift' as the active profile.
 ----
 
 NOTE: The default profile *minishift* cannot be deleted.
 
 [[example-workflows-profiles]]
-== Example Workflows
+== Example Workflow for Profile Configuration
 
-To create a new profile with specific configurations, run:
+You have two options to create a new profile and configure its xref:../using/basic-usage.adoc#persistent-configuration[persistent configuration].
+The first option is to implicitly create the new profile by making it the active profile using the `profile set` command.
+Once the profile is active you can run any `minishift config` command.
+Lastly, start the instance:
 
 ----
-$ minishift profile set foo
-$ minishift config set memory 4GB
-$ minishift config set cpus 2
+$ minishift profile set profile-demo
+$ minishift config set memory 8GB
+$ minishift config set cpus 4
 $ minishift addon enable anyuid
 $ minishift start
 ----
 
-The alternative is to execute a series of commands each specifying the targeted profile explicitly using the `--profile` flag.
-
-Example:
+The alternative is to execute a series of commands each specifying the targeted profile explicitly using the `--profile` flag:
 
 ----
-$ minishift --profile foo config set memory 4GB
-$ minishift --profile foo config set cpus 2
-$ minishift --profile foo addon enable anyuid
-$ minishift --profile foo minishift start
+$ minishift --profile profile-demo config set memory 8GB
+$ minishift --profile profile-demo config set cpus 4
+$ minishift --profile profile-demo addon enable anyuid
+$ minishift --profile profile-demo minishift start
 ----
--- a/docs/source/_topic_map.yml
+++ b/docs/source/_topic_map.yml
@@ -27,6 +27,8 @@ Topics:
         File: index
       - Name: Basic Usage
         File: basic-usage
+      - Name: Profiles
+        File: profiles
       - Name: Image Caching
         File: image-caching
       - Name: Add-ons
@@ -35,8 +37,6 @@ Topics:
         File: host-folders
       - Name: Minishift Docker Daemon
         File: docker-daemon
-      - Name: Profiles
-        File: profiles
       - Name: Experimental Features
         File: experimental-features
   - Name: Interacting with OpenShift
--- a/docs/source/using/index.adoc
+++ b/docs/source/using/index.adoc
@@ -4,13 +4,13 @@ include::variables.adoc[]
 :icons:
 
 This section describes how to use the {project} CLI.
-It covers the basic commands as well as advanced features like image caching, add-ons, host folders and profiles.
+It covers the basic commands as well as advanced features like profiles, image caching, add-ons and host folders.
 
 - xref:../using/basic-usage.adoc#[Basic Usage]
+- xref:../using/profiles.adoc#[Profiles]
 - xref:../using/image-caching.adoc#[Image Caching]
 - xref:../using/addons.adoc#[Add-ons]
 - xref:../using/host-folders.adoc#[Host Folders]
-- xref:../using/profiles.adoc#[Profiles]
 - xref:../using/docker-daemon.adoc#[{project} Docker Daemon]
 - xref:../using/experimental-features.adoc#[Experimental Features]
 


docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,23 @@ toc::[]
...
+
+Each profile is completely independent of other profiles and things like cache, add-ons and persistent config are specific to only one profile.
+This means if you have set some specific configuration parameters for a profile, these settings does not apply to other profiles. 
+
+Currently we do not have any global config for profiles.
+So if you want to have a common configuration for all your profiles then you need to apply the settings for each profile.
+However you can use xref:../using/managing-minishif.adoc#environment-variables[environment variables] as a workaround to use same configuration settings for all profiles.
+
+
+Profile feature introduced a global --profile flag and few commands i.e. `profile list`, `profile set`, `profile delete`.
```
anjannath: --profile should be inside backticks `--profile`

docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,23 @@ toc::[]
...
-Each {project} instance can be created with different configurations and users can switch between different instances.
+Each {project} instance can be created with different configurations e.g memory, CPU, disk-size , add-ons etc and users can switch between different instances.
+Profile feature would help users to test different configurations without deleting an existing {project} instance.
+
+Each profile is completely independent of other profiles and things like cache, add-ons and persistent config are specific to only one profile.
+This means if you have set some specific configuration parameters for a profile, these settings does not apply to other profiles. 
+
+Currently we do not have any global config for profiles.
+So if you want to have a common configuration for all your profiles then you need to apply the settings for each profile.
+However you can use xref:../using/managing-minishift.adoc#environment-variables[environment variables] as a workaround to use same configuration settings for all profiles.
```
gbraad: same or some?

docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,23 @@ toc::[]
...
+
+
+Profile feature introduced a global --profile flag and few xref:../command-ref/minishift_profile.adoc#[commands] for listing, deleting and setting an active profile.
```
gbraad: The

anjannath:
@LalatenduMohanty  A note saying the profile set command creats a profile if it didn't exist earlier can also be added i guess.

docs/source/using/profiles.adoc
LalatenduMohanty: @anjannath yes, will fix it.

docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,23 @@ toc::[]
...
+
+The --profile global flag is used with other {project} commands to run the command against a specific profile.
+
+NOTE: The default profile name is `minishift`. It will be created automatically and will present bydefault.
```
gbraad: by default

docs/source/using/profiles.adoc
```diff
@@ -64,3 +80,27 @@ $ minishift profile delete PROFILE_NAME
...
+Alternate method to create the profile 'foo'
+
+----
+$ minishift --profile foo config set memory 4GB
+$ minishift --profile foo config set cpus 2
+$ minishift --profile foo addon enable anyuid
+$ minishift --profile foo minishift start
+----
+
+NOTE: once you have started multiple profiles, you should use profile set command to switch between profiles.
```
gbraad: use `profile set` ?

docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,23 @@ toc::[]
 == Overview
 
 {project} allows to create multiple instances of {project}.
-Each {project} instance can be created with different configurations and users can switch between different instances.
+Each {project} instance can be created with different configurations e.g memory, CPU, disk-size , add-ons etc and users can switch between different instances.
+Profile feature would help users to test different configurations without deleting an existing {project} instance.
+
+Each profile is completely independent of other profiles and things like cache, add-ons and persistent config are specific to only one profile.
```
praveenkumar: "specific to only one profile" => "specific to that profile."

docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,23 @@ toc::[]
...
+This means if you have set some specific configuration parameters for a profile, these settings does not apply to other profiles. 
+
+Currently we do not have any global config for profiles.
+So if you want to have a common configuration for all your profiles then you need to apply the settings for each profile.
```
praveenkumar: "each profile" => "each profile separately"

docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,26 @@ toc::[]
 == Overview
 
 {project} allows to create multiple instances of {project}.
```
hferentschik: "Minishift allows to create multiple instances of Minishift." - that does not sound right. Maybe multiple instances of the Minshift VM? Also it should be "creating" not "to create".

docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,26 @@ toc::[]
...
-Each {project} instance can be created with different configurations and users can switch between different instances.
+Each {project} instance can be created with different configurations e.g memory, CPU, disk-size , add-ons etc and users can switch between different instances.
```
hferentschik: There is a logical error here. So far we have not introduced 'profile' at all. Here we talk still "Each Minishift instance". Then next sentence is "Profile feature would help ...". At some stage, we need to say something like" an instance of the Minishift VM together with all its configuration and state information is called a Minsihift profile". 

Also I am missing the clarification that for now we have a "share nothing" architecture between profiles.

docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,26 @@ toc::[]
...
+Profile feature would help users to test different configurations without deleting an existing {project} instance.
+
+Each profile is completely independent of other profiles and things like cache, add-ons and persistent config are specific to only one profile.
```
hferentschik: Ok, here we talk about being independent. I think this needs to come earlier. Also I suggest using the "share nothing" term as well.

docs/source/using/profiles.adoc
```diff
@@ -12,7 +12,26 @@ toc::[]
...
+This means if you have set some specific configuration parameters for a profile, these settings do not apply to other profiles.
+
+User can have mutiple profiles but only one profile will be active at one point of time i.e. by default {project} commands will run against the active profile.
```
hferentschik: "You can ..." - we try as much as possible to directly address the user. 
I am missing a bit of a red thread in this and the following sentences. IMO they are not connected so well

hferentschik: I think we are not consistent in the use of "instance" vs "profile". For example Create Profile section. Since we defined what a profile is and the section is called "creating profile", it should really be "you can create a new profile. ..."

docs/source/using/profiles.adoc
```diff
@@ -64,3 +83,27 @@ $ minishift profile delete PROFILE_NAME
 ----
 
 NOTE: The default profile 'minishift'  can not be deleted.
+
+[[example-workflows-to create-new-profiles]]
+== Example workflows to create a new profile
```
hferentschik: Capitalisation (applies to all headings)

hferentschik: I'd call this just "Example Workflow"

docs/source/using/profiles.adoc
```diff
@@ -64,3 +83,27 @@ $ minishift profile delete PROFILE_NAME
...
+Alternate method to create the profile 'foo'
+
+----
+$ minishift --profile foo config set memory 4GB
+$ minishift --profile foo config set cpus 2
+$ minishift --profile foo addon enable anyuid
+$ minishift --profile foo minishift start
+----
+
+NOTE: once you have started multiple profiles, you should use xref:set-active-profiles[`profile set`] command to switch between profiles.
```
hferentschik: captial O - Once

hferentschik: "you can" instead of "you should". Maybe make it a TIP as well

LalatenduMohanty: @hferentschik I intentionally used `you should` as I do not users to use `--profile` after the profile is created as we know it has some side effects.

LalatenduMohanty:
@jorgemoralespou You might be interested in this. Please take a look if you have time.

jorgemoralespou:
SGTM.

docs/source/using/profiles.adoc
```diff
@@ -11,8 +11,36 @@ toc::[]
...
+
+Each {project} profile can be created with different configurations e.g memory, CPU, disk-size , add-ons etc and users can switch between different profiles.
+Profile feature would help users to test different configurations without deleting an existing {project} instance.
+
+Each profile is completely independent of other profiles and they share nothing between them.
+Things like cache, add-ons and persistent config are specific to a single profile.
+This means if you have set some specific configuration parameters for a profile, these settings do not apply to other profiles.
+
+User can have mutiple profiles but only one profile will be active at one point of time i.e. by default {project} commands will run against the active profile.
+When a profile is started it becomes the active profile by-default. You can change the active profile using xref:switching-profiles[`profile set`] command.
```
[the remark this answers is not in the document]
LalatenduMohanty: > As a side note, is there a particular reason that we're using xref:switching-profiles as opposed to creating an xref to the command reference for minishift profile set?

Yes, as I want the users to read the section related to it. Command reference docs do not have any extra context for users.

docs/source/using/profiles.adoc
```diff
@@ -64,3 +92,27 @@ $ minishift profile delete PROFILE_NAME
 ----
 
 NOTE: The default profile 'minishift'  can not be deleted.
+
+[[example-workflows-profiles]]
+== Example Workflows
+
+To create a new profile foo with specific configurations, run:
```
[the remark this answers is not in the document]
LalatenduMohanty: @robin-owen As per The OpenShift Documentation Guidelines it should be *minishift* as it is an example name.

docs/source/using/profiles.adoc
```diff
@@ -11,8 +11,36 @@ toc::[]
...
+----
+$ export MINISHIFT_MEMORY=8G
+----
+
+Profile feature introduced a global `--profile` flag and few xref:../command-ref/minishift_profile.adoc#[commands] for listing, deleting and setting an active profile.
+
+The `--profile` global flag is used with other {project} commands to run a command against a specific profile.
+
+NOTE: The default profile name is `minishift`. It will be created automatically and will be present by default.
+If an user has not created any profiles explictly and not using --profile flag then the commands will run against the default profile.
```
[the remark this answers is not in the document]
LalatenduMohanty: Interesting , I did not know this. https://www.quora.com/Is-it-an-user-or-a-user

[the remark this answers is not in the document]
LalatenduMohanty: > "Commands are run in the default profile if no profiles have been created and the --profile flag is not given."

This changes the meaning and does not sound correct. Commands are run against a profile and not in the profile.

docs/source/using/profiles.adoc
```diff
@@ -52,7 +80,7 @@ To switch between profiles or to set a profile as active, run:
 $ minishift profile set PROFILE_NAME
 ----
 
-When you have multiple profiles you can use `--profile` flag with any minishift command to run the command against a specific profile.
+NOTE: You can run `profile set` command for a non existing profile and it will make the profile as the active profile.
```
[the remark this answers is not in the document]
LalatenduMohanty: > "NOTE: Given a non-existing profile, the profile set command creates the given profile and sets it as the active profile."

Actually it does not create the profile. Will fix it.

LalatenduMohanty:
@robin-owen Please take another look as I have made changes as per your comments.

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
-[[pofile-overview]]
+[[profile-overview]]
 == Overview
 
-{project} allows to create multiple instances of {project}.
-Each {project} instance can be created with different configurations and users can switch between different instances.
+The profile feature allows creating multiple instances of {project}. 
 
-[[creating-profile]]
+A profile is an instance of the {project} VM along with all of its configuration and state information.
```
hferentschik: "state information" == "state" (state information sounds odd).

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
+
+Each {project} profile can be created with its own configuration (memory, CPU, disk size, add-ons, and so on) and completely independent of other profiles.
```
hferentschik: "can be created" -> "is created"

hferentschik: "and *is* completely independent of other profiles"

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
+Profiles share nothing between them. Things like cache, add-ons and persistent config are specific to a single profile.
```
hferentschik: One sentence per line. In this case we might want to drop "Profiles share nothing between them". This is implied by "completely independent"

hferentschik: "Things like cache, add-ons and persistent config are specific to a single profile." -> "In particular cached data, add-ons and persistent configuration are scoped per profile."

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
+This means if you have set some specific configuration parameters for a profile, these settings do not apply to other profiles.
```
hferentschik: I think this sentence is not not needed.

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
+
+Currently, we cannot set any global configuration for profiles.
```
hferentschik: Here I am losing the red thread again. I think this whole discussion should be moved into a dedicated TIP or section.

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
+If you want to apply a common configuration to all your profiles then you need to apply the settings for each profile separately.
+As a workaround, xref:../using/basic-usage.adoc#environment-variables[environment variables] can be used.
+
+To set 8GB memory for all profiles on Linux/macOS, run:
+
+----
+$ export MINISHIFT_MEMORY=8G
+----
+
+You can have multiple profiles and can switch between them. However only one profile can be active at any time.
```
hferentschik: "You can have multiple profiles" - that's the whole point, right? No need to repeat this. You can just say:
"Only one profile can be active at a given time and you can use the `profile set` command to switch between them"

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
+As a workaround, xref:../using/basic-usage.adoc#environment-variables[environment variables] can be used.
+
+To set 8GB memory for all profiles on Linux/macOS, run:
+
+----
+$ export MINISHIFT_MEMORY=8G
+----
+
+You can have multiple profiles and can switch between them. However only one profile can be active at any time. 
+you do not need to delete an existing {project} instance to test a different configuration.
```
hferentschik: why is this relevant?

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
+
+By default {project} commands will run against the active profile.
```
hferentschik: This is kind of repetitive to the sentence above. I guess this can somehow be connected in a better way.

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
+When a profile is started it becomes the active profile. You can change the active profile using xref:switching-profiles[`profile set`] command.
```
hferentschik: I think some of this can go into its own sections which will also create sub headings for us, making it easier to find things. For an "overview" this goes now into too much detail imo.

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
+The profile feature introduced a global `--profile` flag and xref:../command-ref/minishift_profile.adoc#[commands] for listing, deleting and setting the active profile.
+
+The `--profile` global flag is used with other {project} commands to run a command against a specific profile.
+
+NOTE: Commands are run against the default profile If you are not using `--profile` flag  and not set a different active profile.
+
+[[creating-profiles]]
 == Creating Profiles
 
 You can create a new instance of {project} with xref:../command-ref/minishift_start.adoc#[`minishift start --profile`] command.
```
hferentschik: --profile is a global falg, and we usually print it after `minishift`. Let's be consistent in its use.

docs/source/using/profiles.adoc
robin-owen: This information is important enough that I think a section is appropriate.

docs/source/using/profiles.adoc
```diff
@@ -8,18 +8,45 @@ include::variables.adoc[]
...
 
-NOTE: The default profile is `minishift`. It will be present by-default and user does not need to create it specifically.
+NOTE: The default profile is *minishift*. It will be present by-default and user does not need to create it specifically.
```
robin-owen: Remove the hyphen in "by-default", please. There's also no need to address a user (directly or indirectly):

"This profile will be present by default and does not need to be explicitly created."

Note: I used "explicitly" rather than "specifically" here because the creation of the *minishift* profile is implicit.

docs/source/using/profiles.adoc
hferentschik: Right, here is probably a good point to leave the overview. The overview should define the feature and give some context. That we have done now. The rest should be in dedicated sections, preferably in some logical order.

docs/source/using/profiles.adoc
LalatenduMohanty: I think I can remove this.

LalatenduMohanty:
@robin-owen @hferentschik Fixed some of the things you pointed out. Take a look again.

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,47 @@ include::variables.adoc[]
...
+
+The current profile being used is called the active profile.
+By default {project} commands will run against the active profile.
+You can run the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command to findout the active profile.
+
+The profile feature introduced a global `--profile` flag and xref:../command-ref/minishift_profile.adoc#[commands] for listing, deleting and setting the active profile.
+
+The `--profile` global flag is used with other {project} commands to run a command against a specific profile.
+
+NOTE: Commands are run against the default profile if you are not using `--profile` flag and not set a different active profile.
```
robin-owen: "the `--profile` flag" and "have not set":

"NOTE: Commands are run against the default profile if you are not using the `--profile` flag and have not set a different active profile."

docs/source/using/profiles.adoc
```diff
@@ -43,16 +61,17 @@ $ minishift profile list
 [[switching-profiles]]
 == Switching Profiles
 
-With the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] you can switch between profiles or set a profile as active.
-When a profile is active, it is not required to pass `--profile` flag to run minishift commands against that profile.
+With the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] you can switch between profiles.
+However only one profile can be active at any time.
```
robin-owen: "However" seems unnecessary to me. I'd consider removing it.

docs/source/using/profiles.adoc
```diff
@@ -63,4 +82,41 @@ To delete a profile, run:
...
+Alternate method to create the profile *foo*:
+
+----
+$ minishift --profile foo config set memory 4GB
+$ minishift --profile foo config set cpus 2
+$ minishift --profile foo addon enable anyuid
+$ minishift --profile foo minishift start
+----
+
+Note: Once you have started multiple profiles, you should use the xref:switching-profiles[`profile set`] command to switch between them.
```
robin-owen: I'd consider using `IMPORTANT` as the admonition here.

docs/source/using/profiles.adoc
```diff
@@ -43,16 +61,17 @@ $ minishift profile list
 [[switching-profiles]]
 == Switching Profiles
 
-With the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] you can switch between profiles or set a profile as active.
-When a profile is active, it is not required to pass `--profile` flag to run minishift commands against that profile.
+With the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] you can switch between profiles.
```
robin-owen: I would rewrite this sentence so that we immediately tell the user _what_ they can do followed by _how_ they can do it:

"You can switch between profiles with the xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] command."

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,47 @@ include::variables.adoc[]
...
+A profile is an instance of the {project} VM along with all of its configuration and state.
+
+Each {project} profile can be created with its own configuration (memory, CPU, disk size, add-ons, and so on) and completely independent of other profiles.
+Profiles share nothing between them.
+Things like cache, add-ons and persistent config are specific to a single profile.
+This means if you have set some specific configuration parameters for a profile, these settings do not apply to other profiles.
+
+The current profile being used is called the active profile.
+By default {project} commands will run against the active profile.
+You can run the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command to findout the active profile.
```
robin-owen: Please change "findout" to "find" or "determine".

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,47 @@ include::variables.adoc[]
...
-Each {project} instance can be created with different configurations and users can switch between different instances.
+The profile feature allows creating multiple instances of {project}. 
 
-[[creating-profile]]
+A profile is an instance of the {project} VM along with all of its configuration and state.
+
+Each {project} profile can be created with its own configuration (memory, CPU, disk size, add-ons, and so on) and completely independent of other profiles.
+Profiles share nothing between them.
+Things like cache, add-ons and persistent config are specific to a single profile.
+This means if you have set some specific configuration parameters for a profile, these settings do not apply to other profiles.
```
robin-owen: These four sentences feel like they're saying essentially the same thing in slightly different (more nuanced) ways. At some point, I'd like to rewrite this so that we convey the same information without feeling like we're repeating ourselves.

docs/source/using/profiles.adoc
```diff
@@ -43,16 +61,17 @@ $ minishift profile list
 [[switching-profiles]]
 == Switching Profiles
 
-With the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] you can switch between profiles or set a profile as active.
-When a profile is active, it is not required to pass `--profile` flag to run minishift commands against that profile.
+With the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] you can switch between profiles.
+However only one profile can be active at any time.
+When the profile is active, you do not need to use `--profile` flag to run `minishift` commands against the active profile.
```
robin-owen: "When the profile is active" is redundant -- the sentence reads well without it: "You do not need to use the `--profile` flag to run `minishift` commands against the active profile."

docs/source/using/profiles.adoc
hferentschik: I think now is the time. Let's do this right to start with.

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,47 @@ include::variables.adoc[]
...
+Things like cache, add-ons and persistent config are specific to a single profile.
+This means if you have set some specific configuration parameters for a profile, these settings do not apply to other profiles.
+
+The current profile being used is called the active profile.
+By default {project} commands will run against the active profile.
+You can run the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command to findout the active profile.
+
+The profile feature introduced a global `--profile` flag and xref:../command-ref/minishift_profile.adoc#[commands] for listing, deleting and setting the active profile.
+
+The `--profile` global flag is used with other {project} commands to run a command against a specific profile.
```
hferentschik: The `--profile` global flag can be specified with any {project} command to run it against the specific profile bypassing the currently active profile.

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,47 @@ include::variables.adoc[]
...
+Each {project} profile can be created with its own configuration (memory, CPU, disk size, add-ons, and so on) and completely independent of other profiles.
+Profiles share nothing between them.
+Things like cache, add-ons and persistent config are specific to a single profile.
+This means if you have set some specific configuration parameters for a profile, these settings do not apply to other profiles.
+
+The current profile being used is called the active profile.
+By default {project} commands will run against the active profile.
+You can run the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command to findout the active profile.
+
+The profile feature introduced a global `--profile` flag and xref:../command-ref/minishift_profile.adoc#[commands] for listing, deleting and setting the active profile.
```
hferentschik: I talk first about the commands and then introduce --global flag.

docs/source/using/profiles.adoc
hferentschik: why "should". What alternative do I have to switch?

docs/source/using/profiles.adoc
hferentschik: I think it could be as simple as:

"Each {project} profile is created with its own configuration (memory, CPU, disk size, add-ons, and so on) and is independent of other profiles. If you have common configuration options you want to apply to all profiles, you have to do this manually per profile. See xref:../using/profiles.adoc#global-config-profiles[Global Configuration Options]"

Effectively we skip the three repeating sentences.

docs/source/using/profiles.adoc
```diff
@@ -63,4 +82,41 @@ To delete a profile, run:
...
+$ minishift --profile foo minishift start
+----
+
+Note: Once you have started multiple profiles, you should use the xref:switching-profiles[`profile set`] command to switch between them.
+
+[[global-config-profiles]]
+== Global Configuartion For profiles
+
+Currently, we cannot set any global configuration for profiles.
+If you want to apply a common configuration to all your profiles then you need to apply the settings for each profile separately.
```
hferentschik: An example would be nice here.

docs/source/using/profiles.adoc
```diff
@@ -63,4 +82,41 @@ To delete a profile, run:
...
+As a workaround, xref:../using/basic-usage.adoc#environment-variables[environment variables] can be used.
```
hferentschik: "In some cases  xref:../using/basic-usage.adoc#environment-variables[environment variables] can be used."

It won't work for all settings. Needs to be elaborated a bit more.

docs/source/using/profiles.adoc
```diff
@@ -63,4 +82,41 @@ To delete a profile, run:
...
+
+To set 8GB memory for all profiles on Linux/macOS, run:
```
hferentschik: "To allocate 8GB of memory for all profiles use the MINISHIFT_MEMOR environment variable. How you set the environment variable is OS specific. For Linux or macOS run:"

Alternatively give both options Linux/macOS and Windows.

Currently it sounds a bit like it only works for Linux/macOS.

docs/source/using/profiles.adoc
robin-owen: That works for me as long as the same essential information is being conveyed to the reader.

docs/source/using/profiles.adoc
hferentschik: I am wondering whether we need the --profile flag as part of the overview or whether this is also not a good candidate for a subsection "Executing commands against non active profile".

LalatenduMohanty: > I talk first about the commands and then introduce --global flag.

IMO we should be talking about global flag first as it can be used with any other command. 

> I am wondering whether we need the --profile flag as part of the overview or whether this is also not a good candidate for a subsection "Executing commands against non active profile".

It would like to leave it in overview. Not sure if we need a dedicated section.

docs/source/using/profiles.adoc
LalatenduMohanty: I do not think I should give an example here. The right thing to do is to point the users to the right documentation for environment variables.

robin-owen: @LalatenduMohanty Agreed, the example seems unnecessary. We should simply guide the reader to the appropriate information.

@hferentschik @LalatenduMohanty Considering the note about elaboration made, I think it makes sense to elaborate on the use of environment variables within their own documentation. In my opinion, we shouldn't need to go into any _detail_ regarding environment variables in the documentation for profiles.

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,45 @@ include::variables.adoc[]
...
+NOTE: Commands are run against the default profile if you are not using the `--profile` flag and have not set a different active profile.
+
+[[creating-profiles]]
 == Creating Profiles
 
-You can create a new instance of {project} with xref:../command-ref/minishift_start.adoc#[`minishift start --profile`] command.
+You can create a new instance of {project} with xref:../command-ref/minishift_start.adoc#[`minishift --profile PROFILE_NAME start`] command.
 
-NOTE: The default profile is `minishift`. It will be present by-default and user does not need to create it specifically.
+NOTE: The default profile is *minishift*. It will be present by default and user does not need to create it specifically.
```
robin-owen: No need to address the user, whether directly or indirectly: We also want one sentence per line. In the case of admonitions, I find that it's easier to quickly see and edit them when they're put in explicit blocks:

```adoc
[NOTE]
====
The default profile is *minishift*.
It will be present by default and does not need to be created.
====
```

docs/source/using/profiles.adoc
```diff
@@ -43,16 +59,17 @@ $ minishift profile list
 [[switching-profiles]]
 == Switching Profiles
 
-With the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] you can switch between profiles or set a profile as active.
-When a profile is active, it is not required to pass `--profile` flag to run minishift commands against that profile.
+You can switch between profiles with the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] command.
```
robin-owen: I'd remove "help of": "You can switch between profiles with the `minishift profile set` command."

docs/source/using/profiles.adoc
```diff
@@ -43,16 +59,17 @@ $ minishift profile list
...
+At any time only one profile can be the active profile.
```
robin-owen: This is technically correct, but awkward to read. Suggested rewrite: "Only one profile can be active at any time."

If it's more technically accurate to specifically reference "the active profile" versus "one profile can be active", we can simply move "at any time" to the end of the sentence rather than the beginning.

docs/source/using/profiles.adoc
```diff
@@ -43,16 +59,17 @@ $ minishift profile list
...
+You do not need to use `--profile` flag to run {project} commands against the active profile.
```
robin-owen: "the `--profile` flag"

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,45 @@ include::variables.adoc[]
...
 
-{project} allows to create multiple instances of {project}.
-Each {project} instance can be created with different configurations and users can switch between different instances.
+The profile feature allows creating multiple instances of {project}. 
 
-[[creating-profile]]
+A profile is an instance of the {project} VM along with all of its configuration and state.
+
+Each {project} profile can be created with its own configuration (memory, CPU, disk size, add-ons, and so on) and completely independent of other profiles.
+If you have common configuration options you want to apply to all profiles, you have to do this manually per profile. See xref:../using/profiles.adoc#setting-global-config-profiles[Setting Global Configuration Options]
```
robin-owen: One sentence per line, please. Also, a period is needed after the `xref` to Setting Global Configuration Options.

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,50 @@ include::variables.adoc[]
 
 toc::[]
 
-[[pofile-overview]]
+[[profile-overview]]
 == Overview
 
-{project} allows to create multiple instances of {project}.
-Each {project} instance can be created with different configurations and users can switch between different instances.
+The profile feature allows creating multiple instances of {project}.
```
hferentschik: Even so syntactically correct, it does not "flow" so well imo. I also don't like that we have to stand-alone sentences to start the overview. I suggest to start more like this:

"A profile is an instance of the Minishift VM along with all of its configuration and state. The profile feature allows you to create and manage these isolated instances of Minishift."

@robin-owen, @LalatenduMohanty - WDYT?

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,50 @@ include::variables.adoc[]
...
 == Overview
 
-{project} allows to create multiple instances of {project}.
-Each {project} instance can be created with different configurations and users can switch between different instances.
+The profile feature allows creating multiple instances of {project}. 
 
-[[creating-profile]]
+A profile is an instance of the {project} VM along with all of its configuration and state.
+
+Each {project} profile can be created with its own configuration (memory, CPU, disk size, add-ons, and so on) and completely independent of other profiles.
```
hferentschik: "Each {project} profile *is* created with its own configuration" - it is not really a "can" here. It is its own configuration.

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,50 @@ include::variables.adoc[]
...
+If you have common configuration options you want to apply to all profiles, you have to do this manually per profile.
+See xref:../using/profiles.adoc#setting-global-config-profiles[Setting Global Configuration Options].
```
hferentschik: The titles don't match

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,50 @@ include::variables.adoc[]
...
+
+The current profile being used is called the active profile.
```
hferentschik: This needs a bit more of lead in imo. Also I would try to word things along these lines:

"Unless the global  `--profile` flag is used, Minishift commands are executed against the so-called active profile. You can determine the active profile by using `minishift profile list`.  You can execute single commands against a non-active profile, by using the `--profile` flag, eg `minishift --profile my-profile console --url`.

On top of the  `--profile` flag, there are commands for setting , listing and deleting  the active profile. These commands are described in the following sections."

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,50 @@ include::variables.adoc[]
...
+
+The `--profile` global flag can be specified with any {project} command to run it against the specific profile, bypassing the currently active profile.
+
+NOTE: Commands are run against the default profile if you are not using the `--profile` flag and have not set a different active profile.
+
+[[creating-profiles]]
 == Creating Profiles
 
-You can create a new instance of {project} with xref:../command-ref/minishift_start.adoc#[`minishift start --profile`] command.
+You can create a new instance of {project} with the xref:../command-ref/minishift_start.adoc#[`minishift --profile PROFILE_NAME start`] command.
```
hferentschik: Again, there is a disconnect here. We are talking about profiles and how to create them, but then the section starts with "you can create a new instance of {project}". I think we need to motiviate this better.

"There are two ways to create a new profile, explicitly or implicitly. 

=== Explicit Profile Creation 

bla bla bla

    $ minishift profile set my-profile

=== Implicit Profile creation

bla bla bla

    $ minishift --profile my-profile start

docs/source/using/profiles.adoc
```diff
@@ -8,29 +8,50 @@ include::variables.adoc[]
...
 ----
 
+NOTE: When a profile is started it becomes the active profile.
+
 [[listing-profiles]]
 == Listing Profiles
 
-You can list all existing profiles with xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command.
+You can list all existing profiles with the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command.
 You can also see the active profile in the output.
```
hferentschik: The active profile is highlighted.

docs/source/using/profiles.adoc
```diff
@@ -43,16 +64,17 @@ $ minishift profile list
 [[switching-profiles]]
 == Switching Profiles
 
-With the help of xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] you can switch between profiles or set a profile as active.
-When a profile is active, it is not required to pass `--profile` flag to run minishift commands against that profile.
+You can switch between profiles with the xref:../command-ref/minishift_profile_set.adoc#[`minishift profile set`] command.
+Only one profile can be active at any time.
+You do not need to use the `--profile` flag to run {project} commands against the active profile.
```
hferentschik: This sentence is not necessary here and distracts from what we are trying to say. Let's just remove it.

docs/source/using/profiles.adoc
```diff
@@ -43,16 +64,17 @@ $ minishift profile list
...
 
 To switch between profiles or to set a profile as active, run:
```
hferentschik: "or"? Is switching profiles and setting the active profile not the same thing? Jsut "To switch between profiles"

docs/source/using/profiles.adoc
```diff
@@ -63,4 +85,33 @@ To delete a profile, run:
 $ minishift profile delete PROFILE_NAME
 ----
 
-NOTE: The default profile 'minishift'  can not be deleted.
+NOTE: The default profile *minishift* cannot be deleted.
+
+[[example-workflows-profiles]]
+== Example Workflows
```
hferentschik: I'd move this section last.

docs/source/using/profiles.adoc
```diff
@@ -63,4 +85,33 @@ To delete a profile, run:
...
+
+To create a new profile *foo* with specific configurations, run:
```
hferentschik: @LalatenduMohanty you want to keep "foo"?

hferentschik: We need to explain what's happening here. Here we use the the set command to switch to the new profile, making it the active one. Then we execute a series of commands against the active profile.

docs/source/using/profiles.adoc
```diff
@@ -63,4 +85,33 @@ To delete a profile, run:
...
+
+----
+$ minishift profile set foo
+$ minishift config set memory 4GB
+$ minishift config set cpus 2
+$ minishift addon enable anyuid
+$ minishift start
+----
+
+Alternate method to create the profile *foo*:
```
hferentschik: The alternative is to execute a series of commands each specifying the targeted profile explicitly using the --profile flag

docs/source/using/profiles.adoc
```diff
@@ -63,4 +85,33 @@ To delete a profile, run:
...
+$ minishift --profile foo config set memory 4GB
+$ minishift --profile foo config set cpus 2
+$ minishift --profile foo addon enable anyuid
+$ minishift --profile foo minishift start
+----
+
+[[setting-global-config-profiles]]
+== Setting Global Configuration for Profiles
+
+Currently, we cannot set any global configuration for profiles.
```
hferentschik: we == you

docs/source/using/profiles.adoc
```diff
@@ -63,4 +85,33 @@ To delete a profile, run:
...
+If you want to apply a common configuration to all your profiles then you need to apply the settings for each profile separately.
```
hferentschik: reference to example-workflows-profiles which should follow after this section

docs/source/using/profiles.adoc
```diff
@@ -63,4 +85,33 @@ To delete a profile, run:
...
+In some cases xref:../using/basic-usage.adoc#environment-variables[environment variables] can be used.
```
hferentschik: +1

docs/source/using/profiles.adoc
LalatenduMohanty: @hferentschik `minishift profile set my-profile` it does not create any profile but sets it as active profile. I think there a difference between creating a profile(where artifacts gets created) and setting active profile.

LalatenduMohanty: I think you have mistakenly reversed the use of `Explicit` and `Implicit` above.

LalatenduMohanty: I guess we need to separate the section between creating a profile (implicitly) and starting of a profile. Let see what I came back with.

docs/source/using/profiles.adoc
```diff
@@ -63,4 +111,36 @@ To delete a profile, run:
...
 
-NOTE: The default profile 'minishift'  can not be deleted.
+NOTE: The default profile *minishift* cannot be deleted.
+
+[[setting-global-config-profiles]]
+== Setting Global Configuration for Profiles
+
+Currently, you cannot set any global configuration for profiles.
+If you want to apply a common configuration to all your profiles then you need to apply the settings for each profile separately.
+See xref:../using/profiles.adoc#setting-global-config-profiles[Setting Global Configuration For Profiles].
```
hferentschik: You realise you are referring to yourself here. 

In the intro you say: "If you have common configuration options you want to apply to all profiles, you have to do this manually per profile See Setting Global Configuration For Profiles."

One here we offer no solution and even reference to the very section we are reading.

hferentschik: I think we need to be more explicit and give an example for the common env use case. For example the CPU count and memory size is probably something people want to have for all profiles. How can this be done.

docs/source/using/profiles.adoc
```diff
@@ -8,52 +8,100 @@ include::variables.adoc[]
...
 
-[[creating-profile]]
+Each {project} profile is created with its own configuration (memory, CPU, disk size, add-ons, and so on) and completely independent of other profiles.
+If you have common configuration options you want to apply to all profiles, you have to do this manually per profile.
+See xref:../using/profiles.adoc#setting-global-config-profiles[Setting Global Configuration For Profiles].
+
+The current profile being used is called the active profile.
+Unless the global `--profile` flag is used, {project} commands are executed against the active profile.
+You can determine the active profile by using the xref:../command-ref/minishift_profile_list.adoc#[`minishift profile list`] command. 
+You can execute single commands against a non-active profile, by using the `--profile` flag.
```
hferentschik: I don't think this fits into the overview

docs/source/using/profiles.adoc
```diff
@@ -8,52 +8,100 @@ include::variables.adoc[]
...
+----
+
+On top of the `--profile` flag, there are xref:../command-ref/minishift_profile.adoc#[commands] for listing, deleting and setting the active profile.
+These commands are described in the following sections.
+
+[[creating-profiles]]
 == Creating Profiles
 
-You can create a new instance of {project} with xref:../command-ref/minishift_start.adoc#[`minishift start --profile`] command.
+A profile is created when any artifacts (example: configuration files) asociated with it gets created.
```
hferentschik: "when any artifacts (example: configuration files) asociated with it gets created"? What is that supposed to mean? This is not something the user does.

docs/source/using/profiles.adoc
```diff
@@ -8,52 +8,100 @@ include::variables.adoc[]
...
+There are two ways to create a new profile.
+
+==== Using the `--profile` flag
+
+When you run any {project} command with the `--profile` flag the profile gets created if it does not exist.
+
+Example:
+----
+$ minishift --profile PROFILE_NAME cofig set PROPERTY_NAME PROPERTY_VALUE
```
hferentschik: there is a typo "cofig". Should we not use the start command here?

docs/source/using/profiles.adoc
```diff
@@ -8,52 +8,100 @@ include::variables.adoc[]
...
+----
+
+=== Using `profile set` command.
 
-NOTE: The default profile is `minishift`. It will be present by-default and user does not need to create it specifically.
+You can run the `profile set` command for a non existing profile to set it as the active profile.
+The atrifacts for the profile are automatically created when you subsequently run any `minishift` command.
```
hferentschik: That's not really relevant for the user. Artifacts is ambiguous in this case. Also there is a typo 'atrifacts'

docs/source/using/profiles.adoc
```diff
@@ -8,52 +8,100 @@ include::variables.adoc[]
...
+====
+The default profile is *minishift*.
+It will be present by default and it does not need to be created.
+====
+
+[[starting-a-profile]]
+== Starting A Profile
+
+You can start the active profile with xref:../command-ref/minishift_start.adoc#[`minishift start`] command.
+If the profile is a non active profile, you need to use the `--profile` flag.
```
hferentschik: "To start a non active profile, you need to use the `--profile` flag."

docs/source/using/profiles.adoc
LalatenduMohanty: I suggest we leave it here. It might be useful to users (who are not that technical or knows minishift) to get the overview

hferentschik:
@LalatenduMohanty @robin-owen I've created another pull request against @LalatenduMohanty's changes here - https://github.com/LalatenduMohanty/minishift/pull/1

Easier than going a bit in circles. I tried to expand on the examples and add some actual console output as well. I also expanded on the example workflow describing what it actually does. I also tried to clarify the 'active' profile and make the whole docs a bit more compact.

PR #1618, opened 2017-10-25
Issue #1258 Restructure ISO cache path

### LINKED ISSUE #1258: ISO caching need to be restructure to have version information
Currently we don't have any version directory structure for ISO cache and moving forward if we want to decouple our ISO releases with minishift then we need to put some plan for it. 

- Should have `/cache/iso/<iso_verison>/minishift.iso` structure.
- Have a default iso version flag/variable on minishift side so that we understand what should be be downloaded and compatible.

@gbraad please add your points.

praveenkumar:
Fix #1258

### DIFF
--- a/pkg/minishift/util/iso_path.go
+++ b/pkg/minishift/util/iso_path.go
@@ -1,0 +1,49 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
+	"net/url"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+func GetIsoPath(isoURL string) string {
+	// Make sure isoURL validation happen before using this function.
+	// This function ignore any error from url.Parse
+	uri, _ := url.Parse(isoURL)
+
+	switch true {
+	case strings.Contains(filepath.Base(isoURL), minishiftConstants.B2dIsoAlias):
+		return filepath.Join(minishiftConstants.B2dIsoAlias, getIsoVersion(uri.Path))
+	case strings.Contains(filepath.Base(isoURL), minishiftConstants.CentOsIsoAlias):
+		return filepath.Join(minishiftConstants.CentOsIsoAlias, getIsoVersion(uri.Path))
+	case strings.Contains(filepath.Base(isoURL), minishiftConstants.MinikubeIsoAlias):
+		return filepath.Join(minishiftConstants.MinikubeIsoAlias, getIsoVersion(uri.Path))
+	default:
+		// This handle any random URI
+		return filepath.Join("unnamed")
+	}
+
+}
+
+func getIsoVersion(isoName string) string {
+	re := regexp.MustCompile("v[0-9]*.[0-9]*.[0-9]*")
+	return re.FindString(isoName)
+}
--- a/pkg/minishift/util/iso_path_test.go
+++ b/pkg/minishift/util/iso_path_test.go
@@ -1,0 +1,66 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+func TestGetIsoPath(t *testing.T) {
+	testData := []struct {
+		provided string
+		expected string
+	}{
+		{"https://storage.googleapis.com/minikube/iso/minikube-v0.24.0.iso", filepath.Join("minikube", "v0.24.0")},
+		{"https://storage.googleapis.com/minikube/iso/minikube-v0.22.11.iso", filepath.Join("minikube", "v0.22.11")},
+		{"https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso", filepath.Join("b2d", "v1.1.0")},
+		{"https://github.com/minishift/minishift-b2d-iso/releases/download/v1.2.0/minishift-b2d.iso", filepath.Join("b2d", "v1.2.0")},
+		{"https://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso", filepath.Join("centos", "v1.1.0")},
+		{"https://github.com/minishift/minishift-centos-iso/releases/download/v1.3.0/minishift-centos7.iso", filepath.Join("centos", "v1.3.0")},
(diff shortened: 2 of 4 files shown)


pkg/minishift/util/iso_path.go
```diff
@@ -0,0 +1,55 @@
...
+	"regexp"
+	"strings"
+)
+
+func GetIsoPath(isoURL string) string {
+	uri, _ := url.Parse(isoURL)
+	uriPathList := strings.Split(uri.Path, "/")
+
+	// This will only work for minishift-centos and b2d iso which is part of minishift repo
+	if len(uriPathList) == 7 {
```
budhrg: This logic is little weird to me to really on it.

hferentschik: String splitting is really fragile and obfuscating in this case? What's wrong with regexp for example? BTW, there might be other non-canonical URLs (as part of the CDN) which might be valid as well. We might want to deal with these as well.

praveenkumar: @hferentschik Can you provide an non-canonical url example so I can even add it in the unit test also?

pkg/minishift/util/iso_path.go
```diff
@@ -0,0 +1,47 @@
...
+		return filepath.Join(minishiftConstants.MinikubeIsoAlias, getIsoVersion(uri.Path))
+	default:
+		// This handle any random URI
+		return filepath.Join("unnamed")
+	}
+
+}
+
+func getIsoVersion(isoName string) string {
+	re := regexp.MustCompile("v[0-9]*.[0-9]*.[0-9]*")
```
budhrg: This looks more promising logic than previous :+1:

praveenkumar:
retest this please

hferentschik:
@praveenkumar Is this ready for a new review?

hferentschik:
I tried the following:

    $ minishift start --iso-url b2d
    $ minishift start --iso-url minikube

I ended up with

     $ tree ~/.minishift/cache/iso/
     /Users/hardy/.minishift/cache/iso/
     ├── b2d
     │   └── v1.2.0
     │       └── minishift-b2d.iso
     ├── centos
     │   └── v1.3.0
     │       └── minishift-centos7.iso.part
     └── minikube
         └── v0.24.0
             └── minikube-v0.24.0.iso.part

That's looking good. I also verified that aborting a download will restart the download on the next start (hence the .part files).

hferentschik:
Here is one I don't get though:

    $ minishift start --iso-url https://github.com/minishift/minishift-centos-iso/releases/download/v1.2.0/minishift-centos7.iso
    -- Checking the ISO URL ... FAIL
     See the 'Basic Usage' topic (https://docs.openshift.org/latest/minishift/using/basic-usage.html) for more information

What's going on here?

hferentschik:
There is a bug in _checkIsoUrl_ in _start_preflight.go_:

```go
	fmt.Printf("\n   Checking if %s exists ... ", strings.TrimPrefix(isoUrl, "file://"))
	_, err := os.Stat(strings.TrimPrefix(isoUrl, "file://"))
	if err != nil {
		return false
	}
```

At this stage we could be still dealing with an URL and if so, the os.Stat check will fail. Effecitvely, we cannot point to URLs at all anymore. 

Can we fix this as part of this issue? @praveenkumar @anjannath

hferentschik:
Actually the problem starts already here:

```go
		match, _ := regexp.MatchString("^file://[a-zA-Z]:/.+", isoUrl)
		if !match {
			return false
		}
	} else {
		match, _ := regexp.MatchString("^file:///.+", isoUrl)
		if !match {
			return false
		}
	}
```

BTW, why does _checkIsoUrl_ not take the URL as a parameter, so that we could easily test the various cases. This is a simple string matching function. We can cover all scenarios in a simple unit test.

hferentschik:
So there seems to be a pull request for this already - https://github.com/minishift/minishift/pull/1652

hferentschik:
what's about trying to access the URL before creating parts of the path. For example:

    $ minishift start --iso-url https://github.com/minishift/minishift-centos-iso/releases/download/v1.4.0/minishift-centos7.iso

The ISO does not exist and I get

     Error caching the ISO: Received 404 response from https://github.com/minishift/minishift-centos-iso/releases/download/v1.4.0/minishift-centos7.iso while trying to download the ISO

However, on the file system there is already a directory created. Not a blocker, but maybe nice to have.

pkg/minishift/util/iso_path.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
+	"net/url"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+func GetIsoPath(isoURL string) string {
```
hferentschik: GetIsoTypeAndVersion!? Maybe even returning two strings?

LalatenduMohanty: @hferentschik Not sure what advantage we will have if we return two strings separately as the function is not really putting any logic for extracting them.

gbraad: GetIsoCachingPath or something, but it does not have to return version as can be seen with any image as `unnamed`.

hferentschik:
@praveenkumar can you rebase this pull request now?

praveenkumar:
> can you rebase this pull request now?

@hferentschik done.

pkg/minishift/util/iso_path.go
```diff
@@ -0,0 +1,47 @@
...
+	uri, _ := url.Parse(isoURL)
```
hferentschik: So, you ignore an error, because you know in this context that the URL is valid? So this method depends on the caller making sure that the URL is valid first?

hferentschik: If I'd call this method with an invalid URL, I would get a _nil_ uri and a potential panic?

praveenkumar: @hferentschik this method is called from the cluster package and the ISO URI comes from start.go (where we already check for URI validity). I purposefully ignore error here because of there already we check the validity of URI.

hferentschik: maybe we should add a comment. Not that this method ever gets used from another context.  After all, it is in the util package. 

If it is so specific to this particular usecase maybe it should be a package private method there, instead of a public util method.

The alternative is to check the error and either return it or treat it is a default 'unamed' case.

praveenkumar: > maybe we should add a comment. Not that this method ever gets used from another context. After all, it is in the util package.

@hferentschik done.

> If it is so specific to this particular use-case maybe it should be a package private method there, instead of a public util method.

Right now this is used for cluster package so if I make it private to `util` then not able to use for cluster.

> The alternative is to check the error and either return it or treat it is a default 'unamed' case.

Yes, that's best bet, but I was checking what could be error from url.Parse since it accept raw strings https://play.golang.org/p/KuI1HLqcrC not able to get what kind of parse error we can have it here.

pkg/minishift/util/iso_path_test.go
```diff
@@ -0,0 +1,66 @@
...
+
+import (
+	"path/filepath"
+	"testing"
+)
+
+func TestGetIsoPath(t *testing.T) {
+	testData := []struct {
+		provided string
+		expected string
```
LalatenduMohanty: Nitpick: these variable `provided` should be changed to `isoURL`.

LalatenduMohanty:
retest this please

pkg/minishift/util/iso_path_test.go
praveenkumar: @LalatenduMohanty why you think it should be `isoURL` since this is test data set and here we want what we provide and what expect from our tests. I am not convinced by this.

LalatenduMohanty: @praveenkumar It is up to you. Personally `isoURLs` is more appropriate to me. But it shouldnot stop merging this PR.

pkg/minishift/util/iso_path.go
```diff
@@ -0,0 +1,49 @@
...
+	"net/url"
+	"path/filepath"
+	"regexp"
+	"strings"
+)
+
+func GetIsoPath(isoURL string) string {
+	// Make sure isoURL validation happen before using this function.
+	// This function ignore any error from url.Parse
+	uri, _ := url.Parse(isoURL)
```
hferentschik: hmm, at least a comment. Leaves a bad taste in my mouth, but if you think that's enough, so be it

hferentschik:
Rebased and merged.

Issue #1591, opened 2017-10-20
Confusing message when switching profiles


hferentschik:
Following up on issue #1578, I am trying to switch  to an existing profile where the VM is not started:

    $ minishift profile set msa
     oc cli context could not changed for 'msa'. Make sure the profile is in running state or restart if the problem persists.
    Profile 'msa' set as active profile
    $ minishift profile list
     - minishift	Stopped
     - bar		Does Not Exist
     - msa		Stopped	(active)
     - snafu        Does Not Exist

I can now switch to the profile, but still get the message "oc cli context could not changed for 'msa'. Make sure the profile is in running state or restart if the problem persists." Apart of the misleading wording I am wondering why this is relevant. It gives me as a user the impression that something went wrong, but it is not. IMO this message should be removed. Switching to a profile whose VM does not exist or is not running, should just unset any exiting context. The context should then be set when `start` is executed. 

Regarding the message itself. Right now there is at the very least a 'be' missing, as in "oc cli context could not be changed for 'msa'.". However, even then the message is not quite accurate. It should be more like: "No oc CLI context could be activated for the 'msa' profile."

praveenkumar:
@hferentschik I think we should add a check if VM running then only try to set `oc context` and add message which you specified during `minishift set <profile_name>` otherwise this message should be ignored.

hferentschik:
>  I think we should add a check if VM running then only try to set oc context and add message which you specified during minishift set <profile_name> otherwise this message should be ignored.

I am not 100% sure what you mean. Let me see whether I understand you correctly. When we want to switch the oc CLI context as part of a profile switch we check whether the VM is running. If it is running, we try to switch the CLI context. If it succeeds, no worries. If it fails, we display an error. If the VM does not exist or is not started, we unset the current CLI context and no error is reported. Does that align with what you have in mind?

praveenkumar:
> Does that align with what you have in mind?

Yes, that's exactly I have in mind. If a VM doesn't exist or not started then we switch the context but don't show the oc cli context error.

LalatenduMohanty:
> If it is running, we try to switch the CLI context. If it succeeds, no worries. If it fails, we display an error. If the VM does not exist or is not started, we unset the current CLI context and no error is reported

+1. However we do not have any existing code for unsetting the context, so it needs to be written.

hferentschik:
> However we do not have any existing code for unsetting the context, so it needs to be written.

Sure.

anjannath:
[...]

PR #1621, opened 2017-10-26
Issue #1591 Removing misleading  message when switching profiles

anjannath:
Addresses issue #1591

### DIFF
--- a/cmd/minishift/cmd/profile/profile_set.go
+++ b/cmd/minishift/cmd/profile/profile_set.go
@@ -21,13 +21,10 @@ import (
 
 	"github.com/golang/glog"
 	cmdUtil "github.com/minishift/minishift/cmd/minishift/cmd/util"
+	"github.com/minishift/minishift/pkg/minishift/config"
 	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
-)
-
-const (
-	unableToSetOcContextErrorMessage = "Make sure the profile is in running state or restart if the problem persists."
 )
 
 var profileSetCmd = &cobra.Command{
@@ -41,26 +38,38 @@ func runProfile(cmd *cobra.Command, args []string) {
 	validateArgs(args)
 	profileName := args[0]
 
-	if !cmdUtil.IsValidProfile(profileName) {
-		err := profileActions.SetActiveProfile(profileName)
-		if err != nil {
-			atexit.ExitWithMessage(1, err.Error())
-		}
-	} else {
-		err := cmdUtil.SetOcContext(profileName)
+	// Unset the OC CLI context if present for the active profile
+	if cmdUtil.DoesVMExist(config.AllInstancesConfig.ActiveProfile) {
+		err := cmdUtil.RemoveCurrentContext()
 		if err != nil {
 			if glog.V(2) {
 				fmt.Println(fmt.Sprintf("%s", err.Error()))
 			}
-			fmt.Println(fmt.Sprintf("oc cli context could not changed for '%s'. %s", profileName, unableToSetOcContextErrorMessage))
 		}
+	}
 
-		err = profileActions.SetActiveProfile(profileName)
+	if cmdUtil.IsValidProfile(profileName) {
+		err := profileActions.SetActiveProfile(profileName)
 		if err != nil {
 			atexit.ExitWithMessage(1, err.Error())
 		}
+		fmt.Printf("Profile '%s' set as active profile.\n", profileName)
+
+		err = cmdUtil.SetOcContext(profileName)
+		if err != nil {
+			if glog.V(2) {
+				fmt.Println(fmt.Sprintf("%s", err.Error()))
+				fmt.Println(fmt.Sprintf("oc cli context could not changed for '%s'.", profileName))
+			}
+		}
+	} else {
+		err := profileActions.SetActiveProfile(profileName)
+		if err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		}
+		fmt.Printf("Profile '%s' set as active profile.\n", profileName)
 	}
-	fmt.Printf("Profile '%s' set as active profile\n", profileName)
+
 }
 
 func init() {
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -231,6 +231,12 @@ func runStart(cmd *cobra.Command, args []string) {
 		postClusterUp(hostVm, clusterUpConfig)
 		exportContainerImages(hostVm, requestedOpenShiftVersion)
 	}
+	if isRestart {
+		err = cmdUtil.SetOcContext(minishiftConfig.AllInstancesConfig.ActiveProfile)
+		if err != nil {
+			fmt.Println(fmt.Sprintf("Could not set oc CLI context for: %s", profileActions.GetActiveProfile()))
+		}
+	}
 }
 
 // postClusterUp performs configuration action which only need to be run after an initial provision of OpenShift.
--- a/cmd/minishift/cmd/util/oc.go
+++ b/cmd/minishift/cmd/util/oc.go
@@ -17,21 +17,22 @@ limitations under the License.
 package util
 
 import (
+	"encoding/json"
 	"errors"
 	"fmt"
+	"io/ioutil"
+	"os"
 	"path/filepath"
 
 	"github.com/docker/machine/libmachine"
-	configCmd "github.com/minishift/minishift/cmd/minishift/cmd/config"
+	"github.com/golang/glog"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/cache"
-	"github.com/minishift/minishift/pkg/minishift/clusterup"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
 	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
 	"github.com/minishift/minishift/pkg/minishift/oc"
 	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
-	"github.com/spf13/viper"
 )
 
 // CacheOc ensures that the oc binary matching the requested OpenShift version is cached on the host
@@ -54,7 +55,6 @@ func CacheOc(openShiftVersion string) string {
 }
 
 func SetOcContext(profileName string) error {
-
 	profileActions.UpdateProfileConstants(profileName)
 
 	// Need to create the kube config path for the profile for ocrunner to use it.
@@ -78,8 +78,13 @@ func SetOcContext(profileName string) error {
 		return errors.New(fmt.Sprintf("Error getting the IP address: '%s'", err.Error()))
 	}
 
-	requestedOpenShiftVersion := viper.GetString(configCmd.OpenshiftVersion.Name)
-	ocPath := CacheOc(clusterup.DetermineOcVersion(requestedOpenShiftVersion))
+	err, ocPath := GetOcPathForProfile(profileName)
+	if err != nil {
+		if glog.V(2) {
+			fmt.Println(fmt.Sprintf("%s", err.Error()))
+		}
+		return errors.New(fmt.Sprintf("Error getting the oc path for profile '%s'", profileName))
+	}
 
 	ocRunner, err := oc.NewOcRunner(ocPath, kubeConfigPath)
 	if err != nil {
@@ -91,4 +96,45 @@ func SetOcContext(profileName string) error {
 	}
 
 	return nil
+}
+
+//RemoveCurrentContext removes the current context from `machinename_kubeconfig`
+func RemoveCurrentContext() error {
+	ocPath := minishiftConfig.InstanceConfig.OcPath
+	cmd := "config unset current-context"
+
+	ocRunner, err := oc.NewOcRunner(ocPath, filepath.Join(constants.Minipath, "machines", constants.MachineName+"_kubeconfig"))
+	if err != nil {
+		if glog.V(2) {
+			fmt.Println(fmt.Sprintf("%s", err.Error()))
+		}
+		return errors.New(fmt.Sprintf("Error unsetting current-context"))
+	}
+
+	exitCode := ocRunner.RunAsUser(cmd, nil, nil)
+	if exitCode != 0 {
+		return errors.New(fmt.Sprintf("Error during removing current context: %v", err.Error()))
+	}
+	return nil
+}
+
+func GetOcPathForProfile(profileName string) (error, string) {
+	instanceConfigFile := filepath.Join(constants.GetProfileHomeDir(), "machines", profileName+".json")
+	//Check if the file exists
+	_, err := os.Stat(instanceConfigFile)
+	if os.IsNotExist(err) {
+		return nil, ""
+	}
+	raw, err := ioutil.ReadFile(instanceConfigFile)
+	if err != nil {
+		return err, ""
+	}
+
+	var instanceCfg = minishiftConfig.InstanceConfigType{}
+	err = json.Unmarshal(raw, &instanceCfg)
+	if err != nil {
+		fmt.Println(err.Error())
+		return err, ""
+	}
+	return nil, instanceCfg.OcPath
 }
--- a/cmd/minishift/cmd/util/util.go
+++ b/cmd/minishift/cmd/util/util.go
@@ -76,6 +76,12 @@ func GetVMStatus(profileName string) string {
 	return status
 }
 
+func DoesVMExist(profileName string) bool {
+	api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+	defer api.Close()
+	return VMExists(api, profileName)
+}
+
 // UnpackAddons will unpack the default addons into addons default dir
 func UnpackAddons(addonsDir string) error {
 	for _, asset := range DefaultAssets {
--- a/test/integration/features/profile.feature
+++ b/test/integration/features/profile.feature
@@ -81,13 +81,13 @@ Scenario: User should be able to switch between non existing profiles
     When executing "minishift profile set abc" succeeds
       Then stdout should contain
       """
-      Profile 'abc' set as active profile
+      Profile 'abc' set as active profile.
+
       """
     When executing "minishift profile set minishift" succeeds
       Then stdout should contain
       """
-      oc cli context could not changed for 'minishift'. Make sure the profile is in running state or restart if the problem persists.
-      Profile 'minishift' set as active profile
+      Profile 'minishift' set as active profile.
       """
 Scenario: User should not be able to create profile with blank profile name
     When executing "minishift profile set"


cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -26,9 +26,9 @@ import (
 	"github.com/spf13/cobra"
 )
 
-const (
+/*const (
```
hferentschik: -1, if something is not needed anymore, it should be removed. What's your intent with commenting this out?

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -47,15 +47,17 @@ func runProfile(cmd *cobra.Command, args []string) {
 			atexit.ExitWithMessage(1, err.Error())
 		}
 	} else {
-		err := cmdUtil.SetOcContext(profileName)
-		if err != nil {
-			if glog.V(2) {
-				fmt.Println(fmt.Sprintf("%s", err.Error()))
+		if cmdUtil.GetVMStatus(profileName) == "Running" {
```
hferentschik: As I mentioned before, I think we also need to unset the currently active context from the previous profile (if set). 

If I switch to another profile, the current oc CLI context needs to change. If possible it should switch to a new context, but in the case where the profile does not exist or the profile VM is not running, the current CLI context needs to be unset. Otherwise one rins the risk that the user runs oc commands which are intended for a new conext against the old context.

anjannath: @hferentschik i couldn't find any helper function to unset oc cli context, the context will be switched upon `minishift start`.

cmd/minishift/cmd/profile/profile_set.go
anjannath: My bad, left it unknowingly

cmd/minishift/cmd/profile/profile_set.go
anjannath: @hferentschik we would also need a way to find out the current context then.

anjannath:
Sorry centos-ci forgot about you.

cmd/minishift/cmd/profile/profile_set.go
hferentschik: >  i couldn't find any helper function to unset oc cli context, the context will be switched upon minishift start.

Most likely needs to be written. 

> we would also need a way to find out the current context then.

True. We might have to. All this can be done though. How much of the required infrastructure is already in place is a different question.

anjannath: I need to understand first what the setting/unsetting cli context actually do in the code, i'll look into it and reach out to you for further questions.

anjannath: @hferentschik there is `oc config current-context` commnad to get the current context, however i don't know how to unset the context as the context cannot be empty AFAIU.

anjannath:
Updated the PR to get reviews, the error msgs that are shown upon running `oc` commands, such as `oc status` after removing the context are not good.

hferentschik:
> Updated the PR to get reviews,

Ok

>  the error msgs that are shown upon running oc commands, such as oc status after removing the context are not good.

Could you paste the workflow and error messages? Easier for discussion purposes.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -40,22 +36,44 @@ var profileSetCmd = &cobra.Command{
 func runProfile(cmd *cobra.Command, args []string) {
 	validateArgs(args)
 	profileName := args[0]
+	lastActiveProfile := profileActions.GetActiveProfile()
 
 	if !cmdUtil.IsValidProfile(profileName) {
+		if cmdUtil.GetVMStatus(lastActiveProfile) != "Does Not Exist" {
```
hferentschik: That's a tad too many nested if statements here. Also there seems to be quite a bit of duplication between the branches. Can we please refactor this and break into multiple functions + removing some duplication.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -40,22 +36,44 @@ var profileSetCmd = &cobra.Command{
...
 		err := profileActions.SetActiveProfile(profileName)
 		if err != nil {
 			atexit.ExitWithMessage(1, err.Error())
 		}
 	} else {
-		err := cmdUtil.SetOcContext(profileName)
-		if err != nil {
-			if glog.V(2) {
-				fmt.Println(fmt.Sprintf("%s", err.Error()))
+		if cmdUtil.GetVMStatus(lastActiveProfile) != "Does Not Exist" {
```
hferentschik: The state strings should really be constants. Let's try to define them at one place only

anjannath:
The error msg that's shown by `oc`
```
$minishift profile list
- minishift	Does Not Exist	(Active)
$ minishift start
$ minishift profile set abc
Profile 'abc' set as active profile
$ oc status
Error in configuration: 
* context was not found for specified context: minishift
* cluster has no server defined
```

cmd/minishift/cmd/profile/profile_set.go
anjannath: Sure

hferentschik:
> * context was not found for specified context: minishift

Something is not quite right. Why does it say that the 'minishift' profile was not found? it looks to me that we deleted the context information, but left 'minishift' the active context as well. There is something missing in the operations to delete/unset the context.

hferentschik:
@anjannath, FYI the commit message subject, is too long https://circleci.com/gh/minishift/minishift/2561?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

It should not be longer than 90 characters since this is the GitHub ellipsis length. More information is a good thing, just try to keep the subject short and add additional information into the commit message (separated by an empty line from the subject).

anjannath:
well because we used the `oc config delete-context` to delete the context, but the current context is still minishift, because after setting a cli context it is made "active" by `oc config use-context <context-name>`. AFAIK there's no way to unuse the context, so i tried removing the `current-context` from the `~/.kube/config` file, and that gives another horrible looking error.
```
$ oc status
error: Missing or incomplete configuration info.  Please login or point to an existing, complete config file:

  1. Via the command-line flag --config
  2. Via the KUBECONFIG environment variable
  3. In your home directory as ~/.kube/config

To view or setup config directly use the 'config' command.
```

hferentschik:
> AFAIK there's no way to unuse the context, so i tried removing the current-context from the ~/.kube/config file, and that gives another horrible looking error.

My guess is that you broke the config. I am pretty sure one can cleanly remove the context. If not via _oc_, then by loading the config file and manually editing it before saving it back. After all we create this context as well.

anjannath:
I only removed the `current-context` from the config file, here's the diff

```
--- .kube/config	2017-10-30 19:04:10.262127845 +0530
+++ kube_config	2017-10-30 19:03:32.615462965 +0530
@@ -33,11 +33,6 @@
     user: developer/192-168-42-88:8443
   name: /192-168-42-88:8443/developer
 - context:
-    cluster: 192-168-42-88:8443
-    namespace: myproject
-    user: developer/192-168-42-88:8443
-  name: abc
-- context:
     cluster: 127-0-0-1:8443
     namespace: default
     user: system:admin/127-0-0-1:8443
@@ -77,7 +72,7 @@
     namespace: myproject
     user: developer/192-168-42-88:8443
   name: myproject/192-168-42-88:8443/developer
-current-context: abc
+current-context:
 kind: Config
 preferences: {}
 users:
```

cmd/minishift/cmd/start.go
```diff
@@ -230,6 +230,12 @@ func runStart(cmd *cobra.Command, args []string) {
 		postClusterUp(hostVm, clusterUpConfig)
 		exportContainerImages(hostVm, requestedOpenShiftVersion)
 	}
+	if isRestart {
+		err = cmdUtil.SetOcContext(profileActions.GetActiveProfile())
+		if err != nil {
+			fmt.Println(fmt.Sprintf("Could switch oc CLI context for: %s", profileActions.GetActiveProfile()))
+		}
+	}
```
LalatenduMohanty: This is wrong. As user might have started the start with `minishift --profile foo start`. In that case `active` profile is different  than constants.ProfileName. You should use `constants.ProfileName` as it would have the profile name you are interested.

LalatenduMohanty: s/Could switch/Could not switch/

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -61,6 +62,28 @@ func runProfile(cmd *cobra.Command, args []string) {
 		}
 	}
 	fmt.Printf("Profile '%s' set as active profile\n", profileName)
+
+	err := removeOcContext(lastActiveProfile)
+	if err != nil {
+		fmt.Println(err.Error())
+	}
```
LalatenduMohanty: I do not think we should print error here as we are all ready printing it it `removeOcContext`.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -61,6 +62,28 @@ func runProfile(cmd *cobra.Command, args []string) {
 		}
 	}
 	fmt.Printf("Profile '%s' set as active profile\n", profileName)
+
+	err := removeOcContext(lastActiveProfile)
```
LalatenduMohanty: This should be done before setting the new oc context.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -61,6 +62,28 @@ func runProfile(cmd *cobra.Command, args []string) {
...
+	if err != nil {
+		fmt.Println(err.Error())
+	}
+}
+
+func removeOcContext(profileName string) error {
+	if profileName == profileActions.GetActiveProfile() {
+		return nil
+	}
```
LalatenduMohanty: I do not think we need to do this. Without this user can run the same `profile set` command twice and it will redo the every step. Sometimes it is useful when because of some reason things are out of sync.

anjannath: well that was the behaviour that i wanted to avoid, because it would unset the context for the active profile, well it will get reset again when the user starts minishift with `minishift start` but if the vm is already running and the user does profile set the context will be unset. and `oc` commands will not be able to run.

cmd/minishift/cmd/profile/profile_set.go
anjannath: we only print error in `removeOcContext` if logs are enabled `glog.V(2)`

cmd/minishift/cmd/profile/profile_set.go
anjannath: the reson for removing it at last was, if the new context was not able to set we also loose the previous context, which i think would be a bad experience.

LalatenduMohanty: Nope it wont unless user is running the same profile set command over and over again. Because we do not want the previous context for another profile anyway.

cmd/minishift/cmd/profile/profile_set.go
LalatenduMohanty: @anjannath That's fine IMO.

cmd/minishift/cmd/profile/profile_set.go
LalatenduMohanty: IMO we should remove the check as it might help user to sync the context in case it is out of sync for the same active profile.

hferentschik:
@anjannath - so what is the output for the workflow you used inhttps://github.com/minishift/minishift/pull/1621#issuecomment-340436336

anjannath:
@hferentschik now it looks like this:
```
$ minishift profile list
- minishift	Stopped		(Active)
$ minishift profile set test
Profile 'test' set as active profile
$ minishift profile list
- test		Does Not Exist	(Active)
- minishift	Stopped
$ oc status
error: Missing or incomplete configuration info.  Please login or point to an existing, complete config file:

  1. Via the command-line flag --config
  2. Via the KUBECONFIG environment variable
  3. In your home directory as ~/.kube/config

To view or setup config directly use the 'config' command.
```

anjannath:
the integration tests are failing for a broken link in the openshift docs.

hferentschik:
> error: Missing or incomplete configuration info.  Please login or point to an existing, complete config file:

But this still an error right? Something is missing or incomplete.

hferentschik:
I recommend to remove your existing config file away and start with an empty config. Then you diff the file after each of the steps you take. This should help you to detect what the various commands actually do. It also might help you to determine why there is an error reported.

hferentschik:
Maybe this is actually right. It might be _oc_ at fault and its misleading message. They could just say that there is not current context and that you either have to login or point to another config file. We might have to live with this message

anjannath:
Thats right, its the problem eith _oc_. If you query _oc_ about the current using `oc config current-context` then it will say what we desire. But am afraid we will have to live with it for now.

hferentschik:
Looking much better now. However, I think we need to still more work on the message and the order in which they are displayed. Right now I get:

    No oc CLI context could be activated for the 'snafu' profile.
    Profile 'snafu' set as active profile

IMO the order is wrong here. Primarily I want to know " Profile 'snafu' set as active profile". So this message should come first. Then I would expect something like: "No oc CLI context is activated, because the snafu VM does not exist or is not running." Together:

    Profile 'snafu' set as active profile.
    No oc CLI context is activated, because the snafu VM does not exist or is not running. Run 'minishift start first'.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -27,7 +27,7 @@ import (
 )
 
 const (
-	unableToSetOcContextErrorMessage = "Make sure the profile is in running state or restart if the problem persists."
+	vmDoesNotExist = "Does Not Exist"
```
hferentschik: I did not have a local constant in mind, but rather a shared/global constant somewhere in our cmd utility package. This way we can use this constant anywhere we need to compare against the state. 

In fact when it comes to actual libmachine states there are constants for them in github.com/docker/machine/libmachine/state. It is just the "Does not exist" which is the issue. In the cmd util package we have helpers for IsHostRunning. I would expect something like DoesHostExist.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,6 +41,11 @@ func runProfile(cmd *cobra.Command, args []string) {
 	validateArgs(args)
 	profileName := args[0]
 
+	err := removeOcContext(profileActions.GetActiveProfile())
```
hferentschik: I am not sure we should do this regardless. Have you actually tries then switching back to the original profile. It seems deleting the current context and then returning to the profile, leaves me logged out. 

    $ minishift profile set snafu
    No oc CLI context could be activated for the 'snafu' profile.
    Profile 'snafu' set as active profile
    $ minishift profile set minishift
    $  oc get bc
    User "system:anonymous" cannot list buildconfigs in project "myproject"

I was logged in as admin before. Some information is lost here.

anjannath:
I think thats way better then what we show now.

cmd/minishift/cmd/profile/profile_set.go
hferentschik: btw, you are unsetting the current context not removing it. The function name seems misleading.

cmd/minishift/cmd/start.go
hferentschik: probably "could not set or activate"

cmd/minishift/cmd/profile/profile_set.go
anjannath: @hferentschik I couldn't reproduce this, i get an error saying `No resources found`

anjannath:
retest this please

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,26 +37,47 @@ func runProfile(cmd *cobra.Command, args []string) {
 	validateArgs(args)
 	profileName := args[0]
 
+	err := unsetCurrentOcContext(profileActions.GetActiveProfile())
+	if err != nil {
+		fmt.Println(err.Error())
+	}
+
 	if !cmdUtil.IsValidProfile(profileName) {
```
hferentschik: why do you use negated logic in the if statement, especially if you have a if/else? Negated logic is always harder to work out in our brain, since you have to put some mental effort into working out what it actually means. Given that you have both branches either way, why not 'if !cmdUtil.IsValidProfile(profileName)'? Less mental load.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,26 +37,47 @@ func runProfile(cmd *cobra.Command, args []string) {
...
+			fmt.Println(fmt.Sprintf("No oc CLI context is activated for profile, because the '%s' VM does not exist or is not running. Run 'minishift start' first.", profileName))
 		}
 
-		err = profileActions.SetActiveProfile(profileName)
+	}
+
+}
+
+func unsetCurrentOcContext(profileName string) error {
+	if cmdUtil.DoesVMExist(profileName) {
```
hferentschik: Personally I'd take the 'if cmdUtil.DoesVMExist(profileName)' out and do the check on the caller side. This makes the code easier to follow and reveals more clearly that this unsetting of the context is conditional.

cmd/minishift/cmd/util/oc.go
```diff
@@ -92,3 +92,20 @@ func SetOcContext(profileName string) error {
 
 	return nil
 }
+
+//RemoveCurrentContext removes the current context from `machinename_kubeconfig`
+func RemoveCurrentContext() error {
+	requestedOpenShiftVersion := viper.GetString(configCmd.OpenshiftVersion.Name)
```
hferentschik: that does not make any sense. How can you reference _configCmd_ here? This value would only be set as part of the _start_ command. In all other cases you get the default/baseline version. 

So for example, if you start your first cluster against 3.5, we would cache _oc_ 3.5. When you then use profile set to switch to another profile, I would assume you are downloading 3.6 as well and use it to set the context. Don't we write the oc path into the instance config? That's where it should come from. If not, we might have to use the default/baseline _oc_, but at least this should be implicit. @LalatenduMohanty ^^

hferentschik:
$ minishift start
    $ oc whoami
    developer
    $ minishift profile set blabla
    Profile 'blabla' set as active profile
    $ oc whoami
    error: Missing or incomplete configuration info.  Please login or point to an existing, complete config file:

    1. Via the command-line flag --config
    2. Via the KUBECONFIG environment variable
    3. In your home directory as ~/.kube/config

    To view or setup config directly use the 'config' command.
    $ minishift profile set minishift
    $oc whoami
    developer

+1 Example workflow worked. What's about codifying this into a profile integration test?

cmd/minishift/cmd/util/oc.go
gbraad: `configCmd.OpenshiftVersion.Name` is only the referenced name for the configuration parameter.
`viper.GetString()` would read the configured or specified parameter by retrieving the name

for example:

```
$ minishift config set vm-driver virtualbox
$ minishift start
```

will do a pre-flight check with the virtualbox driver:
```
...
-- Starting local OpenShift cluster using 'virtualbox' hypervisor ...
```

In code it does:
https://github.com/minishift/minishift/blob/a9fc054fca4820cb9e249b71e6c28b01a7b123f8/cmd/minishift/cmd/start_preflight.go#L53

So I am not completely following the comment about `configCmd` being an issue here.

LalatenduMohanty:
retest this please

cmd/minishift/cmd/profile/profile_set.go
LalatenduMohanty: @hferentschik I agree. @anjannath Lets fix this in this PR.

cmd/minishift/cmd/profile/profile_set.go
anjannath: Right  @hferentschik moved it out of `unsetCurrentOcContext` now.

cmd/minishift/cmd/profile/profile_set.go
anjannath: ack @LalatenduMohanty Updating the PR with the requested changes.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,26 +38,54 @@ func runProfile(cmd *cobra.Command, args []string) {
...
 		if err != nil {
 			atexit.ExitWithMessage(1, err.Error())
 		}
+		fmt.Printf("Profile '%s' set as active profile\n", profileName)
+
+		if cmdUtil.DoesVMExist(config.AllInstancesConfig.ActiveProfile) {
+			err := unsetCurrentOcContext(config.AllInstancesConfig.ActiveProfile)
+			if err != nil {
+				fmt.Println(err.Error())
+			}
```
LalatenduMohanty: This if block is not required and also it wont be called ever. As if it is not a valid profile the VM wont exist for sure. Profile list is generated from the actual directory structure, so If the VM exists then the profile directory will be present.

anjannath: actually the order is wrong, we need to remove the context of the previously active profile, it has to be done before we set the new profile as active profile.

cmd/minishift/cmd/util/oc.go
LalatenduMohanty: > that does not make any sense. How can you reference configCmd here? This value would only be set as part of the start command. In all other cases you get the default/baseline version.
> 
> So for example, if you start your first cluster against 3.5, we would cache oc 3.5. When you then use profile set to switch to another profile, I would assume you are downloading 3.6 as well and use it to set the context. Don't we write the oc path into the instance config? That's where it should come from. If not, we might have to use the default/baseline oc, but at least this should be implicit

@hferentschik `config.InstanceConfig.OcPath` is set in start.go. So unless a profile is started config.InstanceConfig.OcPath is not set.  When we are trying to remove the context , it means there is fair chance that the old profile will not have the oc binary cached and `config.InstanceConfig.OcPath` would be empty.

cmd/minishift/cmd/util/oc.go
```diff
@@ -92,3 +92,20 @@ func SetOcContext(profileName string) error {
...
+	ocPath := CacheOc(clusterup.DetermineOcVersion(requestedOpenShiftVersion))
```
LalatenduMohanty: During my testing of switching between non existing profiles I did not see oc getting cached. Which is a good thing as we should not download oc while running `profile set command`. But as per code it should download the oc binary. Looking in to the code to understand what is happening here.

hferentschik: How did you test it?

hferentschik:
@LalatenduMohanty @anjannath - where are we on this one now?

anjannath:
I had a bjn discussion with @LalatenduMohanty and gone through the code, and where it was wrong, so now the functions will be using the OcPath present in instanceConfig.OcPath. some refactoring and will update the PR by tomorrow.

gbraad:
Only use the `OcPath` when it has a value... and maybe, we might have to make sure that a restart first clears the value of `OcPath`. I do not think this is currently the cause. Only when it was successful we will see the value (which is correct).

anjannath:
So, after using the `OcPath` in instance config, if we want to switch to an existing and running profile from a profile which doesn't have a VM (and an `OcPath`) we cannot, because there is no _oc_ been cached yet.

```
➜  minishift git:(issue_1591) ✗ minishift profile set test 
/home/anath/.minishift/cache/oc/v3.6.0/oc
/home/anath/.minishift/machines/minishift_kubeconfig
Profile 'test' set as active profile
Error getting information for the VM: test
No oc CLI context is activated for profile, because the 'test' VM does not exist or is not running. Run 'minishift start' first.
➜  minishift git:(issue_1591) ✗ minishift profile set minishift
Error during setting 'minishift' as active profile: The specified path to oc '' does not exist
No oc CLI context is activated for profile, because the 'minishift' VM does not exist or is not running. Run 'minishift start' first.
```
we should use the _oc_ for 'minishift' profile to set the context back to minishift, but since the active profile was 'test', _oc_ in instance config is empty.

i think we should have a helper function which returns the cached _oc_ based on the given profile name from the profile cache directory, and if that doesn't exists we can use the default _oc_.

hferentschik:
@anjannath Using a single approach, like taking _always_ the oc path from instance config is not going to work. For the reasons you described. And yes, you need a helper function which defines an order of things to try. Check source instance config, check target instance config, find any cached oc, download default. Whether we do the "find any cached oc" was part of our resent discussion. I think we said we might not do it for now. Either way, the helper function defines an order to steps to locate _oc_. If all fails, it downloads the default.

LalatenduMohanty:
@hferentschik @anjannath I think we can make it very simple, for example if we are changing from A to B. Then we need to check if VM exists for A if yes, use OcPath from instanceconfig and try to delete the current context. If the VM does not exist then we just go to the next step and see for profile B if the VM exists or not and do the similar.

hferentschik:
@LalatenduMohanty Right, pretty much what I suggested. check A -> check B -> (optional check any) - > download. For now I would also be happy with  check A -> check B  - > download

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,26 +38,54 @@ func runProfile(cmd *cobra.Command, args []string) {
...
 			}
-			fmt.Println(fmt.Sprintf("oc cli context could not changed for '%s'. %s", profileName, unableToSetOcContextErrorMessage))
+			fmt.Println(fmt.Sprintf("No oc CLI context is activated for profile, because the '%s' VM does not exist or is not running. Run 'minishift start' first.", profileName))
+		}
+	} else {
+		if cmdUtil.DoesVMExist(config.AllInstancesConfig.ActiveProfile) {
+			err := unsetCurrentOcContext()
+			if err != nil {
+				fmt.Println(err.Error())
+			}
```
LalatenduMohanty: The piece of code is repeated both in if and else. Which is wrong. It should be present before the if/else starts. 
```
if cmdUtil.DoesVMExist(config.AllInstancesConfig.ActiveProfile) {
			err := unsetCurrentOcContext()
			if err != nil {
				fmt.Println(err.Error())
			}
		}
```

cmd/minishift/cmd/util/oc.go
```diff
@@ -92,3 +87,32 @@ func SetOcContext(profileName string) error {
...
+	exitCode := ocRunner.RunAsUser(cmd, nil, nil)
+	if exitCode != 0 {
+		return errors.New(fmt.Sprintf("Error during removing current context: %v", err.Error()))
+	}
+	return nil
+}
+
+func GetOcPath(profileName string) string {
+	instanceConfigFile := filepath.Join(constants.GetProfileHomeDir(), "machines", profileName+".json")
+	instanceConfig, err := minishiftConfig.NewInstanceConfig(instanceConfigFile)
```
LalatenduMohanty: This is wrong. NewInstanceConfig should be only called in root.go. We need to check if `minishiftConfig.InstanceConfig.OcPath` contains anything. If it does not exist that means the VM does not exist so we need to look for OcPath.

LalatenduMohanty: You can just return `minishiftConfig.InstanceConfig.OcPath` if it exists or null string if it does not exist.

anjannath: the problem with returning `minishiftConfig.InstanceConfig.OcPath` is that, it gives the OcPath of the active profile, but while setting the OcContext we need the OcPath for the profile that we are setting the context to.

hferentschik: You will need to write your own helper. you need to build your own paths and check, not rely on the constants being setup.

anjannath: the function `GetOcPath` loads the `machinename.json` file and reads the `OcPath` from there based on the profilename, if the VM exists this file will be there, if the VM doesn't exist we don't need it anyway. Also, for unsetting the current context we can get the `OcPath` from `instanceConfig.OcPath` only while setting the context we need to call `GetOcPath` because then we need the `OcPath` of the profile we are switching context to, for unsetting we use the `OcPath` of the active profile.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,26 +38,47 @@ func runProfile(cmd *cobra.Command, args []string) {
 	validateArgs(args)
 	profileName := args[0]
 
-	if !cmdUtil.IsValidProfile(profileName) {
+	//need to check if the current active profile has a VM
```
LalatenduMohanty: Nitpick: // Check if the active profile had oc CLI context set or not.

hferentschik:
@anjannath, @LalatenduMohanty  where are we with this?

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,26 +38,47 @@ func runProfile(cmd *cobra.Command, args []string) {
...
+	if cmdUtil.DoesVMExist(config.AllInstancesConfig.ActiveProfile) {
+		err := unsetCurrentOcContext()
+		if err != nil {
+			fmt.Println(err.Error())
+		}
```
LalatenduMohanty: I think we should remove the function unsetCurrentOcContext() and directly call` cmdUtil.RemoveCurrentContext()` here as below. As unsetCurrentOcContext() not adding any value here. Also printing the error for glog.V(2)  is the right thing to do as user do not need to see this and we are not failing the code execution here.   
```
err := cmdUtil.RemoveCurrentContext()
if err != nil {
	if glog.V(2) {
		fmt.Println(fmt.Sprintf("%s", err.Error()))
	}
```

cmd/minishift/cmd/util/oc.go
LalatenduMohanty: @anjannath I understand that it is tricky as you need to get `InstanceConfig.OcPath` but is it not available for the profile you want to set as active.  I am working on the code change required without doing NewInstanceConfig() and will let you know. I do not want to run NewInstanceConfi() as it creates the file if the directory path is present.

cmd/minishift/cmd/profile/profile_set.go
anjannath: +1 to call RemoveCurrentContext directly.

I was making it a function because earlier the VMExists check was part of unsetCurrentContext, we are not doing anything extra in there.

cmd/minishift/cmd/util/oc.go
anjannath: @LalatenduMohanty Sure, we don't want to create a file unnecessarily when the VM for that profile doesn't exists. But before calling `NewInstanceConfig` theres a check for the VM and it will fail if the VM doesn't exist, so we should never hit a situation where the file is not there and `NewInstanceConfig` creates it. But if we can do it without calling that function, that would be _the_ way to do it.

gbraad:
is this still in WIP?

anjannath:
@gbraad Yes, as @LalatenduMohanty mentioned in the [comment](https://github.com/minishift/minishift/pull/1621#discussion_r151409786) we should not call `NewInstanceConfig` in `oc.go`, so when that is resolved it will be done, i am adding [WIP] to this.

gbraad:
@anjannath exactly, as the InstanceConfig on startup only points to the current situation. But in the case of the switch, a different config needs to be read. Waiting until WIP removed.

LalatenduMohanty:
@anjannath Here is the code I had in mind https://github.com/LalatenduMohanty/minishift/commit/b4d66ce029857a43ca63fdc423eef0c644b5ebb7. Please update the PR and we should be good to go.

pkg/minishift/oc/oc.go
```diff
@@ -18,17 +18,19 @@ package oc
...
-	"github.com/minishift/minishift/pkg/util/filehelper"
-	"github.com/pkg/errors"
 	"io"
 	"os"
 	"regexp"
 	"strings"
+
+	"github.com/minishift/minishift/pkg/util/cmd"
+	"github.com/minishift/minishift/pkg/util/filehelper"
+	"github.com/pkg/errors"
```
LalatenduMohanty: @anjannath these are left over changes from some older change.  Please remove these.

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,26 +38,37 @@ func runProfile(cmd *cobra.Command, args []string) {
...
-	} else {
-		err := cmdUtil.SetOcContext(profileName)
+		fmt.Printf("Profile '%s' set as active profile.\n", profileName)
+
+		err = cmdUtil.SetOcContext(profileName)
 		if err != nil {
 			if glog.V(2) {
 				fmt.Println(fmt.Sprintf("%s", err.Error()))
 			}
-			fmt.Println(fmt.Sprintf("oc cli context could not changed for '%s'. %s", profileName, unableToSetOcContextErrorMessage))
```
LalatenduMohanty: IMO`fmt.Println(fmt.Sprintf("oc cli context could not changed for '%s'. %s", profileName))` should be part of the above `if glog.V(2) {` , So that user will come know about context is not set when running debug logs.

anjannath: > IMO `fmt.Println(fmt.Sprintf("oc cli context could not changed for '%s'. %s", profileName))` should be part of the above if glog.V(2)

Done.

pkg/minishift/oc/oc.go
anjannath: Sorry for the delay, i did not see this comment, doing now.

PR #1593, opened 2017-10-20
Issue #1583 Fix localisation issue for active libvirt networks

gbraad:
Fixes #1583

PR #1631, opened 2017-10-27
Adjust ocale to fix the string based network check

### LINKED ISSUE #1583: minishift starts fails on Fedora with German locale
minishift 1.7.0
fedora 26 

Minishift does not start with german locale, when I change to "en_US:UTF-8" it works. 

```
[hygl@localhost ~]$ virsh net-list --all
 Name                 Status     Automatischer Start Bleibend
----------------------------------------------------------

[hygl@localhost ~]$ sudo virsh net-list --all
 Name                 Status     Automatischer Start Bleibend
----------------------------------------------------------
 default              Aktiv      ja            ja

[hygl@localhost ~]$ groups
hygl wheel libvirt docker
[hygl@localhost ~]$ minishift start --metrics
-- Checking if Libvirt default network is present and active ... FAIL
   See the 'Setting Up the Driver Plug-in' topic for more information`
```

fabiand:
Previously we were checking for the (in other languages localized)
string 'active' in order to determine if the default network is active.
This failed whenever the language of the terminal had a translation which
was not the string 'active'.
This patch explicitly sets the language in order to fix this check.

Fixes #1583

### DIFF
--- a/cmd/minishift/cmd/start_preflight.go
+++ b/cmd/minishift/cmd/start_preflight.go
@@ -283,6 +283,9 @@ func checkLibvirtInstalled() bool {
 //checkLibvirtDefaultNetwork returns true if the "default" network is present and active
 func checkLibvirtDefaultNetwork() bool {
 	cmd := exec.Command("virsh", "--connect", "qemu:///system", "net-list")
+	cmd.Env = append(os.Environ(),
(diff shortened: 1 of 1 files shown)


gbraad:
Duplicate of #1593

PR #1640, opened 2017-10-31
Issue #1389 Listing available add-ons in docs

### LINKED ISSUE #1389: Display available add-ons in the documentation
It would be good to display list of available add-ons on the minishift documentation [here](https://docs.openshift.org/latest/minishift/using/addons.html) to bring more visibility.

anjannath:
Fixes #1389

### DIFF
--- a/docs/source/using/addons.adoc
+++ b/docs/source/using/addons.adoc
@@ -241,6 +241,47 @@ You should at least see the *anyuid* add-on listed.
 This is an important add-on that allows you to run images that do not use a pre-allocated UID.
 By default, this is not allowed in OpenShift.
 
+[[table-list-of-default-addons]]
+.Default add-ons
+
+[cols="1,3"]
+|===
+|Add-on Name    | Description
+
+| anyuid
+| Changes the default security context constraints to allow pods to run with arbitrary UID.
+
+| admin-user
+| Creates a user named 'admin' and assigns the *cluster-admin* role to it.  
+| registry-route
+| Creates an edge terminated route for the OpenShift registry.
+
+| xpaas
+| Imports xPaaS templates.
+|===
+
+[[addons-by-community]]
+=== Add-ons by the Community
+
+Apart from the several default add-ons, there are a number of community developed add-ons for {project}.
+Community add-ons can be found in the link:https://github.com/minishift/minishift-addons/tree/master/add-ons[minishift-addons] repository.
+You can get all the information about these add-ons in the link:https://github.com/minishift/minishift-addons[repository].
+The instructions for installing them can be found in the link:https://github.com/minishift/minishift-addons/blob/master/README.adoc[README].
+
+[[installing-addons]]
+== Installing Add-ons
+
+Add-ons are installed with the xref:../command-ref/minishift_addons_install.adoc#[`minishift addons install`] command.
+
+The following example shows how to install an add-on.
+
+[[example-install-addon]]
+.Example: Installing an add-on
+
+----
+$ minishift addons install <path_to_addon_directory>
+----
+
 [[enabling-disabling-addons]]
 == Enabling and Disabling Add-ons
 


docs/source/using/addons.adoc
```diff
@@ -307,6 +307,22 @@ $ minishift addons uninstall admin-user
 Add-on 'admin-user' uninstalled
 ----
 
+[[available-addons]]
+== Available Add-ons
```
hferentschik: There is some lead-in missing. We have not even introduced the minishift-addons repository and suddenly we list add-ons available from this repo? How does the user get them? How do you install them? We will need to duplicate some of the information from the minishift-addons README here. I also argue this section should come last.

docs/source/using/addons.adoc
```diff
@@ -307,6 +307,22 @@ $ minishift addons uninstall admin-user
...
+
+The add-ons link:github.com/minishift/minishift-addons[repository] currently has the following add-ons:
+
+- link:https://github.com/minishift/minishift-addons/tree/master/add-ons/che[Eclipse Che]
```
hferentschik: I am not a big fan of listing add-ons like this, especially with no additional information what the add-on does.
This is so easy to get out of sync. I rather have a generic section about the minishift-addons and how to use it with a link into the add-ons directory. If we curate a manual list, I think some more context would be useful, eg in form of a table.

LalatenduMohanty: -1 For listing add-ons from minishift-addons repository in Minishift documentation as it will be a lot of maintenance to keep this always up to date.  AFAIK the issue https://github.com/minishift/minishift/issues/1389 was about adding information about the default add-ons. @budhrg Can you confirm?

hferentschik: +1 for expanding on the default add-ons and as mentioned inhttps://github.com/minishift/minishift/pull/1640#discussion_r147941078, I am all up for writing a section about the minishift-addon repo. However, in a generic way, w/o listing manually existing add-ons.

anjannath:
in the issue @budhrg mentioned listing the available add-ons, so i figured it was about all the add-ons.

docs/source/using/addons.adoc
anjannath: If we could fetch the list of addons in the repo automatically using some github hooks that'd be great, IMO

budhrg: That would help :+1:

hferentschik: Idea is good, but it needs to integrate into the docs build. Best bet would be some Rake task, but then you still need to insert the data. TBH, I don't think this is worth the effort.

hferentschik:
@anjannath What's the state on this? Are you working on this? If not, let's just close this pull request. It's not an urgent issue anyways.

anjannath:
Yes, i am working on it,since we are eventually going to list them. will have the PR updated by monday.

praveenkumar:
> since we are eventually going to list them.

@anjannath you mean only the default ones not from the minishift-addons repo right?

anjannath:
Yes @praveenkumar, i was thinking of also introducing the addons repo, but not list them, will duplicate the contents of the minishift-addon README.

LalatenduMohanty:
@anjannath Is this still work in progress?

anjannath:
Yes lala, i feel terrible about this, i stare at the screen everyday
thinking what to write, will try to update it by tomorrow.

docs/source/using/addons.adoc
```diff
@@ -241,6 +241,31 @@ You should at least see the *anyuid* add-on listed.
...
+| Creates an edge terminated route for the OpenShift registry.
+
+| xpass
+| Imports xPaaS templates.
+|===
+
+[[addons-by-community]]
+== {project} Add-ons by the community
+
+The {project} community has a number of add-ons in the {project}/addons repo. These addons are community supported.
```
budhrg: @anjannath , Link to README of minishift-addons is missing here I guess.

anjannath: @budhrg Yes, i am still working on this part, will update it soon.

docs/source/using/addons.adoc
```diff
@@ -241,6 +241,47 @@ You should at least see the *anyuid* add-on listed.
...
+
+| xpass
+| Imports xPaaS templates.
+|===
+
+[[addons-by-community]]
+=== Add-ons by the Community
+
+Apart from the several default add-ons, there are a number of community developed add-ons for {project}. We hope that by sharing these add-ons, folks in the community will be able to make the most out of their {project} environment.
+Community add-ons can be found in the link:https://github.com/minishift/minishift-addons/tree/master/add-ons[{project}-addons] repository. You can get all the information about these add-ons in the
```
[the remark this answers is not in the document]
anjannath: it might if the project name changes (which i doubt), but we are using `{project}` to mean 'Minishift' through out the docs.

docs/source/using/addons.adoc
```diff
@@ -241,6 +241,48 @@ You should at least see the *anyuid* add-on listed.
...
+| xpaas
+| Imports xPaaS templates.
+|===
+
+[[addons-by-community]]
+=== Add-ons by the Community
+
+Apart from the several default add-ons, there are a number of community developed add-ons for {project}.
+We hope that by sharing these add-ons, folks in the community will be able to make the most out of their {project} environment.
+Community add-ons can be found in the link:https://github.com/minishift/minishift-addons/tree/master/add-ons[minishift-addons] repository.
```
LalatenduMohanty: We do not need to write the string We hope that by sharing these add-ons, folks in the community will be able to make the most out of their {project} environment. as it is already part of the https://github.com/minishift/minishift-addons/blob/master/README.adoc.

docs/source/using/addons.adoc
```diff
@@ -241,6 +241,48 @@ You should at least see the *anyuid* add-on listed.
...
+You can get all the information about these add-ons in the link:https://github.com/minishift/minishift-addons[repository].
+The instructions for installing them can be found in the link:https://github.com/minishift/minishift-addons/blob/master/README.adoc[README].
+
+[[installing-addons]]
+== Installing Add-ons
```
LalatenduMohanty: What is the intention of adding this section? Is it to add installing section which was not here or provide method to install community add-ons from minishift-addon repository. Because we already hhttps://github.com/minishift/minishift-addons/blob/master/README.adoc#download-use-community-addons

gbraad:
Can we aim to merge this today?

I suggest to remove the duplication from the minishift-addons repo. We can make a simple reference, but we do not need to copy/repeat ourselves.

anjannath:
@gbraad I will make the changes according to the comments, and hopefuly we can merge this today

docs/source/using/addons.adoc
anjannath: @LalatenduMohanty  it is to add the installation section which was missing, now its a complete documentation of the add-ons.

docs/source/using/addons.adoc
```diff
@@ -241,6 +241,47 @@ You should at least see the *anyuid* add-on listed.
...
+
+Add-ons are installed with the xref:../command-ref/minishift_addons_install.adoc#[`minishift addons install`] command.
+
+The following example shows how to install the *prometheus* add-on.
+
+[[example-install-prometheus]]
+.Example: Installing the prometheus add-on
```
LalatenduMohanty: We need to give an generic example here because when we give `prometheus` as an example user might get confused about couple of things, where to get the `prometheus` add-on and what is `prometheus`. So we should say the command to install the add-on is `$ minishift addons install <path_to_addon>`

LalatenduMohanty: I do not think we need a example here as minishift-addon repository has enough documentation around installing add-ons.

anjannath: The example is now `$ minishift addons install <path_to_addon_folder>`

docs/source/using/addons.adoc
```diff
@@ -241,6 +241,47 @@ You should at least see the *anyuid* add-on listed.
...
+
+Add-ons are installed with the xref:../command-ref/minishift_addons_install.adoc#[`minishift addons install`] command.
+
+The following example shows how to install an add-on.
+
+[[example-install-addon]]
+.Example: Installing an add-on
+
+----
+$ minishift addons install <path_to_addon_folder>
```
LalatenduMohanty: Nitpick s/folder/directory/ as I think directory is a better term here.https://superuser.com/questions/187900/what-is-the-difference-between-a-directory-and-a-folder

anjannath: @LalatenduMohanty Thanks, didn't know the difference between them.

Issue #1602, opened 2017-10-23
Remove vmware-fusion mention/support from documentation


praveenkumar:
We should remove VMware-fusion from the host folder documentation since we are not even testing minishift against VMware-fusion and it is confusing for users who refer it and then try out.

hferentschik:
+1

athos-ribeiro:
Should it be removed just from the host folder documentation or from the whole documentation?

gbraad:
Actually, from the whole document. It is not something we can test for as a
target.
At least the reference for the host folders has caused issues of false
expectation before.

gbraad:
![untitled 4](https://user-images.githubusercontent.com/1894/32260342-3f82dc52-bf02-11e7-9220-3d528c756543.jpg)

From the landing page we claim to support VMwarr Fusion. I believe this should be reconsidered in light of effort to make this work (testing, development). /cc: @LalatenduMohanty @praveenkumar @agajdosi

agajdosi:
@gbraad As most of automated tests happen in downstream, we mirror the CDK's requirements, where Win/VB, Win/HyperV, macOS/xhyve, macOS/VB, RHELserver/KVM and RHELworkstation/KVM has been requested.

There should not be a problem with running integration tests as they are with VMware, at least we could try it and see. Once we find it to work or we fix it, the community could start using it if there is a will for it.

I would like to see it automated, but for now I am not sure if we have any options to make it in upstream. For manual starts of integration tests, we can try to to run tests once a sprint I think.

gbraad:
this is the point, it seems VMware Fusion (macOS) is not one of the CDK requirements... or is it? (Note: we do not even talk about VMware Workstation for Windows or Linux).

agajdosi:
@gbraad I can't recall if there is any official document. But AFAIR I have never been asked to test it (same it is for VB on RHELs). When I have started oc CDK 2.x, we had the platforms we have now except later there was a discussion about adding RHEL workstation, so RHEL workstation is one extra platform which has been added.

gbraad:
This needs clarification.

However, minikube has been in discussions to deprecate the vmwarfusion driver anyways, and the workstation driver hasn't seen a lot of development/maintenance. They are in the process of including an updated driver: https://github.com/kubernetes/minikube/pull/2118#issue-268519951

gbraad:
[...]

PR #1643, opened 2017-11-01
Issue #1602 Remove vmware-fusion mention/support from documentation

athos-ribeiro:
Refers to #1602

### DIFF
--- a/docs/source/getting-started/setting-up-driver-plugin.adoc
+++ b/docs/source/getting-started/setting-up-driver-plugin.adoc
@@ -12,7 +12,7 @@ toc::[]
 [[setup-driver-plugin-overview]]
 == Overview
 {project} uses Docker Machine and its driver plug-in architecture to provide a consistent way to manage the OpenShift VM.
-{project} embeds VirtualBox and VMware Fusion drivers so no additional steps are required to use them.
+{project} embeds the VirtualBox driver plugin. No additional steps are required to use it.
 
 However, other drivers require manual installation and setup that are described in the following sections.
 
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -397,6 +397,7 @@ func getDriverOptions(config MachineConfig) (drivers.DriverOptions, error) {
 
 		return createDriverOptions(d, machineConfigOptions)
 	case "vmwarefusion":
+		fmt.Println("VMWare Fusion driver will be deprecated soon. Please consider using other drivers.")
 		if minishiftOs.CurrentOS() != minishiftOs.DARWIN {
 			atexit.ExitWithMessage(1, "vmwarefusion driver is only supported on macOS hosts.")
 		}
--- a/docs/source/getting-started/installing.adoc
+++ b/docs/source/getting-started/installing.adoc
@@ -28,7 +28,6 @@ macOS::
 NOTE: xhyve requires specific installation and configuration steps that are described in the xref:../getting-started/setting-up-driver-plugin.adoc#[Setting Up the Driver Plug-in] section.
 
 - link:https://www.virtualbox.org/wiki/Downloads[VirtualBox]
-- link:https://www.vmware.com/products/fusion[VMware Fusion]
 
 GNU/Linux::
 - link:https://en.wikipedia.org/wiki/Kernel-based_Virtual_Machine[KVM] (default)
--- a/docs/source/using/host-folders.adoc
+++ b/docs/source/using/host-folders.adoc
@@ -33,8 +33,6 @@ These folders are currently not configurable and differ for each driver and OS.
 
 |Virtualbox |Windows |C://Users |/c/Users
 
-|VMWare Fusion |OSX |/Users |/Users
-
 |Xhyve |OSX |/Users |/Users
 |===
 


gbraad:
```
FAIL - commit message does not match "(Issue #[0-9]+ .*|cut v[0-9]+.[0-9]+.[0-9]+)"
```

Commit message should refer to the issue number as follows: "Issue #1602 Remove ..."

praveenkumar:
There are also other places in the doc and also in the code which should be taken care as part of this PR.

Doc section

+ `/docs/source/getting-started/installing.adoc:31:- link:https://www.vmware.com/products/fusion[VMware Fusion]`

Code section

+ https://github.com/minishift/minishift/blob/master/pkg/minikube/cluster/cluster.go#L395-L412
+ https://github.com/minishift/minishift/blob/master/pkg/minikube/machine/client.go#L37-L38

budhrg:
retest this please

budhrg:
> Should they aiso be removed?

@athos-ribeiro Yes please

athos-ribeiro:
Done

LalatenduMohanty:
@minishift/minishift-dev -1 from my side for removing the code related to VMware Fusion. Because we do not want to break the setup of any users who is using VMware fusion or in future if we decide to add the support. I think removing documentation is enough for indicating that it is not supported. Also this issue was only for removing the documentation.

gbraad:
I am for removing the code, but in the sense of a deprecation period. We already kinda mention we do not support the setup when they report an issue involving VMware. At the Minikube side there is:

  * https://github.com/kubernetes/minikube/issues/2058 to drop support

Unless a maintained driver is available the current support is nothing more than flaky. The driver hasn't seen much development. The VMware repo of libmachine has not seen improvements for 2 years on their master: https://github.com/vmware/docker-machine/tree/master/drivers/vmwarefusion. The same for the official repo of libmachine https://github.com/docker/machine/tree/master/drivers/vmwarefusion which has only seen typo changes. Either the code is super-stable... or no-one provides support. The same for the minikube repo: https://github.com/kubernetes/minikube/tree/master/vendor/github.com/docker/machine/drivers/vmwarefusion they synced a year ago and have since not provide any fixes.

There is ongoing work

  * https://github.com/kubernetes/minikube/pull/2118

which is a vmware fusion and workstation driver, but this will not be a drop-in driver for us, as there is likely no RPC support, which we would still need (as we use libmachine unmodified, while minikube has an in-repo modification).

gbraad:
@minishift/minishift-dev Also important to note is that libmachine has no intention to add more functionality: https://github.com/docker/machine/pull/4280#issuecomment-338842944 such as an updated driver.

gbraad:
We will do the code changes in a different PR when we have agreed...

LalatenduMohanty:
retest this please

budhrg:
> I am for removing the code, but in the sense of a deprecation period. We already kinda mention we do not support the setup when they report an issue involving VMware. At the Minikube side there is:

I am too in favour of removing code but certainly not the good idea to remove it completely.
Let's go with deprication.

Depreciation message should goes here as part of of 1.9.1.

https://github.com/minishift/minishift/blob/c677a9853b13c96e6353bacd809da5648d20375f/pkg/minikube/cluster/cluster.go#L396-L412

gbraad:
@budhrg @LalatenduMohanty I am OK with moving the removal back. So this remains a documentation issue... By adding a deprecation notice we can also avoid having to document this further in the documentation.

I will keep an eye on the Minikube development around the driver and see what @anfernee might might do. An alternative approach would be to create an external driver, much like KVM and xhyve has. This way we can avoid direct support.

budhrg:
@athos-ribeiro , could you add the deprecation message 
"VMWare fusion driver is deprecating soon. Please consider using other drivers."

above this line 

https://github.com/minishift/minishift/blob/master/pkg/minikube/cluster/cluster.go#L396

and remove all code changes keeping documentation changes.

/cc @LalatenduMohanty @gbraad

gbraad:
@athos-ribeiro sorry about us going slightly back and forth, but this is because of our userbase and the driver proposal that exists at minikube

athos-ribeiro:
> @athos-ribeiro , could you add the deprecation message

done. I wasn't sure if I should just print the message or log it. I went with the former, but if you prefer otherwise, just let me know.

> and remove all code changes keeping documentation changes.

done

> @athos-ribeiro sorry about us going slightly back and forth, but this is because of our userbase and the driver proposal that exists at minikube

I understand things must be done the right way :) Thanks for the patience guiding me through this PR.

pkg/minikube/cluster/cluster.go
```diff
@@ -393,6 +393,7 @@ func getDriverOptions(config MachineConfig) (drivers.DriverOptions, error) {
 
 		return createDriverOptions(d, machineConfigOptions)
 	case "vmwarefusion":
+		fmt.Println("VMWare fusion driver is deprecating soon. Please consider using other drivers.")
```
robin-owen: VMware Fusion is improperly capitalized.

Personally, I would also prefer "will be deprecated" to "is deprecating".

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -12,7 +12,7 @@ toc::[]
 [[setup-driver-plugin-overview]]
 == Overview
 {project} uses Docker Machine and its driver plug-in architecture to provide a consistent way to manage the OpenShift VM.
-{project} embeds VirtualBox and VMware Fusion drivers so no additional steps are required to use them.
+{project} embeds VirtualBox driver so no additional steps are required to use it.
```
robin-owen: This feels like cleaner writing to me:

```adoc
{project} embeds a VirtualBox driver.
No additional steps are required to use it.
```

At the very least, it should be "a VirtualBox driver".

LalatenduMohanty: @robin-owen a VirtualBox driver" does not sound correct to me as there is only one implementation for the Virtualbox driver. So it should be "the VirtualBox driver".

budhrg:
add to whitelist

docs/source/getting-started/setting-up-driver-plugin.adoc
gbraad: To be nitpicking in that case, it is called a 'driver plugin' in all the documentation and references at docker/machine. This is also what we use in our configuration topic.

robin-owen: > "a VirtualBox driver" does not sound correct to me as there is only one implementation for the Virtualbox driver. So it should be "the VirtualBox driver".

I wasn't sure if there was more than one implementation -- in this case, yes, "the" makes more sense.

> To be nitpicking

This isn't nitpicking if it's more technically accurate to call it "the VirtualBox driver plug-in".

gbraad:
Hope we can merge this soon.

  * [ ] "will be deprecated"
  * [ ] "the VirtualBox driver plugin"

praveenkumar:
@athos-ribeiro ping, Can you please update your PR as per suggestion so we can merge it asap?

athos-ribeiro:
- Fixed the strings as requested
- Rebased against master

praveenkumar:
@LalatenduMohanty changes are already made by @athos-ribeiro, can we merge it now?

LalatenduMohanty:
@athos-ribeiro Thanks for the fix.

Issue #1650, opened 2017-11-02
Minishift start fails when providing ISO URL with http://


robnester-rh:
**Issue:**
Providing a direct URL to an ISO image when executing `minishift start` fails

**Expected results:**
Minishift starts with the specified ISO file

**Actual results:**

```
$ minishift start --iso-url http://artifacts.ci.centos.org/fedora-atomic/minishift/iso/minishift.iso

-- Checking if requested hypervisor 'kvm' is supported on this platform ... OK
-- Checking if KVM driver is installed ... 
   Driver is available at /usr/local/bin/docker-machine-driver-kvm ... 
   Checking driver binary is executable ... OK
-- Checking if Libvirt is installed ... OK
-- Checking if Libvirt default network is present and active ... OK
-- Checking the ISO URL ... FAIL
   See the 'Managing Minishift' topic for more information
```

**Additional info**
```
$ minishift version
minishift v1.8.0+6455fc3
```

gbraad:
This is probably the issue: 

  https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/start_preflight.go#L455

it drops in the else case too soon, and assumes this has to be file://, which means the prefix http?:// is not checked

gbraad:
As a workaround, download the image first and use `file://` instead to specify the image

hferentschik:
Merged by pull request #1652

gbraad:
Just noticed: but the commit description "Issue #1650 Adding test for Windows and simplifying setup" should have included mention of "iso-url". Hopefully next time...

Issue #1651, opened 2017-11-03
Failure in new builds due to external documentation link returning 404


gbraad:
```
06:34:45 - build/openshift/exposing-services.html
06:34:45   *  External link https://docs.openshift.org/latest/dev_guide/getting_traffic_into_cluster.html#using-ingress-IP-self-service failed: 404 No error
```

Ref: https://ci.centos.org/job/minishift-pr/2430/console

gbraad:
The actual documentation at OpenShift is at fault: https://docs.openshift.org/latest/minishift/openshift/exposing-services.html links to a non-existent link https://docs.openshift.org/latest/dev_guide/getting_traffic_into_cluster.html#using-ingress-IP-self-service

hferentschik:
As an at least temporary fix we can link to 3.6 explicitly.

hferentschik:
I guess this is actually what we are after - https://docs.openshift.org/latest/admin_guide/tcp_ingress_external_ports.html#unique-external-ips-ingress-traffic-configure-service

LalatenduMohanty:
I think we replace the link with https://docs.openshift.org/latest/dev_guide/expose_service/expose_internal_ip_nodeport.html#getting-traffic-into-cluster-nodeport

LalatenduMohanty:
Upstream issue: https://github.com/openshift/origin/issues/17165

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/pull/1658

PR #1652, opened 2017-11-03
Issue #1650 Fix preflight check of iso url to allow http and https URI

anjannath:


cmd/minishift/cmd/start_preflight.go
```diff
@@ -448,21 +448,23 @@ func checkIsoUrl() bool {
 	}
 
 	if runtime.GOOS == "windows" {
-		match, _ := regexp.MatchString("^file://[a-zA-Z]:/.+", isoUrl)
+		match, _ := regexp.MatchString("^file://[a-zA-Z]:/.+|^http://[a-zA-Z]:/.+|^https://[a-zA-Z]:/.+", isoUrl)
```
gbraad: `[a-zA-Z]` makes no sense here... but in general, all URLs are case-sensitive.

gbraad:
Likely, the CentOS CI will fail due to #1651

cmd/minishift/cmd/start_preflight.go
```diff
@@ -447,6 +447,10 @@ func checkIsoUrl() bool {
 		return false
 	}
 
+	if strings.HasPrefix(isoUrl, "http://") || strings.HasPrefix(isoUrl, "https://") {
```
hferentschik: So what's about adding some unit tests here? This is a pure string function which can be fully tested. 

Personally, I'd love to see the URL string being passed to _checkIsoUrl_, so that you can easily write tests. If that is not possible, you can also call _viper.SetString_ as part of your test setup and keep the code as is. Either way, it should be easy to add a classic Go table driven test - https://github.com/golang/go/wiki/TableDrivenTests - for this exercising the various cases. 

The test also helps to document the expected values for these URLs, aka shortnames, file path and URLs.

anjannath: well not entirely, we do check if the file exists on disk if its a file path. https://github.com/minishift/minishift/blob/c677a9853b13c96e6353bacd809da5648d20375f/cmd/minishift/cmd/start_preflight.go#L493-L497

cmd/minishift/cmd/start_preflight_test.go
```diff
@@ -0,0 +1,55 @@
...
+package cmd
+
+import (
+	"testing"
+
+	configCmd "github.com/minishift/minishift/cmd/minishift/cmd/config"
+	"github.com/spf13/viper"
+)
+
+func TestIsoUrl(t *testing.T) {
```
hferentschik: +1 Nice with tests :-)

cmd/minishift/cmd/start_preflight_test.go
```diff
@@ -0,0 +1,55 @@
...
+	var isoURLCheck = []struct {
+		in  string
+		out bool
+	}{
+		{"http://github.com/minishift/minishift-centos/minishift-centos.iso", true},
+		{"https://github.com/minishift//minishift/minishift-b2d.iso", true},
+		{"file:///home/foo/bar/bar.iso", true},
```
hferentschik: How is this supposed to pass? You usually don't have this path on your file system. You will need to create a tmp directory and touch an iso file in there. Then write your test against it. See the other test cases which do similar things.

cmd/minishift/cmd/start_preflight_test.go
```diff
@@ -0,0 +1,55 @@
...
+		{"file://home/bar/baz.iso", false},
+		{"file://minishift/hoxier/foo", false},
+		{"/home/joey/chandler/iso.iso", false},
+		{"file:///home/homie/foo", false},
+		{"ram/rahim/anthony.iso", false},
+		{"gopher/minishift/b2d", false},
+		{"file://c:/minishift/gopher.iso", true},
```
hferentschik: You code is OS specific, your tests will have to deal with that as well. See also https://dave.cheney.net/2013/10/12/how-to-use-conditional-compilation-with-the-go-build-tool

gbraad:
retest this please

hferentschik:
@gbraad in its current form the tests will always fail. There are test assumptions which are wrong, for example existence of files which usually don't exist on, as well as OS-specific things which need updating.

cmd/minishift/cmd/start_preflight_test.go
gbraad: @anjannath add a test in a file `start_preflight_test_windows.go` which test for the `file://c:/` situation.

gbraad:
```
   Checking if c:/foo.iso exists ... --- FAIL: TestIsoUrl (0.00s)
	start_preflight_test.go:66: Expected 'true' for given url 'file:///tmp/foo.iso'. Got 'false'.
```

gbraad:
I retriggered to at least get past the documentation check, as this was now fixed.
Looking at it for review and yes, some OS-specific checks should be added to the preflight test.

However, when this would move to validations.go (prio now?), this can be part of the `pkg` tests and it would also make more sense to test in the first place.

anjannath:
retest this please

gbraad:
Filesystem mocks are available; https://github.com/spf13/afero

Haven't used it myself yet, but would like to know if it is easy to use... as this can help with the existence check and dummy files

cmd/minishift/cmd/start_preflight_test.go
anjannath: the tests now will test for both OS, and the regular expression is exercised already, only the `os.stat` was failing.

anjannath:
@gbraad For now i am creating files on disk, but when we move it to validations.go, we can test using the filesystem mock, i looked into it, it should be easy form what i read on their readme.

gbraad:
if it is easy, why not try to add it here anyways?

anjannath:
i already pushing the code before looking into it ;-) i'll work on it now.

hferentschik:
> For now i am creating files on disk, but when we move it to validations.go, we can test using the filesystem mock, 

Personally I stick to what we use already. Why introducing something new here? Mocking the file system does not sounds like a good idea to me.

cmd/minishift/cmd/start_preflight_test.go
```diff
@@ -0,0 +1,83 @@
...
+		if ret != urlTest.out {
+			t.Errorf("Expected '%t' for given url '%s'. Got '%t'.", urlTest.out, urlTest.in, ret)
+		}
+	}
+
+	removeDummyIsoFiles()
+}
+
+func createDummyIsoFiles() {
+	err := os.Mkdir("tmp", 0755)
```
hferentschik: -1, you should use `testDir, err = ioutil.TempDir("", "minishift-test-")`. Please have a look at other places in our codebase where we use this. Where does os.Mkdir create the file anyways. I would think relative to the test, so you might easily end up with unwanted files in the src tree which then accidentally might get committed. 

The alternative to ` ioutil.TempDir` is to use a 'testdata' directory on the same level as the test file. You can then just add the test iso file.

cmd/minishift/cmd/start_preflight_test.go
```diff
@@ -0,0 +1,83 @@
...
+		fmt.Errorf("Failed to create directory")
+	}
+	f, err := os.Create(dummyIsoFile)
+	defer f.Close()
+	if err != nil {
+		fmt.Errorf("Failed to create dummy ISO file")
+	}
+}
+
+func removeDummyIsoFiles() {
```
hferentschik: -1, removal needs to happen via _defer_

cmd/minishift/cmd/start_preflight_test.go
```diff
@@ -0,0 +1,83 @@
...
+	for _, urlTest := range isoURLCheck {
+		viper.Set(configCmd.ISOUrl.Name, urlTest.in)
+		defer viper.Reset()
+		ret := checkIsoUrl()
+		if ret != urlTest.out {
+			t.Errorf("Expected '%t' for given url '%s'. Got '%t'.", urlTest.out, urlTest.in, ret)
+		}
+	}
+
+	removeDummyIsoFiles()
```
hferentschik: this might fail. You want to ensure the cleanup via a _defer_.

cmd/minishift/cmd/start_preflight_test.go
```diff
@@ -0,0 +1,83 @@
...
+		{"c:/gohome/foo.iso", false},
+		{"http:/github.com/minishift/minishift.iso", false},
+		{"file:/home/gopher/minishift-b2d.iso", false},
+	}
+
+	createDummyIsoFiles()
+
+	for _, urlTest := range isoURLCheck {
+		viper.Set(configCmd.ISOUrl.Name, urlTest.in)
+		defer viper.Reset()
```
hferentschik: -1 don't call _defer_ in a loop. _defer_ makes sure the method is guaranteed to execute when the function call ends. Calling _defer_ multiple times does not have an effect, except creating confusion.

cmd/minishift/cmd/start_preflight_test.go
```diff
@@ -0,0 +1,83 @@
...
+		{"https://github.com/minishift//minishift/minishift-b2d.iso", true},
+		{strings.Replace("file://"+dummyIsoFile, "\\", "/", -1), true},
+		{"file://home/bar/baz.iso", false},
+		{"file://minishift/hoxier/foo", false},
+		{"/home/joey/chandler/iso.iso", false},
+		{"file:///home/homie/foo", false},
+		{"ram/rahim/anthony.iso", false},
+		{"gopher/minishift/b2d", false},
+		{"file://d:/gopher/minishift/b2d", false},
+		{"c:/gohome/foo.iso", false},
```
hferentschik: what happened to the Windows tests?

cmd/minishift/cmd/start_preflight_test.go
anjannath: Does it also create a temporary directory in windows?, initially i created the directory in `/tmp` using `os.Mkdir` but then windows doesn't have a temp.

cmd/minishift/cmd/start_preflight_test.go
LalatenduMohanty: 👍

cmd/minishift/cmd/start_preflight_test.go
LalatenduMohanty: @hferentschik I do not see anyway to test the windows path on Linux machines unless we run the unit tests on a Windows machine.

cmd/minishift/cmd/start_preflight_test.go
```diff
@@ -0,0 +1,83 @@
+/*
+Copyright (C) 2016 Red Hat, Inc.
```
budhrg: 2017 as year here.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -477,6 +477,9 @@ func checkIsoUrl() bool {
 	if !strings.HasSuffix(isoUrl, ".iso") {
 		return false
 	}
+	if strings.HasPrefix(isoUrl, "http://") || strings.HasPrefix(isoUrl, "https://") {
```
budhrg: Try to use regex here https://play.golang.org/p/M1MqtTY6iK

cmd/minishift/cmd/start_preflight_test.go
anjannath: Effect of copy paste :-P

cmd/minishift/cmd/start_preflight.go
```diff
@@ -467,17 +467,23 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
 }
 
 // checkIsoUrl checks the Iso url and returns true if the iso file exists
-func checkIsoUrl() bool {
+func checkIsoURL() bool {
```
budhrg: :+1:

cmd/minishift/cmd/start_preflight.go
```diff
@@ -467,17 +467,23 @@ func isMounted(driver drivers.Driver, mountpoint string) (bool, error) {
...
 			return true
 		}
 	}
+
 	if !strings.HasSuffix(isoUrl, ".iso") {
 		return false
 	}
 
+	validScheme := regexp.MustCompile(`^https?://`)
+	if validScheme.MatchString(isoUrl) {
```
budhrg:
```
if validScheme := regexp.MustCompile(`^https?://`); validScheme.MatchString(isoUrl) {
   return true
}
```

cmd/minishift/cmd/start_preflight_test.go
hferentschik: >  I do not see anyway to test the windows path on Linux machines unless we run the unit tests on a Windows machine.

But we do run tests on Windows machines. Appveyor. Just add another test file with the appropriate name or build annotation and the tests specified in there will only run on Windows? We do this already.

cmd/minishift/cmd/start_preflight.go
hferentschik: If you just want to match a string regexp.MatchString is probably a better choice. See also https://gobyexample.com/regular-expressions

anjannath: `regexp.MatchString` is what used in the following string matches. I do not know the advantage or disadvantage of one over the other. this does the job in two lines.

budhrg: `regexp.MatchString` works for me. It's ideal at this point.

`With MatchString, we see if a pattern can match a string. This tells us if the string satisfies the pattern's requirements. We use Compile to create reusable objects.` from  here https://www.dotnetperls.com/regexp-go

anjannath: I think `regexp.MatchString` should be used here, Thanks @hferentschik for the pointer to that.

hferentschik:
@anjannath since I did not know anymore how to explain what I mean, I created a pull request - https://github.com/anjannath/minishift/pull/1. It shows how to do the dummy iso easier and adds Windows-specific tests. Please integrate into your pull request.

It also shows an issue on Windows. @anjannath did you test this on Windows. The moment we are adding OS specific tests we need to make sure that this gets tested. Especially file operations. The difference between OS types in this area is source of many issues.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -490,7 +496,6 @@ func checkIsoUrl() bool {
 		}
 	}
 
-	fmt.Printf("\n   Checking if %s exists ... ", strings.TrimPrefix(isoUrl, "file://"))
 	_, err := os.Stat(strings.TrimPrefix(isoUrl, "file://"))
```
hferentschik: Here is a bug on Windows. You are enforcing a Windows path to be specified with forward slashes, aka file://c:/path/to/iso/my.iso. Only in this format, you let Windows paths get to the os.Stat call. There you trim off the file prefix and do an os.Stat with c:/path/to/iso/my.iso. I believe this is not going to work. Here is the result of the unit tests as run on AppVeyor:

    --- FAIL: TestIsoUrlOnWindows (0.00s)
	start_preflight_windows_test.go:48: Expected 'true' for given url 'file://C:/gopath/src/github.com/minishift/minishift/test/testdata/dummy.iso'. Got 'false'.

When dealing with file path, it is absolute essential to tests this across platforms. At least manually, preferably in a unit/integration test.

On a side node, we also have filehelper.Exists which wraps os.Stat in a function returning true or false.

cmd/minishift/cmd/start_preflight_test.go
hferentschik: @LalatenduMohanty https://github.com/anjannath/minishift/pull/1/commits/1241a13deec53c54b0e2ebdd4894fc602fbe006a

cmd/minishift/cmd/start_preflight.go
anjannath: No, i did not test this in windows locally, that bug i see it now. I will integrate the PR and test it. Thanks @hferentschik

hferentschik:
Merged, thanks

Issue #1629, opened 2017-10-27
Minishift fails to start on Arch Linux with error `failed to initialize KVM: Permission denied`


anjannath:
The complete error message looks something like this: 
```
-- Starting Minishift VM .... FAIL E1027 09:19:46.631295    1869 start.go:356] Error starting the VM: 
Error starting stopped host: [Code-1] [Domain-10] internal error: process exited while connecting to 
monitor: Could not access KVM kernel module: Permission denied
2017-10-27T03:49:46.495880Z qemu-system-x86_64: failed to initialize KVM: Permission denied. 
Retrying.
Error starting the VM: Error starting stopped host: [Code-1] [Domain-10] internal error: process exited 
while connecting to monitor: Could not access KVM kernel module: Permission denied
2017-10-27T03:49:46.495880Z qemu-system-x86_64: failed to initialize KVM: Permission denied
```

This happens because correct permissions are not set for the user who is running minishift.

To fix it, after [installing the docker-machine-driver-kvm](https://docs.openshift.org/latest/minishift/getting-started/setting-up-driver-plugin.html#kvm-driver-install), the current user needs to be added to `kvm` group, and qemu has to be told to use that group to run the instances.

In the qemu configuration file `/etc/libvirt/qemu.conf`.  add `group="kvm"`.
Then restart the machine for the changes to take effect, and NO a log out login won't work.

gbraad:
Can you please explain the steps that were followed that led to this. Were the install instructions followed?

anjannath:
@gbraad Yes, even after following the install instructions in the this error occures in Arch Linux.

gbraad:
Thanks for updating the description. So in short, for some distros, such as Argh Linux, you might need to change the group of the libvirt daemon.

This might be a documentation thing... however, it is also very specific.

anjannath:
@gbraad I think this is something that should go into the FAQ, as you said it's very specific to _Arch Linux_

gbraad:
Well, we do have instructions for Debian?Ubuntu, and Fedora: https://github.com/minishift/minishift/blob/master/docs/source/getting-started/setting-up-driver-plugin.adoc#on-debianubuntu, so a specific _Argh Linux_ comment would be OK

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/pull/1654

PR #1654, opened 2017-11-03
Issue #1629 Documentation of steps for setting up kvm driver on arch linux

anjannath:
addresses issue #1629

### DIFF
--- a/docs/source/getting-started/setting-up-driver-plugin.adoc
+++ b/docs/source/getting-started/setting-up-driver-plugin.adoc
@@ -77,6 +77,33 @@ $ sudo usermod -a -G libvirt <username>
 $ newgrp libvirt
 ----
 
+[[kvm-driver-arch-linux]]
+=== On Arch Linux
+
+. Install *libvirt*, *qemu*, *dnsmasq* and *ebtables* on your system:
++
+----
+$ sudo pacman -S libvirt qemu dnsmasq ebtables
+----
+
+. Add yourself to *kvm* and *libvirt* groups:
++
+----
+$ sudo usemod -a -G kvm,libvirt <username>
+----
+
+. Update libvirt configuration in */etc/libvirt/qemu.conf*:
++
+----
+$ sudo sed -r 's/group=".+"/group="kvm"/1' /etc/libvirt/qemu.conf > /etc/libvirt/qemu.conf
+---- 
+
+.  Update your current session to apply the group change:
++
+----
+$ newgrp libvirt
+----
+
 [[xhyve-driver-install]]
 == xhyve Driver
 


docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -77,6 +77,33 @@ $ sudo usermod -a -G libvirt <username>
 $ newgrp libvirt
 ----
 
+[[kvm-driver-arch-linux]]
+=== On Arch Linux
```
hferentschik: What's the specific part? And why do we want to document it explicitly?

hferentschik: I am using Mint? Can you add instructions for that as well?

anjannath: we got a lot of questions on the IRC about it also an issue got opened, the specific parts are the two seperate packages that has to be installed and making changes to the `/etc/libvirt/qemu.conf`.

hferentschik: Sure, if the majority of @minishift/minishift-dev agrees, let's add it. I'm neutral

LalatenduMohanty:
retest this please

docs/source/getting-started/setting-up-driver-plugin.adoc
LalatenduMohanty: I am in favor of adding it as it involves an extra step. I am going to approve this PR. However we should just add the extra step required for Arch Linux and we should not repeat steps. May be we can take another look later to organize this better.

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -77,6 +77,33 @@ $ sudo usermod -a -G libvirt <username>
...
++
+----
+$ sudo pacman -S libvirt qemu dnsmasq ebtables
+----
+
+. Add yourself to *kvm* and *libvirt* groups:
++
+----
+$ sudo usemod -a -G kvm,libvirt <username>
+----
```
LalatenduMohanty: Any specific reason user needs to be added to both `kvm` and `libvirt` groups? For other distribution we only add to libvirt group.

LalatenduMohanty:
Moving to work in progress after talking to @anjannath.

docs/source/getting-started/setting-up-driver-plugin.adoc
gbraad: Linux Mint and Ubuntu are identical in setup, but Arch Linux involves additional steps. This is particularly an important is distro in Asia which is used by our target audience

gbraad:
What is the reason for the WIP?

LalatenduMohanty:
> What is the reason for the WIP?

https://github.com/minishift/minishift/pull/1654#discussion_r148847460

gbraad:
The reason why both `kvm` and `libvirt` group ? ... I can not answer this... I am not an Argh User

anjannath:
I have to find out and check if it is necessary to add to both the groups, right now don't want to break minishift on my system, need to test other stuffs.

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -77,6 +77,33 @@ $ sudo usermod -a -G libvirt <username>
...
+. Install *libvirt*, *qemu*, *dnsmasq* and *ebtables* on your system:
++
+----
+$ sudo pacman -S libvirt qemu dnsmasq ebtables
+----
+
+. Add yourself to *kvm* and *libvirt* groups:
++
+----
+$ sudo usemod -a -G kvm,libvirt <username>
```
praveenkumar: @anjannath for us it something like below, this looks like a long pending PR, can you do a check + rebase so we can merge it?

```
$ sudo groupmems -g kvm -l
qemu 

$ sudo groupmems -g libvirt -l
prkumar 
```

anjannath: @praveenkumar Sure.. on it now!

anjannath:
```
➜  ~ sudo groupmems -g kvm -l
anath 
➜  ~ sudo groupmems -g libvirt -l
anath 
```
there is no `qemu` user present in Arch Linux by default.

docs/source/getting-started/setting-up-driver-plugin.adoc
anjannath: checked by reinstalling all the packages and doing the configurations, and it is required to be part of both the groups, but still do not know for sure, why the user needs to be added to both the groups. My guess is that, both `kvm` and `libvirt` groups permission is required for running a kvm virtual machine in Arch Linux.

Issue #1498, opened 2017-10-05
Add integration test for deployment of EAP applications


agajdosi:
We should add a test which would check if image streams added by xpaas add-on are working. This check could consist of deploying several application templates from image stream. Planned steps are:

1   deploy template (e.g. oc new-app eap64-basic-s2i)
2    check endopoint of deployed application
3    delete everything in project (e.g. oc delete all --all)
4    repeat

agajdosi:
I have a working prototype for this issue, which deploys and checks the `eap70-basic-s2i` template from xpaas image stream:

```gherkin
  Scenario: Deploy eap70-basic-s2i
   Given Minishift has state "Running"
    When executing "oc new-app eap70-basic-s2i" succeeds
    Then stdout should contain "Success"

  Scenario: Check
    When services "eap-app" rollout successfully
     And body of HTTP request to "/index.jsf" of service "eap-app" in namespace "myproject" contains "Welcome to JBoss!"
     And status code of HTTP request to "/index.jsf" of service "eap-app" in namespace "myproject" is equal to "200"

  Scenario: Delete all in project
    When executing "oc delete all --all" succeeds
    Then stdout should contain
     """
     service "eap-app" deleted
     """
```

Next steps will be to transform this into a Scenario Outline so we could test deployment of large number of JBoss and FSI application templates.

hferentschik:
Merged via pull request #1657

odockal:
@agajdosi Thanks!

agajdosi:
@odockal Thanks for submitting idea for this test. Just want to add - for now the test is not a part of PR acceptance testing and we do not have nightly builds and tests. However the nightlys will come soon and then xpaas will be covered.

odockal:
@agajdosi Sure thing. Alright, please, let me know when nightly will be available.

PR #1657, opened 2017-11-03
Issue #1498 Add integration test for deployment of EAP applications

agajdosi:
This test checks if templates added by xpaas addon are available and if they can be deployed successfully.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -196,14 +196,25 @@ test: vendor $(ADDON_ASSET_FILE)
 	@go test -ldflags="$(VERSION_VARIABLES)" -v $(shell $(PACKAGES))
 
 .PHONY: integration
+integration: GODOG_OPTS = --tags=basic
 integration: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
-	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) --run-before-feature="$(RUN_BEFORE_FEATURE)" --tags basic $(GODOG_OPTS)
+	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
+	--run-before-feature="$(RUN_BEFORE_FEATURE)" $(GODOG_OPTS)
 
 .PHONY: integration_all
+integration_all: GODOG_OPTS = --tags=~coolstore
 integration_all: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
-	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) --run-before-feature="$(RUN_BEFORE_FEATURE)" $(GODOG_OPTS) --tags ~coolstore
+	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
+	--run-before-feature="$(RUN_BEFORE_FEATURE)" $(GODOG_OPTS)
+
+.PHONY: integration_pr
+integration_pr: GODOG_OPTS = --tags=~coolstore\&\&~addon-xpaas
+integration_pr: $(MINISHIFT_BINARY)
+	mkdir -p $(INTEGRATION_TEST_DIR)
+	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
+	--run-before-feature="$(RUN_BEFORE_FEATURE)" $(GODOG_OPTS)
 
 .PHONY: fmt
 fmt:
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -298,7 +298,7 @@ function build_and_test() {
   cd $GOPATH/src/github.com/minishift/minishift
   make prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration_all
+  MINISHIFT_VM_DRIVER=kvm make integration_pr
   echo "CICO: Tests ran successfully"
 
   artifacts_upload_on_pr_and_master_trigger $1;
--- a/docs/source/contributing/developing.adoc
+++ b/docs/source/contributing/developing.adoc
@@ -221,11 +221,16 @@ $ make integration
 
 NOTE: By default `make integration` only runs tests which are tagged as `@basic`.
 
-To run all the test, use the following command:
+To run all tests, use the following command:
 
 ----
 $ make integration_all
 ----
+
+[NOTE]
+====
+There is also a `make integration_pr` target which is being run as part of pull request testing.
+====
 
 ===== Additional Parameters
 
@@ -281,6 +286,12 @@ Set `no-colors` to true to disable ansi colors of Godog's output.
 
 Definitions::
 Set `definitions` to true to print all available step definitions.
+
+[NOTE]
+====
+Passing any value via `GODOG_OPTS` overrides the default tag definition on each integration target.
+Thus in this case `--tags` must be specified manually, otherwise all features will be run.
+====
 
 For example, to run integration tests on two specific feature files using only `@basic` and `@openshift` tags and without ansi colors, the following command can be used:
 
--- a/test/integration/features/addon-xpaas.feature
+++ b/test/integration/features/addon-xpaas.feature
@@ -1,0 +1,35 @@
+@addon-xpaas @addon
+Feature: xpaas add-on
+Xpaas add-on imports xPaaS templates and imagestreams,
+which are then available in OpenShift to the user.
+
+  Scenario: User enables the xpaas add-on
+     When executing "minishift addons enable xpaas" succeeds
+     Then stdout should contain "Add-on 'xpaas' enabled"
+
+  Scenario: User starts Minishift
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start --memory 4GB" succeeds
+     Then Minishift should have state "Running"
+      And stdout should contain
+      """
+      XPaaS OpenShift imagestream and templates installed
+      """
+
+  Scenario Outline: User deploys, checks out and deletes several templates from XpaaS imagestream
+   Given Minishift has state "Running"
+    When executing "oc new-app <template-name>" succeeds
+     And services "<service-name>" rollout successfully
+    Then body of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" contains "<expected-hello>"
+     And status code of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" is equal to "200"
+     And executing "oc delete all --all" succeeds
+
+  Examples: Required information to test the templates
+    | template-name           | service-name   | http-endpoint | expected-hello                        |
+    | datagrid65-basic        | datagrid-app   | /             | Welcome to the JBoss Data Grid Server |
+    | eap64-basic-s2i         | eap-app        | /index.jsf    | Welcome to JBoss!                     |
+    | eap70-basic-s2i         | eap-app        | /index.jsf    | Welcome to JBoss!                     |
+
+  Scenario: User deletes Minishift
+     When executing "minishift delete --force" succeeds
+     Then Minishift should have state "Does Not Exist"


test/integration/features/addon-xpaas.feature
```diff
@@ -0,0 +1,34 @@
...
+
+  Scenario Outline: User deploys, checks out and deletes several templates from XpaaS imagestream
+   Given Minishift has state "Running"
+    When executing "oc new-app <template-name>" succeeds
+     And services "<service-name>" rollout successfully
+    Then body of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" contains "<expected-hello>"
+     And status code of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" is equal to "200"
+     And executing "oc delete all --all" succeeds
+
+  Examples: Values to be used on next Minishift start
```
hferentschik: I guess the example description is not quite right

hferentschik:
@agajdosi See also https://github.com/minishift/minishift/pull/1658

praveenkumar:
@agajdosi How much extra time does it take when running this specific test case, also are we planning to make it as part of each PR of this is something only going to run on nightly?

LalatenduMohanty:
retest this please

gbraad:
retest this please

test/integration/features/addon-xpaas.feature
```diff
@@ -0,0 +1,34 @@
...
+     When executing "minishift start" succeeds
+     Then Minishift should have state "Running"
+      And stdout should contain
+      """
+      XPaaS OpenShift imagestream and templates installed
+      """
+
+  Scenario Outline: User deploys, checks out and deletes several templates from XpaaS imagestream
+   Given Minishift has state "Running"
+    When executing "oc new-app <template-name>" succeeds
```
amitkrout: I don't understand this, how ```<template-name>``` is being parsed ?

hferentschik: It is a so-called Secario Outline - https://github.com/cucumber/cucumber/wiki/Scenario-Outlines.

A scenario is repeated based on some variables defines in the examples table.

hferentschik:
> also are we planning to make it as part of each PR of this is something only going to run on nightly?

I think it would be time to create this nightly build we have been talking about

praveenkumar:
> I think it would be time to create this nightly build we have been talking about

@hferentschik  yes, agree 👍

agajdosi:
@praveenkumar It takes quite a while, sometimes maybe 15+ minutes. I am +1 for nightly test as @hferentschik mentioned.

test/integration/features/addon-xpaas.feature
agajdosi: thanks for a catch, missed it, fixing it now

agajdosi:
I see that deployment of OpenJDK example does not pass the Central-CI and also now I am seeing it on my local workstation. This was not here before, investigating more.

agajdosi:
So OpenJDK is not stable, interesting is that it can pass 6 times in a row and then fail 4 times in a row. I will investigate more by manual checks, where I am still unlucky to reproduce. However for this issue I think I could just remove this one example and later focus on adding it when I will get the root of the issue.

agajdosi:
@hferentschik Suggested changes implemented. The test is not going to be run on pull request. Please take a look when you have some time.

hferentschik:
>  Suggested changes implemented. The test is not going to be run on pull request. Please take a look when you have some time.

Ok, I take a look at it first thing tomorrow.

Makefile
```diff
@@ -198,12 +198,12 @@ test: vendor $(ADDON_ASSET_FILE)
 .PHONY: integration
 integration: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
-	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) --run-before-feature="$(RUN_BEFORE_FEATURE)" --tags basic $(GODOG_OPTS)
+	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) --run-before-feature="$(RUN_BEFORE_FEATURE)" --tags basic "$(GODOG_OPTS)"
```
LalatenduMohanty: @agajdosi Any specific reason for this change? It was working fine previously, right?

agajdosi:
Just rebased on current master.

test/integration/features/addon-xpaas.feature
```diff
@@ -0,0 +1,33 @@
...
+     And services "<service-name>" rollout successfully
+    Then body of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" contains "<expected-hello>"
+     And status code of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" is equal to "200"
+     And executing "oc delete all --all" succeeds
+
+  Examples: Template names and values to be used with each of them
```
LalatenduMohanty: Suggestion: `Required information to test the templates`

agajdosi: ok

Makefile
agajdosi: @LalatenduMohanty It was working, because we did not use multiple tags concatenated with multiple "&&" operands. In this case all the tags between "&&"s got ignored. Due to this I have put it into brackets, which prevents this issue.

agajdosi: @LalatenduMohanty Another possibility would be to use just "positive" tags and keep the negative just for `cdk-only` or `minishift-only`. Maybe this would be more readable. WDYT?

hferentschik: -1, afaict this breaks the use of GODOG_OPTS completely. I actually created an issue - https://github.com/minishift/minishift/issues/1679, but it turns out that the quotes cause the problem. As soon as you provide GODOG_OPTS things fall apart. For example:

    $ make integration GODOG_OPTS="--tags addon-xpaas"

    flag provided but not defined: -tags addon-xpaas
    Usage of /var/folders/39/70n784613gd370g63ggg62lh0000gn/T/    go-build708633840/github.com/minishift/minishift/test/    integration/_test/integration.test:
      -binary string
        	Path to minishift binary
      -definitions
    
      -format string
        	Sets which format godog will use (default "pretty")
      -minishift-args string
        	Arguments to pass to minishift
      -no-colors
        	Disable colors in godog output
      -paths string
        	 (default "./features")
      -run-before-feature string
        	Set of minishift commands to be executed before every feature.     Individual commands must be delimited by a semicolon.
      -stop-on-failure
        	Stop when failure is found
      -tags string
        	Tags for godog test
      -test-dir string
        	Path to the directory in which to execute the tests
      -test.bench regexp
        	run only benchmarks matching regexp
      -test.benchmem
        	print memory allocations for benchmarks
      -test.benchtime d
        	run each benchmark for duration d (default 1s)
      -test.blockprofile file
        	write a goroutine blocking profile to file
      -test.blockprofilerate rate
        	set blocking profile rate (see runtime.SetBlockProfileRate) (    default 1)
      -test.count n
        	run tests and benchmarks n times (default 1)
      -test.coverprofile file
        	write a coverage profile to file
      -test.cpu list
        	comma-separated list of cpu counts to run each test with
      -test.cpuprofile file
        	write a cpu profile to file
      -test.list regexp
        	list tests, examples, and benchmarch maching regexp then exit
      -test.memprofile file
        	write a memory profile to file
      -test.memprofilerate rate
        	set memory profiling rate (see runtime.MemProfileRate)
      -test.mutexprofile string
        	write a mutex contention profile to the named file after     execution
      -test.mutexprofilefraction int
        	if >= 0, calls runtime.SetMutexProfileFraction() (default 1)
      -test.outputdir dir
        	write profiles to dir
      -test.parallel n
        	run at most n tests in parallel (default 4)
      -test.run regexp
        	run only tests and examples matching regexp
      -test.short
        	run smaller test suite to save time
      -test.timeout d
        	panic test binary after duration d (0 means unlimited)
      -test.trace file
        	write an execution trace to file
      -test.v
        	verbose: print additional output
    exit status 2
    FAIL	github.com/minishift/minishift/test/integration	0.067s
    make: *** [integration] Error 1    

I think some work is needed here. Documentation-wise as well as how the integration targets work. Maybe we should actually re-open #1679.

centos_ci.sh
```diff
@@ -298,7 +298,7 @@ function build_and_test() {
   cd $GOPATH/src/github.com/minishift/minishift
   make prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration_all
+  MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="--tags=~coolstore&&~addon-xpaas"
```
hferentschik: I recall an issue where we wanted to make sure that the target called by CI is a direct Makefile target and that we did not want to use custom properties within the CI script. This was to make it clear which features are run in CI jsut be looking at the Makefile. What happened to this issue?

test/integration/features/addon-xpaas.feature
```diff
@@ -0,0 +1,33 @@
+@addon-xpaas @addon
```
hferentschik: +1 Test itself looks great.

centos_ci.sh
agajdosi: @hferentschik I just remember talk about integration_nightly/all, however the new target for CI seems reasonable for me. I will create it and will not use exclusion of tags, but instead will define the test directly tag by tag so it is easily understandable and does not face the issue with multiple ampersands. For now I will just mention in docs that defining multiple exclusive tags has issues and thus we recommend to define tags inclusively.

Possible follow up fix for making the tags more usable could be probably to move the tags from GODOG_OPTS to its own new parameter GODOG_TAGS.

hferentschik:
@agajdosi, regarding the Makefile, I think we are better off using target-specific variables for GODOG_OPTS. See also https://www.gnu.org/software/make/manual/html_node/Target_002dspecific.html. This allows having defaults for differentiating between _integration_ and _integration_all_ target, as well as giving the flexibility to run any combination one wants using GODOG_OPTS from the command line.

Have a  look at https://github.com/agajdosi/minishift/pull/3 and see whether this solves the "problem" which made you adding the quotes.

agajdosi:
@hferentschik Thank for code example on PR and link, missed the target specific variables before. This could be the way, but it still faces the problems I tried to fixed with quotes.

I have modified the code of test runner a little bit to see more info about and found out the ampersands needs to be escaped, otherwise Go interprets them as logical AND and thus all tags after `&&` are interpreted as new command and so they are not parsed as part of `--tags`:

```gherkin
[agajdosi@dhcp-24-156 minishift]$ make integration GODOG_OPTS="--tags=~coolstore\&\&~addon-xpaas"
go install -pkgdir=/home/agajdosi/golang/src/github.com/minishift/minishift/out/bindata -ldflags="-X github.com/minishift/minishift/pkg/version.minishiftVersion=1.8.0 -X github.
com/minishift/minishift/pkg/version.b2dIsoVersion=v1.2.0 -X github.com/minishift/minishift/pkg/version.centOsIsoVersion=v1.3.0 -X github.com/minishift/minishift/pkg/version.mini
kubeIsoVersion=v0.24.0 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v3.6.0 -X github.com/minishift/minishift/pkg/version.commitSha=ba5931a7" ./cmd/minishift
mkdir -p /home/agajdosi/golang/src/github.com/minishift/minishift/out/integration-test
go test -timeout 3600s github.com/minishift/minishift/test/integration --tags=integration -v -args --test-dir /home/agajdosi/golang/src/github.com/minishift/minishift/out/integr
ation-test --binary /home/agajdosi/golang/bin/minishift \
--run-before-feature="" --tags=~coolstore\&\&~addon-xpaas
TAGS: ~coolstore&&~addon-xpaas
Keeping Minishift cache directory '/home/agajdosi/golang/src/github.com/minishift/minishift/out/integration-test/cache' for test run.
```

Will investigate try it out on Windows if we can use this way.

agajdosi:
@hferentschik But I am still not sure where I should go with make targets:

> I recall an issue where we wanted to make sure that the target called by CI is a direct Makefile target and that we did not want to use custom properties within the CI script. This was to make it clear which features are run in CI jsut be looking at the Makefile.

vs. ` MINISHIFT_VM_DRIVER=kvm make integration GODOG_OPTS="--tags=~coolstore&&~addon-xpaas"` in Makefile you have provided in the PR.

Are we ok with defining the CI test run in Makefile with new `integration_ci` target, or better to define it in centos_ci.sh?

hferentschik:
> Are we ok with defining the CI test run in Makefile with new integration_pr target

+1 _integration_pr_ or _instgration_ci_.

hferentschik:
@agajdosi Why do you need the '&'. Can you not use multiple --tag for AND logic and comma seperated for _OR_ logic -  https://github.com/cucumber/cucumber/wiki/Tags#logically-anding-and-oring-tags?

hferentschik:
BTW, the quotes are a problem for Make. If we in the end need quotes as part of the Go command we need to do some escaping/magic in the Makefile.

agajdosi:
> +1 integration_pr or instgration_ci.

ack, great!

@hferentschik I am not able to find the docs for it now, but godog has AND implemented with `&&`, OR with comma as Gherkin has. It might be related to the way how Go's flag package behaves, currently for us only the last `--tags=xyz` is being parsed and used.

> BTW, the quotes are a problem for Make. If we in the end need quotes as part of the Go command we need to do some escaping/magic in the Makefile.

Agree, I would like to not use them, the escaping plan is now without need for addition of any new quotes.

Thanks for consultation, will ping once it is implemented.

agajdosi:
@hferentschik Fixed now, have tested on Fedora and Win7 (cmd+PS), the escaping of ampersands works fine, no need for adding more quotes in this PR. I have not changed the old targets with target specific variables as it works fine, just split the lines for better readability.

For documentation I have just added a mention about fact that `integration_pr` is used for PR tests - it is enough I think as it is not meant for use of developers, but more for CI.

centos_ci.sh
```diff
@@ -298,7 +298,7 @@ function build_and_test() {
   cd $GOPATH/src/github.com/minishift/minishift
   make prerelease synopsis_docs link_check_docs
   # Run integration test with 'kvm' driver
-  MINISHIFT_VM_DRIVER=kvm make integration_all
+  MINISHIFT_VM_DRIVER=kvm make integration_pr
```
praveenkumar: This sound like a plan but the problem is we still don't have a nightly setup for centos CI which means if we merge this PR then other tests will be ignored. I think we should first take care of nightly builds and then pr specific one. @minishift/minishift-dev any other suggestion?

agajdosi: @praveenkumar If we merge this it will ignore the coolstore and this feature, nothing will change in coverage. But I agree, that in next sprint it would be good to work on creating nightly.

budhrg:
LGTM :+1:

centos_ci.sh
budhrg: New target for running PR specific tests is the right direction since at some point we need to decide what tests need to be run in PR.

hferentschik:
@agajdosi 

>  have not changed the old targets with target specific variables as it works fine, just split the lines for better readability.

Can you explain to me then, how I just run the xpaas features? Either to just verify your pull request w/o having to run all the other tests or even more important to develop my own new feature. In the later case, I will be iterating for a while around a single feature file and I want to run just this feature file until I am happy with it. How do I do that.

According to the docs, I can use GODOG_OPTS and hence I try:

    $ make integration GODOG_OPTS="--paths `pwd`/test/integration/features/addon-xpaas.feature"
    go test -timeout 3600s github.com/minishift/minishift/test/integration     --tags=integration -v -args --test-dir /Users/hardy/work/    developer-platform/cdk-3/go/src/github.com/minishift/minishift/out/    integration-test --binary /Users/hardy/work/developer-platform/cdk-3/go/    bin/minishift \
    	--run-before-feature="" --tags=basic --paths /Users/hardy/work/    developer-platform/cdk-3/go/src/github.com/minishift/minishift/test/    integration/features/addon-xpaas.feature
    Log successfully started, logging into: /Users/hardy/work/    developer-platform/cdk-3/go/src/github.com/minishift/minishift/out/    test-results/integration_2017-11-14_15-02-31.log
    Running Integration test in: /Users/hardy/work/developer-platform/cdk-3/    go/src/github.com/minishift/minishift/out/integration-test
    Using binary: /Users/hardy/work/developer-platform/cdk-3/go/bin/minishift
    Feature: xpaas add-on
      Xpaas add-on imports xPaaS templates and imagestreams,
      which are then available in OpenShift to the user.
    Error closing the log: invalid argument
    
    No scenarios
    No steps
    222.138221ms
    PASS
    ok  	github.com/minishift/minishift/test/integration	0.303s    

As you can see, no tests run. The problem is that '--tags basic' is always applied and no matter what I specify in _GODOG_OPTS_, it gets ANDed to the specified tags or paths. 

Using target-specific variables I can override the '--tags basic'.

centos_ci.sh
hferentschik: It beats me why we cannot create this nightly build. This is nothing else than a new entry in https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml with a cron based trigger. @budhrg or am I missing something. Why do we need to wait for another sprint to put this in place?

agajdosi:
@hferentschik On Amit's PR which introduced the integration_all target there was an agreement (it was pulled) that `--tags=basic` and `--tags=~coolstore` will be used as it is now present. Reason was that it will make the use easier, if one for example wants to use just different `--format`. However if one specifies the `--tags` then the previous `--tags` is overwritten. In your case then your command should look like:

`make integration GODOG_OPTS="--paths `pwd`/test/integration/features/addon-xpaas.feature --tags=addon-xpaas"`

I have tested it now and it works for me (have modified the name of the file so I am sure it is not taken from original feature folder):

```gherkin
[agajdosi@dhcp-24-156 minishift]$ make integration GODOG_OPTS="--paths=/home/agajdosi/addon-xpaas.feature --tags=addon-xpaas"
go install -pkgdir=/home/agajdosi/golang/src/github.com/minishift/minishift/out/bindata -ldflags="-X github.com/minishift/minishift/pkg/version.minishiftVersion=1.8.0 -X github.
com/minishift/minishift/pkg/version.b2dIsoVersion=v1.2.0 -X github.com/minishift/minishift/pkg/version.centOsIsoVersion=v1.3.0 -X github.com/minishift/minishift/pkg/version.mini
kubeIsoVersion=v0.24.0 -X github.com/minishift/minishift/pkg/version.openshiftVersion=v3.6.0 -X github.com/minishift/minishift/pkg/version.commitSha=6df129a3" ./cmd/minishift
mkdir -p /home/agajdosi/golang/src/github.com/minishift/minishift/out/integration-test
go test -timeout 3600s github.com/minishift/minishift/test/integration --tags=integration -v -args --test-dir /home/agajdosi/golang/src/github.com/minishift/minishift/out/integr
ation-test --binary /home/agajdosi/golang/bin/minishift \
--run-before-feature="" --tags=basic --paths=/home/agajdosi/addon-xpaas.feature --tags=addon-xpaas
Keeping Minishift cache directory '/home/agajdosi/golang/src/github.com/minishift/minishift/out/integration-test/cache' for test run.
Log successfully started, logging into: /home/agajdosi/golang/src/github.com/minishift/minishift/out/test-results/integration_2017-11-14_15-35-11.log
Running Integration test in: /home/agajdosi/golang/src/github.com/minishift/minishift/out/integration-test
Using binary: /home/agajdosi/golang/bin/minishift
Feature: xpaas add-on at agajdosis home folder
  Xpaas add-on imports xPaaS templates and imagestreams,
  which are then available in OpenShift to the user.

  Scenario: User enables the xpaas add-on                   # /home/agajdosi/addon-xpaas.feature:6
    When executing "minishift addons enable xpaas" succeeds # integration_test.go:726 -> executingMinishiftCommandSucceedsOrFails
    Then stdout should contain "Add-on 'xpaas' enabled"     # integration_test.go:668 -> commandReturnShouldContain

  Scenario: User starts Minishift                          # /home/agajdosi/addon-xpaas.feature:10
    Given Minishift has state "Does Not Exist"             # integration_test.go:162 -> *Minishift.shouldHaveState
```

agajdosi:
@hferentschik Forgot to mention: I just wanted to not change what was agreed by a bigger discussion before. But I am totally fine with your proposal, it is more readable and understandable indeed. So are we going that way? I think it is not a big deal, actually, just tell me :)

hferentschik:
> On Amit's PR which introduced the integration_all target there was an agreement (it was pulled) that --tags=basic and --tags=~coolstore will be used as it is now present. Reason was that it will make the use easier, if one for example wants to use just different --format. 

But using target specific variables you can have both. You can have a reasonable default as well as the ability to slice and dice. 

> However if one specifies the --tags then the previous --tags is overwritten. In your case then your command should look like:
> make integration GODOG_OPTS="--paths `pwd`/test/integration/features/addon-xpaas.feature --tags=addon-xpaas"

And how do I know that? In this case it works because you specify the feature file as well as an additional tag. This is not intuitive. Arguably you have a similar problem when using target specific variables in the sense of that when you use it you also need to specify which tests you are running, but that's still better imo. An alternative is to split  GODOG_OPTS and also introduce GODOG_TESTS. The former is used for formatting parameters and the latter to control things like the formatter. This way one can have the best of two worlds. 

Yet another option is to also add _integration_single_ which forces you to specify GODOG_OPTS and can be used for running single features during development. 

In its current form there is a conflict/overlap between GODOG_OPTS and the --tags flag which is "hardcoded".

hferentschik:
> So are we going that way? 

That would be my choice as an immediate improvement. We could then investigate to split  GODOG_OPTS into  GODOG_OPTS and GODOG_TESTS and/or add another target. Either way, a target specific variable is more understandable imo.

agajdosi:
@hferentschik ack 👍

agajdosi:
@hferentschik Missed your above above comment, will add the target specific variable and we can see later if we want to do bigger refactor.

agajdosi:
@hferentschik Fixed now, please take a look when you have some free time :)

docs/source/contributing/developing.adoc
```diff
@@ -221,12 +221,14 @@ $ make integration
...
 NOTE: By default `make integration` only runs tests which are tagged as `@basic`.
 
-To run all the test, use the following command:
+To run all tests, use the following command:
 
 ----
 $ make integration_all
 ----
 
+NOTE: There is also a `make integration_pr` target which is being run as part of pull request testing.
```
hferentschik: I think @robin-owen tries to convert all NOTE, TIP, etc into the multi-line version.  

\[NOTE]
\====
\====

Best to align.

@robin-owen, why don't you schedule a meeting and give us a short heads up on the most common style violations we use?

agajdosi: @hferentschik Aah, missed that, will then help @robin-owen with this file. Update soon.

+1 for the sync around best practices and common mistakes we do.

hferentschik: @agajdosi - fyi pull request #1684. I guess in the end it does not matter whether you update this here. I take it that @robin-owen needs to rebase this pull request anyways an then he can take care of it then.

> +1 for the sync around best practices and common mistakes we do.

Right, at least the core devs should be aligned on this, otherwise, there will be inconsistencies in no time. 

@robin-owen I think we need two things here. First the additional section in the writing docs document which outlines some of these syles, eg multiline NOTE over NOTE:, use of '{project}' over Minishift, etc. Once this is in place we still should have a short sync were we together go through this list once. Just to make sure everyone read it. @robin-owen WDYT?

docs/source/contributing/developing.adoc
```diff
@@ -282,6 +287,12 @@ Set `no-colors` to true to disable ansi colors of Godog's output.
 Definitions::
 Set `definitions` to true to print all available step definitions.
 
+[NOTE]
```
agajdosi: @hferentschik Based on our conversation around tags I have added this note, it is connected to addition of target specific variables in this PR.

docs/source/contributing/developing.adoc
```diff
@@ -221,12 +221,17 @@ $ make integration
...
 NOTE: By default `make integration` only runs tests which are tagged as `@basic`.
 
-To run all the test, use the following command:
+To run all tests, use the following command:
 
 ----
 $ make integration_all
 ----
 
+[NOTE]
```
hferentschik: Thanks. @robin-owen will have to rebase anyways, but at least you will not forget again :-)

Issue #1613, opened 2017-10-25
Drive letters and command prompts for Windows


gbraad:
In several places of the documentation the usage on Windows is wrongly represented, such as `C:/` as the drive letter and `$` as the command prompt, or `minishift` as the binary:

eg.
https://github.com/minishift/minishift/blob/86c0f8c556db16abb0333c55a566c99a91e88619/docs/source/getting-started/installing.adoc#L70

https://github.com/minishift/minishift/blob/86c0f8c556db16abb0333c55a566c99a91e88619/docs/source/using/basic-usage.adoc#L102

https://github.com/minishift/minishift/blob/a9925e445a317f6e08a1d4d06a3cca11306fdd0a/docs/source/openshift/openshift-client-binary.adoc#L147

This needs to be consistent and suggested to change to:

  * `C:\` when talking about drives and paths
  * `C:\>` for the Command Prompt
  * `PS>` for PowerShell
  * `minishift.exe` when talked about the command

Note: only when specifically is referred to usage on Windows or in a particular shell.

gbraad:
![image](https://user-images.githubusercontent.com/1894/31980258-d8a56416-b97d-11e7-9b9d-ed4bc311bb86.png)

gbraad:
In this case explicitly usage of each shell is explained, but not clear due to the prompt:
```
For example, if you use PowerShell on Windows 7 or 10, modify the above command to:

$ minishift openshift config set --patch '{\"corsAllowedOrigins\": [\".*\"]}'
If you use Command Prompt you might need to use:

$ minishift openshift config set --patch "{\"corsAllowedOrigins\": [\".*\"]}"
```

https://github.com/minishift/minishift/blob/a9925e445a317f6e08a1d4d06a3cca11306fdd0a/docs/source/openshift/openshift-client-binary.adoc#L147

https://github.com/minishift/minishift/blob/a9925e445a317f6e08a1d4d06a3cca11306fdd0a/docs/source/openshift/openshift-client-binary.adoc#L150

gbraad:
@robin-owen WDYT?

robin-owen:
For the most part, I agree completely, and have noted some of these myself. As long as we keep these changes to areas of the documentation that are explicitly referring to usage on Windows, I think this makes perfect sense.

In the example given using `--iso-url`, though, aren't slashes (rather than Windows backslashes) appropriate for a URI?

gbraad:
[...]

PR #1670, opened 2017-11-07
Issue #1613 use consistent Windows drive letters and prompts

robin-owen:
* Use `PS>` consistently for PowerShell
* Use `C:\>` consistently for Command Prompt
* Replace slashes in Windows-specific paths with backslashes
* Refer to the Minishift binary in Windows-specific contexts as `minishift.exe`

### DIFF
--- a/docs/source/getting-started/installing.adoc
+++ b/docs/source/getting-started/installing.adoc
@@ -67,7 +67,7 @@ If you encounter issues related to the hypervisor, see the xref:../troubleshooti
 
 [NOTE]
 ====
-- On Windows operating system, due to issue link:https://github.com/minishift/minishift/issues/236[#236], you need to execute the {project} binary from your local C:/ drive.
+- On the Windows operating system, due to issue link:https://github.com/minishift/minishift/issues/236[#236], you need to execute the {project} binary from your local *_C:\_* drive.
 You cannot run {project} from a network drive.
 ====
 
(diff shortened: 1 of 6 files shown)


LalatenduMohanty:
I think I was wrong about PowerShell on Windows 7. https://blogs.technet.microsoft.com/heyscriptingguy/2011/01/07/how-do-i-install-powershell-on-windows-7-and-other-questions/. @minishift/minishift-dev Can anyone check on Windows7 that it has PowerShell.

gbraad:
I think this is clear enough. Both the situations you mention are actually written as command prompt examples, and are not different in use between this and PowerShell


```diff
 -$ minishift config set warn-check-hyperv-driver true
 +C:\> minishift.exe config set warn-check-hyperv-driver true
```

```diff
 -$ minishift start --iso-url file://c:/path/to/image.iso
 +C:\> minishift.exe start --iso-url file://c:/path/to/image.iso
```

IN all the other cases it is addressed beforehand.

In text, it might be possible to mention that they do not differ.

gbraad:
### Is PowerShell available on Windows 7?

Windows 7/Server 2008 R2 are the first Windows versions to come with PowerShell installed, by default. Windows PowerShell 2.0 needs to be installed on Windows Server 2008 and Windows Vista only. It is already installed on Windows Server 2008 R2 and Windows 7

Ref: https://blogs.technet.microsoft.com/heyscriptingguy/2011/01/07/how-do-i-install-powershell-on-windows-7-and-other-questions/, https://serverfault.com/questions/615275/in-which-version-of-windows-is-powershell-installed-by-default

budhrg:
> Can anyone check on Windows7 that it has PowerShell.

@LalatenduMohanty , Yes the one our QE use has PowerShell.

PR #1681, opened 2017-11-10
Issue #1655 Remove Docker API version info from docker-env command

### LINKED ISSUE #1655: Remove DOCKER_API_VERSION from the docker-env command output
We used to add this variable as part of service-manager (while we were still Vagrant based) to avoid errors like: 

    Error response from daemon: client is newer than server (client API version: 1.22, server API version: 1.20)

See https://github.com/projectatomic/vagrant-service-manager/issues/113

This was an issue with Docker daemon pre 1.12. See also https://github.com/moby/moby/issues/21930#issuecomment-334309028

Now that we are using a later version of Docker in the VM, there is no need for this anymore. Client and server use a negotiation algorithm now to determine the API version.

praveenkumar:
Fix #1655

### DIFF
--- a/cmd/minishift/cmd/env.go
+++ b/cmd/minishift/cmd/env.go
@@ -37,7 +37,7 @@ import (
 )
 
 const (
-	envTmpl = `{{ .Prefix }}DOCKER_TLS_VERIFY{{ .Delimiter }}{{ .DockerTLSVerify }}{{ .Suffix }}{{ .Prefix }}DOCKER_HOST{{ .Delimiter }}{{ .DockerHost }}{{ .Suffix }}{{ .Prefix }}DOCKER_CERT_PATH{{ .Delimiter }}{{ .DockerCertPath }}{{ .Suffix }}{{ .Prefix }}DOCKER_API_VERSION{{ .Delimiter }}{{ .DockerAPIVersion }}{{ .Suffix }}{{ if .NoProxyVar }}{{ .Prefix }}{{ .NoProxyVar }}{{ .Delimiter }}{{ .NoProxyValue }}{{ .Suffix }}{{end}}{{ .UsageHint }}`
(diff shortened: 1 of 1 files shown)


hferentschik:
Easy enough change :-) Did you check connecting to Docker now? Also, double check the assumptions here.

And format the code ;-)

gbraad:
No unit tests? ;-)

praveenkumar:
> Did you check connecting to Docker now? 

@hferentschik  I did docker connectivity for mac and it does works as expected with API version. Will check on Linux and windows. 

> Also, double check the assumptions here.

yes sure.

hferentschik:
I guess it always depends on the version on the client side as well. From the server side the version we are running is supporting the version negotiation. What we might have to do is to get this change tested as part of the DevSuite as well. I believe that's where the need originated in the first place. They might use some Java based Docker client. Best to double check that removing this is not going to break something on there end. Setting it seems odd, but it is not wrong either.

gbraad:
Have recently used DOCKER_HOST a lot (no api version) and this works fine.

hferentschik:
> Have recently used DOCKER_HOST a lot (no api version) and this works fine.

Sure, probably we all proceeded past Docker 1.12. Still worth checking with DevSuite, since they afaik use a custom client.

budhrg:
Updated the title of PR to not merge till we get confirmation from Devstudio team.

hferentschik:
> Updated the title of PR to not merge till we get confirmation from Devstudio team.

+1, but let's use WIP for that

hferentschik:
> till we get confirmation from Devstudio team.

Is the Devstudio team informed? Do they have a way to test this?

budhrg:
> Is the Devstudio team informed? Do they have a way to test this?

Informed last time during Cabal and got that they can test it. Will follow up tomorrow.

praveenkumar:
We got confirmation from Jean-Francois Maury on internal ML about this and it doesn't break any functionality of server adaptor. I am removing the WIP tag now.

gbraad:
can you rebase?

Issue #1568, opened 2017-10-16
Start testing Minishift with Minikube/CentOS ISO as part nightly job


agajdosi:
This is follow-up issue for https://github.com/minishift/minishift/issues/1532.

After the addition of support for Minikube iso into integration tests, we should start testing with the Minikube iso on regular basis. For beginning I suggest to start with nightly job, which will test current master with integration tests on Minikube iso. We can later progress more if results of Minikube will be good enough.

What do you think? Will be happy for any comments and ideas around this.

agajdosi:
As agreed on today sprint planning, we will start testing of Minikube iso on PR test runs, not on nightly test runs.

agajdosi:
As mentioned above, on last sprint planning it was said we should start testing minikube as part of PR test and not wait for nightly being set up. There are 2 possible ways of doing so:

- **change** the basic.feature from being run with b2d.iso to being run with minikube.iso, this will keep the execution time nearly same as it is now, we will loose the coverage of basic.feature on b2d.iso, but as most of its content is covered also in all other features, it basically should not matter
- **add** one extra run of basic.feature, which will run with minikube.iso, in this case we will need to remove some other feature to keep the execution time the same

I think the 1st plan is better, WDYT? @minishift/minishift-dev

praveenkumar:
FYI https://github.com/kubernetes/minikube/pull/2177 , we will soon have, testing from minikube side.

LalatenduMohanty:
@agajdosi 

>change the basic.feature from being run with b2d.iso to being run with minikube.iso, this will keep the execution time nearly same as it is now, we will loose the coverage of basic.feature on b2d.iso, but as most of its content is covered also in all other features, it basically should not matter

We can only go in this route when we make Minikube ISO as the default ISO. 

> add one extra run of basic.feature with minikube.iso, in this case we will need to remove some other feature to keep the execution time the same

I think we should do this for few weeks. It is a temporary arrangement so increase in execution test time should not hurt us badly. @minishift/minishift-dev  WDYT?

hferentschik:
So why can we not just create another job? One which is configured against Minikube. Keep everything in our setup as is until we officially make Minikube the default. 

Am I missing something here?

LalatenduMohanty:
> So why can we not just create another job? One which is configured against Minikube. Keep everything in our setup as is until we officially make Minikube the default.

+1

agajdosi:
[...]

PR #1683, opened 2017-11-10
[WIP] Issue #1568 Start testing minikube iso as part of PR tests

agajdosi:
Taking one commit from: Issue #1498 Add integration test for deployment of EAP applications as this PR needs that code. Will remove it once it is merged in master.

The only change relevant for this PR is in Makefile.


agajdosi:
@LalatenduMohanty Thanks for comments, closing this PR as it seems we are heading towards completely different way of solving the issue.

PR #1686, opened 2017-11-12
[WIP] Issue #1396 Reduce centos ci time

### LINKED ISSUE #1396: Reduce the time of integration test run on ci.centos.org
Currently, the CentOS CI job is taking too much time. Example https://ci.centos.org/job/minishift-pr/1983/console which is making it timeout.

One idea is to split into multi-stage jobs.

Suggestions are welcome.

budhrg:
Fix #1396


Makefile
```diff
@@ -203,6 +203,8 @@ integration: $(MINISHIFT_BINARY)
 .PHONY: integration_all
 integration_all: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
+	export MINISHIFT_ENABLE_EXPERIMENTAL=y
+	$(MINISHIFT_BINARY) config set image-caching true
```
budhrg: Not working. Debugging!

test/integration/features/basic.feature
```diff
@@ -188,6 +188,108 @@ Feature: Basic
      """
   # End of Ruby application ruby-ex deployment
 
+  # User can interact with OpenShift
+
+  Scenario: Service list sub-command
```
budhrg: Just checking the impact of merging

budhrg:
3 mins saving after combining `cmd-openshift.feature into basic.feature`.

hferentschik:
@budhrg, how is changing Cucumber tests helping in this context? What's the plan? Have we ever explored provision our own image? Installing all dependencies takes 4 to 5 minutes which are wasted.

budhrg:
> @budhrg, how is changing Cucumber tests helping in this context? 

Reduced one start of Minishift VM which probably saved ~4 min time here. 

>  Installing all dependencies takes 4 to 5 minutes which are wasted.

Don't think CentOS CI build machine offers any sort of caching. Once Ci job is triggered, it take any random machine from a pool of machines 

https://github.com/minishift/minishift-ci-jobs/blob/52904767e3c83d16d5b0183db7b66252912293ce/minishift-ci-index.yaml#L111-L112

agajdosi:
@budhrg I am fine with adding more into basic.feature, it can save us time as your results show. We could go with one big test for pull requests. It is true, that making it too long is against concepts of Cucumber, but it is better that breaking the isolation of individual features - and so it looks it is our only way. +1

But I am -1 for removing cmd-openshift.feature, we should have features aimed on individual commands. I would think of it as: we have a set of focused, well-done features (cmd-addons,addons-xpaas), which are the basis of our tests. And from them we can copy paste the hackish basic.feature (or other name) - the hacked one ad hoc feature just for saving time on PRs. We can even put it into a separate folder so it does not conflict with the "proper" feature files.

hferentschik:
>  I am fine with adding more into basic.feature, it can save us time as your results show. 

I personally would like to explore other options for speedup first. There is a reasons why it is not in the spirit of Cucumber. The longer the feature file the harder to take it in. There might be the odd things we can move into basic.feature, but as general solution it is imo not a good plan. As I pointed out, there are various other things we can explore.

> and so it looks it is our only way. +1

I don't think this is the case

> But I am -1 for removing cmd-openshift.feature, we should have features aimed on individual commands

+1

gbraad:
We can reduce time in CI by utilizing a build container which will also fix the actual software used for the build process.

hferentschik:
> We can reduce time in CI by utilizing a build container which will also fix the actual software used for the build process

Compilation in container might be viable, but not running the integration tests. Then we are down to nested virtualisation again. To then run the integration tests you need the dependencies on the slave itself.

gbraad:
@hferentschik https://github.com/minishift/minishift/issues/1396#issuecomment-344554551

Issue #1688, opened 2017-11-13
Add vet target to Makefile


hferentschik:
Avoid some common coding pitfalls by using Go's vet tool.

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/pull/1689

PR #1689, opened 2017-11-13
Issue #1688 Introducing vet target and fixing any vet violations

hferentschik:
Addresses issue #1688

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,7 +33,7 @@ ifeq ($(GOOS),windows)
 endif
 MINISHIFT_BINARY ?= $(GOPATH)/bin/minishift$(IS_EXE)
 TIMEOUT ?= 3600s
-PACKAGES := go list ./... | grep -v /vendor
+PACKAGES := go list ./... | grep -v /vendor | grep -v /out
 SOURCE_DIRS = cmd pkg test
 
 # Linker flags
@@ -143,7 +143,7 @@ $(DOCS_SYNOPISIS_DIR)/*.md: vendor $(ADDON_ASSET_FILE)
 synopsis_docs: $(DOCS_SYNOPISIS_DIR)/*.md
 
 .PHONY: prerelease
-prerelease: clean fmtcheck test cross
+prerelease: clean fmtcheck vet test cross
 	$(eval files = $(shell ./scripts/boilerplate/boilerplate.py --rootdir . --boilerplate-dir ./scripts/boilerplate | grep -v vendor))
 	@if test "$(files)" != ""; then \
 		echo "The following files don't pass the boilerplate checks:"; \
@@ -208,6 +208,10 @@ integration_all: $(MINISHIFT_BINARY)
 .PHONY: fmt
 fmt:
 	@gofmt -l -s -w $(SOURCE_DIRS)
+
+.PHONY: vet
+vet:
+	@go vet $(shell $(PACKAGES))
 
 .PHONY: fmtcheck
 fmtcheck:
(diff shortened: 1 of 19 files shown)


Makefile
```diff
@@ -209,6 +209,10 @@ integration_all: $(MINISHIFT_BINARY)
 fmt:
 	@gofmt -l -s -w $(SOURCE_DIRS)
 
+.PHONY: vet
```
praveenkumar: Great 👍

praveenkumar:
@hferentschik looks like it does need a rebase.

Makefile
```diff
@@ -33,7 +33,7 @@ ifeq ($(GOOS),windows)
 endif
 MINISHIFT_BINARY ?= $(GOPATH)/bin/minishift$(IS_EXE)
 TIMEOUT ?= 3600s
-PACKAGES := go list ./... | grep -v /vendor
+PACKAGES := go list ./... | grep -v /vendor | grep -v /out
```
LalatenduMohanty: Does this work in Windows?

LalatenduMohanty:
I think it should be two commits. 1. changes to the Makefile and the second all the changes required for the vat to pass.

hferentschik:
> I am assuming all the changes which are part of PR because of vet violation

Right

> I think it should be two commits. 1. changes to the Makefile and the second all the changes required for the vat to pass.

I disagree. I don't see a value in separating in this case. @LalatenduMohanty are you ok to merge as is?

cmd/minishift/cmd/config/util_test.go
```diff
@@ -57,10 +57,10 @@ func TestSetInt(t *testing.T) {
 	}
 	val, ok := minikubeConfig["cpus"].(int)
 	if !ok {
-		t.Fatalf("Type is not set to int")
+		t.Fatal("Type is not set to int")
```
LalatenduMohanty: So we should not use `Fatalf` because it formats the message, and we are not doing any formatting here. Is that correct?

LalatenduMohanty:
@hferentschik Also do you have the output when you ran `go vet` before fixing these issues. Putting that output here would help to get better context I think.

LalatenduMohanty:
Looks good to me. I think we can merge this once we resolve the merge conflict.

cmd/minishift/cmd/config/util_test.go
hferentschik: Correct. Usually, the 'f' versions of these type of functions do interpolation. The idea behind here is that you are using 'f', but not interpolate. You might have missed something. On the other hand, if you really just have a simple string, use the non 'f' version to make your intention clear.

hferentschik:
> looks like it does need a rebase.

It does

hferentschik:
@praveenkumar, @LalatenduMohanty  - rebased. Maybe one of you could merge once the CI build completes.

hferentschik:
@praveenkumar @LalatenduMohanty - does one of you want to merge?

Makefile
```diff
@@ -143,7 +143,7 @@ $(DOCS_SYNOPISIS_DIR)/*.md: vendor $(ADDON_ASSET_FILE)
 synopsis_docs: $(DOCS_SYNOPISIS_DIR)/*.md
 
 .PHONY: prerelease
-prerelease: clean fmtcheck test cross
+prerelease: clean fmtcheck vet test cross
```
praveenkumar: 👍

praveenkumar:
@hferentschik approved, I will let @LalatenduMohanty  to merge it.

hferentschik:
Thanks. I'll rebase my other pull requests.

Issue #1310, opened 2017-08-28
Improve test coverage in profile.feature


agajdosi:
This is a follow up issue for #126, respectively for PR #1304.

As discussed before there probably will be a basic integration test as part of the PR. But for further proper testing we will need to cover this new functionality in more detail.

gbraad:
Would it be possible for you to show/track the scenario you have envisioned to test this?

agajdosi:
@gbraad Not yet, but as soon as this will have any progress I will update it first here to consult. I am still not well familiar with these new changes.

LalatenduMohanty:
@agajdosi I have updated #126 with some use cases and will add more of them going forward. I think that should give a fair idea about what the feature is going to do .

LalatenduMohanty:
@agajdosi One of the things we should test is each profile can run with different config (i.e. different RAM, CPU, disk etc).

LalatenduMohanty:
Resolved via #1692

PR #1692, opened 2017-11-13
Issue #1310 Improve test coverage in profile.feature

amitkrout:


test/integration/features/profile.feature
```diff
@@ -2,89 +2,162 @@
...
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift start" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: User should be able to list default profile 'minishift'
+    Given Minishift has state "Running"
+     When executing "minishift profile list" succeeds
+     Then stdout should match "-\s*minishift\s*Running\s*\(Active\)"
+
+  Scenario: Getting default profile 'minishift' internal docker registry address
```
hferentschik: How is this relevant in the context of profiles?

test/integration/features/profile.feature
```diff
@@ -2,89 +2,162 @@
...
+     When executing "minishift openshift registry" succeeds
+     Then stdout should be valid IP with port number
+
+  Scenario: User should be able to set a new profile 'foo'
+    Given Minishift has state "Running"
+     When executing "minishift profile set foo" succeeds
+     Then stdout should match "Profile\s*\'foo\'\s*set\s*as\s*active\s*profile"
+
+  Scenario Outline: User should be able to set and get values on profile 'foo'
+    Given Minishift has state "Does Not Exist"
```
hferentschik: How do you get from "running state" in the previous scenario to "does not exist"? Does this test even pass?

test/integration/features/profile.feature
```diff
@@ -2,89 +2,162 @@
...
+    | cpus       | 4          | 4               |
+   
+  Scenario: User should be able to enable default add-ons and start profile 'foo'
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift addons enable anyuid" succeeds
+     Then stdout should contain
+      """
+      Add-on 'anyuid' enable
+      """
+     When executing "minishift start" succeeds
```
hferentschik: Here you start again?

test/integration/features/profile.feature
budhrg: @hferentschik , I think due to `When executing "minishift profile set foo" succeeds` from previous scenario he switched to new profile where VM doesn't exist.

test/integration/features/profile.feature
budhrg: Starting `foo` profile I believe

test/integration/features/profile.feature
amitkrout: @budhrg you are right.
@hferentschik To make the scenario more interactive we can add a good explanatory description to the scenario or can change the steps some thing like ```Profile foo has state "Does Not Exist"```

test/integration/features/profile.feature
amitkrout: @budhrg right

test/integration/features/profile.feature
amitkrout: It was there with the profile feature file is first created. I am not sure the reason behind the scenario probably it is just a check on the vm after minishift start. So i kept it here.

amitkrout:
@budhrg With the current changes i think the CI time will not increase even more than 20sec. The changes comprises some more scenario that won't take more than 30 sec and the remaining scenario with little improvement as it was there. 

@budhrg  +1 for finding proper solution for [#1396 (comment).](https://github.com/minishift/minishift/issues/1396#issuecomment-343748827)

budhrg:
@amitkrout , CI time increased by 3 mins. https://ci.centos.org/job/minishift-pr/2572/ from 45mins to 48mins.

agajdosi:
@budhrg I agree that another start of VM in current CI is not welcome. However I think this should not block us in improving the tests. Having 2 VMs running in parallel is what I see as a key feature of `profile`s and thus I think it should be in this test. But we can talk if the profile.feature should be part of CI, if we should not take some of it into basic.feature. Or if we could remove for example cmd-addon.feature of cmd-config.feature from CI.

test/integration/features/profile.feature
agajdosi: @amitkrout +1 This really needs scenario description. Or I am even thinking about implementing and using a new step definition, which would make it explicit on which profile the status check is being executed, for example:

```gherkin
Given profile "foo" of Minishift has state "Running"
```
@budhrg @hferentschik WDYT?

test/integration/features/profile.feature
```diff
@@ -2,89 +2,162 @@
...
+     Then stdout should contain
+      """
+      -- Applying addon 'anyuid':.
+      """
+      And Minishift should have state "Running"
```
agajdosi: After this step we could check the status of profile "Minishift".

test/integration/features/profile.feature
hferentschik: Just because it was there does not mean that it makes sense. We know the VM is running, it is in our pre-condition. If we want to check whether OpenShift is running we should test that and be explicit in what we want to assert. This assertion seems to be completely out of context for what we are trying to test.

test/integration/features/profile.feature
hferentschik: > To make the scenario more interactive we can add a good explanatory description to the scenario or can change the steps some thing like Profile foo has state "Does Not Exist"

+100. The whole point of Cucumber is that I am able to read the feature file and understand on a high level what you are testing. It should feel like reading a book/play script. I think we pay too little attention to this.

hferentschik:
> I agree that another start of VM in current CI is not welcome. However I think this should not block us in improving the tests. Having 2 VMs running in parallel is what I see as a key feature of profiles and thus I think it should be in this test.

+1 It's the key promise of profiles. It needs to be tested. Reducing build time cannot come before set coverage and "proper" tests. Again, if pull request builds take too long, we need to partition the tests and have multiple jobs. As test coverage will increase that's the only feasible approach. -1 for trying to avoid VM starts where it makes sense in the name of reducing build time.

hferentschik:
I feel there are so many things in there which are not relevant to what we are trying to assert. The whole  Scenario Outline for the config set. Why a scenario outline. In a feature around the config command this might make sense to test the full range of options. What's the purpose in the scope of the feature profile? The longer the test, the harder to grasp what's going on. The test should be concise to what we are trying to test/assert. 

I have a feeling there is a bit too much cut&paste going here.

amitkrout:
> I feel there are so many things in there which are not relevant to what we are trying to assert. The whole Scenario Outline for the config set. Why a scenario outline. In a feature around the config command this might make sense to test the full range of options. What's the purpose in the scope of the feature profile?

@hferentschik This is the [#1310 (comment)](https://github.com/minishift/minishift/issues/1310#issuecomment-329951990) what i am implementing here. Using Scenario Outline i can test the minimum set and get config values on profile start as per the comment.

> The longer the test, the harder to grasp what's going on. The test should be concise to what we are trying to test/assert.

@hferentschik Agree.

hferentschik:
> This is the #1310 (comment) what i am implementing here. 

Except you are not really testing it. You are calling _config set_ commands and making changes to the persistent config. That is all you are doing there. A proper test would be to prepare both VMs with custom cpu count and memory prior to starting them and then verifying the actual values for that. To do so, however, you need to read the information out of the running VM and not from the persistent config. You need to read the cpu and memory settings for example from the proc filesystem on the VM. I don't think a scenario outline is needed here.

agajdosi:
@hferentschik It is being tested, for example on L60: https://github.com/minishift/minishift/pull/1692/files#diff-db8e17a1ae346b091b162c735329b2b7R60 

@amitkrout  What I am afraid of however is how this test will work with CDK, where config is changed, there are troubles in tests... Would be nice if you could figure it out so the tests will be usable for both Minishift and CDK, if not we need to add `@minishift-only` tag.

hferentschik:
>  It is being tested, for example on L60

True, but I still feel that things are not concise. So what's the scenario? "The user can start a second Minishift instance with custom configuration settings". Hang on ...

agajdosi:
@hferentschik About that I have the same feeling.

hferentschik:
```Cucumber
@profile @command
Feature: Profile
  As a user, I can start two independent Minishift instances with isolated configuration

  Scenario: As user, I can start Minishift with default profile
      And executing "minishift start" succeeds
      And executing "minishift profile list" succeeds
     Then stdout should match "-\s*minishift\s*Running\s*\(Active\)"

  Scenario: As user, I can create a new profile 'foo' with independent settings
     Given executing "minishift profile set foo" succeeds
       And stdout should match "Profile\s*\'foo\'\s*set\s*as\s*active\s*profile"
       And executing "minishift config set memory 4096 succeeds
       And executing "minishift config set disk-size 30g succeeds
       And executing "minishift config set cpus 4 succeeds
       And executing "minishift addons enable anyuid" succeeds
       And executing "minishift start" succeeds
       And executing "minishift addons list"
      Then stdout should match "admin-user\s*: disabled\s*P\(0\)"
       And stdout should match "anyuid\s*: enabled\s*P\(0\)"
       And stdout should match "registry-route\s*: disabled\s*P\(0\)"
       And stdout should match "xpaas\s*: disabled\s*P\(0\)"
      When executing "minishift ssh -- less /proc/meminfo" succeeds
      Then stdout should match "MemTotal:\s*4[0-1][0-9]{5}\s*kB"
      When executing "minishift ssh -- sudo fdisk -l | grep Disk" succeeds
      Then stdout should match "Disk \/dev\/sda: 3[0-2]\.?[0-9]{0,2} (GB|GiB)"
      When executing "minishift ssh -- cat /proc/cpuinfo" succeeds
      Then stdout should match "processor\s*: 0"
       And stdout should match "processor\s*: 1"
       And stdout should match "processor\s*: 2"
       And stdout should match "processor\s*: 3"
       And stdout should not match "processor\s*: [4-9]"
```

Here I re-wrote the first part of the feature file. Some of the difference:

* Much fewer scenarios. There is no need to have these fine-grained scenarios which seem to hang loose. I state a scenario "As user, I can create a new profile 'foo' with independent settings" and then  I define a whole set of pre-conditions (all that is needed to get me into the state fulfilling the scenario) followed by assertions for this scenario 
* Much heavier use of preconditions 'Given'
* A more active voice in stating the scenario: "As user, I can do XYZ"

@agajdosi , @amitkrout  - You see where I am coming from?

Also, I would make more use of custom steps. I would have custom steps for:

* the VM has <num> processors
* the VM has <num>GB of RAM
* the VM has <num>GB of disk space
* the profiles <profile1>, <profile2>, <profile3> are defined
* profile <profile> is the active profile

I am not saying all these custom steps are required, but they would make things imo clearer. Especially in the cases where a command is followed by multiple lines of regexp matching, it makes imo sense ue custom steps.

hferentschik:
Using more custom steps, this becomes:

```Cucumber
@profile @command
Feature: Profile
  As a user I can start two independent Minishift instances with isolated configurtation

  Scenario: As user I can start Minishift with default profile
      And executing "minishift start" succeeds
      And executing "minishift profile list" succeeds
     Then stdout should match "-\s*minishift\s*Running\s*\(Active\)"

  Scenario: As user I can create a new profile 'foo' with independet settings
     Given executing "minishift profile set foo" succeeds
       And 'minishift' is the active profile
       And executing "minishift config set memory 4096 succeeds
       And executing "minishift config set disk-size 30g succeeds
       And executing "minishift config set cpus 4 succeeds
       And executing "minishift addons enable anyuid" succeeds
       And executing "minishift start" succeeds
      Then the VM should have 4 cpus
       And the VM should have 4GB of memory
       And the VM should have 30GB of disk space 
      When executing "minishift addons list"
      Then the add-ons anyuid should be enabled
       And the add-ons admin-user, registry-route, xpass should be disabled
```

amitkrout:
@hferentschik @agajdosi Sorry i missed the latest conversion. Will go through closely and find the abstract to implement here.

hferentschik:
Sure. As said, these are just suggestions. Not all scenarios need/should be collapsed and not every regulate expression match should be a custom step. The sweet spot is as always somewhere in the middle and there is of course also a trade-off in time. Nevertheless, I feel the scale sways strongly to one side atm.

test/integration/features/profile.feature
LalatenduMohanty: The test for registry was added because we had a bug where `minishift openshift registry` was not returning correct information for profiles.  I would suggest we keep it, but may be after we done with basic profile testing.

agajdosi:
Based on https://github.com/minishift/minishift/issues/1696 we should also test the subcommands in more depth here, we could probably rename this to `cmd-profile.feature`, similar to cmd-openshift.feature we should focus on valid names for `set` command, try to think of edge cases on all subcommands and test them.

hferentschik:
> we should also test the subcommands in more depth here, we could probably rename this to cmd-profile.feature, similar to cmd-openshift.feature we should focus on valid names for set command, try to think of edge cases on all subcommands and test them.

+1

amitkrout:
@hferentschik I have created an issue #1722 to impliment on all the feature files. I will be happy to take this in next sprint which will be started in couple of days. I think have implemented almost all the review comments, will verify it once from my end and remove the WIP tag. cc_ @agajdosi @LalatenduMohanty @budhrg

test/integration/features/profile.feature
amitkrout: Done

test/integration/features/profile.feature
```diff
@@ -2,89 +2,162 @@
...
+     When executing "minishift config set <property> "<value>"" succeeds
+     Then stdout of command "minishift config get <property>" is equal to "<expectedvalue>"
+
+  Examples: Values to be used on next Minishift start
+    | property   | value      | expectedvalue   |
+    | memory     | 4096       | 4096            |
+    | disk-size  | 30g        | 30g             |
+    | cpus       | 4          | 4               |
+   
+  Scenario: User should be able to enable default add-ons and start profile 'foo'
```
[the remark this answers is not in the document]
amitkrout: Done

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
+
+  Scenario: As user, I can check that disk-size value was applied on profile 'foo'
+    Given Profile "foo" of Minishift has state "Running"
+     When executing "minishift ssh -- sudo fdisk -l | grep Disk" succeeds
+     Then stdout should match "Disk \/dev\/sda: 3[0-2]\.?[0-9]{0,2} (GB|GiB)"
+
+  Scenario: As user, I can check that cpus value was applied on profile 'foo'
+    Given Profile "foo" of Minishift has state "Running"
+     When executing "minishift ssh -- cat /proc/cpuinfo" succeeds
+     Then stdout should match "processor\s*: 0"
```
agajdosi: Here it could be a multi line match.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
-    """
-
-Scenario: User can not delete default profile 'minishift'
-    When executing "minishift profile delete minishift --force"
-    Then exitcode should equal "1"
-     And stderr should contain
-     """
-     Default profile 'minishift' can not be deleted
-     """
+  Scenario: As user, I can list 'minishift' as the active profile
```
agajdosi: Scenarios should describe a general behaviour of product, here the behaviour/feature of minishift is not that it can list minishift as active profile, the feature is that minishift can list active profile, in context of previous scenarios this profile is named 'minishift' but that does not matter for whole product, this should be just checked in the steps.

Also this applies for a lot of scenario texts in this PR.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
+    Given Profile "minishift" of Minishift has state "Stopped"
+     When executing "minishift profile list" succeeds
+     Then stdout should match "-\s*foo\s*Running"
+      And stdout should match "-\s*minishift\s*Stopped\s*\(Active\)"
+
+  Scenario: As user, I can use profile alias command 'instance' to list 'minishift' as the active profile
+    This alias instance of profile can be applied to all the commands instead of profile.
+    Here only list is demonstrated with this command.
+    Given Profile "minishift" of Minishift has state "Stopped"
+     When executing "minishift instance list" succeeds
```
agajdosi: +1

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
+    Given Minishift has state "Stopped"
+     When executing "minishift delete --force" succeeds
+     Then Minishift should have state "Does Not Exist"
+     When executing "minishift ip"
+     Then exitcode should equal "1"
+
+  Scenario: As user, I can switch between non existing profiles
+    Given Profile "minishift" of Minishift has state "Does Not Exist"
+     When executing "minishift profile list" succeeds
+      Then stdout should match "-\s*minishift\s*Does Not Exist\s*\(Active\)"
```
agajdosi: is the indentation broken just here on GH or are some steps really off?

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
-    When executing "minishift profile list" succeeds
-    Then stdout should contain
-    """
-    - minishift	Running		(Active)
-    """
-  Scenario: Getting default profile internal docker registry address
-     Given Minishift has state "Running"
-      When executing "minishift openshift registry" succeeds
-      Then stdout should be valid IP with port number
+  Scenario: As user, I can't create profile with blank profile name
```
agajdosi: -1 for joining the two cases into one scenario.

test/integration/integration_test.go
```diff
@@ -160,6 +160,8 @@ func FeatureContext(s *godog.Suite) {
 	// steps to execute `minishift` commands
 	s.Step(`^Minishift (?:has|should have) state "(Does Not Exist|Running|Stopped)"$`,
 		minishift.shouldHaveState)
+	s.Step(`^Profile (.*) of Minishift (?:has|should have) state "(Does Not Exist|Running|Stopped)"$`,
```
agajdosi: I am not sure about the `of Minishift`, could it be:
```
Profile (.*) (?:has|should have) state "(Does Not Exist|Running|Stopped)"
```
?

Also first matching group should be in brackets I think to outline the name out of the sentence. I am wondering how is it that it works for you when in feature.file there is a step: `Given Profile "minishift" of Minishift has state "Does Not Exist"` which uses brackets.

test/integration/minishift.go
```diff
@@ -56,6 +56,15 @@ func (m *Minishift) shouldHaveState(expected string) error {
 	return nil
 }
 
+func (m *Minishift) shouldHaveProfileState(profile string, expected string) error {
```
agajdosi: looks good to me, +1

test/integration/util/util.go
```diff
@@ -161,6 +161,11 @@ func (m *MinishiftRunner) GetStatus() string {
 	return strings.Trim(cmdOut, " \n")
 }
 
+func (m *MinishiftRunner) GetProfileStatus(profileName string) string {
```
agajdosi: 👍

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
+
+  Scenario: As user, I can delete profile 'foo'
+    Given Profile "minishift" of Minishift has state "Stopped"
+     When executing "minishift profile delete foo --force" succeeds
+     Then stdout should contain
+      """
+      Profile 'foo' deleted successfully
+      """
+
+  Scenario: As user, I can't delete default profile 'minishift'
```
agajdosi: Can one delete absolutely non-existing profile? +1 for adding this scenario

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
+     Then stdout should match "admin-user\s*: disabled\s*P\(0\)"
+      And stdout should match "anyuid\s*: disabled\s*P\(0\)"
+      And stdout should match "registry-route\s*: disabled\s*P\(0\)"
+      And stdout should match "xpaas\s*: disabled\s*P\(0\)"
 
-  Scenario: Getting profile foo internal docker registry address
-     Given Minishift has state "Running"
-      When executing "minishift openshift registry" succeeds
-      Then stdout should be valid IP with port number
+  Scenario: As user, I can get default profile 'minishift' internal docker registry address
```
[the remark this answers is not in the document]
amitkrout: Much better..

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
-    """
-  Scenario: Getting default profile internal docker registry address
-     Given Minishift has state "Running"
-      When executing "minishift openshift registry" succeeds
-      Then stdout should be valid IP with port number
+  Scenario: As user, I can't create profile with blank profile name
+    Given Profile "minishift" of Minishift has state "Does Not Exist"
+     When executing "minishift profile set"
+      Then exitcode should equal "1"
+      Then stderr should contain
```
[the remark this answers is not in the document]
amitkrout: yup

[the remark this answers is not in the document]
amitkrout: Better naming convention coverage will be added once the pr #1723 is merged. For the time being i will split this.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
+      """
+      A profile name must be provided. Run `minishift profile list` for a list of existing profiles.
+      """
+     When executing "minishift profile set '  '"
+      Then exitcode should equal "1"
+      Then stderr should contain
```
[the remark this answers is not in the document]
amitkrout: Agree

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
+      """
+      A profile name must be provided. Run `minishift profile list` for a list of existing profiles.
+      """
+  
+  Scenario: As user, I can start Minishift with default profile 'minishift'
+    Given Profile "minishift" of Minishift has state "Does Not Exist"
+     When executing "minishift start" succeeds
+     Then Profile "minishift" of Minishift should have state "Running"
+      And executing "minishift profile list" succeeds
+      Then stdout should match "-\s*minishift\s*Running\s*\(Active\)"
```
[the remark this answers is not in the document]
amitkrout: Looks precise

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
-      Then stdout should be valid IP with port number
+  Scenario: As user, I can get default profile 'minishift' internal docker registry address
+    Given Profile "minishift" of Minishift has state "Running"
+     When executing "minishift openshift registry" succeeds
+     Then stdout should be valid IP with port number
 
-  Scenario: User should be able to list 'foo' and 'minishift'
-      When executing "minishift profile list" succeeds
+  Scenario: As user, I can create a new profile 'foo' with independent settings
+    Given executing "minishift profile set foo" succeeds
```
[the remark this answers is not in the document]
amitkrout: Thanks..i missed it

test/integration/features/profile.feature
```diff
@@ -2,104 +2,176 @@
...
+      And executing "minishift addons enable anyuid" succeeds
+      And executing "minishift start" succeeds
       Then stdout should contain
       """
-      - foo		Running		(Active)
-      - minishift	Running
+      -- Applying addon 'anyuid':.
       """
+      And Profile "foo" of Minishift should have state "Running"
+
```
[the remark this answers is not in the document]
amitkrout: Yes..

test/integration/features/profile.feature
amitkrout: Nice to have this scenario

amitkrout:
#1748

test/integration/integration_test.go
amitkrout: > I am not sure about the ```of Minishift```, could it be:
```Profile (.*) (?:has|should have) state "(Does Not Exist|Running|Stopped)"```
?

Not bad. Profile itself describe it as the instance. Would be fine with  ```Profile (.*) (?:has|should have) state "(Does Not Exist|Running|Stopped)"```. 

> Also first matching group should be in brackets I think to outline the name out of the sentence. I am wondering how is it that it works for you when in feature.file there is a step: Given Profile "minishift" of Minishift has state "Does Not Exist" which uses brackets

Not sure what you mean here

**Comment by amitkrout** [comment]
> please also pay attention if this can be run on with CDK.

This need really some hacks. Probably will get some refinement and sync on feature file after checking with downstream. Honestly till now i don't check this with downstream bits. AFAIU there would be addons related  scenario on profile need ```@minishift-only``` tag. I have some solution like enabling addons for example ```registry-route``` on profile which will be satisfied for both upstream and downstream. 

I will keep the pr status as [WIP] as it required some more refinement. Will check it on monday

**Comment by amitkrout** [comment]
Done with the change. Due to the on going issue #1748, i can not add the specific test scenario to the feature file. That can be added later to the feature file once the issue is fixed. Hence removing [WIP] tag.

File: test/integration/features/profile.feature [hunk 21]
```diff
@@ -2,104 +2,251 @@
...
+    | '20'        |
+    | 'random001z'|
+    | '00_XYZ-'   |
+    | 'foo'       |
 
-  Scenario: Getting profile foo internal docker registry address
-     Given Minishift has state "Running"
-      When executing "minishift openshift registry" succeeds
-      Then stdout should be valid IP with port number
+  Scenario: As user, I can list all profile
```
hferentschik: This should be part of the previous scenario since it really depends on it.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+       And executing "minishift profile list" succeeds
+       And stdout should match "-\s*minishift\s*Running\s*\(Active\)"
 
-Scenario: Deleting Minishift
-   Given Minishift has state "Running"
-    When executing "minishift delete --force" succeeds
-    Then Minishift should have state "Does Not Exist"
-    When executing "minishift ip"
-    Then exitcode should equal "1"
+  Scenario: As user, I can get internal docker registry address of default profile 'minishift'
```
hferentschik: This really does not belong here. Why is the internal docker registry specific to the profile feature?

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
-    When executing "minishift profile set abc" succeeds
-      Then stdout should contain
-      """
-      Profile 'abc' set as active profile.
+  Scenario: As user, I can apply independent settings in profile 'foo'
+    Given Profile "foo" has state "Does Not Exist"
+     And executing "minishift profile set foo" succeeds
+     Then stdout should match "Profile\s*\'foo\'\s*set\s*as\s*active\s*profile"
+      When executing "minishift config set memory 4096" succeeds
+       And executing "minishift config set disk-size 30g" succeeds
```
hferentschik: Where is the assertion that this was persisted in the config? Either via the `config view` command or maybe even better by asserting against the JSON file. We so this already somewhere else. You are not really verifying your scenario here.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+    Given Profile "foo" has state "Running"
+     And executing "minishift profile set foo" succeeds
+     Then stdout should match "Profile\s*\'foo\'\s*set\s*as\s*active\s*profile"
+     When executing "minishift addons list"
+     Then stdout should match "admin-user\s*: disabled\s*P\(0\)"
+      And stdout should match "anyuid\s*: disabled\s*P\(0\)"
+      And stdout should match "registry-route\s*: enabled\s*P\(0\)"
+      And stdout should match "xpaas\s*: disabled\s*P\(0\)"
+
+  Scenario: Checking that memory value was applied on profile 'foo'
```
hferentschik: Why different scenarios for checking things like memory and memory? There is so much unnecessary repetition imo

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+    Given Profile "foo" has state "Running"
+     And executing "minishift profile set foo" succeeds
```
hferentschik: So why do you execute this all the time?

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+
+  Scenario: As user, I can use profile alias command 'instance' to list 'minishift' as the active profile
+    This alias instance of profile can be applied to all the commands instead of profile.
+    Here only list is demonstrated with this command.
+    When executing "minishift instance list" succeeds
+    #Only validating the running and stopped profile
+    Then stdout should match "-\s*foo\s*Running\s*\(Active\)"
+     And stdout should match "-\s*minishift\s*Stopped"
+
+  Scenario: As user, I can use profile alias command 'profiles' to list 'minishift' as the active profile
```
hferentschik: What's the difference to the above?

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+  Scenario: As user, I can not run delete command on non existing profile
+    When executing "minishift profile delete XYZ --force"
      Then exitcode should equal "1"
-     Then stderr should contain
+      And stderr should contain
+      """
+      Error: 'XYZ' is not a valid profile
+      """
+
+  Scenario: As user, I can't delete default profile 'minishift'
```
hferentschik: "cannot" is better than "can't" (applies across the whole file)

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+    And stdout should match "-\s*minishift\s*Stopped\s*\(Active\)"
+    Then executing "minishift profile delete minishift --force"
+     And exitcode should equal "1"
+     And stderr should contain
      """
-     A profile name must be provided. Run `minishift profile list` for a list of existing profiles.
+     Default profile 'minishift' can not be deleted
      """
 
+  Scenario: As user, I can delete Minishift vm
```
hferentschik: VM

test/integration/features/profile.feature
amitkrout: Awesome..Scenario would be like
Scenario Outline: As user, I can create profile with alphanumeric character including '_' and '-' in profile name
[...]
And executing "minishift profile list" succeeds
Then stdout should match "-\s*<profilename>\s*Does Not Exist\s*\(Active\)"

Will check the implimentation

test/integration/features/profile.feature
amitkrout: Yup..this should be moved. Better place for this scenario would be, if we add it to issue like #1318 where the registry validation will be done.

test/integration/features/profile.feature
amitkrout: Each scenario should be satisfied the precondition and less depend on the previous scenario. 
For example let foo remains an active profile at the end of the scenario execution. For the very next scenario as a user  i am not aware of the current active profile though the profile context is transferred form the previous step. So to make the steps more descriptive each time i check this as a precondition when required.

test/integration/features/profile.feature
amitkrout: Each scenario should be satisfied the precondition and less depend on the previous scenario. 
For example let foo remains an active profile at the end of the scenario execution. For the very next scenario as a user  i am not aware of the current active profile though the profile context is transferred form the previous step. So to make the steps more descriptive each time i check this as a precondition when required.

test/integration/features/profile.feature
amitkrout: This is just a check whether profile alias command ```profiles``` and ```instance``` atleast works on list execution.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
 Feature: Profile
   As a user I can perform basic operations of Minishift with profile feature
 
-  Scenario: Starting Minishift with default profile
-     Given Minishift has state "Does Not Exist"
-      When executing "minishift start" succeeds
-      Then Minishift should have state "Running"
+  Scenario Outline: As user, I can't create profile with blank profile name
+    Given Profile "minishift" has state "Does Not Exist"
```
agajdosi: Profile here should be without a capital letter.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+    | '&foo'      |
+    | '*te$st^'   |
+    | 'foo 123'   |
+    | 'foo    '   |
 
-Scenario: Starting Minishift with profile foo
-    When executing "minishift start --profile foo" succeeds
-    Then Minishift should have state "Running"
+  Scenario Outline: As user, I can create profile with alphanumeric character including '_' and '-' in profile name
+    NOTE: Though single character like a-z, A-Z or single digit between 0-9 are consider as alphanumeric, will not be supported here.
```
agajdosi: The single alphanumeric character should be then tested in previous scenario outline.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
       """
-
-  Scenario: User should be able set 'minishift' as the active profile
-      When executing "minishift profile set minishift" succeeds
+     When executing "minishift profile set minishift" succeeds
       Then stdout should contain
       """
       Profile 'minishift' set as active profile
-      """
+      """
```
agajdosi: extra whitespaces

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+     Then stdout should match "Profile\s*\'foo\'\s*set\s*as\s*active\s*profile"
+      When executing "minishift config set memory 4096" succeeds
+       And executing "minishift config set disk-size 30g" succeeds
+       And executing "minishift config set cpus 4" succeeds
+       And executing "minishift addons enable registry-route" succeeds
+         
+  @minishift-only
+  Scenario: As user, I can check the config settings and add-ons list that should not be changed in profile 'minishift' 
+    Given Profile "foo" has state "Does Not Exist"
+     And executing "minishift profile set foo" succeeds
```
agajdosi: `Scenario: As user, I can check the config settings and add-ons list that should not be changed in profile 'minishift' ` but you are not really checking the `minishift` profile, but the `foo` profile instead.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+    Given Profile "foo" has state "Running"
+     And executing "minishift profile set foo" succeeds
+     Then stdout should match "Profile\s*\'foo\'\s*set\s*as\s*active\s*profile"
+     When executing "minishift ssh -- sudo fdisk -l | grep Disk" succeeds
+      Then stdout should match "Disk \/dev\/sda: 3[0-2]\.?[0-9]{0,2} (GB|GiB)"
+
+  Scenario: Checking that cpus value was applied on profile 'foo'
+    Given Profile "foo" has state "Running"
+     And executing "minishift profile set foo" succeeds
+      Then stdout should match "Profile\s*\'foo\'\s*set\s*as\s*active\s*profile"
```
agajdosi: `Given, And, Then, When` is not right. Should be `Given, And, And...`, `When, And, And...`, `Then, And, And...`. We might change the matching groups if the step sounds bad with `Given` statement.

test/integration/features/profile.feature
agajdosi: I would suggest to bring this to the beginning of feature. No need to have it in the middle of the file where it breaks the logical flow, which is quite hard to follow in this feature.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+
+  Scenario: As user, I can use profile alias command 'profiles' to list 'minishift' as the active profile
+    This alias 'profiles' of profile can be applied to all the commands instead of profile.
+    Here only list is demonstrated with this command.
+    When executing "minishift profiles list" succeeds
+    #Only validating the running and stopped profile
+    Then stdout should match "-\s*foo\s*Running\s*\(Active\)"
+     And stdout should match "-\s*minishift\s*Stopped"
+
+  Scenario Outline: As user, I can delete all created profile
```
agajdosi: profiles

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+    NOTE: Though single character like a-z, A-Z or single digit between 0-9 are consider as alphanumeric, will not be supported here.
+    See https://github.com/minishift/minishift/issues/1748 for more details.
+    Given Profile "minishift" has state "Does Not Exist"
+     When executing "minishift profile set <profilename>" succeeds
+     Then stdout should contain
+      """
+      Profile <profilename> set as active profile.
+      """
+  
+  Examples: Good profile name
```
agajdosi: correct profile names

test/integration/features/profile.feature
```diff
@@ -2,104 +2,251 @@
...
+    
+  Scenario Outline: As user, I can't create profile with special character in profile name
+    Given Profile "minishift" has state "Does Not Exist"
+     When executing "minishift profile set <profilename>" fails
+     Then stderr should contain
+      """
+      Profile names must consist of alphanumeric characters only.
+      """
+  
+  Examples: Bad profile name
```
agajdosi: wrong profile names

test/integration/features/profile.feature
agajdosi: I think we can have this as part of the `Scenario Outline: As user, I can create profile with alphanumeric character including '_' and '-' in profile name.`

test/integration/integration_test.go
```diff
@@ -160,6 +160,8 @@ func FeatureContext(s *godog.Suite) {
 	// steps to execute `minishift` commands
 	s.Step(`^Minishift (?:has|should have) state "(Does Not Exist|Running|Stopped)"$`,
 		minishift.shouldHaveState)
+	s.Step(`^Profile (.*) (?:has|should have) state "(Does Not Exist|Running|Stopped)"$`,
```
agajdosi: There is no need to have Profile with capital letter, it is not a name as Minishift is. All other step definitions starts with minuscule.

amitkrout: True

test/integration/features/profile.feature
amitkrout: Done. Then we need to modify it once the pr #1749 is merged.

test/integration/features/profile.feature
amitkrout: There is some missing text in the scenario description imo. 

This should be something like 
```
Scenario: As user, I can check the config settings and add-ons list that should not be changed in profile 'minishift' from active profile 'foo'
```

test/integration/features/profile.feature
amitkrout: Will remove it

test/integration/features/profile.feature
amitkrout: Done. Then we need to modify it once the pr #1749 is merged.

test/integration/features/profile.feature
amitkrout: However in this case this should not be allowed as this is not a feature of profile. It just a check on config applied on running VM. Moreover this is completely depends on the previous scenario, adding this as a scenario is to avoid the step skip if any failure occurs on any config applied.

test/integration/features/profile.feature
amitkrout: This is what we are checking the config is applied after the VM (foo) is start. The idea behind keeping it in scenario is to avoid the step skip if any failure occur.

test/integration/features/profile.feature
amitkrout: Currently i am working on the issue #1722 as per the suggestion made on previous comment(https://github.com/minishift/minishift/pull/1692#issuecomment-344227795).

amitkrout: For validation ```config get``` is applied to the property name (memory, disk-size and cpus) in the same scenario. Changes are pushed to the pr.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,229 @@
...
+
+  Scenario Outline: As user, I can delete all created profiles
+    When executing "minishift profile delete <profilename> --force" succeeds
+     Then stdout should contain
      """
-    When executing "minishift profile set '  '"
+     Profile '<profilename>' deleted successfully
+     """
+  
+  Examples: Profile name
```
agajdosi: This could be shortened I think, there is some repetition, for example `00_XYZ-` covers quite a lot. With `00_xYz-` it is even better. I would also suggest to use something like that, which is more complicated, instead of `foo` in all other cases. We can use something which looks like somebody is actually using, for example`small_VM-2` or `python_VM-1`.

What I like here is `foo-`, `profile_`, I am curious how is it with `_profile` or `-profile`.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,229 @@
...
-    Then stdout should contain
-    """
-    - minishift	Does Not Exist	(Active)
-    """
-    When executing "minishift profile set abc" succeeds
-      Then stdout should contain
-      """
-      Profile 'abc' set as active profile.
+  @minishift-only
+  Scenario: As user, I can check the config settings and add-ons list that should not be changed in profile 'minishift' from active profile 'foo'
```
agajdosi: If we sacrifice the memory check, which is not really crucial, and move the addon check to the individual scenario, then this scenario could be without `@minishift-only`.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,229 @@
...
-    - minishift	Does Not Exist	(Active)
-    """
-    When executing "minishift profile set abc" succeeds
-      Then stdout should contain
-      """
-      Profile 'abc' set as active profile.
+  @minishift-only
+  Scenario: As user, I can check the config settings and add-ons list that should not be changed in profile 'minishift' from active profile 'foo' 
+    Given profile "foo" has state "Does Not Exist"
+     And executing "minishift profile set foo" succeeds
```
agajdosi: But... but wait. This scenario should check the setting on default `minishift` profile, should not?

test/integration/features/profile.feature
```diff
@@ -2,104 +2,229 @@
...
+     """
+     Stopping local OpenShift cluster...
+     Cluster stopped.
      """
-     A profile name must be provided. Run `minishift profile list` for a list of existing profiles.
+     And executing "minishift --profile minishift status" succeeds
+      Then stdout should match "Minishift:\s*Stopped\nProfile:\s*minishift\nOpenShift:\s*Stopped"
+
+  Scenario Outline: As user, I can delete all created profiles
+    When executing "minishift profile delete <profilename> --force" succeeds
```
agajdosi: Here it is quite hidden that you are deleting also the VM with the deletion of the profile. This really needs to go to a separate scenario. It is big difference when one is deleting profiles with not existing VM and profiles with actually running VM.

agajdosi:
Also please take a look on @hferentschik's comments, suggestions are still not implemented.
https://github.com/minishift/minishift/pull/1692#issuecomment-344226675

test/integration/features/profile.feature
amitkrout: Nothing is hidden here. AFAIU whatever the profile is created of status ```Does Not Exist|Running|Stopped``` it really does not mater what is happening while deleting as we are not capturing any process in the background. The only validation ```Profile '<profilename>' deleted successfully``` is done in ```stdout``` which would be same for all the profile.
```
dhcp-0-120:v1.9.0 amit$ ./minishift profile list
- asdf		Running
- minishift	Running
- wer		Does Not Exist	(Active)
```
```
dhcp-0-120:v1.9.0 amit$ ./minishift profile delete wer --force
VM for profile 'wer' does not exist
Deleted:  /Users/amit/.minishift/profiles/wer
Profile 'wer' deleted successfully
Switching to default profile 'minishift' as the active profile.
```
```
dhcp-0-120:v1.9.0 amit$ ./minishift profile delete asdf --force
Deleted:  /Users/amit/.minishift/profiles/asdf
Profile 'asdf' deleted successfully
```

test/integration/features/profile.feature
```diff
@@ -2,104 +2,229 @@
...
+    When executing "minishift ssh -- cat /proc/cpuinfo" succeeds
+     Then stdout should match "processor\s*: 0"
+      And stdout should match "processor\s*: 1"
+      And stdout should match "processor\s*: 2"
+      And stdout should match "processor\s*: 3"
+      And stdout should not match "processor\s*: [4-9]"
+
+  Scenario: As user, I can execute command on non active running profile 'minishift' from active profile 'foo'
+    Given profile "foo" has state "Running"
+     And executing "minishift profile set foo" succeeds
```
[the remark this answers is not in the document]
amitkrout: Nice to have it. Will check the implementation

test/integration/features/profile.feature
```diff
@@ -2,104 +2,229 @@
...
+     Then stdout should match "MemTotal:\s*4[0-1][0-9]{5}\s*kB"
+
+  Scenario: Checking that disk-size value was applied on profile 'foo'
+    When executing "minishift ssh -- sudo fdisk -l | grep Disk" succeeds
+     Then stdout should match "Disk \/dev\/sda: 3[0-2]\.?[0-9]{0,2} (GB|GiB)"
+
+  Scenario: Checking that cpus value was applied on profile 'foo'
+    When executing "minishift ssh -- cat /proc/cpuinfo" succeeds
+     Then stdout should match "processor\s*: 0"
+      And stdout should match "processor\s*: 1"
```
gbraad: And stdout should match "processor\s*: [1-4]"

test/integration/features/profile.feature
amitkrout: That is what being checked here. User can check config setting irrespective of the  profile is currently being used. for example from active profile foo i can see config setting of default profile minishift by using the command ```minishift --profile minishift config get cpus```

test/integration/features/profile.feature
amitkrout: -profile won't be a valid usecase. ```minishift profile``` interprets this as flag. We can take some better permutation/combination using the separator ```_``` and ```-```. For more details i would suggest to have a look on #1749

amitkrout:
> Also please take a look on @hferentschik's comments, suggestions are still not implemented.
#1692 (comment)

@agajdosi I have already created a issue #1722. Will be implemented later

amitkrout:
> Indentation is still bad.

@agajdosi Which Indentation format you are following. Do we have any cucumber support doc on indentation? please share for reference

test/integration/features/profile.feature
agajdosi: @amitkrout sure, but can that be passed when quotes are used? that should be possible AFAIU. But the discussion is ongoing on this: https://github.com/minishift/minishift/pull/1749#issuecomment-347812553

test/integration/features/profile.feature
agajdosi: @amitkrout I think these are two different cases:

- deleting profile which has not any VM Running or Stopped
- deleting profile which has VM Running or Stopped

In second case the Minishift has somehow first delete the VM, then has to delete the profile. Would be nice to keep these scenarios separately I think, wdyt?

agajdosi: However it is true that the validation will be the same. But the 2 scenarios will describe the situation more clearly I think and in case of any failure on one of them, it would be clear that their difference somehow matters.

test/integration/features/profile.feature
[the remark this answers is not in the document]
amitkrout: There is no direct command as of now to know the profile is a active one. It comes with ```set``` command and become the default active profile. So steps like ```”foo” should be a active profile``` need some nice logic of course. 

The only way i am seeing ahead is list out all the profiles then run some nice string operation on list for validation. @agajdosi @gbraad  Please suggest if you have any better resolution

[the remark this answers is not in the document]
agajdosi: @amitkrout I think your suggestion is correct, I do not see any shortcut or better solution here.

test/integration/features/profile.feature
amitkrout: I spend some time to know what really is happening. I took some sample of Active/running/does not exist profiles then i ran various scenarios on profile delete subcommand. 
```
Case 1: 
Deleting profile having no VM created from a active profile
$ ./minishift profile delete pro --force 
Profile 'pro' deleted successfully.
```
```
Case 2: 
Deleting profile having no VM created from a active profile having VM up and running
$ ./minishift profile delete pro1 --force 
Profile 'pro' deleted successfully.
```
```
Case 3: 
Deleting the active profile having no VM
$ ./minishift profile delete pro3 --force 
Profile 'pro1' deleted successfully.
Switching to default profile 'minishift' as the active profile.
```
```
Case 3: 
Deleting the profile from default active profile minishift (Both the profile has VM up and running)
$ ./minishift profile delete pro4 --force 
Profile ‘pro4’ deleted successfully.
```
In all the above cases ```Profile <profilename> deleted successfully``` is common printed output and no other output except ```Switching to default profile 'minishift' as the active profile.``` like case 2

So no need to split the scenario. Then by adding a precondition to the very next step such as ```Given profile ”minishift” is an active``` will resolve the concern.

NOTE : I executed the cases on master

test/integration/features/profile.feature
amitkrout: > but can that be passed when quotes are used? that should be possible AFAIU

Just saw the question. No it should not work. Interpreter still consider this as a flag

amitkrout:
All comments are addressed cc_ @agajdosi.

gbraad:
@minishift/minishift-dev please review or remove stale comments/status

test/integration/features/profile.feature
```diff
@@ -2,104 +2,211 @@
...
+
+  Scenario: As user, I can start Minishift with default profile 'minishift'
+    Given profile "minishift" has state "Does Not Exist"
+    And profile "minishift" is an active profile
+     When executing "minishift start" succeeds
+     Then profile "minishift" should have state "Running"
+     And executing "minishift profile list" succeeds
+     Then stdout should match "-\s*minishift\s*Running\s*\(Active\)"
+
+  Scenario: As user, I can apply independent settings in profile 'foo'
```
hferentschik: So in this case, given your assertions, where is the proof that the settings are independent? You make some changes to the foo profile and assert that they apply. But to show that these settings are independent you need to also assert that the settings for another profile are unchanged. 

So  a proper test would look imo like:

* Assert that the memory of profile minishift is <n>GB
* Start a new profile 
* Set a new memory size for the new profle
* Assert that the new profile has the specified memory setting
* Assert that the minishift profile still has a memory setting of <n> GB

If you skip the whole business around disk-size, cpus, etc you end up with a test which is about the same length, but actually verifies what you want to show. There is not really much value in setting 4 different config values. What you want to show can be shown with a single one.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,211 @@
...
+     And stdout should match "4096"
+     Then executing "minishift config get disk-size" succeeds
+     And stdout should match "30g"
+     Then executing "minishift config get cpus" succeeds
+     And stdout should match "4"
+     Then executing "minishift addons list"
+     And stdout should match "registry-route\s*: enabled\s*P\(0\)"
+
+  @minishift-only
+  Scenario: As user, I can check the config settings and add-ons list that should not be changed in profile 'minishift' from active profile 'foo'
```
hferentschik: What is the purpose of this scenario. If the previous scenario would be written in a better way, then this would be pretty much unnecessary. Also, why introducing something Minishift specific here. There is really no need for that. We should use  @minishift-only sparingly.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,211 @@
...
+     When executing "minishift --profile minishift config get cpus" succeeds
+     Then stdout should match "<nil>"
+     When executing "minishift --profile minishift addons list"
+     Then stdout should match "admin-user\s*: disabled\s*P\(0\)"
+     And stdout should match "anyuid\s*: disabled\s*P\(0\)"
+     And stdout should match "registry-route\s*: disabled\s*P\(0\)"
+     And stdout should match "xpaas\s*: disabled\s*P\(0\)"
+
+  Scenario: As a user, I can start profile 'foo'
+    Given profile "foo" has state "Does Not Exist"
```
hferentschik: That's really not important.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,211 @@
...
+    And profile "foo" is an active profile
+     When executing "minishift start" succeeds
+     Then stdout should contain
      """
-     Default profile 'minishift' can not be deleted
+     Add-on 'registry-route' created docker-registry route.
      """
+     And profile "foo" should have state "Running"
+  
+  @minishift-only
```
hferentschik: Do we need to be Minishift specific here. Is the exact match of add-ons important here? Can we not change the regexp in a way that it matches CDK and Minishift. In the add-on tests I would expect that we use @minishift-only, since there we want to assert something which differs. In this feature file we are interested in the profile feature. No need to use @minishift-only

test/integration/features/profile.feature
```diff
@@ -2,104 +2,211 @@
...
+  Scenario: As user, I can list add-ons in profile 'foo' 
+    Given profile "foo" has state "Running"
+    And profile "foo" is an active profile
+     When executing "minishift addons list"
+     Then stdout should match "admin-user\s*: disabled\s*P\(0\)"
+     And stdout should match "anyuid\s*: disabled\s*P\(0\)"
+     And stdout should match "registry-route\s*: enabled\s*P\(0\)"
+     And stdout should match "xpaas\s*: disabled\s*P\(0\)"
+
+  Scenario: Checking that memory value was applied on profile 'foo'
```
hferentschik: I am still not a fan of this, but if you think this makes sense, so bit it.
My concerns are:

* Scenarios should be user-driven. This is not the case here
* If nothing else these three scenarios could be rolled into one

amitkrout:
@hferentschik Thanks for the nice explanatory review comments. I merged and removed some unnecessary config checks/scenario which are not specific to this feature. Please comment if things are not good yet. cc_ @minishift/minishift-dev

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
-    When executing "minishift profile list" succeeds
-    Then stdout should contain
-    """
-    - minishift	Running		(Active)
-    """
-  Scenario: Getting default profile internal docker registry address
-     Given Minishift has state "Running"
-      When executing "minishift openshift registry" succeeds
-      Then stdout should be valid IP with port number
+  Scenario Outline: As user, I cannot create profile with blank profile name
```
hferentschik: nitpick - "As user, I cannot create a profile..."

Same applies to the following couple of scenarios.

test/integration/minishift.go
hferentschik: I think it is more "profileShouldHaveState" or maybe even better "ProfileShouldBeInState".

test/integration/minishift.go
```diff
@@ -56,6 +56,28 @@ func (m *Minishift) shouldHaveState(expected string) error {
...
+	actual := m.runner.GetProfileStatus(profile)
+	if !strings.Contains(actual, expected) {
+		return fmt.Errorf("Profile %s of Minishift state did not match. Expected: %s, Actual: %s", profile, expected, actual)
+	}
+
+	return nil
+}
+
+func (m *Minishift) isAnActiveProfile(profileName string) error {
```
hferentschik: "isAnActiveProfile" -> "isTheActiveProfile"

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
+     When executing "minishift --profile minishift config get memory" succeeds
+     Then stdout should match "<nil>"
+     And executing "minishift --profile minishift addons list" succeeds
+     And stdout should match "registry-route\s*: disabled\s*P\(0\)"
+     And executing "minishift addons enable registry-route" succeeds
+     And executing "minishift config set memory 5120" succeeds
+     And executing "minishift config get memory" succeeds
+     And stdout should match "5120"
+     When executing "minishift start" succeeds
+     Then stdout should contain
```
LalatenduMohanty: This test is little confusing to me. In the scenario we are saying that it is about `apply independent settings in profile 'foo'` but in the test we are applying settings to both `minishift` and `foo` and we are verifying the settings of memory for profile `foo`. Do we really need to set config for `minishift`.

amitkrout: yes, we need to set the atleast some config setting in profile ```foo``` to verify that it should not be reflected in default profile ```minishift```.

> but in the test we are applying settings to both minishift and foo

No. We are applying setting only for foo and verifying with the default setting of minishift.

test/integration/minishift.go
amitkrout: Yup, second one is even better however i feel “profileShouldHaveState” is more align with function like ```shouldHaveState```. Also in feature context I find the similar kind of alignment in naming of function. So I think It would be great to keep ```profileShouldHaveState```

test/integration/minishift.go
amitkrout: Ahh..I missed it in last review comment. Will update it

test/integration/features/profile.feature
amitkrout: Hmm :) I took sometime to get a better scenario description infact I got of some them however non of them were convincing

amitkrout:
@LalatenduMohanty Ahh..that is why i don't see my user name in contribution list. Not sure why ```amit@dhcp35-153.lab.eng.blr.redhat.com``` is being picked 
```
 $ git config --list
user.name=Amit Rout
user.email=arout@redhat.com
[...]
$
```
will fix it

amitkrout:
> will fix it

@LalatenduMohanty Done

test/integration/features/profile.feature
hferentschik: The descriptions are good now. Reads quite well now imo. Well done. This really just a typo.

test/integration/features/profile.feature
hferentschik: @LalatenduMohanty, you need to do something. The scenario is that you can do something *independently*. If you just modify settings for profile 'foo' and assert that they apply how can you make sure that these changes are not applied another existing profile.  In this case you are not asserting your actual scenario. As @amitkrout is saying you need to make sure that something else (eg the default mininishift profile) stays untouched.

test/integration/minishift.go
hferentschik: >  So I think It would be great to keep profileShouldHaveState

I am fine with that, but this is not what you have. You have 'shouldHaveProfileState'.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
+     And executing "minishift ssh -- less /proc/meminfo" succeeds
+     And stdout should match "MemTotal:\s*5[0-1][0-9]{5}\s*kB"
 
-Scenario: Deleting Minishift
-   Given Minishift has state "Running"
-    When executing "minishift delete --force" succeeds
-    Then Minishift should have state "Does Not Exist"
-    When executing "minishift ip"
-    Then exitcode should equal "1"
+  Scenario: As user, I can execute command on non active running profile 'minishift' from active profile 'foo'
```
hferentschik: "As user, I can execute command on non active running profile 'minishift' from active profile 'foo'" -> "As user, I can execute a command against a non active profile"

test/integration/minishift.go
hferentschik: This really reads as "profile state" which could conjure the idea of a new type of state specific to a profile. What we are after is the state of the Minishift VM for a given profile.

test/integration/features/profile.feature
amitkrout: Short and precise

budhrg:
Need consistent alignment of `Given`, `When`, and `Then`

screenshot from 2017-12-07 14-58-49

See expected alignment https://github.com/minishift/minishift/blob/master/test/integration/features/basic.feature#L37-L78

amitkrout:
@budhrg Much better...Done

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
-      """
+  Scenario: As user, I can switch between existing profiles
+    When executing "minishift profile set Test-123" succeeds
+    Then profile "Test-123" should be the active profile
+    When executing "minishift profile set minishift" succeeds
+    Then profile "minishift" should be the active profile
+
+  Scenario: As user, I can start Minishift with default profile 'minishift'
+    Given profile "minishift" has state "Does Not Exist"
+    And profile "minishift" is the active profile
```
budhrg: @amitkrout , Still missing this alignment :)

amitkrout: @budhrg It should act as a single logic as both the lines are precondition. IMO it should fall under same indentation 
```
Given - precondition
And - precondition
 When - Statement
```
looks more reasonable than 
```
Given - precondition
 And - precondition
 When - statement
```

I am not in the favor of alignment you are suggesting here

amitkrout:
Seems the pr is ready for merge cc_ @minishift/minishift-dev

test/integration/features/profile.feature
hferentschik: TBH, I don't really fancy it too much either. However, we do have the same alignment across all feature files. I would open for a discussion to change this, but until then the sensible thing to do is to stick with the alignment we have.

hferentschik: FYI, Using the Cucumber feature file formatter in the keywords are left aligned. We have right alignment. I prefer the former, but as said, the latter is what is chosen and should be used until we change it.


This is what we use:
```Cucumber
Scenario: As user, I can start Minishift with default profile 'minishift'
   Given profile "minishift" has state "Does Not Exist"
     And profile "minishift" is the active profile
```

This would be the alternative:
This is what we use:
```Cucumber
Scenario: As user, I can start Minishift with default profile 'minishift'
   Given profile "minishift" has state "Does Not Exist"
   And   profile "minishift" is the active profile
```

agajdosi: Influential `The Cucumber Book` uses aligning to left:
```gherkin
Scenario: Successful sign up
New users should get a confirmation email and be greeted
personally by the site once signed in.
  Given I have chosen to sign up
  When I sign up with valid details
  Then I should receive a confirmation email
  And I should see a personalized greeting message
```
We can switch to left. But really this is not a debate for this issue, if somebody wants to switch, we should create a separate issue. In all cases we need to be consistent.

Left or right, but IMHO there is no need to add a meaning to the indentation as the block logic is somehow part of the Gherkin keywords. Also this is the first time I actually realized what's going on with these not aligned steps, before I was just thinking it is chaotic rest from the fast-writing of the feature. Also I do not recall any example using this around Gherkin community.

agajdosi: Created #1815 follow-up issue for the formatting.

amitkrout: @agajdosi Thanks for creating the issue. 
@budhrg Updated the ```Given And``` indentation suggestion atm. Will update all the feature file once we agree on a common formatting rule cc_ @hferentschik

gbraad: I am against reformatting...

gbraad:
> ready for merge

Well, I haven't seen people remove their review, or approve them. So it is not ready for merge, but rather "ready for review" I assume?

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
+    | '*te$st^'   |
+    | 'foo 123'   |
+    | 'foo    '   |
 
-  Scenario: Getting profile foo internal docker registry address
-     Given Minishift has state "Running"
-      When executing "minishift openshift registry" succeeds
-      Then stdout should be valid IP with port number
+  Scenario Outline: As user, I can create profile with alphanumeric character including '_' and '-' in profile name
+    When executing "minishift profile set <profilename>" succeeds
```
gbraad: Wrong indent as "Given" is not used.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
+    | foo         |
 
-  Scenario: User should be able set 'minishift' as the active profile
-      When executing "minishift profile set minishift" succeeds
-      Then stdout should contain
-      """
-      Profile 'minishift' set as active profile
-      """
+  Scenario: As user, I can switch between existing profiles
+    When executing "minishift profile set Test-123" succeeds
```
gbraad: Wrong indent as "Given" is not used.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
+    Then profile "Test-123" should be the active profile
+    When executing "minishift profile set minishift" succeeds
+    Then profile "minishift" should be the active profile
+
+  Scenario: As user, I can start Minishift with default profile 'minishift'
+    Given profile "minishift" has state "Does Not Exist"
+     And profile "minishift" is the active profile
```
gbraad: Wrong indent

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
+     When executing "minishift start" succeeds
+     Then profile "minishift" should have state "Running"
+     When executing "minishift profile list" succeeds
+     Then stdout should match "-\s*minishift\s*Running\s*\(Active\)"
 
-Scenario: User should be able to list 'minishift' as the active profile
-    When executing "minishift profile list" succeeds
+  Scenario: As user, I can apply independent settings in profile 'foo'
+    When executing "minishift profile set foo" succeeds
```
gbraad: Wrong indent as "Given" is not used.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
+    | 20          |
+    | random001z  |
+    | 00_XYZ-     |
+    | test_Pro-01 |
+    | foo         |
 
-Scenario: User can not delete default profile 'minishift'
-    When executing "minishift profile delete minishift --force"
+  Scenario: As user, I cannot run delete command on non existing profile
+    When executing "minishift profile delete XYZ --force"
```
gbraad: Wrong indent as "Given" is not used.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
+     Error: 'XYZ' is not a valid profile
      """
 
-Scenario: Deleting Minishift
-   Given Minishift has state "Running"
-    When executing "minishift delete --force" succeeds
-    Then Minishift should have state "Does Not Exist"
-    When executing "minishift ip"
+  Scenario: As user, I cannot delete default profile 'minishift'
+    When executing "minishift profile delete minishift --force"
```
gbraad: Wrong indent as "Given" is not used.

test/integration/features/profile.feature
```diff
@@ -2,104 +2,161 @@
...
+     And stderr should contain
      """
-     A profile name must be provided. Run `minishift profile list` for a list of existing profiles.
+     Default profile 'minishift' can not be deleted
      """
 
+  Scenario Outline: As user, I can use 'profile' alias to execute profile subcommand
+    'instance' and 'profiles' can be applied independently to all the subcommands of profile.
+    Here subcommand list is only demonstrated with this command.
+    When executing "minishift <profilealias> list" succeeds
```
gbraad: Wrong indent as "Given" is not used.

test/integration/features/profile.feature
amitkrout: @gbraad AFAIU Given is used for precondition check, as a user i can directly run profile set without any profile precondition and moreover its not a good idea to add ```Given``` clause in scenario outline.

test/integration/features/profile.feature
amitkrout: Given is not required here, is it ? other wise given will be overused

test/integration/features/profile.feature
amitkrout: This what we are following the indentation atm. 
See https://github.com/minishift/minishift/pull/1692/files#r155476964 - Referred by @budhrg  

However we have no such doc for indentation rule in feature file. I would suggest lets continue the indentation concern in separate Issue #1815.

test/integration/features/profile.feature
amitkrout: unnecessary or overused

test/integration/features/profile.feature
amitkrout: unnecessary or overused

test/integration/features/profile.feature
amitkrout: unnecessary or overused

test/integration/features/profile.feature
amitkrout: unnecessary or overused

amitkrout:
@gbraad 
> Identation is off for cases where no "Given" is used in the feature. Maybe picky, but it is noticable... since another "And" was not aligned, maybe you can handle this? 

For indentation problem we should have a fair common agreement (Issue #1815) and document it for future reference 

> Waiting for the other reviews to be cleared...

I am also waiting ;)

LalatenduMohanty:
@minishift/minishift-dev I have dismissed the older reviews as these are addressed  in the PR. Please review and give review comments if any, else I will merge it in Monday morning.

gbraad:
You miss a space each time you do NOT use Given. I never said you have to
use given. Just the indentation is not at the same level. What tool do you
use to edit these files? Set guide or ruler lines.

On Friday, December 8, 2017, amitkrout <notifications@github.com> wrote:
> @amitkrout commented on this pull request.
>
> ________________________________
>
> In test/integration/features/profile.feature:
>
>>
> -Scenario: User should be able to list 'minishift' as the active profile
> - When executing "minishift profile list" succeeds
> + Scenario: As user, I can apply independent settings in profile 'foo'
> + When executing "minishift profile set foo" succeeds
>
> unnecessary or overused
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.<
https://ci6.googleusercontent.com/proxy/th54OvL1ZOA2zRfAp77R-FAEHHZHTLofD88u81-CmWXAoSdPAsPtw7Bk6MyD0_bTnIq28mdZcFgK9e7RS2YQWfMWWG2gAekYmqWD5n_8nbGw8ONd74AgI8ooKlvc2SrldZ1xRmPMIUjEpbVAcV6aMzQ3q9fNbg=s0-d-e1-ft#https://github.com/notifications/beacon/AAAHZnlF1exQMCHDnrCk95mr9cJaNBsCks5s-NWngaJpZM4QbvEa.gif
>

-- 

   [ Doing Open Source Matters ]

gbraad:
I responded to #1815, as an example is shown. This is the current markup,
and I do use guidelines when editing the files.

On Friday, December 8, 2017, amitkrout <notifications@github.com> wrote:
> @gbraad
>
> Identation is off for cases where no "Given" is used in the feature.
Maybe picky, but it is noticable... since another "And" was not aligned,
maybe you can handle this?
>
> For indentation problem we should have a fair common agreement (Issue
#1815) and document it for future reference
>
> Waiting for the other reviews to be cleared...
>
> I am also waiting ;)
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub, or mute the thread.<
https://ci4.googleusercontent.com/proxy/cRk0tTg7cQBcS4Cx2WMG01krrGMf24uRLZ90PUWZzPDYbMbvA5Da7l4LXVGPaxKUnxFVnUs9LmuxAc__uDOwykbAOkBzOOQsZ-_Aw5k9SdLaHPdBMTlQJrIm-4pKzw8Z0dNOnKmBdMavcnUSrQ9QYEh2rU5-eA=s0-d-e1-ft#https://github.com/notifications/beacon/AAAHZnEdlkO48i9ugF-Ped2VYvygKQ6Nks5s-NiNgaJpZM4QbvEa.gif
>

-- 

   [ Doing Open Source Matters ]

amitkrout:
@gbraad 
> What tool do you use to edit these files? Set guide or ruler lines.

TBH Whatever the basic indentation guidelines are available i just follow but never used any tool for it. Found two indentation extension ```(Gherkin Indent 2.0.0 and Cucumber (Gherkin) Full Support)``` for vs code ```1.18.1```. Now i can see the indentation line in the feature file.

> You miss a space each time you do NOT use Given.

Applied the change

> I never said you have to use given. Just the indentation is not at the same level. 

I am sorry, i misread it

gbraad:
Be sure ` And` is prepended by a space, as that way all the task definitions align.

Always use your editor to it fullest potential... learn and know as much as possible about it.

amitkrout:
> Be sure And is prepended by a space, as that way all the task definitions align.

Done...Thanks :)

gbraad:
@LalatenduMohanty, You originally approved it. Please have a look if this is still the case... if so, go ahead with merging

Issue #1696, opened 2017-11-14
`minishift profile set` does not report if an empty string/white space is passed as profile name.


anjannath:
```
$ minishift profile set ' '
Profile ' ' set as active profile
$ minishift profile list
-  		Does Not Exist	(Active)
- minishift	Does Not Exist
#but we can also delete the profile
$ minishift profile delete ' '
You are deleting the active profile. It will remove the VM and all related artifacts. Do you want to continue [y/N]?: y
VM for profile ' ' does not exist
Deleted:  /home/anath/.minishift/profiles/ 
Profile ' ' deleted successfully
Switching to default profile 'minishift' as the active profile.
```
we should give an error saying profile name is empty.

amitkrout:
Is there any naming convention checks on creating profile name ?

anjannath:
Not right now, but we could check if the profile name is an empty string or just a bunch of spaces and exit with a message.

hferentschik:
Merged via pull request #1697

PR #1697, opened 2017-11-14
Issue #1696 add check to see if profile name is empty

anjannath:


cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,6 +42,10 @@ func runProfile(cmd *cobra.Command, args []string) {
 	validateArgs(args)
 	profileName := args[0]
 
+	if strings.TrimSpace(profileName) == "" {
```
hferentschik: -1, there is already a _validateArgs_ function. Any checks should probably happen there. Also please add a unit test.

anjannath: Moved the check to validateArgs function but that function doesn't return anything, just exits with a message, so it means we have to add integration test?? @hferentschik

cmd/minishift/cmd/profile/profile_set.go
```diff
@@ -41,6 +42,10 @@ func runProfile(cmd *cobra.Command, args []string) {
 	validateArgs(args)
```
LalatenduMohanty: As @hferentschik  suggested, we need to add the logic to `validateArgs(args)` in profile.go.

anjannath: moved the check to _validateArgs_.

hferentschik: +1, now please add some tests. Especially everything around validateArgs is easily testable since it is purely dependent on the CLI input (no running VM needed, etc). See also https://github.com/minishift/minishift/blob/4b9ddbf40f525d63a86eac087b1406997b1a5ce9/cmd/minishift/cmd/addon/addon_apply_test.go

You can call runProfile directly and then assert that the system exits with a specific error message.

anjannath: Thanks @hferentschik for the example, will add tests now. and about the other pull requests the arch linux documentation one got merged yesterday, i have updated #1621 PTAL when you get time, will have to do some work on #1640

test/integration/features/profile.feature
```diff
@@ -89,3 +89,16 @@ Scenario: User should be able to switch between non existing profiles
       oc cli context could not changed for 'minishift'. Make sure the profile is in running state or restart if the problem persists.
       Profile 'minishift' set as active profile
       """
+Scenario: User should not be able to create profile with blank profile name
+    When executing "minishift profile set"
+     Then exitcode should equal "1"
+     Then stdout should contain
```
LalatenduMohanty: Is this test passing locally? I think it should be `Then stderr should contain`

anjannath: Right, i see, i can't run integration tests locally takes a lot of time, and the system become unusable :( So pushed it.

anjannath: Is the unit test fine? Checks only the empty string condition, not testing for correct argument.

cmd/minishift/cmd/profile/profile_test.go
```diff
@@ -0,0 +1,40 @@
...
+package profile
+
+import (
+	"testing"
+
+	"github.com/minishift/minishift/pkg/testing/cli"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+)
+
+func Test_validate_args_exit_when_profilename_empty(t *testing.T) {
```
hferentschik: +1 Why don't you add tests for the other two cases as a good measure?

hferentschik: Or you can do a table based test again

test/integration/features/profile.feature
gbraad: You could always push with `[WIP]` as state, and check the CI results for the message (as CI is able to run the integration tests).

test/integration/features/profile.feature
```diff
@@ -89,3 +89,17 @@ Scenario: User should be able to switch between non existing profiles
       oc cli context could not changed for 'minishift'. Make sure the profile is in running state or restart if the problem persists.
       Profile 'minishift' set as active profile
       """
+Scenario: User should not be able to create profile with blank profile name
```
hferentschik: Given that we cover this in unit tests, there is not much value here, but I guess for completeness sake.

test/integration/features/profile.feature
```diff
@@ -89,3 +89,17 @@ Scenario: User should be able to switch between non existing profiles
...
+    When executing "minishift profile set"
+     Then exitcode should equal "1"
+     Then stderr should contain
+     """
+     You must provide the profile name. Run `minishift profile list` to view profiles
```
hferentschik: Could you please change the message? 

"A profile name must be provided. Run `minishift profile list` for a list of existing profiles"

test/integration/features/profile.feature
```diff
@@ -89,3 +89,17 @@ Scenario: User should be able to switch between non existing profiles
...
+     """
+    When executing "minishift profile set '  '"
+     Then exitcode should equal "1"
+     Then stderr should contain
+     """
+     You must provide the profile name. Run `minishift profile list` to view profiles
+     """
+
```
hferentschik: newline missing. Let's make GitHub happy.

cmd/minishift/cmd/profile/profile_set.go
gbraad: agree with @hferentschik here... I see many issues on your name and several unfinished PRs. Try not to start new work, unless it is working/merged.

Especially as you can use the time in-between to look at some usage and general knowledge about OpenShift.

cmd/minishift/cmd/profile/profile_test.go
anjannath: > Or you can do a table based test again

I also thought of it, so i tried it but for some reason the tests always pass when it should failed, i reckon because the function exits the loop never moves foreward.. is that right?

test/integration/features/profile.feature
anjannath: Will change this.

test/integration/features/profile.feature
anjannath: Sure :) Will do..

cmd/minishift/cmd/profile/profile_test.go
hferentschik: what did you try. It might help to put it back, so that we can have a look. Or just make it three separate tests. I'm fine with that.

cmd/minishift/cmd/profile/profile_test.go
```diff
@@ -0,0 +1,40 @@
...
+	tmpMinishiftHomeDir := cli.SetupTmpMinishiftHome(t)
```
hferentschik: I don't think your test needs that. A test should be as minimal as possible. You are just doing input validation, there is no dependence on the Minishift home dir.

anjannath: Yes that crossed my mind too, but the `cli.TearDown` function needs a `tmpMinishiftHomedir` as the first argument, and if the teardown is not called i get lots of error that i don't understand, passing an empty string seems to work, is it okay to pass empty string to `cli.TearDown`

cmd/minishift/cmd/profile/profile_test.go
anjannath: I lost the code that i wrote but here's roughly what i did, i tried what we did for the iso url test, making a struct like this:
```
type testData struct {
        in string
        out string
        exitCode int
}

then looped over it calling validArgs and registering a exit handler, then matched the actual output with `testData.out`.

File: cmd/minishift/cmd/profile/profile.go [hunk 9]
```diff
@@ -17,12 +17,14 @@ limitations under the License.
...
 import (
+	"strings"
+
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
 )
 
 const (
-	emptyProfileMessage  = "You must provide the profile name. Run `minishift profile list` to view profiles"
+	emptyProfileMessage  = "A profile name must be provided. Run 'minishift profile list' for a list of existing profiles"
```
hferentschik: full stop after 'profiles'. Please also add a full stop after the other error message.

anjannath:
@hferentschik This is ready now i belive, made the required changes as you suggested.

hferentschik:
Rebased and merged. Thanks.

PR #1710, opened 2017-11-17
Issue #1662 Add profile line into start log

### LINKED ISSUE #1662: No indication of profile being started
I would like to have as first line of the logs when you start, something like:

```
-- Starting minishift profile XYZ
```

ulyssessouza:
Fixes #1662

- Add profile line into start log

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -165,6 +165,7 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	proxyConfig := handleProxies()
 
+	fmt.Println(fmt.Sprintf("-- Starting profile '%s'", minishiftConfig.AllInstancesConfig.ActiveProfile))
 	fmt.Printf("-- Starting local OpenShift cluster")
 	hostVm := startHost(libMachineClient)
 	registrationUtil.RegisterHost(libMachineClient)


budhrg:
add to whitelist

gbraad:
We might consider to combine the two lines "Starting profile ..." and "Starting local OpenShift"...
The last one gets appended with the hypervistor info, but we can change this by breaking it further up.

WDYT?

cmd/minishift/cmd/start.go
```diff
@@ -165,6 +165,7 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	proxyConfig := handleProxies()
 
+	fmt.Println(fmt.Sprintf("-- Starting profile '%s'", minishiftConfig.AllInstancesConfig.ActiveProfile))
```
LalatenduMohanty: You need to use `constants.ProfileName` instead of `minishiftConfig.AllInstancesConfig.ActiveProfile` as user can start the profile with the `--profile flag` e.g. `minishift --profile foo start`

LalatenduMohanty: There will be use cases where active profile is `bar` but user is trying to start profile `foo`.

ulyssessouza: Nice! I'm gonna change that

gbraad:
Can you try to rebase, something like

```
git rebase -i HEAD~3
```

and (s)quash your changes into a single commit?

Issue #1507, opened 2017-10-06
Switch dependency management from Glide to dep


hferentschik:
Glide is in support mode and no new feature development happens there. The community has aligned behind [dep](https://github.com/golang/dep) which behaves similar and is a candidate to be part of the official Go toolchain.

From the Glide homepage:

> The Go community now has the dep project to manage dependencies. Please consider trying to migrate from Glide to dep. If there is an issue preventing you from migrating please file an issue with dep so the problem can be corrected. Glide will continue to be supported for some time but is considered to be in a state of support rather than active feature development.

gbraad:
The glide importer has since been removed: `cmd/dep/glide_importer.go` => https://github.com/golang/dep/blob/f88c2339516836dad313a6b8b6de8430f0cfa65f/cmd/dep/glide_importer.go (older version) It is a one-shot anyways... Import once, if it worked we move on...

hferentschik:
I am not quite sure why this is part of the next sprint. Even though it is something we should do eventually, there is no need for doing this right now. This can easily become a time sink.

There are plenty of things which have a much higher value imo.

gbraad:
> There are plenty of things which have a much higher value imo.

such as #1714

hferentschik:
Or the "default" Virtual Switch.

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/pull/1715

PR #1715, opened 2017-11-17
Issue #1507 Switch dependency management from Glide to Dep

praveenkumar:
- [x] Need to update documents
- [x] Need to adjust all the CI.


hferentschik:
Are there any changes in the general workflow?

praveenkumar:
> Are there any changes in the general workflow?

@hferentschik  no, good thing is that dep tool init will grab most of the stuff from glide because they support glide and godep directly. Once I make change for centos CI that also make sure we don't have any regression over this.

praveenkumar:
retest this please

hferentschik:
Good stuff. Any noticeable change at all,  eg speed to get dependencies? Does it like Glide cache the dependencies in your home directory?

hferentschik:
@praveenkumar Any chance Dep supports [Trash](https://github.com/rancher/trash) as well?

hferentschik:
> No, it doesn't do any caching locally, always check the current vendor directory.

Are you sure? So if I do a `make clean` which removes the vendor directory and then a `make vendor`, it has to download each dependency  from the web again? How could that ever be faster than having a local cache?

praveenkumar:
> Are you sure? So if I do a make clean which removes the vendor directory and then a make vendor, it has to download each dependency from the web again? How could that ever be faster than having a local cache?

Looks like as per docs https://github.com/golang/dep/blob/master/docs/FAQ.md#why-is-dep-slow it does cache locally to `$GOPATH/pkg/dep` and till there is any change in the dependency sha it uses this every time. It doesn't have a dedicated command to remove that cache though like we had in `glide` and that is the reason I thought it doesn't do any local cache.

hferentschik:
> Looks like as per docs https://github.com/golang/dep/blob/master/docs/FAQ.md#why-is-dep-slow it does cache locally to $GOPATH/pkg/dep and till there is any change in the dependency sha it uses this every time.

Ahh, here we do. Interesting location for the cache

 > It doesn't have a dedicated command to remove that cache though like we had in glide and that is the reason I thought it doesn't do any local cache.

That's a bit of a shame

docs/source/contributing/developing.adoc
```diff
@@ -89,19 +89,19 @@ For instructions on setting up IDEA, see link:http://hadihariri.com/2015/09/30/s
...
-=== Installing Glide
+[[install-dep]]
+=== Installing Dep
 
-Before you can use Glide you need to download and install it from GitHub:
+Before you can use Dep you need to download and install it from GitHub:
 
 ----
-$ go get github.com/Masterminds/glide
+$ go get github.com/golang/dep
```
LalatenduMohanty: This need to be fixed as `go get github.com/golang/dep/cmd/dep` as `go get github.com/golang/dep` does not work for me.

LalatenduMohanty:
Once cached, I think it is faster than glide. The PR works for me except the documentation change.

docs/source/contributing/developing.adoc
```diff
@@ -114,33 +114,27 @@ To install the dependencies, you can run the following command:
...
-Glide will recognize that there is no lock file and recalculate the required dependencies.
+. Delete *_Gopkg.lock_* and re-create the vendor directory by running `make vendor`.
+Dep will recognize that there is no lock file and recalculate the required dependencies.
 
-. Check-in the updated *_glide.yaml_* and *_glide.lock_* files.
+. Check-in the updated *_Gopkg.toml_* and *_Gopkg.lock_* files.
 
 . Test that everything still compiles with the new lock file in place by running `make clean && make`.
 
-[TIP]
```
hferentschik: What's about a tip regarding Dep caching?

docs/source/contributing/developing.adoc
```diff
@@ -114,31 +114,31 @@ To install the dependencies, you can run the following command:
...
+. Check-in the updated *_Gopkg.toml_* and *_Gopkg.lock_* files.
 
 . Test that everything still compiles with the new lock file in place by running `make clean && make`.
 
 [TIP]
 ====
-The Glide cache located under *_~/.glide/cache_* can get corrupted in some cases.
-If you see Glide errors during `make vendor`, you can clear the Glide cache via `glide cc`.
+The Dep cache located under *_$GOPATH/pkg/dep_*.
+If you see any Dep errors during `make vendor`, you can remove local cached directory and try again.
```
praveenkumar: We wrote this tip for glide because we were getting an error during `glide install` but right now we are not sure we need for dep but I have it this as a placeholder and if this change then we will update accordingly.

Issue #1408, opened 2017-09-19
Make Minishift cache global shared between all profiles


gbraad:
Profile needs a shared cache solution. It is too slow to have to redownload everything every time

```
9.20 MiB / 33.92 MiB [============>--------------------------------------------]  27.11% 23m19s
```

gbraad:
With the `--iso-url` option, you can bypass the download, by specifying an ISO from a cache which is already downloaded. If #1409 is available, this can also work for the `oc` client.

hferentschik:
> So, we are using ~/.minishift/cache for the global cache right?

most likely

gbraad:
> most likely

A little late to ask, but it concerns me. Has this been decided/can it be answered ? It is a too vague and should be given before we can start on an implementation. My 2c. I can assume it would have to be... but...

gbraad:
Fixed with 07c8570

LalatenduMohanty:
PR  reference https://github.com/minishift/minishift/pull/1716

Issue #1144, opened 2017-07-17
Allow cache to be cleared without having defined an instance


gbraad:
It seems the command: `minishift delete --clear-cache` only works when an instance has been defined.

     Running this command requires an existing 'minishift' VM, but no VM is defined.

This does not seem like the expected behaviour.

LalatenduMohanty:
@gbraad Just to be clear, you are saying `minishift delete` fine to check if VM is defined before deleting it , but in case of `minishift delete --clear-cache` it should bypass the check and just delete the cache if VM is not defined?

gbraad:
Well, this is a question we need an answer to... else we need to advice to do `rm -rf ~/.minishift/cache` which defeats the purpose of this command.

But also, should clearing the cache really remove the VM? It seems to be a command that does two things: deleting the instance, and removing the cache. Sounds like a side-effect.

LalatenduMohanty:
I see , you mean --clear-cache operation should be done independently, right?  Example: if user has deleted the instance but then want to delete the cache how does he do it.  In practice if we just talk about ISO and OC user should not need to clean the cache unless it is an uninstall. For uninstall we suggest deleting the `.minishift` which is ok I think. With respect to images , I think image command should provide some command for image management.

gbraad:
currently, we can have corrupted downloads in the cache. this is the reason why we have a `--clear-cache` command. if downloads are checked, this might become irrelevant.

although, at the moment using this command is not error-free, as it suffers from a side-effect (existence of a VM is needed).

budhrg:
@anjannath Can you have a look?

anjannath:
@budhrg Sure, looking into it

anjannath:
We still need to fix this.. right?

gbraad:
Fixed with 675539a

Issue #1402, opened 2017-09-18
Refactor code of function UpdateProfileConstants


LalatenduMohanty:
In function `UpdateProfileConstants()` we change some variables from `pkg/minikube/contstants/constants.go` [1] . It is because cluster.GetHostStatus(api) uses `constants.MachineName` directly.

However we should not change these outside of `cmd/minishift/cmd/root.go`. We should find out how we can handle this in a better way.  

```
func UpdateProfileConstants(profileName string) {
	constants.ProfileName = profileName
	constants.MachineName = constants.ProfileName
	constants.Minipath = constants.GetProfileHomeDir()
}
```
[1] https://github.com/minishift/minishift/blob/master/pkg/minishift/profile/profile.go#L74-L83

budhrg:
Moreover, `UpdateProfileConstants()` should be renamed to some `UpdateProfileVariables()` or similar lines.

LalatenduMohanty:
In commit 07c8570fed7edc6e254b6d208fc93f83507820c9 we have removed the dependency on `UpdateProfileConstants` in `GetVMStatus`. 
```diff
iff --git a/cmd/minishift/cmd/util/util.go b/cmd/minishift/cmd/util/util.go
index 4214c14b..8da123f8 100644
--- a/cmd/minishift/cmd/util/util.go
+++ b/cmd/minishift/cmd/util/util.go
@@ -24,11 +24,12 @@ import (
        "github.com/docker/machine/libmachine/drivers"
        "github.com/docker/machine/libmachine/state"
        "github.com/minishift/minishift/pkg/minikube/cluster"
-       "github.com/minishift/minishift/pkg/minikube/constants"
        profileActions "github.com/minishift/minishift/pkg/minishift/profile"
        "github.com/minishift/minishift/pkg/util/os/atexit"

+       cmdState "github.com/minishift/minishift/cmd/minishift/state"
        "github.com/minishift/minishift/out/bindata"
+       "github.com/minishift/minishift/pkg/minikube/constants"
 )

 var (
@@ -67,10 +68,11 @@ func ExitIfNotRunning(driver drivers.Driver, machineName string) {

 func GetVMStatus(profileName string) string {
        var status string
-       profileActions.UpdateProfileConstants(profileName)
-       api := libmachine.NewClient(constants.Minipath, constants.MakeMiniPath("certs"))
+       profileDirs := cmdState.NewMinishiftDirs(constants.GetProfileHomeDir(profileName))
+
[...]

PR #1716, opened 2017-11-17
Issue #1408 Using a global cache for all profiles

hferentschik:
Addresses issue #1408 and #1144

Functionally working, but I want to do some more changes around _config.InstanceDirs_. There are several things which need straightening out. Also needs a doc update. 

Overall a global cache is, however, not hard to achieve.


budhrg:
Functionally working for me in Linux F26. I can see that when I create another profile after having existing default profile, the new VM is created with cached ISO and oc.

Also, If I create any profile without having default profile, it cache to the global cache location.

hferentschik:
@minishift/minishift-dev WIP status lifted :-)

hferentschik:
@budhrg now it is rebased :-)

hferentschik:
retest this please

gbraad:
Please rebase, so the two extraneous commits can be disregarded

LalatenduMohanty:
> we should update the minishift delete docs --clear-cache behavior https://docs.openshift.org/latest/minishift/using/basic-usage.html#minishift-delete-overview

Specially now because there is global cache and there local cache which is instance specific. 
I am not sure when the local cache will be used. If it will not be used then do we need instance specific `clear-cache`.

hferentschik:
Rebased.

hferentschik:
> IMO we should update the minishift delete docs --clear-cache behavior

Sure. There is #1144 to track this. I guess one needs to decide how we want to do this. Really, in the long run we probably need some cache specific commands which are not tied to _delete_ at all. 
I think we need to think this through a bit more. Since there is really no harm in the current behaviour,  I think we still could go ahead. @LalatenduMohanty do you have something concrete in mind?

LalatenduMohanty:
@hferentschik  Looks like `minishift delete --clear-cache` is deleting the global cache even if the active profile is a different one. Is this expected?

```
$ minishift profile list
- 1st-java-project	Does Not Exist
- minishift		Does Not Exist
- test			Running		(Active)

$ minishift delete --clear-cache
You are deleting the Minishift VM: 'test'. Do you want to continue [y/N]?: y
Removed the cache at: /Users/lmohanty/.minishift/cache
Deleting the Minishift VM...
Minishift VM deleted.

$ tree ~/.minishift/profiles/test/cache/
/Users/lmohanty/.minishift/profiles/test/cache/
├── images
│   ├── blobs
│   │   └── sha256
│   │       ├── 2fdfe1cd78c20d05774f0919be19bc1a3e4729bce219968e4188e7e0f1af679d
│   │       ├── 338d6757fb11f950e39ad4cf9da50b5eb2c22db64058022491e3e3d3776f25be
│   │       ├── 3d8f7fb2019749b501d60207453997186d96d85a50d4c5ac763b356760462065
│   │       ├── 501d1a8f0487e93128df34ea349795bc324d5e0c0d5112e08386a9dfaff620be
│   │       ├── 5a1f719285eb1aaf9459acfe5c4b01d7813426e36f09b61d965f1a86141a74ef
│   │       ├── 5fd7209ce0c3ff0bf7571ec3abeb49ab21ef85ab01e9bea29e9f3201d37a8ec4
│   │       ├── 68beca863b32d3ff6b689a41c676be2490ab7392e3ccdb8e1cfe1d37202909e8
│   │       ├── 69852451c4471fd59ca2baa054120c9fe622df4147c17eadce9cd81a37cb2fac
│   │       ├── a42bccc56fc8ff314e67f75aa86b5a349f610378e790bd65a281e259050630ae
│   │       ├── a6f02728622186556d5842e44cd73b5e8b81ac43025e629e7a120351f23b41ea
│   │       ├── bb1fb6e9d0509afa26e52b75bdf144faea07d2473c9d5c693427b6f09b6c3ba5
│   │       ├── c577c840a769fe3ecc2f7617640c03eca9ddad9b762208d3c9453fee8aa3d327
│   │       ├── d4f4bcfdfdef569fea25d1aae1626df533393709e575a68f5be55941eeffd3a2
│   │       ├── ec2751d932c2d68ef6993e70266d31d00b1de7672921db2488632e8276fa0b24
│   │       └── ff04a77efbee3d708c4ba5c379e00f72130a0477916d876673c7985660644523
│   ├── index.json
│   └── oci-layout
├── iso
│   ├── b2d
│   │   └── v1.2.0
│   │       └── minishift-b2d.iso
│   └── minishift-b2d.iso
└── oc
    └── v3.6.0
        ├── darwin
        │   └── oc
        └── oc

9 directories, 21 files

$ tree ~/.minishift/cache
/Users/lmohanty/.minishift/cache [error opening dir]

0 directories, 0 files
```

LalatenduMohanty:
Also #1144 is for changing the behavior around `--clear-cache`. But we need to add documentation for the change we are doing in this PR i.e. the concept of global cache vs local cache and what should `--clear-cache` should delete.

gbraad:
I think the scenarios need to be described and detailed in the issue. This also makes reviewing easier.

hferentschik:
> But we need to add documentation for the change we are doing in this PR i.e. the concept of global cache vs local cache and what should --clear-cache should delete.

So you think the changes I made to the docs are not enough? Did you actually see them?

hferentschik:
> I think the scenarios need to be described and detailed in the issue

Not sure what you are talking about? Which scenarios? The difference is that now all profiles share the same cache.

LalatenduMohanty:
> So you think the changes I made to the docs are not enough? Did you actually see them?

I saw that you have added a tip in to profile documentation. But I feel we need to change delete message for `$ minishift delete --clear-cache` and the documentation for `minishift delete`

Currently  it gives following message and it does not communicate about effects of deleting the cache. 
```
$ minishift delete --clear-cache
You are deleting the Minishift VM: 'minishift'. Do you want to continue [y/N]?: N
```

It should be something as below
```
$ minishift delete --clear-cache
You are deleting the Minishift VM: 'minishift'. Do you want to continue [y/N]?: y
You are deleting cache for all profiles: Do you want to continue [y/N]?: y
```
May be we can do with one confirmation. I am not sure at this point.

hferentschik:
> It should be something as below

> $ minishift delete --clear-cache
> You are deleting the Minishift VM: 'minishift'. Do you want to continue [y/N]?: y
> You are deleting cache for all profiles: Do you want to continue [y/N]?: y

> May be we can do with one confirmation. I am not sure at this point.

Fair enough. I talked w/ @budhrg about this before as well. We for sure can add something to the message. I thought I had done this already.  Do you really think we need to confirmations though?

gbraad:
> scenarios

  * `m start` pulls ISO image
  * `m delete` removes the instance
  * `m start --profile new` does NOT pull the ISO image
  * `m delete --clear-cache` should remove cache 
    * instance cache?

Mostly about the last scenario(s). What is expected to happen? As this will also have to result in integration tests.

gbraad:
>  Do you really think we need to confirmations though?

'two' i guess... well, if `--clear-cache` would work regardless of having an instance, this can be avoided. but In this case, an instance i s required to even execute the `--clear-cache` command... so, yes... unfortunately the current implementation 'forces' this.

hferentschik:
> 'two' i guess... well, if --clear-cache would work regardless of having an instance, this can be avoided. but In this case, an instance i s required to even execute the --clear-cache command... so, yes... unfortunately the current implementation 'forces' this.

We could just change the order and do the cache clear before the instance delete.

LalatenduMohanty:
> We could just change the order and do the cache clear before the instance delete.

Thats good idea. We should check for `clearCache` before `util.ExitIfUndefined(api, constants.MachineName)` and it should ask for a confirmation (separate than VM delete confirmation). It should respect `--force`.

hferentschik:
> Thats good idea. We should check for clearCache before util.ExitIfUndefined(api, constants.MachineName) and it should ask for a confirmation (separate than VM delete confirmation). It should respect --force.

I'll try to update the pull request asap. Latest tomrrow.

hferentschik:
Updated pull request. Changed the implementation of `delete` to first handle the cache deletion (paired with user interaction). Also added tests for delete command.

hferentschik:
@gbraad @LalatenduMohanty - pull requests addresses now #1408 and #1144

hferentschik:
retest this please

hferentschik:
@LalatenduMohanty all tests passing again.

budhrg:
Tested and functionally working.

Also, on `m delete --clear-cache` we are getting expected message and working
```
$ minishift delete --clear-cache
This will delete the cache content for all profiles. Do you want to continue [y/N]?: n
You are deleting the Minishift VM: 'budh'. Do you want to continue [y/N]?: y
Deleting the Minishift VM...
Minishift VM deleted.

$ tree ~/.minishift/cache/
/home/budhram/.minishift/cache/
├── [budhram   4.0K]  images
│   ├── [budhram   4.0K]  blobs
│   │   └── [budhram   4.0K]  sha256
│   │       ├── [budhram   1.7K]  338d6757fb11f950e39ad4cf9da50b5eb2c22db64058022491e3e3d3776f25be
│   │       ├── [budhram   5.9M]  3d8f7fb2019749b501d60207453997186d96d85a50d4c5ac763b356760462065
│   │       ├── [budhram    981]  5a1f719285eb1aaf9459acfe5c4b01d7813426e36f09b61d965f1a86141a74ef
│   │       ├── [budhram   1.7K]  5fd7209ce0c3ff0bf7571ec3abeb49ab21ef85ab01e9bea29e9f3201d37a8ec4
│   │       ├── [budhram    420]  68beca863b32d3ff6b689a41c676be2490ab7392e3ccdb8e1cfe1d37202909e8
│   │       ├── [budhram   132M]  69852451c4471fd59ca2baa054120c9fe622df4147c17eadce9cd81a37cb2fac
│   │       ├── [budhram   115M]  a42bccc56fc8ff314e67f75aa86b5a349f610378e790bd65a281e259050630ae
│   │       ├── [budhram    823]  a6f02728622186556d5842e44cd73b5e8b81ac43025e629e7a120351f23b41ea
│   │       ├── [budhram    823]  bb1fb6e9d0509afa26e52b75bdf144faea07d2473c9d5c693427b6f09b6c3ba5
│   │       ├── [budhram   2.1K]  d4f4bcfdfdef569fea25d1aae1626df533393709e575a68f5be55941eeffd3a2
│   │       ├── [budhram    60M]  ec2751d932c2d68ef6993e70266d31d00b1de7672921db2488632e8276fa0b24
│   │       └── [budhram    69M]  ff04a77efbee3d708c4ba5c379e00f72130a0477916d876673c7985660644523
│   ├── [budhram    904]  index.json
│   └── [budhram     31]  oci-layout
├── [budhram   4.0K]  iso
│   └── [budhram   4.0K]  b2d
│       └── [budhram   4.0K]  v1.2.0
│           └── [budhram    40M]  minishift-b2d.iso
└── [budhram   4.0K]  oc
    ├── [budhram   4.0K]  v3.6.0
        └── [budhram   4.0K]  linux
            └── [budhram   179M]  oc


11 directories, 17 files

$ minishift delete --clear-cache
This will delete the cache content for all profiles. Do you want to continue [y/N]?: y
Removed cache content at: /home/budhram/.minishift/cache

$ tree ~/.minishift/cache/
/home/budhram/.minishift/cache/ [error opening dir]

0 directories, 0 files
```

cmd/minishift/cmd/start_test.go
```diff
@@ -34,11 +34,12 @@ import (
 
 	"bytes"
 
+	"github.com/minishift/minishift/cmd/testing/cli"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minikube/tests"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
 	instanceState "github.com/minishift/minishift/pkg/minishift/config"
-	"github.com/minishift/minishift/pkg/testing/cli"
+	pkgTest "github.com/minishift/minishift/pkg/testing"
```
praveenkumar: @hferentschik below you put `pkgTesting` when imported this package and here `pkgTest`, can we have it same across?

cmd/minishift/state/instance_dirs.go
```diff
@@ -37,20 +38,21 @@ type MinishiftDirs struct {
...
-		Tmp:        constants.MakeMiniPath("tmp"),
-		Config:     constants.MakeMiniPath("config"),
+		Home:       baseDir,
+		Certs:      filepath.Join(baseDir, "certs"),
+		Machines:   filepath.Join(baseDir, "machines"),
+		Addons:     filepath.Join(baseDir, "addons"),
+		Logs:       filepath.Join(baseDir, "logs"),
+		Tmp:        filepath.Join(baseDir, "tmp"),
+		Config:     filepath.Join(baseDir, "config"),
+		Cache:      cacheDir,
```
LalatenduMohanty: 👍

pkg/minikube/cluster/cluster.go
```diff
@@ -132,18 +132,18 @@ func DeleteHost(api libmachine.API) error {
 	return m.ToError()
 }
 
-// GetHostStatus gets the status of the host VM.
-func GetHostStatus(api libmachine.API) (string, error) {
+// GetHostStatus gets the status of the host VM with the specified name.
+func GetHostStatus(api libmachine.API, machine string) (string, error) {
```
LalatenduMohanty: Initially we had planned that we will not change this as it would make difficult for us to use Minikube as a pkg. Hence we had https://github.com/minishift/minishift/issues/1402. I wanted to give a serious try to use minikube as a lib. So little skeptical with this change.

pkg/minikube/constants/constants.go
```diff
@@ -60,14 +60,14 @@ func MakeMiniPath(fileName ...string) string {
 	return filepath.Join(args...)
 }
 
-// GetProfileHomeDir determines the Minishift home directory for the profile.
-func GetProfileHomeDir() string {
-	if ProfileName != DefaultProfileName {
+// GetProfileHomeDir determines the base directory for the specified profile name
+func GetProfileHomeDir(profile string) string {
```
LalatenduMohanty: 👍

cmd/minishift/cmd/start_test.go
LalatenduMohanty: @praveenkumar I am not sure what is your suggestion? Do you mean we should have "pkgTest" when we import "github.com/minishift/minishift/pkg/testing" ? Because now we have testing in `pkg` as well as `cmd`.

cmd/testing/cli/harness.go
```diff
@@ -102,3 +105,35 @@ func TearDown(testDir string, tee *Tee) {
 		}
 	}
 }
+
+// PrepareStdinResponse creates a temproary file with a prepared content which then is used as os.Stdin to test user input.
+// The orignal os.Stdin file handle is returned as well as the path of the file containing the canned stdin responses.
+func PrepareStdinResponse(s string, t *testing.T) (*os.File, string) {
```
LalatenduMohanty: This is not used in this commit. But overall I can see delete_test.go uses PrepareStdinResponse() in another commit. Is this fine or we should keep them together?

cmd/minishift/cmd/delete_test.go
```diff
@@ -0,0 +1,95 @@
...
+func prepareCacheDir(t *testing.T) (string, *pgkTesting.Tee) {
+	// setup the test Minishift home directory and make sure we have a cache directory
+	tmpMinishiftHomeDir := cli.SetupTmpMinishiftHome(t)
+	os.MkdirAll(state.InstanceDirs.Cache, os.ModePerm)
+
+	// create a tee to keep the test silent and make sure that the temporary Minishift home directory gets cleaned up
+	tee := cli.CreateTee(t, true)
+
+	return tmpMinishiftHomeDir, tee
+}
```
LalatenduMohanty: Really liked the unit tests. Thanks.

hferentschik: yeah, I like the end result as well. Thanks. Went through a couple of iterations. That's also the context in which I realized that some other tests were not quite right.

cmd/testing/cli/harness.go
hferentschik: That's fine. It's a helper really. I should have staged in hunks. But in this case it really does not matter.

pkg/minikube/cluster/cluster.go
hferentschik: Sure, I am aware of this. However, this is a minor change and when it comes to #1402 we can always write a couple of adapters. In fact we will need to do this anyways. The improved code structure outweighs imo this particular change. 

I can later also explain what I mean with adapters. I think this be the best way to address #1402.

cmd/minishift/cmd/start_test.go
hferentschik: You guy realize that there is an import conflict. The only reason I do it, is that there are two 'testing' packages imported/needed here.

pkg/minikube/constants/constants.go
hferentschik: using constants in methods is bad. We should rather pass parameters. The more functions rely internally on constants, the tighter the coupling gets and the harder to change.

pkg/minikube/cluster/cluster.go
gbraad: An abstraction (or override) can be used to feed the function with the constant if needed. So this change is minimal.

cmd/minishift/cmd/start_test.go
praveenkumar: @hferentschik I know about having two testing package which will import conflict here my suggestion was something below. like you have it  in `cmd/testing/cli/harness.go` file for same commit. It just a name nothing else.

```
pkgTesting "github.com/minishift/minishift/pkg/testing"
```

cmd/minishift/state/instance_dirs.go
gbraad: Finally see my suggestion implemented 👍

hferentschik:
Awesome. Thanks for merging. Frees up some of my attention to look back at my sshfs branch. Need to rebase now and continue my work there.

PR #1723, opened 2017-11-20
Issue #1714 Profile name should be alpha-numeric

### LINKED ISSUE #1714: Only alpha-numeric characters should be allowed as a valid profile name
As discussed with @LalatenduMohanty and @gbraad , there need to put some check around some special character.

for example : 
 ```. ? / & * % # $ @```

praveenkumar:
Fix #1714

### DIFF
--- a/cmd/minishift/cmd/profile/profile.go
+++ b/cmd/minishift/cmd/profile/profile.go
@@ -19,6 +19,7 @@ package profile
 import (
 	"strings"
 
+	cmdUtil "github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
 )
@@ -26,6 +27,7 @@ import (
 const (
 	emptyProfileMessage  = "A profile name must be provided. Run `minishift profile list` for a list of existing profiles."
 	extraArgumentMessage = "You have provided more arguments than required. You must provide a single profile name."
+	invalidNameMessage   = "Profile names must consist of alphanumeric characters only."
 )
 
 var ProfileCmd = &cobra.Command{
@@ -45,5 +47,8 @@ func validateArgs(args []string) {
 		atexit.ExitWithMessage(1, extraArgumentMessage)
 	} else if strings.TrimSpace(args[0]) == "" {
 		atexit.ExitWithMessage(1, emptyProfileMessage)
+	}
+	if !cmdUtil.IsValidProfileName(args[0]) {
+		atexit.ExitWithMessage(1, invalidNameMessage)
 	}
 }
--- a/cmd/minishift/cmd/util/util.go
+++ b/cmd/minishift/cmd/util/util.go
@@ -18,6 +18,7 @@ package util
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/drivers"
@@ -96,4 +97,10 @@ func IsValidProfile(profileName string) bool {
 		}
 	}
 	return false
+}
+
+// IsValidProfileName return true if profile name follow ^\w+[\w+-]+$
+func IsValidProfileName(profileName string) bool {
+	rr := regexp.MustCompile(`^\w+[\w+-]+$`)
+	return rr.MatchString(profileName)
 }
--- a/cmd/minishift/cmd/util/util_test.go
+++ b/cmd/minishift/cmd/util/util_test.go
@@ -1,0 +1,44 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+import (
+	"testing"
+)
+
+func TestIsValidProfileName(t *testing.T) {
+	testData := []struct {
+		profileName string
+		expected    bool
+	}{{"test", true},
+		{".", false},
+		{"@foo", false},
+		{"foo123", true},
+		{"123foo", true},
+		{"foo_123", true},
+		{"foo@123", false},
+		{"", false},
+		{"-foo", false},
+		{"--hell", false},
+		{"foo-123", true}}
+	for _, v := range testData {
+		got := IsValidProfileName(v.profileName)
+		if got != v.expected {
+			t.Errorf("Expected *%s* profileName to be %t, Got %t", v.profileName, v.expected, got)
+		}
+	}
+}
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -213,6 +213,9 @@ func initializeProfile() string {
 	for i, arg := range os.Args {
 		if arg == "--"+profileFlag {
 			profileName = os.Args[i+1]
+			if !cmdUtil.IsValidProfileName(profileName) {
+				atexit.ExitWithMessage(1, "Profile names must consist of alphanumeric characters only.")
+			}
 			break
 		}
 	}
--- a/docs/source/using/profiles.adoc
+++ b/docs/source/using/profiles.adoc
@@ -27,7 +27,15 @@ These commands are described in the following sections.
 [[creating-profiles]]
 == Creating Profiles
 
-There are two ways to create a new profile.
+There are two ways to create a new profile. 
+
+[NOTE]
+====
+Profile names must consist of alphanumeric characters only, for example:
+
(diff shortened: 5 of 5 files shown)


cmd/minishift/cmd/start_test.go
```diff
@@ -512,7 +512,7 @@ func setUp(t *testing.T) {
 	viper.Set("host-volumes-dir", hostVolumesDirectory)
 	viper.Set("host-pv-dir", hostPvDirectory)
 
-	provision.SetDetector(&tests.MockDetector{&tests.MockProvisioner{Provisioned: true}})
+	provision.SetDetector(&tests.MockDetector{Provisioner: &tests.MockProvisioner{Provisioned: true}})
```
praveenkumar: This changed happen to make `go vet` happy.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -135,7 +135,7 @@ func (m *AddOnManager) Enable(addonName string, priority int) (*addon.AddOnConfi
 	addOn.SetEnabled(true)
 	addOn.SetPriority(priority)
 
-	return &addon.AddOnConfig{addonName, true, float64(priority)}, nil
+	return &addon.AddOnConfig{Name: addonName, Enabled: true, Priority: float64(priority)}, nil
```
praveenkumar: This changed happen to make `go vet` happy.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -164,7 +164,7 @@ func (m *AddOnManager) Disable(addonName string) (*addon.AddOnConfig, error) {
 
 	addOn.SetEnabled(false)
 
-	return &addon.AddOnConfig{addonName, false, float64(addOn.GetPriority())}, nil
+	return &addon.AddOnConfig{Name: addonName, Enabled: false, Priority: float64(addOn.GetPriority())}, nil
```
praveenkumar: This changed happen to make `go vet` happy.

cmd/minishift/cmd/start_test.go
hferentschik: Why would that only happen now? Are you running a different version of Go?

cmd/minishift/cmd/util/util.go
```diff
@@ -97,3 +98,9 @@ func IsValidProfile(profileName string) bool {
 	}
 	return false
 }
+
+// IsValidProfileName return true if profile name follow ^[a-zA-Z]+[0-9]*$
+func IsValidProfileName(profileName string) bool {
+	rr := regexp.MustCompile(`^[a-zA-Z]+[0-9]*$`)
```
hferentschik: I thought we generally allow alpha-numeric. What's the reason for enforcing that the profile name starts with letters?

cmd/minishift/cmd/util/util_test.go
```diff
@@ -0,0 +1,40 @@
...
+
+func TestIsValidProfileName(t *testing.T) {
+	testData := []struct {
+		profileName string
+		expected    bool
+	}{{"test", true},
+		{".", false},
+		{"@foo", false},
+		{"foo123", true},
+		{"123foo", false},
```
hferentschik: Why is this not ok?

cmd/minishift/cmd/profile/profile.go
```diff
@@ -19,13 +19,15 @@ package profile
...
 
+	cmdUtil "github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
 )
 
 const (
 	emptyProfileMessage  = "A profile name must be provided. Run `minishift profile list` for a list of existing profiles."
 	extraArgumentMessage = "You have provided more arguments than required. You must provide a single profile name."
+	invalidNameMessage   = "Profile name only allow alphanumeric characters."
```
hferentschik: "Profile name only allow alphanumeric characters." -> "Profile names must consist of alphanumeric characters only."

This does not match, however, the implementation, which is that the profile name is alphanumeric AND starts with at least one [:alpha:] character. 

We also need to update the docs and add at least a note there.

cmd/minishift/cmd/util/util.go
hferentschik: you don't like character classes like {:alpha:] and [:alnum:]?

Instead of repeating the regexp in the docs, I rather state the intention in words.

hferentschik:
@praveenkumar Please fix the commit message as well

cmd/minishift/cmd/util/util.go
hferentschik: Why is '-' or '_' not allowed? It is nice to at least be able to use some form of "separator" to make names more readable. IMO [:word:] is the character class I would match against, without any exception.

cmd/minishift/cmd/util/util_test.go
LalatenduMohanty: This my concern to me as well. I believe `123foo` is a valid profilename e.g. `1stjavaproject`

cmd/minishift/cmd/util/util.go
LalatenduMohanty: > Why is '-' or '_' not allowed? It is nice to at least be able to use some form of "separator" to make names more readable. IMO [:word:] is the character class I would match against, without any exception.

Agree. Can we cross check this with libvirt, xhyve and Virtualbox? I do not think HyperV is an issue as it uses uuid for VM name.

cmd/minishift/cmd/start_test.go
praveenkumar: @hferentschik I am using 1.8.3 only.

```
$ go version
go version go1.8.3 linux/amd64
```

gbraad:
libvirt does not like non-alpha-numeric in the name field of the Domain XML: https://libvirt.org/formatdomain.html#elementsMetadata

gbraad:
Can we not put the `vet`-satisfying changes in a separate commit, as at the moment it taints the actual change.

cmd/minishift/cmd/start_test.go
hferentschik: So why does that not appear until now? AFAICT the vetting target is enabled for a while and you are effectively changing some unrelated code.

cmd/minishift/cmd/util/util.go
```diff
@@ -97,3 +98,9 @@ func IsValidProfile(profileName string) bool {
 	}
 	return false
 }
+
+// IsValidProfileName return true if profile name follow ^[a-zA-Z]+[0-9]*$
```
hferentschik: the docs are not in sync with the impl

cmd/minishift/cmd/util/util.go
```diff
@@ -97,3 +98,9 @@ func IsValidProfile(profileName string) bool {
...
+func IsValidProfileName(profileName string) bool {
+	rr := regexp.MustCompile(`^\w+$`)
```
hferentschik: Can we not att the '-' as well? Then we won't have to make so many changes to the docs either.

jorgemoralespou: Yes, please. the "-" is something very needed in profile names.

praveenkumar:
I removed `go vet` changes, will send it to a different PR which will address the same issue in PR which @hferentschik worked #1688.

cmd/minishift/cmd/util/util_test.go
```diff
@@ -0,0 +1,42 @@
...
+		expected    bool
+	}{{"test", true},
+		{".", false},
+		{"@foo", false},
+		{"foo123", true},
+		{"123foo", true},
+		{"foo_123", true},
+		{"foo@123", false},
+		{"", false},
+		{"foo-123", true}}
```
LalatenduMohanty: We need to add `--foo` or `-foo123` as invalid profile names as `-` should be only used as separator.

cmd/minishift/cmd/profile/profile.go
```diff
@@ -19,13 +19,15 @@ package profile
...
 
+	cmdUtil "github.com/minishift/minishift/cmd/minishift/cmd/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/cobra"
 )
 
 const (
 	emptyProfileMessage  = "A profile name must be provided. Run `minishift profile list` for a list of existing profiles."
 	extraArgumentMessage = "You have provided more arguments than required. You must provide a single profile name."
+	invalidNameMessage   = "Profile names must consist of alphanumeric characters only."
```
LalatenduMohanty: `Profile names` or `Profile name` ? as we are printing the error for one profile name

```
$ minishift profile set *hello
Profile names must consist of alphanumeric characters only.
```
@robin-owen WDYT?

LalatenduMohanty: @minishift/minishift-dev any comments?

anjannath: both works fine in this context IMO. using `names` generalises it that all profiles should be alpha numeric and `name` would mean the profile name supplied to the command. both are correct statements. if we want to use `name` then ` The profile name must consist of alphanumeric characters only`

LalatenduMohanty:
I did a mea culpa. The docs mention `Profile names must consist of alphanumeric characters only` which should have been `Profile names can consist of alphanumeric characters and "-" only`

Issue #1503, opened 2017-10-05
Improvement: use a more generic solution for the VMDriver usage


gbraad:
In #1316 (settings_windows.go) I do:

```go
func ConfigureNetworking(machineName string, networkSettings NetworkSettings) {
	// Instruct the user that this does not work for other Hypervisors on Windows
	if viper.GetString("vm-driver") != "hyperv" {
		fmt.Println(configureIPAddressMessage, configureIPAddressFailure)
		return
	}
```

this means we need to rely on `spf13/viper`. In #1288 it was suggested to find an alternative way to handle this situation more elegantly. Something like:

`minishiftConfig.VMDriver` or `minishiftConfig.IsHyperV`

brian-avery:
@gbraad If I understand correctly, this issue is purely about creating a constant for the VMDriver rather than relying on reading the config each time, correct? I'd like to take this one.

gbraad:
Either in code we have to consider an abstraction, to prevent reading the configuration each time, or consider something that sets a configuration for us in the form of booleans `IsHyperV`, `IsVirtualBox`,

Haven't looked into it too much to decide what is a nicer approach.

budhrg:
Resolved via #1728 1728

gbraad:
Reopening, as the solution asked for is not implemented

LalatenduMohanty:
@gbraad I see that you have added the code as part of #1457 in #1849 . I am in the opinion of closing the issue as we had merged a code which addresses the issue in a way and #1457 is an further improvement.

LalatenduMohanty:
Also we should be conservative re-opening an existing issue if it is closed because of an code change as it breaks our git commit history.

Issue #1457, opened 2017-09-28
Follow-up feature to assign/use the IP address previously used: set IP address as static


gbraad:
With #1316 and https://github.com/minishift/minishift-centos-iso/pull/164 + https://github.com/minishift/minishift-centos-iso/pull/168 it is possible to assign an IP address for re-use when the virtual machine starts.

This task will allow the user to force the current address to become `static`, so on restart no new address will be assigned.

### Needed tasks

  * [x] allow to assign address to multiple interfaces  
         `DEVICE=[interface]` using multiple files named `/var/lib/networking?` => https://github.com/minishift/minishift-centos-iso/issues/180
  * [x] allow interface to be configured to fallback to dhcp  
         `dhclient [interface]` => added a `USEDHCP` option
  * [x] #1496 feature detection during provisioning; 'minishift-set-ipaddress'
  * [x] #1503 use a more generic solution for the VMDriver usage

### Needs investigation

  * [ ] #1765 set IP address for VirtualBox over the forwarded SSH port

---

**Note**: You need at least ISO: http://artifacts.ci.centos.org/minishift/minishift-centos-iso/master/39/ or higher 

This ISO will also allow an address to be 'fixed' by creating a file called: `/var/lib/minishift/networking` containing something like:

```ini
DEVICE=eth0
IPADDR=10.0.75.128
NETMASK=24
GATEWAY=10.0.75.1
DNS1=8.8.8.8
DNS2=8.8.4.4
```

However, this is currently only implemented for use with Hyper-V. Something like the following should be considered to make this work for the other hypervisors.

  * [x] Allow `minishift ip --static` to lock the IP address that has been assigned on startup
  * [ ] Allow `minishift ip --dhcp` to unlock the IP address to be assigned by DHCP
  * [ ] Use `minishift start --network-ipaddress ...` to set the secondary address, which on provisioning is available as secondary and on reboot will be used.

### Overview

| Hypervisor  | Device           | Remark           |
| --- | --- | --- |
| HyperV | `eth0` | attached to whatever vswitch you choose |
| KVM | `eth1` | `eth0` is attached to `default` and `eth1` to `docker-machines` (both on DHCP) |
| VirtualBox | `eth1` | `eth0` used for local communication; assigned `10.0.2.15` by DHCP |
| xhyve | `eth0` | |

budhrg:
> Allow minishift ip --static to lock the IP address that has been assigned on startup

[...]

PR #1728, opened 2017-11-21
Issue #1503 Use vmdriver variable instead of reading from config

brian-avery:
startHost is already reading the VM driver from the config.  We can pass this value in directly rather than reading it out of the config a second time.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -348,7 +348,7 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 
 		// Configure networking on startup only works on Hyper-V
 		if networkSettings.IPAddress != "" {
-			minishiftNetwork.ConfigureNetworking(constants.MachineName, networkSettings)
+			minishiftNetwork.ConfigureNetworking(constants.MachineName, machineConfig.VMDriver, networkSettings)
 		}
 	}
 
--- a/pkg/minishift/network/settings_windows.go
+++ b/pkg/minishift/network/settings_windows.go
@@ -25,7 +25,6 @@ import (
 	hvkvp "github.com/gbraad/go-hvkvp"
 	"github.com/golang/glog"
 	"github.com/minishift/minishift/pkg/minishift/shell/powershell"
-	"github.com/spf13/viper"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


praveenkumar:
@brian-avery Thanks for your contribution can you please squash the commit message and also there is typo in commit message "of reading frmo config" => "of reading from config"

budhrg:
add to whitelist

budhrg:
@brian-avery , Thanks for the contributions. Changes looks good to me. Let's squash the commits and fix typo. Then, I believe we are good to merge.

cmd/minishift/cmd/start.go
```diff
@@ -342,7 +342,7 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 
 		// Configure networking on startup only works on Hyper-V
 		if networkSettings.IPAddress != "" {
-			minishiftNetwork.ConfigureNetworking(constants.MachineName, networkSettings)
+			minishiftNetwork.ConfigureNetworking(constants.MachineName, machineConfig.VMDriver, networkSettings)
```
praveenkumar: 👍

pkg/minishift/network/settings_windows.go
```diff
@@ -34,9 +33,9 @@ const (
 	networkingMessageName = "PROVISION_NETWORKING"
 )
 
-func ConfigureNetworking(machineName string, networkSettings NetworkSettings) {
+func ConfigureNetworking(machineName string, vmDriver string, networkSettings NetworkSettings) {
 	// Instruct the user that this does not work for other Hypervisors on Windows
-	if viper.GetString("vm-driver") != "hyperv" {
+	if vmDriver != "hyperv" {
```
praveenkumar: 👍

brian-avery:
Thanks. Squash should be complete now. Let me know how it looks.

praveenkumar:
@brian-avery Thanks, looks good. Let CI  test runs and then we will merge it.

budhrg:
LGTM as well :+1:

pkg/minishift/network/settings_windows.go
gbraad: I am not a huge fan of string-based checks. Especially as all the follow-up code has to check this to decide on hypervisor specific solutions. Will address this in #1457

PR #1731, opened 2017-11-21
Issue #1695 Restrict profile set and start command to create profiles

### LINKED ISSUE #1695: New profile xyz is being created on passing the command "./minishift --profile xyz status"
```Minishift version : 1.8.0/master```
The command ```./minishift --profile xyz status``` creating the profile xyz rather throwing the exception something like ```Profile is not created. Create the profile first```
```
$ ./minishift profile list
- foo        Running        (Active)
- minishift    Running

$ ./minishift --profile xyz status
Does Not Exist

$ ./minishift profile list
- foo        Running        (Active)
- minishift    Running
- xyz        Does Not Exist
$
```

### PR DESCRIPTION
**Pull request by praveenkumar**
Fix #1695

### DIFF
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -52,8 +52,11 @@ import (
 
 const (
 	showLibmachineLogs    = "show-libmachine-logs"
+	profileCmd            = "profile"
 	profileFlag           = "profile"
+	profileSetCmd         = "set"
 	enableExperimentalEnv = "MINISHIFT_ENABLE_EXPERIMENTAL"
+	invalidProfileName    = "Profile names must consist of alphanumeric characters only."
 )
 
 var viperWhiteList = []string{
@@ -72,6 +75,8 @@ var RootCmd = &cobra.Command{
 			err                    error
 			isAddonInstallRequired bool
 		)
+
+		checkForValidProfileOrExit(cmd)
 
 		constants.MachineName = constants.ProfileName
 		constants.Minipath = constants.GetProfileHomeDir(constants.ProfileName)
@@ -206,18 +211,47 @@ func initConfig() {
 	setupViper()
 }
 
+// initializeProfile always return profile name based on below checks.
+// 1. If profile set <PROFILE_NAME> is used then return PROFILE_NAME
+// 2. If --profile <PROFILE_NAME> then return PROFILE_NAME
+// 3. If no profile command or flag then return active profile name.
 func initializeProfile() string {
 	var (
-		profileName   string
-		err           error
-		activeProfile string
+		profileName     string
+		err             error
+		activeProfile   string
+		profileCmdAlias = []string{
+			"profiles",
+			"instance",
+		}
 	)
 
+	// Check if profileCmd is part of os.Args so that it takes preference instead `--profile` argument
+	var isProfileCmdUsed bool
+	for _, arg := range os.Args {
+		if arg == profileCmd || arg == profileCmdAlias[0] || arg == profileCmdAlias[1] {
+			isProfileCmdUsed = true
+		}
+	}
+
 	for i, arg := range os.Args {
-		if arg == "--"+profileFlag {
-			profileName = os.Args[i+1]
-			if !cmdUtil.IsValidProfileName(profileName) {
-				atexit.ExitWithMessage(1, "Profile names must consist of alphanumeric characters only.")
+		if !isProfileCmdUsed {
+			// This will match if `--profile` flag is used
+			if arg == "--"+profileFlag {
+				profileName = os.Args[i+1]
+				break
+			}
+		}
+		// This will match if we used profile or it's alias commands
+		if arg == profileCmd || arg == profileCmdAlias[0] || arg == profileCmdAlias[1] {
+			// This make sure if user specify profile command without any subcommand then
+			// it should not panic with out of index error
+			if len(os.Args) <= i+2 {
+				break
+			}
+			// For use cases when minishift profile set PROFILE_NAME is used
+			if os.Args[i+1] == profileSetCmd {
+				profileName = os.Args[i+2]
 			}
 			break
 		}
@@ -330,6 +364,22 @@ func setDefaultActiveProfile() {
 		// Otherwise minishift will be the active profile irrespective of what user chooses
 		if constants.ProfileName == constants.DefaultProfileName {
 			cmdUtil.SetOcContext(constants.DefaultProfileName)
+		}
+	}
+}
+
+// checkForValidProfileOrExit checks if a profile exist or not when --profile flag used.
+// If profile not exist then it will error out with message.
+func checkForValidProfileOrExit(cmd *cobra.Command) {
+	if !cmdUtil.IsValidProfileName(constants.ProfileName) {
+		atexit.ExitWithMessage(1, invalidProfileName)
+	}
+	if cmd.Parent() != nil {
+		// This condition true for each command execpt `minishift profile <subcommand>` and `minishift start ...``
+		if cmd.Parent().Name() != profileCmd && cmd.Name() != startCmd.Name() {
+			if !cmdUtil.IsValidProfile(constants.ProfileName) {
+				atexit.ExitWithMessage(1, fmt.Sprintf("Profile: %s doesn't exist, Use `minishift profile set %s` or `minishift start --profile %s` to create", constants.ProfileName, constants.ProfileName, constants.ProfileName))
+			}
 		}
 	}
 }
--- a/test/integration/features/profile.feature
+++ b/test/integration/features/profile.feature
@@ -11,8 +11,9 @@ Feature: Profile
 
   Examples: Empty profile name
     | profilename |
+    |             |
     | ''          |
-    | ' '         |
+
     
   Scenario Outline: As user, I cannot create profile with special character in profile name
      When executing "minishift profile set <profilename>" fails
@@ -23,6 +24,7 @@ Feature: Profile
 
   Examples: Wrong profile names
     | profilename |
+    | ' '         |
     | '.'         |
     | '-'         |
     | '#$'        |
@@ -160,3 +162,29 @@ Feature: Profile
      Then Minishift should have state "Does Not Exist"
      When executing "minishift ip"
      Then exitcode should equal "1"
+
+  Scenario: As user, I cannot create profile when used along with profile <subcommand> --profile name
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift profile list --profile foo" succeeds
+     Then stdout should not contain
+       """
+       foo
+       """
+
+  Scenario Outline: Test all minishift commands along with --profile flag to make sure it doesn't create
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift <commands> --profile foo"
+     Then exitcode should equal "1"
+      And stderr should contain
+       """
+       Profile: foo doesn't exist, Use `minishift profile set foo` or `minishift start --profile foo` to create
+       """
+  Examples: minishift commands
+    | commands  |
+    | ip        |
+    | status    |
+    | delete    |
+    | logs      |
+    | stop      |
+    | version   |
+    
--- a/cmd/minishift/cmd/root_test.go
+++ b/cmd/minishift/cmd/root_test.go
@@ -134,3 +134,27 @@ func TestViperAndFlags(t *testing.T) {
 		cli.UnsetValues(testOption)
 	}
 }
+
+func TestInitializeProfile(t *testing.T) {
+	oldArgs := os.Args
+	defer func() { os.Args = oldArgs }()
+	Testdata := []struct {
+		Args   []string
+		Result string
+	}{
+		{[]string{"minishift", "profile", "set", "hello"}, "hello"},
+		{[]string{"minishift", "profiles", "set", "hello"}, "hello"},
+		{[]string{"minishift", "instance", "set", "hello"}, "hello"},
+		{[]string{"minishift", "start", "--profile", "hello"}, "hello"},
+		{[]string{"minishift", "ip", "--profile", "hello"}, "hello"},
+		{[]string{"minishift", "status", "--profile", "hello"}, "hello"},
+	}
+
+	for _, testInput := range Testdata {
+		os.Args = testInput.Args
+		got := initializeProfile()
+		if got != testInput.Result {
+			t.Errorf("Got %s, Expected %s as profile Name", got, testInput.Result)
+		}
+	}
+}
--- a/docs/source/using/profiles.adoc
+++ b/docs/source/using/profiles.adoc
@@ -44,7 +44,7 @@ The use of underscore ( *_* ) and hyphen ( *-* ) are allowed as separator.
 
 === Using the `--profile` Flag
 
-When you run any {project} command with the `--profile` flag the profile gets created if it does not exist, for example:
+When you run {project} `start` command with the `--profile` flag the profile gets created if it does not exist, for example:
 
 ----
 $ minishift --profile profile-demo start

### CODE REVIEW

**Comment by gbraad** [comment]
Are you currently working on this?

**Comment by praveenkumar** [comment]
@gbraad just waiting for #1408 to be in and then I will resume the work.

**Comment by praveenkumar** [comment]
@minishift/minishift-dev now `m start --profile` and `m profile set` only going to create the profile structure.

**Comment by gbraad** [comment]
#1408 got merged, ...

**Comment by praveenkumar** [comment]
@gbraad rebased :-s

File: cmd/minishift/cmd/root.go [hunk 1]
```diff
@@ -333,3 +360,29 @@ func setDefaultActiveProfile() {
 		}
 	}
 }
+
+// validateProfile check if a profile exist or not when used --profile flag
```
budhrg: validateProfile checks if a profile exist or not when --profile flag is used.

cmd/minishift/cmd/root.go
```diff
@@ -333,3 +360,29 @@ func setDefaultActiveProfile() {
...
+				if arg == "--"+profileFlag {
+					profileName = os.Args[i+1]
+					break
+				}
+			}
+			if profileName != "" {
+				if !cmdUtil.IsValidProfile(profileName) {
+					atexit.ExitWithMessage(1, fmt.Sprintf("Profile: %s doesn't exist, Use *minishift profile set %s* to create", profileName, profileName))
+				}
+				constants.ProfileName = profileName
```
budhrg: The function `validateProfile` is not reflecting this step. If I don't look into the code of validateProfile then, I will not know where my profile is getting set.
I would highly recommend to move this line with other similar line `constants.MachineName = constants.ProfileName`.

cmd/minishift/cmd/root.go
```diff
@@ -333,3 +365,29 @@ func setDefaultActiveProfile() {
 		}
 	}
 }
+
+// isProfileExist checks if a profile exist or not when --profile flag used.
+// If profile exist then return ProfileName provided by --profile option.
+func isProfileExist(cmd *cobra.Command) string {
```
budhrg: :+1:

cmd/minishift/cmd/root.go
```diff
@@ -73,6 +75,13 @@ var RootCmd = &cobra.Command{
 			isAddonInstallRequired bool
 		)
 
+		// If profileName is not an empty string then we need to update
+		// constants.ProfileName to make sure it will create requested directory structure.
+		profileName := isProfileExist(cmd)
```
budhrg: Much better! Reads quite well. Thanks.

cmd/minishift/cmd/root.go
```diff
@@ -73,6 +75,13 @@ var RootCmd = &cobra.Command{
...
+		if profileName != "" {
+			constants.ProfileName = profileName
+		}
```
LalatenduMohanty: I do not think we need to assign `constants.ProfileName ` again. Because we do it in initconfig() and by the time the control flow reaches here it would have exited in isProfileExist if the profile does not exist.

cmd/minishift/cmd/root.go
```diff
@@ -213,11 +222,34 @@ func initializeProfile() string {
...
+		if arg == profileFlag {
+			// This make sure if user specify *profile* flag then
+			// it should not panic with out of index error
+			if len(os.Args) <= i+2 {
+				break
+			}
+			// This will match if after *profile* we used *set* command
+			// Then it will have profileName as whatever we provided as set command parameter.
+			if os.Args[i+1] == profileSetFlag {
+				profileName = os.Args[i+2]
```
LalatenduMohanty: We have the same code in initializeProfile() i.e. 
```
if os.Args[i+1] == profileSetFlag {
				profileName = os.Args[i+2]
```
We should create a function for this and use it both the places.

cmd/minishift/cmd/root.go
praveenkumar: >I do not think we need to return profileName from isProfileExist(cmd). Because at this stage constant.ProfileName would have the correct profile name. The only thing we need to do is, we need to check if --profile PROFILE_NAME is a valid profile (with directory structure).

@LalatenduMohanty No at this point `constant.ProfileName` only have either active profile info or default profile. It doesn't have any info about `--profile` flag parameter until we set it.

> I understand that we need to do it in PersistentPreRun because we want to take help of Viper to see if it is part of minishift start --profile.
In initializeProfile(), the --profile takes priority than active profile. So it will always return the correct profile name. If it returns null then constants.ProfileName would have default profile name.

@LalatenduMohanty Sorry, here you already contradicting your first argument. You are absolutely right but to get it we return the parameter which is part of profile flag.

> isProfileExist sounds like a function which returns boolean value. Also the name does not communicate the intent of the function.

Agree, might be `setExistProfile` => return profile name if exist?

cmd/minishift/cmd/root.go
praveenkumar: @LalatenduMohanty No, what if profile exist? Then flow will be following.

Assume profile `foo` exist and not as active profile.

```
$ minishift --profile list
foo 
minishift (active)

$ minishift status --profile foo
initconfig() will set the constants.ProfileName to minishift because it's active.
isProfileExist() validate that --profile flag is used so check if valid if yes then return so set back constants.Profile as foo to make sure now all the operation happen on this profile.
```

cmd/minishift/cmd/root.go
praveenkumar: @LalatenduMohanty This is still `initializeProfile()` function code only where you think this logic is used again?

cmd/minishift/cmd/root.go
```diff
@@ -333,3 +361,16 @@ func setDefaultActiveProfile() {
...
 }
+
+// checkProfileExist checks if a profile exist or not when --profile flag used.
+// If profile not exist then it will error out with message.
+func checkProfileExist(cmd *cobra.Command) {
+	if cmd.Parent() != nil {
+		// This condition true for each command execpt *m profile* and *m start ...*
+		if cmd.Parent().Name() != profileFlag && cmd.Name() != startFlag {
+			if !cmdUtil.IsValidProfile(constants.ProfileName) {
+				atexit.ExitWithMessage(1, fmt.Sprintf("Profile: %s doesn't exist, Use *minishift profile set %s* to create", constants.ProfileName, constants.ProfileName))
```
LalatenduMohanty: Please update the error message to include `minishift start --profile PROFILE_NAME`. At this point we are only saying `minishift profile set`. E.g. `Use minishift start --profile or minishift profile set ` to create profile.

cmd/minishift/cmd/root.go
```diff
@@ -73,6 +75,8 @@ var RootCmd = &cobra.Command{
 			isAddonInstallRequired bool
 		)
 
+		checkProfileExist(cmd)
```
LalatenduMohanty: I think validateProfileExists() would be a better name.

praveenkumar: @LalatenduMohanty For me `checkProfileExist` make more sense because we are just checking if profile exist or not but when we say `validateProfileExists` then what we validating in profile?

cmd/minishift/cmd/root.go
```diff
@@ -333,3 +361,16 @@ func setDefaultActiveProfile() {
 		}
 	}
 }
+
+// checkProfileExist checks if a profile exist or not when --profile flag used.
+// If profile not exist then it will error out with message.
+func checkProfileExists(cmd *cobra.Command) {
+	if cmd.Parent() != nil {
+		// This condition true for each command execpt *m profile* and *m start ...*
+		if cmd.Parent().Name() != profileFlag && cmd.Name() != startFlag {
```
gbraad: +1

cmd/minishift/cmd/root.go
```diff
@@ -333,3 +361,16 @@ func setDefaultActiveProfile() {
 		}
 	}
 }
+
+// checkProfileExist checks if a profile exist or not when --profile flag used.
+// If profile not exist then it will error out with message.
+func checkProfileExists(cmd *cobra.Command) {
```
gbraad: checkForValidProfileOrExit

cmd/minishift/cmd/root.go
LalatenduMohanty: @praveenkumar nvm.

LalatenduMohanty:
Seems like it is not working for `profile` commands
```
$ minishift profile list
- hello		Does Not Exist
- minishift	Does Not Exist	(Active)
```
```
$ minishift profile list --profile xyz
- hello		Does Not Exist
- minishift	Does Not Exist	(Active)
- xyz		Does Not Exist
```

cmd/minishift/cmd/root.go
```diff
@@ -213,14 +217,38 @@ func initializeProfile() string {
...
+				break
+			}
 			profileName = os.Args[i+1]
 			if !cmdUtil.IsValidProfileName(profileName) {
 				atexit.ExitWithMessage(1, "Profile names must consist of alphanumeric characters only.")
 			}
 			break
 		}
+		// This will match if we used *profile* command
+		if arg == profileFlag {
```
LalatenduMohanty: This would introduce bugs for the aliases for `minishift profile` command i.e. `instance` and `profiles`.

LalatenduMohanty:
retest this please

LalatenduMohanty:
@praveenkumar I would highly recommend adding an integration test for a negative test case e.g.  https://github.com/minishift/minishift/pull/1731#issuecomment-350729853

LalatenduMohanty:
While testing the PR I saw 
```
 minishift status --profile hello
Profile: hello doesn't exist, Use *minishift profile set hello* or *minishift start --profile hello* to create
```
You can see the `*` in the output of the command to provide differentiation that these are commands. I checked in existing Minishift code and found below

```
test/integration/features/cmd-openshift.feature:      Run 'oc status' to view your app.
cmd/minishift/cmd/addon/addon.go:       emptyAddOnError      = "You must specify an add-on name. Run `minishift addons list` to view installed add-ons."
cmd/minishift/cmd/addon/addon_disable.go:       emptyDisableError           = "You must specify an add-on name. Run `minishift addons list` to view installed add-ons."
test/integration/features/profile.feature:      A profile name must be provided. Run `minishift profile list` for a list of existing profiles.
```
As you can see we are consistent about how we represent the command in the CLI output, which is bad. IMO I feel commands should be inside back ticks e.g. \`minishift profile list\` . @minishift/minishift-dev  WDYT?

budhrg:
> IMO I feel commands should be inside back ticks e.g. `minishift profile list`

Why? Back tick form is for rendering command in markdown format. 
I am fine with single quote.

LalatenduMohanty:
> Why? Back tick form is for rendering command in markdown format.
> I am fine with single quote.

Single quotes are already used for giving importance to particular words

```
$ minishift profile set hello
Profile 'hello' set as active profile.

$ minishift docker-env
Running this command requires an existing 'minishift' VM, but no VM is defined.
```

budhrg:
> Single quotes are already used for giving importance to particular words

For displaying command as well:

```
$ minishift stop -h
Stops the running local OpenShift cluster. This command stops the Minishift
VM but does not delete any associated files. To start the cluster again, use the 'minishift start' command.
```

LalatenduMohanty:
> For displaying command as well:

So we are not consistent. Lets decide what we want to use and then fix the existing ones.

budhrg:
> So we are not consistent. Lets decide what we want to use and then fix the existing ones.

Created issue #1833

@praveenkumar , plz update as per your preference or  wait till #1833 is fixed :)

cmd/minishift/cmd/root.go
```diff
@@ -333,3 +366,16 @@ func setDefaultActiveProfile() {
 		}
 	}
 }
+
+// checkForValidProfileOrExit checks if a profile exist or not when --profile flag used.
+// If profile not exist then it will error out with message.
+func checkForValidProfileOrExit(cmd *cobra.Command) {
+	if cmd.Parent() != nil {
+		// This condition true for each command execpt `minishift profile <subcommand>` and `minishift start ...``
+		if cmd.Parent().Name() != profileFlag && cmd.Name() != startFlag {
```
LalatenduMohanty: We need to take profile aliases in to account i.e. instance, profiles.

cmd/minishift/cmd/root.go
```diff
@@ -333,3 +366,16 @@ func setDefaultActiveProfile() {
 		}
 	}
 }
+
+// checkForValidProfileOrExit checks if a profile exist or not when --profile flag used.
+// If profile not exist then it will error out with message.
+func checkForValidProfileOrExit(cmd *cobra.Command) {
+	if cmd.Parent() != nil {
+		// This condition true for each command execpt `minishift profile <subcommand>` and `minishift start ...``
```
LalatenduMohanty: s/condition/condition will be/

cmd/minishift/cmd/root.go
```diff
@@ -333,3 +366,16 @@ func setDefaultActiveProfile() {
 		}
 	}
 }
+
+// checkForValidProfileOrExit checks if a profile exist or not when --profile flag used.
+// If profile not exist then it will error out with message.
```
LalatenduMohanty: s/not/does not/

cmd/minishift/cmd/root.go
```diff
@@ -213,14 +222,38 @@ func initializeProfile() string {
...
 		}
+		// This will match if we used profile command
+		if arg == profileFlag || arg == profileAlias[0] || arg == profileAlias[1] {
+			// This make sure if user specify profile command without any subcommand then
+			// it should not panic with out of index error
+			if len(os.Args) <= i+2 {
+				break
+			}
+			// This will match if after profile command we used set subcommand
+			// Then it will have profileName as whatever we provided as set command parameter.
```
LalatenduMohanty: This comment is not easy to understand. Suggestion, For use cases when `--profile PROFILE_NAME` is used with `minishift profile set` command.

test/integration/features/profile.feature
```diff
@@ -160,3 +162,28 @@ Feature: Profile
...
+  Scenario: As user, I cannot create profile when used along with profile <subcommand> --profile name
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift profile list --profile foo" succeeds
+     Then stdout should not contain
+        """
+        foo
+        """
+
+  Scenario Outline: Test all minishift commands along with --profile flag to make sure it doesn't create
+    Given Minishift has state "Does Not Exist"
```
LalatenduMohanty: 👍 . Awesome.

test/integration/features/profile.feature
```diff
@@ -160,3 +162,28 @@ Feature: Profile
...
+      Profile: foo doesn't exist, Use `minishift profile set foo` or `minishift start --profile foo` to create
+      """
+  Examples: minishift commands
+    | commands  |
+    | ip        |
+    | status    |
+    | delete    |
+    | logs      |
+    | stop      |
+    | version   |
```
LalatenduMohanty: @budhrg @agajdosi asked a empty line here for one of my PRs. Not sure why.

gbraad:
I only use backticks in commands when it is markdown or IRC
In our fedback to the use I rather see single-quotes to indicate a command

test/integration/features/profile.feature
budhrg: Yes, when we see below mark in Github, we need empty line.
screenshot from 2017-12-12 15-15-02

See reason https://stackoverflow.com/questions/5813311/no-newline-at-end-of-file

agajdosi: @LalatenduMohanty I am sorry, my mistake, most of our features and features on other projects do not have a new line at the end, for sure it has no effect on actual functionality

test/integration/features/profile.feature
```diff
@@ -160,3 +162,28 @@ Feature: Profile
      Then Minishift should have state "Does Not Exist"
      When executing "minishift ip"
      Then exitcode should equal "1"
+
+  Scenario: As user, I cannot create profile when used along with profile <subcommand> --profile name
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift profile list --profile foo" succeeds
+     Then stdout should not contain
```
amitkrout: Then stdout should not contain
  """
  foo
  """

amitkrout: @praveenkumar Please refer the below link and do the indentation accordingly 

https://github.com/minishift/minishift/blob/master/test/integration/features/profile.feature#L19

test/integration/features/profile.feature
```diff
@@ -160,3 +162,28 @@ Feature: Profile
...
+        """
+        foo
+        """
+
+  Scenario Outline: Test all minishift commands along with --profile flag to make sure it doesn't create
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift <commands> --profile foo"
+     Then exitcode should equal "1"
+      And stderr should contain
+      """
```
amitkrout: prepended by a space

test/integration/features/profile.feature
```diff
@@ -160,3 +162,28 @@ Feature: Profile
      Then Minishift should have state "Does Not Exist"
      When executing "minishift ip"
      Then exitcode should equal "1"
+
+  Scenario: As user, I cannot create profile when used along with profile <subcommand> --profile name
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift profile list --profile foo" succeeds
+     Then stdout should not contain
+        """
```
amitkrout: @praveenkumar Please refer the below link and do the indentation accordingly 

https://github.com/minishift/minishift/blob/master/test/integration/features/profile.feature#L19

LalatenduMohanty:
> I only use backticks in commands when it is markdown or IRC
> In our feedback to the user I rather see single-quotes to indicate a command

@gbraad lets discuss that in #1833

gbraad:
Any movement on this? it seems the last issue is with a newline in `profile-feature` ? => https://github.com/minishift/minishift/pull/1731#discussion_r156315335

cmd/minishift/cmd/root.go
budhrg: Ya forgot about it.

praveenkumar: @LalatenduMohanty We don't need it here because this handled in the `initializeProfile()` method. Do try once and let me know if you still think it needed.

```
$ minishift instance/profiles set/list/delete 

=> then checkForValidProfileOrExit() function will check first it profile flag which is not then it will check IsvalidProfile which either doesn't exist and it exit then return.
```

cmd/minishift/cmd/root.go
```diff
@@ -213,14 +222,37 @@ func initializeProfile() string {
...
 	)
 
+	// If `minishift profile set` or `minishift start --profile` command is used.
 	for i, arg := range os.Args {
+		// This will match if `--profile` flag is used
 		if arg == "--"+profileFlag {
+			// This make sure if user specify `--profile` flag without any name
+			// it should not panic with out of index error
+			if len(os.Args) <= i+1 {
+				break
```
LalatenduMohanty: We do not need this check as Cobra/Viper takes care of this. In my testing I found this as dead code.

cmd/minishift/cmd/root.go
```diff
@@ -213,14 +222,37 @@ func initializeProfile() string {
...
 			}
 			break
 		}
+		// This will match if we used profile or it's alias commands
+		if arg == profileFlag || arg == profileAlias[0] || arg == profileAlias[1] {
+			// This make sure if user specify profile command without any subcommand then
+			// it should not panic with out of index error
+			if len(os.Args) <= i+2 {
+				break
+			}
```
LalatenduMohanty: We do not need this check as Cobra/Viper takes care of this. In my testing I found this as dead code.

cmd/minishift/cmd/root.go
```diff
@@ -213,14 +222,37 @@ func initializeProfile() string {
...
+			// For use cases when minishift profile set PROFILE_NAME is used
+			if os.Args[i+1] == profileSetFlag {
+				profileName = os.Args[i+2]
+				if !cmdUtil.IsValidProfileName(profileName) {
+					atexit.ExitWithMessage(1, "Profile names must consist of alphanumeric characters only.")
+				}
+			}
+			break
```
LalatenduMohanty: These code change is completely changing the behavior of initializeProfile which might not be intended.  Just take example of below commands
* minishift profile list --profile xyz
* minishift --profile xyz profile list
* minishift profile delete xyz 
InitilizeProfile behaves differently around these commands which is not expected IMO. 

Also it fails for `$ minishift --profile xyz profile list` i.e. it creates the `xyz` profile.

cmd/minishift/cmd/root.go
LalatenduMohanty: @praveenkumar we need to fix the logic of this function depending upon what we plan to do with initializeProfile function.

cmd/minishift/cmd/root.go
```diff
@@ -213,14 +222,37 @@ func initializeProfile() string {
 		activeProfile string
 	)
 
+	// If `minishift profile set` or `minishift start --profile` command is used.
```
LalatenduMohanty: We are not doing anything related to `start` command in this function, so the comment is confusing to me.

cmd/minishift/cmd/root.go
praveenkumar: @LalatenduMohanty  this is not dead code. Try to comment it out and then do `minishift profile` you will get out of index runtime panic.

cmd/minishift/cmd/root.go
praveenkumar: @LalatenduMohanty Good catch 👍 , I need to implement another flag to make sure if `profile` and `--profile` args are there then `profile` always take preference.

cmd/minishift/cmd/root.go
LalatenduMohanty: @praveenkumar you are right. This check is required in this place.

cmd/minishift/cmd/root.go
```diff
@@ -52,10 +52,18 @@ import (
 
 const (
 	showLibmachineLogs    = "show-libmachine-logs"
+	profileCmd            = "profile"
 	profileFlag           = "profile"
```
LalatenduMohanty: @praveenkumar I thought we agreed on `profileFlag           = "--profile"` . Any reason why we have not done that?

LalatenduMohanty: The issue is we can not have `profileCmd` and `profileFlg` with same value i.e. `profile` , it looks odd.

cmd/minishift/cmd/root.go
```diff
@@ -333,3 +368,19 @@ func setDefaultActiveProfile() {
...
+
+// checkForValidProfileOrExit checks if a profile exist or not when --profile flag used.
+// If profile not exist then it will error out with message.
+func checkForValidProfileOrExit(cmd *cobra.Command) {
+	if !cmdUtil.IsValidProfileName(constants.ProfileName) {
+		atexit.ExitWithMessage(1, invalidProfileName)
+	}
+	if cmd.Parent() != nil {
+		// This condition true for each command execpt `minishift profile <subcommand>` and `minishift start ...``
+		if cmd.Parent().Name() != profileCmd && cmd.Name() != startCmd.Name() {
```
LalatenduMohanty: 👍 , Viper/Cobra takes care of the aliases here.

praveenkumar:
> The issue is we can not have profileCmd and profileFlag with the same value i.e. profile, it looks odd.

@LalatenduMohanty Yes, but problem is if do `profileFlag` is used directly in the https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/root.go#L176 as a persistent flag.If we use  `profileFlag = "--profile"` then it will become `----profile` as global flags :(

cmd/minishift/cmd/root.go
```diff
@@ -52,10 +52,18 @@ import (
...
 	profileFlag           = "profile"
+	profileSetCmd         = "set"
 	enableExperimentalEnv = "MINISHIFT_ENABLE_EXPERIMENTAL"
+	invalidProfileName    = "Profile names must consist of alphanumeric characters only."
 )
 
+var profileCmdAlias = []string{
+	"profiles",
+	"instance",
+}
```
LalatenduMohanty: I think we should move this in to initializeProfile() as it is only used in it and we do not need these as global variables

LalatenduMohanty:
> @LalatenduMohanty Yes, but problem is if do profileFlag is used directly in the https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/root.go#L176 as a persistent flag.If we use profileFlag = "--profile" then it will become ----profile as global flags :(

That's unfortunate. I am not able to get any generic name for both profileFlag and profileCmd. 

The other alternative I see that we can have local variable `profileCmd` for initiliizeProfile and use `profileFlag` for checkForValidProfileOrExit. As in  checkForValidProfileOrExit. `if cmd.Parent().Name() != profileFlag` is readable and not confusing.

cmd/minishift/cmd/root.go
```diff
@@ -206,18 +216,43 @@ func initConfig() {
...
+// 2. If --profile <PROFILE_NAME> then return PROFILE_NAME
+// 3. If no profile command or flag then return active profile name.
 func initializeProfile() string {
 	var (
 		profileName   string
 		err           error
 		activeProfile string
 	)
 
+	// Check if profileFlag is part of os.Args so that it takes preference instead `--profile` argument
```
LalatenduMohanty: I think you mean profileCmd . If yes, please fix it.

cmd/minishift/cmd/root.go
gbraad: I do not think this is odd... writing the full flag in my opinion is, as this puts logic and an implementatiuon detail inside the flagname, while this should be generic logic.

gbraad:
+1

PR #1737, opened 2017-11-22
Issue #1729 Fix minishift architecture diagram rendering issue

### LINKED ISSUE #1729: Minishift architecture diagram is not rendering due to permission issue
Again getting permission issue for image rendering. It was fixed as part of https://github.com/minishift/minishift/pull/1700 but not sure what happened after latest sync.

Link - https://docs.openshift.org/latest/minishift/using/basic-usage.html

screenshot from 2017-11-21 11-26-18

budhrg:
Fix #1729

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -266,7 +266,7 @@ function perform_release() {
   MINISHIFT_VM_DRIVER=kvm make integration_all
   exit_on_failure "$?" "Integration tests failed."
 
-  make link_check_docs IMAGE_UID=$(id -u) # Test docs builds and all links are valid
+  make link_check_docs # Test docs builds and all links are valid
   exit_on_failure "$?" "Documentation build failed."
 
   create_release_commit
@@ -279,7 +279,7 @@ function perform_release() {
   add_release_notes;
   exit_on_failure "$?" "Failed to update release notes of Minishift v$RELEASE_VERSION. Try to manually update the release notes here - https://github.com/${REPO_OWNER}/minishift/releases/tag/v$RELEASE_VERSION."
 
-  make gen_adoc_tar IMAGE_UID=$(id -u)
+  make gen_adoc_tar
   exit_on_failure "$?" "Documentation tarball build failed."
 
   docs_tar_upload $1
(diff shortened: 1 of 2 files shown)


budhrg:
retest this please

budhrg:
Doc hosted from this PR http://artifacts.ci.centos.org/minishift/minishift/pr/1737/openshift-origin/latest/minishift/using/basic-usage.html

LalatenduMohanty:
@budhrg I feel we are fixing the cause the effect but not the root cause as we do not know why the presence of image directory caused the rendering to fail.

docs/Rakefile
```diff
@@ -4,6 +4,9 @@ require 'rubygems/package'
 task :default => :gen
 
 BUILD_DIR          = "build"
+DITAA_IMAGES       = [
```
hferentschik: -1 This does not feel right and it is also fragile.

hferentschik:
>  I feel we are fixing the cause the effect but not the root cause as we do not know why the presence of image directory causing the rendering to fail.

Exactly. Similar to the first round we had to "fix" this issue.

centos_ci.sh
```diff
@@ -279,7 +279,7 @@ function perform_release() {
   add_release_notes;
   exit_on_failure "$?" "Failed to update release notes of Minishift v$RELEASE_VERSION. Try to manually update the release notes here - https://github.com/${REPO_OWNER}/minishift/releases/tag/v$RELEASE_VERSION."
 
-  make gen_adoc_tar IMAGE_UID=$(id -u)
+  make gen_adoc_tar
```
hferentschik: +1, I thought this was already removed when we changed the image. Not use in this.

docs/Rakefile
hferentschik: We discussed this before. This is problematic out of two reasons:

* It requires you to know that there is a special case. What if we add a ditaa image into another topic? Then things will break again (provided that we are actually fixing a problem here which I doubt)
* Even more problematic is that all images are excluded. What if someone wants to add a static image to the topic. Locally it all would render fine, but the image would disappear when the docs would get build as part of the openshift.org build via the tarball.

hferentschik:
See pull request #1751. Let's close this pull request and merge #1751. The changes in #1751 make sense regardless it will fix the issue we see on the Jenkins build. 

@budhrg maybe you can test it before merging. 

It does not quite explain why the image cannot exist.

hferentschik:
Superseded by #1751.

Issue #1748, opened 2017-11-24
Unable to set profile name with single letter alphanumeric character


amitkrout:
As per the current changes profile name should be alphanumeric. However there is an issue of profile name having single char or a single digit number. In both the case minishift fails to set the profile.
Case 1:
```
Amits-MacBook-Pro:bin amit$ ./minishift profile set 'a'
Profile names must consist of alphanumeric characters only.
Amits-MacBook-Pro:bin amit$ ./minishift profile set 'aa'
Profile 'aa' set as active profile.
```
Case 2:
```
Amits-MacBook-Pro:bin amit$ ./minishift profile set '1'
Profile names must consist of alphanumeric characters only.
Amits-MacBook-Pro:bin amit$ ./minishift profile set '11'
Profile '11' set as active profile.

amitkrout:
#1692

LalatenduMohanty:
I think single alpha numeric character is fine with any hypervisor. So we should fix it asap.

LalatenduMohanty:
> We do need to check if _ is vaild VM name for different hypervisors otherwise we might force it to use minimum 2 characters.

You mean *-* right? not *_*. I do not think *-* should be allowed as a single character for profile name as *-* should be only used as a separator.

praveenkumar:
> You mean - right? not _. 

@LalatenduMohanty No, I meant `_` only because write now we are using `\w` which is word range and contain `_` also so if we allow single character that means `_` shouldn't have any issue.

LalatenduMohanty:
I am afraid `_` (underscore) is not accepted as a valid character in KVM/libvirt. I need to first cross check it.

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/pull/1749/

thisiskatewinsle:
Thanks for sharing the link I never read about it before. In return, I would like to share [nhs argos discount](https://www.voucherpro.co.uk/argos.co.uk ) and some other discount codes that will be beneficial for everyone visiting here. If you are planning to travel you can go through the [travelodge discount code student](https://www.voucherpro.co.uk/travelodge.co.uk), it is easy to avail.

jamescharles111:
With the latest <a href="https://www.voucherpro.co.uk/gardening-naturally.com">gardening naturally voucher code</a>, you can enjoy a greener and more sustainable lifestyle without breaking the bank.

jamescharles111:
[...]

PR #1749, opened 2017-11-24
Issue #1748 Enable single alphanumeric character for profile name

praveenkumar:


cmd/minishift/cmd/util/util_test.go
```diff
@@ -32,6 +32,9 @@ func TestIsValidProfileName(t *testing.T) {
 		{"foo_123", true},
 		{"foo@123", false},
 		{"", false},
+		{"a", true},
```
LalatenduMohanty: We should add `{"1", true}` to test single digit profile names.

praveenkumar:
> We need to fix the documentation also as it does not talk about _ at all.

@LalatenduMohanty I don't follow, why we want to explicit talk about `_` when already tell alphanumeric in the documentations?

cmd/minishift/cmd/util/util_test.go
praveenkumar: @LalatenduMohanty Since regex is for word range so testing a digit or alphabet would be same but here we are testing single character (be it digit or alphabet).

amitkrout:
Tested. Working as expected 👍

LalatenduMohanty:
> @LalatenduMohanty I don't follow, why we want to explicit talk about _ when already tell alphanumeric in the documentation?

In general alphanumeric character means only alphabets and numbers [1]  anything else is exception. 
[1] https://www.quora.com/What-are-alpha-and-numeric-characters

I consider `_` is not a alphanumeric character and it is not mentioned in https://github.com/minishift/minishift/blob/master/docs/source/using/profiles.adoc#creating-profiles

cmd/minishift/cmd/util/util_test.go
LalatenduMohanty: That's the beauty of unit tests. If we cover all possible combinations of profile name in the unit test then in future eve if we change the regular expression in the function , it will make sure that we do not introduce a regression.

docs/source/using/profiles.adoc
```diff
@@ -31,10 +31,10 @@ There are two ways to create a new profile.
 
 [NOTE]
 ====
-Profile name can only consist of alphanumeric characters and *-* as a separator, for example:
+Profile name can only consist of alphanumeric characters (underscore *_* allowed) and *-* as a separator, for example:
```
hferentschik: I think the () placement is a bit off here. Closing ) after allowed. Why not rewrite this a bit:

"Profile names can only consist of alphanumeric characters.  
The use of *_* and *-* as a separator is allowed. For example"

docs/source/using/profiles.adoc
```diff
@@ -31,10 +31,10 @@ There are two ways to create a new profile.
...
 
-- foo, bar123, demo-profile and 1stProfile are considered as valid profile names.
-- $foo, bar#@, profile@123 are invalid profile names.
+- foo, bar_123, demo-profile and 1stProfile are considered as valid profile names.
```
hferentschik: I find the used examples a bit too casual. I don't think it is good style to use things like foo and bar in the public facing documentation. Let's try to use some more real examples.

docs/source/using/profiles.adoc
```diff
@@ -31,10 +31,10 @@ There are two ways to create a new profile.
...
+- $foo, bar#@, profile@123, -foo are invalid profile names.
```
hferentschik: So why disallows '-foo'. I guess I understand now what you wanted to say in your explanation, but that is not clear at all. Effectively _foo would be ok, but -foo not?

hferentschik:
> I don't follow, why we want to explicit talk about _ when already tell alphanumeric in the documentation?

Because for someone not familiar with regular expressions and the fact that '_' belongs to the alphanumeric character class this is not clear. For most people alphanumeric probably means [a-zA-Z-0-9]. Hence it is better to be explicit. However, in its current form the docs are confusing and there is also the restriction that you cannot use '-' as a first character which I don't quite understand.

praveenkumar:
> However, in its current form the docs are confusing and there is also the restriction that you cannot use '-' as a first character which I don't quite understand.

if we allow `-` also as first character then an user can also create a profile named as `-h` or `--help` and then you can never able to delete it :)

hferentschik:
> if we allow - also as first character then an user can also create a profile named as -h or --help and then you can never able to delete it :)

why not? That's what '--' is for right?

docs/source/using/profiles.adoc
```diff
@@ -31,10 +31,11 @@ There are two ways to create a new profile.
 
 [NOTE]
 ====
-Profile name can only consist of alphanumeric characters and *-* as a separator, for example:
+Profile name can only consist of alphanumeric characters.
+The use of underscore *_* and *-* as a separator is allowd. For example:
```
hferentschik: allowd == allowed

docs/source/using/profiles.adoc
```diff
@@ -31,10 +31,11 @@ There are two ways to create a new profile.
...
 
-- foo, bar123, demo-profile and 1stProfile are considered as valid profile names.
-- $foo, bar#@, profile@123 are invalid profile names.
+- myprofile, myprofile_123, demo-profile and 1stProfile are considered as valid profile names.
+- $myprofile, myprofile#@, profile@123, -myprofile are invalid profile names.
```
hferentschik: By the docs it is not clear why '-myprofile' should be disallowed.

gbraad: it should be described that this can be interpreted as flag/option (and also might be disallowed by the underlying hypervisor tools). However, do we really need to explain the technical details?

gbraad:
>  but having this exception just makes it harder for us,

prevention is the best solution. be defensive. if the criteria does not match, fail... as in the worst case, the hypervisor will not allow this.

docs/source/using/profiles.adoc
LalatenduMohanty: IMO `-` and `_` should be only allowed as separator.  @hferentschik why do you think `-myprofile` is a valid profile name?

cmd/minishift/cmd/util/util.go
```diff
@@ -105,8 +105,8 @@ func IsValidProfile(profileName string) bool {
 	return false
 }
 
-// IsValidProfileName return true if profile name follow ^\w+[\w+-]+$
+// IsValidProfileName return true if profile name follow ^[a-zA-Z0-9]+[\w+-]*$
 func IsValidProfileName(profileName string) bool {
-	rr := regexp.MustCompile(`^\w+[\w+-]+$`)
+	rr := regexp.MustCompile(`^[a-zA-Z0-9]+[\w+-]*$`)
```
LalatenduMohanty: 👍

docs/source/using/profiles.adoc
LalatenduMohanty: I think we can remove the line about invalid profile names. As it adds to the confusion that why these are invalid names.

hferentschik: TBH, I think we can remove the whole example section.

gbraad:
Agree with @hferentschik that they do not add value, but rather invite unnecessay technical explaination... since we do not want to provide, it might be best to remove this.

docs/source/using/profiles.adoc
```diff
@@ -31,10 +31,8 @@ There are two ways to create a new profile.
 
 [NOTE]
 ====
-Profile name can only consist of alphanumeric characters and *-* as a separator, for example:
-
-- foo, bar123, demo-profile and 1stProfile are considered as valid profile names.
-- $foo, bar#@, profile@123 are invalid profile names.
+Profile name can only consist of alphanumeric characters.
+The use of underscore *_* and *-* as a separator is allowed.
```
budhrg: `The use of underscore (`_`) and hyphen (`-`) are allowed as a separator.`

LalatenduMohanty:
retest this please

LalatenduMohanty:
CI failed because nip.io is down today. https://ci.centos.org/job/minishift-pr/2853/consoleFull
```
     """
12:15:24     And status code of HTTP request to "/" of service "ruby-ex" in namespace "ruby" is equal to "200"                                 # integration_test.go:795 -> getRoutingUrlAndVerifyHTTPResponse
12:15:24     Server returned error on url: http://ruby-ex-ruby.192.168.42.30.nip.io
```

PR #1753, opened 2017-11-26
Issue #1570 Nightly job integration in centos_ci.sh script

### LINKED ISSUE #1570: Add nightly test in CentOS CI
Required for running https://github.com/minishift/minishift/issues/1568 and coolstore checks.

Also beginning to address https://github.com/minishift/minishift/issues/1396.

budhrg:
Fix #1570

Job is ready in CentOS CI, last successfully performed here https://ci.centos.org/job/minishift-nightly/10/console with same checks as PR build.

This PR add necessary changes for nightly build.

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -308,6 +308,14 @@ function prepare_for_proxy() {
   firewall-cmd --zone=public --add-port=$INTEGRATION_PROXY_CUSTOM_PORT/tcp;
 }
 
+function perform_nightly() {
+  setup_kvm_docker_machine_driver;
+  cd $GOPATH/src/github.com/minishift/minishift
+  make prerelease synopsis_docs link_check_docs
+  MINISHIFT_VM_DRIVER=kvm make integration_all GODOG_OPTS=""
+  echo "CICO: Tests ran successfully"
+}
+
 if [[ "$UID" = 0 ]]; then
   setup_build_environment;
 else
@@ -322,6 +330,8 @@ else
     docs_tar_upload $RSYNC_PASSWORD;
   elif [[ "$JOB_NAME" = "minishift-release" ]]; then
     perform_release $RSYNC_PASSWORD;
+  elif [[ "$JOB_NAME" = "minishift-nightly" ]]; then
+    perform_nightly;
   else
     build_and_test $RSYNC_PASSWORD;
   fi
--- a/docs/source/contributing/ci.adoc
+++ b/docs/source/contributing/ci.adoc
@@ -1,3 +1,5 @@
+include::variables.adoc[]
+
 = Minishift CI
 :data-uri:
 :imagesdir: images
@@ -39,9 +41,11 @@ The artifacts of a successful pull request build can be found at link:http://art
 
 On top of building the master branch and pull requests, CentOS CI is also used to build the tar bundle used for xref:../contributing/writing-docs.adoc#integration-with-docs-openshift-org[integration with docs.openshift.org], as well as to build xref:../contributing/releasing.adoc#automated-release[releases].
 
+The link:https://ci.centos.org/job/minishift-nightly[nightly job] runs daily at midnight.
+It executes all integration tests against the _master_ branch of the {project} GitHub repository.
 
 |link:https://ci.centos.org/job/minishift/[master], link:https://ci.centos.org/job/minishift-pr/[pull requests], link:https://ci.centos.org/job/minishift-docs/[docs],
-link:https://ci.centos.org/job/minishift-release/[release]
+link:https://ci.centos.org/job/minishift-release/[release], link:https://ci.centos.org/job/minishift-nightly[nightly]
 
 |link:https://github.com/minishift/minishift/blob/master/centos_ci.sh[centos_ci.sh]
 


centos_ci.sh
```diff
@@ -308,6 +308,14 @@ function prepare_for_proxy() {
   firewall-cmd --zone=public --add-port=$INTEGRATION_PROXY_CUSTOM_PORT/tcp;
 }
 
+function perform_nightly() {
+  setup_kvm_docker_machine_driver;
+  cd $GOPATH/src/github.com/minishift/minishift
+  make prerelease synopsis_docs link_check_docs
+  MINISHIFT_VM_DRIVER=kvm make integration_all
```
budhrg: @agajdosi , I hope this is the correct check for nightly build.

budhrg:
Adding documentation now...

docs/source/contributing/ci.adoc
```diff
@@ -60,3 +60,9 @@ link:https://ci.centos.org/job/minishift-release/[release]
 |link:https://travis-ci.org/minishift/minishift/branches[master], link:https://travis-ci.org/minishift/minishift/pull_requests[pull requests]
 |link:https://github.com/minishift/minishift/blob/master/.travis.yml[.travis.yml]
 |===
+
+[[nightly-job]]
+== Nightly Job
+
+A link:https://ci.centos.org/job/minishift-nightly[nightly job] has been configured to be run daily at midnight (2:20 AM UTC) which executes all tests against the `master` branch from the link:https://github.com/minishift/minishift[minishift] repo.
```
budhrg: @minishift/minishift-dev Please review this part of documentation.

anjannath: Nitpick: s/to be run/to run

centos_ci.sh
amitkrout: @budhrg yes,it should be. integration_all will run all the feature file including addon-xpaas.feature except the coolstore feature file.

budhrg: @amitkrout , I think we want to run coolstore as well. So is this what I need to pass ?

```
MINISHIFT_VM_DRIVER=kvm make integration_all GODOG_OPTS="-tags coolstore"
```

hferentschik: @budhrg No, this would just run the coolstore. Try empty GODOG_OPTS. Something like:

    MINISHIFT_VM_DRIVER=kvm make integration_all GODOG_OPTS=""

amitkrout: @budhrg No. This is not the way to run all the feature including coolstore. 

> MINISHIFT_VM_DRIVER=kvm make integration_all GODOG_OPTS="-tags coolstore"

This will run only coolstore feature. AFAIU this is not the requirement here. I would suggest use ```MINISHIFT_VM_DRIVER=kvm make integration_all``` to run all the feature file except coolstore as there is an open issue on coolstore repo https://github.com/jbossdemocentral/coolstore-microservice/. See https://github.com/jbossdemocentral/coolstore-microservice/issues/90 for more details.

docs/source/contributing/ci.adoc
hferentschik: 2:20 am UTC? Is there somewhere a timezone which with 20min offset? Why don't we run it midnight UTC?

hferentschik: So why is this nightly in a new section instead of incorporating it into the table as part of the CentOS CI row?

LalatenduMohanty: @budhrg Why 2:20 AM UTC? Why not 00:00AM UTC.

LalatenduMohanty: s/A nightly job has been configured to be run/The nightly test  job runs daily/. 

Suggestion:
The nightly test  job runs daily at 2:20AM UTC). It executes all tests against the master branch of the Minishift github repository.

LalatenduMohanty: > So why is this nightly in a new section instead of incorporating it into the table as part of the CentOS CI row?

@hferentschik Personally I feel a separate section makes sense as  nightly tests should not be considered as 100% continuous integration.

budhrg: @LalatenduMohanty @hferentschik , I had specified midnight in the job configuration here https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml#L248. Not sure how it is starting after 2:20 hrs from midnight. See `System clock time` timestamps here https://ci.centos.org/job/minishift-nightly/10/consoleFull.

screenshot from 2017-11-27 18-11-33

centos_ci.sh
gbraad: we will run ALL tests in nightl, including coolstore

docs/source/contributing/ci.adoc
budhrg: Done with this change here https://github.com/minishift/minishift-ci-jobs/commit/8f964f783ab2d05412ae4a99be9f421c603f754f

hferentschik:
> Personally I feel a separate section makes sense as nightly tests should not be considered as 100% continuous integration.

The table lists jobs per CI System. There are already docs jobs listed there as well. Adding this new section breaks the whole idea and design of this page.

docs/source/contributing/ci.adoc
hferentschik: there is still a disconnect here. The config is for midnight even if you changed from @midnight to 0 0 * * *. So what's going on? What timezone is set on the slave? What's the timezone on the master? Do we need to set the proper time zone on the slave as part of initialization? Do we need to enable ntp. IMO it is not ok, to just say  "it starts 2:20AM UTC" even though midnight is configured. It is also not ok to make this assumption based on a single run like this. Who says this time displayed is directly related to the trigger? 

Let's find out what's going on here and do this properly.

budhrg: @hferentschik Confirmed with @bstinsonmhk that default time zone `UTC` is set on all CI machines. 

> Do we need to set the proper time zone on the slave as part of initialization?

Setting time zone seems starting jenkins server altogether which might not be feasible here https://wiki.jenkins.io/display/JENKINS/Change+time+zone.

> Who says this time displayed is directly related to the trigger?

With `0 0 * * *` should start the job at sharp 00:00 AM as `@midnight` in terms of Jenkins world means run job in between 00:00 AM to 2:59 AM. 
See https://stackoverflow.com/a/17204211/1120530.

budhrg:
@amitkrout @LalatenduMohanty @hferentschik 

```
MINISHIFT_VM_DRIVER=kvm make integration_all GODOG_OPTS=""
```

is doing the trick to run `coolstore` tests as well.

See below job though it got failed due to timeout at 60mins.
https://ci.centos.org/job/minishift-pr/2788/consoleFull

amitkrout:
> MINISHIFT_VM_DRIVER=kvm make integration_all GODOG_OPTS=""

@budhrg yes this will include all the feature file. In the [make file argument](https://github.com/minishift/minishift/blob/master/Makefile#L206) GODOG_OPTS=""
 will overwrite ```GODOG_OPTS = --tags=~coolstore```. If there is no specific tag passed through the argument then make integration_all will include all the feature file for execution.

> See below job though it got failed due to timeout at 60mins.

See issue #1327

budhrg:
Good to go now. The nightly job is starting at 00:01 AM UTC
screenshot from 2017-11-28 12-08-16

/cc @hferentschik @LalatenduMohanty

docs/source/contributing/ci.adoc
```diff
@@ -41,7 +41,7 @@ On top of building the master branch and pull requests, CentOS CI is also used t
 
 
 |link:https://ci.centos.org/job/minishift/[master], link:https://ci.centos.org/job/minishift-pr/[pull requests], link:https://ci.centos.org/job/minishift-docs/[docs],
-link:https://ci.centos.org/job/minishift-release/[release]
+link:https://ci.centos.org/job/minishift-release/[release], link:https://ci.centos.org/job/minishift-nightly[nightly]
```
LalatenduMohanty: 👍

centos_ci.sh
```diff
@@ -308,6 +308,14 @@ function prepare_for_proxy() {
   firewall-cmd --zone=public --add-port=$INTEGRATION_PROXY_CUSTOM_PORT/tcp;
 }
 
+function perform_nightly() {
+  setup_kvm_docker_machine_driver;
+  cd $GOPATH/src/github.com/minishift/minishift
+  make prerelease synopsis_docs link_check_docs
+  MINISHIFT_VM_DRIVER=kvm make integration_all GODOG_OPTS=""
```
agajdosi: Is this passing? I am suspicious of coolstore.feature, which was not fixed yet. And also of timeout, which is 3600s per default (https://github.com/minishift/minishift/blob/master/Makefile#L35). I would suggest to increase it so we will not hit the limit on next additions to integration test suite. This would be done by adding: `TIMEOUT=7200s` for example. More here: https://docs.openshift.org/latest/minishift/contributing/developing.html#additional-parameters.

amitkrout: Even i am surprised how the it CI passed with the command ```MINISHIFT_VM_DRIVER=kvm make integration_all GODOG_OPTS=""```

>I would suggest to increase it so we will not hit the limit on next additions to integration test suite. This would be done by adding: TIMEOUT=7200s

@budhrg For running integration_all test what is the time taken by CI ?

budhrg: @amitkrout , not known yet.

docs/source/contributing/ci.adoc
hferentschik: So what happened to the extra explanation? There was useful content in the previous version of this change. It's all gone now.

hferentschik: You did not update the description column

docs/source/contributing/ci.adoc
```diff
@@ -39,9 +41,11 @@ The artifacts of a successful pull request build can be found at link:http://art
 
 On top of building the master branch and pull requests, CentOS CI is also used to build the tar bundle used for xref:../contributing/writing-docs.adoc#integration-with-docs-openshift-org[integration with docs.openshift.org], as well as to build xref:../contributing/releasing.adoc#automated-release[releases].
 
+The link:https://ci.centos.org/job/minishift-nightly[nightly job] runs daily at midnight.
+It executes all tests against the _master_ branch of the {project} github repository.
```
hferentschik: integration tests

hferentschik: github => GitHub

hferentschik:
We also need some sort of notification scheme. If the nightly build fails, how do we devs know about it?

budhrg:
@hferentschik  

I have updated the PR with your suggestions and rebased with master as well. 

> We also need some sort of notification scheme. If the nightly build fails, how do we devs know about it?

It will be part of https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml and no changes required from `centos_ci.sh` side.

budhrg:
@hferentschik  

Build notification PR is sent here https://github.com/minishift/minishift-ci-jobs/pull/34

hferentschik:
Merged, thanks.

Issue #1756, opened 2017-11-27
Document X.509 cert issue when VM IP change during start => stop => start


praveenkumar:
Currently, some of the users are facing an issue like below when try to start a stopped minishift VM and IP get changed. We do have some issue to resolve it as part of code but this will take time (check #1515 , #1457 , #391)

In meantime, we should document this behavior in the troubleshooting section of the document. 

```
Error: Get https://192.168.99.100:8443/healthz/ready: x509: certificate is valid for 10.0.2.15, 127.0.0.1, 172.17.0.1, 172.30.0.1, 192.168.99.101, not 192.168.99.100 Error during 'cluster up' execution: Error starting the cluster.
```

gbraad:
@hferentschik Why is the title changed to "certificate regeneration" ? This is something we can not do at the moment, so it is remains an issue. When you want a regenerate, we have to remove ALL configuration, which means we can not regenerate without loss of configuration.

praveenkumar:
@gbraad I again changed the issue to make sure what we are dealing with as part of document changes.

gbraad:
thanks, as this is essential for the changelog/release notes

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/pull/1757/

PR #1757, opened 2017-11-27
Issue #1756 Document x509 certificate issue when VM restart

praveenkumar:


docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -59,4 +59,35 @@ For example:
...
\ No newline at end of file
+----
+
+[[x509-certificate-is-valid-for-10.0.2.15-127.0.0.1-192.168.99.100-not-192.168.99.101]]
+== x509 certificate is valid for 10.0.2.15, 127.0.0.1, 172.17.0.1, 172.30.0.1, 192.168.99.100, not 192.168.99.101
+
+If you started a stopped {project} VM then due to IP change you might hit following x509 certificate error:
+
+----
+$ minishift.exe start
```
anjannath: I think this should be just `minishift` ?

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -59,4 +59,35 @@ For example:
...
+     No log available from "origin" container
+
+   Caused By:
+     Error: Get https://192.168.99.101:8443/healthz/ready: x509: certificate is valid for 10.0.2.15, 127.0.0.1, 172.17.0.1, 172.30.0.1, 192.168.99.100, not 192.168.99.101
+Error during 'cluster up' execution: Error starting the cluster.
+----
+
+The reason is openshift cluster certificates contain IP when {project} VM created and a restart may change the VM IP for which certificates are not valid anymore.
+
+Workaround: Delete existing {project} VM and start again.
```
LalatenduMohanty: s/existing/the existing/

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -59,4 +59,35 @@ For example:
...
+   Error: cannot access master readiness URL https://192.168.99.101:8443/healthz/ready
+   Details:
+     No log available from "origin" container
+
+   Caused By:
+     Error: Get https://192.168.99.101:8443/healthz/ready: x509: certificate is valid for 10.0.2.15, 127.0.0.1, 172.17.0.1, 172.30.0.1, 192.168.99.100, not 192.168.99.101
+Error during 'cluster up' execution: Error starting the cluster.
+----
+
+The reason is openshift cluster certificates contain IP when {project} VM created and a restart may change the VM IP for which certificates are not valid anymore.
```
LalatenduMohanty: The reason for the above error is  that OpenShift cluster certificates contain the IP  of the {project} VM. The certificates are generated only when {project} VM is freshly started (not for restart). So when a restart changes the VM IP the certificates are not valid anymore.

robin-owen: > The certificates are generated only when {project} VM is freshly started (not restart).

@LalatenduMohanty By "freshly started (not restart)", I assume that the certificates are generated the first time that `minishift start` is run, not for any following runs of `minishift start`, correct? If that's the case, I'd like to be explicit about that.

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -59,4 +59,35 @@ For example:
 
 ----
 minishift start --iso-url centos
-----
\ No newline at end of file
+----
+
+[[x509-certificate-is-valid-for-10.0.2.15-127.0.0.1-192.168.99.100-not-192.168.99.101]]
```
robin-owen: I'm not sure if we can do anything about this, but this anchor name is unwieldy.

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -59,4 +59,35 @@ For example:
...
+== x509 certificate is valid for 10.0.2.15, 127.0.0.1, 172.17.0.1, 172.30.0.1, 192.168.99.100, not 192.168.99.101
+
+If you started a stopped {project} VM then due to IP change you might hit following x509 certificate error:
```
robin-owen: No need to mention "due to the IP change" here. The reason for the error can be explained in greater detail later.

I'd also replace "hit" with another word (such as "encounter"); "hit" has a more conversational tone. Similarly, I would use "X.509" instead of "x509".

There's also tense confusion between "if you started" and "you might encounter". I'd recommend rewriting this as follows:

"Starting a stopped {project} VM can produce the following X.509 certificate error:"

docs/source/troubleshooting/troubleshooting-misc.adoc
LalatenduMohanty: @robin-owen yes, thats right.

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -59,4 +59,37 @@ For example:
...
+     No log available from "origin" container
+
+   Caused By:
+     Error: Get https://192.168.99.101:8443/healthz/ready: x509: certificate is valid for 10.0.2.15, 127.0.0.1, 172.17.0.1, 172.30.0.1, 192.168.99.100, not 192.168.99.101
+Error during 'cluster up' execution: Error starting the cluster.
+----
+
+The reason for the above error is that OpenShift cluster certificates contain the IP of the {project} VM.
+The certificates are generated only when {project} VM is freshly started.
+So when a restart changes the {project} VM IP the certificates are not valid anymore.
```
robin-owen: "The certificates become invalid when a restart changes the IP of the {project} VM."

gbraad: Maybe the cause might make this clearer: The lease on the IP address expires, and therefore the VM's IP address will change. It is not the restart itself that changes the address... and it does not always happen.

TL;dr => "Upon restart the {project} VM might get assigned a new IP address. If this happens the certificate is not valid anymore".

docs/source/troubleshooting/troubleshooting-misc.adoc
LalatenduMohanty: @praveenkumar I would suggest the anchor name as `invalid-x509-certificate`. The anchor is used for pointing to the exact section e.g. `external-network-check` is the anchor in https://docs.openshift.org/latest/minishift/troubleshooting/troubleshooting-getting-started.html#external-network-check

budhrg: Agree, with @LalatenduMohanty  to have general anchor name.

praveenkumar: @LalatenduMohanty @budhrg I thought anchor name should be something which easily appears when user search for an error and an error message would be good as an anchor. I defer it to @robin-owen view.

robin-owen: I also agree with @LalatenduMohanty and @budhrg here. We want to keep the title the same, just change the anchor name itself.

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -59,4 +59,37 @@ For example:
...
+     No log available from "origin" container
+
+   Caused By:
+     Error: Get https://192.168.99.101:8443/healthz/ready: x509: certificate is valid for 10.0.2.15, 127.0.0.1, 172.17.0.1, 172.30.0.1, 192.168.99.100, not 192.168.99.101
+Error during 'cluster up' execution: Error starting the cluster.
+----
+
+The reason for the above error is that OpenShift cluster certificates contain the IP of the {project} VM.
+The certificates are generated only when {project} VM is freshly started.
+Restart the {project} VM might get assigned a new IP address if this happens the certificate become invalid.
```
LalatenduMohanty: s/Restart the {project} VM might get assigned a new IP address if this happens the certificate become invalid./After restart the {project} VM might be assigned a new IP address and if this happens the certificate become invalid./

jjwallaby:
This occurs for my experience  on laptop when change  external network provider.   The solution you mention is not acceptable to 'delete project' and restart as you loose ALL your work and existing containers. How poor is this openshift.  This is a show-stopper.

gbraad:
This is due to `oc cluster up` generating certificates. In a real
world situation, the address of the server would not change, but when
you switch network yoiu might end up with a new IP address and this
will mismatch. We have workaround available for this, like pinning the
IP address, or you could set the VM up on your local NAT bridge, etc.
It is not ideal, we are aware, but at this moment the problem is that
`oc cluster up` as scenario will not regenerate the certificates when
this happens.

PR #1777, opened 2017-12-01
Issue #1772 Allow non-alphanumeric characters to password during hostfolder mount

### LINKED ISSUE #1772: Hostfolder password with a non-alphanumeric character like $ fails to mount
### Issue Description
I create a share folder via SMB from `Share` window of `System Preference` on my macOS. However the `minishift hostfolder mount` command got a `FAILED` result while mounting this shared folder.

and from the discussion of #1539 , i believe the issue occurs because of the special character such as dollar sign `$` existing in my password.

### Debugging info

```
# sudo mount -t cifs //192.168.99.1/local /mnt/sda1/local -o username=eliuhy,password=xxxxxx,context=system_u:object_r:svirt_sandbox_file_t:s0
# xxxxxx is a password that contains dollar sign $
...
SSH cmd err, output: exit status 32: mount error(13): Permission denied
Refer to the mount.cifs(8) manual page (e.g. man mount.cifs)
...
```

### Version info
```
$ minishift version
minishift v1.8.0+6455fc3
```

praveenkumar:
Fixes #1772

### DIFF
--- a/pkg/minishift/hostfolder/hostfolder.go
+++ b/pkg/minishift/hostfolder/hostfolder.go
@@ -32,6 +32,7 @@ import (
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
 	miniutil "github.com/minishift/minishift/pkg/minishift/util"
 	"github.com/minishift/minishift/pkg/util"
+	minishiftStrings "github.com/minishift/minishift/pkg/util/strings"
 )
 
 const (
@@ -322,11 +323,11 @@ func mountCifsHostfolder(driver drivers.Driver, hostfolder *config.HostFolder) e
 	}
 
 	cmd := fmt.Sprintf(
-		"sudo mount -t cifs %s %s -o username=%s,password=%s",
+		"sudo mount -t cifs %s %s -o username=%s,password='%s'",
 		hostfolder.Options["uncpath"],
 		hostfolder.Mountpoint(),
 		hostfolder.Options["username"],
-		password)
+		minishiftStrings.EscapeSingleQuote(password))
 
 	if minishiftConfig.InstanceConfig.IsRHELBased {
 		cmd = fmt.Sprintf("%s,context=system_u:object_r:svirt_sandbox_file_t:s0", cmd)


pkg/minishift/hostfolder/hostfolder.go
```diff
@@ -322,11 +323,11 @@ func mountCifsHostfolder(driver drivers.Driver, hostfolder *config.HostFolder) e
 	}
 
 	cmd := fmt.Sprintf(
-		"sudo mount -t cifs %s %s -o username=%s,password=%s",
+		"sudo mount -t cifs %s %s -o username=%s,password='%s'",
```
hferentschik: We could split this already rather long method up a bit and extract for example a method which returns the command string, given some sensible input. This allows us to test everything up to the point where the command is actually executed. This way we can codify our expectations on how this command looks like when there are for example special characters in the password.

hferentschik:
retest this please

pkg/minishift/hostfolder/hostfolder.go
gbraad: I would rather see that happen as part of a coordinated refactor

gbraad:
retest this please

pkg/minishift/hostfolder/hostfolder.go
gbraad: I also believe such a chance will unnecessarily conflict with the refactoring that happens in #317, such as commit: https://github.com/hferentschik/minishift/commit/718caf9403e8ece32adf7b42bbbc6f66c41aa333

gbraad:
The commit title contains a typo, but I'll rectify this during rebase

pkg/minishift/hostfolder/hostfolder.go
LalatenduMohanty: IMO we should have a separate issue for this. If @praveenkumar is up for it , he can fix both the issues at one PR.

LalatenduMohanty:
> If @praveenkumar is up for it , he can fix both the issues at one PR.

@praveenkumar Do you want to refactor the code as per the comment https://github.com/minishift/minishift/pull/1777#discussion_r154422752 in this PR?

praveenkumar:
> @praveenkumar Do you want to refactor the code as per the comment #1777 (comment) in this PR?

@LalatenduMohanty I prefer it to handle when we do refactor so we can have a single function around commands which include user inputs.

gbraad:
Both issues? I am sorry, but I do not see 'both issues'.

gbraad:
Merged as f6cdb9597ea106d6469b32361e08c67d6454ca10

Issue #1776, opened 2017-12-01
Update default OpenShift Origin version to 3.7.1 in Minishift


LalatenduMohanty:
Now that Origin v3.7.0 is released we should move to the latest GA version.

hferentschik:
There is also the service catalog question. Per 3.7 it should be enabled er default. Does this work? With the upgrade does the service catalog get provisioned automatically? If so, there is no need anymore for the experimental flag and it should be removed from the code and docs.

praveenkumar:
> There is also the service catalog question. Per 3.7 it should be enabled er default. Does this work? With the upgrade does the service catalog get provisioned automatically? If so, there is no need anymore for the experimental flag and it should be removed from the code and docs.

Problem is `oc cluster up` still have this as an experimental feature like logging/metrics so I still think we should not expose this by default.

```
$ oc version
oc v3.7.0+7ed6862
kubernetes v1.7.6+a08f5eeb62
features: Basic-Auth GSSAPI Kerberos SPNEGO

Server https://192.168.110.218:8443
openshift v3.7.0+7ed6862
kubernetes v1.7.6+a08f5eeb62

$ oc cluster up -h | grep catalog
      --service-catalog=false: Install service catalog (experimental).
```

jorgemoralespou:
@LalatenduMohanty Service Catalog should no longer be an experimental flag, as it is a GA feature in OpenShift 3.7 and should be exposed and available.

praveenkumar:
@jorgemoralespou As I said https://github.com/minishift/minishift/issues/1776#issuecomment-349870979 since for `oc cluster up` it's experimental flag then we can't go ahead and make it permanent and internally use same logic :(

jorgemoralespou:
@praveenkumar. Totally disagree. The same way logging and metrics are
experimental and they are exposed. They are marked experimental because
provisioning of the feature is not really bullet proof, but to be honest,
service catalog is more reliable than metrics or logging, since it's
developed by same team as oc cluster.

praveenkumar:
> Totally disagree. The same way logging and metrics are experimental and they are exposed. They are marked experimental because provisioning of the feature is not really bullet proof, but to be honest, service catalog is more reliable than metrics or logging, since it's developed by same team as oc cluster.

@jorgemoralespou right but even metrics and logging has still experimented from the minishift side. If you want same with `service catalog` then we can expose that option but our help string will show that it experiment same as oc cluster up.

```
[...]

PR #1779, opened 2017-12-01
Issue #1776 Updating Origin version to v3.7.0

LalatenduMohanty:


LalatenduMohanty:
retest this please

budhrg:
retest this please

Makefile
```diff
@@ -14,7 +14,7 @@
 
 # Various versions - Minishift, default OpenShift, default B2D ISO
 MINISHIFT_VERSION = 1.9.0
-OPENSHIFT_VERSION = v3.6.0
+OPENSHIFT_VERSION = v3.7.0
```
praveenkumar: 👍

hferentschik:
retest this please

LalatenduMohanty:
> If we only do this change then we are breaking our master to use < 3.7.0 version of OpenShift. Because as per Ben currently, you can't provision a 3.6 cluster using 3.7 client binary and in our code we only download oc client binary if it is greater than our specified version, otherwise we use same client to provision.

Good point, I will fix it with #1417

hferentschik:
There is also the service catalog question. Per 3.7 it should be enabled er default. Does this work? With this upgrade did the service catalog get provisioned automatically? If so, there is no need anymore for the experimental flag and it should be removed from the code and docs.

PR #1786, opened 2017-12-04
Issue #1785 Validate commit message as part of CI builds

### LINKED ISSUE #1785: Document the availability of commit message validation
It seems that only [Circle CI](https://github.com/minishift/minishift/blob/master/circle.yml#L41) is [validating commit message](https://docs.openshift.org/latest/minishift/contributing/developing.html#commit-message-requirements). Document this behavior.

budhrg:
Fix #1785

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
 - export PATH=$PATH:/tmp/godep/
 
 script:
-- make prerelease
+- make validate_commits prerelease
 
 notifications:
   webhooks: http://minibot.19cf262c.svc.dockerapp.io:9009/hubot/travis-ci
--- a/circle.yml
+++ b/circle.yml
@@ -35,10 +35,10 @@ dependencies:
 ## Customize test commands
 test:
     override:
-        - make prerelease:
+        - make validate_commits:
                 pwd:
                     ../src/github.com/minishift/minishift
-        - make validate_commits:
+        - make prerelease:
                 pwd:
                     ../src/github.com/minishift/minishift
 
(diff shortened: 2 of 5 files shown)


budhrg:
Regex for commit message validation is wrong. See https://github.com/minishift/minishift/issues/1788

circle.yml
```diff
@@ -35,10 +35,10 @@ dependencies:
 ## Customize test commands
 test:
     override:
-        - make prerelease:
+        - make validate_commits:
```
gbraad: Is this change for CircleCI related to priority or ordering?

.travis.yml
```diff
@@ -19,7 +19,7 @@ install:
 - export PATH=$PATH:/tmp/godep/
 
 script:
-- make prerelease
+- make validate_commits prerelease
```
hferentschik: -1, this is not going to work. Travis is setting up the get repo to be in a detached state. The current validation target is not going to work here.

hferentschik:
-1 Only Circle CI sets the build branch up in a way that the current git validation message target will work. effectively it uses the same setup as a developer would have. 

All other CI system work with a detached HEAD which will drip the validation.

hferentschik:
I recommend closing this pull request. If anything we can add a comment to  circle.yml as well as the CI adoc page on the fact that commit message validation is only enabled on one CI system.

gbraad:
Defer

gbraad:
Please record what the needs and restrictions are to using git-validation. It does not seem to handle detached or shallow checkouts well... ?

Issue #1261, opened 2017-08-11
Restructure code around ValidateOpenshiftMinVersion


LalatenduMohanty:
Currently ValidateOpenshiftMinVersion() [1]  is in pkg/minishift/clusterup/clusterup.go but it should be in 
https://github.com/minishift/minishift/blob/master/pkg/minishift/openshift/version/openshift_versions.go as this package supposed to contains function around OpenShift versions. 

Also the function name ` ValidateOpenshiftMinVersion` does not sound correct as it is actually doing a semantic version comparison between two versions and just not with minimum supported version. 

[1] https://github.com/minishift/minishift/blob/master/pkg/minishift/clusterup/clusterup.go#L282-L299

budhrg:
Still to do.

LalatenduMohanty:
Fixed via #1806

PR #1806, opened 2017-12-06
Issue #1417 Use 3.6.0 for versions < v3.7.0 

### LINKED ISSUE #1417: Minishift should use oc binary of 3.6 or greater, depending upon the OpenShift version needs to be provisioned
With OpenShift origin issue of [`Stop allowing oc cluster up version X to instantiate a server version X-1`](https://github.com/openshift/origin/issues/15731), we need to follow-up plan for the impact of minishift behavior and how to tackle it.

LalatenduMohanty:
Fixes #1417

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -21,6 +21,8 @@ import (
 	"runtime"
 	"strings"
 
+	"os"
+
 	"github.com/asaskevich/govalidator"
 	units "github.com/docker/go-units"
 	"github.com/docker/machine/libmachine"
@@ -44,6 +46,7 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/hostfolder"
 	minishiftNetwork "github.com/minishift/minishift/pkg/minishift/network"
 	"github.com/minishift/minishift/pkg/minishift/openshift"
+	openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"
 	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
 	"github.com/minishift/minishift/pkg/util"
@@ -54,7 +57,6 @@ import (
 	"github.com/spf13/cobra"
 	flag "github.com/spf13/pflag"
 	"github.com/spf13/viper"
-	"os"
 )
 
 const (
@@ -645,7 +647,7 @@ func ensureNotRunning(client *libmachine.Client, machineName string) {
 func validateOpenshiftVersion() {
 	requestedVersion := viper.GetString(configCmd.OpenshiftVersion.Name)
 
-	valid, err := clusterup.ValidateOpenshiftMinVersion(requestedVersion, constants.MinOpenshiftSupportedVersion)
+	valid, err := openshiftVersion.IsGreaterOrEqualToBaseVersion(requestedVersion, constants.MinimumSupportedOpenShiftVersion)
 	if err != nil {
 		atexit.ExitWithMessage(1, err.Error())
 	}
@@ -653,7 +655,7 @@ func validateOpenshiftVersion() {
 	if !valid {
 		fmt.Printf("Minishift does not support OpenShift version %s. "+
 			"You need to use a version >= %s\n", viper.GetString(configCmd.OpenshiftVersion.Name),
-			constants.MinOpenshiftSupportedVersion)
+			constants.MinimumSupportedOpenShiftVersion)
 		atexit.Exit(1)
 	}
 
--- a/pkg/minikube/constants/constants.go
+++ b/pkg/minikube/constants/constants.go
@@ -25,17 +25,19 @@ import (
 )
 
 const (
-	APIServerPort                = 8443             // Port that the API server should listen on
-	MiniShiftEnvPrefix           = "MINISHIFT"      // Prefix for the environmental variables
-	MiniShiftHomeEnv             = "MINISHIFT_HOME" // Environment variable used to change the Minishift home directory
-	VersionPrefix                = "v"
-	MinOpenshiftSupportedVersion = "v1.4.1"
-	DefaultMemory                = "2GB"
-	DefaultCPUS                  = 2
-	DefaultDiskSize              = "20GB"
-	UpdateMarkerFileName         = "updated"
-	DefaultMachineName           = "minishift"
-	DefaultProfileName           = "minishift"
+	APIServerPort                 = 8443             // Port that the API server should listen on
+	MiniShiftEnvPrefix            = "MINISHIFT"      // Prefix for the environmental variables
+	MiniShiftHomeEnv              = "MINISHIFT_HOME" // Environment variable used to change the Minishift home directory
+	VersionPrefix                 = "v"
+	MinOpenshiftSupportedVersion  = "v1.4.1"
+	MinimumOcBinaryVersion        = "v3.6.0" // For provisioning OpenShift versions < v3.7.0 , we need to use oc v3.6.0. Refer #1417
+	BackwardIncompatibleOcVersion = "v3.7.0" // From v3.7.0 oc binary will not provision older versions of OpenShift instance
+	DefaultMemory                 = "2GB"
+	DefaultCPUS                   = 2
+	DefaultDiskSize               = "20GB"
+	UpdateMarkerFileName          = "updated"
+	DefaultMachineName            = "minishift"
+	DefaultProfileName            = "minishift"
 )
 
 var (
--- a/pkg/minikube/constants/constants.go
+++ b/pkg/minikube/constants/constants.go
@@ -25,19 +25,19 @@ import (
 )
 
 const (
-	APIServerPort                 = 8443             // Port that the API server should listen on
-	MiniShiftEnvPrefix            = "MINISHIFT"      // Prefix for the environmental variables
-	MiniShiftHomeEnv              = "MINISHIFT_HOME" // Environment variable used to change the Minishift home directory
-	VersionPrefix                 = "v"
-	MinOpenshiftSupportedVersion  = "v1.4.1"
-	MinimumOcBinaryVersion        = "v3.6.0" // For provisioning OpenShift versions < v3.7.0 , we need to use oc v3.6.0. Refer #1417
-	BackwardIncompatibleOcVersion = "v3.7.0" // From v3.7.0 oc binary will not provision older versions of OpenShift instance
-	DefaultMemory                 = "2GB"
-	DefaultCPUS                   = 2
-	DefaultDiskSize               = "20GB"
-	UpdateMarkerFileName          = "updated"
-	DefaultMachineName            = "minishift"
-	DefaultProfileName            = "minishift"
+	APIServerPort                    = 8443             // Port that the API server should listen on
+	MiniShiftEnvPrefix               = "MINISHIFT"      // Prefix for the environmental variables
+	MiniShiftHomeEnv                 = "MINISHIFT_HOME" // Environment variable used to change the Minishift home directory
+	VersionPrefix                    = "v"
+	MinimumSupportedOpenShiftVersion = "v1.4.1"
+	MinimumOcBinaryVersion           = "v3.6.0" // For provisioning OpenShift versions < v3.7.0 , we need to use oc v3.6.0. Refer #1417
+	BackwardIncompatibleOcVersion    = "v3.7.0" // From v3.7.0 oc binary will not provision older versions of OpenShift instance
+	DefaultMemory                    = "2GB"
+	DefaultCPUS                      = 2
+	DefaultDiskSize                  = "20GB"
+	UpdateMarkerFileName             = "updated"
+	DefaultMachineName               = "minishift"
+	DefaultProfileName               = "minishift"
 )
 
 var (
--- a/pkg/minishift/clusterup/clusterup.go
+++ b/pkg/minishift/clusterup/clusterup.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"fmt"
 
-	"github.com/blang/semver"
 	"github.com/docker/machine/libmachine/host"
 	"github.com/docker/machine/libmachine/provision"
 	"github.com/golang/glog"
@@ -37,6 +36,7 @@ import (
 	"github.com/minishift/minishift/pkg/minishift/addon/manager"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
 	"github.com/minishift/minishift/pkg/minishift/oc"
+	openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"
 
 	"regexp"
 
@@ -168,7 +168,7 @@ func EnsureHostDirectoriesExist(host *host.Host, dirs []string) error {
 // If the requested OpenShift version is >= v3.7.0, we align the oc version with the requested OpenShift version.
 // Check Minishift github issue #1417 for details.
 func DetermineOcVersion(requestedVersion string) string {
-	valid, _ := ValidateOpenshiftMinVersion(requestedVersion, constants.BackwardIncompatibleOcVersion)
+	valid, _ := openshiftVersion.IsGreaterOrEqualToBaseVersion(requestedVersion, constants.BackwardIncompatibleOcVersion)
 	if !valid {
 		requestedVersion = constants.MinimumOcBinaryVersion
 	}
@@ -289,23 +289,4 @@ func getConfigClusterName(hostname string, ip string, port int) string {
 		return "local-cluster"
 	}
 	return fmt.Sprintf("%s:%d", strings.Replace(ip, ".", "-", -1), port)
-}
-
-func ValidateOpenshiftMinVersion(version string, minVersion string) (bool, error) {
-	v, err := semver.Parse(strings.TrimPrefix(version, constants.VersionPrefix))
-	if err != nil {
-		return false, errors.New(fmt.Sprintf("Invalid version format '%s': %s", version, err.Error()))
-	}
-
-	minSupportedVersion := strings.TrimPrefix(minVersion, constants.VersionPrefix)
-	versionRange, err := semver.ParseRange(fmt.Sprintf(">=%s", minSupportedVersion))
-	if err != nil {
-		fmt.Println("Not able to parse version info", err)
-		return false, err
-	}
-
-	if versionRange(v) {
-		return true, nil
-	}
-	return false, nil
 }
--- a/pkg/minishift/clusterup/clusterup.go
+++ b/pkg/minishift/clusterup/clusterup.go
@@ -42,7 +42,6 @@ import (
 
 	"github.com/minishift/minishift/pkg/util"
 	minishiftStrings "github.com/minishift/minishift/pkg/util/strings"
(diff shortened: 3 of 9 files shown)


pkg/minikube/constants/constants.go
```diff
@@ -25,17 +25,19 @@ import (
...
-	DefaultDiskSize              = "20GB"
-	UpdateMarkerFileName         = "updated"
-	DefaultMachineName           = "minishift"
-	DefaultProfileName           = "minishift"
+	APIServerPort                 = 8443             // Port that the API server should listen on
+	MiniShiftEnvPrefix            = "MINISHIFT"      // Prefix for the environmental variables
+	MiniShiftHomeEnv              = "MINISHIFT_HOME" // Environment variable used to change the Minishift home directory
+	VersionPrefix                 = "v"
+	MinOpenshiftSupportedVersion  = "v1.4.1"
+	MinimumOcBinaryVersion        = "v3.6.0"
```
budhrg: Can we use `MinOcBinaryVersion` as like `MinOpenshiftSupportedVersion`

pkg/minishift/clusterup/clusterup.go
```diff
@@ -168,9 +167,9 @@ func EnsureHostDirectoriesExist(host *host.Host, dirs []string) error {
 // If the requested version is < the base line version we use baseline oc to provision the requested OpenShift version.
 // If the requested OpenShift version is >= the baseline, we align the oc version with the requested OpenShift version.
 func DetermineOcVersion(requestedVersion string) string {
-	valid, _ := ValidateOpenshiftMinVersion(requestedVersion, version.GetOpenShiftVersion())
+	valid, _ := ValidateOpenshiftMinVersion(requestedVersion, constants.BackwardIncompatibleOcVersion)
 	if !valid {
-		requestedVersion = version.GetOpenShiftVersion()
+		requestedVersion = constants.MinimumOcBinaryVersion
```
hferentschik: I have a hard time matching the name of these variables to what is going on in the code. Also the docs are not in sync afaikt.

LalatenduMohanty: @hferentschik I have gone through the existing docs to see if any documentation needs changes and found none. But I need to add the information about the changes this PR brings in.

hferentschik: I mean the godocs of the method

hferentschik: Also, I am referring to the naming of the methods. Reading _DetermineOcVersion_ and its godocs I have a hard time mapping methods like _ValidateOpenshiftMinVersion_ into this. Or constants like constants.MinimumOcBinaryVersion or BackwardIncompatibleOcVersion for that matter. 

I think _ValidateOpenshiftMinVersion_ should be renamed as well as the constants. At the very least the constants should get a godoc.

hferentschik:
@LalatenduMohanty, there is a _provision_earlier_version.feature_ which provisions OpenShift 1.5.1. Why don't we extend this a bit, make it a _Scenario Outline_ and let it provisions all major versions we are interested in. At least 1.5.1, 3.6, 3.7. We can extend the assertions to check the downloaded oc version as well which would directly link to this issue. 

The only downside is the time this takes, but now that we have a nightly build in place, we can add it there. WDYT?

LalatenduMohanty:
> there is a provision_earlier_version.feature which provisions OpenShift 1.5.1. Why don't we extend this a bit, make it a Scenario Outline and let it provisions all major versions we are interested in. At least 1.5.1, 3.6, 3.7. We can extend the assertions to check the downloaded oc version as well which would directly link to this issue.

Sounds good to me.

pkg/minishift/clusterup/clusterup.go
LalatenduMohanty: We have an issue https://github.com/minishift/minishift/issues/1261 to fix validateOpenshiftMinVersion. I will trying fix it in this issue.

LalatenduMohanty:
> At least 1.5.1, 3.6, 3.7. We can extend the assertions to check the downloaded oc version as well which would directly link to this issue.

Regarding how I can check if the expected oc version is getting cached , I was thinking the following

* Clearcache
* minishift start --openshift-version v1.5.1
* Check the cache directory to see if v3.6.0 is cached or not. 

Or do we have better solution than this? @minishift/minishift-dev @amitkrout @agajdosi 
Also as you mentioned it will take little more time , so I hope everyone is ok with it.

LalatenduMohanty:
Or we can do the following 

```
$ minishift start --openshift-version v1.5.1
$ minishift status
Minishift:  Running
Profile:    minishift
OpenShift:  Running (openshift v1.5.1+7b451fc)
DiskUsage:  6% of 17.9G

**Comment by praveenkumar** [comment]
> Or do we have better solution than this?

@LalatenduMohanty, why do you to perform all the steps, when you provision any version using minishift the 'OcPath', have detail about which `oc` binary we used and runing version against it give you client version what you are looking to.

**Comment by praveenkumar** [comment]
> Or we can do the following

This is not give you which version of client you used :(

**Comment by amitkrout** [comment]
> minishift delete --clear-cache --force
@LalatenduMohanty 
Why you want to delete it though its not created in a fresh setup ?
I think even we find the oc version in the m start log which is being used for provisioning.

**Comment by LalatenduMohanty** [comment]
> Why you want to delete it though its not created in a fresh setup ?

To make sure I do not get any false positives from the previous tests. 

> I think even we find the oc version in the m start log which is being used for provisioning.
Nope, it is OpenShift server version. Check below output.
```
$ minishift start --openshift-version v1.5.1
--
xxxxx
   Version: v1.5.1
   host-volumes-dir ... OK
   version ... OK
   routing-suffix ... OK
   host-config-dir ... OK
   host-data-dir ... OK
   host-pv-dir ... OK
Starting OpenShift using openshift/origin:v1.5.1 ...
Pulling image openshift/origin:v1.5.1
Pulled 0/3 layers, 3% complete
Pulled 1/3 layers, 56% complete
Pulled 2/3 layers, 71% complete
Pulled 3/3 layers, 100% complete
Extracting
Image pull complete
OpenShift server started.

    https://192.168.64.200:8443

You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin

```

**Comment by amitkrout** [comment]
ahh..without ```--openshift-version``` flag m start log list out oc version being provisioned. Not an issue

IMO the steps would be something like 
-     minishift delete --clear-cache --force
-     minishift start --openshift-version v1.5.1
-     oc version

If some cross verification is needed then more steps can be added

File: test/integration/features/provision_various_versions.feature [hunk 3]
```diff
@@ -0,0 +1,25 @@
+@provision-various-versions @openshift @minishift-only
```
LalatenduMohanty: Creating a separate file as we want to run it in the nightly test.

cmd/minishift/cmd/start.go
```diff
@@ -645,7 +647,7 @@ func ensureNotRunning(client *libmachine.Client, machineName string) {
 func validateOpenshiftVersion() {
 	requestedVersion := viper.GetString(configCmd.OpenshiftVersion.Name)
 
-	valid, err := clusterup.ValidateOpenshiftMinVersion(requestedVersion, constants.MinOpenshiftSupportedVersion)
+	valid, err := openshiftVersion.IsGTEBaseVersion(requestedVersion, constants.MinOpenshiftSupportedVersion)
```
gbraad: GTE => Greater Than or Equal? Too ambiguous...
Please, provide better naming

pkg/minishift/clusterup/clusterup.go
```diff
@@ -165,12 +164,13 @@ func EnsureHostDirectoriesExist(host *host.Host, dirs []string) error {
...
 
 // DetermineOcVersion returns the oc version to use.
-// If the requested version is < the base line version we use baseline oc to provision the requested OpenShift version.
-// If the requested OpenShift version is >= the baseline, we align the oc version with the requested OpenShift version.
+// If the requested version is < v3.7.0 we will use oc binary v3.6.0 to provision the requested OpenShift version.
+// If the requested OpenShift version is >= v3.7.0, we align the oc version with the requested OpenShift version.
+// Check Minishift github issue #1417 for details.
 func DetermineOcVersion(requestedVersion string) string {
-	valid, _ := ValidateOpenshiftMinVersion(requestedVersion, version.GetOpenShiftVersion())
+	valid, _ := openshiftVersion.IsGTEBaseVersion(requestedVersion, constants.BackwardIncompatibleOcVersion)
```
gbraad: GTE => Greater Than or Equal? Too ambiguous...
Please, provide better naming

LalatenduMohanty: @gbraad What about IsGreaterThanVersion or IsGreaterThanBaseVersion

budhrg: @LalatenduMohanty @gbraad , I would say `IsGreaterThanEqualToBaseVersion` as `IsGreaterThanBaseVersion` or `IsGreaterThanVersion` is incomplete and one could easily think that this function is skipping equally check.

gbraad: Can this be solved in a more explicit way? This feels confusing

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -136,3 +137,24 @@ func isPrerelease(tag string) bool {
...
+
+// IsGTEBaseVersion (Is greater than equal to base version) returns true
+// if the version is greater or equal to the base version
+func IsGTEBaseVersion(version string, baseVersion string) (bool, error) {
+	v, err := semver.Parse(strings.TrimPrefix(version, constants.VersionPrefix))
+	if err != nil {
+		return false, errors.New(fmt.Sprintf("Invalid version format '%s': %s", version, err.Error()))
+	}
+
+	minSupportedVersion := strings.TrimPrefix(baseVersion, constants.VersionPrefix)
```
praveenkumar: This should be changed to baseVersion instead `minSupportedVersion` now.

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -136,3 +137,24 @@ func isPrerelease(tag string) bool {
...
+	versionRange, err := semver.ParseRange(fmt.Sprintf(">=%s", minSupportedVersion))
```
praveenkumar: same here.

pkg/minishift/clusterup/clusterup.go
praveenkumar: @LalatenduMohanty `IsGreaterOrEqualTo`

test/integration/features/provision_earlier_version.feature
```diff
@@ -11,6 +11,12 @@ Feature: Provision an older major release
      """
      openshift v1.5.1
      """
+      And JSON config file "machines/minishift.json" contains key "OcPath" with value matching "v3.6.0"
+     When executing "minishift oc-env" succeeds
+     Then stdout should contain
+      """
+      v3.6.0
```
praveenkumar: 👍

test/integration/features/provision_various_versions.feature
```diff
@@ -0,0 +1,29 @@
...
+     When executing "minishift oc-env" succeeds
+     Then stdout should contain
+      """
+      <OcVersion>
+      """
+Examples:
+    | serverVersion | OcVersion |
+    | v1.5.1        |  v3.6.0   |
+    | v3.6.0        |  v3.6.0   |
+    | v3.7.0        |  v3.7.0   |
```
praveenkumar: Since we agreed to have all major supported version then shouldn't we also add `1.4.1` here?

test/integration/features/provision_various_versions.feature
```diff
@@ -0,0 +1,29 @@
+@provision-various-versions @openshift @minishift-only
+Feature: Provision all major Openshift versions
+  As a user I can provision major versions of openshift
+
+Scenario Outline:: Provision all major OpenShift versions
```
agajdosi: There is an extra colon, also the Scenario Outline should be indented by 2 spaces.

test/integration/features/provision_various_versions.feature
amitkrout: @LalatenduMohanty This will fail as each time it will provision --openshift-version on same minishift instance. To avoid the failure  need to add ```executing "minishift delete --force" succeeds``` at the end of the scenario.

This would be like:
```
@provision-various-versions @openshift @minishift-only
Feature: Provision all major Openshift versions
  As a user I can provision major versions of openshift

Scenario Outline:: Provision all major OpenShift versions
     When executing "minishift start --openshift-version <serverVersion>" succeeds
     When executing "minishift openshift version" succeeds
     Then stdout should contain
     """
     openshift <serverVersion>
     """
      And JSON config file "machines/minishift.json" contains key "OcPath" with value matching "<OcVersion>"
     When executing "minishift oc-env" succeeds
     Then stdout should contain
      """
      <OcVersion>
      """
     Then executing "minishift delete --force" succeeds
      And Minishift should have state "Does Not Exist"
Examples:
    | serverVersion | OcVersion |
    | v1.5.1        |  v3.6.0   |
    | v3.6.0        |  v3.6.0   |
    | v3.7.0        |  v3.7.0   |
```

cmd/minishift/cmd/start.go
```diff
@@ -645,7 +647,7 @@ func ensureNotRunning(client *libmachine.Client, machineName string) {
 func validateOpenshiftVersion() {
 	requestedVersion := viper.GetString(configCmd.OpenshiftVersion.Name)
 
-	valid, err := clusterup.ValidateOpenshiftMinVersion(requestedVersion, constants.MinOpenshiftSupportedVersion)
+	valid, err := openshiftVersion.IsGreaterOrEqualToBaseVersion(requestedVersion, constants.MinOpenshiftSupportedVersion)
```
gbraad: `MinimumSupportedOpenShiftVersion` => but wasn't this the base version?

test/integration/features/provision_various_versions.feature
```diff
@@ -0,0 +1,26 @@
+@provision-various-versions @openshift @minishift-only
+Feature: Provision all major Openshift versions
```
agajdosi: Capital S in `OpenShift` is missing.

test/integration/features/provision_various_versions.feature
```diff
@@ -0,0 +1,26 @@
...
+  As a user I can provision major versions of openshift
```
agajdosi: Also capitals in `OpenShift`.

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -136,3 +137,23 @@ func isPrerelease(tag string) bool {
 	}
 	return false
 }
+
+// IsGreaterOrEqualToBaseVersion returns true if the version is greater or equal to the base version
```
gbraad: Why do we actually need to provide the base version as a parameter?

test/integration/features/provision_various_versions.feature
```diff
@@ -0,0 +1,26 @@
...
+     """
+      And JSON config file "machines/minishift.json" contains key "OcPath" with value matching "<ocVersion>"
+     When executing "minishift oc-env" succeeds
+     Then stdout should contain
+      """
+      <ocVersion>
+      """
+     When executing "minishift delete --force" succeeds
+     Then Minishift should have state "Does Not Exist"
+Examples:
```
agajdosi: Should be indented by 2 spaces as it is part of Feature block.

test/integration/features/provision_various_versions.feature
```diff
@@ -0,0 +1,26 @@
...
+    | serverVersion | ocVersion |
+    | v1.5.1        |  v3.6.0   |
+    | v3.6.0        |  v3.6.0   |
+    | v3.7.0        |  v3.7.0   |
```
agajdosi: Please add a newline.

test/integration/features/provision_various_versions.feature
```diff
@@ -0,0 +1,27 @@
...
+      <ocVersion>
+      """
+     When executing "minishift delete --force" succeeds
+     Then Minishift should have state "Does Not Exist"
+Examples:
+    | serverVersion | ocVersion |
+    | v1.4.1        |  v3.6.0   |
+    | v1.5.1        |  v3.6.0   |
+    | v3.6.0        |  v3.6.0   |
+    | v3.7.0        |  v3.7.0   |
```
agajdosi: Please add a newline.

budhrg: New line needed here

test/integration/features/provision_various_versions.feature
amitkrout: Examples: -> Examples: OpenShift and oc versions

test/integration/features/provision_various_versions.feature
```diff
@@ -0,0 +1,27 @@
...
+     """
+     openshift <serverVersion>
+     """
+      And JSON config file "machines/minishift.json" contains key "OcPath" with value matching "<ocVersion>"
+     When executing "minishift oc-env" succeeds
+     Then stdout should contain
+      """
+      <ocVersion>
+      """
+     When executing "minishift delete --force" succeeds
```
amitkrout: Little confusion...We are deleting the minishift instance not clearing the cache where oc is stored. Each time the loop runs it will keep the oc binary might causes conflict. To mke the clean run each time i would suggest ```When executing "minishift delete --force --clear-cache" succeeds``` would be a better resolution.

LalatenduMohanty: @amitkrout I wanted to use the cached ISO to reduce execution time, so did not cleared the cache. Also if we do anot want to use cache in any test, we can clear cache first.

LalatenduMohanty: > Each time the loop runs it will keep the oc binary that might causes conflict

@amitkrout it is not as you can see the test. Also it should not conflict.

test/integration/features/provision_various_versions.feature
LalatenduMohanty: Added v1.4.1

pkg/minishift/openshift/version/openshift_versions.go
praveenkumar: @gbraad It basically a compare function of 2 different openshift versions.

amitkrout:
Looks good to me, however the look and feel of the feature file is not so impressive i mean indentation. @LalatenduMohanty i would suggest please have a look to the [comment ](https://github.com/minishift/minishift/issues/1815#issuecomment-350140221) so that you can make the indentation much better or probably see this [pr feature file ](https://github.com/minishift/minishift/pull/1692/files#diff-db8e17a1ae346b091b162c735329b2b7)

test/integration/features/provision_various_versions.feature
amitkrout: Got it...thanks for the clarification

LalatenduMohanty:
@amitkrout Fixed the indentation as you suggested.

pkg/minishift/openshift/version/openshift_versions.go
gbraad: That is clear, but the naming of the variables confused... this is improved... except for one case.

LalatenduMohanty:
@gbraad I agree, lets fix this after 1.10. I will send a PR

Issue #1675, opened 2017-11-09
Preflight check for Hyper-V setup too restrictive


hferentschik:
I have a new fresh Windows install, enabled Hyper-V and configured the external switch. I even set _HYPERV_VIRTUAL_SWITCH_. With or without this variable, I cannot pass the driver check:

    PS C:\Users\Administrator\Desktop> .\minishift.exe start
    -- Checking if requested hypervisor 'hyperv' is supported on this platform ..
    -- Checking if Hyper-V driver is installed ... OK
    -- Checking if Hyper-V driver is configured to use a Virtual Switch ... FAIL 

I then skip the driver checks via:

    > .\minishift.exe config set warn-check-hyperv-driver true

After that Minishift starts as expected.

gbraad:
Please let me know what you have set as `HYPERV_VIRTUAL_SWITCH` ?

gbraad:
The option `m config set warn-check-hyperv-driver true` will not fail the steps, but give the output... So I would like to know what the output was after setting this.

gbraad:
A possible option for default: #1676

gbraad:
Accidently closed

hferentschik:
I will update once I am back to Hyper-V. For now, I am using VirtualBox. AFAIR there was no additonal output w/ or w/o warn-check-hyperv-driver set. When it was not set, the check just failed. 

I set HYPERV_VIRTUAL_SWITCH via:

     set HYPERV_VIRTUAL_SWITCH=External

as per docs https://docs.openshift.org/latest/minishift/getting-started/setting-up-driver-plugin.html#hyper-v-driver

gbraad:
Either must have been a typo, or a syntax error: In the command prompt, do not use `"`. Looking forward to the output...

For me it is currently "cannot reproduce" (using Hyper-V every day).

hferentschik:
I think the problem is:

    PS C:\Users\Administrator> Get-Command Get-VM
    Get-Command : The term 'Get-VM' is not recognized as the name of a cmdlet,     function, script file, or operable program.
    Check the spelling of the name, or if a path was included, verify that the     path is correct and try again.
    At line:1 char:1
    + Get-Command Get-VM
    + ~~~~~~~~~~~~~~~~~~
        + CategoryInfo          : ObjectNotFound: (Get-VM:String) [    Get-Command], CommandNotFoundException
[...]

PR #1822, opened 2017-12-11
Issue #1472 Pre-flight check to see if 'virtualbox' hypervisor is installed

### LINKED ISSUE #1472: Preflight: check if hypervisor is installed (VBox, libvirt, etc)
For Hyper-V we check if the actual hypervisor is installed (by checking for the management tools `get-vmhost`), but we can also check for `VBoxManage` for VirtualBox and `virsh` for KVM/Libvirt. The check of VirtualBox might have to involve `GOOS`-checks and specific paths as I do not think it gets added to the system paths.

we might have to duplicate some of the effort as done in: https://github.com/docker/machine/blob/47068ae5853d2f7136155b09673ec4153d89972d/drivers/virtualbox/virtualbox_windows.go#L44, https://github.com/docker/machine/blob/47068ae5853d2f7136155b09673ec4153d89972d/drivers/virtualbox/virtualbox_linux.go#L35 and https://github.com/docker/machine/blob/47068ae5853d2f7136155b09673ec4153d89972d/drivers/virtualbox/virtualbox_darwin.go#L25 (seems darwin and linux are both in path)

### Todo
  * [x] KVM / Libvirt check  
    Merged with 2b005fb7ad8a16b00d91b5174a66e8e041c26145
  * [x] HyperV  
    Issue #1429 - af7bffc54d5e030af37a38419756dc82e2a159a0
  * [ ] VirtualBox

### Note
Addresses #1467

anjannath:
Fixes #1472


gbraad:
Will have to try this... since it covers several platforms, this can be a tricky test for review ;-)

Gopkg.lock
```diff
@@ -419,6 +418,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "126ca98597875822b477375d68ee4318150c6ebd9faf1dd5ab368fbe1fe99e9d"
+  inputs-digest = "f41ce71fcc26465ebf6a3b4171311624087e86431cd519fd26c839c4434ab180"
```
LalatenduMohanty: @anjannath Not sure why you need to change the digest.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -374,6 +388,22 @@ func checkHypervDriverUser() bool {
...
+		return false
+	}
+	fi, _ := os.Stat(path)
+	if fi.Mode()&os.ModeSymlink != 0 {
+		path, err = os.Readlink(path)
+		if err != nil {
+			return false
+		}
+	}
+	return true
```
LalatenduMohanty: -1, We do not need Xhyve to be installed on macOS if we have the driver installed as the driver contains the Xhyve also.

Gopkg.lock
anjannath: this happened because of the `registry` package. the function `registry.OpenKey()` is undefined in non windows OS, so i did a `dep ensure` to update the dependency thinking it was version issue. 
I can revert the changes in the file if needed.

cmd/minishift/cmd/config/config.go
```diff
@@ -100,17 +100,20 @@ var (
...
+	SkipCheckXHyveDriver    = createConfigSetting("skip-check-xhyve-driver", SetBool, nil, nil, true)
+	WarnCheckXHyveDriver    = createConfigSetting("warn-check-xhyve-driver", SetBool, nil, nil, true)
+	SkipCheckHyperVDriver   = createConfigSetting("skip-check-hyperv-driver", SetBool, nil, nil, true)
+	WarnCheckHyperVDriver   = createConfigSetting("warn-check-hyperv-driver", SetBool, nil, nil, true)
+	SkipCheckIsoUrl         = createConfigSetting("skip-check-iso-url", SetBool, nil, nil, true)
+	WarnCheckIsoUrl         = createConfigSetting("warn-check-iso-url", SetBool, nil, nil, true)
+	SkipCheckVMDriver       = createConfigSetting("skip-check-vm-driver", SetBool, nil, nil, true)
+	WarnCheckVMDriver       = createConfigSetting("warn-check-vm-driver", SetBool, nil, nil, true)
+	SkipCheckVBoxInstalled  = createConfigSetting("skip-check-vbox-installed", SetBool, nil, nil, true)
+	WarnCheckVBoxInstalled  = createConfigSetting("warn-check-vbox-installed", SetBool, nil, nil, true)
```
LalatenduMohanty: The names of the flag does not look consistent with other similar flag names. I think the purpose of `WarnCheckVBoxInstalle` is similar as `WarnCheckHyperVDriver`.

anjannath: Yes they are similar, theseis flags can be used to skip a preflight test, so i wanted to make it explicit that we are skipping the VBox installation check. hence named it so.

cmd/minishift/cmd/start_preflight.go
anjannath: Yes i didn't know about it, we don't need this check.

gbraad: We would need a check to see if the platform has supported for running a hypervisor... This is the HypervisorKit or similar...

Gopkg.lock
gbraad: We need to run platform specific tests? If so, define these as a new function which can be overriden by a platform specific source file.

cmd/minishift/cmd/start_preflight.go
anjannath: To check if the platform supports VM driver we have a check already here, https://github.com/minishift/minishift/blob/84bbae500a15995e16843e22f67ef1d6c203424e/cmd/minishift/cmd/start_preflight.go#L53-L58

Gopkg.lock
anjannath: I did that, at least i think i did that, waiting for your review.

hferentschik: -1, this change is a change in a generated file. You cannot edit the lock file. You need to modify Gopkg.toml and make sure dependency resolution will work.

anjannath:
i did not edit it, it got automatically generated when i did `dep ensure`

gbraad:
> This is not the correct way to mange dependecies.

Is there a description what the correct way is? If so, please share... If not, we need to create this...

cmd/minishift/cmd/start_preflight.go
hferentschik: you need to check for the existence of _docker-machine-driver-xhyve_ on the path. That's the actual driver which needs to be installed manually and which needs to have the suid bit set.

cmd/minishift/cmd/start_preflight_notwin.go
```diff
@@ -0,0 +1,53 @@
...
+	}
+	fi, _ := os.Stat(path)
+	if fi.Mode()&os.ModeSymlink != 0 {
+		path, err = os.Readlink(path)
+		if err != nil {
+			return false
+		}
+	}
+	// check in the default applications path in macOS
+	if runtime.GOOS == "darwin" {
```
hferentschik: Where is this coming from? I'd argue this is not correct. And why does it matter? Is the central point not to find VBoxManage on the path?

hferentschik:
I am very sceptical about this. I see already a whole bunch of false positives popping up. I don't think we fix an actual problem here. Let's rather focus on aligning with the Minikube driver usage and the ability to manage xhyve driver installation as part of our code. That would provide actual user value and once we go down this road this code would become obsolete.

hferentschik:
> i did not edit it, it got automatically generated when i did dep ensure.

how did this come about? Did you have the lock file removed? As long as the lock file exists `dep ensure` should vendor the locked dependencies. If not, something is wrong here.

cmd/minishift/cmd/start_preflight_notwin.go
anjannath: this check is to ensure that, if virtualbox in mac is removed by removing the virtualbox app from finder, the script `VBoxManage` is still on the path, but cannot access the original `VBoxManage` binary,  which is in the location `/Applications/VirtualBox.app/Contents/MacOS/VBoxManage` to care for that situation this check is added.

This prevents where we find the `VBoxManage` in path but cannot access the actual binary.
The right way to remove virtualbox is by using the unistaller in the dmg image.

gbraad:
> I see already a whole bunch of false positives popping up. 

Such as? Please file issues accordingly... So far the feedback on one of those reports have npot provided a clear answer #1675 that this issue exists.

anjannath:
well the changes came when i wanted to add the `"golang.org/x/sys"` for the `registry` package to Gopkg.toml it already had that dependecy listed but i added again then did `dep ensure` that might have caused it. Anyway i'll update the PR without this file.

cmd/minishift/cmd/start_preflight_notwin.go
hferentschik: This implies it is in the default location. This does not have to be the case. I personally structure my /Application directory instead of flat out adding everything in there. This effectively would brake things on my machine. 

Given that install locations are often configurable and up to the user, I am against doing checks against fixed paths. Checks should be against required binaries on the PATH.

hferentschik: If you want to make sure that _VBoxManage_ works, you can either check whether it is a symbolic link and its target exist or you can execute a known command, eg `VBoxManage -v`. Verify that there is a valid output. This is generic and works for Linux nicely as well, where the _VBosManage_ binary could be a symbolic link as well.

hferentschik:
> Such as? Please file issues accordingly... So far the feedback on one of those reports have not provided a clear answer (eg. #1675) that this issue exists.

The referenced issues is reproducible. It is not just a hypothetical issue. 

> Constructive and detailed feedback please, not just assumptions and wild guesses.

What do you mean? I believe my feedback is very much constructive and to the point. Take the macOS case. In the current pull request, it would break things for non-default installations. I think I explained this quite in detail. And this are the type of issues I am concerned about.

hferentschik:
> We do NOT point to the driver, but libmachine does. So even an automated installation is not a guarantee of success.

That's not correct. libmachine is using the PATH to detect drivers. If we control the installation and know which driver we want to use, we can very much determine which driver gets used, even if the user does a manual install as well.

> Even if we have an installation command, we would have to check the existance of a previous driver... 

Not necessarily (see above). I do not even have an explicit command in mind. I thought we could handle this implicitly. Similar to how we implicitly cache _oc_ or the ISO.

> We can't just assume that the user does the right thing.

This is exactly why I want to do it for him

> He might already have minikube working, or docker-machine. If we install a new driver, we might even cause a conflict.

I don't think so. The whole idea is that we isolate the drivers in MINISHIFT_HOME. If this is done correctly we even reduce the number of potential conflicts.

cmd/minishift/cmd/start_preflight_notwin.go
anjannath: in the machine i tried `VBoxManage' on the path was not a symbolic link but a shell script which would `exec` the `VBoxManage` binary inside Application.
It already checks for it on the path,  and checks if it is a symlink and follows it. 

I added this as another check to really make sure it is there, the situation you mentioned where it is installed in a different location will make this fail. Probably running a command like `VBoxManage -v` is the way to go.

anjannath:
Updating the PR to leave `Gopkg.lock` file and removing the check for `XHyve` on the path as this is already taken care of. Now this PR only checks if VirtualBox is installed in all the platforms we support.

cmd/minishift/cmd/start_preflight_notwin.go
gbraad: When `VBoxManage` is still on the path, this might have been caused by an issue in which the application got removed from `/Applications` on macOS, but not uninstalled. The included driver from Libmachine is actually wrong and uses this as a fallback case.

hferentschik: Let's execute 'VBoxManage -v' then. That's the interaction point between the driver and VirtualBox. Making assumptions about the install location is imo wrong.

gbraad:
> I don't think so. The whole idea is that we isolate the drivers in MINISHIFT_HOME. If this is done correctly we even reduce the number of potential conflicts.

and this is not the behaviour libmachine expects, unless we invoke minishift again with an altered PATH.

cmd/minishift/cmd/start_preflight_notwin.go
hferentschik: So what is the user chooses user only install (not for all users). In this case ~/Applications is used.

hferentschik: > I believe that when xhyve is selected, the driver will eventually fail as it finds VBoxManage... right? This is the behaviour I saw

Fair enough. I am ok with trying to verify whether the VirtualBox install is working. However, this should be done in a way which is not too restrictive and makes wrong assumptions.

cmd/minishift/cmd/start_preflight_notwin.go
```diff
@@ -0,0 +1,46 @@
...
+func checkVBoxInstalled() bool {
+	// check if VBoxManage is really installed by executing a command
+	cmd := exec.Command("VBoxManage", "-v")
+	stdOutStdError, err := cmd.CombinedOutput()
+	if err != nil {
+		return false
+	}
+	stdOut := fmt.Sprintf("%s", stdOutStdError)
+	stdOut = strings.TrimSpace(stdOut)
+	match, err := regexp.MatchString("^[0-9]\\.[0-9]\\.[0-9]+[a-z][0-9]*$", stdOut)
```
hferentschik: Do we really have to match the version string. That seems brittle. It should be enough to check the exit code of the command. If it is '0' we should be good. If not something went wrong. Either not installed at all or some other setup issue.

cmd/minishift/cmd/config/config.go
```diff
@@ -100,17 +100,18 @@ var (
...
+	WarnCheckKVMDriver     = createConfigSetting("warn-check-kvm-driver", SetBool, nil, nil, true)
+	SkipCheckXHyveDriver   = createConfigSetting("skip-check-xhyve-driver", SetBool, nil, nil, true)
+	WarnCheckXHyveDriver   = createConfigSetting("warn-check-xhyve-driver", SetBool, nil, nil, true)
+	SkipCheckHyperVDriver  = createConfigSetting("skip-check-hyperv-driver", SetBool, nil, nil, true)
+	WarnCheckHyperVDriver  = createConfigSetting("warn-check-hyperv-driver", SetBool, nil, nil, true)
+	SkipCheckIsoUrl        = createConfigSetting("skip-check-iso-url", SetBool, nil, nil, true)
+	WarnCheckIsoUrl        = createConfigSetting("warn-check-iso-url", SetBool, nil, nil, true)
+	SkipCheckVMDriver      = createConfigSetting("skip-check-vm-driver", SetBool, nil, nil, true)
+	WarnCheckVMDriver      = createConfigSetting("warn-check-vm-driver", SetBool, nil, nil, true)
+	SkipCheckVBoxInstalled = createConfigSetting("skip-check-vbox-installed", SetBool, nil, nil, true)
```
hferentschik: So really you just add two lines here, right? The other changes are due to adding an additional whitespace for alignment, or?

anjannath: Yes, just added the last two lines, VSCode aligns and hence the other changes.

cmd/minishift/cmd/start_preflight_notwin.go
anjannath: Thanks i'll check the exit code than. This crossed my mind too, when they change the versioning format it'll fail.

hferentschik: Right.

anjannath: @hferentschik Do you know why the integration tests are failing?

anjannath: I spent some time on this today, couldn't find any easy way to get the exit code.

anjannath:
retest this please

cmd/minishift/cmd/config/config.go
hferentschik: ok

cmd/minishift/cmd/start_preflight_notwin.go
hferentschik: Another thing. The only VirtualBox specific thing in this function is the actual command 'VBoxManage -v'. In its current form, we cannot unit test it at all because we cannot really make any assumptions whether VirtualBox is installed.

If you, however, extract a generic _CommandExecutesSuccessfully(cmd []string) bool_ method you can.

_checkVBoxInstalled()_ becomes:

```Go
func checkVBoxInstalled() bool {
    return CommandExecutesSuccessfully([]string{"VBoxManage", "v"})
}
```

_ CommandExecutesSuccessfully_ could be placed in one of your util packages. Most importantly, it can be tested. We can execute it with:

- CommandExecutesSuccessfully([]string{"thiscommanddoesexist"})
- CommandExecutesSuccessfully([]string{"pwd"})

With a bit of setup code, we can create a temp directory and add a dummy file and make it executable. Then add the temp directory to the PATH and execute _CommandExecutesSuccessfully_ against this path. 

Bottom line, one can test pretty much all possible scenarios of _CommandExecutesSuccessfully_ given confidence in that this method works.

cmd/minishift/cmd/start_preflight_windows.go
```diff
@@ -0,0 +1,83 @@
...
+	if filehelper.Exists(filepath.Join("C:", "Program Files", "Oracle", "VirtualBox", vboxCmd)) {
+		return true
+	}
+	// Look in windows registry
+	if p, err := findVBoxInstallDirInRegistry(); err != nil {
+		if filehelper.Exists(filepath.Join(p, vboxCmd)) {
+			return true
+		}
+	}
+	return false
```
hferentschik: What happened to the fallback? This is not at all the same as in virtualbox_windows.go. There the methods ends with - `return detectVBoxManageCmdInPath() //fallback to path`. This is not here at all, making your code too stringent again.

anjannath: Yes @hferentschik the fallback to checking in path is not present here. Will add it.

cmd/minishift/cmd/start_preflight_notwin.go
anjannath: The `GetExitCode()` func in `docker/pkg/system` doesn't seem to work as expected, it returns 0 all the time. However i found another piece of code on [stackoverflow](https://stackoverflow.com/a/40770011). Will refactor this and use it, also going to implement  _CommandExecutesSuccessfully()_ but it will also return the exitcode.

cmd/minishift/cmd/util/util.go
```diff
@@ -112,3 +115,44 @@ func IsValidProfileName(profileName string) bool {
 	rr := regexp.MustCompile(`^[a-zA-Z0-9]+[\w+-]*$`)
 	return rr.MatchString(profileName)
 }
+
+// RunAndGetExitCode returns the exit code of a command
+func RunAndGetExitCode(command string, args ...string) int {
```
hferentschik: I don't think this is a _cmd_ specific util method. It should probably go somewhere under _pkg/util_, since it is useful in other cases as well.

cmd/minishift/cmd/util/util_test.go
```diff
@@ -48,3 +48,36 @@ func TestIsValidProfileName(t *testing.T) {
...
 	}
 }
+
+func TestRunAndGetExitCode(t *testing.T) {
+	testData := []struct {
+		command  string
+		exitCode int
+	}{
+		{"blahh", 1},
+		{"make", 2},
```
hferentschik: Why do you expect '2' here? How do you know whether or not 'make' is installed?

cmd/minishift/cmd/util/util.go
```diff
@@ -112,3 +115,44 @@ func IsValidProfileName(profileName string) bool {
...
+
+// CommandExecutesSuccessfully returns true if the command executed based on the exit code
+func CommandExecutesSuccessfully(cmd string, args ...string) bool {
+	// check in the path if cmd	is present
+	path, err := exec.LookPath(cmd)
+	if err != nil {
+		return false
+	}
+	fi, _ := os.Stat(path)
+	if fi.Mode()&os.ModeSymlink != 0 {
```
hferentschik: What is the purpose of this? Why even have this check?

cmd/minishift/cmd/util/util.go
hferentschik: That said, there is already the Runner interface under _pkg/util/runner.go_. It looks like it already does what we need. No need to re-invent the wheel.

anjannath: Yes, earlier i thought it had extra stuff cause we use it for `OcRunner` but it does the same thing, i will modify this to make use of `Runner` here.

cmd/minishift/cmd/util/util.go
anjannath: To check if the binary in the `PATH` is a symlink, and check if the target exists.

cmd/minishift/cmd/util/util_test.go
anjannath: when make is run without there being a `Makefile` it returns error code 2, well to run the tests we do `make test` so i figured `make` has to be installed... is that a bad assumption?

cmd/minishift/cmd/util/util.go
hferentschik: +1

pkg/util/utils_test.go
```diff
@@ -136,3 +139,32 @@ func TestFriendlyDuration(t *testing.T) {
...
+		{"blahh", false},
+		{dummybinary, true},
+	}
+	for _, v := range testData {
+		got := CommandExecutesSuccessfully(v.command)
+		if got != v.exists {
+			t.Errorf("Expected %v Got %v", v.exists, got)
+		}
+	}
+}
```
anjannath: This test should fail, as the `dummybinary*` is not present in the `PATH` if `exec.LookPath()` only looks for the file in `PATH`, maybe it behaves differently when absolute path is passed.

.gitignore
```diff
@@ -24,6 +24,9 @@ _testmain.go
 *.test
 *.prof
 
+#don't ignore exe under test/testdata
```
hferentschik: I don't think this is a good idea. The ignore file gets more and more complicated. Really. there is no need to exclude "*.exe" from the start. The only executable ever build dynamically is in the _out_ directory which is already excluded. 

So instead of making things more complicated, start making things simpler and remove the '*.exe' exclusion. 

Mind you, we inherited the ignore file from Minikube and never really took the time to go through it. There are several entries in which could be removed.

cmd/minishift/cmd/start_preflight_windows.go
```diff
@@ -0,0 +1,76 @@
...
+	// Look in default installation path for VirtualBox version > 5
+	if filehelper.Exists(filepath.Join("C:", "Program Files", "Oracle", "VirtualBox", vboxCmd)) {
+		return true
+	}
+	// Look in windows registry
+	if p, err := findVBoxInstallDirInRegistry(); err != nil {
+		return filehelper.Exists(filepath.Join(p, vboxCmd))
+	}
+	// fallback to search in the path and execute
+	return util.CommandExecutesSuccessfully("VBoxManage", "-v")
```
hferentschik: 👍

pkg/util/runner_test.go
```diff
@@ -0,0 +1,47 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
```
hferentschik: For new files from our side, let's use the Red Hat license header

pkg/util/runner_test.go
```diff
@@ -0,0 +1,47 @@
...
+	"runtime"
+	"testing"
+)
+
+func Test_Runner_Run(t *testing.T) {
+	var runner Runner = &RealRunner{}
+	var stdOut, stdErr io.Writer
+
+	currentDir, _ := os.Getwd()
+	dummybinary := filepath.Join(currentDir, "..", "..", "test", "testdata", "dummybinary")
```
hferentschik: I don't think this is quite right as test. You want to just pass the command name, not the fully specified path. If you do so, you are bypassing the whole lookup  part. 

Pass just the binary name and without additional modifications, it should not work. Then in a second test, adjust the PATH first and then it should work.

BTW, what binaries are we talking about here? What do they do?

hferentschik: I think it would be better to have a _testdata_ directory under _pkg/util_ directly. Adding all testdata under the same directory does not really scale so well. For some terst data which is shared between many tests it might make sense, but otherwise I think we should use testdata local to the test. Golang ignores directories called _testdata_ for the actual binary build. It is only used for tests. See also _/pkg/minishift/docker/image/testdata_.

pkg/util/utils.go
```diff
@@ -166,3 +168,26 @@ func FriendlyDuration(d time.Duration) time.Duration {
 	d2 := (d / time.Nanosecond) * (time.Nanosecond)
 	return d2
 }
+
+// CommandExecutesSuccessfully returns true if the command executed based on the exit code
+func CommandExecutesSuccessfully(cmd string, args ...string) bool {
+	var runner Runner = &RealRunner{}
+	var stdOut, stdErr io.Writer
+	// check in the path if cmd	is present
+	path, err := exec.LookPath(cmd)
```
hferentschik: This is not needed. The underlying _exec.Command_ of the _Runner_ takes care of this.

pkg/util/utils.go
```diff
@@ -166,3 +168,26 @@ func FriendlyDuration(d time.Duration) time.Duration {
...
+	if err != nil {
+		return false
+	}
+	fi, _ := os.Stat(path)
+	if fi.Mode()&os.ModeSymlink != 0 {
```
hferentschik: There is not need for this check either. All required is to do a _runner.Run_ and check the error code.

pkg/util/utils_test.go
hferentschik: See comment above.

pkg/util/runner_test.go
anjannath: Okay

.gitignore
anjannath: Sure, i was also thinking the same and could not find any '.exe' files elsewhere, but wanted to keeo things the way they were.

pkg/util/runner_test.go
anjannath: > BTW, what binaries are we talking about here? What do they do?

The binaries are dead simple they just print "Minishift", using `fmt.Println()`, just to check the exitcode returned by `Runner`.

pkg/util/utils.go
anjannath: Okay

pkg/util/runner_test.go
anjannath: > Pass just the binary name and without additional modifications, it should not work. Then in a second test, adjust the PATH first and then it should work.

Just to make sure i understand what you mean, you want the _command_ to be directly passed to `Runner`? But `Runner` takes the full path, i did try passing only the _command_ and faced some issue and @praveenkumar helped me to understand that it takes the commnad path and then it converts it to a absolute path.

pkg/util/utils_test.go
anjannath: Will fix this as per this [comment](https://github.com/minishift/minishift/pull/1822/files#r158220129).

pkg/util/runner_test.go
hferentschik: Oh, I see. I guess we are talking about `path, _ := filepath.Abs(commandPath)`. Actually, I would consider this a bug or at least a bad implementation choice. I don't think we need to make an absolute path at this point.  It prevents us from utilizing the full functionality of exec.Command.

Also, for the case we use it, we already have the full path. Mostly we need it for `oc` and there we get the full path from instance config. 

I'd recommend removing this line creating the absolute path.

hferentschik: > Just to make sure i understand what you mean, you want the command to be directly passed to Runner? 

Yes

> But Runner takes the full path, it doesn't do a exec.LookPath()

It should. See comment above. Also, check the Godoc of `exec.Command`: "If name contains no path separators, Command uses LookPath to resolve name to a complete path if possible. Otherwise it uses name directly as Path"

pkg/util/utils.go
hferentschik: See comment above ;-)

pkg/util/runner_test.go
anjannath: Thanks @hferentschik will make the change in `Runner`

anjannath: This doesn't seem to help, for some reason `Runner` is not returning the correct exitcode when its called from the unit test function.

However it works fine in the actual Pre-flight check.

hferentschik: > This doesn't seem to help, for some reason Runner is not returning the correct exitcode when its called from the unit test function.

So you are removing the line _path, _ := filepath.Abs(commandPath)_, right? And the test is not returning what you are after? What is it returning and what do you expect? 

Note, the test here is not quite the right if you want to test the lookup via the path. You are building an absolute path, so the lookup won't be made. There are several test cases you need to cover.

pkg/util/utils_test.go
hferentschik: So right, these tests are not correct.

pkg/util/runner_test.go
anjannath: > So you are removing the line path, _ := filepath.Abs(commandPath), right? And the test is not returning what you are after? What is it returning and what do you expect?

Yes and yes, its not returning the correct exit code, returns 1 every time.

> Note, the test here is not quite the right if you want to test the lookup via the path. You are building an absolute path, so the lookup won't be made.

Yes and that is the issue, the _exec.Command()_ and _exec.LookPath()_ is behaving weired in case of the unit tests.

hferentschik:
@anjannath Not sure what your problems are. Here is a pull request against your code -https://github.com/anjannath/minishift/pull/2 - with some cleanup and the changes I am trying to explain. For all I can tell all seems to be working as expected. What am I missing?

pkg/util/runner_test.go
hferentschik: https://github.com/anjannath/minishift/pull/2

anjannath: The changes in the PR [anjannath#2](https://github.com/anjannath/minishift/pull/2) fixed the issue, i was using _ls_ and _dir_ for the command lookup test, in place of adding the dummybinaries to the PATH, and that wasn't working.

anjannath:
Thanks @hferentschik i have incorporated your commit in this PR. now this is working as expected.

anjannath:
retest this please

pkg/util/utils_test.go
```diff
@@ -136,3 +139,87 @@ func TestFriendlyDuration(t *testing.T) {
...
+		dummyBinaryPath = filepath.Join(dummyBinaryPath, "dummybinary_windows.exe")
+	default:
+		t.Fatal("Unpexpected OS")
+	}
+
+	testData := []struct {
+		command string
+		exists  bool
+	}{
+		{"blahh", false},
```
hferentschik: Let's make this an absolute path to fit the test,

    {filepath.Join(currentDir, "..", "..", "test", "testdata", "blahh"), false},

Other than that, I am ok now.

hferentschik:
retest this please

hferentschik:
@anjannath, seems CentOS CI tests is failing due to a broken link to CentOS CI nightly.

cmd/minishift/cmd/start_preflight_notwin.go
```diff
@@ -0,0 +1,25 @@
+// +build !windows
```
LalatenduMohanty: This is not intended I think. Can you please remove it?

pkg/util/utils_test.go
anjannath: Okay

cmd/minishift/cmd/start_preflight_notwin.go
anjannath: This is needed, to prevent it from getting compiled in case of building for windows.

hferentschik: It is needed. There are two OS specific versions of _checkVBoxInstalled_

anjannath:
Yes @hferentschik i have tested it on all the three platforms and it worked for me.

LalatenduMohanty:
@anjannath @hferentschik When I am testing the PR I am seeing

```diff
diff --git a/Gopkg.lock b/Gopkg.lock
index 007f3a4b..4d3d9a15 100644
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -381,10 +381,9 @@
   revision = "442624c9ec9243441e83b374a9e22ac549b5c51d"
 
 [[projects]]
-  branch = "master"
   name = "golang.org/x/sys"
   packages = ["unix","windows","windows/registry"]
-  revision = "53aa286056ef226755cd898109dbcdaba8ac0b81"
+  revision = "43e60d72a8e2bd92ee98319ba9a384a0e9837c08"
 
 [[projects]]
   branch = "master"
@@ -418,6 +417,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "6d90e09d82cfe1cea527bbbc9753c1e9ad8ca2b7e2f567e1e2809767fb017df2"
+  inputs-digest = "0aef932b275def0df646c10b2df61d5070a5678ec76e6ae33f019283dab85361"
   solver-name = "gps-cdcl"
   solver-version = 1
```
We have seen similar in https://github.com/minishift/minishift/pull/1859#discussion_r158943333 . The sha changed because of the use of `"golang.org/x/sys"` (interestingly we have sys as dependency but was not used).

cmd/minishift/cmd/start_preflight.go
```diff
@@ -109,6 +109,14 @@ func preflightChecksBeforeStartingHost() {
 			"Checking if user is a member of the Hyper-V Administrators group",
 			false, configCmd.WarnCheckHyperVDriver.Name,
 			driverErrorMessage)
+	case "virtualbox":
+		preflightCheckSucceedsOrFails(
+			configCmd.SkipCheckVBoxInstalled.Name,
+			checkVBoxInstalled,
+			"Checking if VirtualBox is installed",
+			false, configCmd.WarnCheckVBoxInstalled.Name,
+			driverErrorMessage)
```
LalatenduMohanty: The driver error message is not appropriate in this case.
```
driverErrorMessage := "See the 'Setting Up the Driver Plug-in' topic (https://docs.openshift.org/latest/minishift/getting-started/setting-up-driver-plugin.html) for more information"
```
As it points to the driver plugin documentation which do not have any information on Virtualbox. It should point to https://docs.openshift.org/latest/minishift/getting-started/installing.html#install-prerequisites

anjannath: Yeah, this should be changed, changing the error message. and about [#1822 comment](https://github.com/minishift/minishift/pull/1822#issuecomment-354450507) the `windows/registry` package is used from`"golang.org/x/sys"`, interestigly i don't see the changes to `gopkg.lock` in my local branch. How do i take care of this?

anjannath:
retest this please

cmd/minishift/cmd/start_preflight.go
```diff
@@ -49,6 +49,7 @@ const (
 // preflightChecksAfterStartingHost is executed before the startHost function.
 func preflightChecksBeforeStartingHost() {
 	driverErrorMessage := "See the 'Setting Up the Driver Plug-in' topic (https://docs.openshift.org/latest/minishift/getting-started/setting-up-driver-plugin.html) for more information"
+	vBoxErrorMessage := "See the 'Installing Minishift' topic (https://docs.openshift.org/latest/minishift/getting-started/installing.html#install-prerequisites) for more information"
```
LalatenduMohanty: s/Installing Minishift/Installing prerequisites for Minishift/

LalatenduMohanty: s/vBoxErrorMessage/prerequisiteErrorMessage/

anjannath: Making the changes.

LalatenduMohanty:
Thanks @anjannath

Issue #1834, opened 2017-12-12
Set explicit version of osext dependency


hferentschik:

LalatenduMohanty:
Resolved via #1835

PR #1835, opened 2017-12-12
Issue #1834 Adding explicit revision for osext dependency

hferentschik:


Gopkg.toml
```diff
@@ -70,6 +70,7 @@ ignored = ["github.com/Sirupsen/logrus"]
 
 [[constraint]]
   name = "github.com/kardianos/osext"
+  revision = "ae77be60afb1dcacde03767a8c37337fad28ac14"
```
praveenkumar: LGTM 👍 , since the package doesn't have any release so revision is better.

Issue #1816, opened 2017-12-07
Add a retry logic to the verifyHTTPResponse() step


agajdosi:
This is upstream follow up issue for the [CDK-222](https://issues.jboss.org/browse/CDK-222). In combination of slower host and more complicated application on OpenShift, the time for the http endpoint of the application to be accessible is longer and thus some tests may fail.

To fix this behaviour and secure that the result reported from tests is relevant and true, we should add a retry logic to this step, so when it takes more time for app to be ready, the tests will wait and try it again in reasonable time.

gbraad:
I do not think a retry logic is correct. If a 500 or 503 is returned, the service failed. Bashing it another time is just not a proper test. I'd rather see an explicit backing off (sleep) as step or part of a deployment... WDYT?

agajdosi:
@gbraad I was thinking about sleep also, but how much we want to have? 30 seconds? In most cases the tests passes just fine as they are now, on some machines maybe just 3 or 5 seconds would help. On the machine I am facing a problem 5 second sleep helps. So I understand the retry as more sophisticated sleep in this case.

But well, I understand also that having the sleep mentioned directly in the Gherkin feature maybe is what we want to have, to be transparent and understandable.

hferentschik:
Sleeping is never a good option. Does this demo use a readiness probe? If so, should the rollout not wait for the readiness probe to report success. In this case it should be save to hit the endpoint. 

I think before adding random sleeps we should investigate what's going on here.

gbraad:
it is not a 'sleep' that needs to happen, but an ease-off. It seems the service is slow to setup correctly. So when roll-out succeeded, the service is not available yet?

retriggering the service seems wrong... so, what about a health check before checking the services or as suggested maybe a readiness-probe? Although, these might lead to the same result I guess.

agajdosi:
> retriggering the service seems wrong (which was suggested before)

No @gbraad, I have never suggested to add a redeployment of the service/app. I have suggested to add a retry logic to the the verifyHTTPResponse() which is the step that checks the http response. If the response is wrong, the step should wait a little bit and try once more.

> Does this demo use a readiness probe? If so, should the rollout not wait for the readiness probe to report success.

@hferentschik Did not check yet. My first plan was to add a retry logic to http response check, so one does not need to care if the application she wants to test use the readiness probes or not. But I agree, the readiness probe check could be part of the rollout step - if it is possible to handle the applications which do not use the probe.

[...]

PR #1808, opened 2017-12-06
Issue #1776 Moving default Origin version to v3.7.1

LalatenduMohanty:
Fix #1776

PR #1838, opened 2017-12-12
Issue #1816 - Add a retry logic to integration tests

agajdosi:
Addresses issue #1816

### DIFF
--- a/test/integration/features/addon-xpaas.feature
+++ b/test/integration/features/addon-xpaas.feature
@@ -19,9 +19,10 @@ which are then available in OpenShift to the user.
   Scenario Outline: User deploys, checks out and deletes several templates from XpaaS imagestream
    Given Minishift has state "Running"
     When executing "oc new-app <template-name>" succeeds
-     And services "<service-name>" rollout successfully
-    Then body of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" contains "<expected-hello>"
-     And status code of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" is equal to "200"
+     And executing "oc set probe dc/<service-name> --readiness --get-url=http://:8080<http-endpoint>" succeeds
+     And service "<service-name>" rollout successfully
+    Then with up to "10" retries with wait period of "500ms" the "body" of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" contains "<expected-hello>"
+     And with up to "10" retries with wait period of "500ms" the "status code" of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" is equal to "200"
      And executing "oc delete all --all" succeeds
 
   Examples: Required information to test the templates
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -35,12 +35,12 @@ Feature: Basic
     and OpenShift web console should be accessible.
 
     Given Minishift has state "Running"
-     When status code of HTTP request to "OpenShift" at "/healthz" is equal to "200"
-     Then body of HTTP request to "OpenShift" at "/healthz" contains "ok"
-      And status code of HTTP request to "OpenShift" at "/healthz/ready" is equal to "200"
-      And body of HTTP request to "OpenShift" at "/healthz/ready" contains "ok"
-      And status code of HTTP request to "OpenShift" at "/console" is equal to "200"
-      And body of HTTP request to "OpenShift" at "/console" contains "<title>OpenShift Web Console</title>"
+     When "status code" of HTTP request to "/healthz" of OpenShift instance is equal to "200"
+     Then "body" of HTTP request to "/healthz" of OpenShift instance contains "ok"
+      And "status code" of HTTP request to "/healthz/ready" of OpenShift instance is equal to "200"
+      And "body" of HTTP request to "/healthz/ready" of OpenShift instance contains "ok"
+      And "status code" of HTTP request to "/console" of OpenShift instance is equal to "200"
+      And "body" of HTTP request to "/console" of OpenShift instance contains "<title>OpenShift Web Console</title>"
 
   Scenario Outline: User can set, get, view and unset values in configuration file
     User is able to setup persistent configuration of Minishift using "config" command
@@ -143,10 +143,12 @@ Feature: Basic
     When executing "oc login --username=developer --password=developer" succeeds
      And executing "oc new-project ruby" succeeds
      And executing "oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-ex.git" succeeds
-     And services "ruby-ex" rollout successfully
      And executing "oc expose svc/ruby-ex" succeeds
-    Then status code of HTTP request to "/" of service "ruby-ex" in namespace "ruby" is equal to "200"
-     And body of HTTP request to "/" of service "ruby-ex" in namespace "ruby" contains "Welcome to your Ruby application on OpenShift"
+     And executing "oc set probe dc/ruby-ex --readiness --get-url=http://:8080" succeeds
+     And service "ruby-ex" rollout successfully
+
+    Then with up to "10" retries with wait period of "500ms" the "status code" of HTTP request to "/" of service "ruby-ex" in namespace "ruby" is equal to "200"
+     And with up to "10" retries with wait period of "500ms" the "body" of HTTP request to "/" of service "ruby-ex" in namespace "ruby" contains "Welcome to your Ruby application on OpenShift"
 
     When executing "oc delete project ruby" succeeds
     Then stdout should contain "project "ruby" deleted"
--- a/test/integration/features/cmd-openshift.feature
+++ b/test/integration/features/cmd-openshift.feature
@@ -7,9 +7,9 @@ cluster in VM provided by Minishift.
      Given Minishift has state "Does Not Exist"
       When executing "minishift openshift service list" succeeds
       Then stdout should contain
-      """
-      Running this command requires an existing 'minishift' VM, but no VM is defined.
-      """
+       """
+       Running this command requires an existing 'minishift' VM, but no VM is defined.
+       """
 
   Scenario: Minishift start
   Minishift must be started in order to interact with OpenShift via "minishift openshift" command
@@ -35,9 +35,9 @@ cluster in VM provided by Minishift.
   Scenario: User deploys nodejs example application from OpenShift repository
       When executing "oc new-app https://github.com/openshift/nodejs-ex -l name=myapp" succeeds
       Then stdout should contain
-      """
-      Run 'oc status' to view your app.
-      """
+       """
+       Run 'oc status' to view your app.
+       """
 
   @minishift-only
   Scenario: Getting information about OpenShift and kubernetes versions
@@ -45,11 +45,11 @@ cluster in VM provided by Minishift.
      Given Minishift has state "Running"
       When executing "minishift openshift version" succeeds
       Then stdout should match
-      """
-      ^openshift v[0-9]+\.[0-9]+\.[0-9]+\+[0-9a-z]{7}
-      kubernetes v[0-9]+\.[0-9]+\.[0-9]+\+[0-9a-z]{10}
-      etcd [0-9]+\.[0-9]+\.[0-9]+
-      """
+       """
+       ^openshift v[0-9]+\.[0-9]+\.[0-9]+\+[0-9a-z]{7}
+       kubernetes v[0-9]+\.[0-9]+\.[0-9]+\+[0-9a-z]{10}
+       etcd [0-9]+\.[0-9]+\.[0-9]+
+       """
 
   Scenario: Getting address of internal docker registry
   Prints the host name and port number of the OpenShift registry to the standard output.
@@ -81,9 +81,9 @@ cluster in VM provided by Minishift.
   Scenario: User creates route to the service
       When executing "oc expose svc/nodejs-ex" succeeds
       Then stdout should contain
-      """
-      route "nodejs-ex" exposed
-      """
+       """
+       route "nodejs-ex" exposed
+       """
 
   Scenario: Getting existing service with route
       When executing "minishift openshift service nodejs-ex" succeeds
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -35,6 +35,7 @@ import (
 	"strconv"
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/DATA-DOG/godog"
 	"github.com/DATA-DOG/godog/gherkin"
@@ -192,7 +193,7 @@ func FeatureContext(s *godog.Suite) {
 		variableShouldNotBeEmpty)
 
 	// steps for rollout check
-	s.Step(`^services "([^"]*)" rollout successfully$`,
+	s.Step(`^services? "([^"]*)" rollout successfully$`,
 		minishift.rolloutServicesSuccessfully)
 
 	// steps for proxying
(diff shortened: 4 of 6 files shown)


test/integration/features/addon-xpaas.feature
```diff
@@ -19,7 +19,8 @@ which are then available in OpenShift to the user.
   Scenario Outline: User deploys, checks out and deletes several templates from XpaaS imagestream
    Given Minishift has state "Running"
     When executing "oc new-app <template-name>" succeeds
-     And services "<service-name>" rollout successfully
+     And service "<service-name>" rollout successfully
+     And user waits "5" seconds
```
hferentschik: why do you need a sleep? How do you know that sleeping this time interval will work? we explicitly avoided to use sleep.

hferentschik:
Instead of a sleep have you investigated adding readiness probes?

gbraad:
> Sleeps are never a good idea.

sleeps are workarounds 👍 although more explicit, they seldom 'solve' the problem.
It is likely used here as an ease-off period...

@agajdosi what is the actual problem... ?

agajdosi:
@hferentschik @gbraad The problem is that tests are to fast for the OCP in CDK - on same machines and sometimes. However the chance on Jenkins slaves got quite high and the tests are failing - basic.feature was removed from the pipeline and addon-xpaas.feature can't be added ([CDK-224](https://issues.jboss.org/browse/CDK-224), [CDK-222](https://issues.jboss.org/browse/CDK-222)).

hferentschik:
> The problem is that tests are to fast for the OCP in CDK

This sounds very much that there is no readiness probe in place. 

The problem with sleep is always that it is a guess. You also tend to use something which somewhere lies at the boundary between passing and failing. You don't want to waste time waiting, but you also need it to pass. Any timeout becomes then prone to unnecessary test failures. 

Even if readiness probes don't cut it, I am against single sleep. In this case, a retry is a much safer option. It allows keeping wasted time low while also adding an appropriate buffer for the overall timeout. We have the retry step already defined, but luckily we don't have to use it atm (at least what I remember).

test/integration/features/addon-xpaas.feature
```diff
@@ -19,7 +19,8 @@ which are then available in OpenShift to the user.
   Scenario Outline: User deploys, checks out and deletes several templates from XpaaS imagestream
    Given Minishift has state "Running"
     When executing "oc new-app <template-name>" succeeds
-     And services "<service-name>" rollout successfully
+     And executing "oc set probe dc/<service-name> --readiness --get-url=http://:8080<http-endpoint>" succeeds
```
hferentschik: 👍 Does this help to pass the test on the slow machine?

test/integration/features/basic.feature
```diff
@@ -143,8 +143,10 @@ Feature: Basic
     When executing "oc login --username=developer --password=developer" succeeds
      And executing "oc new-project ruby" succeeds
      And executing "oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-ex.git" succeeds
-     And services "ruby-ex" rollout successfully
      And executing "oc expose svc/ruby-ex" succeeds
+     And executing "oc set probe dc/ruby-ex --readiness --get-url=http://:8080" succeeds
```
hferentschik: 👍

test/integration/features/cmd-openshift.feature
```diff
@@ -79,11 +79,15 @@ cluster in VM provided by Minishift.
       Then stderr should contain "You must specify the name of the service."
 
   Scenario: User creates route to the service
+     Given executing "oc set probe dc/node-js --readiness --get-url=http://:8080" succeeds
```
hferentschik: Typo - dc/nodejs-ex not dc/node-js

agajdosi: @hferentschik thanks

test/integration/features/addon-xpaas.feature
agajdosi: @hferentschik Unfortunately not, it now also started to fail on centos-ci.

agajdosi:
@gbraad I have updated the branch, it is now using the health checks in support with retries. Their timeouts are set to 100ms, 400ms and 2500ms. These should give us some image of how long it take for the service to be accessible after rollout. The retries are being reported to console. We can make the delays shorter if all goes well.

Is it accessible for you this way?

Makefile
```diff
@@ -216,6 +216,7 @@ integration_all: $(MINISHIFT_BINARY)
 
 .PHONY: integration_pr
 integration_pr: GODOG_OPTS = --tags=~coolstore\&\&~addon-xpaas
+integration_pr: GODOG_OPTS = --tags=addon-xpaas,basic,cmd-openshift
```
agajdosi: this is just for testing purposes, will be removed later

test/integration/features/addon-xpaas.feature
LalatenduMohanty: @agajdosi I am confused by your comment, do you mean manually readiness probe is working fine but not in integration tests?

test/integration/integration_test.go
```diff
@@ -725,6 +730,22 @@ func succeedsOrFails(execute commandRunner, command string, expectedResult strin
...
+		err = verifyHTTPResponse(partOfResponse, url, urlSuffix, assertion, expected)
+		if err == nil {
+			break
+		}
+		fmt.Printf("HTTP check (%v) has failed, trying again in %v ms. Error: %v\n", i+1, delay[i], err)
+		time.Sleep(time.Millisecond * time.Duration(delay[i]))
+	}
+
+	return err
+}
```
LalatenduMohanty: @agajdosi This should get exposed to the cucumber test steps i.e. when we read through tests it should be clear that we have some waiting time with n times of retry. Then it would be transparent to folks who are writing tests.Because from the test we should be able to control the total time we are waiting.  See below as an similar example. I will give it more thought and we can discuss how we want to achive that. 

```
/ steps to execute `oc` commands
	s.Step(`^executing "oc (.*)" retrying (\d+) times with wait period of (\d+) seconds$`,
		minishift.executingRetryingTimesWithWaitPeriodOfSeconds)
```

test/integration/features/addon-xpaas.feature
agajdosi: @LalatenduMohanty TBH I can't tell right now. One possibility is that integration tests plays a role, another is that I was just unlucky when trying to replicate, third is that the it has something to do with network and time of the day.

The script I use:
```
oc new-app centos/ruby-22-centos7~https://github.com/openshift/ruby-ex.git
oc expose svc/ruby-ex
oc set probe dc/ruby-ex --readiness --get-url=http://:8080
oc rollout status deploymentconfig ruby-ex --watch
curl http://ruby-ex-myproject.$(minishift ip).nip.io
```

will try it today once more and also on Windows.

test/integration/integration_test.go
agajdosi: @LalatenduMohanty Ack.

test/integration/integration_test.go
```diff
@@ -725,11 +736,57 @@ func succeedsOrFails(execute commandRunner, command string, expectedResult strin
...
+			break
+		}
+		if i == retryCount {
+			fmt.Printf("HTTP check (%v) has failed. Error: %v\n", i+1, err)
+			break
+		}
+
+		delay := (i + 1) * retryMaxOverallTime / divisor
+		fmt.Printf("HTTP check (%v) has failed, trying again in %v ms. Error: %v\n", i+1, delay, err)
+		time.Sleep(time.Millisecond * time.Duration(delay))
```
LalatenduMohanty: @agajdosi I think we would be fine if the the unit is second. Millisecond is too granular.

test/integration/features/addon-xpaas.feature
```diff
@@ -19,9 +19,10 @@ which are then available in OpenShift to the user.
   Scenario Outline: User deploys, checks out and deletes several templates from XpaaS imagestream
    Given Minishift has state "Running"
     When executing "oc new-app <template-name>" succeeds
-     And services "<service-name>" rollout successfully
-    Then body of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" contains "<expected-hello>"
-     And status code of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" is equal to "200"
+     And executing "oc set probe dc/<service-name> --readiness --get-url=http://:8080<http-endpoint>" succeeds
+     And service "<service-name>" rollout successfully
+    Then with up to "3" retries in "1000ms" the "body" of HTTP request to "<http-endpoint>" of service "<service-name>" in namespace "myproject" contains "<expected-hello>"
```
LalatenduMohanty: /retries in/retries with wait period of/

test/integration/integration_test.go
agajdosi: @LalatenduMohanty I agree seconds are friendlier but here we are really handling processes which happen it time frame of milliseconds. But if needed I can add a converter so it can be defined in both seconds and milliseconds.

test/integration/features/addon-xpaas.feature
agajdosi: @LalatenduMohanty ack, fixing it now

test/integration/util/util.go
```diff
@@ -80,6 +81,11 @@ func runCommand(command string, commandPath string) (stdOut string, stdErr strin
 	return
 }
 
+func WaitForNSeconds(seconds int) error {
```
gbraad: Why wrapping this up in a named function when it is just a sleep, especially since there is only one usecase for this.

test/integration/util/util.go
```diff
@@ -80,6 +81,11 @@ func runCommand(command string, commandPath string) (stdOut string, stdErr strin
...
+	time.Sleep(time.Duration(seconds) * time.Second)
+	return nil
```
gbraad: return nil... while the function has error as return value.
why have a value in the first place? or just place the sleep inline instead.

gbraad:
I agree with @hferentschik , as I believe there is something else wrong here.
As said, any sleep or retry logic feels like a workaround.

What is our aim with this fix? It feels like we are trying to implememt something make the test pass, ... while to me the actual problem is somewhere else. This wasn't needed before.

hferentschik:
> I agree with @hferentschik , as I believe there is something else wrong here.
As said, any sleep or retry logic feels like a workaround.

> What is our aim with this fix? It feels like we are trying to implememt something to make the test pass, ... while to me the actual problem is somewhere else. This wasn't needed before.

I fully agree.

LalatenduMohanty:
@gbraad @hferentschik I am in the favor for asking OpenShift team about why we need this deplay. However I tired to reproduce the issue but failed as manually things are fine. So I am in favor of the retry at this point  but we can keep discussing the issue with OpenShift team and if it get resolved then we can update our tests.

test/integration/util/util.go
agajdosi: If I am not mistaken, Godog requires return value of type error. Will check that.

gbraad: From what I see, it is not godog calling this function (as part of a step definition), so I do not think this is necessary.

agajdosi:
I have created an issue for this in Origin tracker now: https://github.com/openshift/origin/issues/17934. The reason it was not before was that I have tried to reproduce it manually by a script to help the Origin team to understand, but I was not lucky with that.

> It feels like we are trying to implememt something to make the test pass, ...

If the issue was reproducible manually I would not suggest changes to tests.

LalatenduMohanty:
@hferentschik @gbraad Also it this point we do not know if the issue is from "oc cluster up" side or from OpenShift in general.  IMO it should not be a blocker for moving to v3.7.0 as it is out already as without this PR I do not see how  the integration tests will pass for v3.7.0.

budhrg:
retest this please

gbraad:
I'd rather have the test fail/disabled than adding a workaround to make it pass,as this would give the wrong impression of a fix

agajdosi:
@gbraad Removing affected scenarios gives us zero information about the issue, accepting there 500ms, 1s or 5s of maximum timeout (I prefer lower variants) will give us at least the info if the issue is resolved after first 100ms retry or after 50th retry. (And as this is not happening on all machines all the time, it would really help us I think to get more information about the issue.) However both solutions share the same conceptual background: we are updating the tests to pass, disabling or adding timeout, does not matter imho.

The only and true contrary solution to the 2 above is to call this blocker for #1808 and until https://github.com/openshift/origin/issues/17934 is fixed do not merge the PR and stay with oc v3.6. For that lets just close this issue.

budhrg:
retest this please

agajdosi:
Rebased.

budhrg:
@LalatenduMohanty @praveenkumar @gbraad If all agree then, we can merge this PR. CI is passing consistently now.

test/integration/util/util.go
LalatenduMohanty: @agajdosi can please fix this? the `return nil` is not required and error return type is not required.

LalatenduMohanty:
@minishift/minishift-dev I am planning to merge this PR after @agajdosi fixes one review comment. Once the PR is merged I will create another issue to track removal for the retry for the tests once it  is not required.

budhrg:
> I am planning to merge this PR after @agajdosi fixes one review comment. Once the PR is merged I will create another issue to track removal for the retry for the tests once it is not required.

:+1:

test/integration/util/util.go
agajdosi: With no return value:

```go
func WaitForNSeconds(seconds int) {
	time.Sleep(time.Duration(seconds) * time.Second)
}
````

the step

```go
	s.Step(`^user (?:waits|waited) "(\d+)" seconds?$`,
		util.WaitForNSeconds)
```
throws an error when used in a feature:

```
panic: expected handler to return only one value, but it has: 0

goroutine 1 [running]:
github.com/minishift/minishift/vendor/github.com/DATA-DOG/godog.(*Suite).Step(0xc4203219e0, 0x9acfc0, 0xc4204476a0, 0x9ab5c0, 0xaa2078)
        /home/agajdosi/golang/src/github.com/minishift/minishift/vendor/github.com/DATA-DOG/godog/suite.go:103 +0x842
github.com/minishift/minishift/test/integration.FeatureContext(0xc4203219e0)
        /home/agajdosi/golang/src/github.com/minishift/minishift/test/integration/integration_test.go:301 +0x1da7
github.com/minishift/minishift/test/integration.TestMain.func1(0xc4203219e0)
        /home/agajdosi/golang/src/github.com/minishift/minishift/test/integration/integration_test.go:105 +0x2b
github.com/minishift/minishift/vendor/github.com/DATA-DOG/godog.(*runner).run(0xc42034a980, 0x14)
        /home/agajdosi/golang/src/github.com/minishift/minishift/vendor/github.com/DATA-DOG/godog/run.go:75 +0xaa
github.com/minishift/minishift/vendor/github.com/DATA-DOG/godog.RunWithOptions(0xa7c7e9, 0x9, 0xaa1ee8, 0x0, 0x0, 0x0, 0xc420300210, 0x10, 0xa79027, 0x6, ...)
        /home/agajdosi/golang/src/github.com/minishift/minishift/vendor/github.com/DATA-DOG/godog/run.go:167 +0x83b
github.com/minishift/minishift/test/integration.TestMain(0xc420487f20)
        /home/agajdosi/golang/src/github.com/minishift/minishift/test/integration/integration_test.go:113 +0x30f
main.main()
        github.com/minishift/minishift/test/integration/_test/_testmain.go:38 +0xf7
exit status 2
FAIL    github.com/minishift/minishift/test/integration 0.058s
make: *** [Makefile:206: integration] Error 1
```

Still I can remove the function as it is not needed, just wanted to add a step which could be practical for debugging purposes. Let me know please @LalatenduMohanty @gbraad

gbraad: can be removed, as it does not provide any additional debugging option

agajdosi:
@gbraad, @LalatenduMohanty Sleep wrapping function transformed into anonymous function directly in step definition. However the need for the return value of type error remains.

LalatenduMohanty:
Raised #1895

Issue #1837, opened 2017-12-12
Getting import logs error on first time start when caching enabled


budhrg:
### General information

  * Minishift version: 1.10.0
  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce

  1. Enable image caching with `minishift config set image-caching true`
  2. Do `minishift start`

### Expected
Should not get any import logs error on first time `minishift start`

### Actual
Getting weird import logs error on first time `minishift start`

### Logs

```
-- Checking available disk space ... 0% used OK
   Importing openshift/origin:v3.6.0  FAIL
   Importing openshift/origin-docker-registry:v3.6.0  FAIL
   Importing openshift/origin-haproxy-router:v3.6.0  FAIL
  WARN: Import of cached images failed. Continuing without importing images. Error: Unable to import image 'openshift/origin:v3.6.0' because it is not available in the local cache.
Unable to import image 'openshift/origin-docker-registry:v3.6.0' because it is not available in the local cache.
Unable to import image 'openshift/origin-haproxy-router:v3.6.0' because it is not available in the local cache. 
```

LalatenduMohanty:
Resolved via #1839

Issue #1803, opened 2017-12-05
Add ability to force a pull of images during image export


hferentschik:
ATM, during an `image export` the only criteria of whether an image needs to be (re-)exported is whether there is an entry in the OCI index. Images which change (_latest_ tag) will not get updated. An option should be provided to force a re-export, eg

    $ minishift image export --validate

budhrg:
Not sure `--validate` will convey the right message to force re-export the image.

gbraad:
same for me; `validate` does not convey the meaning to force a re-export and actually pulling `:latest` again.
Validate to me means to 'check the validity or accuracy of the requested image`, with accuracy meaning more along the lines of the hash (to ensure the image is not corrupted).

hferentschik:
Note the "eg". Could be "forced" (I was leaning to this initially, but it seems a bit too strong), or "re-pull" or ....

gbraad:
`re-pull` comes closer... but likely without the dash...

hferentschik:
Well, not quite. Since "re-pull" implies that the image gets pulled again from the Docker daemon side, but that is not necessarily the case.

gbraad:
in that case, since it will not 're-pull' from the registry, it might be more appropriate to call it as the alternatives: `--refresh`... :-/ `--freshen`... :-s `--update` :-|

praveenkumar:
If we check from docker side `docker pull` always checks the layers and if any layer updated it re-pull. If we want to implement the same behavior with `export` then it going to depend on the internet. So I think `--refresh` or `freshen` looks like a good option.

jorgemoralespou:
--overwrite sounds better?

```
minishift image export --overwrite
```

budhrg:
> minishift image export --overwrite

:+1: Sounds better

LalatenduMohanty:
Resolved via #2436

Issue #1802, opened 2017-12-05
Add ability to clear a given image from the cache


hferentschik:
At the moment there is no way to selectively remove an image from the image cache. It's all or nothing (in fact you need to clear the whole cache). We need a way to selective delete an image, eg

    $ minishift image prune <image names>

This most likely will need changes to underlying container/images.

praveenkumar:
This is something I would like to work on once we have a proper milestone to it.

anjannath:
@praveenkumar since you are pre-occupied with the 3.10 work, would you like me to work on it?

praveenkumar:
@anjannath https://github.com/minishift/minishift/pull/2336 3.10 work is mostly done and waiting for feedback mostly. I want to work on this because this is the codebase which is dark for me, implementing this feature might give some insight for this codebase.

LalatenduMohanty:
Resolved via #2381

PR #1839, opened 2017-12-12
Issue #1837 Making sure start command does not print misleading messages

hferentschik:
Addresses #1837

- Making sure that on empty cache the start command is not printing misleading error messages
- Changing ImageHandler.ExportImages and ImageHandlerImportImages to also returns the list of successfully imported/exported images

### DIFF
--- a/cmd/minishift/cmd/image/import.go
+++ b/cmd/minishift/cmd/image/import.go
@@ -100,9 +100,13 @@ func importImage(cmd *cobra.Command, args []string) {
 		ImageMissStrategy: image.Skip,
 	}
 
-	err = handler.ImportImages(imageCacheConfig)
+	importedImages, err := handler.ImportImages(imageCacheConfig)
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Container image import failed:\n%v", err))
+	}
+
+	if len(importedImages) < len(normalizedImageNames) {
+		atexit.ExitWithMessage(1, "At least one image could not be imported.")
 	}
 }
 
--- a/pkg/minishift/docker/image/image_handler.go
+++ b/pkg/minishift/docker/image/image_handler.go
@@ -16,13 +16,36 @@ limitations under the License.
 
 package image
 
+// ProgressStatus defines a status code for import/export operations.
+type ProgressStatus int
+
+const (
+	OK ProgressStatus = iota
+	FAIL
+	CACHE_MISS
+)
+
+func (ps ProgressStatus) String() string {
+	switch ps {
+	case OK:
+		return "OK"
+	case FAIL:
+		return "FAIL"
+	case CACHE_MISS:
+		return "CACHE MISS"
+	}
+	return ""
+}
+
 // ImageHandler is responsible for the import and export of images into the Docker daemon of the VM
 type ImageHandler interface {
 	// ImportImages imports cached images from the host into the Docker daemon of the VM.
-	ImportImages(config *ImageCacheConfig) error
+	// The method returns the list of successfully imported images and an error if one occurred.
+	ImportImages(config *ImageCacheConfig) ([]string, error)
 
 	// ExportImages exports the images specified as part of the ImageCacheConfig from the VM to the host.
-	ExportImages(config *ImageCacheConfig) error
+	// The method returns the list of successfully exported images and an error if one occurred.
+	ExportImages(config *ImageCacheConfig) ([]string, error)
 
 	// IsImageCached returns true if the specified image is cached, false otherwise.
 	IsImageCached(config *ImageCacheConfig, image string) bool
--- a/cmd/minishift/cmd/image/export.go
+++ b/cmd/minishift/cmd/image/export.go
@@ -97,7 +97,7 @@ func exportImage(cmd *cobra.Command, args []string) {
 		ImageMissStrategy: image.Pull,
 	}
 
-	err = handler.ExportImages(imageCacheConfig)
+	_, err = handler.ExportImages(imageCacheConfig)
 	if err != nil {
 		msg := fmt.Sprintf("Container image export failed:\n%v", err)
 		if logToFile {
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -418,9 +418,9 @@ func importContainerImages(driver drivers.Driver, api libmachine.API, openShiftV
 		CachedImages: images,
 		Out:          os.Stdout,
 	}
-	err = handler.ImportImages(config)
+	_, err = handler.ImportImages(config)
 	if err != nil {
-		fmt.Println(fmt.Sprintf("  WARN: Import of cached images failed. Continuing without importing images. Error: %s ", err.Error()))
+		fmt.Println(fmt.Sprintf("  WARN: At least one image could not be imported. Error: %s ", err.Error()))
 	}
 }
 
--- a/pkg/minishift/docker/image/oci_image_handler.go
+++ b/pkg/minishift/docker/image/oci_image_handler.go
@@ -17,7 +17,6 @@ limitations under the License.
 package image
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -25,6 +24,7 @@ import (
 
 	"bytes"
 	"encoding/json"
+	"errors"
 	"github.com/containers/image/copy"
 	"github.com/containers/image/oci/layout"
 	"github.com/containers/image/signature"
@@ -84,17 +84,18 @@ func NewLocalOnlyOciImageHandler() (*OciImageHandler, error) {
 }
 
 // ImportImages imports cached images from the host into the Docker daemon of the VM.
-func (handler *OciImageHandler) ImportImages(config *ImageCacheConfig) error {
+func (handler *OciImageHandler) ImportImages(config *ImageCacheConfig) ([]string, error) {
 	out := handler.getOutputWriter(config)
+	importedImages := []string{}
 
 	policyContext, err := handler.getPolicyContext()
 	if err != nil {
-		return fmt.Errorf("Error creating security context: %s", err.Error())
+		return importedImages, fmt.Errorf("Error creating security context: %s", err.Error())
 	}
 
 	availableImages, err := handler.GetDockerImages()
 	if err != nil {
-		return err
+		return importedImages, err
 	}
 
 	multiError := util.MultiError{}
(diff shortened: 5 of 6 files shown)


cmd/minishift/cmd/image/import.go
```diff
@@ -100,10 +100,14 @@ func importImage(cmd *cobra.Command, args []string) {
...
 	}
 
-	err = handler.ImportImages(imageCacheConfig)
+	importedImages, err := handler.ImportImages(imageCacheConfig)
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Container image import failed:\n%v", err))
 	}
+
+	if len(importedImages) < len(normalizedImageNames) {
+		atexit.ExitWithMessage(1, "At least one image could not be imported.")
```
gbraad: The problem is; which image failed? Why did it fail?

The last question, of why, will be hard as the error is not available. But the naming should be able to be determined as " A ∩ B ". Because I might have to re-export this one?

pkg/minishift/docker/image/image_handler.go
```diff
@@ -16,13 +16,36 @@ limitations under the License.
 
 package image
 
+// ProgressStatus defines a status code for import/export operations.
+type ProgressStatus int
+
+const (
+	Ok ProgressStatus = iota
```
gbraad: A little bit of nitpicking perhaps; but Ok is a term that is about 99% of the cases written as `OK`. Since this is also a constant, it suits better. This makes it more obvious that is relates to

  * https://github.com/minishift/minishift/pull/1839/files#diff-0d2b6075538057707499cc3316570998R108
  * https://github.com/minishift/minishift/pull/1839/files#diff-0d2b6075538057707499cc3316570998R402

and the other situations as:
  * https://github.com/minishift/minishift/pull/1839/files#diff-0d2b6075538057707499cc3316570998R115
  * etc ...

cmd/minishift/cmd/image/import.go
hferentschik: > The problem is; which image failed? Why did it fail?

That's already displayed. I explicitly choose this more generic format for now.  I don't think any information is lost here. However, as you said, the information is there if one wants to extend it at some stage.

> Progress is shown, right? and the indication shows fail when an import fails... right?

Right

> If so, maybe it might be better to show: "Unable to continue since at least one image failed to import".

To continue with what. The commands end here.

pkg/minishift/docker/image/image_handler.go
hferentschik: > but Ok is a term that is about 99% of the cases written as OK. Since this is also a constant,

So what's about the other constants? FAIL and CACHE_MISS? Or you want to just rename Ok? I thought we at some stage said we don't use the all capital spelling for constants. Either way, one would have an inconsistency. 

If you feel strongly about it, I rename it. I don't understand what you mean with the references code diffs though. I am not quite sure how Ok vs OK matters there.

cmd/minishift/cmd/image/import.go
gbraad: > Unable to continue

If the command exits, you have to express this to the user. Using "At least one image could not be imported." to me is not a clear exit condition by itself if expressed this way.

pkg/minishift/docker/image/image_handler.go
gbraad: Rename ALL. In the rest of the code, these variables do not stand out or are clear in their origin.

LalatenduMohanty:
While testing the PR saw this:
```
   Importing openshift/origin:v3.6.0 . FAIL
   Importing openshift/origin-docker-registry:v3.6.0  FAIL
   Importing openshift/origin-haproxy-router:v3.6.0 . FAIL
  WARN: Import of cached images failed. Continuing without importing images. Error: Unable to import image 'openshift/origin:v3.6.0' because it is not available in the local cache.
Unable to import image 'openshift/origin-docker-registry:v3.6.0' because it is not available in the local cache.
Unable to import image 'openshift/origin-haproxy-router:v3.6.0' because it is not available in the local cache. 
```

I can still see the same issue as mentioned in #1837

LalatenduMohanty:
Removed my last comment as it was a false alarm. I see below
```
   Importing openshift/origin:v3.6.0  CACHE MISS
   Importing openshift/origin-docker-registry:v3.6.0  CACHE MISS
   Importing openshift/origin-haproxy-router:v3.6.0  CACHE MISS
```
But I do not think the output, specially `CACHE MISS` word, I think just `Importing openshift/origin:v3.6.0 ... Missing` would be fine.

hferentschik:
> But I do not like the output, specially CACHE MISS word, I think just Importing openshift/origin:v3.6.0 ... Missing would be fine.

Does "missing" potentially not hint towards a more serious issue? Cache miss is imo more concise.

cmd/minishift/cmd/image/import.go
hferentschik: > If the command exits, you have to express this to the user. Using "At least one image could not be imported." to me is not a clear exit condition by itself if expressed this way.

You have to look at all the messages which appear as part of running the command.

pkg/minishift/docker/image/image_handler.go
hferentschik: I renamed the constants. Not sure I follow your line of thoughts though ;-)

LalatenduMohanty:
@hferentschik From a pure aesthetic point of view the output need to be tweaked. See below:
```
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ..................... OK
   Pinging 8.8.8.8 ... OK
   Importing openshift/origin:v3.6.1  CACHE MISS
   Importing openshift/origin-docker-registry:v3.6.1  CACHE MISS
   Importing openshift/origin-haproxy-router:v3.6.1  CACHE MISS
```
should be 

```
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ..................... OK
   Pinging 8.8.8.8 ... OK
   Importing openshift/origin:v3.6.1  CACHE MISS
   Importing openshift/origin-docker-registry:v3.6.1 ... CACHE MISS
   Importing openshift/origin-haproxy-router:v3.6.1 ... CACHE MISS
```
or
```
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ..................... OK
   Pinging 8.8.8.8 ... OK
   Importing openshift/origin:v3.6.1  CACHE MISS
   Importing openshift/origin-docker-registry:v3.6.1 : CACHE MISS
   Importing openshift/origin-haproxy-router:v3.6.1 : CACHE MISS
```

LalatenduMohanty:
Looking at the code , required code is there to print the progress dots but because it is failing so fast before printing anything from progress dots. Not sure what we can do here. But it somehow looks odd.

LalatenduMohanty:
I manually deleted one blob and here is the error 
```
   Importing openshift/origin:v3.6.1 ............... OK
   Importing openshift/origin-docker-registry:v3.6.1 . FAIL
   Importing openshift/origin-haproxy-router:v3.6.1 ..... OK
  WARN: At least one image could not be imported. Error: Error determining manifest MIME type for oci:/home/lmohanty/.minishift/cache/images:openshift/origin-docker-registry:v3.6.1: open /home/lmohanty/.minishift/cache/images/blobs/sha256/617a9102d02b4888e3ef55fccfa7cce5d480afbbf69db9ca9d08786e5a5896a4: no such file or directory 
```
But I can see the docker-registry running the VM. Did it download the image again? Should it cache again?

```
$ minishift ssh "docker ps"
CONTAINER ID        IMAGE                                                                                                      COMMAND                  CREATED              STATUS              PORTS               NAMES
5e88794f72c2        openshift/origin-docker-registry@sha256:462acb1b568584fa6d84e4dd2b1ccea637f73b474ab6d0f3c87faa62aea322ab   "/bin/sh -c '/usr/bin"   46 seconds ago       Up 45 seconds                           k8s_registry_docker-registry-1-g65ds_default_882a6fbc-e3ef-11e7-8353-1ae7cb81792d_0
134225d4746e        openshift/origin-pod:v3.6.1                                                                                "/usr/bin/pod"           About a minute ago   Up About a minute                       k8s_POD_docker-registry-1-g65ds_default_882a6fbc-e3ef-11e7-8353-1ae7cb81792d_0
84041cfd4508        80baf7d123bd                                                                                               "/usr/bin/openshift-r"   About a minute ago   Up About a minute                       k8s_router_router-1-nqpvd_default_84776d18-e3ef-11e7-8353-1ae7cb81792d_0
da30d4c3e595        openshift/origin-pod:v3.6.1                                                                                "/usr/bin/pod"           About a minute ago   Up About a minute                       k8s_POD_router-1-nqpvd_default_84776d18-e3ef-11e7-8353-1ae7cb81792d_0
2110af36b25b        openshift/origin:v3.6.1                                                                                    "/usr/bin/openshift s"   3 minutes ago        Up 3 minutes                            origin
```

cmd/minishift/cmd/image/import.go
LalatenduMohanty: @gbraad  I think @hferentschik is right here, check the output https://github.com/minishift/minishift/pull/1839#issuecomment-352414018 , though it is coming from start.go. But as per code it actually prints, ok, fail or cache miss against each image. So we are fine here. 

```
$ minishift images import --all
   Importing alpine:latest . FAIL
Container image import failed:
Error determining manifest MIME type for oci:/home/lmohanty/.minishift/cache/images:alpine:latest: open /home/lmohanty/.minishift/cache/images/blobs/sha256/c577c840a769fe3ecc2f7617640c03eca9ddad9b762208d3c9453fee8aa3d327: no such file or directory
```

gbraad:
```diff
-- Checking available disk space ... 0% used OK
-   Importing openshift/origin:v3.6.1 ............... OK
+-- Importing container images from cache ...
+   Importing openshift/origin:v3.6.1 ............... OK
   Importing openshift/origin-docker-registry:v3.6.1 . FAIL
   Importing openshift/origin-haproxy-router:v3.6.1 ..... OK
  WARN: At least one image could not be imported. Error: Error determining manifest MIME type for oci:/home/lmohanty/.minishift/cache/images:openshift/origin-docker-registry:v3.6.1: open /home/lmohanty/.minishift/cache/images/blobs/sha256/617a9102d02b4888e3ef55fccfa7cce5d480afbbf69db9ca9d08786e5a5896a4: no such file or directory 
```

cmd/minishift/cmd/image/import.go
gbraad: It was the message itself, but since the command actually 'continues' it can be ignored. But somehow this causes the instance to start with an incomplete set of images.

I do notice the alignments jump around here...
```
$ minishift images import --all
   Importing alpine:latest . FAIL
```
this is due to re-using the function for both the start.go logic and the cmd.

LalatenduMohanty:
@hferentschik I am fine with the PR but I need a explanation for https://github.com/minishift/minishift/pull/1839#issuecomment-352414018. After that I will approve the PR.

gbraad:
@LalatenduMohanty what was the confusion related to 'continue' or not: https://github.com/minishift/minishift/pull/1839#discussion_r157506540

gbraad:
`CACHE MISS` in the startup message is maybe a little too technical, but covers for the moment what happens. I am OK with the PR, ...

hferentschik:
@LalatenduMohanty 

> I manually deleted one blob and here is the error

You cannot do that atm. You effectively invalidated the cache. This relates to issues #1802   and #1803. We need a way to selectively delete images as well as forcing a re-export. The latter would solve the issues. ATM, the index file is considered the truth, meaning if the index states the image is there, but you delete a blob out of the underlying Merkle tree, you won't be able to import this image. #1803 is pretty straightforward and should be tackled asap.

> But I can see the docker-registry running the VM. Did it download the image again

It needs to, right? It is one of the core images. Different if it would be a custom image.

> Should it cache again? 

See #1803. Atm, there is no verification of whether the image data is correct. Would be nice to add, not sure what containers/image offers in this regards though.

hferentschik:
> After that I will approve the PR.

Thanks. Please merge.

LalatenduMohanty:
> TM, the index file is considered the truth, meaning if the index states the image is there, but you delete a blob out of the underlying Merkle tree, you won't be able to import this image. #1803 is pretty straightforward and should be tackled asap.

Yes, I observed the same when I was doing further testing.

Issue #1841, opened 2017-12-13
Minishift 1.10.0 can not provision OpenShift v3.6.1


paoloantinori:
### General information

  * Minishift version: `minishift v1.10.0+10461c6`
  * OS: `Linux `
  * Hypervisor: `KVM `

### Steps to reproduce

  1. `minishift start --memory 2048 --cpus 2 --show-libmachine-logs  --openshift-version v3.6.1  --alsologtostderr`
  2. 
  3. 
  4. 

### Expected
Minishift being able to handle `v3.6.1` correctly

### Actual
```
FAIL
   Error: could not start OpenShift container "origin"
   Details:
     No log available from "origin" container

Error during 'cluster up' execution: Error starting the cluster.
```

### Logs

The issue reported in logs is very specific and it's related to the lack of an attribute:

```
F1213 11:39:01.486397    2973 start_allinone.go:99] Server could not start: Couldn't init admission plugin "openshift.io/ImagePolicy": [openshift.io/ImagePolicy.resolutionRules[0].policy: Required value: a policy must be specified for this resource, openshift.io/ImagePolicy.resolutionRules[1].policy: Required value: a policy must be specified for this resource, openshift.io/ImagePolicy.resolutionRules[2].policy: Required value: a policy must be specified for this resource, openshift.io/ImagePolicy.resolutionRules[3].policy: Required value: a policy must be specified for this resource, openshift.io/ImagePolicy.resolutionRules[4].policy: Required value: a policy must be specified for this resource]
```

I'm trying to patch config to put those required values there with this command:

```
JSON='{"admissionConfig":{"pluginConfig":{"openshift.io/ImagePolicy":{"configuration":{"resolutionRules":[{"localNames":true,"policy":"AttemptRewrite"}]}}}}}'
minishift openshift config set --patch $JSON --target master
```

But that fails since `origin` container is stopped.

Now, I'm not sure that my patch is all I need to make 3.6.1 to work, anyhow I'm looking for a workflow that can allow me to patch master configuration before starting it, so that I can accomplish something lik what's defined in this scenario.

LalatenduMohanty:
@paoloantinori It looks like a bug in Minishift we introduced recently. We will fix it asap. Till then 
you can do following manual steps to get it working.

* Download the oc client for Linux from https://github.com/openshift/origin/releases/tag/v3.6.1
 * Extract the oc binary and replace the existing oc binary with the new one ~/.minishift/cache/oc/v3.6.0/linux/
[...]

PR #1845, opened 2017-12-14
Issue #1841 Fix to provision OpenShift 3.6.1

praveenkumar:


LalatenduMohanty:
@praveenkumar you need to fix `func Test_DetermineOcVersion`

gbraad:
what if `3.6.2` comes along? Does would mean we have to force a new point release to make it work... right?

praveenkumar:
> what if 3.6.2 comes along? Does would mean we have to force a new point release to make it work... right?

I think minor version can be provisioned without any problem. This problem occurred there was some bug on `oc cluster up` but I am hoping it not going to happen so frequently. If again occurred then we might force new point.

gbraad:
Do we have a related BZ issue for this? Might be important to have this as part of the issue description or even commit description.

I see no further issue here... however, this will result in a point release of Minishift, and if so, what else would have to go in? I guess the start error message for caching?

LalatenduMohanty:
> what if 3.6.2 comes along? Does would mean we have to force a new point release to make it work... right?

I was also thinking the same. Ideally the version should be the last in the 3.6.x series. 

> Do we have a related BZ issue for this? Might be important to have this as part of the issue description or even commit description.

I think we should file an issue on this i.e. using 3.6.0 binary we can not provision 3.6.1  as this is an issue for sure.

LalatenduMohanty:
Raised https://github.com/openshift/origin/issues/17821

LalatenduMohanty:
@praveenkumar @gbraad IMO we should change `OPENSHIFT_VERSION = v3.6.0` to v3.6.1 in the Makefile. I understand that we would move to 3.7.0 in #1776  but with this PR v3.6.0 in Makefile does not look consistent in a way.

gbraad:
I agree

Issue #1805, opened 2017-12-06
Make dep installation as part of Makefile same as go-bindata


praveenkumar:
We should add dep installation as part of Makefile, same way we are doing for `go-bindata`.

LalatenduMohanty:
We should not add the installtion as part of Makefile as it is not good for our CI runs. Refer https://github.com/minishift/minishift/pull/1846#issuecomment-352390953

PR #1846, opened 2017-12-15
[WIP] Issue #1805 Install dep in Makefile

LalatenduMohanty:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -84,7 +84,8 @@ __check_defined = \
 .PHONY: $(GOPATH)/bin/minishift$(IS_EXE)
 $(GOPATH)/bin/minishift$(IS_EXE): $(ADDON_ASSET_FILE) vendor
 	go install -tags "$(BUILD_TAGS)" -pkgdir=$(ADDON_BINDATA_DIR) -ldflags="$(VERSION_VARIABLES)" ./cmd/minishift
-vendor:
+vendor: $(GOPATH)/bin/dep
+	echo $(shell dep version)
 	dep ensure -v
 
 $(ADDON_ASSET_FILE): $(GOPATH)/bin/go-bindata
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
Tested in Fedora 27 and it works fine there.

LalatenduMohanty:
Tested on macOS and it working fine.

LalatenduMohanty:
> I am able to build from master it when manually downloading the dep version 0.3.2 in GOPATH/bin.

@budhrg Where did the dep got installed in Windows bydefault?

LalatenduMohanty:
@budhrg I think the last three dots `...` might be causing issue in Windows along with need for `$(GOPATH)/bin/dep$(IS_EXE)` . Will check in Windows.

LalatenduMohanty:
Also we need see if auto installation of this is causing any issue in CI runs as we have specific version pinned in CI runs for dep.

Makefile
```diff
@@ -84,7 +84,7 @@ __check_defined = \
 .PHONY: $(GOPATH)/bin/minishift$(IS_EXE)
 $(GOPATH)/bin/minishift$(IS_EXE): $(ADDON_ASSET_FILE) vendor
 	go install -tags "$(BUILD_TAGS)" -pkgdir=$(ADDON_BINDATA_DIR) -ldflags="$(VERSION_VARIABLES)" ./cmd/minishift
-vendor:
+vendor: $(GOPATH)/bin/dep
 	dep ensure -v
```
budhrg: @LalatenduMohanty Could you add Add `echo $(shell dep version)` here to see which version of dep is installed?

LalatenduMohanty:
Looks like CI (e.g. appveyor) will start using the dep version from master as `go get` always installs from master.  We do not want to use  `dep` from master as it will break our CI when the master gets broken. For the same reason we have pinned version of `dep` in CI. Hence closing the PR.
```
echo dep:  version     : devel  build date  :   git hash    :   go version  : go1.8.3  go compiler : gc  platform    : windows/amd64
dep: version : devel build date : git hash : go version : go1.8.3 go compiler : gc platform : windows/amd64
```

gbraad:
We do the same for go-bindata?!?

PR #1852, opened 2017-12-18
Issue #1568 - Add nightly tests with for Minikube and CentOS images

### LINKED ISSUE #1568: Start testing Minishift with Minikube/CentOS ISO as part nightly job

agajdosi:
Fix #1568

### DIFF
--- a/centos_ci.sh
+++ b/centos_ci.sh
@@ -342,7 +342,7 @@ function perform_master() {
 
 function perform_nightly() {
   make prerelease synopsis_docs link_check_docs
-  MINISHIFT_VM_DRIVER=kvm make integration_all
+  MINISHIFT_ISO_URL=$1 MINISHIFT_VM_DRIVER=kvm make integration_all
 }
 
 if [[ "$UID" = 0 ]]; then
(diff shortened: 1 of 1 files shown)


centos_ci.sh
```diff
@@ -371,7 +371,11 @@ else
   elif [[ "$JOB_NAME" = "minishift-release" ]]; then
     perform_release $RSYNC_PASSWORD;
   elif [[ "$JOB_NAME" = "minishift-nightly" ]]; then
-    perform_nightly;
+    perform_nightly "b2d";
+  elif [[ "$JOB_NAME" = "minishift-nightly-minikube" ]]; then
+    perform_nightly "minikube";
```
praveenkumar: I didn't see any job exist https://github.com/minishift/minishift-ci-jobs/blob/master/minishift-ci-index.yaml so first we should update our JJB file and then move forward. cc @budhrg

agajdosi:
Rebased, needed changes in minishift/minishift-ci-jobs merged.

Issue #1850, opened 2017-12-18
minikube iso not working anymore because of docker version update.


praveenkumar:
### Steps to reproduce

  1. Clean iso cache (remove if you already have an old minikube iso)
  2.  minishift start --iso-url minikube

### Expected

It should start as expected.

### Actual

It is failing to start docker daemon because systemd unit file changed for a new version of docker.

### Logs

```
About to run SSH command:
sudo systemctl -f restart docker
SSH cmd err, output: exit status 1: Job for docker.service failed because the control process exited with error code.
See "systemctl  status docker.service" and "journalctl  -xe" for details.
```

agajdosi:
Same behaviour here with latest master.

LalatenduMohanty:
@praveenkumar I thought we have tests in Minikube ISO where it checks for Minishift compatibility. Why did we miss the issue.

praveenkumar:
@LalatenduMohanty Yesterday we had meeting with minikube folks and they admitted it happens from their end and will not happen in future. According to them they never knew 0.24 already released one and they updated without bump version. In future, if there is any iso change and minishift test failing then they will always be going to add us to review something they promised yesterday.

LalatenduMohanty:
Resolved via #1853

PR #1853, opened 2017-12-19
Issue #1850 Update minishift provisioner to work with latest minikube-iso

praveenkumar:


LalatenduMohanty:
@budhrg @agajdosi There should be some way we should be able to trigger the CI tests with Minikube ISO or atleast kickstart the nightly manually for this PR to test.

LalatenduMohanty:
@praveenkumar The commit message says its `minikube provisioner` but it is `minishift` provisioner right?

Issue #1861, opened 2017-12-27
minishift version should ignore profile details


praveenkumar:
* Minishift version: current master
  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce

  1.  Delete any running VM
  2.  rm -fr ~/.minishift
  3.  minishift version

### Expected
minishift v1.10.0+77422c5c

### Actual
Profile: minishift doesn't exist, Use `minishift profile set minishift` or `minishift start --profile minishift` to create

LalatenduMohanty:
Resolved via #1862

PR #1862, opened 2017-12-27
Issue #1861 Ignore checkForValidProfileOrExit function for default profile

praveenkumar:
Default profile name 'minishift' shouldn't be part of checkForValidProfileOrExit function because this is already a valid profile name and globally true.


gbraad:
we should pay more attention to testing... integration did not catch this?

PR #1871, opened 2017-12-29
Issue #1860 Add poll to verify persistent-volume-setup job to successful

praveenkumar:


pkg/minishift/clusterup/clusterup.go
```diff
@@ -243,13 +243,21 @@ func configurePersistentVolumes(hostPvDir string, addOnManager *manager.AddOnMan
 		time.Sleep(1 * time.Second)
 	}
 
-	// verify the job succeeded
-	out = new(bytes.Buffer)
-	err = new(bytes.Buffer)
-	exitStatus := ocRunner.Run("get job persistent-volume-setup -n default -o 'jsonpath={ .status.succeeded }'", out, err)
+	// poll the success status of persistent-volume-setup job verify.
+	for {
```
hferentschik: I think we should get away from the endless loop here as well as above and also introduce another exit condition. There is a chance this will hang forever. Better safe than sorry.

Why is this condition needed? Does it work? I have a vague memory that there was an issue with it?

praveenkumar: > I think we should get away from the endless loop here as well as above and also introduce another exit condition

@hferentschik Another exit condition you mean some kind of time limit?

> Why is this condition needed? Does it work? I have a vague memory that there was an issue with it?

Looks like when a pod goes from active => success state there is a fraction of time when it not updated the state correctly and this was was causing issue for our CI for a long time :(

budhrg: @hferentschik Verified that it is working locally in my CentOS box(where the issue #1860 is reproducible) and in CI job https://ci.centos.org/job/minishift-pr/3144/consoleFull 
```
00:04:59.913 
00:04:59.913   Scenario: Starting Minishift                 # features/basic.feature:28
00:04:59.914     Given Minishift has state "Does Not Exist" # integration_test.go:165 -> *Minishift.shouldHaveState
00:07:47.661     When executing "minishift start" succeeds  # integration_test.go:710 -> executingMinishiftCommandSucceedsOrFails
00:07:48.206     Then Minishift should have state "Running" # integration_test.go:165 -> *Minishift.shouldHaveState
00:07:48.829 
```

> There is a chance this will hang forever. Better safe than sorry

+1 here to get rid of for loop. I am also feeling this infinite for loop might have later bite us.

budhrg: In my CentOS CI box, this is the behavior
```
[...]
-- Waiting for persistent volumes to be created ... [ Waited some time like 35-50 seconds] OK 
```
and worked.

I am just wondering if by any possible way we can be stuck here for infinite wait period. 
I was thinking of some kind of timeout here.

hferentschik: And why should there be a dealy between the job executing and it reported success? Something is not adding up. Have you talked with @csrwng to understand what's going on here?

hferentschik: @praveenkumar, right I am talking about a timeout. Without there is a risk for an endless loop.

hferentschik: > Verified that it is working locally in my CentOS box

"It works" is one thing. I am trying to understand why it would be required and why it would be required now. It did work so far, right? 

> Looks like when a pod goes from active => success state there is a fraction of time when it not updated the state correctly and this was causing issue for our CI for a long time :(

For a long time? @csrwng, is this a expected behaviour that there is a delay between job going from "active" to "success".

hferentschik: >  here to get rid of for loop. I am also feeling this infinite for loop might have later bite us.

Not necessarily getting rid of the loop, but at least ensuring that it eventually will exit one way or another.

hferentschik: > Looks like when a pod goes from active => success state there is a fraction of time when it not updated the state correctly and this was causing issue for our CI for a long time :(

Which test exactly? Does it make sense to change our code for that? For our purposes, we don't have to wait. The PVs are created. Maybe we should change the test instead?

budhrg: @hferentschik So we found that in 3.6.*, the first for loop is taking time, now with 3.7.*, the first for loop is taking no time and 2nd for loop is taking time. 
Hence, there is definitely a change of behavior. 
/cc @csrwng

hferentschik: > So we found that in 3.6.*, the first for loop is taking time, now with 3.7.*, the first for loop is taking no time and 2nd for loop is taking time.

That sounds like a bug/regression. Let's find out what's going on before changing code because "it works".

LalatenduMohanty: @hferentschik We have looked in to it and the behavior is not wrong. The difference if behavior is only seen on a buffy machine because the creation of directories happens faster (I think) hence it quickly moves to the next check to see if it is successful. It is really hard to reproduce it locally i.e. I could reproduce 1 out of 20 times. Also this behavior is very specific to "oc cluster up". The Production deployment of OpenShift is not effected. So I feel we should go ahead with this PR. 

The only difference is it is taking time to give if the operation is successful

Issue #1891, opened 2018-01-08
Error message to export an image is different for minikube and b2d ISO


praveenkumar:
### Steps to reproduce

```
$ ./minishift start --iso-url b2d
[..]
You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin

$  ./minishift image export foo:latest
Exporting foo:latest... FAIL
Container image export failed:
Error running command 'docker pull foo:latest': Process exited with status 1 
Pulling repository docker.io/library/foo
Error: image library/foo:latest not found


$ ./minishift start --iso-url minikube
[...]
You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin

$ ./minishift image export foo:latest
Exporting foo:latest.. FAIL
Container image export failed:
Error running command 'docker pull foo:latest': Process exited with status 1 
Error response from daemon: repository foo not found: does not exist or no pull access
```

This is happen because the docker version which is part of b2d and minikube is different. This also the one of reason where minikube nightly builds are failing in integration test. https://github.com/minishift/minishift/blob/master/test/integration/features/cmd-image.feature#L71-L73 senario is depend on docker daemon output instead our program error message.

LalatenduMohanty:
Resolved via #1892

PR #1892, opened 2018-01-08
Issue #1891 Fix error message to export an image for image feature

praveenkumar:


LalatenduMohanty:
@minishift/minishift-dev @agajdosi @amitkrout Here is another example where I wanted to run the integration runs on top of Minikube and CentOS ISO before merging.

LalatenduMohanty:
Raised #1893

Issue #1858, opened 2017-12-22
`minishift config unset` does not return error message if an invalid property name is passed


anjannath:
### General information

  * Minishift version: master
  * OS: All (Linux / macOS / Windows) 
  * Hypervisor: All (KVM / xhyve / Hyper-V / VirtualBox)

### Steps to reproduce

  1. `minishift config unset foo`

### Expected
An error message saying foo is not a valid property or the syntax for `minishift config unset` command

### Actual
nothing happens, and the user has no idea if the value was unset

LalatenduMohanty:
Does this look ok
```
$ minishift config unset dummyconfig
Cannot find property name dummyconfig
```

anjannath:
Yes, this looks fine, also the `minidhift config set` command doesn't show a message on succes, but i am a bit unsure if we want to show a message saying something like `iso url property successfuly set`. I feel it might be an overkill. @LalatenduMohanty WDYT?

LalatenduMohanty:
@anjannath yeah we can print success message.

LalatenduMohanty:
Resolved by https://github.com/minishift/minishift/commit/fe75c68dd7e547c8739997f762b3963caef9cb1c #1903

PR #1903, opened 2018-01-10
Issue #1858 Print success and failure message for `minishift config unset`

anjannath:


cmd/minishift/cmd/config/unset.go
```diff
@@ -44,10 +44,18 @@ func init() {
...
+	}
 	m, err := ReadConfig()
 	if err != nil {
 		return err
 	}
+	if m[name] == nil {
+		return fmt.Errorf("Property name '%s' is not set", name)
+	}
 	delete(m, name)
+	fmt.Printf("Property name '%s' successfuly unset\n", name)
```
LalatenduMohanty: @hferentschik I remember you are saying providing output to all commands related for saying success was counter productive to you. I am not sure why we came to that conclusion. Can you take a look here?

praveenkumar: @LalatenduMohanty my understanding is we should only provide direct feedback related to unset command when a user try to unset a config which either not set or exist. For successfully unset we don't need to include this message.

LalatenduMohanty:
Merged https://github.com/minishift/minishift/commit/fe75c68dd7e547c8739997f762b3963caef9cb1c

Issue #1798, opened 2017-12-05
Make image-caching enabled by default


LalatenduMohanty:
With merge of https://github.com/minishift/minishift/pull/1685 caching is optional and user can enable the container image caching by running `minishift config set image-caching true`. However we need to make it the enable by default as it will help saving users dependency on internet. Less dependency will help us towards improving the offline use of Minishift.  Before we make the the image caching enabled by default we need to run integration tests with `minishift config set image-caching true`.

PR #1954, opened 2018-01-30
Issue #1799 Make image caching default and run integration tests 

### LINKED ISSUE #1799: Run integration tests with image caching enabled
To achieve #1798, we need to run integration tests with image caching enabled

budhrg:
Fix #1799 and #1798 

The verifying of cached images is not based on the suggestion provided by Hardy here https://github.com/minishift/minishift/issues/1799#issuecomment-349320277.
It is more based on the output of `minishift image list` suggested by @agajdosi . 

In both cases, we had to poll for something like in monitoring PID case, we have to poll for process end.

Tested following cases:
- Fresh start
- Fresh start with image-caching true
- Fresh start with image-caching false

### DIFF
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -283,12 +283,17 @@ func initializeProfile() string {
 	return ""
 }
 
+func setViperConfigDefaults() {
+	viper.SetDefault(configCmd.ImageCaching.Name, constants.DefaultImageCaching)
+}
+
 func setupViper() {
 	viper.SetEnvPrefix(constants.MiniShiftEnvPrefix)
 	// Replaces '-' in flags with '_' in env variables
 	// e.g. show-libmachine-logs => $ENVPREFIX_SHOW_LIBMACHINE_LOGS
 	viper.SetEnvKeyReplacer(strings.NewReplacer("-", "_"))
 	viper.AutomaticEnv()
+	setViperConfigDefaults()
 	setFlagsUsingViper()
 }
 
--- a/docs/source/using/image-caching.adoc
+++ b/docs/source/using/image-caching.adoc
@@ -125,19 +125,19 @@ We recommend using this feature with caution.
 [[implicit-image-caching]]
 == Implicit Image Caching
 
-Image caching can be configured to automatically occur during the creation of the {project} VM while executing `minishift start`.
-To enable this feature you need to enable the `image-caching` property in the persistent configuration using the xref:../command-ref/minishift_config_set#[`minishift config set`] command:
-
-----
-$ minishift config set image-caching true
-----
-
-Once enabled, caching occurs in a background process, the first time you use the xref:../command-ref/minishift_start#[`minishift start`] command.
+Image caching is enabled by default for {project}.
+It occurs in a background process after the xref:../command-ref/minishift_start#[`minishift start`] command is completed for the first time.
 Once the images are cached under *_$MINISHIFT_HOME/cache/images_*, successive {project} VM creations will use these cached images.
+
+To disable this feature you need to disable the `image-caching` property in the persistent configuration using the xref:../command-ref/minishift_config_set#[`minishift config set`] command:
+
+----
+$ minishift config set image-caching false
+----
 
 [NOTE]
 ====
-Enabling implicit image caching will transparently add the required OpenShift images to the list of cached images as specified per `cache-images` configuration option.
+Implicit image caching will transparently add the required OpenShift images to the list of cached images as specified per `cache-images` configuration option.
 See xref:../using/image-caching.adoc#persisting-image-names[Persisting Cached Image Names].
 ====
 
@@ -146,7 +146,7 @@ See xref:../using/image-caching.adoc#persisting-image-names[Persisting Cached Im
 Each time an image exporting background process runs, a log file is generated under *_$MINISHIFT_HOME/logs_* which can be used to verify the progress of the export.
 ====
 
-You can disable the caching of the OpenShift images by setting `image-caching` to `false` or removing the setting altogether using xref:../command-ref/minishift_config_unset#[`minishift config unset`]:
+You can re-enable the caching of the OpenShift images by setting `image-caching` to `true` or removing the setting altogether using xref:../command-ref/minishift_config_unset#[`minishift config unset`]:
 
 ----
 $ minishift config unset image-caching
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -27,8 +27,11 @@ Feature: Basic
 
   Scenario: Starting Minishift
     Given Minishift has state "Does Not Exist"
+      And image caching is disabled
      When executing "minishift start" succeeds
      Then Minishift should have state "Running"
+     When executing "minishift image list" succeeds
+     Then stdout should be empty
 
   Scenario: OpenShift is ready after startup
     After startup of Minishift OpenShift instance should respond correctly on its html endpoints
--- a/test/integration/features/cmd-addons.feature
+++ b/test/integration/features/cmd-addons.feature
@@ -82,6 +82,7 @@ Feature: Addons command and its subcommands
   Scenario: Starting Minishift with anyuid and xpaas add-ons enabled
   The addons are being applied in correct order.
     Given Minishift has state "Does Not Exist"
+      And image caching is disabled
      When executing "minishift start" succeeds
      Then Minishift should have state "Running"
      Then stdout should match
--- a/test/integration/features/cmd-config.feature
+++ b/test/integration/features/cmd-config.feature
@@ -203,6 +203,8 @@ user defined options which changes default behaviour of Minishift.
   Scenario: Minishift informs about starting with correct setup of memory, disk and CPU
   Note: Minishift rounds the values for the report to make it more readable.
         However original non-rounded values are used for the startup.
+    Given Minishift has state "Does Not Exist"
+      And image caching is disabled
      When executing "minishift start" succeeds
      Then stdout should match "Memory\s*:\s*3 GB"
      Then stdout should match "Disk size\s*:\s*25 GB"
--- a/test/integration/features/cmd-image.feature
+++ b/test/integration/features/cmd-image.feature
@@ -13,8 +13,11 @@ Feature: Basic image caching test
      """
      Running this command requires an existing 'minishift' VM, but no VM is defined.
      """
-
      When executing "minishift start" succeeds
+      And image export completes with 3 images
+      And container image "openshift\/origin-haproxy-router:v[0-9]+\.[0-9]+\.[0-9]+" is cached
+      And container image "openshift\/origin-docker-registry:v[0-9]+\.[0-9]+\.[0-9]+" is cached
+      And container image "openshift\/origin:v[0-9]+\.[0-9]+\.[0-9]+" is cached
       And executing "minishift image export alpine:latest" succeeds
      Then stdout of command "minishift image list" contains "alpine:latest"
 
@@ -22,7 +25,8 @@ Feature: Basic image caching test
      When executing "minishift delete --force" succeeds
      Then stdout of command "minishift image list" contains "alpine:latest"
 
-  Scenario: As a user I can import a container image from the local cache into a running Minishift instance
+  Scenario: As a user I can reuse the cached images on next start and also import a container image from the local
+    cache into a running Minishift instance.
     Note: In this scenario we use alpine:latest which was cached in the previous scenario
 
     Given Minishift has state "Does Not Exist"
@@ -36,8 +40,12 @@ Feature: Basic image caching test
      """
 
      When executing "minishift start" succeeds
-      And executing "minishift image list --vm" succeeds
-      And stdout should not contain "alpine:latest"
+     Then stdout should match "Importing 'openshift\/origin:v[0-9]+\.[0-9]+\.[0-9]+' [\.]+ OK"
+      And stdout should match "Importing 'openshift\/origin-docker-registry:v[0-9]+\.[0-9]+\.[0-9]+' [\.]+ OK"
+      And stdout should match "Importing 'openshift\/origin-haproxy-router:v[0-9]+\.[0-9]+\.[0-9]+' [\.]+ OK"
+
+     When executing "minishift image list --vm" succeeds
+     Then stdout should not contain "alpine:latest"
       And executing "minishift image import alpine:latest" succeeds
       And executing "minishift image list --vm" succeeds
      Then stdout should contain "alpine:latest"
--- a/test/integration/integration_test.go
+++ b/test/integration/integration_test.go
@@ -176,6 +176,11 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^([^"]*) of command "minishift (.*)" (contains|does not contain) "(.*)"$`,
 		commandReturnContains)
 
+	// setting image caching operation
+	s.Step(`^image caching is (disabled|enabled)$`, minishift.setImageCaching)
+	s.Step(`^image export completes with (\d+) images$`, minishift.imageExportShouldComplete)
+	s.Step(`^container image "(.*)" is cached$`, minishift.imageShouldHaveCached)
+
 	// steps to execute `oc` commands
 	s.Step(`^executing "oc (.*)" retrying (\d+) times with wait period of (\d+) seconds$`,
 		minishift.executingRetryingTimesWithWaitPeriodOfSeconds)
@@ -402,12 +407,7 @@ func ensureTestDirEmpty() {
 	}
 
 	for _, file := range files {
-		fullPath := filepath.Join(testDir, file.Name())
-		if filepath.Base(file.Name()) == "cache" {
-			fmt.Println(fmt.Sprintf("Keeping Minishift cache directory '%s' for test run.", fullPath))
-			continue
-		}
-		os.RemoveAll(fullPath)
+		os.RemoveAll(filepath.Join(testDir, file.Name()))
 	}
 }
 
@@ -415,6 +415,7 @@ func cleanTestDirConfiguration() {
 	var foldersToClean []string
 	foldersToClean = append(foldersToClean, filepath.Join(testDir, "addons"))
 	foldersToClean = append(foldersToClean, filepath.Join(testDir, "config"))
+	foldersToClean = append(foldersToClean, filepath.Join(testDir, "cache/images"))
 
 	for index := range foldersToClean {
 		err := os.RemoveAll(foldersToClean[index])
--- a/test/integration/minishift.go
+++ b/test/integration/minishift.go
@@ -190,6 +190,47 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
 	return nil
 }
 
+func (m *Minishift) setImageCaching(operation string) error {
+	enabled := "true"
+	if operation == "disabled" {
+		enabled = "false"
+	}
+
+	return m.executingMinishiftCommand(fmt.Sprintf("config set image-caching %s", enabled))
+}
+
+func (m *Minishift) imageExportShouldComplete(noOfImages int) error {
+	// poll till the output of the `minishift image list` shows number of cached images
+	timeout := time.NewTimer(20 * time.Minute)
+
+outerPollActive:
+	for {
+		select {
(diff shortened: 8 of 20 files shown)


docs/source/using/image-caching.adoc
```diff
@@ -146,8 +146,4 @@ See xref:../using/image-caching.adoc#persisting-image-names[Persisting Cached Im
 Each time an image exporting background process runs, a log file is generated under *_$MINISHIFT_HOME/logs_* which can be used to verify the progress of the export.
 ====
 
-You can disable the caching of the OpenShift images by setting `image-caching` to `false` or removing the setting altogether using xref:../command-ref/minishift_config_unset#[`minishift config unset`]:
-
-----
-$ minishift config unset image-caching
-----
+You can again enable the caching of the OpenShift images by setting `image-caching` to `true`.
```
budhrg: Not sure should we use `minishift config unset image-caching` here.

test/integration/minishift.go
```diff
@@ -190,6 +190,43 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
...
+	if operation == "disabled" {
+		enabled = "false"
+	}
+
+	return m.executingMinishiftCommand(fmt.Sprintf("config set image-caching %s", enabled))
+}
+
+func (m *Minishift) imageExportShouldComplete() error {
+	// poll till the output of the `minishift image list` shows three OpenShift cached images
+	timeout := time.NewTimer(20 * time.Minute)
```
budhrg: Took ~15 min in my laptop.

test/integration/minishift.go
```diff
@@ -190,6 +190,43 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
 	return nil
 }
 
+func (m *Minishift) setImageCaching(operation string) error {
```
anjannath: can we please modify this to:
```go
func (m *Minishift) setImageCaching(operation string) error {
    return m.executingMinishiftCommand(fmt.Sprintf("config set image-caching %s", operation))
}
```
writing `setImageCaching("true")` kinda feels more intuitive to me.
I haven't gone through the feature files.

test/integration/integration_test.go
```diff
@@ -176,6 +176,11 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^([^"]*) of command "minishift (.*)" (contains|does not contain) "(.*)"$`,
 		commandReturnContains)
 
+	// setting image caching operation
+	s.Step(`^image caching is (disabled|enabled)$`, minishift.setImageCaching)
```
agajdosi: Here maybe the 2 line format we use in other definitions should be followed. (However in current 1 line format it is quite nicely readable. We might reconsider.)

test/integration/minishift.go
```diff
@@ -190,6 +190,43 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
...
+	// poll till the output of the `minishift image list` shows three OpenShift cached images
+	timeout := time.NewTimer(20 * time.Minute)
+
+outerPollActive:
+	for {
+		select {
+		case <-timeout.C:
+			return errors.New("Timed out in getting the number of default cached images")
+		default:
+			cmdOut, _, _ := m.runner.RunCommand("image list")
```
agajdosi: Maybe it would be safer to also check the cmdErr and exitCode. If a three line error occurs it would pass.

test/integration/minishift.go
```diff
@@ -190,6 +190,43 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
...
+			cmdOut = strings.TrimRight(cmdOut, "\n")
+			if len(strings.Split(cmdOut, "\n")) == noOfImages {
```
agajdosi: If higher than expected I think we could return error?

test/integration/minishift.go
```diff
@@ -190,6 +190,43 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
...
+	enabled := "true"
+	if operation == "disabled" {
+		enabled = "false"
+	}
+
+	return m.executingMinishiftCommand(fmt.Sprintf("config set image-caching %s", enabled))
+}
+
+func (m *Minishift) imageExportShouldComplete(noOfImages int) error {
+	// poll till the output of the `minishift image list` shows three OpenShift cached images
```
agajdosi: `three` should be changed to `number of images` or `noOfImages`.

test/integration/minishift.go
anjannath: This is fine, as it is aligned with the step `And image caching is disabled`.

test/integration/integration_test.go
budhrg: Don't have the strong opinion on it. My general rule is readability. 
With long lines, two lines is good but for smaller line one line is good.

test/integration/minishift.go
budhrg: @agajdosi Yes, forgot to update comment :) when I added argument.

test/integration/minishift.go
budhrg: It will break from the loop whenever it reaches `noOfImages` but it would be worthing thinking if `image list` will give me result greater than `noOfImages` then whether we should break out of loop or return error.

agajdosi: @budhrg If the number of images is greater than noOfImages I would suggest to error as we clearly define we expect it to be for example 3 images. However it is quite possible that even if it would download 4 images, they will not finish in same window of 5 seconds and it will still pass.

test/integration/minishift.go
```diff
@@ -190,6 +190,47 @@ func (m *Minishift) executingMinishiftCommand(command string) error {
...
+		case <-timeout.C:
+			return errors.New("Timed out in getting the number of default cached images")
+		default:
+			cmdOut, _, _ := m.runner.RunCommand("image list")
+			cmdOut = strings.TrimRight(cmdOut, "\n")
+			numOfLines := len(strings.Split(cmdOut, "\n"))
+			if numOfLines == noOfImages {
+				break outerPollActive
+			}
+			if numOfLines > noOfImages {
```
budhrg: This is added in case, `minishift image list` shows 5 images upfront and we are expecting 3.

test/integration/integration_test.go
agajdosi: @budhrg True, lets keep it 👍

test/integration/minishift.go
agajdosi: @budhrg You are right. This is fine then.

cmd/minishift/cmd/start.go
```diff
@@ -395,8 +395,32 @@ func addActiveProfileInformation() {
...
+		return false, err
+	}
+	for k := range config {
+		if k == configCmd.ImageCaching.Name {
+			defaultBehavior = viper.GetBool(configCmd.ImageCaching.Name)
+		}
+	}
+
+	return defaultBehavior, nil
+}
```
praveenkumar: Can we move this function to `cmd/util/util.go`? I don't like expending `start.go` day by day. Logic wise 👍

docs/source/using/image-caching.adoc
```diff
@@ -125,19 +125,19 @@ We recommend using this feature with caution.
 [[implicit-image-caching]]
 == Implicit Image Caching
 
-Image caching can be configured to automatically occur during the creation of the {project} VM while executing `minishift start`.
-To enable this feature you need to enable the `image-caching` property in the persistent configuration using the xref:../command-ref/minishift_config_set#[`minishift config set`] command:
+Image caching has been configured to automatically occur during the creation of the {project} VM while executing `minishift start`.
+After the completion of `minishift start`, caching occurs in a background process, the first time you use the xref:../command-ref/minishift_start#[`minishift start`] command.
```
praveenkumar: `After the completion of `minishift start`, caching occurs` => `After the completion of `minishift start`, image caching occurs`

docs/source/using/image-caching.adoc
```diff
@@ -125,19 +125,19 @@ We recommend using this feature with caution.
...
+Once the images are cached under *_$MINISHIFT_HOME/cache/images_*, successive {project} VM creations will use these cached images.
+
+To disable this feature you need to disable the `image-caching` property in the persistent configuration using the xref:../command-ref/minishift_config_set#[`minishift config set`] command:
 
 ----
-$ minishift config set image-caching true
+$ minishift config set image-caching false
```
praveenkumar: 👍

test/integration/features/basic.feature
```diff
@@ -29,6 +29,10 @@ Feature: Basic
     Given Minishift has state "Does Not Exist"
      When executing "minishift start" succeeds
      Then Minishift should have state "Running"
+      And image export completes with 3 images
+      And container image "openshift/origin-haproxy-router:v3.6.1" is cached
+      And container image "openshift/origin-docker-registry:v3.6.1" is cached
+      And container image "openshift/origin:v3.6.1" is cached
```
praveenkumar: These all should have a regex not hardcoded tag like v3.6.1 because right now it works as our default OpenShift version is 3.6.1 but when we update our default version then it will fail 👎

test/integration/features/cmd-addons.feature
```diff
@@ -82,6 +82,7 @@ Feature: Addons command and its subcommands
   Scenario: Starting Minishift with anyuid and xpaas add-ons enabled
   The addons are being applied in correct order.
     Given Minishift has state "Does Not Exist"
+      And image caching is disabled
      When executing "minishift start" succeeds
```
amitkrout: IMO some kind of start log validation required to confirm that it uses the cached image.  
For example: 
```When executing "minishift start" succeeds
    Then stdout should contain/match "caching image string"
```

test/integration/features/cmd-config.feature
```diff
@@ -203,6 +203,7 @@ user defined options which changes default behaviour of Minishift.
   Scenario: Minishift informs about starting with correct setup of memory, disk and CPU
   Note: Minishift rounds the values for the report to make it more readable.
         However original non-rounded values are used for the startup.
+    Given image caching is disabled
```
amitkrout: This will prevent the default configuration to run if this feature is only executed i,e make integration GODOG_OPTS="-tags cmd-config". 
Same is applicable for all feature file

test/integration/features/basic.feature
amitkrout: +1 for regex

test/integration/features/basic.feature
```diff
@@ -29,6 +29,10 @@ Feature: Basic
     Given Minishift has state "Does Not Exist"
      When executing "minishift start" succeeds
```
amitkrout: ```Importing openshift/origin-docker-registry:v3.6.1 . CACHE MISS``` is little hard to understand. Would you please confirm the behaviors what i  am seeing in ```m start``` log. 
```make integration GODOG_OPTS="-tags basic"```
```                        
[...]    
    Importing openshift/origin:v3.6.1 . CACHE MISS
    Importing openshift/origin-docker-registry:v3.6.1 . CACHE MISS
    Importing openshift/origin-haproxy-router:v3.6.1  CACHE MISS
    Version: v3.6.1
     [...]
```

Comment (anjannath) [reply test/integration/features/basic.feature]: @amitkrout `CACHE MISS` means the images are not cached yet, i.e you are running `minishift start` for the first time with `image-caching` so this time it will cache the images and from the next start it will use the cached images, and you should not get this again.

Comment (budhrg) [reply test/integration/features/basic.feature]: It is expected at the moment. I think we shouldn't show it on first fresh start.

Comment (budhrg) [reply test/integration/features/basic.feature]: @amitkrout As per discussion with @LalatenduMohanty , let's wait for some time . If community reports back, then we can change this behavior.

File: test/integration/features/cmd-config.feature (hunk 14 above)
Comment (budhrg) [reply test/integration/features/cmd-config.feature]: Yes, a side effect of this current implementation. Can't do much here I guess.

Comment (amitkrout) [reply test/integration/features/cmd-config.feature]: Actually its blocking a single feature file run with default configuration. This should be handle somehow smartly otherwise single feature run will be executed without image caching enable (Here basic can run with default image caching enabled). 

We should wait for more feedback on this, else will document it or open a new issue to refactor it later 
@budhrg WDYT ?

File: test/integration/features/basic.feature (hunk 15 above)
Comment (amitkrout) [reply test/integration/features/basic.feature]: Thanks for the clarification. I was confused with ```CACHE MISS```. Can we use better phrase for this like ```SEARCHING CACHE```
@robin-owen Please suggest

File: test/integration/features/cmd-addons.feature (hunk 13 above)
Comment (agajdosi) [reply test/integration/features/cmd-addons.feature]: @amitkrout I think it is not going to cache neither use the cached images as the image caching is disabled.

File: test/integration/features/cmd-config.feature (hunk 14 above)
Comment (agajdosi) [reply test/integration/features/cmd-config.feature]: AFAIU this strategy is similar to what we did with proxy. It could be run with all features, but to be clear, we made it into just one feature, which focuses on it. So it is expected that `cmd-config.feature` to not run with image caching. However it is true, that in basic.feature we do not test that much - for example a second startup on which we would check that the caching actually happens is missing. This might be fixed in follow up issue in `cmd-image.feature`.

File: test/integration/features/cmd-image.feature [hunk 16]
```diff
@@ -58,7 +58,9 @@ Feature: Basic image caching test
       And JSON config file "config/config.json" contains key "cache-images" with value matching "[alpine:latest]"
 
      When executing "minishift start" succeeds
-     Then stdout of command "minishift image list --vm" contains "alpine:latest"
+     Then Minishift should have state "Running"
+      And image export completes with 4 images
```
agajdosi: Just realized this might be problematic, if we run this without the `basic.feature` it will fail. In test/integration/integration_test.go in cleanTestDirConfiguration() we should probably add `cache/images` to be also removed after the feature to secure isolation between features.

test/integration/features/basic.feature
```diff
@@ -29,6 +29,10 @@ Feature: Basic
...
      Then Minishift should have state "Running"
+      And image export completes with 3 images
+      And container image "openshift/origin-haproxy-router:v[0-9]+\.[0-9]+\.[0-9]+" is cached
+      And container image "openshift/origin-docker-registry:v[0-9]+\.[0-9]+\.[0-9]+" is cached
+      And container image "openshift/origin:v[0-9]+\.[0-9]+\.[0-9]+" is cached
```
praveenkumar: @amitkrout , @agajdosi  FYI from downstream we need to change it for sure because images have different name and tags

amitkrout: @praveenkumar ack. Once this pr is merged, will create a jira and change it accordingly in downstream

test/integration/integration_test.go
```diff
@@ -402,19 +407,15 @@ func ensureTestDirEmpty() {
 	}
 
 	for _, file := range files {
-		fullPath := filepath.Join(testDir, file.Name())
-		if filepath.Base(file.Name()) == "cache" {
-			fmt.Println(fmt.Sprintf("Keeping Minishift cache directory '%s' for test run.", fullPath))
-			continue
-		}
-		os.RemoveAll(fullPath)
+		os.RemoveAll(filepath.Join(testDir, file.Name()))
```
agajdosi: 👍

test/integration/features/cmd-image.feature
```diff
@@ -36,8 +40,12 @@ Feature: Basic image caching test
      """
 
      When executing "minishift start" succeeds
-      And executing "minishift image list --vm" succeeds
-      And stdout should not contain "alpine:latest"
+     Then stdout should match "Importing 'openshift\/origin:v[0-9]+\.[0-9]+\.[0-9]+' [\.]+ OK"
```
budhrg: Here, verifying that the importing of images happen successfully.

test/integration/integration_test.go
```diff
@@ -402,19 +407,15 @@ func ensureTestDirEmpty() {
...
 	}
 }
 
 func cleanTestDirConfiguration() {
 	var foldersToClean []string
 	foldersToClean = append(foldersToClean, filepath.Join(testDir, "addons"))
 	foldersToClean = append(foldersToClean, filepath.Join(testDir, "config"))
+	foldersToClean = append(foldersToClean, filepath.Join(testDir, "cache"))
```
agajdosi: This will slow the runs a little bit because we will download the `oc` on each time, but nothing critical I think. Alternative could be to delete just `cache/images`.

docs/source/using/image-caching.adoc
```diff
@@ -125,19 +125,19 @@ We recommend using this feature with caution.
...
 ----
 
-Once enabled, caching occurs in a background process, the first time you use the xref:../command-ref/minishift_start#[`minishift start`] command.
-Once the images are cached under *_$MINISHIFT_HOME/cache/images_*, successive {project} VM creations will use these cached images.
-
 [NOTE]
 ====
-Enabling implicit image caching will transparently add the required OpenShift images to the list of cached images as specified per `cache-images` configuration option.
+Implicit image caching will transparently add the required OpenShift images to the list of cached images as specified per `cache-images` configuration option.
```
robin-owen: 👍

docs/source/using/image-caching.adoc
```diff
@@ -146,7 +146,7 @@ See xref:../using/image-caching.adoc#persisting-image-names[Persisting Cached Im
 Each time an image exporting background process runs, a log file is generated under *_$MINISHIFT_HOME/logs_* which can be used to verify the progress of the export.
 ====
 
-You can disable the caching of the OpenShift images by setting `image-caching` to `false` or removing the setting altogether using xref:../command-ref/minishift_config_unset#[`minishift config unset`]:
+You can again enable the caching of the OpenShift images by setting `image-caching` to `true` or removing the setting altogether using xref:../command-ref/minishift_config_unset#[`minishift config unset`]:
```
robin-owen: I'd suggest "re-enable" instead of "again enable".

docs/source/using/image-caching.adoc
```diff
@@ -125,19 +125,19 @@ We recommend using this feature with caution.
 [[implicit-image-caching]]
 == Implicit Image Caching
 
-Image caching can be configured to automatically occur during the creation of the {project} VM while executing `minishift start`.
-To enable this feature you need to enable the `image-caching` property in the persistent configuration using the xref:../command-ref/minishift_config_set#[`minishift config set`] command:
+Image caching has been configured to automatically occur during the creation of the {project} VM while executing `minishift start`.
+After the completion of `minishift start`, image caching occurs in a background process, the first time you use the xref:../command-ref/minishift_start#[`minishift start`] command.
```
robin-owen: I know you didn't introduce this sentence, but I think we can shorten it while keeping the same meaning:

```adoc
Image caching occurs in a background process after the `minishift start` command is completed for the first time.
```

test/integration/minishift.go
amitkrout: I do not think this will be help full by counting the caching image number IMO. 

Consider this Usecase:
1. minishift start with default configuration (i mean image caching enabled by default) 
2. Export one image (minishift image export ...)
3. minishift delete
4. minishift start
What i am expecting here is that all four images will be cached, whereas the validation is done on number of lines i.e three.

budhrg: @amitkrout , 

> What i am expecting here is that all four images will be cached

This assumption is wrong here. Only openshift images will get automatically imported for next start. You need to import other images explicitely.

See the scenario https://github.com/budhrg/minishift/blob/issue-1799/test/integration/features/cmd-image.feature#L28-L51

cmd/minishift/cmd/root.go
```diff
@@ -283,12 +283,17 @@ func initializeProfile() string {
 	return ""
 }
 
+func setViperDefaultKeyValues() {
```
LalatenduMohanty: @budhrg I would suggest renaming the function as `setViperConfigDefaults` as `KeyValues` in `setViperDefaultKeyValues` does not make much sense to me.

test/integration/minishift.go
LalatenduMohanty: @amitkrout `noOfImages` is an argument to the function and the value comes from `s.Step(`^image export completes with (\d+) images$`, minishift.imageExportShouldComplete)` e.g. `And image export completes with 3 images`.So we are good here.

amitkrout: @LalatenduMohanty @budhrg thanks for the explanation. Actually was little confused with import and export.

PR #1974, opened 2018-02-03
Issue #1672 Show error when addon dir name is not equal to its name

### LINKED ISSUE #1672: Unable to uninstall add-on when directory name is different than add-on name in meta data
**Steps to reproduce**
- Create a addon directory `demoaddon` : `mkdir demoaddon`
- Create addon file : 
```
cat <<EOF > demoaddon/demoaddon.addon
# Name: demo-addon
# Description: Demo addon
# Required-Vars: USER,NAMESPACE
# Var-Defaults: NAMESPACE=kube-system

echo User is #{USER} and namespace is #{NAMESPACE}.
EOF
```
- Install addon
```
$ minishift addons install demoaddon
Addon 'demo-addon' installed
```
- List addons
``` 
$ minishift addons list
- demo-addon     : disabled   P(0)
```
- Uninstall addon
```
$ minishift addons uninstall demo-addon
Cannot Uninstall the add-on demo-addon: Unable to find addon directory /home/budhram/.minishift/addons/demo-addon
```

budhrg:
Fix #1672 and #2014 

Tried with following addon
```
$ mkdir demoaddon
$ cat demoaddon/demoaddon.addon
# Name: demo-addon
# Description: Demoaddon description
# Url: https://someurl.html
# Required-Vars: FOO
# Var-Defaults: FOO=abc

echo This is demo addon with variable FOO whose value is #{FOO}
```

###  Installing addon
```
$ minishift addon install demoaddon
Add-on installation failed with the error: Unable to parse specified addon: Add-on directory name 'demoaddon' should match to addon name 'demo-addon'
```

### DIFF
--- a/pkg/minishift/addon/parser/addon_parser.go
+++ b/pkg/minishift/addon/parser/addon_parser.go
@@ -83,19 +83,22 @@ func (parser *AddOnParser) getAddOnContent(addOnDir, fileSuffix string) (addon.A
 	}
 	if addonReader != nil {
 		meta, commands, err = parser.parseAddOnContent(addonReader)
+		var name string
+		if meta != nil {
+			name = meta.Name()
+		}
 		if err != nil {
-			name := ""
-			if meta != nil {
-				name = meta.Name()
-			}
 			return nil, nil, NewParseError(err.Error(), name, addOnDir)
+		}
+		if filepath.Base(addOnDir) != name {
+			return nil, nil, NewParseError(fmt.Sprintf("Add-on directory name should match to addon name"), "", addOnDir)
 		}
 	}
 
 	return meta, commands, nil
 }
 
-// Parse parses the addon files containing in a directory provided and returns an AddOn instance.
+// Parse parses the addon files containing in a directory provided via addOnDir and returns an AddOn instance.
 // If an error occurs, the error is returned.
 func (parser *AddOnParser) Parse(addOnDir string) (addon.AddOn, error) {
 	meta, commands, err := parser.getAddOnContent(addOnDir, ".addon")
--- a/pkg/minishift/addon/parser/addon_parser.go
+++ b/pkg/minishift/addon/parser/addon_parser.go
@@ -70,42 +70,44 @@ func NewAddOnParser() *AddOnParser {
 	return &parser
 }
 
-// Parse takes as parameter a reader containing an addon definition and returns an AddOn instance.
-// If an error occurs, the error is returned.
-func (parser *AddOnParser) Parse(addOnDir string) (addon.AddOn, error) {
-	addonInstallReader, err := parser.getAddOnContentReader(addOnDir, ".addon")
-	if err != nil {
-		return nil, err
-	}
-	addonRemoveReader, err := parser.getAddOnContentReader(addOnDir, ".addon.remove")
-	if err != nil {
-		return nil, err
-	}
-	meta, commands, err := parser.parseAddOnContent(addonInstallReader)
-	if err != nil {
-		name := ""
-		if meta != nil {
-			name = meta.Name()
-		}
-		return nil, NewParseError(err.Error(), name, addOnDir)
-	}
+func (parser *AddOnParser) getAddOnContent(addOnDir, fileSuffix string) (addon.AddOnMeta, []command.Command, error) {
+	var (
+		meta     addon.AddOnMeta
+		commands []command.Command
+		err      error
+	)
 
-	var removeCommands []command.Command
-	var metaDataForAddonRemove addon.AddOnMeta
-	if addonRemoveReader != nil {
-		metaDataForAddonRemove, removeCommands, err = parser.parseAddOnContent(addonRemoveReader)
+	addonReader, err := parser.getAddOnContentReader(addOnDir, fileSuffix)
+	if err != nil {
+		return nil, nil, err
+	}
+	if addonReader != nil {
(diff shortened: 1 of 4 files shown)


anjannath:
Works fine for me, showing error when addon have different directory name then the addon name.
```zsh
➜  ~ minishift addons install demoaddon 
Add-on installation failed with the error: Unable to parse specified addon: Add-on directory name should match to addon name
```

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -70,42 +70,47 @@ func NewAddOnParser() *AddOnParser {
...
 		if err != nil {
-			name := ""
-			if metaDataForAddonRemove != nil {
-				name = metaDataForAddonRemove.Name()
-			}
-			return nil, NewParseError(err.Error(), name, addOnDir)
+			return nil, nil, NewParseError(err.Error(), name, addOnDir)
+		}
+		if filepath.Base(addOnDir) != name {
+			return nil, nil, NewParseError(fmt.Sprintf("Add-on directory name should match to addon name"), addOnDir, "")
```
praveenkumar: This should be `return nil, nil, NewParseError("Add-on directory name should match to addon name","", addOnDir)` because you are not doing any formatting and `NewParseError` definition says `NewParseError(msg string, name string, dir string)` so `addOnDir` sould be last one.

pkg/minishift/addon/parser/addon_parser_test.go
```diff
@@ -277,3 +277,18 @@ func Test_multiple_addon_definitions_create_error(t *testing.T) {
...
+	meta, _, err := testParser.getAddOnContent(testAddonDir, ".addon")
+	assert.NoError(t, err, "Add-on directory name should match to addon name")
+	assert.Equal(t, testAddonName, meta.Name())
+}
+
+func TestInvalidAddonDirAndName(t *testing.T) {
+	testAddonName := "addon-dir-mismatch"
+	testAddonDir := filepath.Join(basepath, "..", "..", "..", "..", "test", "testdata", "testaddons", testAddonName)
+	_, _, err := testParser.getAddOnContent(testAddonDir, ".addon")
+	assert.EqualError(t, err, "Add-on directory name should match to addon name")
```
praveenkumar: 👍

pkg/minishift/addon/parser/addon_parser_test.go
```diff
@@ -277,3 +277,18 @@ func Test_multiple_addon_definitions_create_error(t *testing.T) {
 
 	assert.Nil(t, addon, "No addon should have been returned")
 }
+
+func TestValidAddonDirAndName(t *testing.T) {
```
praveenkumar: why we even need this test? Since add-on test is already in place which anyway going to use the same parse and also validate the directory mismatch. Let's remove this test.

praveenkumar:
@budhrg I am not sure putting commit for a refactor along with a working PR is good until we have a separate issue for it. what you can do like hardy used to do in this case in a single commit you can add what you did. This is not a blocker per se and I don't want to block this PR for this .@LalatenduMohanty, @gbraad  wdyt?

```
Issue #1692 Show error when addon dir name is not equal to its name

    - Refactored Addon parser parse function
```

budhrg:
> I am not sure putting commit for a refactor along with a working PR is good until we have a separate issue for it. what you can do like hardy used to do in this case in a single commit you can add what you did. 

IMO, this would help to separate actual fix change-set from refactoring work.
Now, here there are two commits https://github.com/minishift/minishift/pull/1974/commits.
Clicking on `Issue #1672 Show error when addon dir name is not equal to its name` commit shows me actual changes for this issue fix and looks fine and in fact less code changes than refactoring. 
If we get into single commit, this will be lost.

I am happy to create new issue if needed but I don't think we should create issue for such.

pkg/minishift/addon/parser/addon_parser.go
```diff
@@ -83,19 +83,22 @@ func (parser *AddOnParser) getAddOnContent(addOnDir, fileSuffix string) (addon.A
 	}
 	if addonReader != nil {
 		meta, commands, err = parser.parseAddOnContent(addonReader)
+		name := ""
```
LalatenduMohanty: personally, I would have done `var name string`

LalatenduMohanty:
> @budhrg I am not sure putting commit for a refactor along with a working PR is good until we have a separate issue for it. what you can do like hardy used to do in this case in a single commit you can add what you did.

+1 I agree. Both ways work for me.

LalatenduMohanty:
The reason we keep issues in the commit message because it helps us to backtrack changes for each issue, but when we have code changes that not related to that issue it is against the original reason we put issue numbers in the commit message. But when working on the actual bug we can find some code to refactor and it should be fine in one commit as long as it is not a lot of code. There is no way to quantify how much code is a lot of code, so it would depend on PR to PR. If the PR submitter and reviewer agrees to it then it should be fine.

budhrg:
> There is no way to quantify how much code is a lot of code, so it would depend on PR to PR. If the PR submitter and reviewer agrees to it then it should be fine.

Then, I wouldn't agree in this PR :smile:  to include refactored code in same commit since I am losing the fix in the refactoring change set.
See below, the red mark is the actual fix and other change set is refactoring. I cann't even guess what fix I did here.

screenshot from 2018-02-12 17-57-22

Issue #1970, opened 2018-02-02
Improve packaging capabilities of code around integration tests


agajdosi:
### General information
Currently the integration tests can't be easily used outside of the Minishift repository.

### Current
The `integration_test.go` is mixture of both code which starts the tests and code which defines them. This applies to `integration` package as whole.

### Expected
Move all the code which defines tests and covers its functionality into separate package. This package will be imported into the `integration_test.go` which itself will be very short and will not carry anything extra for the actual tests. If one would need to use the newly created integration test package she could import it the same way as in `integration_test.go` file.

### Relevant issues
https://github.com/minishift/minishift-addons/issues/67
https://github.com/minishift/minishift/pull/1967
https://github.com/minishift/minishift/issues/1968

agajdosi:
Proof of concept at https://github.com/agajdosi/minishift/commit/1bafb75a3a28cd0d452aa46fa53dbac557ac0210. Any comments appreciated.

gbraad:
Can the expected workflow be described? for instance, what does an addon developer need to do to make this work for him?

LalatenduMohanty:
@agajdosi One if the challenge we will have is to keep the code for the integration tests always up to date with respect to Minishift code repository and having a separate package for it can help.

agajdosi:
@gbraad @LalatenduMohanty The workflow I plan is that: in https://github.com/minishift/minishift-addons/pull/85 I add just minimal `integration-test.go` file, which imports the package from `github.com/minishift/minishift/test/integration/testsuite` + addon developer - like JPinkney can load his own package containing his own step definitions.

However even without any custom packaging developer could write gherkin feature files using all the steps which are available here in Minishift, IMHO if I am correct that could be enough to write basic smoke tests for addons.

When one then wants to run the tests, she would need Go 1.8+, dep, make on path, downloaded Minishift binary and having prerequisites for Minishift installed. Then one would need to run `dep` to vendor packages (I have there `make vendor` target for that), after its vendored, one can place the Minishift binary into `testing/bin` and run just `make integration` or `make integration MINISHIFT_BINARY=<path-to-binary>` if she wants to use custom path.

I am also thinking about having a shortcut for the tags, so ideally for one addon one would run: `make integration ADDON=cors` and that's it!

You can check https://github.com/agajdosi/minishift-addons/blob/718a503dbae1a48b9ec1cb21791d4505ea8486d8/CONTRIBUTING.adoc where I have added information about testing.

[...]

PR #1975, opened 2018-02-05
Issue #1970 Moving integration test code to a separate go package

agajdosi:
This PR tries to move as much as possible from the `integration_test.go` file into a new package `testsuite`, leaving in `integration_test.go` only calls to flag parsing, test preparation and initialization of actual Godog test with godog.RunWithOptions().

The aim is to provide the ability to run integration tests outside of the Minishift repository using the `testsuite`, `util` and `proxy` packages and customized `integration_test.go` in which one can add custom step definitions and flags.

This PR works right now but it amplifies my feeling that the packages in the `test/integration` are a little bit chaotically organized and that it will need some care in future.

Please let me know your ideas around this change, I am now trying to implement it onto `minishift-addons`.


agajdosi:
To test the packaging ability which this PR brings you can use the https://github.com/agajdosi/minishift-addons/tree/67:

1. Pull the branch
2. run `dep ensure -v`
3. run `make integration MINISHIFT_BINARY=<path-to-minishift-binary>`

This should trigger a test which only checks minishift status and runs a step which definition lies in minishift-addons and not in Minishift repository. If the test passes we are fine.

agajdosi:
@minishift/minishift-dev If you have time, please take a look, any feedback or ack is welcomed as this blocks the implementation of the Che addon integration test in the `minishift/minishift-addons` repository. Thanks :)

agajdosi:
@LalatenduMohanty promised ping :) please take a look once you have time. All seems to work fine, I am just not sure why it fails in circleci, locally the `make validate_commits` passes fine.

test/integration/testsuite/prepare.go
```diff
@@ -0,0 +1,214 @@
...
+			fmt.Println("Test run using CentOS iso image.")
+			isoName = "centos"
+		default:
+			fmt.Print("Using full path for iso image.")
+			isoName = determineIsoFromFile(isoUrl)
+		}
+	}
+}
+
+func determineIsoFromFile(isoUrl string) string {
```
praveenkumar: @agajdosi Shouldn't we just validate the IsoURL our and return as it is instead of returning the `b2d, minikube..etc.`. what if I have a custom iso to test and uploaded it some location then integration test might just return `b2d`? I just find it odd.

agajdosi: @praveenkumar You can name it to end with `centos7.iso`, then it should be fine.

> Shouldn't we just validate the IsoURL our and return as it is

Later integration tests needs `b2d`, `minikube`, `centos` or `rhel` to determine in some cases during integration tests.

gbraad: too much logic... and I do not see a real need to check if CentOS, Minikube, etc from the prepare code... the test suite should have as little information about this as possible.

agajdosi: > the test suite should have as little information about this as possible.

I agree @gbraad, but for example for Docker config I have not found any other way of checking it than cat-ing it. And unfortunately the path differs on all platforms: [catDockerConfigFile()](https://github.com/agajdosi/minishift/blob/dbcf2c4cbb9f396f63d7e5d9c1cd4f8186b5e91a/test/integration/testsuite/testsuite.go#L708).

I suggest to create an issue for the issue you see here and we can then change the code. However for now I am not adding it, instead just moving it.

praveenkumar: @agajdosi Alright so I understand that some of the integration tests behave as per iso so you need to make sure that shouldn't fail in case of custom ISO URI. Fair enough.

This complete PR is just shuffling code base in a package structure nothing more right?

praveenkumar:
@agajdosi Also in the commit message instead `Issue #1970 Put itests into package` use ` Issue #1970 Put integrations tests into package `.

test/integration/integration_test.go
```diff
@@ -19,94 +19,27 @@ limitations under the License.
...
-	copyOcFrom       string
-
-	// Godog options
-	godogFormat              string
-	godogTags                string
-	godogShowStepDefinitions bool
-	godogStopOnFailure       bool
-	godogNoColors            bool
-	godogPaths               string
+	"github.com/minishift/minishift/test/integration/testsuite"
```
LalatenduMohanty: I think we need a better name for this. I can not come up with any better name. @minishift/minishift-dev ?

test/integration/testsuite/prepare.go
agajdosi: @praveenkumar Yes, it is just a shuffle.

> so you need to make sure that shouldn't fail in case of custom ISO URI

True, currently it is just 1 step which is specific + it is handy for the tag handling, but to be prepared there is this functions so we have the information which can be used if any new test is iso specific.

test/integration/integration_test.go
agajdosi: @LalatenduMohanty Agree, it is not a strong name. Trying to follow Go advices to keep it short this was the best I realized, but there might be better, would be happy to find it :)

agajdosi:
@praveenkumar Commit message changed, thanks for noticing.

test/integration/integration_test.go
gbraad: I actually think it is an OK name

agajdosi:
retest this please

docs/source/contributing/developing.adoc
```diff
@@ -360,6 +360,15 @@ This information is often enough to find and debug the reason for a failure.
 However, for cases which require further investigation, the integration test also logs more detailed progress into a log file.
 This file is located at *_$GOPATH/github.com/minishift/minishift/out/test-results/integration_YYYY-MM-DD_HH:MM:SS.log_*.
 
+
+[[integration-tests-as-package]]
+==== Using Integration Tests as a Package
+
+Integration tests can be also used as a package outside of the {project} repository.
+You can import it as `github.com/minishift/minishift/test/integration/testsuite`.
+As a starting point for your project you can copy and use the `test/integration/integration_test.go` file.
```
budhrg: Hyperlink `test/integration/integration_test.go` to respective file is needed.

agajdosi: @budhrg thanks for suggestion, should be fixed now: `link:https://github.com/minishift/minishift/blob/master/test/integration/integration_test.go[`test/integration/integration_test.go`]`

Issue #1986, opened 2018-02-07
Use a user friendly error message when no *.addon.remove file available


amitkrout:
### General information

  * Minishift version: master 
  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce
```
  1. ./minishift start
  2. ./minishift addons apply anyuid
  3. ./minishift addons remove anyuid
      No anyuid.addon.remove file found for 'anyuid' add-on.
  ```

### Expected
```
   $ ./minishift addons remove anyuid
      Unable to remove addon: 'anyuid'. No anyuid.addon.remove file is found.
```

Out put should be something like as mentioned above suggested by @gbraad 

### Actual
$ ./minishift addons remove anyuid
No anyuid.addon.remove file found for 'anyuid' add-on.

By seeing this output user can have a feeling of something went wrong in the addon remove process which is actually baseless. Moreover this addons can not be removed once applied, hence no .remove file is available.    

### Logs

You can start Minishift with `minishift start --show-libmachine-logs -v5` to collect logs.
Please consider posting this on http://gist.github.com/ and post the link in the issue.

praveenkumar:
I added `good first issue` label, if we don't get a newbie contributor then will take it during '1.15.0` milestone.

amitkrout:
@praveenkumar i want to take this. This way i get some more exposer to minishift addons and its work flow

praveenkumar:
@amitkrout I set milestone `1.15.0` so if you want to do it now please adjust the milestone accordingly.

LalatenduMohanty:
@anjannath @minishift/minishift-dev Usually for issues which are not planned for a milestone but it is getting fixed because it is easy or someone wants to fix it then we do not want to add the current milestone to it. But when the issue gets resolved we should add the correct milestone. Adding the current milestone means it needs to be fixed in this milestone because it is important.

anjannath:
@LalatenduMohanty Ahh, I didn't know it was done like that, Thanks for making it clear, Will remember it.

PR #1991, opened 2018-02-07
Issue #1986 User friendly error message for *.addon.remove file

amitkrout:
Fix : #1986

### DIFF
--- a/cmd/minishift/cmd/addon/addon.go
+++ b/cmd/minishift/cmd/addon/addon.go
@@ -25,7 +25,7 @@ const (
 
 	emptyAddOnError      = "You must specify an add-on name. Run 'minishift addons list' to view installed add-ons."
 	noAddOnMessage       = "No add-on with the name '%s' is installed."
-	noRemoveAddOnMessage = "No %s.addon.remove file found for '%s' add-on."
+	noRemoveAddOnMessage = "Unable to remove addon '%s'. No %s.addon.remove file is found."
 )
 
 var AddonsCmd = &cobra.Command{
--- a/test/integration/features/cmd-addons.feature
+++ b/test/integration/features/cmd-addons.feature
@@ -37,7 +37,7 @@ Feature: Addons command and its subcommands
      When executing "minishift addons remove anyuid" succeeds
      Then stdout should contain
       """
-      No anyuid.addon.remove file found for 'anyuid' add-on.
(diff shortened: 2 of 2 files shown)


amitkrout:
This failure is coming from file ```provision_earlier_version.feature``` in starting minishift, resulting some dependent steps skip and failure. The same feature file executed successfully in my local fedora27. So In re-triggering there is a great chance of not hitting the same failure issue. 

This king of failure need to be documented in the upstream as ```Integration test most common failure``` IMO. It would be really helpful for the new contributor mostly newbie for example if the pr fails with these errors they should not be in panic. @minishift/minishift-dev Thoughts ?

LalatenduMohanty:
retest this please

cmd/minishift/cmd/addon/addon.go
```diff
@@ -25,7 +25,7 @@ const (
 
 	emptyAddOnError      = "You must specify an add-on name. Run 'minishift addons list' to view installed add-ons."
 	noAddOnMessage       = "No add-on with the name '%s' is installed."
-	noRemoveAddOnMessage = "No %s.addon.remove file found for '%s' add-on."
+	noRemoveAddOnMessage = "Unable to remove addon: '%s'. No %s.addon.remove file is found."
```
LalatenduMohanty: s/addon: '%s'/e addon '%s'/. We should remove the `:`.

LalatenduMohanty:
> This king of failure need to be documented in the upstream as Integration test most common failure IMO. It would be really helpful for the new contributor mostly newbie for example if the pr fails with these errors they should not be in the panic. @minishift/minishift-dev Thoughts ?

Actually, I have not seen this failure much in the past. Not sure if others have seen it. IMO we can not take any test failure as known issue or common failure. So we should do something to the fix the test itself and not be documenting this as known issues.

amitkrout:
> Actually, I have not seen this failure much in the past. Not sure if others have seen it. 

This is not consistent in ci even local system i seldom face such failure. 
 
> IMO we can not take any test failure as known issue or common failure. So we should do something to the fix the test itself and not be documenting this as known issues.

Sure, in the next appearance of such failure, we will file a github issue.

cmd/minishift/cmd/addon/addon.go
amitkrout: Done

PR #2000, opened 2018-02-09
Issue #1994 Show app builds logs during integration tests on failure

### LINKED ISSUE #1994: Show app builds logs during integration tests on failure
As part of this PR https://github.com/minishift/minishift/pull/1951, we exposed logs of `stdout` and `stderr` which is not helping much IMO

screenshot from 2018-02-08 12-24-32

It would be great if we get logs of origin container and application container logs on failure.
Really getting difficult to debug https://github.com/minishift/minishift/issues/1857 and https://github.com/minishift/minishift/issues/1993

budhrg:
Fix #1994

This will help in debugging rollout steps which is often used in application deployment which is the reason of nightly jobs failure in Minikube ISO.

screenshot from 2018-02-09 16-33-34

### DIFF
--- a/test/integration/testsuite/minishift.go
+++ b/test/integration/testsuite/minishift.go
@@ -268,12 +268,21 @@ func (m *Minishift) checkServiceRolloutForSuccess(service string, timeout int, d
 	if strings.Contains(cmdOut, expected) {
 		done <- rolloutMessage{passed: true, stdErr: cmdErr, stdOut: cmdOut}
 	} else {
+		// get application's build logs if rollout fails
+		command = fmt.Sprintf("logs bc/%s", service)
+		m.executingOcCommand(command)
+
+		lastCmdResult := getLastCommandOutput()
+
+		cmdOut += fmt.Sprintf("\n Service build output logs: %s\n", lastCmdResult.StdOut)
+		cmdErr += fmt.Sprintf("\n Service build error logs: %s\n", lastCmdResult.StdErr)
+
 		done <- rolloutMessage{passed: false, stdErr: cmdErr, stdOut: cmdOut}
 	}
 }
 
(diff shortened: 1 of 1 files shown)


test/integration/minishift.go
```diff
@@ -263,6 +263,15 @@ func (m *Minishift) checkServiceRolloutForSuccess(service string, timeout int, d
 		})
 	m.mutex.Unlock()
 
+	// get application's build logs
+	command = fmt.Sprintf("logs bc/%s", service)
```
LalatenduMohanty: @budhrg This is going to happen irrespective of pass or fail, isn't it? Should we try to make it optional i.e. enable it somehow if you want to get more information?

gbraad:
Try adding entries to github.com in /etc/hosts

budhrg:
> Try adding entries to github.com in /etc/hosts

Of build slave?

LalatenduMohanty:
> Try adding entries to github.com in /etc/hosts

It can not be the problem as preflight check passes (i.e. minishift start works). It has to be something else IMO.

test/integration/minishift.go
budhrg: Done @LalatenduMohanty

budhrg:
@LalatenduMohanty We get build logs only on error. See
### Success
screenshot from 2018-02-10 13-42-46
### Failure
screenshot from 2018-02-10 13-43-17

gbraad:
On Saturday, February 10, 2018, Lalatendu Mohanty <notifications@github.com>
wrote:
> Try adding entries to github.com in /etc/hosts
>
> It can not be the problem as preflight check passes (i.e. minishift start
works). It has to be something else IMO.

Preflight doesn't detect intermittent issues with networking. A failing
resolve to github.com indicates a networking issues

-- 

   [ Doing Open Source Matters ]

PR #2010, opened 2018-02-12
[test] #1776 Check container status after Openshift restart

agajdosi:


agajdosi:
Testing now 1 second delay, will close the PR after.

agajdosi:
@LalatenduMohanty It needs more than a 1 second.

PR #2013, opened 2018-02-12
Issue #2012 Print OpenShift restart message post its restart operation

### LINKED ISSUE #2012: Print OpenShift restart message post its restart operation
### Expected

Right now we are printing message of `Restarting OpenShift` before running restart operation through commander.
https://github.com/minishift/minishift/blob/c5ad41134f490ff6d3c5ff9ae7cf5f7285bbf8c4/pkg/minishift/openshift/openshift.go#L71-L78

And with using this message in the `cmd-openshift.feature` 
https://github.com/minishift/minishift/blob/c5ad41134f490ff6d3c5ff9ae7cf5f7285bbf8c4/test/integration/features/cmd-openshift.feature#L34-L35

it is giving false assumption that at this point Openshift has been restarted.

### Actual

The message should be printed after restart operation.

budhrg:
Fix #2012

### DIFF
--- a/cmd/minishift/cmd/openshift/restart.go
+++ b/cmd/minishift/cmd/openshift/restart.go
@@ -56,4 +56,6 @@ func runRestart(cmd *cobra.Command, args []string) {
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error restarting the OpenShift cluster: %s", err.Error()))
 	}
+
+	fmt.Println("OpenShift restarted successfully")
 }
--- a/pkg/minishift/openshift/openshift.go
+++ b/pkg/minishift/openshift/openshift.go
@@ -69,11 +69,11 @@ func (t *OpenShiftPatchTarget) localConfigFilePath() (string, error) {
 }
 
 func RestartOpenShift(commander docker.DockerCommander) (bool, error) {
-	fmt.Println("Restarting OpenShift")
 	ok, err := commander.Restart(constants.OpenshiftContainerName)
 	if err != nil {
 		return false, err
 	}
+
 	return ok, nil
 }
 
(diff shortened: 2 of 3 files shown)


pkg/minishift/openshift/openshift.go
```diff
@@ -69,11 +69,11 @@ func (t *OpenShiftPatchTarget) localConfigFilePath() (string, error) {
 }
 
 func RestartOpenShift(commander docker.DockerCommander) (bool, error) {
-	fmt.Println("Restarting OpenShift")
 	ok, err := commander.Restart(constants.OpenshiftContainerName)
 	if err != nil {
 		return false, err
 	}
+	fmt.Println("OpenShift restarted successfully")
```
LalatenduMohanty: I would suggest putting the success message in cmd/openshift/restart.go as ideally, the package should only return to cmd and cmd should success or failure log to console. I know we are not consistent in the code about it, but I think that's the right thing to do.

gbraad:
Not understanding the commit message... please use `"` or "," to inidcate your intent.

pkg/minishift/openshift/openshift.go
gbraad: agree with @LalatenduMohanty to move this out of the code block. Since we already retrurn a`bool` to indicate the state, we can easily handle this outside this function.

cmd/minishift/cmd/openshift/restart.go
```diff
@@ -56,4 +56,6 @@ func runRestart(cmd *cobra.Command, args []string) {
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error restarting the OpenShift cluster: %s", err.Error()))
 	}
+
+	fmt.Println("OpenShift restarted successfully")
```
anjannath: :+1:

test/integration/features/cmd-openshift.feature
```diff
@@ -32,7 +32,7 @@ cluster in VM provided by Minishift.
         after next start of container.
      Given stdout of command "minishift ssh -- "docker inspect --format={{.State.FinishedAt}} origin"" is equal to "0001-01-01T00:00:00Z"
       When executing "minishift openshift restart" succeeds
-      Then stdout should contain "Restarting OpenShift"
+      Then stdout should contain "OpenShift restarted successfully"
```
praveenkumar: :+1:

PR #2032, opened 2018-02-14
Issue #1971 Add-on to change the value of 'max-pods' in node config

### LINKED ISSUE #1971: Change limit pod/cpu
Statement: 

"I experiment from time to time such problem on my machine using "oc cluster up"  (openshift origin 3.7.x)

`0/1 nodes are available: 1 Insufficient pods.`

If I look to the to stats generated by the command "oc get node localhost", then I can see that the limit of the pods created has been reached.

As the convention followed by openshift is to limit the number of the pods "available" to 10pods/cpu, I'm wondering how we could change this value when a cluster is started ?"

According to the openshift team, such limit can be managed using this parameter

"this limit can be changed in the node configuration, see: https://docs.openshift.com/container-platform/3.7/admin_guide/manage_nodes.html#admin-guide-max-pods-per-node
Looking at https://github.com/openshift/origin/blob/master/docs/cluster_up_down.md#configuration it seems that the configuration file should be available under /var/lib/origin/openshift.local.config with oc cluster up. You can use --use-existing-config argument when restarting your cluster to avoid regenerating the configuration."

anjannath:
Fixes #1971

### DIFF
--- a/addons/max-pod/max-pod.addon
+++ b/addons/max-pod/max-pod.addon
@@ -1,0 +1,11 @@
+# Name: max-pod
+# Description: Add-on to update the maximum pods value (defaults to 20)
+# Required-Vars: MAX_PODS
+# Var-Defaults: MAX_PODS=20
+
+ssh docker exec -t origin /usr/bin/openshift ex config patch /var/lib/origin/openshift.local.config/node-localhost/node-config.yaml '--patch={"kubeletArguments": { "max-pods": ["#{MAX_PODS}"] } }' > tmpconfig
+
+ssh cat tmpconfig | sed "s/'/\'/" | sed "s/.$//" | sudo tee /var/lib/minishift/openshift.local.config/node-localhost/node-config.yaml > /dev/null
+ssh rm tmpconfig
+
+docker restart origin
--- a/docs/source/using/addons.adoc
+++ b/docs/source/using/addons.adoc
@@ -255,6 +255,8 @@ By default, this is not allowed in OpenShift.
 | Creates a user named 'admin' and assigns the *cluster-admin* role to it.  
 | registry-route
 | Creates an edge terminated route for the OpenShift registry.
+| max-pod
+| Modify the value of 'max-pods', set the desired value using `--addon-env MAX_PODS=`
 
 | xpaas
 | Imports xPaaS templates.
--- a/cmd/minishift/cmd/util/util.go
+++ b/cmd/minishift/cmd/util/util.go
@@ -33,7 +33,7 @@ import (
 )
 
 var (
-	DefaultAssets = []string{"anyuid", "admin-user", "xpaas", "registry-route"}
+	DefaultAssets = []string{"anyuid", "admin-user", "xpaas", "registry-route", "max-pod"}
 )
 
 func VMExists(client *libmachine.Client, machineName string) bool {
--- a/pkg/minishift/addon/manager/addon_manager_test.go
+++ b/pkg/minishift/addon/manager/addon_manager_test.go
@@ -62,7 +62,7 @@ func Test_create_addon_manager(t *testing.T) {
 
 	addOns := manager.List()
 
-	expectedNumberOfAddOns := 4
+	expectedNumberOfAddOns := 5
 	assert.Len(t, addOns, expectedNumberOfAddOns)
 }
 
--- a/test/integration/features/basic.feature
+++ b/test/integration/features/basic.feature
@@ -6,7 +6,7 @@ Feature: Basic
     When executing "minishift addons install --defaults" succeeds
     Then stdout should contain
      """
-     Default add-ons 'anyuid, admin-user, xpaas, registry-route' installed
+     Default add-ons 'anyuid, admin-user, xpaas, registry-route, max-pod' installed
      """
 
   Scenario: User can enable the anyuid add-on
--- a/test/integration/features/cmd-addons.feature
+++ b/test/integration/features/cmd-addons.feature
@@ -44,7 +44,7 @@ Feature: Addons command and its subcommands
      When executing "minishift addons install --defaults" succeeds
      Then stdout should contain
       """
-      Default add-ons 'anyuid, admin-user, xpaas, registry-route' installed
+      Default add-ons 'anyuid, admin-user, xpaas, registry-route, max-pod' installed
       """
      When executing "minishift addons list" succeeds
      Then stdout should contain "admin-user"


addons/max-pod/max-pod.addon
```diff
@@ -0,0 +1,15 @@
...
+# Required-Vars: MAX_PODS  
+# OpenShift-Version: >3.6.0
+# Var-Defaults: MAX_PODS=20
+
+
+ssh docker exec -t origin /usr/bin/openshift ex config patch /var/lib/origin/openshift.local.config/node-localhost/node-config.yaml '--patch={"kubeletArguments": { "max-pods": ["#{MAX_PODS}"] } }' > tmpconfig
+
+ssh cat tmpconfig | sed "s/'/\'/" | sed "s/.$//" | sudo tee /var/lib/minishift/openshift.local.config/node-localhost/node-config.yaml > /dev/null
+ssh rm tmpconfig
+
```
LalatenduMohanty: @anjannath you can directly use `docker exec` command like https://github.com/minishift/minishift-addons/blob/master/add-ons/cors/cors.addon.

addons/max-pod/max-pod.addon
```diff
@@ -0,0 +1,15 @@
...
+docker stop origin
+docker start origin
```
LalatenduMohanty: docker restart origin

praveenkumar:
And why this is part of this repo instead `minishift-addon` one?

addons/max-pod/max-pod.addon
LalatenduMohanty: Also I think you can remove the intermediate file and directly apply the sed.

LalatenduMohanty:
> And why this is part of this repo instead minishift-addon one?

IMO the issue is part of core Minishift and the solution should be part of core Minishift. Not sure what others think @minishift/minishift-dev .

anjannath:
The reason for it not in the `add-on` repo because, it is not something like an application, its more similar to the add-ons in this repo like the `admin-user` or `anyuid` where we are changing openshift configs. 

That being said, i have a bit of hesitation to make it an addon and want people to use the commad line `minishift openshift config --patch` to patch the node config file. The command line takes care of backing up the previous config file and restores it if anything goes wrong. In the addon we are not doing that :(. And we might end up with a corrupt config file, whic would stop openshift from starting.

addons/max-pod/max-pod.addon
anjannath: Thanks @LalatenduMohanty i will try with that

addons/max-pod/max-pod.addon
```diff
@@ -0,0 +1,8 @@
+# Name: max-pod
+# Description: Add-on to change the value of max-pods
```
budhrg: `Add-on to update the maximum pod value`

anjannath: @budhrg Thanks.. updating the description, WDYT about the name?

budhrg: Name is fine. :+1:

LalatenduMohanty:
@anjannath  The add-on is not part of defaults as `minishift addons install --defaults` does not installs it.

anjannath:
@LalatenduMohanty Yeah, sorry i forgot to add it to defaultAssests, can you please try again, It should install now.

addons/max-pod/max-pod.addon
anjannath: @LalatenduMohanty we need to use `ssh` as the patching process interects both with the container and the host vm.

LalatenduMohanty:
@anjannath there seems to be issue with the default install and some kind of bug in general

```
$ rm -rf ~/.minishift/addons

$ minishift addons list
Error installing default add-ons : Asset max-pod not found

$ minishift addons list
```
`Error installing default add-ons : Asset max-pod not found` is bug and we need to fix it.

```
$ minishift addon install --defaults
Default add-ons 'anyuid, admin-user, xpaas, registry-route, max-pod' installed

$ minishift addons list

```
Still no sign of `max-pod`.

anjannath:
@LalatenduMohanty It should have worked, also the ci didn't fail (made changes to the feature file to have 'max-pod' as default addon).
Did you do `make clean` before building this branch? you need to, as `go-bindata` needs to pack the addon.

budhrg:
> Did you do make clean before building this branch? you need to, as go-bindata needs to pack the addon.

I was also getting similar error as @LalatenduMohanty is getting. 
Doing `make clean` and `make`, I am not getting any error.

addons/max-pod/max-pod.addon
```diff
@@ -0,0 +1,11 @@
+# Name: max-pod
+# Description: Add-on to update the maximum pods value
```
budhrg: Could you add value as well here in the description so that it is available in the verbose view.

budhrg:
Functionally working. Able to see the node's configuration to have max pod
```
$ minishift openshift config view --target node
allowDisabledDocker: false
apiVersion: v1
[...]
imageConfig:
  format: openshift/origin-${component}:v3.7.1
  latest: false
iptablesSyncPeriod: 30s
kind: NodeConfig
kubeletArguments:
  cgroup-driver:
  - cgroupfs
  max-pods:
  - "20"
masterClientConnectionOverrides:
  acceptContentTypes: application/vnd.kubernetes.protobuf,application/json
  burst: 40
  contentType: application/vnd.kubernetes.protobuf
  qps: 20
[...]
```

addons/max-pod/max-pod.addon
anjannath: I could, but the user is supposed to pass that value, the add-on is to allow the user to change the 'max-pods' value, not set it to a fixed 20. maybe we should just remove the `Default-Vars` ? WDYT?

budhrg: Then, updated description as 
```
Add-on to update the maximum pods value (default to 20)
```

budhrg: Also, need documentation update.

docs/source/using/addons.adoc
```diff
@@ -255,6 +255,8 @@ By default, this is not allowed in OpenShift.
 | Creates a user named 'admin' and assigns the *cluster-admin* role to it.  
 | registry-route
 | Creates an edge terminated route for the OpenShift registry.
+| max-pod
```
anjannath: Hi @robin-owen can you please take a look, i am not very satisfied with the description :-P

robin-owen: What's there is true, but doesn't help the reader understand what the add-on does. Something along the lines of "adjusts the maximum number of pods allocated for each CPU core" could work. :)

anjannath: Aha.. now there's a shortcoming in this addon, to set the number of pods/CPU core, 'pods-per-core' value has to be modified. both the properties have an effect on the number of pods a node can run.

ref: https://docs.openshift.com/container-platform/3.5/admin_guide/manage_nodes.html#admin-guide-max-pods-per-node

LalatenduMohanty:
@anjannath  I can not move past 18 pods even of I used 6GB for the VM and max pods was 30. 
 
```
$ minishift start --memory 6GB

$ minishift addon apply  --addon-env MAX_PODS=30 max-pod

$ minishift openshift config view --target node
allowDisabledDocker: false
apiVersion: v1
authConfig:
  authenticationCacheSize: 1000
  authenticationCacheTTL: 5m
  authorizationCacheSize: 1000
  authorizationCacheTTL: 5m
dnsBindAddress: ""
dnsDomain: cluster.local
dnsIP: 172.30.0.1
dnsNameservers: null
dnsRecursiveResolvConf: ""
dockerConfig:
  dockerShimRootDirectory: /var/lib/dockershim
  dockerShimSocket: /var/run/dockershim.sock
  execHandlerName: native
enableUnidling: true
imageConfig:
  format: openshift/origin-${component}:v3.7.1
  latest: false
iptablesSyncPeriod: 30s
kind: NodeConfig
kubeletArguments:
  cgroup-driver:
  - systemd
  max-pods:
  - "30"
```

image


image

addons/max-pod/max-pod.addon
```diff
@@ -0,0 +1,11 @@
+# Name: max-pod
+# Description: Add-on to update the maximum pods value (defaults to 20)
```
budhrg: Looks good now

LalatenduMohanty:
@anjannath I would suggest you to close this PR and once you are sure about the issue please re-open the PR with required changes.

Issue #1969, opened 2018-02-02
Error during post cluster up configuration: Error reading config file ~/.minishift/machines/minishift_kubeconfig


indrayam:
### General information

  * Minishift version: v1.12.0+daa0943
  * OS: macOS
  * Hypervisor: VirtualBox

I am running 5.2.6 r120293 of VirtualBox. I already have minikube, gcloud (with kubectl) and Docker for Mac (Edge Channel) with Kubernetes support installed. I connect to k8s clusters hosted at GKE, locally on minikube and by Docker for Mac. Which is to say, I have an existing ~/.kube/config file.

I wanted to install minishift to try out the Istio Tutorial hosted by RedHat [here](https://github.com/redhat-developer-demos/istio-tutorial)

### Steps to reproduce

  1. Download and installed minishift in `/usr/local/bin`
  2. Run `(prompt)>minishift start --vm-driver=virtualbox`
  3. Things seem to work well until I get the following error:
```
Error during post cluster up configuration: Error reading config file /Users/anasharm/.minishift/machines/minishift_kubeconfig
```

### Expected
I was hoping that the post cluster configuration would work. It seems like I can run most of the commands (like `minishift status` and `oc get projects`) and they all seem to work right, but I saw this open issue - [#1827](https://github.com/minishift/minishift/issues/1827). Sure enough, I ran into the same issue as well...

```
(prompt)>minishift openshift service nodejs-ex --in-browser
Error getting valid namespaces user belongs to.: exit status 1
```

### Actual
Since there is no `minishift_kubeconfig` file, guess the error was expected. The `oc` command did not work. I found the binary inside `~/.minishift/cache/oc` folder and had to create a symlink in `/usr/local/bin`. 

### Logs
Output from `minishift start --show-libmachine-logs -v5` command is at this location: [https://gist.github.com/indrayam/51f31abbbd67314a083e98d846ccc9e0](https://gist.github.com/indrayam/51f31abbbd67314a083e98d846ccc9e0)

praveenkumar:
@indrayam This might happen because the way we parse the ~/.kube/config and extract minishift specific details but in your case looks like we are not parsing it well enough and `minishift_kubeconfig` file become corrupted. Can you please provide how is your ~/.kube/config file looks like (make sure you remove all the private token from it before pasting it)?

indrayam:
Here you go @praveenkumar 
[config.txt](https://github.com/minishift/minishift/files/1691158/config.txt)

[...]

PR #2034, opened 2018-02-15
Issue #1969 Resolve kube config parsing issue for post cluster tasks

praveenkumar:


budhrg:
I see few notable changes in the `minishift_kubeconfig` file generated when using this PR

### Current
```yaml
$ cat ~/.minishift/machines/minishift_kubeconfig 
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: some-data
    server: https://192.168.42.238:8443
  name: 192-168-42-238:8443
contexts:
- context:
    cluster: 192-168-42-238:8443
    namespace: default
    user: system:admin/127-0-0-1:8443
  name: default/192-168-42-238:8443/system:admin
current-context: default/192-168-42-238:8443/system:admin
users:
- name: system:admin/127-0-0-1:8443
  user:
    client-certificate-data: some-data
    client-key-data: some-data
```

### From PR
```yaml
$ cat ~/.minishift/machines/minishift_kubeconfig 
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: some-data
    server: https://192.168.42.127:8443
  name: 192-168-42-127:8443
contexts:
- context:
    cluster: 192-168-42-127:8443
    namespace: myproject                           <--------------------diff
    user: system:admin/192-168-42-127:8443
  name: myproject/192-168-42-127:8443/system:admin      <--------------------diff
current-context: myproject/192-168-42-127:8443/system:admin      <--------------------diff
kind: Config
preferences: {}
users:
- name: system:admin/192-168-42-127:8443
  user:
    client-certificate-data: some-data
    client-key-data: some-data
```

I didn't get with application deployment and servicing application url where it will be tested but doubt it might not work as it is tied to `myproject` namespace.

praveenkumar:
> @praveenkumar , have you tested with app deployement and using service command?

```
$ ./minishift openshift service list
|-----------|-----------------|----------|---------------------------------------------------|--------|
| NAMESPACE |      NAME       | NODEPORT |                     ROUTE-URL                     | WEIGHT |
|-----------|-----------------|----------|---------------------------------------------------|--------|
| default   | router          |          |                                                   |        |
|           | docker-registry |          |                                                   |        |
|           | kubernetes      |          |                                                   |        |
| myproject | cakephp-ex      |          | http://cakephp-ex-myproject.192.168.64.100.nip.io |        |
|-----------|-----------------|----------|---------------------------------------------------|--------|
```

budhrg:
I tested with a different namespace and looks working to me. So `namespace` in the `minishift_kubeconfig` is not playing any role here.

pkg/minikube/kubeconfig/config_test.go
```diff
@@ -1,72 +0,0 @@
-/*
```
budhrg: Since, we are removing unit test to verify kubeconfig entries. 
It would be good to verify it as part of integration test.

praveenkumar: @budhrg as part of integration test it's already verified because this is the step which part of `postclusterup` and if there is any issue with kubeconfig file then start minishift fails. Even with the deploying and app or getting the routes of applications all depend on this file.  I am not sure what else you want to add as part of test.

budhrg: As this time, I am approving as during application deployment, it get tested.

I was thinking of testing the content of minishift_kubeconfig file after start succeed.

praveenkumar: @budhrg Testing minishift_kubeconfig context doesn't make sense because it changes with each VM creation, if you think only to test structure that again not something useful as a test, but as I told if our generated config have any issue then it will fail in every step (be it starting minishift, getting the service/route for service ... etc.)

pkg/minikube/kubeconfig/config.go
```diff
@@ -60,41 +59,20 @@ func GetConfigPath() string {
...
+	// There is another easy way to get config for current context
+	// oc login -u system:admin
+	// oc config view --minify --raw=true
+	// We need to login as system:admin because then only config view will have client-certificate-data
+	// and client-key-data which is associated with admin and all the operation we do as oc runner.
+	var buffer bytes.Buffer
+	cmdArgs := []string{"login", "-u", "system:admin"}
+	runner.Run(nil, os.Stderr, ocPath, cmdArgs...)
+	cmdArgs = []string{"config", "view", "--minify", "--raw=true"}
+	runner.Run(&buffer, os.Stderr, ocPath, cmdArgs...)
```
LalatenduMohanty: @praveenkumar Where did this come from? any pointers?

praveenkumar: @LalatenduMohanty you mean the command? it is in the kubernetes and openshift documentations https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/ and https://docs.openshift.org/latest/cli_reference/manage_cli_profiles.html#cli-reference-manage-cli-profiles

LalatenduMohanty: @praveenkumar Yes, that's what I was looking for.

LalatenduMohanty:
@praveenkumar Thanks.

Issue #1933, opened 2018-01-25
minishift registry-route addon seems to block docker push to minishift registry


cunningt:
### General information

  * Minishift version: 
minishift v1.11.0+4459917, but reproduces on 1.10 as well
  * OS: Linux / macOS / Windows
Mac OS X 10.13.2
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox
xhyve

tcunning@tcunning-OSX:~/src/cunningt/prometheus-asb-installer/fuseprometheus-apb$  docker version
Client:
 Version:	17.12.0-ce
 API version:	1.24 (downgraded from 1.35)
 Go version:	go1.9.2
 Git commit:	c97c6d6
 Built:	Wed Dec 27 20:03:51 2017
 OS/Arch:	darwin/amd64

Server:
 Engine:
  Version:	1.12.6
  API version:	1.24 (minimum version )
  Go version:	go1.8.3
  Git commit:	85d7426/1.12.6
  Built:	Tue Oct 24 15:40:21 2017
  OS/Arch:	linux/amd64
  Experimental:	false

### Steps to reproduce

tcunning@tcunning-OSX:~/src/cunningt/prometheus-asb-installer/fuseprometheus-apb$ minishift addons list
- admin-user     : disabled   P(0)
- anyuid         : disabled   P(0)
- registry-route : disabled   P(0)
- xpaas          : disabled   P(0)

```
tcunning@tcunning-OSX:~/src/cunningt/prometheus-asb-installer/fuseprometheus-apb$ minishift addons enable registry-route
Add-on 'registry-route' enabled
tcunning@tcunning-OSX:~/src/cunningt/prometheus-asb-installer/fuseprometheus-apb$ minishift start --openshift-version=v3.7.0 --iso-url centos --memory 4GB
-- Starting profile 'minishift'
-- Checking if requested hypervisor 'xhyve' is supported on this platform ... OK
-- Checking if xhyve driver is installed ... 
   Driver is available at /usr/local/bin/docker-machine-driver-xhyve
   Checking for setuid bit ... OK
-- Checking the ISO URL ... OK
-- Starting local OpenShift cluster using 'xhyve' hypervisor ...
-- Minishift VM will be configured with ...
   Memory:    4 GB
   vCPUs :    2
   Disk size: 20 GB

   Downloading ISO 'https://github.com/minishift/minishift-centos-iso/releases/download/v1.4.0/minishift-centos7.iso'
[...]

PR #2035, opened 2018-02-16
Issue #1933 Allow docker push from exposed registry-route

praveenkumar:
Openshift >= v3.7.0 we don't need to specify the port number `443` for registry route.

### DIFF
--- a/addons/registry-route/registry-route.addon
+++ b/addons/registry-route/registry-route.addon
@@ -16,4 +16,9 @@ docker cp origin:/var/lib/origin/openshift.local.config/master/ca.crt /etc/docke
 echo Add-on '#{addon-name}' created docker-registry route. Please run following commands to login to the OpenShift docker registry:
 echo $ eval $(minishift docker-env)
 echo $ eval $(minishift oc-env)
+echo
+echo If your deployed version of OpenShift is < 3.7.0 use.
 echo $ docker login -u developer -p `oc whoami -t` docker-registry-default.#{routing-suffix}:443
+echo
+echo If your deployed version of OpenShift is >= 3.7.0 use.
+echo $ docker login -u developer -p `oc whoami -t` docker-registry-default.#{routing-suffix}
--- a/pkg/minishift/openshift/registry.go
+++ b/pkg/minishift/openshift/registry.go
@@ -23,13 +23,14 @@ import (
 
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	instanceState "github.com/minishift/minishift/pkg/minishift/config"
+	openshiftVersionCheck "github.com/minishift/minishift/pkg/minishift/openshift/version"
 )
 
-func GetDockerRegistryInfo(registryAddonEnabled bool) (string, error) {
+func GetDockerRegistryInfo(registryAddonEnabled bool, openshiftVersion string) (string, error) {
 	var registryInfo string
 	var err error
 	if isRegistryRouteEnabled(registryAddonEnabled) {
-		registryInfo, err = fetchRegistryRoute()
+		registryInfo, err = fetchRegistryRoute(openshiftVersion)
 	} else {
 		registryInfo, err = fetchRegistryService()
 	}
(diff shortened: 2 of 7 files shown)


addons/registry-route/registry-route.addon
```diff
@@ -16,4 +16,9 @@ docker cp origin:/var/lib/origin/openshift.local.config/master/ca.crt /etc/docke
 echo Add-on '#{addon-name}' created docker-registry route. Please run following commands to login to the OpenShift docker registry:
 echo $ eval $(minishift docker-env)
 echo $ eval $(minishift oc-env)
+echo
+echo If your deployed version of OpenShift is < 3.7.0 use.
 echo $ docker login -u developer -p `oc whoami -t` docker-registry-default.#{routing-suffix}:443
+echo
+echo If your deployed version of OpenShift is >= 3.7.0 use.
+echo $ docker login -u developer -p `oc whoami -t` docker-registry-default.#{routing-suffix}
```
LalatenduMohanty: +1 works for me.

pkg/minishift/openshift/registry.go
```diff
@@ -51,10 +52,17 @@ func isRegistryRouteEnabled(registryAddonEnabled bool) bool {
 	return true
 }
 
-func fetchRegistryRoute() (string, error) {
+func fetchRegistryRoute(openshiftVersion string) (string, error) {
 	namespace := "default"
 	route := "route/docker-registry"
 	cmdArgText := fmt.Sprintf("get -o jsonpath={.spec.host}:443 %s -n %s --config=%s", route, namespace, constants.KubeConfigPath)
+	valid, err := openshiftVersionCheck.IsGreaterOrEqualToBaseVersion(openshiftVersion, "v3.7.0")
```
LalatenduMohanty: @praveenkumar Can you add a comment saying what changed in v3.7.0? Also should n't we use a variable instead of directly using hard coded string "v3.7.0"?

LalatenduMohanty: I mean `constants.BackwardIncompatibleOcVersion`.

LalatenduMohanty:
@praveenkumar Looks like the PR is not working for versions < 3.7.0.
```
$ minishift addon enable registry-route
Add-on 'registry-route' enabled

$ minishift start --openshift-version v3.6.1
```
But `minishift openshift registry` returning `docker-registry-default.192.168.64.11.nip.io` where as I was expecting `docker-registry-default.192.168.64.11.nip.io:443` as the version is less than v3.7.0

```
$ minishift openshift registry
docker-registry-default.192.168.64.11.nip.io
```
To cross check I tested master branch it is returning correct value. 
```
$ git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.

$ make

$ minishift openshift registry
docker-registry-default.192.168.64.11.nip.io:443
```

praveenkumar:
> Also it seems there is no integration test with registry-route add-on to check output of minishift openshift registry. Please add one and you can test for versions greater and less than 3.7.0.

@LalatenduMohanty @amitkrout is creating integration test around registry addon #1318 and that's another reason we found this issue.

amitkrout:
> Also it seems there is no integration test with registry-route add-on to check output of minishift openshift registry. Please add one and you can test for versions greater and less than 3.7.0.

@LalatenduMohanty This usecase will be added to #1318 for sure. Once this pr is merged i will start work on the test integration side. Probably in the beginning of next sprint will have the pr ready.

praveenkumar:
retest this please.

praveenkumar:
Created #2060 for doc decision around it.

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -55,6 +56,25 @@ func GetOpenshiftVersion(host *host.Host) (string, error) {
 	return dockerCommander.Exec(" ", minishiftConstants.OpenshiftContainerName, "openshift", "version")
 }
 
+func GetOpenshiftVersionWithoutK8sAndEtcd(sshCommander provision.SSHCommander) (string, error) {
+	dockerCommander := docker.NewVmDockerCommander(sshCommander)
+	versionInfo, err := dockerCommander.Exec(" ", minishiftConstants.OpenshiftContainerName, "openshift", "version")
+	if err != nil {
+		return "", err
+	}
```
LalatenduMohanty: @praveenkumar I do not think we should run `dockerCommander.Exec(" ", minishiftConstants.OpenshiftContainerName, "openshift", "version") ` in two functions, when the command is exactly the same. I would suggest we reuse GetOpenshiftVersion()  then parse the output to just get the OpenShift version information here. 

Also it might be a good idea to rename `GetOpenshiftVersion` to `GetOpenshiftClusterVersion` and `GetOpenshiftVersionWithoutK8sAndEtcd` to `GetOpenshiftVersion`

gbraad:
> Also it might be a good idea to rename GetOpenshiftVersion to
GetOpenshiftClusterVersion and GetOpenshiftVersionWithoutK8sAndEtcd to
GetOpenshiftVersion

Sounds bad to reuse a previously used name.

-- 

   [ Doing Open Source Matters ]

addons/registry-route/registry-route.addon
anjannath: This information will be only shown when the add-on is applied, but for this particular addon we need to enable it and then start minishift.

pkg/minishift/openshift/version/openshift_versions.go
praveenkumar: @LalatenduMohanty If you see the signature of both function then it different. If you think we should call `GetOpenshiftVersion` from `GetOpenshiftVersionWithoutK8sAndEtcd` this is not possible atm. 

```
GetOpenshiftVersion(host *host.Host) (string, error)
GetOpenshiftVersionWithoutK8sAndEtcd(sshCommander provision.SSHCommander) (string, error)
```

@LalatenduMohanty @gbraad I don't think to execute a single command in both the function have any issue. I am also netural about naming part, let me know if you think we should change then will change it.

addons/registry-route/registry-route.addon
LalatenduMohanty: @anjannath We will update the documentation about it in #2060

praveenkumar: @anjannath This information should be displayed if a user try to apply or enable it. For this add-on also you can use it as apply.

Issue #2039, opened 2018-02-19
Image caching is not disabled for profile foo in cmd-profile.feature file


amitkrout:
### General information

  * Minishift version: 
  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce

  1. 
  2. 
  3. 
  4. 

### Expected
According to issue #1799 fix it is fine to disable image caching before minishift start due to ci time constraint except cmd-image feature as image caching test is fully implemented in ```cmd-image.feature```. We know in cmd-profile feature two VMs are being started. The problem here is before staring the minishift vm in cmd-profile image caching is disabled which is fine however missed to disable caching before foo profile start.

### Actual
[cmd-profile.feature](https://github.com/minishift/minishift/blob/master/test/integration/features/cmd-profile.feature#L86)

### Logs

You can start Minishift with `minishift start --show-libmachine-logs -v5` to collect logs.
Please consider posting this on http://gist.github.com/ and post the link in the issue.

Issue #1889, opened 2018-01-05
Update profile tests


budhrg:
Update profile tests to work for both upstream and downstream. 

See https://issues.jboss.org/browse/CDK-234.

```
# In profile.feature
{code}
Scenario: As user, I can apply independent settings in profile 'foo'
[...]    
Then stdout should match "<nil>"                                          # integration_test.go:677 -> commandReturnShouldMatchRegex
    Output did not match. Expected: '<nil>', Actual: '4096
```
And
```
Scenario: As user, I can execute a command against a non active profile
[...]    
And stdout should contain                                                                   # integration_test.go:649 -> commandReturnShouldContainContent
      """
      Stopping local OpenShift cluster...
      Cluster stopped.
      """
    Output did not match. Expected: 'Stopping local OpenShift cluster...
Cluster stopped.', Actual: 'Stopping local OpenShift cluster...
Unregistering machine
Cluster stopped.
'
```

Here, we found in `profile.feature`, two test cases are related to minishift only. 
We should search other test cases and tag them `@minishift-only` to avoid failure in downstream.

agajdosi:
We should however focus on tagging `@minishift-only` only in inevitable cases, meaning that some longer scenarios has to be split or edited.

LalatenduMohanty:
we should tag them minishift only once we have these tests modified for CDK and merged in CDK  in separate files to avoid merge conflicts.

PR #2044, opened 2018-02-19
 Issue #1889 #2039 Update profile test to disable image caching 

amitkrout:
Fix : #1889 #2039 

I added tag @minishift-tag to the feature files. Some feature files can be restructured without loosing the usecase integrity to run on both upstream and downstream use. Files like ```cmd-image, provision-earlier-version``` are only for upstream integration test as this is very difficult to restructure and keep the usecase integrity intact for both testing.

So this pr is dealing with these feature files, and as a solution to the downstream it avoid merge conflict with same downstream feature file. In the downstream repo same file should be created with name cdk-*.feature and @cdk-only tag. For example 

cdk-cmd-image.feature
```
@cmd-image @command @cdk-only
Feature: Basic image caching test
  As a user I am able to import and export container images from a local OCI repository
  located in the $MINISHIFT_HOME/cache directory

  //Modify the scenarios as per downstream use

[...]
```

For those specific scenario which are tagged @minishift-only in feature file should be kept in separate feature file like ```cdk-specific.feature``` and is modified accordingly for downstream only. @agajdosi i can see a long pending pr [1] in downstream which almost resolves the same. Would you please delete and resend it.

This is how we can avoid merge conflict for our downstream sync job.
[1] https://gitlab.cee.redhat.com/minishift/minishift/merge_requests/295

### DIFF
--- a/test/integration/features/cmd-profile.feature
+++ b/test/integration/features/cmd-profile.feature
@@ -77,28 +77,26 @@ Feature: Profile commands
   Scenario: As user, I can apply independent settings in profile 'foo'
      When executing "minishift profile set foo" succeeds
      Then profile "foo" should be the active profile
-     When executing "minishift --profile minishift config get memory" succeeds
-     Then stdout should match "<nil>"
+      And executing "minishift --profile minishift config set cpus 4" succeeds
       And executing "minishift --profile minishift addons list" succeeds
       And stdout should match "registry-route\s*: disabled\s*P\(0\)"
       And executing "minishift addons enable registry-route" succeeds
       And executing "minishift config set memory 5120" succeeds
       And executing "minishift config get memory" succeeds
       And stdout should match "5120"
+      And image caching is disabled
      When executing "minishift start" succeeds
-     Then stdout should contain
-      """
-      Add-on 'registry-route' created docker-registry route.
-      """
+     Then stdout should match "vCPUs\s*:\s*2"
       And profile "foo" should have state "Running"
-     When executing "minishift --profile minishift config get memory" succeeds
-     Then stdout should match "<nil>"
+     When executing "minishift --profile minishift config get cpus" succeeds
+     Then stdout should match "4"
       And executing "minishift --profile minishift addons list" succeeds
       And stdout should match "registry-route\s*: disabled\s*P\(0\)"
       And executing "minishift addons list" succeeds
       And stdout should match "registry-route\s*: enabled\s*P\(0\)"
-      And executing "minishift ssh -- cat /proc/meminfo" succeeds
-      And stdout should match "MemTotal:\s*5[0-1][0-9]{5}\s*kB"
+      And executing "minishift ssh -- cat /proc/cpuinfo" succeeds
+      And stdout should match "processor\s*: [0-3]"
+      And stdout should not match "processor\s*: [4-9]"
 
   Scenario: As user, I can execute a command against a non active profile
     Given profile "foo" is the active profile
@@ -107,7 +105,6 @@ Feature: Profile commands
       And executing "minishift --profile minishift stop" succeeds
       And stdout should contain
        """
-       Stopping local OpenShift cluster...
        Cluster stopped.
        """
      When executing "minishift --profile minishift status" succeeds


amitkrout:
@agajdosi @praveenkumar I am moving the tag [WIP]. Please review it

test/integration/features/cmd-profile.feature
```diff
@@ -75,28 +75,26 @@ Feature: Profile commands
...
-     When executing "minishift --profile minishift config get memory" succeeds
-     Then stdout should match "<nil>"
+      And executing "minishift --profile minishift config set cpus 4" succeeds
       And executing "minishift --profile minishift addons list" succeeds
       And stdout should match "registry-route\s*: disabled\s*P\(0\)"
       And executing "minishift addons enable registry-route" succeeds
       And executing "minishift config set memory 5120" succeeds
       And executing "minishift config get memory" succeeds
       And stdout should match "5120"
+      And image caching is disabled
```
amitkrout: Fix of issue #2039

test/integration/features/cmd-profile.feature
```diff
@@ -75,28 +75,26 @@ Feature: Profile commands
...
      When executing "minishift start" succeeds
-     Then stdout should contain
-      """
-      Add-on 'registry-route' created docker-registry route.
-      """
+     Then stdout should match "vCPUs\s*:\s*2"
```
praveenkumar: @amitkrout how come vCPUs is 2 here instead 4?

test/integration/features/cmd-profile.feature
```diff
@@ -75,28 +75,26 @@ Feature: Profile commands
...
+     Then stdout should match "4"
       And executing "minishift --profile minishift addons list" succeeds
       And stdout should match "registry-route\s*: disabled\s*P\(0\)"
       And executing "minishift addons list" succeeds
       And stdout should match "registry-route\s*: enabled\s*P\(0\)"
-      And executing "minishift ssh -- cat /proc/meminfo" succeeds
-      And stdout should match "MemTotal:\s*5[0-1][0-9]{5}\s*kB"
+      And executing "minishift ssh -- cat /proc/cpuinfo" succeeds
+      And stdout should match "processor\s*: [0-3]"
+      And stdout should not match "processor\s*: [4-9]"
```
praveenkumar: I don't get why we are removing memory part and adding CPU part for this test.

test/integration/features/cmd-profile.feature
amitkrout: I am not adding the any CPUs to foo profile. I added the CPUs to default minishift profile, so it wont reflect in profile foo

test/integration/features/cmd-profile.feature
amitkrout: I am doing this change keeping https://issues.jboss.org/projects/CDK/issues/CDK-253 in my mind. With this change i can minimize the tagging effort i mean same feature file will run for both upstream and downstream

test/integration/features/cmd-image.feature
```diff
@@ -1,4 +1,4 @@
-@cmd-image @command
+@cmd-image @command @minishift-only
```
LalatenduMohanty: @amitkrout I am not sure why is this should be only minishift specific test?

LalatenduMohanty:
@amitkrout Can you please update your commit message to include #2039 as you are also fixing it?

test/integration/features/cmd-image.feature
amitkrout: @LalatenduMohanty My understanding here is the major scenarios are not for downstream i.e. image exporting and importing which is the key feature of image caching. There is no point keeping the rest scenarios without this and it is very difficult to split such major scenario for tagging. So i tagged whole file minishift specific.

This file will be skipped if i run this file with downstream binary which is a plus IMO rather throwing error. And its good enough to modify the same file for downstream use.

amitkrout:
> Can you please update your commit message to include #2039 as you are also fixing it?

Done

test/integration/features/cmd-image.feature
budhrg: @amitkrout @LalatenduMohanty +1 for modifying it for downstream use. Mostly below scenario
```
     Then stdout should match "Importing 'openshift\/origin:v[0-9]+\.[0-9]+\.[0-9]+' [\.]+ OK"
      And stdout should match "Importing 'openshift\/origin-docker-registry:v[0-9]+\.[0-9]+\.[0-9]+' [\.]+ OK"
      And stdout should match "Importing 'openshift\/origin-haproxy-router:v[0-9]+\.[0-9]+\.[0-9]+' [\.]+ OK"
```

test/integration/features/coolstore.feature
```diff
@@ -6,14 +6,17 @@ Feature: Cool Store
   After which I will import image streams
   Finally I will deploy the Cool Store
   
-  Scenario: User enables the 'xpaas' add-on and creates a Minishift instance
+  @minishift-only
```
agajdosi: Tag again is not truly needed. However +1 for splitting up the scenario.

test/integration/features/addon-xpaas.feature
```diff
@@ -3,6 +3,7 @@ Feature: xpaas add-on
 Xpaas add-on imports xPaaS templates and imagestreams,
 which are then available in OpenShift to the user.
 
+  @minishift-only
```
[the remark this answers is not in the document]
amitkrout: True. I just put for a check for upstream as by default it is disabled, so enable it. Enabling the enabled addon reports the same output. For time being I think its not required.

test/integration/features/coolstore.feature
amitkrout: tag removed

test/integration/features/cmd-image.feature
LalatenduMohanty: @amitkrout I am not favor of marking this with minishift-only tag. In downstream we should modify the tests and if the rebase causes more issues, we can mark explicit tests (which fails for downstream) and add cdk-only scenarios in future.

test/integration/features/cmd-profile.feature
amitkrout: @praveenkumar Keeping this change won't change the test scenario ```(same validation but with different config values)```, even this would reduce your merge effort

Is this fine for you ? otherwise i will revert back to original state

test/integration/features/coolstore.feature
amitkrout: IMO this is not the right pr for splitting the scenario. Will create an issue unless there is a hard requirement of splitting. As of now let the original file be as it is.

praveenkumar:
@amitkrout is it still in WIP? if not then please remove that.

test/integration/features/cmd-profile.feature
praveenkumar: @amitkrout ack, It's fine.

amitkrout:
@LalatenduMohanty All review comments addressed. Now its ready to merge

test/integration/features/provision_various_versions.feature
```diff
@@ -1,4 +1,4 @@
-@provision-various-versions
+@provision-various-versions @openshift @minishift-only
```
amitkrout: @LalatenduMohanty Does this tag ```@minishift-only``` is worth having to keep here until we manage the merge conflict volume ?
The same [question](https://github.com/minishift/minishift/pull/2061#discussion_r170177659) is asked by @budhrg in this pr #2061.

LalatenduMohanty: @amitkrout The question of whether we should put the `minishift-only` tag or not, depends on what is the best way to handle changes required for downstream. It seems the easiest path forward is modify the same file for downstream. Though it might cause merge conflicts in long term, but unless we have a lot of merge conflicts it will work. 

So the answer is , we do not need the `minishift-only` tag here.

amitkrout:
@LalatenduMohanty I have removed WIP tag. This is kind of a almost dead pr where we have the fix but not satisfying what it was created for. 

Work done in this pr:
Fix : #2039
Modified the feature file that will run on both upstream and downstream (Intention was to optimize ```minishift-only``` tagging). I want to keep this change as it reduces some merge effort.

amitkrout:
@LalatenduMohanty Done

LalatenduMohanty:
Thanks @amitkrout

Issue #2054, opened 2018-02-21
Total memory re fails to validate applied memory in cmd-config feature


agajdosi:
Upstream issue for CDK JIRA https://issues.jboss.org/browse/CDK-256. The cmd-config.feature fails on Windows 10 slave with Hyper-V due to memory regex check:

```
$make integration MINISHIFT_BINARY=/c/cdk3.4alpha1/minishift GODOG_OPTS="-tags cmd-config"
--
[...]
Scenario: Checking that memory value was applied               # features\a.feature:214
When executing "minishift ssh -- cat /proc/meminfo" succeeds # testsuite.go:587 -> github.com/minishift/minishift/test/integration/testsuite.executingMinishiftCommandSucceedsOrFails
Then stdout should match "MemTotal:\s*3[4-5][0-9]{5} kB"     # testsuite.go:565 -> github.com/minishift/minishift/test/integration/testsuite.commandReturnShouldMatchRegex
Output did not match. Expected: 'MemTotal:\s*3[4-5][0-9]{5} kB', Actual: 'MemTotal:        3765156 kB
MemFree:         1216932 kB
MemAvailable:    2705220 kB
Buffers:           30672 kB
[...]
```

LalatenduMohanty:
@agajdosi So it only fails in Windows 10, right? Wondering why we catch the issue now and why not in the past?

agajdosi:
@LalatenduMohanty Interestingly the VM starts with `--memory 3500` so I am curious how in the JIRA we got 3700000kb.

LalatenduMohanty:
Resolved via #2055

PR #2055, opened 2018-02-21
Issue #2054 Total memory regex fails in cmd-config.feature

agajdosi:


agajdosi:
Hi @amitkrout, please validate this fix on the machine on which have faced the issue.

test/integration/features/cmd-config.feature
```diff
@@ -213,7 +213,7 @@ user defined options which changes default behaviour of Minishift.
 
   Scenario: Checking that memory value was applied
      When executing "minishift ssh -- cat /proc/meminfo" succeeds
-     Then stdout should match "MemTotal:\s*3[4-5][0-9]{5} kB"
+     Then stdout should match "MemTotal:\s*3[4-7][0-9]{5} kB"
```
budhrg: This can be `[4-9]` right? How can we sure `[4-7]` is correct here?

budhrg: Now we have agreed on `MemTotal:\s*3[0-9]{6} kB`

test/integration/features/cmd-config.feature
```diff
@@ -213,7 +213,7 @@ user defined options which changes default behaviour of Minishift.
 
   Scenario: Checking that memory value was applied
      When executing "minishift ssh -- cat /proc/meminfo" succeeds
-     Then stdout should match "MemTotal:\s*3[4-5][0-9]{5} kB"
+     Then stdout should match "MemTotal:\s*3[4-9][0-9]{5} kB"
```
budhrg: @agajdosi , I think we agreed on `Then stdout should match "MemTotal:\s*3[0-9]{6} kB"`.

budhrg: As per comment here https://issues.jboss.org/browse/CDK-256?focusedCommentId=13537041&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-13537041, `MemTotal` can start with 3 or reach to 5.
 
@amitkrout @agajdosi , I think we should not worry about `3` or `4` or `5` here. It is hypervisor behavior. 

Let's go with `Then stdout should match "MemTotal:\s*\d[0-9]{6} kB"`

amitkrout: @budhrg Right, We have no control over issue which is coming from hypervisor side. I am ok with this patch

agajdosi: @budhrg @amitkrout then I think we need to cover also a situation with 4[0-1]**** kbs as this sometimes happens on Windows 10.

> The pass rate is 60% with the patch (failed 2 out of 5 run). In one of the test run i can see memory increase to 41* on applied 3500mb. Seems the hypervisor is responsible for inconsistent in "MemTotal" output on same applied memory in config file.

agajdosi:
Hi @LalatenduMohanty @amitkrout, expected value for 3500mb, should be around `3.584.000kb`, I have modified the regexp to match interval of 3.300.000 to 3.699.999 kb. Bigger deviation we should not tolerate IMHO. For the Hyper-V bug I think we can document it and ignore it in the nightly tests once the hyper-V slave is ready again.

Other take might be to disable the test completely. wdyt?

gbraad:
I think you could test for a value that is at least bigger than a certain
threshold. This feels less likely to fail... As we are not test h the
supervisor drivers,  it just if the setting has been applied and worked in
some way

-- 

   [ Doing Open Source Matters ]

amitkrout:
Have some observation with these config set memory values ```4096``` and ```4GB```. This is what i am seeing the respective value applied in ```/proc/meminfo```

```
Config set memory : 4096        Applied memory : 4299584 kB
Config set memory : 4GB         Applied memory : 4029248 kB
```
Seems hyperv considerable increasing the the memory value are passed in mbs with respect to the same memory in GBs.  Instead of passing memory ```3500```, set the memory with 3GB probably solve the problem IMO.

agajdosi:
@gbraad I have now checked on the Win 10 (using ) by myself and it seems, that in some cases the value is taken properly - the hyperV shows 3500mb and /proc/meminfo shows 3.400.000 kb, however in some cases the HyperV shows 4096mb and /proc/meminfo shows 4.060.000 kb.

I did just few runs (had to borrow the machine to Budh then) but I realized that the fialing starts are probably those, which are run directly after setup-cdk: `setup-cdk; config set memory 3500; start` fails, while after that when I delete the machine and start again, the proper values are applied.

Will investigate more.

test/integration/features/cmd-config.feature
```diff
@@ -211,10 +211,6 @@ user defined options which changes default behaviour of Minishift.
      Then stdout should match "vCPUs\s*:\s*3"
       And Minishift should have state "Running"
 
-  Scenario: Checking that memory value was applied
-     When executing "minishift ssh -- cat /proc/meminfo" succeeds
-     Then stdout should match "MemTotal:\s*3[4-5][0-9]{5} kB"
-
```
praveenkumar: As per yesterday discussion, this looks good 👍

amitkrout:
No other reference found for ```minishift ssh -- cat /proc/meminfo``` in remaining feature file. So LGTM :+1:

Issue #2057, opened 2018-02-21
Mention the commit message format in `CONTRIBUTING.adoc`


anjannath:
First time contributors always mess up the commit message and we have to go back and forth on correcting it, which is not a _very_ big deal, but as a first step, we should add this information to `CONTRIBUTING.md` which would make it more visible.

PR #2062, opened 2018-02-23
Issue #2057 Update `CONTRIBUTING.adoc` to contain the format for Commit message

anjannath:
Fixes: #2057

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -18,7 +18,8 @@ Thank you for your contributions! Please follow this process to submit a patch:
 . Fork the repository and create a topic branch.
 . Refer to the link:https://docs.openshift.org/latest/minishift/contributing/developing.html[developer documentation] for guidelines and tips on how to build and test Minishift.
 . Submit a pull request with the proposed changes.
-  To make reviewing and integration as smooth as possible, please run the `make` targets link:https://docs.openshift.org/latest/minishift/contributing/developing.html#unit-tests[`test`], https://docs.openshift.org/latest/minishift/contributing/developing.html#format-source-code[`fmtcheck`] and https://docs.openshift.org/latest/minishift/contributing/developing.html#commit-messages[`validate_commits`] prior to submitting the pull request.
+.. The required format for the git commit message is 'Issue #<issue_no> message'.
+.. To make reviewing and integration as smooth as possible, please run the `make` targets link:https://docs.openshift.org/latest/minishift/contributing/developing.html#unit-tests[`test`], https://docs.openshift.org/latest/minishift/contributing/developing.html#format-source-code[`fmtcheck`] and https://docs.openshift.org/latest/minishift/contributing/developing.html#commit-messages[`validate_commits`] prior to submitting the pull request.
 
 [[questions]]
 == Questions?


CONTRIBUTING.adoc
```diff
@@ -18,7 +18,7 @@ Thank you for your contributions! Please follow this process to submit a patch:
...
 . Submit a pull request with the proposed changes.
-  To make reviewing and integration as smooth as possible, please run the `make` targets link:https://docs.openshift.org/latest/minishift/contributing/developing.html#unit-tests[`test`], https://docs.openshift.org/latest/minishift/contributing/developing.html#format-source-code[`fmtcheck`] and https://docs.openshift.org/latest/minishift/contributing/developing.html#commit-messages[`validate_commits`] prior to submitting the pull request.
+  Format for Commit message: *Issue #<issue_no> message*. The regex `^(Issue #[0-9]+ .*|cut v[0-9]+\.[0-9]+\.[0-9]+)$$` is used to validate commit messages. To make reviewing and integration as smooth as possible, please run the `make` targets link:https://docs.openshift.org/latest/minishift/contributing/developing.html#unit-tests[`test`], https://docs.openshift.org/latest/minishift/contributing/developing.html#format-source-code[`fmtcheck`] and https://docs.openshift.org/latest/minishift/contributing/developing.html#commit-messages[`validate_commits`] prior to submitting the pull request.
```
anjannath: Is it OK to put the regex here? its for contributors, they would know exactly what pattern they have to match.

budhrg: Agree. Regex is not required here

LalatenduMohanty: @anjannath In the current changes are not acceptable as makes the bullet point almost like a paragraph. Please break it in to two bullet points under `Submit a pull request with the proposed changes under `Submit a pull request with the proposed changes`. 
e.g. 
* Submit a pull request with the proposed changes
  * The required format for the git commit message should be *Issue #<issue_no> commit message*
  *  To make reviewing and integration as smooth as possible, please run the make targets test, fmtcheck and validate_commits prior to submitting the pull request.

anjannath: Okay @LalatenduMohanty

CONTRIBUTING.adoc
```diff
@@ -18,7 +18,8 @@ Thank you for your contributions! Please follow this process to submit a patch:
 . Fork the repository and create a topic branch.
 . Refer to the link:https://docs.openshift.org/latest/minishift/contributing/developing.html[developer documentation] for guidelines and tips on how to build and test Minishift.
 . Submit a pull request with the proposed changes.
-  To make reviewing and integration as smooth as possible, please run the `make` targets link:https://docs.openshift.org/latest/minishift/contributing/developing.html#unit-tests[`test`], https://docs.openshift.org/latest/minishift/contributing/developing.html#format-source-code[`fmtcheck`] and https://docs.openshift.org/latest/minishift/contributing/developing.html#commit-messages[`validate_commits`] prior to submitting the pull request.
+.. The required format for the git commit message is *Issue #<issue_no> message*.
```
LalatenduMohanty: @anjannath We do not need to make `Issue #<issue_no> message` bold. As it carries the same importance as other guidelines.

anjannath: My understanding was we need to differentiate it as the format for the commit message and not part of the running sentance. Making it italics/bold will give it that differance.

LalatenduMohanty: @anjannath Then it would be a competition between what needs to be bold. At this point there is only the bold format but if there are several bold strings in the document in future then the overall readability will impacted. So I would suggest to keep it simple.

LalatenduMohanty: @robin-owen WDYT?

robin-owen: @anjannath I agree with @LalatenduMohanty here, although I would add that you should encapsulate the information in quotation marks so that it is clear what, specifically, the commit message requirements are. :)

anjannath: Thanks @robin-owen, I'll put the commit message format in quotation marks.

PR #2075, opened 2018-03-01
Issue #2074 Refactor naming around openshift checks

### LINKED ISSUE #2074: Refactor naming around pre-flight check for oc tag on github
This is mostly required for https://issues.jboss.org/browse/CDK-259 where we will generalize the names like 
- `checkOcTagOnGithub` to `checkOcReleaseVersion`
- `SkipCheckOpenShiftGithubRelease` to `SkipCheckOpenShiftRelease`
- `skip-check-openshift-github-release` to `skip-check-openshift-release`

/cc @LalatenduMohanty

budhrg:
Fix #2074 

In downstream, we will just modify `checkOpenShiftRelease` to be based on openshift mirror url.


budhrg:
@LalatenduMohanty Please review it as it is critical for downstream.
/cc @praveenkumar

cmd/minishift/cmd/start_preflight.go
```diff
@@ -534,8 +534,8 @@ func validateOpenshiftVersion() bool {
 	return true
 }
 
-// checkOcTagOnGithub return true if specified version of OpenShift released.
-func checkOcTagOnGithub() bool {
+// checkOpenShiftRelease return true if specified version of OpenShift released
+func checkOpenShiftRelease() bool {
```
praveenkumar: @budhrg @LalatenduMohanty  What are you planning to do here? Do you folks going to overwrite this function with downstream bits?

budhrg: Yes @praveenkumar.

Or create a new function downstream and use it instead of modify above. 

I think the latter approach would be nice since we would have one line of modification which might have potential merge conflict.

praveenkumar: @budhrg Right that what I suggest to have a new function downstream instead modify above and call that in the `preflightCheckSucceedsOrFails` then I don't think modifying the name of this function is required.

budhrg: Sure. Make sense

cmd/minishift/cmd/config/config.go
```diff
@@ -101,22 +101,22 @@ var (
...
+	SkipCheckIsoUrl           = createConfigSetting("skip-check-iso-url", SetBool, nil, nil, true, nil)
+	WarnCheckIsoUrl           = createConfigSetting("warn-check-iso-url", SetBool, nil, nil, true, false)
+	SkipCheckVMDriver         = createConfigSetting("skip-check-vm-driver", SetBool, nil, nil, true, nil)
+	WarnCheckVMDriver         = createConfigSetting("warn-check-vm-driver", SetBool, nil, nil, true, false)
+	SkipCheckVBoxInstalled    = createConfigSetting("skip-check-vbox-installed", SetBool, nil, nil, true, nil)
+	WarnCheckVBoxInstalled    = createConfigSetting("warn-check-vbox-installed", SetBool, nil, nil, true, false)
+	SkipCheckOpenShiftVersion = createConfigSetting("skip-check-openshift-version", SetBool, nil, nil, true, nil)
+	WarnCheckOpenShiftVersion = createConfigSetting("warn-check-openshift-version", SetBool, nil, nil, true, false)
+	SkipCheckOpenShiftRelease = createConfigSetting("skip-check-openshift-release", SetBool, nil, nil, true, nil)
+	WarnCheckOpenShiftRelease = createConfigSetting("warn-check-openshift-release", SetBool, nil, nil, true, false)
```
LalatenduMohanty: This is better than `WarnCheckOpenShiftGithubReleas` or `SkipCheckOpenShiftGithubRelease`. @praveenkumar  WDYT? `WarnCheckOpenShiftRelease` or `SkipCheckOpenShiftRelease` can used for downstream too.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -534,8 +534,8 @@ func validateOpenshiftVersion() bool {
...
 	client := github.Client()
 	_, _, err := client.Repositories.GetReleaseByTag("openshift", "origin", viper.GetString(configCmd.OpenshiftVersion.Name))
```
LalatenduMohanty: @praveenkumar  any specific reason we did not use `GetRelease` in place of  `GetReleaseByTag`. Techincally we need to check the release not the tag in Origin github repository?

cmd/minishift/cmd/start_preflight.go
LalatenduMohanty: Having a separate function for OCP oc binary makes sense. However I think the function name should be `checkOriginReleaseTag`. Also we should be checking for releases not tags in Origin repository so the function name should be `checkOriginRelease` and the downstream equivalent should be `checkOcpRelease`

budhrg: > Also we should be checking for releases not tags in Origin repository so the function name should be checkOriginRelease and the downstream equivalent should be checkOcpRelease

@LalatenduMohanty this is not the scope of the issue. We can address it in follow up issue.

LalatenduMohanty: @budhrg  It is not much work actually. But if you do not want to take it up in this issue thats fine too.

cmd/minishift/cmd/config/config.go
praveenkumar: 👍

cmd/minishift/cmd/start_preflight.go
```diff
@@ -51,11 +51,11 @@ func preflightChecksBeforeStartingHost() {
 	prerequisiteErrorMessage := "See the 'Installing Prerequisites for Minishift' topic (https://docs.openshift.org/latest/minishift/getting-started/installing.html#install-prerequisites) for more information"
 
 	preflightCheckSucceedsOrFails(
-		configCmd.SkipCheckOpenShiftGithubRelease.Name,
-		checkOcTagOnGithub,
-		fmt.Sprintf("Checking if requested OpenShift version '%s' is valid on GitHub", viper.GetString(configCmd.OpenshiftVersion.Name)),
-		configCmd.WarnCheckOpenShiftGithubRelease.Name,
-		fmt.Sprintf("%s is not a valid OpenShift version on GitHub", viper.GetString(configCmd.OpenshiftVersion.Name)),
+		configCmd.SkipCheckOpenShiftRelease.Name,
+		checkOriginRelease,
```
praveenkumar: 👍

cmd/minishift/cmd/start_preflight.go
```diff
@@ -534,8 +534,8 @@ func validateOpenshiftVersion() bool {
 	return true
 }
 
-// checkOcTagOnGithub return true if specified version of OpenShift released.
-func checkOcTagOnGithub() bool {
+// checkOriginRelease return true if specified version of OpenShift is released
+func checkOriginRelease() bool {
```
praveenkumar: 👍

PR #2078, opened 2018-03-03
Issue #2056 Check writable permission before downloading during update

### LINKED ISSUE #2056: 'minishift update' should check write permissions before downloading
### General information

  * Minishift version: 1.12.0
  * OS: Linux
  * Hypervisor: KVM

### Steps to reproduce

  1. Run `minishift update`

### Expected
`minishift update` should check if it has write permissions to the target directory for installation before downloading the latest release, not after. Once this fails, I need to run `sudo minishift update` again which again downloads the binary. This is a bit painful because downloads from GitHub are generally slow.

### Actual
```console
$ minishift update
Do you want to update from 1.12.0 to 1.13.1 now? [y/N]: y
Downloading https://github.com/minishift/minishift/releases/download/v1.13.1/minishift-1.13.1-linux-amd64.tgz
 5.19 MiB / 5.19 MiB [=============================================================================================] 100.00% 0s
 65 B / 65 B [=====================================================================================================] 100.00% 0s
Update failed: open /usr/local/bin/.minishift.new: permission denied
```

budhrg:
Fix #2056 

Need to test in mac and Windows.

### DIFF
--- a/pkg/minishift/update/update.go
+++ b/pkg/minishift/update/update.go
@@ -36,6 +36,8 @@ import (
 	"github.com/minishift/minishift/pkg/util/archive"
 	pb "gopkg.in/cheggaaa/pb.v1"
 
+	"github.com/kardianos/osext"
+	"github.com/minishift/minishift/pkg/util"
 	githubutils "github.com/minishift/minishift/pkg/util/github"
 	"github.com/minishift/minishift/pkg/version"
 )
@@ -77,6 +79,11 @@ func IsNewerVersion(localVersion, latestVersion semver.Version) bool {
 // It returns an error if any of these functions fails at any point.
 func Update(latestVersion semver.Version) error {
 	var extName string
+
+	path, _ := osext.Executable()
+	if writable := util.IsDirectoryWritable(path); !writable {
+		return fmt.Errorf("Directory '%s' binary doesn't have write permission.\nPlease fix the permission issue and try running the command again.", filepath.Dir(path))
+	}
 
 	// Temporary directory to store archive contents
 	tmpDir, err := ioutil.TempDir("", "download")
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -19,6 +19,8 @@ package util
 import (
 	"fmt"
 	"io"
+	"os"
+	"path/filepath"
 	"strings"
 	"time"
 )
@@ -177,4 +179,18 @@ func CommandExecutesSuccessfully(cmd string, args ...string) bool {
 		return true
 	}
 	return false
+}
+
+// IsDirectoryWritable checks if the directory is writable or not
+// by trying creating a temporary file
+// Return true if directory is writable else false
+func IsDirectoryWritable(path string) bool {
+	tmpFilePath := filepath.Join(path, "tmp")
+	_, err := os.Create(tmpFilePath)
+	defer os.Remove(tmpFilePath)
+	if err != nil {
+		return false
+	}
+
+	return true
 }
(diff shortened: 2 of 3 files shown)


pkg/util/utils.go
```diff
@@ -178,3 +179,17 @@ func CommandExecutesSuccessfully(cmd string, args ...string) bool {
 	}
 	return false
 }
+
+// IsDirectoryWritable checks if the directory is writable or not
+// by trying creating a temporary file
+// Return true if directory is writable else false
+func IsDirectoryWritable(path string) bool {
```
budhrg: Couldn't find a direct logic hence using it. Let me know if have good solution.

anjannath: I think this is good, it finds out the truth if the directory is actually wirtable or not.
I was thinking about using `Stat()` and get the `Mode().Perm()` but this will only work in unix like systems.
> IsReadable and IsWritable wouldn't be reliable enough... the filesystem can be mounted read-only, or corrupt, or there could be extended attribute ACLs on the file, etc.  The only reliable strategy is to just try to open it for read or write and see if it fails.


Found this mail threaed on the golang-dev mailing list abou it: https://groups.google.com/forum/#!msg/golang-dev/qfenPFNYdYY/8tkDvE3KaUQJ

budhrg: If `Brad Fitzpatrick` is saying then I am sure it is one of good approach :)

pkg/minishift/update/update.go
```diff
@@ -78,6 +80,11 @@ func IsNewerVersion(localVersion, latestVersion semver.Version) bool {
 func Update(latestVersion semver.Version) error {
 	var extName string
 
+	path, _ := osext.Executable()
+	if writable := util.IsDirectoryWritable(path); !writable {
+		return fmt.Errorf("Directory '%s' containing binary doesn't have write permission.\nPlease fix the permission issue and try running the command again.", filepath.Dir(path))
+	}
```
LalatenduMohanty: s/containing binary// i.e. `Directory '%s' doesn't have write permission.\nPlease fix the permission issue and try running the command again.`

budhrg: Sure

budhrg: Done

LalatenduMohanty:
@budhrg looks like the appveyor test failed without any reason can you do another force push to see if we can reproduce the issue. 
```
-- PASS: TestSplitAndTrim (0.00s)
760
PASS
761
ok  	github.com/minishift/minishift/pkg/util/strings	0.093s
762
?   	github.com/minishift/minishift/pkg/version	[no test files]
763
make: *** [test] Error 1
764
Command exited with code 2
```

pkg/util/utils_test.go
```diff
@@ -206,3 +207,24 @@ func Test_command_executes_unsuccessfully_with_command_lookup(t *testing.T) {
...
+	assert.True(t, IsDirectoryWritable(fmt.Sprintf("%s/foo", testDir)))
+}
+
+func TestNonWritableDirectory(t *testing.T) {
+	testDir, err := ioutil.TempDir("", "minishift-test-")
+	assert.NoError(t, err)
+	defer os.RemoveAll(testDir)
+
+	if runtime.GOOS == "windows" {
+		testDir = "C:/Windows/System" // using some known non-writable directory
```
budhrg: `os.Chmod` is not working in Windows.

pkg/util/utils.go
```diff
@@ -178,3 +180,17 @@ func CommandExecutesSuccessfully(cmd string, args ...string) bool {
...
+	tmpFilePath := filepath.Join(path, "tmp")
+	_, err := os.Create(tmpFilePath)
+	defer os.Remove(tmpFilePath)
```
containscafeine: `defer` statement should be called after the error is checked

PR #2079, opened 2018-03-03
[WIP] Confirm install choco install make

budhrg:


budhrg:
The issue is intermittent. Will debug again when it is more consistent.

Issue #2092, opened 2018-03-07
Make Makefile self-documenting


hferentschik:

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/pull/2093

Issue #959, opened 2017-05-27
Allow hostfolder configuration non interactively 


ruifortes:
Hi.
Sorry for the off-topic post but since the command I'm struggling is from minishift I hope yu don't mind me posting.
I not at all proficient with linux but surely this is a n brainer for you guys.
I'm trying to call "minishift hostfolder add" in a shell script (wel...actually a npm script) that automatically fills all input prompts.
I followed [this](https://askubuntu.com/questions/338857/automatically-enter-input-in-command-line) instructions but something gets broken and although "add" works the "mount" fails.

Heres the line I use:
`printf '//192.168.1.33/datafolder\n/mnt/sda1/datafolder\nsomeuser\nsomepass\n\n' | minishift hostfolder add mymount`
It adds but mount fails.
if I enter manually it works

Can someone tell me how to fill all input in a single command?
Thanks

gbraad:
Can you check the config file for the VM ? what does the entry contan for your defined hostfolder?

BTW, we will offer parameter options in a future increment.

hferentschik:
@gbraad Do you want to rename this issue to match what we are planning to do, aka "Add command line switches for `minishift hostfolder add`" or shall we create a new issue and reference it from here?

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/pull/2093

PR #2093, opened 2018-03-07
Issue #317 Refactoring existing hostfolder code and implementing sshfs hostfolders

hferentschik:
Addresses issue #317, #959 and #2092.

- Add hidden sftpd command
- Refactor existing hostfolder actions
- Introducing hostfolder.Manager as entry point for hostfolder operations
- Introducing HostFolder interface as common abstraction for any host folder type
- Split impl details between CIFS and SSHFS
- Allow to configure host folder interactively as well as non-interactively (Issue #959)
- Adding docs


hferentschik:
Tested on macOS against default and centos ISO. Needs testing against Minikube and RHEL ISO as well as testing on Windows. For the latter, some special handling might be required. 

There are a few follow up tasks which needs addressing. I will create them once merged.

LalatenduMohanty:
@hferentschik I was testing the PR on macOS (yet to fix my Windows machine) and encountered this issue. 

When I try to open a file through the hostfolder mount in the VM I get this everytime (after deleting the swp file)

```
$ vi /mnt/sda1/myshare/hello
E325: ATTENTION
Found a swap file by the name "/mnt/sda1/myshare/.hello.swp"
             dated: Thu Mar  8 04:17:15 2018
         [cannot be read]
While opening file "/mnt/sda1/myshare/hello"
             dated: Thu Mar  8 04:05:38 2018

(1) Another program may be editing the same file.  If this is the case,
    be careful not to end up with two different instances of the same
    file when making changes.  Quit, or continue with caution.
(2) An edit session for this file crashed.
    If this is the case, use ":recover" or "vim -r /mnt/sda1/myshare/hello"
    to recover the changes (see ":help recovery").
    If you did this already, delete the swap file "/mnt/sda1/myshare/.hello.swp"
    to avoid this message.
"/mnt/sda1/myshare/hello" 1L, 6C
```

I am not sure why I am seeing this.

LalatenduMohanty:
@hferentschik However I really like the seamless experience as compared to CIFS. 👍

cmd/minishift/cmd/config/config.go
```diff
@@ -95,6 +95,7 @@ var (
 	// Host Folders
 	HostFoldersMountPath = createConfigSetting("hostfolders-mountpath", SetString, nil, nil, true, nil)
 	HostFoldersAutoMount = createConfigSetting("hostfolders-automount", SetBool, nil, nil, true, nil)
+	HostFoldersSftpPort  = createConfigSetting("hostfolders-sftp-port", SetInt, []setFn{validations.IsPositive}, nil, true, nil)
```
gbraad: Improvement cange to make sureport is between 1-65536, or 1024-65536

cmd/minishift/cmd/hostfolder/add.go
```diff
@@ -20,45 +20,223 @@ import (
 	"fmt"
 	"runtime"
 
-	hostfolderActions "github.com/minishift/minishift/pkg/minishift/hostfolder"
+	"errors"
+	miniConfig "github.com/minishift/minishift/pkg/minishift/config"
+	hf "github.com/minishift/minishift/pkg/minishift/hostfolder"
```
gbraad: Not a huge fan of abbrev. when concerning vars or libs.

cmd/minishift/cmd/hostfolder/add.go
```diff
@@ -20,45 +20,223 @@ import (
...
+	}
 
-		if err != nil {
-			atexit.ExitWithMessage(1, fmt.Sprintf("Failed to mount host folder: %s", err.Error()))
+	switch shareType {
+	case hf.CIFS.String():
+		if interactive {
+			addCIFSInteractive(hostFolderManager, name)
+		} else {
+			addCIFSNonInteractive(hostFolderManager, name)
```
gbraad: 👍

cmd/minishift/cmd/hostfolder/add.go
budhrg: +1

cmd/minishift/cmd/hostfolder/add.go
```diff
@@ -20,45 +20,223 @@ import (
...
+	instanceOnlyFlag     = "instance-only"
+	usersShareFlag       = "users-share"
+)
+
 var (
 	instanceOnly bool
 	usersShare   bool
+	interactive  bool
+	shareType    string
+	source       string
```
gbraad: Source and target are globally derined, but used in hthe add code... why this choice?

gbraad: Ahh, now i see... They come from the arguments on start

cmd/minishift/cmd/hostfolder/add.go
```diff
@@ -20,45 +20,223 @@ import (
...
+
+	return nil
+}
+
+func addCIFSInteractive(manager *hf.Manager, name string) error {
+	var uncPath string
+	if usersShare {
+		uncPath = "[determined on startup]"
+	} else {
+		uncPath = util.ReadInputFromStdin("UNC path")
```
gbraad: Check for tty support is also relevant here. i would suggest to error out immediately on entry of this function when no tty support.. Or even earlier, like the switch for types

cmd/minishift/cmd/hostfolder/add.go
```diff
@@ -20,45 +20,223 @@ import (
...
+		return fmt.Errorf(nonSupportedTtyError)
+	}
+
+	password := util.ReadPasswordFromStdin("Password")
+	password, err := util.EncryptText(password)
+	if err != nil {
+		return err
+	}
+
+	domain := util.ReadInputFromStdin("Domain")
```
gbraad: Would also need tty support... right?

cmd/minishift/cmd/hostfolder/add_test.go
```diff
@@ -0,0 +1,145 @@
...
+	assert.NoError(t, err, "Unexpected error setting instance config")
+
+	tee := cli.CreateTee(t, true)
+	defer cli.TearDown(tmpMinishiftHomeDir, tee)
+	defer viper.Reset()
+
+	atexit.RegisterExitHandler(cli.VerifyExitCodeAndMessage(t, tee, 1, fmt.Sprintf(unknownType, "snafu")))
+	source = "/home/johndoe"
+	target = "/var/tmp"
+	shareType = "snafu"
```
gbraad: Better than TARFUN or FUBAR

cmd/minishift/cmd/hostfolder/list.go
```diff
@@ -17,38 +17,68 @@ limitations under the License.
...
+	"os"
+	"text/tabwriter"
 )
 
-var hostfolderListCmd = &cobra.Command{
+var listCmd = &cobra.Command{
 	Use:   "list",
 	Short: "Lists the defined host folders.",
-	Long:  `Lists an overview of the defined host folders that can be mounted to a running OpenShift cluster.`,
+	Long:  `Lists an overview of the defined host folders that can be mounted into the Minishift VM.`,
```
gbraad: Why minishift vm here, while in previous files this was changed to openshift cluster...

cmd/minishift/cmd/hostfolder/mount.go
```diff
@@ -32,8 +31,8 @@ var (
 	mountAll bool
 )
 
-var hostfolderMountCmd = &cobra.Command{
-	Use:   "mount HOSTFOLDER_NAME",
+var mountCmd = &cobra.Command{
+	Use:   "mount HOST_FOLDER_NAME",
 	Short: "Mounts the specified host folder to the running OpenShift cluster.",
```
gbraad: Here it is cluster...

cmd/minishift/cmd/hostfolder/mount.go
```diff
@@ -67,6 +68,6 @@ var hostfolderMountCmd = &cobra.Command{
 }
 
 func init() {
-	HostfolderCmd.AddCommand(hostfolderMountCmd)
-	hostfolderMountCmd.Flags().BoolVarP(&mountAll, "all", "a", false, "Mounts all defined host folders to the running OpenShift cluster.")
+	HostFolderCmd.AddCommand(mountCmd)
+	mountCmd.Flags().BoolVarP(&mountAll, "all", "a", false, "Mounts all defined host folders into the Minishift VM.")
```
gbraad: Why VM?

cmd/minishift/cmd/hostfolder/sftpd.go
```diff
@@ -0,0 +1,193 @@
+/*
+Copyright (C) 2017 Red Hat, Inc.
```
gbraad: 2018

cmd/minishift/cmd/hostfolder/sftpd.go
```diff
@@ -0,0 +1,193 @@
...
+func runSftp(cmd *cobra.Command, args []string) {
+	serverConfig := serverConfig()
+
+	port := viper.GetInt(config.HostFoldersSftpPort.Name)
+	if port == 0 {
+		port = serverPort
+	}
+
+	// Once a ServerConfig has been configured, connections can be accepted.
+	listener, err := net.Listen("tcp", fmt.Sprintf("0.0.0.0:%d", port))
```
gbraad: Can we be more restrictive than to bind on all interfaces?

cmd/minishift/cmd/hostfolder/umount.go
```diff
@@ -17,24 +17,21 @@ limitations under the License.
...
 )
 
-var hostfolderUmountCmd = &cobra.Command{
-	Use:   "umount HOSTFOLDER_NAME",
-	Short: "Unmount a host folder from the running OpenShift cluster.",
-	Long:  `Unmount a host folder from the running OpenShift cluster. This command does not remove the host folder definition or the host folder itself.`,
+var umountCmd = &cobra.Command{
+	Use:   "umount HOST_FOLDER_NAME",
+	Short: "Umount a host folder from the running Minishift VM.",
+	Long:  `Umount a host folder from the running Minishift VM. This command does not remove the host folder config or the host folder itself.`,
```
gbraad: Ditto

pkg/minishift/hostfolder/hostfolder_manager.go
```diff
@@ -0,0 +1,262 @@
...
+type MountInfo struct {
+	Name       string
+	Type       string
+	Source     string
+	MountPoint string
+	Mounted    bool
+}
+
+// Manager is the central point for all operations around managing hostfolders.
+type Manager struct {
```
gbraad: more related to config than management

cmd/minishift/cmd/hostfolder/sftpd.go
gbraad: 2018

LalatenduMohanty:
I also saw changes required to Gopkg.lock after `make clean;make`

```diff
 git diff
diff --git a/Gopkg.lock b/Gopkg.lock
index 0b9b0ab9..7d72da70 100644
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -216,7 +216,7 @@

 [[projects]]
   name = "github.com/magiconair/properties"
-  packages = ["."]
+  packages = [".","assert"]
   revision = "c3beff4c2358b44d0493c7dda585e7db7ff28ae6"
   version = "v1.7.6"

@@ -447,6 +447,6 @@
 [solve-meta]
   analyzer-name = "dep"
   analyzer-version = 1
-  inputs-digest = "56f18a532b86c7741de018c586555a7e21d193213d341c08dad3a3c296eaaa90"
+  inputs-digest = "2a66ab61973ffea916212a52397b72af4a939485f12657eb1e99c5278b6def5d"
   solver-name = "gps-cdcl"
   solver-version = 1
```

cmd/minishift/cmd/hostfolder/umount.go
budhrg: @gbraad Didn't get your comment here. Are you pointing `umount` ?

cmd/minishift/cmd/hostfolder/util.go
```diff
@@ -0,0 +1,95 @@
...
+	"strings"
+)
+
+const (
+	HostfoldersDefaultPath  = "/mnt/sda1"
+	HostfoldersMountPathKey = "hostfolders-mountpath"
+)
+
+var (
+	optionsPattern  = regexp.MustCompile(`^([a-z]+=.*,)?([a-z]+=.*)$`)
```
budhrg: I think `optionsPattern/keyValuePattern` need to be renamed to something like `optionsRegex` since when we say pattern I was thinking only regex pattern.

pkg/util/strings/strings.go
```diff
@@ -115,3 +115,7 @@ func SplitAndTrim(s string, separator string) ([]string, error) {
 
 	return cleanSplit, nil
 }
+
+func ConvertSlashes(input string) string {
```
budhrg: Can we rename to specific like `ConvertToForwardSlash` ?

pkg/minishift/hostfolder/hostfolder_test.go
```diff
@@ -17,54 +17,10 @@ limitations under the License.
...
-			"mountpoint": "",
-			"uncpath":    "//127.0.0.1/Users",
-			"username":   "joe@pillow.us",
-			"password":   "am!g@4ever",
-			"domain":     "DESKTOP-RHAIMSWIN",
-		},
-	}
-}
-
-func TestHostfolderIsMounted(t *testing.T) {
```
budhrg: Does this mean we are not testing CIFS implementation?

gbraad: Cifs needs to have samba setup or use Windows. This for now an issue to setup...

cmd/minishift/cmd/hostfolder/umount.go
gbraad: Ditto for use of VM again, while also to OpenShift cluster exist in this pr.

cmd/minishift/cmd/hostfolder/add.go
hferentschik: Not quite sure what you mean

cmd/minishift/cmd/hostfolder/add.go
hferentschik: you guys tell me. This code I jsut refaxcrtored. The original code from hostfolder.go looks like this:

```Golang
func Add(name string, allInstances bool) error {
	if isHostfolderDefinedByName(name) {
		return fmt.Errorf("Already have a host folder defined for: '%s'", name)
	}

	uncpath := util.ReadInputFromStdin("UNC path")
	if len(uncpath) == 0 {
		return fmt.Errorf("No remote path has been given")
	}
	mountpoint := readInputForMountpoint(name)
	username := util.ReadInputFromStdin("Username")
	if !util.IsTtySupported() {
		return fmt.Errorf(NonSupportedTtyError)
	}
	password := util.ReadPasswordFromStdin("Password")
	domain := util.ReadInputFromStdin("Domain")
	password, err := util.EncryptText(password)
	if err != nil {
		return err
	}

	addToConfig(newCifsHostFolder(
		name,
		uncpath,
		mountpoint,
		username, password, domain),
		allInstances)

	return nil
}
```

I find it also odd why this check is only on Username, but tbh I don't know. Happy to do as you suggest and wrap the whole interactive process into a check for TTY support. 

@praveenkumar anything on this one? I believe you might be the one who added it

cmd/minishift/cmd/hostfolder/add_test.go
hferentschik: :-)

cmd/minishift/cmd/hostfolder/list.go
hferentschik: Other way around. I tried to changed it from OpenShift cluster to Minishift VM and missed to do so in the mount command. I believe that the right way to describe this is to mount something into the VM or into the filesystem of the VM. OpenShift is not even involved here and talking about OpenShift cluster might also create wrongly some association to persistent volumes.

cmd/minishift/cmd/hostfolder/mount.go
hferentschik: This is the one I actually forgot to change.

cmd/minishift/cmd/hostfolder/mount.go
hferentschik: Because hostfolders is about the VM and its file system, not the OpenShift cluster at all.

cmd/minishift/cmd/hostfolder/sftpd.go
hferentschik: well, it depends. I created this file last year.

cmd/minishift/cmd/hostfolder/sftpd.go
hferentschik: maybe, needs some doing. I'd go with this for now and create a follow up issue. There are some caveats here.

hferentschik:
> I also saw changes required to Gopkg.lock after make clean;make

My wrong. Wrong accidental import by the IDE

cmd/minishift/cmd/hostfolder/mount.go
gbraad: Also my thinking, so rather see this being consistent

Makefile
```diff
@@ -196,43 +196,48 @@ clean:
...
+fmtcheck: ## Checks for style violation using gofmt
 	@gofmt -l -s $(SOURCE_DIRS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
 
 .PHONY: validate_commits
-validate_commits: $(GOPATH)/bin/git-validation
+validate_commits: $(GOPATH)/bin/git-validation ## Validates commit messages match pattern ^(Issue #[0-9]+ .*|cut v[0-9]+\.[0-9]+\.[0-9]+)
 	@# Need to add $$ to avoid shell interpretation/evaluation
 	git-validation -q -run short-subject,message_regexp='^(Issue #[0-9]+ .*|cut v[0-9]+\.[0-9]+\.[0-9]+)$$' -range $(START_COMMIT_MESSAGE_VALIDATION)...
+
+.PHONY: help
```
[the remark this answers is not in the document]
hferentschik: I thought you would like it :-)

cmd/minishift/cmd/hostfolder/util.go
hferentschik: I am not following you on this one

pkg/util/strings/strings.go
hferentschik: This is really an existing function. I just moved it. Sure it can be renamed, but personally I don't think it is required. That said, there is actually a bug here which I think should be addressed in a follow up issue.

pkg/minishift/hostfolder/hostfolder_test.go
hferentschik: We really never tested CIFS as such. This was mocked test and due to the changes of the code I removed it. 

I agree that there are more tests missing. There are several areas where I would like to add more tests, but as discussed I don't have the bandwidth for that right now. 

Generally,  I would start with a set of integration tests for this. This would allow to test the whole range of hostfolder commands end to end. The SSHFS host folders have no prerequistes and work on all OS types, so we can write a single set of integration tests. I think doing this would give most value for money. Basically go through the cycle of:

- minishift hostfolder list -> nothing there
- minishift hostfolder add ... -> host folder defined
- minishift hostfolder list -> host folder defined, but not mounted
- minishift hostfolder mount -> host folder gets mounted, runs some commands that the file system is mounted and writable, etc
- minishift hostfolder list -> host folder defined and mounted
- minishift hostfolder umount
- minishift hostfolder delete

All this should be easily doable in a single cmd-hostfolder.feature.

If we are talking tests, I think that's what really missing here.
cc @agajdosi

hferentschik:
I was able to test this on Windows 10 as well now. Needed to make another minor adjustment, but now it works.

praveenkumar:
minikube doesn't have `sshfs` by default so this is something we need to add to the ISO. Other than that I tested this PR on the OSx with RHEL and b2d ISO and mount works as expected. Code review is still pending from my side.

cmd/minishift/cmd/hostfolder/util.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/minishift/minishift/pkg/minishift/hostfolder"
+	"github.com/minishift/minishift/pkg/util"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/viper"
+	"regexp"
+	"strings"
+)
+
+const (
+	HostfoldersDefaultPath  = "/mnt/sda1"
```
praveenkumar: This hard-coded disk path not going to work with hyperkit or xhyve when using raw disk and it changes to `/mnt/vda1`.  As per code, this is only used when a user try to use interactive mode. Can we not add this as part of default string and let user provide the complete path? We can, however, convey some way to use /mnt/sda1 or /mnt/vda1 but don't append ourself.

cmd/minishift/cmd/hostfolder/add.go
```diff
@@ -20,45 +20,231 @@ import (
...
-			}
-			err = hostfolderActions.Add(args[0], !instanceOnly)
+func init() {
+	HostFolderCmd.AddCommand(addCmd)
+	addCmd.Flags().StringVarP(&shareType, shareTypeFlag, "t", "cifs", "The host folder type. Allowed types are [cifs|sshfs]. SSHFS is experimental.")
+	addCmd.Flags().StringVar(&source, sourceFlag, "", "The source of the host folder.")
+	addCmd.Flags().StringVar(&target, targetFlag, "", "The target (mount point) of the host folder.")
+	addCmd.Flags().StringVar(&options, optionsFlag, "", "Host folder type specific options.")
+	addCmd.Flags().BoolVar(&instanceOnly, instanceOnlyFlag, false, "Defines the host folder only for the current Minishift instance.")
+	addCmd.Flags().BoolVarP(&interactive, interactiveFlag, "i", false, "Allows to interactively provide the required parameters.")
```
gbraad: This changes the current behaviour and is NOT self-explanatory to those who already use it

gbraad:
On a clean startup
```
$ minishift hostfolder add tmp -i --type sshfs
source path: /tmp/
Mountpoint [/mnt/sda1/tmp]: 
$ minishift hostfolder mount tmp
Error mounting the host folder: error occured while mounting host folder: ssh command error:
command : sudo sshfs docker@192.168.42.1:/tmp/ /mnt/sda1/tmp -o IdentityFile=/home/docker/.ssh/id_rsa -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p 2022
err     : exit status 1
output  : read: Connection reset by peer
```

praveenkumar:
@hferentschik I am also able to see the file ownership inside the VM for xhyve driver is not as expected as @LalatenduMohanty pointed out https://github.com/minishift/minishift/pull/2093#issuecomment-371428106 here.

LalatenduMohanty:
> On a clean startup

I also get the issue, but when I again tried it worked fine. Not sure what caused the first issue. May be the sftp server did not respond.

LalatenduMohanty:
I am also seeing the swap issue on Fedora 27+KVM+CentOS https://github.com/minishift/minishift/pull/2093#issuecomment-371428106

cmd/minishift/cmd/hostfolder/list.go
LalatenduMohanty: I think Minishift VM is more transparent.

pkg/minishift/hostfolder/hostfolder_test.go
budhrg: Make sense @hferentschik

hferentschik:
I rebased onto master. I also addressed the sftp server shutdown using a counter as discussed. Looking good to me now. Going to respond to comments.

hferentschik:
> minikube doesn't have sshfs by default so this is something we need to add to the ISO.

ohh, I thought they do. Will adding this be an issue? I guess we can add a note to the docs for now.

> Other than that I tested this PR on the OSx with RHEL and b2d ISO and mount works as expected. 

Sweet.

cmd/minishift/cmd/config/config.go
hferentschik: > Maybe something we can do later if @hferentschik doesn't have time to spend here.

Right, this would be one of the follow up issues I will create. I don't think it makes sense for me to spend time on these type of things now.

hferentschik: > Also we need a config option as hostfolders-type for sshfs or CIFS cc 

For what would you need that? I actually don't think this is needed. I am not sure what you try to achieve with it.

cmd/minishift/cmd/hostfolder/add.go
hferentschik: I renamed it to _hostFolderConfig_.

cmd/minishift/cmd/hostfolder/add.go
hferentschik: True, this changes the current behaviour. However, I think it was the 'hostfolder add' which was the odd one. Really the default mode for all commands should be non-interactive. 

I don't think this should be a big issue for anyone who is using CIFS already. It is also more connected to #959 than #317.

Eventually, I would not only make the default non-interactive but also switch the default type from CIFS to SSHFS. That's after we gained some experience with it.

@minishift/minishift-dev ^^ any more thoughts on what the default should be?

cmd/minishift/cmd/hostfolder/add.go
hferentschik: Got you

hferentschik: moved the check up

cmd/minishift/cmd/hostfolder/mount.go
hferentschik: It should be now

cmd/minishift/cmd/hostfolder/util.go
hferentschik: This is existing code. I have not changed that, just moved. That said, this is just the default string displayed. It is used when the user just hits enter. The user can at any stage provide a full path. 

It makes sense to give the user some hint on where he might want to mount to. I think the docs could be extended in this area.

If this particular suggestion causes issues, I suggest creating a follow-up issue. Either we don't display any default at all or do something specific depending on the VM. Right now, I don't see this as a directly related.

hferentschik:
Addresses (hopefully) outstanding comments. Either by making small adjustments or by responding to the comment. Let me know what you think.

hferentschik:
Removing the WIP status from my side. I would be happy to merge as is and create a few follow up issues.

cmd/minishift/cmd/hostfolder/add.go
gbraad: If #959 would solve the interactive flow, than maybe it should not have been addressed here. Just my 2c as it does change bahaviour.

gbraad: Eventually we will suggest SSHFS, and later move to SSHFS as default, But for now, we have been experiencing different results on platforms, from unexplained errors to super-pleasant experience.

cmd/minishift/cmd/hostfolder/util.go
gbraad: Originally suggesting was to use `/mnt/data` instead of `/mnt/sda1`, but this was never addressed. Maybe it is now a good time to address this... in a new issue.

cmd/minishift/cmd/config/config.go
gbraad: I think they are looking for a user-set default. So if I ONLY use sshfs, I can set this as a the 'default' type.

gbraad:
I have never seen it work... on different hosts: centos 7, fedora 27, ubuntu... all for the same error:

```
$ minishift hostfolder mount tmp -v5
-- minishift version: v1.14.0+aafadfd7
sftpd running with pid 35570
sudo sshfs docker@192.168.42.1:/tmp/ /mnt/sda1/tmp -o IdentityFile=/home/docker/.ssh/id_rsa -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p 2022
Error mounting the host folder: error occured while mounting host folder: ssh command error:
command : sudo sshfs docker@192.168.42.1:/tmp/ /mnt/sda1/tmp -o IdentityFile=/home/docker/.ssh/id_rsa -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p 2022
err     : exit status 1
output  : read: Connection reset by peer
```

pkg/minishift/hostfolder/sshfs_hostfolder.go
```diff
@@ -0,0 +1,188 @@
...
+		return err
+	}
+
+	ip, err := h.hostIP(driver)
+	if err != nil {
+		return err
+	}
+
+	cmd := fmt.Sprintf(
+		"sudo sshfs docker@%s:%s %s -o IdentityFile=%s -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p %d",
```
gbraad: and what if I do not ave a `docker` user on my host?

gbraad:
It does not property handled the error state when the folder is already mounted:

```
$ minishift hostfolder mount --all
-- Mounting host folders
$ minishift hostfolder mount tmp
Error mounting the host folder: error occured while mounting host folder: ssh command error:
command : sudo sshfs docker@192.168.42.1:/tmp/ /mnt/sda1/tmp -o IdentityFile=/home/docker/.ssh/id_rsa -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p 2022
err     : exit status 1
output  : fuse: mountpoint is not empty
fuse: if you are sure this is safe, use the 'nonempty' mount option

$    
```

pkg/minishift/hostfolder/hostfolder_test.go
gbraad: * minishift hostfolder mount -> host folder gets mounted, runs some commands that the file system is mounted and writable, etc
  * minishift hostfolder mount -> host folder errors out with 'Mounted already'.

hferentschik:
> It does not property handled the error state when the folder is already mounted:

Right, good catch. That I can easily add. Turns out there is a check for that in CIFS, but not in SSFS. I take care of this one.

hferentschik:
> So we have possible firewall issues.

Sure.
 
> I suggest to document the need to open port 2022 on the host, 

I can do that.

> make the error message more clear; "Unable to connect. Please make sure port {SFTP_PORT} is opened in your firewall", or something.

Well, I am not so sure whether we can generally conclude that it will always be a firewall issue. Ther error might be other issues. Let me think about it.

> I suggest in a follow-up to use port knocking to determine if the port is reachable... for now, error messaging might suffice.

Right, that would be what I suggest as well.

hferentschik:
Pushed an update

* Handling umounting of a not mounted host folder more gracefully
* Make mounting more resilient 
* Added notes about Minikube ISO and firewall settings to docs

hferentschik:
Rebased. Where do we stand on this one? @minishift/minishift-dev

LalatenduMohanty:
@hferentschik I am going to merge this today. Can you please put list of issues we need to create?

cmd/minishift/cmd/config/config.go
hferentschik: I don't think that makes a lot of sense. This is not like the start command where typing flags become tedious. If you go down this road where do you draw the line?

cmd/minishift/cmd/hostfolder/add.go
hferentschik: > If #959 would solve the interactive flow, than maybe it should not have been addressed here. 

It was always the plan to address the interactive vs non-ineractive flow as part of the issue. It was a natural fall-out of the refactoring. Both modes are there. One just needs to set the default for the `interactive` flag to true and you have the old behaviour. The question is what makes most sense. IMO non-interactive. However, if making the default interactive gets us closer to merging this, I can do it. I was basically hoping for some more opinion on this one (@minishift/minishift-dev ). ATM, it seems to be 1:1. Anyone wants to weigh in to push the scale one way or the other?

cmd/minishift/cmd/hostfolder/add.go
hferentschik: Done

hferentschik:
> I am going to merge this today.

Awesome. Thanks a ton.

>  Can you please put list of issues we need to create?

Absolutely. Stay tuned. I have several, more or less directly realted to this pull request. Most of them are minor really.

cmd/minishift/cmd/hostfolder/util.go
LalatenduMohanty: @gbraad agree. We should have a follow up issue for this.

pkg/util/strings/strings.go
LalatenduMohanty: Ok, another followup issue :)

hferentschik:
Thanks 👍

Issue #2108, opened 2018-03-09
Document use of setx for Hyper-V virtual switch


robin-owen:
The Hyper-V virtual switch documentation in **_getting-started/setting-up-driver-plugin.adoc_** uses `set` to set environment variables for Command Prompt and `$env` to set them for PowerShell. Both of these methods will change the environment for the current shell session, but will not persist. The `setx` utility can set environment variables in the registry to persist these settings, but will not affect the current shell. As such, both `set` and `setx` must be used for Command Prompt, and both `$env` and `setx` must be used for PowerShell.

LalatenduMohanty:
Resolved by #2109

PR #2109, opened 2018-03-09
Issue #2108 Document use of setx for Hyper-V virtual switch

robin-owen:


budhrg:
> This is to persist the config accross restarts.. right?

@anjannath For enabling nested virtualization.

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -209,6 +209,7 @@ For example, on Command Prompt use:
 +
 ----
 C:\> set HYPERV_VIRTUAL_SWITCH=External (Wireless)
+C:\> setx HYPERV_VIRTUAL_SWITCH %HYPERV_VIRTUAL_SWITCH%
```
budhrg: I think this will persist the setting and we should add message like
```
If you want to persist this environment variable, perform as below:
```

something on this line.

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -222,4 +223,5 @@ However, on PowerShell you need to use the quotes:
 +
 ----
 PS> $env:HYPERV_VIRTUAL_SWITCH="External (Wireless)"
+PS> setx HYPERV_VIRTUAL_SWITCH %HYPERV_VIRTUAL_SWITCH%
```
budhrg: Same as https://github.com/minishift/minishift/pull/2109/files#r173818592

anjannath: This is not working in case of `Posh`, It is working in `cmd` .
For some reason PowerShell is not expanding `%HYPERV_VIRTUAL_SWITCH%`
```
C:\Users\CDK\Documents> .\minishift.exe start --iso-url centos
-- Starting profile 'centos'
-- Checking if requested OpenShift version 'v3.7.1' is valid ... OK
-- Checking if requested OpenShift version 'v3.7.1' is supported ... OK
-- Checking if requested hypervisor 'hyperv' is supported on this platform ... OK
-- Checking if Hyper-V driver is installed ... OK
-- Checking if Hyper-V driver is configured to use a Virtual Switch ... OK
-- Starting local OpenShift cluster using 'hyperv' hypervisor ...
   Memory:    2 GB
   vCPUs :    2
   Disk size: 20 GB
-- Starting Minishift VM ..... FAIL E0313 02:53:43.604921     864 start.go:369] Error starting the VM: Error creating the VM. Error with pre-create check: "vswitch \"%HYPERV_VIRTUAL_SWITCH%\" not found". Retrying.
Error starting the VM: Error creating the VM. Error with pre-create check: "vswitch \"%!H(MISSING)YPERV_VIRTUAL_SWITCH%!\(MISSING)" not found"
```

robin-owen:
@anjannath Yes, this is to persist the environment variable across both restarts and shell instances. The documentation has been expanded to provide this information.

Also, the steps have been reorganized so that they may be done either in the same shell instance or across different shell instances instead of making the assumption that they would be performed one after the other.

@anjannath @budhrg Please see the updated information. :)

Thank you!

docs/source/getting-started/setting-up-driver-plugin.adoc
robin-owen: That's my mistake -- PowerShell would need to use `$env:HYPERV_VIRTUAL_SWITCH` to expand the environment variable, not `%HYPERV_VIRTUAL_SWITCH%` like in Command Prompt.

This has been rewritten to avoid using shell expansion due to the fact that the environment variable may not be defined when using `setx` for some users.

docs/source/getting-started/setting-up-driver-plugin.adoc
```diff
@@ -223,3 +223,12 @@ However, on PowerShell you need to use the quotes:
 ----
 PS> $env:HYPERV_VIRTUAL_SWITCH="External (Wireless)"
 ----
++
+Using `set` in Command Prompt or `$env` in PowerShell will only affect the currently running shell session.
```
anjannath: :+1:

docs/source/getting-started/setting-up-driver-plugin.adoc
anjannath: Thanks @robin-owen for the clarification..

Issue #199, opened 2016-12-19
Error accessing web console in Safari browser after minishift start 


mmalina:
I tried out Minishift 1.0.0-beta.1 on macOS Sierra with VirtualBox.
Started with this command:
./minishift --vm-driver=virtualbox start

After it successfully started, I entered https://192.168.99.100:8443 in my browser.
It asked me which certificate to use. There were two offered, I chose the first, not sure what this is about. Then I allowed everything when prompted.
But in the end, I keep being redirected to the error page:
https://192.168.99.100:8443/console/error?error_description=Unable%20to%20load%20details%20about%20the%20server.%20If%20the%20problem%20continues,%20please%20contact%20your%20system%20administrator.&error=API_DISCOVERY

mmalina:
Update: I tried to access the console again in an anonymous web browser window and it just keeps Loading... (4 minutes already)

hferentschik:
> It asked me which certificate to use.

What exactly do you mean? I don't think I have seen this. It should behave as current CDK. OpenShift uses self signed certificates, so you will need to add a security exception, but I am not sure what this "selection" you describe is about. Can you re-produce the problem and how? Can you give some more context about what you had to do. Messages displayed or maybe a screenshot?

If you want to try from scratch, delete your ~/.minishift directory and start over. Does the problem still occur?

mmalina:
So I tried again. In Chrome, it works exactly as you describe - I am told that the connection is not safe (self-issued cert) and am able to add an exception and then see the console ok.
But in Safari, this doesn't work for me right now.
When I entered the url now, first thing the browser showed me was this:
https://www.dropbox.com/s/mangb10uejssdxt/Screenshot%202016-12-20%2010.47.50.png?dl=0
Strange.
Then I got this:
https://www.dropbox.com/s/dc3ctnxu10c7rf1/Screenshot%202016-12-20%2010.48.10.png?dl=0
And when I clicked Continue, I got the same message again and again and again and was not able to access the console at all.
I guess it's a problem of Safari, but I'm not exactly sure what's the problem here.
Restarting the browser may help, I will try that.

hferentschik:
> I guess it's a problem of Safari, but I'm not exactly sure what's the problem here.

Now that you mention it, I remember seeing this once as well. I forgot about it, since I usually use Firefox or Chrome. I am also not sure what the cause for this is. There should really by no difference.
[...]

PR #2113, opened 2018-03-11
Issue #199 Adding Safari browser issue to troubleshooting

LalatenduMohanty:

### DIFF
--- a/docs/source/troubleshooting/troubleshooting-getting-started.adoc
+++ b/docs/source/troubleshooting/troubleshooting-getting-started.adoc
@@ -138,3 +138,18 @@ Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/mini
  65 B / 65 B [===================================================================================================================================================] 100.00% 0s
 Update failed: open /usr/bin/.minishift.new: permission denied
 ----
+
+[[openshift-web-console-does-not-work-in-older-versions-of-safari]]
+== OpenShift web console does not work with older versions of Safari
+
+`minishift console` does not work on older versions of Safari web browser such as version 10.1.2 (12603.3.8).
+Attempting to access the web console results in the following error:
+
+----
+Error unable to load details about the server
+----
+
+Retry after updating it to the latest version or use Firefox or Chrome browser for this.
+Version 11.0.3 (13604.5.6) has been tested and works with OpenShift web console.
+You can use `minishift console --url` to get the web console URL.
+


docs/source/troubleshooting/troubleshooting-getting-started.adoc
```diff
@@ -138,3 +138,14 @@ Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/mini
  65 B / 65 B [===================================================================================================================================================] 100.00% 0s
 Update failed: open /usr/bin/.minishift.new: permission denied
 ----
+
+[[safari-openshift-web-console-issue]]
+== Safari browser issue with OpenShift web console
+
+In some versions of Safari web browser e.g. version 10.1.2 (12603.3.8), OpenShift web console URL i.e. `minishift console` gives following error:
```
praveenkumar: Instead of `In some versions` => `In older versions of`

praveenkumar: Instead of In some versions => In older versions of

docs/source/troubleshooting/troubleshooting-getting-started.adoc
```diff
@@ -138,3 +138,14 @@ Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/mini
  65 B / 65 B [===================================================================================================================================================] 100.00% 0s
 Update failed: open /usr/bin/.minishift.new: permission denied
 ----
+
+[[safari-openshift-web-console-issue]]
+== Safari browser issue with OpenShift web console
+
+OpenShift web console URL i.e. `minishift console` does not work on older versions of Safari web browser e.g. version 10.1.2 (12603.3.8) and its gives following error:
```
anjannath: Typo s/its/it ..

docs/source/troubleshooting/troubleshooting-getting-started.adoc
```diff
@@ -138,3 +138,14 @@ Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/mini
...
+
+----
+Error unable to load details about the server
+----
+
+Re-try after updating it to the latest version or use Firefox or Chrome browser for this. You can use `minishift console --url` to get the web console URL.
```
budhrg: It would be good if we specify the version which works here.

```
Re-try after updating it to the latest version like bla bla or use Firefox. ...
```

docs/source/troubleshooting/troubleshooting-getting-started.adoc
```diff
@@ -138,3 +138,14 @@ Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/mini
  65 B / 65 B [===================================================================================================================================================] 100.00% 0s
 Update failed: open /usr/bin/.minishift.new: permission denied
 ----
+
+[[safari-openshift-web-console-issue]]
+== Safari browser issue with OpenShift web console
```
robin-owen: We can be more specific and say that the OpenShift web console does not work with some versions of Safari here. "Safari browser issue" is very vague.

docs/source/troubleshooting/troubleshooting-getting-started.adoc
robin-owen: The use of i.e. and e.g. here confuses the sentence rather than clarifying it. We can try rewriting it with something along these lines:

```adoc
The OpenShift web console does not work on older versions of the Safari web browser such as version 10.1.2 (12603.3.8).
Attempting to access the web console results in the following error:
```

docs/source/troubleshooting/troubleshooting-getting-started.adoc
robin-owen: Please put each sentence on its own line. :)

Also, I agree with @budhrg here, and you can simply write "Retry" without the hyphen.

docs/source/troubleshooting/troubleshooting-getting-started.adoc
gbraad: Ins't "Safari browser issue with OpenShift web console" not a little TOO generic? 
s/of Safari/of the Safari
Also, eg, can not be part of the same sentence here.

docs/source/troubleshooting/troubleshooting-getting-started.adoc
gbraad: Agree

LalatenduMohanty:
@robin-owen @gbraad @budhrg Take another look.

docs/source/troubleshooting/troubleshooting-getting-started.adoc
```diff
@@ -138,3 +138,18 @@ Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/mini
  65 B / 65 B [===================================================================================================================================================] 100.00% 0s
 Update failed: open /usr/bin/.minishift.new: permission denied
 ----
+
+[[openshift-web-console-doesnot-work-in-older-safari]]
```
robin-owen: Can we make this ID match the heading better?

docs/source/troubleshooting/troubleshooting-getting-started.adoc
```diff
@@ -138,3 +138,18 @@ Downloading https://github.com/minishift/minishift/releases/download/v1.2.0/mini
...
+== OpenShift web console does not work with older versions of Safari
+
+OpenShift web console URL i.e. `minishift console` does not work on older versions of Safari web browser such as version 10.1.2 (12603.3.8).
```
robin-owen: Using "i.e." here doesn't really make sense. What we really want to say here is something along the lines of "as accessed by", not that the URL *is* ("id est") `minishift console`.

We should also use "the" before "OpenShift web console URL". :)

LalatenduMohanty:
@robin-owen Updated the PR.

Issue #2069, opened 2018-02-27
"minishift openshift version list" should only display released bits from Origin


amitkrout:
### General information

  * Minishift version: 
  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce

  1. Find out all the listed openshift version available (```minishift openshift version list```) 
  2. minishift start --openshift-version v3.8.0-alpha.1
      -- Starting profile 'minishift'
      -- Checking if requested OpenShift version 'v3.8.0-alpha.1' is valid on GitHub ... FAIL
          v3.8.0-alpha.1 is not a valid OpenShift version on GitHub
 
### Expected
In the origin release page https://github.com/openshift/origin/releases, i can see those bits ```v3.9.0-alpha.2, v3.9.0-alpha.1 ...``` are not released however only the tags are created. There should be a proper check in the master that ```minishift openshift version list``` should not display the unreleased origin version.  
```
$ ./minishift openshift version list
The following OpenShift versions are available: 
	- v1.4.1
	- v1.5.0
	- v1.5.1
	- v3.6.0
	- v3.6.1
	- v3.7.0
	- v3.7.1
	- v3.8.0-alpha.0
	- v3.8.0-alpha.1
	- v3.9.0
	- v3.9.0-alpha.0
	- v3.9.0-alpha.1
	- v3.9.0-alpha.2
	- v3.9.0-alpha.3
	- v3.9.0-alpha.4
$ 
```
Only tags are created in release page 
<img width="868" alt="111" src="https://user-images.githubusercontent.com/19851115/36716341-12ee85ea-1bc0-11e8-8bcb-84195257dad7.png">

### Actual
m start failed

### Logs

You can start Minishift with `minishift start --show-libmachine-logs -v5` to collect logs.
Please consider posting this on http://gist.github.com/ and post the link in the issue.

gbraad:
https://github.com/openshift/origin/releases/tag/v3.8.0-alpha.1 contains no published binaries. this is due to the async publishing of docker images and the client binaries,

gbraad:
I suggest to change the error message to a slightly less technical approach:

> "The oc binaries for this release are not available"

vmtyler:
isn't there a way to only list releases that have images and client binaries available?

anjannath:
We could check both docker hub and github releses and then display the list, effectively performing an intersection between the two results.

LalatenduMohanty:
[...]

PR #2120, opened 2018-03-14
Issue #2069 Print origin versions from github releases instead of dockerhub

anjannath:
Fixes: #2069

```
$ minishift openshift version list
	- v1.4.1
	- v1.5.0
	- v1.5.1
	- v3.6.0
	- v3.6.1
	- v3.7.0
	- v3.7.1
	- v3.9.0-alpha.3
```

### DIFF
--- a/pkg/minishift/openshift/version/openshift_versions.go
+++ b/pkg/minishift/openshift/version/openshift_versions.go
@@ -35,18 +35,8 @@ import (
 	"github.com/minishift/minishift/pkg/util"
 )
 
-type ImageTags struct {
+type releaseTags struct {
 	Name string `json:"name"`
-}
-
-type ImageInfo struct {
-	Size         int         `json:"size"`
-	Architecture string      `json:"architecture"`
-	Variant      interface{} `json:"variant"`
-	Features     interface{} `json:"features"`
-	Os           interface{} `json:"os"`
-	OsVersion    interface{} `json:"os_version"`
-	OsFeatures   interface{} `json:"os_features"`
 }
 
 func GetOpenshiftVersion(sshCommander provision.SSHCommander) (string, error) {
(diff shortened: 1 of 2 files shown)


praveenkumar:
@anjannath you don't need to query the docker hub images but only the github releases since all the github releases are surely available to docker hub (or you think there is some mismatch?) It not good to execute 2 different query and then interset them to get same result.

anjannath:
@praveenkumar I don't see any mismatch, but this is to be on the safe side, rather then just assuming they will upload the images. This way we won't ever list any version that is missing either release.

But if you strongly feel it should fetch from only one place, that is okay.. since i don't see any mismatch as of now.

praveenkumar:
> But if you strongly feel it should fetch from only one place, will change it.. since i don't see any mismatch as of now.

@anjannath I will not in future also because whatever GitHub release they do they need to upload image to docker hub otherwise release not even going to work.

anjannath:
@praveenkumar Cool.. Making the change: i also saw that its the same list if only github release is queried.

anjannath:
@praveenkumar Done

budhrg:
@anjannath , watch out for log of travis CI, some valid issue:
```
ok  	github.com/minishift/minishift/pkg/minishift/openshift	0.011s
--- FAIL: TestPrintUpStreamVersions (3.69s)
	Error Trace:	openshift_versions_test.go:42
	Error:      	Received unexpected error:
	            	*json.UnmarshalTypeError
	            	json: cannot unmarshal object into Go value of type []version.releaseTags
	            	&json.UnmarshalTypeError{Value:"object", Type:(*reflect.rtype)(0x7ecd40), Offset:1, Struct:"", Field:""}
	Test:       	TestPrintUpStreamVersions
	Error Trace:	openshift_versions_test.go:49
	Error:      	"" does not contain "v1.4.1"
	Test:       	TestPrintUpStreamVersions
	Error Trace:	openshift_versions_test.go:50
	Error:      	"" does not contain "v3.7.1"
	Test:       	TestPrintUpStreamVersions
FAIL	github.com/minishift/minishift/pkg/minishift/openshift/version	4.232s
```

gbraad:
I do think an intersect is relevant... at the moment there might not be an issue... but it is all about failing early.

If I can detect if GitHub is down, or DockerHub is down I am for this. Because a i/o timeout or a 404 is a different error. And if a timeout occurs, I can accordingly fail beforehand... instead of having to wait until the VM has started and fails to pull the image

anjannath:
@budhrg Yes, it appears in the list because `v3.9.0-aplha.3` is both greater than minimum version and the default version, i have modified the code to handle this. PTAL when you have time.

```
[redhat@localhost minishift]$ minishift openshift version list
	- v1.4.1
	- v1.5.0
	- v1.5.1
	- v3.6.0
	- v3.6.1
	- v3.7.0
	- v3.7.1
```

praveenkumar:
> Pre-release tag - v3.9.0-alpha.3 should not be part of version list output right?

Why not this is a release which have released binary available to github and also images are available to docker hub.

praveenkumar:
> If I can detect if GitHub is down, or DockerHub is down I am for this. Because a i/o timeout or a 404 is a different error. And if a timeout occurs, I can accordingly fail beforehand... instead of having to wait until the VM has started and fails to pull the image

@gbraad Right, this is something we have now part of our pre-fight check right, we do need to update it bit more.

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -104,37 +104,33 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) error
 	return nil
 }
 
+// PrintUpStreamVersions prints the origin versions which satisfies the following conditions:
+// 	1. Major versions greater than or equal to the minimum supported and default version
+//	2. Pre-release versions greater than or equal to the minimum supported version which
+//	   do not have a major version
```
anjannath: @budhrg Will these comments do?

budhrg: @anjannath 
Change `Pre-release versions greater than or equal to the minimum supported version which do not have a major version` to 
```
Pre-release versions greater than default version
```

anjannath: @budhrg Thanks.. that's much simpler :)

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -35,7 +35,7 @@ import (
 	"github.com/minishift/minishift/pkg/util"
 )
 
-type ImageTags struct {
+type releaseTags struct {
```
praveenkumar: @anjannath There is no use of following `ImangeInfo` struct now also. L42-L50 

```golang
 type ImageInfo struct {
 	Size         int         `json:"size"`
 	Architecture string      `json:"architecture"`
 	Variant      interface{} `json:"variant"`
 	Features     interface{} `json:"features"`
 	Os           interface{} `json:"os"`
 	OsVersion    interface{} `json:"os_version"`
 	OsFeatures   interface{} `json:"os_features"`
 }
```

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -104,37 +104,32 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) error
 	return nil
 }
 
+// PrintUpStreamVersions prints the origin versions which satisfies the following conditions:
+// 	1. Major versions greater than or equal to the minimum supported and default version
+//	2. Pre-release versions greater than default version
```
praveenkumar: 👍

pkg/minishift/openshift/version/openshift_versions_test.go
```diff
@@ -101,3 +101,21 @@ func TestIsGreaterOrEqualToBaseVersion(t *testing.T) {
 		}
 	}
 }
+
+func TestIsPrerelease(t *testing.T) {
```
praveenkumar: 👍

pkg/minishift/openshift/version/openshift_versions.go
anjannath: Aha.. Removing it, shouldn't i have got an error while compiling..?  :-/

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -35,20 +35,10 @@ import (
...
-type ImageInfo struct {
-	Size         int         `json:"size"`
-	Architecture string      `json:"architecture"`
-	Variant      interface{} `json:"variant"`
-	Features     interface{} `json:"features"`
-	Os           interface{} `json:"os"`
-	OsVersion    interface{} `json:"os_version"`
-	OsFeatures   interface{} `json:"os_features"`
-}
-
```
praveenkumar: 👍

Issue #1919, opened 2018-01-18
admin-user addon fails if admin user exists on apply


jorgemoralespou:
### General information

  * Minishift version: 1.11
  * OS: macOS
  * Hypervisor: xhyve

### Steps to reproduce

  1. minishift start
  2. minishift login -u admin -p admin
  3. minishift addon apply admin-user
  4. ERROR
![image](https://user-images.githubusercontent.com/78350/35092976-3f8bd128-fc41-11e7-90a2-7584f37bc689.png)

### Expected
Addon applied succesfully

### Actual
Addon is not applied

### Qualification
The addons need to be idempotent, so there needs to be a feature when writing addons that ignores a failure on a command error. I think I already requested this.

e.g.

```
oc create user admin --full-name=admin
```
Will fail on error

```
!oc create user admin --full-name=admin
```

Will not fail on error, and continue.

cc/  @LalatenduMohanty @hferentschik

praveenkumar:
#1130

elyscape:
Another related issue is that `minishift addon remove admin-user` doesn't fully work. It removes the User object but not the Identity object that is created if you log in as that user, resulting in the following:
```shellsession
scapeless:~ elyscape$ minishift addons apply admin-user
-- Applying addon 'admin-user':..

scapeless:~ elyscape$ oc get user admin
NAME      UID                                    FULL NAME   IDENTITIES
admin     df65bdfe-060c-11e8-9aad-8e64ce944971   admin       
scapeless:~ elyscape$ oc login -u admin
Authentication required for https://192.168.64.9:8443 (openshift)
Username: admin
Password: 
Login successful.

scapeless:~ elyscape$ oc get user admin
NAME      UID                                    FULL NAME   IDENTITIES
admin     df65bdfe-060c-11e8-9aad-8e64ce944971   admin       anypassword:admin
scapeless:~ elyscape$ oc get identity anypassword:admin
NAME                IDP NAME      IDP USER NAME   USER NAME   USER UID
anypassword:admin   anypassword   admin           admin       df65bdfe-060c-11e8-9aad-8e64ce944971
scapeless:~ elyscape$ minishift addons remove admin-user
-- Removing addon 'admin-user':.
admin user deleted

scapeless:~ elyscape$ minishift addons apply admin-user
-- Applying addon 'admin-user':..

scapeless:~ elyscape$ oc login -u admin
Authentication required for https://192.168.64.9:8443 (openshift)
Username: admin
Password: 
[...]

Issue #1130, opened 2017-07-13
Add-on command that will skip errors (not fail on errors)


jorgemoralespou:
When developing addons, you might find yourself that you want to install something on a namespace that might or might not exist. Probably many oc commands that you execute you might want to fail or continue.

I want to have a way to express on the addon definition that I don't care about a specific failure, so I can continue processing the addon.

e.g.:

````
oc adm new-project example
# Now if I execute the creation, I don't want to fail
!oc adm new-project example
````

Possibly, you might want to capture the output of the error, depending of it is stdout or stderr both or none, so there could exist four variants of the command:

&oc: shows both, stdout and stderr
1&oc: only shows stdout
2&oc: only shows stderr
!oc: don't show any, but don't fail

This is really helpful for debugging purposes. But I could accept living without this if the output is logged to the minishift.log so it can be later seen, or depends on --loglevel

Of course, notation is suggested. To be discussed when implementing

jorgemoralespou:
Example of where this can be really handy can be found in the [prometheus addon](https://github.com/jorgemoralespou/minishift-addons/tree/master/prometheus).

The addon can be installed in different projects as the project is specified as a parameter to the addon, in the form of --addon-env namespace=<my-namespace>

When you choose a namespace that exists, you don't need to create it, but when you use a non-existent namespace, you need to create it. It would be helpful to use "oc new-project" or "oc adm new-project" with the passed namespace, and in case it does exist not fail, and continue.

hferentschik:
Interesting. What is Prometheus anyways? Never heard.

jorgemoralespou:
@hferentschik are you kidding? Metrics, to be the default in openshift soon, replacing hawkular. Although that's not the point. This applies to multiple add-ons.

hferentschik:
> Metrics, to be the default in openshift soon, replacing hawkular.

That's sounds good. Hawkular never really worked for me, even when it "properly" provisioned.

jorgemoralespou:
@LalatenduMohanty This is preventing from creating addons. I have some addons in the pipeline that can't work as can't be run twice and safe removal is not an option.

e.g. Validate that kube-system namespace is there, if not create it, if there is, use it. I can't just delete that namespace on removal, so the namespace might or might not exist. I would hope this is done soon. It's been delayed for more than  months and prevents addons authoring.

praveenkumar:
[...]

PR #2130, opened 2018-03-15
Issue #1130 Add placeholder to ignore command exit status for addon

praveenkumar:
To test this PR.

1. create a test addon or use existing addon (admin-user)

```
$ minishift apply addon admin-user

# Again run same apply command to make this admin-user addon to fail
$ ./minishift addon apply admin-user 
-- Applying addon 'admin-user':.Error from server (AlreadyExists): users.user.openshift.io "admin" already exists
Error applying the add-on: Error executing command 'oc create user admin --full-name=admin'.
$ echo $?
1

# Now add ignore error placeholder to command
$ cat ~/.minishift/addons/admin-user/admin-user.addon
# Name: admin-user
# Maintainer: kumarpraveen.nitdgp@gmail.com

!oc create user admin --full-name=admin
oc adm policy add-cluster-role-to-user cluster-admin admin

echo "Hello world"

$ ./minishift addon apply admin-user 
-- Applying addon 'admin-user':.Error from server (AlreadyExists): users.user.openshift.io "admin" already exists
.
"Hello world"
$ echo $?
0
```

### DIFF
--- a/docs/source/using/addons.adoc
+++ b/docs/source/using/addons.adoc
@@ -38,7 +38,9 @@ oc adm policy add-scc-to-group anyuid system:authenticated                      
 
 [NOTE]
 ====
-Comment lines, starting with the '#' character, can be inserted anywhere in the file.
+- Comment lines, starting with the '#' character, can be inserted anywhere in the file.
+- Command starting with `!` character ignores the execution failure.
+With the help of this, add-on will be idempotent i.e. a command can be executed multiple times without changing the final behavior of the add-on.
 ====
 
 Enabled add-ons are applied during xref:../command-ref/minishift_start.adoc#[`minishift start`], immediately after the initial cluster provisioning is successfully completed.
--- a/pkg/minishift/addon/command/command.go
+++ b/pkg/minishift/addon/command/command.go
@@ -26,17 +26,21 @@ type Command interface {
 	String() string
 }
 
-type doExecute func(ec *ExecutionContext) error
+type doExecute func(ec *ExecutionContext, ignoreError bool) error
 
 type defaultCommand struct {
 	Command
 
-	rawCommand string
-	fn         doExecute
+	rawCommand  string
+	fn          doExecute
+	ignoreError bool
 }
 
 func (c *defaultCommand) Execute(ec *ExecutionContext) error {
-	err := c.fn(ec)
+	err := c.fn(ec, c.ignoreError)
+	if c.ignoreError {
+		return nil
+	}
 	if err != nil {
 		return err
 	}
@@ -45,4 +49,8 @@ func (c *defaultCommand) Execute(ec *ExecutionContext) error {
 
 func (c *defaultCommand) String() string {
 	return c.rawCommand
+}
+
+func (c *defaultCommand) IgnoreError() bool {
+	return c.ignoreError
 }
--- a/pkg/minishift/addon/command/oc_command.go
+++ b/pkg/minishift/addon/command/oc_command.go
@@ -28,20 +28,24 @@ type OcCommand struct {
 	*defaultCommand
 }
 
-func NewOcCommand(command string) *OcCommand {
-	defaultCommand := &defaultCommand{rawCommand: command}
+func NewOcCommand(command string, ignoreError bool) *OcCommand {
+	defaultCommand := &defaultCommand{rawCommand: command, ignoreError: ignoreError}
 	ocCommand := &OcCommand{defaultCommand}
 	defaultCommand.fn = ocCommand.doExecute
 	return ocCommand
 }
 
-func (c *OcCommand) doExecute(ec *ExecutionContext) error {
+func (c *OcCommand) doExecute(ec *ExecutionContext, ignoreError bool) error {
 	// split off the actual 'oc' command. We are using our cached oc version to run oc commands
 	cmd := strings.Replace(c.rawCommand, "oc ", "", 1)
 	cmd = ec.Interpolate(cmd)
 	fmt.Print(".")
 
 	commander := ec.GetOcCommander()
+	if ignoreError {
+		commander.Run(ec.Interpolate(cmd), ioutil.Discard, ioutil.Discard)
+		return nil
+	}
 	exitStatus := commander.Run(ec.Interpolate(cmd), ioutil.Discard, os.Stdin)
 	if exitStatus != 0 {
 		return errors.New(fmt.Sprintf("Error executing command '%s'.", c.String()))
--- a/pkg/minishift/addon/command/docker_command.go
+++ b/pkg/minishift/addon/command/docker_command.go
@@ -25,14 +25,14 @@ type DockerCommand struct {
 	*defaultCommand
 }
 
-func NewDockerCommand(command string) *DockerCommand {
-	defaultCommand := &defaultCommand{rawCommand: command}
+func NewDockerCommand(command string, ignoreError bool) *DockerCommand {
+	defaultCommand := &defaultCommand{rawCommand: command, ignoreError: ignoreError}
 	dockerCommand := &DockerCommand{defaultCommand}
 	defaultCommand.fn = dockerCommand.doExecute
 	return dockerCommand
 }
 
-func (c *DockerCommand) doExecute(ec *ExecutionContext) error {
+func (c *DockerCommand) doExecute(ec *ExecutionContext, ignoreError bool) error {
 	commander := ec.GetDockerCommander()
 	cmd := ec.Interpolate(c.rawCommand)
 	fmt.Print(".")
--- a/pkg/minishift/addon/command/echo_command.go
+++ b/pkg/minishift/addon/command/echo_command.go
@@ -26,14 +26,14 @@ type EchoCommand struct {
 	*defaultCommand
 }
 
-func NewEchoCommand(command string) *EchoCommand {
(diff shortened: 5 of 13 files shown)


praveenkumar:
@jorgemoralespou Can you take the artifact from https://ci.appveyor.com/project/minishift-bot/minishift/build/2451/artifacts and let us know if that works as expected for you? I went with your suggestion `!` as ignoreError character  :)

pkg/minishift/addon/command/command.go
```diff
@@ -31,12 +31,16 @@ type doExecute func(ec *ExecutionContext) error
...
-	fn         doExecute
+	rawCommand  string
+	fn          doExecute
+	ignoreError bool
 }
 
 func (c *defaultCommand) Execute(ec *ExecutionContext) error {
 	err := c.fn(ec)
+	if c.ignoreError {
+		return nil
```
gbraad: this is the whole change... ;-)
the rest is related to passing the variable and tests

gbraad (approved):

gbraad:
however... no tests to verify IgnoreError?

pkg/minishift/addon/command/command.go
budhrg: :) ya

docs/source/using/addons.adoc
```diff
@@ -109,7 +109,8 @@ This can be useful in cases where you know that a command such as `oc` might tak
 
 [NOTE]
 ====
-Trying to use an undefined command will cause an error when the add-on gets parsed.
+- Trying to use an undefined command will cause an error when the add-on gets parsed.
+- Exit status of a Command start with `!` is ignored.
```
budhrg: @praveenkumar I didn't get this line. Could you elaborate?

May be it should be written as
```
- Failure of command starting with `!` is ignored 
```

brgnepal (dismissed):
This change is not visible. 

I would recommand to create a section as `Special characters` and describe `#` and `!` there.
I felt above not sure what others @minishift/minishift-dev think of.

docs/source/using/addons.adoc
budhrg: Below section I feel should come after `Built-in Variables` section.
```
=== Special characters

Add-on file contains some special character that has special meaning. 
Below are few special characters which one can use to have special behavior:
- `#` : This character is reserved for comment character.
- `!` : When used at the beginning of any command, the failure of the command will be ignored and will not exit the addon execution
```
Somewhere on the similar lines.

@robin-owen could you help here to improve it?

jorgemoralespou:
@paveenkumar I've tried it and the fact that it prints the error is annoying. When I saw the error I didn't even tried to see if it worked. I was going to comment here as a not working when I realized in your example you also show the error, but it's working :-(
Also, there's a **.** after the addon applying message but in a new line, which also looks odd.

I would just not show errors, as that's what this command was meant for, and if there's an error, log it through logger ad info/warn level.

docs/source/using/addons.adoc
robin-owen: I think a section is certainly appropriate, but I think placing it directly before Variable Interpolation makes the most sense. The special characters we're talking about here are quite relevant to the available commands, and having them documented in between documentation about variables feels out of place to me.

As for the wording itself, it could use some improvement. Once we have the placement worked out, I can send a PR to praveenkumar:issue_1130 to assist with the writing. :)

praveenkumar: @budhrg @robin-owen problem is `#` is also used as part of metadata definition for so might be we need rewording around it, Meanwhile, I added `- Execution failure of command starting with `!` is ignored` which at this moment feel more appropriate to me.

pkg/minishift/addon/command/oc_command.go
```diff
@@ -28,20 +28,24 @@ type OcCommand struct {
...
-func (c *OcCommand) doExecute(ec *ExecutionContext) error {
+func (c *OcCommand) doExecute(ec *ExecutionContext, ignoreError bool) error {
 	// split off the actual 'oc' command. We are using our cached oc version to run oc commands
 	cmd := strings.Replace(c.rawCommand, "oc ", "", 1)
 	cmd = ec.Interpolate(cmd)
 	fmt.Print(".")
 
 	commander := ec.GetOcCommander()
+	if ignoreError {
+		commander.Run(ec.Interpolate(cmd), ioutil.Discard, ioutil.Discard)
```
anjannath: Do we want the error to be shown if run with logging enabled?

jorgemoralespou:
I haven't tested an update on this, but I would think that the error needs to be loved and an indication need to make it an info, and not an error.

praveenkumar:
@jorgemoralespou Please take the artifacts from http://artifacts.ci.centos.org/minishift/minishift/pr/2130/ and now we just ignored the error and also don't put it to stdout so now if run with ignore error placeholder you will not see any kind of error/info/warning message for specific command.

pkg/minishift/addon/command/oc_command.go
praveenkumar: @anjannath this complete command interface stuff doesn't have any logger, that's another reason we are sending stdin/stderr to `ioutil.Discard`.

anjannath: I see.. maybe we can do something about it after #169 is resolved

anjannath (approved):
:+1:

docs/source/using/addons.adoc
budhrg: As I said documentation for `!` is just a line and easily not visible :)

budhrg: WDYT @LalatenduMohanty ?

LalatenduMohanty (changes requested):
@praveenkumar I think you should fix https://github.com/minishift/minishift/issues/1919 along with this PR.

docs/source/using/addons.adoc
```diff
@@ -109,7 +109,8 @@ This can be useful in cases where you know that a command such as `oc` might tak
 
 [NOTE]
 ====
-Trying to use an undefined command will cause an error when the add-on gets parsed.
+- Trying to use an undefined command will cause an error when the add-on gets parsed.
+- Execution failure of command starting with `!` is ignored
```
LalatenduMohanty: Two comments
1. The description of `!` is not enough. Adding `!` makes a command execution idempotent i.e. a command can be executed multiple times without changing the final behavior of the add-on. 

2. The note should be with `Comment lines, starting with the '#' character, can be inserted anywhere in the file.`. Line number 42

LalatenduMohanty (changes requested):

docs/source/using/addons.adoc
LalatenduMohanty: @budhrg @robin-owen I am fine with a separate section too. But may be once we add more syntax in add-on.

budhrg: > The note should be with Comment lines, starting with the '#' character, can be inserted anywhere in the file.. Line number 42

Let's add the `!` note here as it is more visible than the current form. 

> I am fine with a separate section too. But may be once we add more syntax in add-on.

Sure.

docs/source/using/addons.adoc
```diff
@@ -38,7 +38,8 @@ oc adm policy add-scc-to-group anyuid system:authenticated
 
 [NOTE]
 ====
-Comment lines, starting with the '#' character, can be inserted anywhere in the file.
+- Comment lines, starting with the '#' character, can be inserted anywhere in the file.
+- Commands start with `!` character  makes execution idempotent i.e. a command can be executed multiple times without changing the final behavior of the add-on.
```
budhrg: `Command starting with !...`

LalatenduMohanty: @praveenkumar  I did not want you to remove the information about it ignores the command failure.  I should have been more clear. Sorry for the confusion 
```
Commands starting with `!` character  ignores the execution failure. With the help of this command execution in add-on will be idempotent i.e. a command can be executed multiple times without changing the final behavior of the add-on.
```

LalatenduMohanty (changes requested):

praveenkumar:
> I think you should fix #1919 along with this PR.

@LalatenduMohanty I will create another PR for this issue it needs to update both addon and remove part of admin-user. I want this to be in first and then send another.

pkg/minishift/addon/command/oc_command.go
LalatenduMohanty: @praveenkumar You could have ignored the exitStatus too instead of calling the function with ioutil.Discard for stderr. Why did you choose to call the `Run` separately?

praveenkumar: @LalatenduMohanty If you not call Run then command execution doesn't going to happen, also ignoring exitStatus mean we are just ignoring the command exit status but if you look how `Run` works then it put stderr to stdin so only way to make it work the way @jorgemoralespou wants to redirect stderr => dev/null which ioutil.Discard do.

LalatenduMohanty: Right, I modified the code and saw this
```
$ minishift addon apply admin-user
-- Applying addon 'admin-user':.Error from server (AlreadyExists): users.user.openshift.io "admin" already exists
.
```

LalatenduMohanty (approved):

thomasmoree:
Follow the journey of [AI Ideology](https://aiideology.com/) in exploring the evolving relationship between humanity and artificial intelligence.

PR #2148, opened 2018-03-18
[WIP] Issue #2141 Remove vmwarefusion deprecation code

### LINKED ISSUE #2141: Remove vmwarefusion depreciation code
We have already removed the `vmwarefusion` from all of our documentation with https://github.com/minishift/minishift/issues/1602.

I think we need to remove all code specific to `vmwarefusion` from our code base.
See https://github.com/minishift/minishift/search?utf8=%E2%9C%93&q=vmwarefusion&type=

tylerauerbeck:
Fixes #2141 

To cleanup Gopkg.lock, I also ran `dep ensure -update`, which appears to have updated some packages. Let me know if this isn't desired and I can back this out.

### DIFF
--- a/pkg/minikube/cluster/cluster.go
+++ b/pkg/minikube/cluster/cluster.go
@@ -368,9 +368,6 @@ func getDriverOptions(config MachineConfig) interface{} {
 	switch config.VMDriver {
 	case "virtualbox":
 		driver = createVirtualboxHost(config)
-	case "vmwarefusion":
-		fmt.Println("VMWare Fusion driver will be deprecated soon. Please consider using other drivers.")
-		driver = createVMwareFusionHost(config)
 	case "kvm":
 		driver = createKVMHost(config)
 	case "xhyve":
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -133,7 +133,7 @@
     "pkg/system",
     "pkg/truncindex"
   ]
-  revision = "f5dc7fe58d10b581aee6cbf400ab69d2f6b86ab6"
+  revision = "aaa10cba3c246a9186e2160f4e12b21d8cf77c94"
 
 [[projects]]
   name = "github.com/cpuguy83/go-md2man"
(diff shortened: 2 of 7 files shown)


pkg/minikube/cluster/cluster.go
```diff
@@ -368,9 +368,6 @@ func getDriverOptions(config MachineConfig) interface{} {
 	switch config.VMDriver {
 	case "virtualbox":
 		driver = createVirtualboxHost(config)
-	case "vmwarefusion":
-		fmt.Println("VMWare Fusion driver will be deprecated soon. Please consider using other drivers.")
-		driver = createVMwareFusionHost(config)
```
praveenkumar: That means we are not even supporting vmwarefusion :(, Recently https://github.com/machine-drivers/docker-machine-driver-vmware is already added and can be used.

gbraad:
We have to see if @anfarnee can provide an update PR to support the new
driver. Or can we make this work?

-- 

   [ Doing Open Source Matters ]

pkg/minikube/cluster/cluster.go
tylerauerbeck: @praveenkumar Ah. I may have misunderstood the issue then. So we were just looking to remove the deprecation warnings themselves and not the different cases / functions that implemented vmwarefusion?

If that's the case I can roll back those changes. But has the updated driver been implemented already?

gbraad:
The orginal issue was related to removing the driver support. But in the
time we discussed this, we also discussed with minikube and someone at
VMware. We were going to remove it, but since we have a supported driver,
we are currently investigating in adding the support back in (modified).

-- 

   [ Doing Open Source Matters ]

LalatenduMohanty:
@minishift/minishift-dev @tylerauerbeck I think we can close the issue #2141 without any fix as we will try to add VMWare Fusion support again. WDYT?

praveenkumar:
I do agree, let's close it and create a new one to do a POC  on updated VMware driver side.

budhrg:
+1 for closing.

tylerauerbeck:
+1. Go ahead and close it.

budhrg:
@tylerauerbeck  Thanks again for the PR. Really appreciate. 
Closing now in the interest of supporting vmwarefusion in near future.

Issue #2125, opened 2018-03-14
Add port verification for hostfolders-sftp-port config option


hferentschik:
SSHFS host-folders introduces `hostfolders-sftp-port` as a persistent config option to set the port for the sftp server. At the moment, there is only an `isPositive` validation configured in `config.go`.  We should create a port specific validation which verifies a that the option is in a valid port range.

gbraad:
this should check a defined range of ports which are allowed, eg. 1024-65536 so that we can tell user at early stage about which port should be used for config.

PR #2152, opened 2018-03-19
Issue #2125 Validation for `hostfolders-sftp-port` config option

anjannath:
Fixes: #2125

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -95,7 +95,7 @@ var (
 	// Host Folders
 	HostFoldersMountPath = createConfigSetting("hostfolders-mountpath", SetString, nil, nil, true, nil)
 	HostFoldersAutoMount = createConfigSetting("hostfolders-automount", SetBool, nil, nil, true, nil)
-	HostFoldersSftpPort  = createConfigSetting("hostfolders-sftp-port", SetInt, []setFn{validations.IsPositive}, nil, true, nil)
+	HostFoldersSftpPort  = createConfigSetting("hostfolders-sftp-port", SetInt, []setFn{validations.IsValidPort}, nil, true, nil)
 
 	// Image caching
 	ImageCaching = createConfigSetting("image-caching", SetBool, nil, nil, true, true)
--- a/pkg/minishift/config/validations.go
+++ b/pkg/minishift/config/validations.go
@@ -187,3 +187,31 @@ func IsValidNetmask(name string, mask string) error {
 
 	return nil
 }
+
+func IsValidPort(name string, p string) error {
+	port, err := strconv.Atoi(p)
+	if err != nil {
+		return fmt.Errorf("%s:%v", name, err)
+	}
+	if numInRange(port, 1024, 65536) && isPortAccessible(port) {
+		return nil
+	}
+	return fmt.Errorf("Port %d is inaccessible please use a port in the range (1024-65536)", port)
+}
+
+func numInRange(num int, start int, end int) bool {
+	if num >= start && num <= end {
+		return true
+	}
+	return false
+}
+
+func isPortAccessible(port int) bool {
+	hostWithPort := fmt.Sprintf(":%d", port)
+	listener, err := net.Listen("tcp", hostWithPort)
+	if err != nil {
+		return false
+	}
+	listener.Close()
+	return true
+}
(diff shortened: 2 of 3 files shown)


cmd/minishift/cmd/config/config.go
```diff
@@ -95,7 +95,7 @@ var (
 	// Host Folders
 	HostFoldersMountPath = createConfigSetting("hostfolders-mountpath", SetString, nil, nil, true, nil)
 	HostFoldersAutoMount = createConfigSetting("hostfolders-automount", SetBool, nil, nil, true, nil)
-	HostFoldersSftpPort  = createConfigSetting("hostfolders-sftp-port", SetInt, []setFn{validations.IsPositive}, nil, true, nil)
+	HostFoldersSftpPort  = createConfigSetting("hostfolders-sftp-port", SetInt, []setFn{validations.IsPositive, validations.IsValidPort}, nil, true, nil)
```
praveenkumar: We don't need `validations.IsPositive` anymore for port validations.

pkg/minishift/config/validations.go
```diff
@@ -187,3 +187,35 @@ func IsValidNetmask(name string, mask string) error {
...
+func numInRange(num int, start int, end int) bool {
+	if num >= start && num <= end {
+		return true
+	}
+	return false
+}
+
+func isPortAccessible(port int) error {
+	hostWithPort := fmt.Sprintf(":%d", port)
+	lnr, err := net.Listen("tcp", hostWithPort)
```
praveenkumar: What is `lnr` instead of this can we use `conn` (sort form of connection)?

praveenkumar: Also does this block if port is blocked by firewalld, issue which @gbraad was facing that time?

anjannath: `lnr` was meant to be listener ;-), because it returns a listener. I guess if it is blocked by firewalld it will be able to still bind to the port, as host is localhost, but i'll verify this behavior.

pkg/minishift/config/validations.go
```diff
@@ -187,3 +187,35 @@ func IsValidNetmask(name string, mask string) error {
...
+}
+
+func numInRange(num int, start int, end int) bool {
+	if num >= start && num <= end {
+		return true
+	}
+	return false
+}
+
+func isPortAccessible(port int) error {
```
budhrg: The first thing when I see this type of function `isPortAccessible`, I except bool to be returned.
Like https://github.com/minishift/minishift/blob/master/pkg/util/utils.go#L187

Please make the changes.

gbraad: bool expected...

anjannath: Yeah that is true, but i made that choice because i wanted to print the error, if it failed so needed to return it to the upper level. But the error is of only one type here, which is bind error so, I am okay with returning a bool. Making the change. :)

anjannath: Yes.. making the change

pkg/minishift/config/validations.go
```diff
@@ -187,3 +187,31 @@ func IsValidNetmask(name string, mask string) error {
...
+}
+
+func isPortAccessible(port int) bool {
+	hostWithPort := fmt.Sprintf(":%d", port)
+	listener, err := net.Listen("tcp", hostWithPort)
+	if err != nil {
+		return false
+	}
+	listener.Close()
+	return true
```
LalatenduMohanty: I am little skeptical about this function. I am not sure if we need this.

anjannath:
@lalatenduM We are checking if the port is free, so that `sftp` server can bind to the port. that being said, when the server actually tries to bind this is handled, i didn't think about that earlier, now when you mentioned it came to my mind, we still want to fail 
 earlier.. as in the case of `minishift config set` this would help. So i say lets keep it.

gbraad:
Approved. LGTM

-- 

   [ Doing Open Source Matters ]

pkg/minishift/config/validations.go
gbraad: Yes, we need... As user is allowed to change the port. And no feedback is given during mount if start of sftpd would fail. So early fail is a better approach until we can handle async feedback (if ever)

gbraad:
Please rebase

anjannath:
Rebased, but travis is failing.. :-(

Issue #2150, opened 2018-03-19
Disk validation fails in cmd-config.feature on MacOS


agajdosi:
Probably caused by #2097, feature file needs a small update. The error log:

```
    Then stdout should match "Disk \/dev\/sda: 2[4-6]\.?[0-9]{0,2} (GB|GiB)" # testsuite.go:565 -> /testsuite.commandReturnShouldMatchRegex
    Output did not match. Expected: 'Disk \/dev\/sda: 2[4-6]\.?[0-9]{0,2} (GB|GiB)', Actual: 'Disk /dev/vda: 26.2 GB, 26214400000 bytes, 51200000 sectors
Disk label type: dos
Disk identifier: 0x6a5948ab
Disk /dev/loop0: 0 MB, 16384 bytes, 32 sectors
Disk /dev/loop1: 4 MB, 4304896 bytes, 8408 sectors
Disk /dev/loop2: 292 MB, 292593664 bytes, 571472 sectors
Disk /dev/loop3: 10.7 GB, 10737418240 bytes, 20971520 sectors
Disk /dev/loop4: 536 MB, 536870912 bytes, 1048576 sectors
Disk /dev/mapper/live-rw: 10.7 GB, 10737418240 bytes, 20971520 sectors
Disk /dev/mapper/live-base: 10.7 GB, 10737418240 bytes, 20971520 sectors
Disk /dev/mapper/live-osimg-min: 10.7 GB, 10737418240 bytes, 20971520 sectors
'
```

I am checking the rest of feature files if there are not other issues.

agajdosi:
Resolved with #2154.

PR #2154, opened 2018-03-19
Issue #2150 Disk validation fails in cmd-config.feature on MacOS

agajdosi:
Also updating Godog to latest release which fixes incomplete failure report when scenario outlines are used.

### DIFF
--- a/test/integration/features/cmd-config.feature
+++ b/test/integration/features/cmd-config.feature
@@ -213,7 +213,7 @@ user defined options which changes default behaviour of Minishift.
 
   Scenario: Checking that disk-size value was applied
      When executing "minishift ssh -- sudo fdisk -l | grep Disk" succeeds
-     Then stdout should match "Disk \/dev\/sda: 2[4-6]\.?[0-9]{0,2} (GB|GiB)"
+     Then stdout should match "Disk \/dev\/.da: 2[4-6]\.?[0-9]{0,2} (GB|GiB)"
 
   Scenario: Checking that cpus value was applied
      When executing "minishift ssh -- cat /proc/cpuinfo" succeeds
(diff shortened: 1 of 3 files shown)


test/integration/features/cmd-config.feature
```diff
@@ -213,7 +213,7 @@ user defined options which changes default behaviour of Minishift.
 
   Scenario: Checking that disk-size value was applied
      When executing "minishift ssh -- sudo fdisk -l | grep Disk" succeeds
-     Then stdout should match "Disk \/dev\/sda: 2[4-6]\.?[0-9]{0,2} (GB|GiB)"
+     Then stdout should match "Disk \/dev\/(s|v)da: 2[4-6]\.?[0-9]{0,2} (GB|GiB)"
```
praveenkumar: @agajdosi Can't we use `?da` as regex where we don't care what is the first char? `?da` is what we are using as part of disk check for preflight checks.

budhrg: Yes `?da` is better choice +1

agajdosi: @praveenkumar @budhrg Good point, missed that. Updating now.

test/integration/features/cmd-config.feature
```diff
@@ -213,7 +213,7 @@ user defined options which changes default behaviour of Minishift.
 
   Scenario: Checking that disk-size value was applied
      When executing "minishift ssh -- sudo fdisk -l | grep Disk" succeeds
-     Then stdout should match "Disk \/dev\/sda: 2[4-6]\.?[0-9]{0,2} (GB|GiB)"
+     Then stdout should match "Disk \/dev\/.da: 2[4-6]\.?[0-9]{0,2} (GB|GiB)"
```
praveenkumar: 👍

budhrg:
LGTM :+1:

Issue #2155, opened 2018-03-20
Update docker-machine version from 0.13.0 to 0.15.0


praveenkumar:
Recently docker-machine released 0.15.0 which have some improvement around hyperV side. we need to update to latest version and check if there is any conflict.

- https://github.com/docker/machine/releases/tag/v0.15.0

anjannath:
I guess this is still to do..

praveenkumar:
1.14.0 have an issue with `hyperV` so either we use the revision which has a fix for this or waits for 1.14.1 release. check https://github.com/docker/machine/commit/c79bd4ad7404b3c17ff3fd14991dc3f4cddf792e

gbraad:
Fix will never come as a release I guess:
https://github.com/docker/machine/issues/4424#issuecomment-393414375

LalatenduMohanty:
Resolved via #2516

PR #2159, opened 2018-03-21
[WIP]Issue #2155 Update docker-machine from 0.13.0 to 0.14.0

praveenkumar:


Gopkg.toml
```diff
@@ -117,11 +117,6 @@ ignored = ["github.com/Sirupsen/logrus"]
   name = "github.com/docker/go-connections"
   revision = "3ede32e2033de7505e6500d6c868c2b9ed9f169d"
 
-# Get dependencies for hyperkit
-[[constraint]]
-  name = "github.com/machine-driver/docker-machine-driver-hyperkit"
```
anjannath: Why do we not need this anymore? Did we needed this earlier?

gbraad:
image

Issues with the new libmachine; it seems a check they introduced conflicts for me
"Hyper-V PowerShell Module is not available"
This check was introduced to prevent a conflict with the get-vm module for VMware

gbraad:
Not much difference in speed. It still seems to use a fixed disksize for the initial disk creation (for adding userdata), after this it needs to convert... which still takes the same amount of time

Gopkg.lock
```diff
@@ -399,15 +386,6 @@
...
   revision = "2788f0dbd16903de03cb8186e5c7d97b69ad387b"
 
-[[projects]]
-  name = "github.com/machine-drivers/docker-machine-driver-hyperkit"
-  packages = [
-    "pkg/drivers",
-    "pkg/hyperkit"
-  ]
-  revision = "88bae774eacefa283ef549f6ea6bc202d97ca07a"
-  version = "v1.0.0"
```
LalatenduMohanty: @praveenkumar why are we removing this?

praveenkumar: @LalatenduMohanty not needed anymore since we are adding directly the driver struct to codebase same as we add for xhyve and others.

gbraad:
it seems Minikube has a similar problem... they updated in December, https://github.com/kubernetes/minikube/commit/ee00eda89fbddaee6ffc2510ec777a661782298d#diff-cba8b87c461928ce65e3c6c4a7147d71 but it is hard to track which version they actually have.

praveenkumar:
> it seems Minikube has a similar problem... they updated in December, kubernetes/minikube@ee00eda#diff-cba8b87c461928ce65e3c6c4a7147d71 but it is hard to track which version they actually have.

@gbraad right, they updated on the December which is way before the 0.14.0 release but might be hyperV  stuff was in that time.

> At least I noticed they have the issue with the PowerShell module names: kubernetes/minikube#2639

Right.

praveenkumar:
Closing this PR and going to send 2 separate one for it.

PR #2160, opened 2018-03-21
Issue #1919 Make admin-user addon as idempotent

### LINKED ISSUE #1919: admin-user addon fails if admin user exists on apply

praveenkumar:
Fix #1919

### DIFF
--- a/addons/admin-user/admin-user.addon
+++ b/addons/admin-user/admin-user.addon
@@ -1,5 +1,5 @@
 # Name: admin-user
 # Description: Create admin user and assign the cluster-admin role to it.
 
-oc create user admin --full-name=admin
+!oc create user admin --full-name=admin
 oc adm policy add-cluster-role-to-user cluster-admin admin
--- a/addons/admin-user/admin-user.addon.remove
+++ b/addons/admin-user/admin-user.addon.remove
@@ -2,4 +2,5 @@
 # Description: Remove admin user addon
 
 oc delete user admin
+!oc delete identity anypassword:admin
 echo admin user deleted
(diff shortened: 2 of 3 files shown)


LalatenduMohanty:
Thanks.

jorgemoralespou:
@LalatenduMohanty will this add-on work with previous releases of minishift? Are we confident that because the addon is packaged with minishift there will be no issues? What about the non default addons? What's the strategy we'll recommend with those?

praveenkumar:
> will this add-on work with previous releases of minishift? 

@jorgemoralespou Previous released version have different backed addon which works as expected.

> Are we confident that because the addon is packaged with minishift there will be no issues? 

Package version of addon doesn't have any issue.

> What about the non default addons? What's the strategy we'll recommend with those?

We didn't update any non-default addons yet but in future, if we use this capability then somehow need to mention which version of minishift required for specific addon. I think we do have an issue around this and something we should add soon as part of addon metadata.

Issue #2140, opened 2018-03-16
3.9 will break registry-route addon


jorgemoralespou:
```
$ minishift addon apply registry-route
-- Applying addon 'registry-route':.Error applying the add-on: Error executing command 'ssh command error:
command : docker exec -t origin /usr/bin/openshift admin ca create-server-cert --signer-cert=/var/lib/origin/openshift.local.config/master/ca.crt --signer-key=/var/lib/origin/openshift.local.config/master/ca.key --signer-serial=/var/lib/origin/openshift.local.config/master/ca.serial.txt --hostnames='docker-registry-default.192.168.64.44.nip.io,docker-registry-default.192.168.64.44.nip.io:443,docker-registry.default.svc.cluster.local,172.30.1.1' --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key
err     : exit status 1
output  : Error: unknown command "admin" for "openshift"
Run 'openshift --help' for usage.

$ minishift version
minishift v1.14.0+1ec5877

$ minishift openshift version
openshift v3.9.0-alpha.3+78ddc10
kubernetes v1.9.1+a0ce1bc657
etcd 3.2.8
```

LalatenduMohanty:
@jorgemoralespou Thanks, do you know when it is coming out?

jorgemoralespou:
End of month

anjannath:
I did some poking around and it seems the `openshift admin` command is no there anymore, we have to use `oc adm`, `oc cli`.

We need to run these commands from inside the origin container, right now `oc` is not run inside the origin container AFAIK.

budhrg:
> I did some poking around and it seems the openshift admin command is not there anymore, we have to use oc adm, oc create.

If this is the case, we also need to think about the backward compatibility of the addon as how we are going to support it for version < 3.9.0.

anjannath:
> we also need to think about the backward compatibility of the addon as how we are going to support it for version < 3.9.0

Yes, and also the parts that interact with OpenShift. #2156 is also caused because of this, probably we need to check all the sub-commands of `minishift openshift`.

praveenkumar:
Looks they removed some of the commands from the openshift binary. We need to find an alternative for this now.

```
docker@minishift:~$ docker run --rm -it openshift/origin:v3.9.0-alpha.3 help
OpenShift Application Platform 

The OpenShift distribution of Kubernetes helps you build, deploy, and manage your applications on top of Docker
[...]

PR #2161, opened 2018-03-21
Issue #2140 Add entrypoint to run openshift command for origin container

praveenkumar:
Added WIP just to make sure we test it thoroughly with all the openshift versions.

Related to #2140

### DIFF
--- a/addons/registry-route/registry-route.addon
+++ b/addons/registry-route/registry-route.addon
@@ -1,8 +1,8 @@
 # Name: registry-route
 # Description: Create an edge terminated route for the OpenShift registry
 
-openshift admin ca create-server-cert --signer-cert=/var/lib/origin/openshift.local.config/master/ca.crt --signer-key=/var/lib/origin/openshift.local.config/master/ca.key --signer-serial=/var/lib/origin/openshift.local.config/master/ca.serial.txt --hostnames='docker-registry-default.#{routing-suffix},docker-registry-default.#{routing-suffix}:443,docker-registry.default.svc.cluster.local,172.30.1.1' --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key
-openshift cli create route edge docker-registry --service=docker-registry --ca-cert=/var/lib/origin/openshift.local.config/master/ca.crt --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key --hostname='docker-registry-default.#{routing-suffix}'
+openshift adm ca create-server-cert --signer-cert=/var/lib/origin/openshift.local.config/master/ca.crt --signer-key=/var/lib/origin/openshift.local.config/master/ca.key --signer-serial=/var/lib/origin/openshift.local.config/master/ca.serial.txt --hostnames='docker-registry-default.#{routing-suffix},docker-registry-default.#{routing-suffix}:443,docker-registry.default.svc.cluster.local,172.30.1.1' --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key
+openshift create route edge docker-registry --service=docker-registry --ca-cert=/var/lib/origin/openshift.local.config/master/ca.crt --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key --hostname='docker-registry-default.#{routing-suffix}'
 
 ssh sudo mkdir -p /etc/docker/certs.d/docker-registry-default.#{routing-suffix}:443
 ssh sudo chown docker /etc/docker/certs.d/docker-registry-default.#{routing-suffix}:443
--- a/docs/source/using/addons.adoc
+++ b/docs/source/using/addons.adoc
@@ -92,7 +92,7 @@ The `oc` command is executed as *system:admin*.
 ====
 
 openshift::
-If the add-on command starts with `openshift`, you can run the *oc* binary within the container that runs OpenShift.
+If the add-on command starts with `openshift`, it uses the `oc` binary present in the OpenShift container to execute the command.
 This means that any file parameters or other system-specific parameters must match the environment of the container instead of your host.
 
 docker::
--- a/docs/source/using/addons.adoc
+++ b/docs/source/using/addons.adoc
@@ -92,7 +92,7 @@ The `oc` command is executed as *system:admin*.
 ====
 
 openshift::
-If the add-on command starts with `openshift`, you can run the *openshift* binary within the container that runs OpenShift.
+If the add-on command starts with `openshift`, you can run the *oc* binary within the container that runs OpenShift.
 This means that any file parameters or other system-specific parameters must match the environment of the container instead of your host.
 
 docker::
--- a/pkg/minishift/addon/command/openshift_command.go
+++ b/pkg/minishift/addon/command/openshift_command.go
@@ -36,13 +36,13 @@ func NewOpenShiftCommand(command string, ignoreError bool) *OpenShiftCommand {
 }
 
 func (c *OpenShiftCommand) doExecute(ec *ExecutionContext, ignoreError bool) error {
-	// split off the actual 'oc' command. We are using our cached oc version to run oc commands
+	// split off the actual 'openshift' command. We are using origin container to run those commands
 	cmd := strings.Replace(c.rawCommand, "openshift ", "", 1)
 	cmd = ec.Interpolate(cmd)
 	fmt.Print(".")
 
 	commander := ec.GetDockerCommander()
-	_, err := commander.Exec("-t", constants.OpenshiftContainerName, constants.OpenshiftExec, ec.Interpolate(cmd))
+	_, err := commander.Exec("-t", constants.OpenshiftContainerName, constants.OpenshiftOcExec, ec.Interpolate(cmd))
 	if err != nil {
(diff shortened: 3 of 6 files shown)


praveenkumar:
@anjannath Check the patch apply command also using the artifact of this PR.

anjannath:
Sure @praveenkumar Will test this today.

praveenkumar:
@budhrg you forgot to execute `minishift addon install --defaults` using same artifacts because this PR also contain the change in the registry addon.

budhrg:
The PR is working for me for v3.7.2 and v3.9.0-alpha.3.
Make sure you do 
```
$ make clean; make
```
and perform testing.

anjannath:
@praveenkumar Do we also need to test if other add-ons are working? since the working of the `openshift` command has changed.

praveenkumar:
> Do we also need to test if other add-ons are working? 

@anjannath I did test that and looks like there is no addon in the minishift-addon repo which uses `openshift` command directly, I think there is on but that atm is good.

anjannath:
@praveenkumar Sounds good!

gbraad:
LGTM

pkg/minishift/addon/command/openshift_command.go
```diff
@@ -36,13 +36,13 @@ func NewOpenShiftCommand(command string, ignoreError bool) *OpenShiftCommand {
...
 func (c *OpenShiftCommand) doExecute(ec *ExecutionContext, ignoreError bool) error {
-	// split off the actual 'oc' command. We are using our cached oc version to run oc commands
+	// split off the actual 'openshift' command. We are using origin container to run those commands
 	cmd := strings.Replace(c.rawCommand, "openshift ", "", 1)
 	cmd = ec.Interpolate(cmd)
 	fmt.Print(".")
 
 	commander := ec.GetDockerCommander()
-	_, err := commander.Exec("-t", constants.OpenshiftContainerName, constants.OpenshiftExec, ec.Interpolate(cmd))
+	_, err := commander.Exec("-t", constants.OpenshiftContainerName, constants.OpenshiftOcExec, ec.Interpolate(cmd))
```
anjannath: :+1:

LalatenduMohanty: Why are we using `oc` here? Shouldn't we use `openshift` as it is for `openshift` command in add-on?

pkg/minishift/constants/constants.go
```diff
@@ -22,6 +22,7 @@ const (
 	MinikubeIsoAlias       = "minikube"
 	OpenshiftContainerName = "origin"
 	OpenshiftExec          = "/usr/bin/openshift"
+	OpenshiftOcExec        = "/usr/bin/oc"
```
LalatenduMohanty: Now that we have `oc binary` we should create a follow up issue about which places in the code it should use this `oc binary` and we should fix it.

pkg/minishift/addon/command/openshift_command.go
```diff
@@ -36,13 +36,13 @@ func NewOpenShiftCommand(command string, ignoreError bool) *OpenShiftCommand {
 }
 
 func (c *OpenShiftCommand) doExecute(ec *ExecutionContext, ignoreError bool) error {
-	// split off the actual 'oc' command. We are using our cached oc version to run oc commands
```
gbraad: Why does it show `our cached oc version` but we actually do not run it as such?

pkg/minishift/addon/command/openshift_command.go
praveenkumar: No, all commands which are running inside the container should be part of OpenShift commander and we can separate oc commander (which interact with what we downloaded and cached) and OpenShift commander which interact with origin container.

pkg/minishift/constants/constants.go
praveenkumar: Not sure what you mean here. As I think all the interaction we have with OpenShift commander.

pkg/minishift/addon/command/openshift_command.go
praveenkumar: @gbraad it was old comment which I think during implementation hardy copy pasted same text from oc commander.

gbraad:
I see, the certificates are needed... but the invocation with `openshift` which results in running `oc` in the container is not straightforward at the moment. feels icky.

pkg/minishift/addon/command/openshift_command.go
gbraad: this code needs a quick review overall...

but to continue, let's split out the preconditions to when we can run `openshift` or `oc`, or does running `oc` always works?

pkg/minishift/addon/command/openshift_command.go
LalatenduMohanty: @praveenkumar I do not understand. As per documentation the openshift command should this `If the add-on command starts with openshift, you can run the openshift binary within the container that runs OpenShift. This means that any file parameters or other system-specific parameters must match the environment of the container instead of your host.` And I am not how `oc` can do this?

addons/registry-route/registry-route.addon
```diff
@@ -1,8 +1,8 @@
...
 
-openshift admin ca create-server-cert --signer-cert=/var/lib/origin/openshift.local.config/master/ca.crt --signer-key=/var/lib/origin/openshift.local.config/master/ca.key --signer-serial=/var/lib/origin/openshift.local.config/master/ca.serial.txt --hostnames='docker-registry-default.#{routing-suffix},docker-registry-default.#{routing-suffix}:443,docker-registry.default.svc.cluster.local,172.30.1.1' --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key
-openshift cli create route edge docker-registry --service=docker-registry --ca-cert=/var/lib/origin/openshift.local.config/master/ca.crt --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key --hostname='docker-registry-default.#{routing-suffix}'
+openshift adm ca create-server-cert --signer-cert=/var/lib/origin/openshift.local.config/master/ca.crt --signer-key=/var/lib/origin/openshift.local.config/master/ca.key --signer-serial=/var/lib/origin/openshift.local.config/master/ca.serial.txt --hostnames='docker-registry-default.#{routing-suffix},docker-registry-default.#{routing-suffix}:443,docker-registry.default.svc.cluster.local,172.30.1.1' --cert=/etc/secrets/registry.crt --key=/etc/secrets/registry.key
```
LalatenduMohanty: @praveenkumar We can replace `openshift adm ca create-server-cert` with `oc adm ca create-server-cert` right? But the oc has to be inside the origin container right?

pkg/minishift/openshift/openshift.go
```diff
@@ -91,7 +92,17 @@ func Patch(target OpenShiftPatchTarget, patch string, commander docker.DockerCom
...
 	patchCommand := fmt.Sprintf("ex config patch %s --patch='%s'", containerConfigPath, patch)
-	result, err := commander.Exec("-t", constants.OpenshiftContainerName, constants.OpenshiftExec, patchCommand)
+	executer := constants.OpenshiftExec
+	// OpenShift >= 3.9 we need to run `ex config patch` using oc binary inside the origin container instead openshift server binary.
+	valid, err := openshiftVersionCheck.IsGreaterOrEqualToBaseVersion(openshiftVersion, "v3.9.0-alpha.3")
+	if err != nil {
+		return false, err
+	}
+	if valid {
+		executer = constants.OpenshiftOcExec
```
budhrg: :+1:

docs/source/using/addons.adoc
```diff
@@ -92,7 +92,7 @@ The `oc` command is executed as *system:admin*.
 ====
 
 openshift::
-If the add-on command starts with `openshift`, you can run the *openshift* binary within the container that runs OpenShift.
+If the add-on command starts with `openshift`, you can run the *oc* binary within the container that runs OpenShift.
```
LalatenduMohanty: 👍 . At least we are transparent now about what we are doing.

budhrg: This will create confusion which one to use. `oc adm` or `openshift adm`.

We need to give more info about when to use which command.
screenshot from 2018-03-27 11-36-36

LalatenduMohanty: @praveenkumar Here is my suggestion:
If the add-on command starts with `openshift`, it uses the `oc` binary that is present in the OpenShift container.
This means that any file parameters or other system-specific parameters must match the environment of the container instead of your host. 
This command should be used when successful execution of the `oc` command needs to change something in the OpenShift container.

gbraad: well, that was the suggestion I made; `openshift` should indicate a remote `oc` instance; in this case happening inside the origin server container. while `oc` will start a local `oc` from `OcPath`.

LalatenduMohanty: Another attempt: 

If the add-on command starts with openshift, it uses the oc binary present in the OpenShift container to execute the command.
This is useful when the oc command required to be inside the OpenShift container for successful execution of the add-on command. 
This means that any file parameters or other system-specific parameters must match the environment of the container instead of your host.

gbraad:
Approved now...

Issue #2174, opened 2018-03-26
RHEL iso on minishfit is not re-registered after once stopped and unregistered


anjannath:
### General information

  * Minishift version: v1.14.0
  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce

  1.  Start minishift with RHEL iso
  2. Stop minishift
  3. Start minishift

### Expected
The rhel iso is registered when minishift is started the second time.

### Actual
Registration doesn't happen / unregistering fails while `minishfit stop`

Ref: https://issues.jboss.org/projects/CDK/issues/CDK-270

PR #2176, opened 2018-03-26
Issue #2174 Handle unregistration properly for rhel iso

anjannath:
Fixes: #2174 

Ref: https://issues.jboss.org/projects/CDK/issues/CDK-270

### DIFF
--- a/pkg/minishift/registration/redhat.go
+++ b/pkg/minishift/registration/redhat.go
@@ -130,6 +130,10 @@ func (registrator *RedHatRegistrator) Unregister(param *RegistrationParameters) 
 			"sudo -E subscription-manager unregister"); err != nil {
 			return err
 		}
+		if _, err := registrator.SSHCommand(
+			"sudo -E subscription-manager clean"); err != nil {
+			return err
+		}
 	} else {
 		return err
 	}


pkg/minishift/registration/redhat.go
```diff
@@ -130,6 +130,14 @@ func (registrator *RedHatRegistrator) Unregister(param *RegistrationParameters)
 			"sudo -E subscription-manager unregister"); err != nil {
 			return err
 		}
+		if _, err := registrator.SSHCommand(
+			"sudo -E subscription-manager remove --all"); err != nil {
```
praveenkumar: Why we need this as per subscription manager doc, it clearly indicates that `unregister` should strip its subscriptions. 

- https://access.redhat.com/documentation/en-us/red_hat_subscription_management/1/html/quick_registration_for_rhel/un-registering

LalatenduMohanty: Copying the error from @anjannath 
```
$ sudo -E subscription-manager status
Consumer profile "67845724-d9be-4048-95fa-e81a162c8761" has been deleted from the server. You can use command clean or unregister to remove local profile.
```
So my guess is it is related  to the `local profile` not being cleaned properly

LalatenduMohanty:
@anjannath Any update from Ondrej?

anjannath:
@gbraad Clean removes the local profiles for RHSM, this was added to care for the case where, the registration is canceled in the RHSM server but local profiles are present, it is going to use the old info for again for registering, which i believe caused the issue.

anjannath:
@LalatenduMohanty Onderj have replied on the jira issue, seems like he's seeing some regression, so i rebased the changes again on top of latest release and given him the binary and iso to test..

anjannath:
Here's the latest [comment](https://issues.jboss.org/browse/CDK-270?focusedCommentId=13553753&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-13553753) on the jira issue. The binary which he's referring to, which i gave him last, had the registration fix in it. We are good to merge this i guess..

budhrg:
@anjannath He mentioned `Although, registration issue was still there.` in the comment. Not sure it means it has been fixed.

anjannath:
@budhrg He said that for the beta bits, which do not have the fix so registration issue was there... what he confirmed for in the beta bits is that the regression which he was seeing before is not there anymore..

budhrg:
@anjannath 
> that's what i understood from his comment, we could ask him to further clarify this..

Sure. Clarification is needed.

LalatenduMohanty:
@praveenkumar It is waiting for your review. Please review.

pkg/minishift/registration/redhat.go
```diff
@@ -130,6 +130,10 @@ func (registrator *RedHatRegistrator) Unregister(param *RegistrationParameters)
 			"sudo -E subscription-manager unregister"); err != nil {
 			return err
 		}
+		if _, err := registrator.SSHCommand(
+			"sudo -E subscription-manager clean"); err != nil {
+			return err
+		}
```
praveenkumar: :+1:

Issue #2178, opened 2018-03-26
Update minishift centos iso version to 1.8.0


praveenkumar:
Recently released 1.8.0 centos iso, we need to update to it.

anjannath:
Resolved via #2179

PR #2179, opened 2018-03-26
Issue #2178 Update centos ISO from 1.7.0 to 1.8.0

praveenkumar:


gbraad:
Maybe for next time, we should issue a minor release... since only from
this point on, we are able to deliver a working Hyper-V deployment.

Lessons learned...

PR #2190, opened 2018-03-27
Issue #2189 Fix minishift update command

### LINKED ISSUE #2189: "minishift" update is failing
### General information

  * Minishift version: v1.15.0
  * OS: Linux / macOS / Windows

### Steps to reproduce

  1. Compile minishift from source after the commit from https://github.com/minishift/minishift/pull/2078
  2.  Change Makefile to keep minishift version less than v.1.15.0
  3.  run `minishift update`
 
### Actual
```
$ minishift update
Do you want to update from 1.14.0 to 1.15.0 now? [y/N]: y
Update failed: Directory '/Users/lmohanty/code/gocode/bin' binary doesn't have write permission.
Please fix the permission issue and try running the command again.
```

budhrg:
Fix #2189 

Test it as : 
```
# vim Makefile and change MINISHIFT_VERSION = 1.14.0
$ make
$ minishift update
```

### DIFF
--- a/pkg/minishift/update/update.go
+++ b/pkg/minishift/update/update.go
@@ -81,8 +81,8 @@ func Update(latestVersion semver.Version) error {
 	var extName string
 
 	path, _ := osext.Executable()
-	if writable := util.IsDirectoryWritable(path); !writable {
-		return fmt.Errorf("Directory '%s' binary doesn't have write permission.\nPlease fix the permission issue and try running the command again.", filepath.Dir(path))
+	if writable := util.IsDirectoryWritable(filepath.Dir(path)); !writable {
+		return fmt.Errorf("Directory '%s' doesn't have write permission.\nPlease fix the permission issue and try running the command again.", filepath.Dir(path))
 	}
 
 	// Temporary directory to store archive contents
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -187,10 +187,10 @@ func CommandExecutesSuccessfully(cmd string, args ...string) bool {
 func IsDirectoryWritable(path string) bool {
 	tmpFilePath := filepath.Join(path, "tmp")
 	_, err := os.Create(tmpFilePath)
-	defer os.Remove(tmpFilePath)
 	if err != nil {
 		return false
 	}
+	defer os.Remove(tmpFilePath)
 
 	return true
 }


pkg/util/utils.go
```diff
@@ -187,10 +187,10 @@ func CommandExecutesSuccessfully(cmd string, args ...string) bool {
 func IsDirectoryWritable(path string) bool {
 	tmpFilePath := filepath.Join(path, "tmp")
 	_, err := os.Create(tmpFilePath)
-	defer os.Remove(tmpFilePath)
 	if err != nil {
 		return false
 	}
+	defer os.Remove(tmpFilePath)
```
anjannath: Why move this line?

pkg/minishift/update/update.go
```diff
@@ -81,7 +81,7 @@ func Update(latestVersion semver.Version) error {
 	var extName string
 
 	path, _ := osext.Executable()
-	if writable := util.IsDirectoryWritable(path); !writable {
+	if writable := util.IsDirectoryWritable(filepath.Dir(path)); !writable {
 		return fmt.Errorf("Directory '%s' binary doesn't have write permission.\nPlease fix the permission issue and try running the command again.", filepath.Dir(path))
```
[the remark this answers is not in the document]
LalatenduMohanty: We should remove the word `binary` from the sentence.

anjannath:
LGTM :+1:

pkg/util/utils.go
LalatenduMohanty: @anjannath Because we should handle the error from the `os.Create` and then call anything else. It is just more logical.

pkg/minishift/update/update.go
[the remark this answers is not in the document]
gbraad: not a big problem, but sure

pkg/util/utils.go
gbraad: it does not make sense to move it... defer is just acting like a queue

budhrg: It is not having significant impact here but actually it does if we defer the function:
#### When defer is after error check
```
  _, err := os.Create("/usr/bin/tmp")
  if err != nil {
    fmt.Println("err before defer", err)
    return false
  }
  defer func(){
    fmt.Println("err in defer: ", err)
    os.Remove(tmpFilePath)
  }()

  # Output
  err before defer open /usr/bin/tmp: permission denied
```     

#### When defer is before error check
```
  _, err := os.Create("/usr/bin/tmp")
  defer func(){
    fmt.Println("err in defer: ", err)
    os.Remove(tmpFilePath)
    }()
  if err != nil {
    fmt.Println("err before defer", err)
    return false
  }

  # Output
  err before defer open /usr/bin/tmp: permission denied
  err in defer:  open /usr/bin/tmp: permission denied
```

Check https://play.golang.org/p/pluqU0OQlEq

Issue #2188, opened 2018-03-27
Improve usage message for adding hostfolder interactively


anjannath:
### General information

  * Minishift version: v1.15.0+e07e789
  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce

  1. `minishift hostfolder add --interactive`
  2. `minishift hostfolder add --type sshfs --interactive`
  3. `minishift hostfolder add --type cifs --interactive`

### Expected
Should get prompt to enter required information (target folder, source, etc..)

### Actual
`Usage: minishift hostfolder add --type TYPE --source SOURCE --target TARGET HOST_FOLDER_NAME`

gbraad:
you need to add a name... I had reported this on the initial PR... so it IS an issue now?!?

LalatenduMohanty:
Resolved by #2193

PR #2193, opened 2018-03-28
Issue #2188 Update usage message for interactive use of hostfolder add

praveenkumar:


praveenkumar:
retest this please.

anjannath:
This still needs some brain cycles... ideally we should ask for all the info when its interactive mode, Including the type and hostfolder name.
```
$ minishfit hostfolder add --interactive
Type: ...
Source: ...
Target: ...
Name: ...
```

praveenkumar:
@anjannath if you go through the code that is not something we can do because according to your type it ask required information like for sshfs it asks source and target but in case of cifs it ask different info :(

cmd/minishift/cmd/hostfolder/add.go
```diff
@@ -32,6 +32,7 @@ import (
 
 const (
 	usage                = "Usage: minishift hostfolder add --type TYPE --source SOURCE --target TARGET HOST_FOLDER_NAME"
+	interactiveModeUsage = "Usage: minishift hostfolder add --interactive --type TYPE [sshfs|cifs (Default: cifs)] HOST_FOLDER_NAME"
```
anjannath: :+1:

anjannath:
@praveenkumar I see... but it can be done i guess, based on the interactive input of type, we can ask for the type the first thing, and based on that decide what info to ask the user.. anyway this needs to happen on a separate issue, The message looks fine now, better than before.

LalatenduMohanty:
@praveenkumar  I do not think the original issue is getting fixed by this PR. 
```
$ minishift hostfolder add --interactive
Usage: minishift hostfolder add --interactive --type TYPE [sshfs|cifs (Default: cifs)] HOST_FOLDER_NAME
```

I still do not see the prompt asking me to give input for the arguments.

LalatenduMohanty:
> @praveenkumar I do not think the original issue is getting fixed by this PR.

#2219

Issue #2204, opened 2018-03-30
Documentation does not mention Fixed/Static IP only works on CentOS/RHEL based images


gbraad:
The Static/Fixed IP functionality only works with CentOS/RHEL images, similar to how Hyper-V works: https://github.com/minishift/minishift/blob/master/docs/source/using/experimental-features.adoc#assign-ip-address-to-hyper-v This seems to have been forgotten in the documentation.

PR #2205, opened 2018-03-30
Issue #2204 fixed IP only works with CentOS or RHEL ISO

gbraad:


docs/source/using/experimental-features.adoc
```diff
@@ -92,9 +92,9 @@ Failing to do so will result in connectivity issues.
 [[set-fixed-ip]]
 == Set Fixed IP Address
 
-[NOTE]
+[IMPORTANT]
 ====
-Currently not supported on KVM as the driver plugin relies on the DHCP offer to determine the IP address.
+This only works with the CentOS/RHEL based image and is currently not supported on KVM as the driver plugin relies on the DHCP offer to determine the IP address.
```
LalatenduMohanty: s/based image/ ISO image/

gbraad: This is to stay consistent with previously written documentation:

https://github.com/minishift/minishift/blob/519429e3b3bd202a7a06509c84f4a13a1c0260dd/docs/source/using/experimental-features.adoc#assign-ip-address-to-hyper-v

gbraad: in that case it should be `{project}'s RHEL/CentOS-based ISO images.`, as it would avoid any confusion

gbraad: Checking, and I see https://github.com/minishift/minishift/blob/9ab21d6d93b4507e9ee086fcec4d862d4f100544/docs/source/using/profiles.adoc#L29

... so, ISO images or ISOs ? /cc @robin-owen

robin-owen: I'm fine with ISOs in the referenced line in **_profiles.adoc_** due to the use of "container images" directly following the use of "ISOs". Considering we deal with multiple possible definitions of "image", however, I would generally advocate being as explicit as possible.

We can always open a new issue to help disambiguate the use of the word "image".

gbraad: going to use ISOs

gbraad:
Moved to new PR #2236

PR #2218, opened 2018-04-02
Issue #273 Do not provision the VM with OpenShift

### LINKED ISSUE #273: Add  '--no-provision' flag to minishift start
When the flag is set, only the VM is started and OpenShift does not get provisioned yet. You can use provided Docker daemon for "other" use cases or you can pre-seed the daemon with images prior then provisioning OpenShift. The exact details need to be worked out.

gbraad:
Fixes #273 Add experimental support to not provision the VM.

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -97,6 +97,9 @@ var (
 	HostFoldersAutoMount = createConfigSetting("hostfolders-automount", SetBool, nil, nil, true, nil)
 	HostFoldersSftpPort  = createConfigSetting("hostfolders-sftp-port", SetInt, []setFn{validations.IsValidPort}, nil, true, nil)
 
+	// No Provision
+	NoProvision = createConfigSetting("no-provision", SetBool, nil, nil, true, nil)
+
 	// Image caching
 	ImageCaching = createConfigSetting("image-caching", SetBool, nil, nil, true, true)
 	CacheImages  = createConfigSetting("cache-images", SetSlice, nil, nil, false, nil)
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -61,7 +61,6 @@ import (
 const (
 	commandName             = "start"
 	defaultInsecureRegistry = "172.30.0.0/16"
-	hostfoldersAutoMountKey = "hostfolders-automount"
 )
 
 var (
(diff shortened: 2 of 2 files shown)


cmd/minishift/cmd/config/config.go
```diff
@@ -97,6 +97,9 @@ var (
 	HostFoldersAutoMount = createConfigSetting("hostfolders-automount", SetBool, nil, nil, true, nil)
 	HostFoldersSftpPort  = createConfigSetting("hostfolders-sftp-port", SetInt, []setFn{validations.IsValidPort}, nil, true, nil)
 
+	// No Provision
+	NoProvision = createConfigSetting("no-provision", SetBool, nil, nil, true, nil)
```
praveenkumar: :+1:

cmd/minishift/cmd/start.go
```diff
@@ -578,6 +583,7 @@ func initStartFlags() *flag.FlagSet {
 	startFlagSet.AddFlag(nameServersFlag)
 
 	if minishiftConfig.EnableExperimental {
+		startFlagSet.Bool(configCmd.NoProvision.Name, false, "Do not provision the VM with OpenShift (experimental)")
```
praveenkumar: :+1:

LalatenduMohanty:
@gbraad  The only thing I am not sure about it that, the `oc` binary gets downloaded (if not cached) for --no-provision. But without OpenShift cluster it does not have any value. So seems unnecessary to download the `oc` binary. Either we fix this in this PR or create an issue to fix this later. WDYT?

praveenkumar:
> The only thing I am not sure about it that, the oc binary gets downloaded (if not cached) for --no-provision. But without OpenShift cluster it does not have any value. So seems unnecessary to download the oc binary.

@LalatenduMohanty  @gbraad It does have value if you start the vm without provision then also you can use `minishift docker-env` and `minishift oc-env` then use download oc binary to just start cluster as per your requirement. This actually help us also to verify if there is any issue when doing plain `oc cluster up`. I would like to have it as it is without removing capablity of downloading oc binary.

LalatenduMohanty:
> This actually help us also to verify if there is any issue when doing plain oc cluster up. I would like to have it as it is without removing capablity of downloading oc binary.

Fair enough.

LalatenduMohanty:
@gbraad Thanks.

Issue #1564, opened 2017-10-13
openshift list service seems to report only http URLs


jeffmaury:
I have defined a route to the registry but with a TLS option.
The web console reports https://docker-registry-default.192.168.99.100.nip.io/ 
but 
`minishift openshift service list` reports:
```
|-----------|-----------------|----------|------------------------------------------------------|--------|
| NAMESPACE |      NAME       | NODEPORT |                      ROUTE-URL                       | WEIGHT |
|-----------|-----------------|----------|------------------------------------------------------|--------|
| default   | kubernetes      |          |                                                      |        |
|           | router          |          |                                                      |        |
|           | docker-registry |          | http://docker-registry-default.192.168.99.100.nip.io |        |
|-----------|-----------------|----------|------------------------------------------------------|--------|
```

anjannath:
Is this still relevent or it got solved? Anyways, I will try to reproduce this and report here.

praveenkumar:
@anjannath I don't think we resolved this and this is still valid.

anjannath:
This might be caused due to github's pagination.. 

https://developer.github.com/v3/guides/traversing-with-pagination/

gbraad:
???

The issue here is about the service list...

On Friday, April 20, 2018, Anjan Nath <notifications@github.com> wrote:
> This might be caused due to github's pagination..
>
> https://developer.github.com/v3/guides/traversing-with-pagination/
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.<
https://ci6.googleusercontent.com/proxy/zQfqzjvhWmO3xEqCGbL4SAQQlCCxxnq74DNNxMh7-3dlZnqD9B5jo_GHo0tdEOfio1le3ZzJCr22FwWieLe5O8fPRn4NvY_bi-oC_JQcIgNQc2_bRD9G9hABLVIo2UxApQuQ_SNiJL90ZoVnfDvEgRGyRi7VTw=s0-d-e1-ft#https://github.com/notifications/beacon/AAAHZhPyAUWjcQKko-ZHKoW2Mv4eMTGRks5tqaU3gaJpZM4P4cYW.gif
>

-- 

   Gerard Braad | http://gbraad.nl
   [ Doing Open Source Matters ]

anjannath:
Yes. :) It was only printing http URLs, secured routes were also prefixed
with http. That's what #2237 fixes + the refactor

>

PR #2237, opened 2018-04-04
Issue #1564 Print TLS service routes in o/p of `minishift openshift service list` command

anjannath:
Fixes: #1564 

- [x] Refactor `filterAndUpdateServiceSpecs`

### DIFF
--- a/cmd/minishift/cmd/openshift/service.go
+++ b/cmd/minishift/cmd/openshift/service.go
@@ -33,11 +33,11 @@ import (
 )
 
 var (
-	namespace string
-	inbrowser bool
-	https     bool
-	url       bool
-	service   string
+	namespace   string
+	inbrowser   bool
+	https       bool
+	url         bool
+	serviceName string
 )
 
 // serviceCmd represents the service command
@@ -67,21 +67,21 @@ var serviceCmd = &cobra.Command{
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error getting IP: %s", err.Error()))
 		}
 
-		service = args[0]
+		serviceName = args[0]
 
-		serviceSpecs, err := openshift.GetServiceSpecs(namespace)
+		services, err := openshift.GetServices(namespace)
 		if err != nil {
 			atexit.ExitWithMessage(1, err.Error())
 		}
 
 		if url {
-			stdOutURL(serviceSpecs, ip)
+			stdOutURL(services, ip)
 		}
 		if inbrowser {
-			openInBrowser(serviceSpecs, ip)
+			openInBrowser(services, ip)
 		}
 		if !url && !inbrowser {
-			printToStdOut(serviceSpecs, ip)
+			printToStdOut(services, ip)
 		}
 	},
 }
@@ -94,36 +94,36 @@ func init() {
 	OpenShiftCmd.AddCommand(serviceCmd)
 }
 
-func openInBrowser(serviceSpecs []openshift.ServiceSpec, ip string) {
-	serviceURL := getServiceURL(serviceSpecs, ip)
+func openInBrowser(services []openshift.Service, ip string) {
+	serviceURL := getServiceURL(services, ip)
 	fmt.Fprintln(os.Stdout, "Opening the route/NodePort "+serviceURL+" in the default browser...")
 	browser.OpenURL(serviceURL)
 }
 
-func stdOutURL(serviceSpecs []openshift.ServiceSpec, ip string) {
-	serviceURL := getServiceURL(serviceSpecs, ip)
+func stdOutURL(services []openshift.Service, ip string) {
+	serviceURL := getServiceURL(services, ip)
 	fmt.Fprintln(os.Stdout, serviceURL)
 }
 
-func getServiceURL(serviceSpecs []openshift.ServiceSpec, ip string) string {
+func getServiceURL(services []openshift.Service, ip string) string {
 	serviceURL := ""
-	namespaceList := isServiceInMultipleNamespace(serviceSpecs, service)
+	namespaceList := isServiceInMultipleNamespace(services, serviceName)
 	if len(namespaceList) == 0 {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Service '%s' does not exist", service))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Service '%s' does not exist", serviceName))
 	}
 	if len(namespaceList) > 1 {
 		namespaces := strings.TrimSpace(strings.Join(namespaceList, ", "))
-		atexit.ExitWithMessage(1, fmt.Sprintf("Service '%s' exists in multiple namespaces (%s), you need to chose a specific namespace using -n <namespace>.", service, namespaces))
+		atexit.ExitWithMessage(1, fmt.Sprintf("Service '%s' exists in multiple namespaces (%s), you need to chose a specific namespace using -n <namespace>.", serviceName, namespaces))
 	}
 
-	for _, serviceSpec := range serviceSpecs {
-		if serviceSpec.Name == service {
-			if serviceSpec.URL != nil {
-				serviceURL = serviceSpec.URL[0]
+	for _, service := range services {
+		if service.Name == serviceName {
+			if service.URL != nil {
+				serviceURL = service.URL[0]
 				return serviceURL
 
-			} else if serviceSpec.NodePort != "" {
-				nodePortURL := fmt.Sprintf("%s:%s", ip, serviceSpec.NodePort)
+			} else if service.NodePort != "" {
+				nodePortURL := fmt.Sprintf("%s:%s", ip, service.NodePort)
 				urlScheme := "http://"
 				if https {
 					urlScheme = "https://"
@@ -131,40 +131,40 @@ func getServiceURL(serviceSpecs []openshift.ServiceSpec, ip string) string {
 				serviceURL = urlScheme + nodePortURL
 				return serviceURL
 			} else {
-				atexit.ExitWithMessage(1, fmt.Sprintf("Service '%s' in namespace '%s' does not have route associated which can be opened in the browser.", serviceSpec.Name, serviceSpec.Namespace))
+				atexit.ExitWithMessage(1, fmt.Sprintf("Service '%s' in namespace '%s' does not have route associated which can be opened in the browser.", service.Name, service.Namespace))
 			}
 		}
 	}
 	return serviceURL
 }
 
-func isServiceInMultipleNamespace(serviceSpecs []openshift.ServiceSpec, service string) []string {
+func isServiceInMultipleNamespace(services []openshift.Service, serviceName string) []string {
 	namespceList := []string{}
-	for _, serviceSpec := range serviceSpecs {
-		if serviceSpec.Name == service {
-			namespceList = append(namespceList, serviceSpec.Namespace)
+	for _, service := range services {
+		if service.Name == serviceName {
+			namespceList = append(namespceList, service.Namespace)
 		}
 	}
 	return namespceList
 }
 
-func printToStdOut(serviceSpecs []openshift.ServiceSpec, ip string) {
+func printToStdOut(services []openshift.Service, ip string) {
 	var data [][]string
 	var urls, weights string
 
-	for _, serviceSpec := range serviceSpecs {
-		if serviceSpec.Name == service {
-			nodePortURL := serviceSpec.NodePort
+	for _, service := range services {
+		if service.Name == serviceName {
+			nodePortURL := service.NodePort
 			if nodePortURL != "" {
 				nodePortURL = fmt.Sprintf("%s:%s", ip, nodePortURL)
 			}
-			if serviceSpec.URL != nil {
-				urls = strings.Join(serviceSpec.URL, "\n")
+			if service.URL != nil {
+				urls = strings.Join(service.URL, "\n")
 			}
-			if serviceSpec.Weight != nil {
-				weights = strings.Join(serviceSpec.Weight, "\n")
+			if service.Weight != nil {
+				weights = strings.Join(service.Weight, "\n")
 			}
-			data = append(data, []string{serviceSpec.Namespace, serviceSpec.Name, nodePortURL, urls, weights})
+			data = append(data, []string{service.Namespace, service.Name, nodePortURL, urls, weights})
 		}
 	}
 	table := tablewriter.NewWriter(os.Stdout)
--- a/pkg/minishift/openshift/service.go
+++ b/pkg/minishift/openshift/service.go
@@ -33,7 +33,12 @@ var (
 	runner util.Runner = &util.RealRunner{}
 )
 
-type Service struct {
+const (
+	http  = "http://%s"
+	https = "https://%s"
+)
+
+type ServiceSpec struct {
 	Items []struct {
 		Metadata struct {
 			Name string `json:"name"`
@@ -46,7 +51,7 @@ type Service struct {
 	} `json:"items"`
 }
 
-type Route struct {
+type RouteSpec struct {
 	Items []struct {
 		Spec struct {
 			AlternateBackends []struct {
@@ -54,7 +59,10 @@ type Route struct {
 				Weight int    `json:"weight"`
 			} `json:"alternateBackends"`
 			Host string `json:"host"`
-			To   struct {
+			Tls  struct {
+				Termination string `json:"termination"`
+			} `json:"tls"`
+			To struct {
 				Name   string `json:"name"`
 				Weight int    `json:"weight"`
 			} `json:"to"`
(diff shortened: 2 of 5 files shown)


anjannath:
retest this please

pkg/minishift/openshift/service.go
```diff
@@ -214,15 +218,29 @@ func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace
...
 				} else {
-					serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf("http://%s", routeURL))
-					serviceSpecMap[service.Name].Weight = append(serviceSpecMap[service.Name].Weight, service.Weight)
+					if service.Tls != "" {
+						serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf("https://%s", routeURL))
+						serviceSpecMap[service.Name].Weight = append(serviceSpecMap[service.Name].Weight, service.Weight)
+					} else {
+						serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf("http://%s", routeURL))
+						serviceSpecMap[service.Name].Weight = append(serviceSpecMap[service.Name].Weight, service.Weight)
+					}
```
praveenkumar: you need to add a unit test also for this change, check https://github.com/minishift/minishift/blob/master/pkg/minishift/openshift/service_test.go we have a dummy testdata for route and service spec you can add another data which have tls and then check if it returns `https` uri for route.

anjannath: Okay

anjannath: @praveenkumar Added the unit test, PTAL

anjannath:
@gbraad Please take a look at the re-factor... :-)

pkg/minishift/openshift/service.go
```diff
@@ -238,6 +220,37 @@ func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace
 	return serviceSpecs
 }
 
+func addStuffToServiceSpecMap(serviceSpecMap map[string]*ServiceSpec, url string, service ServiceWeight, namespace string, np string) {
+	const (
+		http  = "http://%s"
+		https = "https://%s"
+	)
+
+	if _, ok := serviceSpecMap[service.Name]; !ok {
```
gbraad: sorry, but this is confusing

pkg/minishift/openshift/service.go
```diff
@@ -238,6 +220,37 @@ func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace
...
+		if service.Tls != "" {
+			serviceSpecMap[service.Name] = fillServiceSpec(service.Name, url, np, service.Weight, namespace, https)
+		} else {
+			serviceSpecMap[service.Name] = fillServiceSpec(service.Name, url, np, service.Weight, namespace, http)
+		}
+	} else {
+		if service.Tls != "" {
```
gbraad: it would be more readable if ALWAYS http is assumed, unless TLS is set

gbraad:
```
serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf(http, url))
if service.Tls != "" {
  serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf(https, url))
}
```

This will be not slower or less-performant... and might even be optimized away. it makes the code more readable...

budhrg: Agree with this comment.

gbraad: also not sure what append actually does here... does it ADD or just format something?

maybe it is the structure... but it feels unclear to me.

pkg/minishift/openshift/service.go
```diff
@@ -195,35 +199,13 @@ func calculateWeight(a float64, b float64) string {
...
-						Name:     service.Name,
-						URL:      []string{fmt.Sprintf("http://%s", routeURL)},
-						NodePort: serviceNodePort[service.Name],
-						Weight:   []string{service.Weight},
-					}
-				} else {
-					serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf("http://%s", routeURL))
-					serviceSpecMap[service.Name].Weight = append(serviceSpecMap[service.Name].Weight, service.Weight)
-				}
+				addStuffToServiceSpecMap(serviceSpecMap, routeURL, service, namespace, serviceNodePort[service.Name])
```
gbraad: why named `Stuff`?

pkg/testing/testdata/testdata.go
```diff
@@ -419,6 +466,55 @@ const (
...
+                "resourceVersion": "3217",
+                "selfLink": "/apis/route.openshift.io/v1/namespaces/myproject/routes/ruby-ex-tls",
+                "uid": "0eeabedc-372d-11e8-853f-567fb20565f6"
+            },
+            "spec": {
+                "host": "ruby-ex-tls-myproject.192.168.42.14.nip.io",
+                "port": {
+                    "targetPort": "8080-tcp"
+                },
+                "tls": {
```
gbraad: this is tested? if so... where can I see the expected outcome?

anjannath: yes, this is tested.. the outcome would be we have the route starting with https instead of http.. i added a  service and route spec in `pkg/testing/testdata.go`

pkg/minishift/openshift/service.go
anjannath: It appends, so this part of the code executes if the service is already present in serviceSecMap, then we just append the new route and weight to the existing values..

pkg/minishift/openshift/service.go
anjannath: Not able to find a better name.. i suck at naming, suggestions are welcome!

anjannath: Ohh, did you mean just `addToServiceSpecMap`? :-)

pkg/minishift/openshift/service.go
anjannath: But that will make it show two routes, suppose the service does not have insecure route? In that case also it will have the http route.. while actually it was not defined..

anjannath: might be a good thing to keep the full route in the `ServiceWeight` struct.. name needs to be changed, in that case we can avoid the check for the `Tls` field...

anjannath:
@gbraad Did some further refactor..

pkg/minishift/openshift/service.go
```diff
@@ -238,6 +220,60 @@ func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace
...
+}
+
+//serviceInServiceSpecMap returns true if the service is present in serviceSpecMap
+func serviceInServiceSpecMap(serviceSpecMap map[string]*ServiceSpec, name string) bool {
+	_, ok := serviceSpecMap[name]
+	return ok
+}
+
+func appendUrlToServiceSpec(serviceSpecMap map[string]*ServiceSpec, service ServiceWeight, url string) {
+	if service.Tls != "" {
```
gbraad: In the next part you use:

```diff
+	const (
+		http  = "http://%s"
+		https = "https://%s"
+	)
```

why not re-use this?

pkg/minishift/openshift/service.go
```diff
@@ -238,6 +220,60 @@ func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace
...
+		serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf("https://%s", url))
```
gbraad: ^^^ "https://"

pkg/minishift/openshift/service.go
anjannath: Yeah.. I should re-use this...

pkg/minishift/openshift/service.go
```diff
@@ -238,6 +220,60 @@ func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace
...
+	if !serviceInServiceSpecMap(serviceSpecMap, service.Name) {
+		addSpecToServiceSpecMap(serviceSpecMap, service, url, namespace, np)
+	} else {
+		appendUrlToServiceSpec(serviceSpecMap, service, url)
+		appendWeightToServiceSpec(serviceSpecMap, service.Name, service.Weight)
+	}
+}
+
+//serviceInServiceSpecMap returns true if the service is present in serviceSpecMap
+func serviceInServiceSpecMap(serviceSpecMap map[string]*ServiceSpec, name string) bool {
```
gbraad: name as `isBlahbla` as it is a bool

pkg/minishift/openshift/service.go
```diff
@@ -238,6 +220,60 @@ func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace
 	return serviceSpecs
 }
 
+func serviceHasNodePort(serviceNodePort map[string]string, name string) bool {
```
gbraad: try to use `is` or `has` as naming to start with

anjannath: what about `doesServiceHaveNodePort` ?

gbraad: conventions are to use `is` or `has`

anjannath: Done.. `hasNodePort` :-)

pkg/minishift/openshift/service.go
```diff
@@ -238,6 +225,56 @@ func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace
...
+func addSpecToServiceSpecMap(serviceSpecMap map[string]*ServiceSpec, service ServiceWeight, url, namespace, np string) {
+
+	if service.Tls != "" {
+		serviceSpecMap[service.Name] = fillServiceSpec(service.Name, url, np, service.Weight, namespace, https)
+	} else {
+		serviceSpecMap[service.Name] = fillServiceSpec(service.Name, url, np, service.Weight, namespace, http)
+	}
+}
+
+func fillServiceSpec(name, url, np, weight, namespace, routeType string) *ServiceSpec {
```
gbraad: I would place this above the `addSpec`

pkg/minishift/openshift/service.go
```diff
@@ -192,46 +205,65 @@ func calculateWeight(a float64, b float64) string {
...
-						Name:     service.Name,
-						URL:      []string{fmt.Sprintf("http://%s", routeURL)},
-						NodePort: serviceNodePort[service.Name],
-						Weight:   []string{service.Weight},
-					}
-				} else {
-					serviceSpecMap[service.Name].URL = append(serviceSpecMap[service.Name].URL, fmt.Sprintf("http://%s", routeURL))
-					serviceSpecMap[service.Name].Weight = append(serviceSpecMap[service.Name].Weight, service.Weight)
-				}
+			if hasNodePort(serviceNodePort, service.Name) {
```
gbraad: `serviceNodePort` is a little bit unfortunately named. It almost feels, because of the name, that THIS test will alwats return `true` anyway.

gbraad: the origin of the name is:
```
	serviceNodePort, err := getService(namespace)
```
and I think this can be renamed to `service`. or if that is too generic, `serviceForNamespace`, etc.

pkg/minishift/openshift/service.go
```diff
@@ -75,6 +84,10 @@ type ServiceSpec struct {
 	Weight    []string
 }
 
+type ServiceSpecMap struct {
+	S map[string]*ServiceSpec
```
budhrg: `S` this is called code smell. Please give proper member name, should be at least 3 letters.

pkg/minishift/openshift/service.go
```diff
@@ -192,46 +205,65 @@ func calculateWeight(a float64, b float64) string {
...
+	return fmt.Sprintf(http, url)
+}
+
+func NewServiceSpecMap() *ServiceSpecMap {
+	return &ServiceSpecMap{S: make(map[string]*ServiceSpec)}
+}
+
+func (s *ServiceSpecMap) Add(service ServiceWeight, routeURL, namespace, nodePort string) {
+	if s.HasService(service.Name) {
+		s.S[service.Name].URL = append(s.S[service.Name].URL, genUrl(service, routeURL))
```
budhrg: See how it is being used here.   What is `s.S` means here?

pkg/minishift/openshift/service.go
gbraad: > S this is called code smell. 

so you will also call `i` from now on `iter`, or `iterator`? Darn, seems I need to find an alternative to call `IP`, as it is not 'at least' three long.

In short, this is not a necessity. every name should be in context. Here he used it as it gets encapsulated... perhaps a function `getSpecs()` or just `get()` would be clearer even... but sure, I would not have called it 'code smell'.

pkg/minishift/openshift/service.go
gbraad: simple solution... but I hate to see 'implementation' details in naming (things like a ServiceString, EnableBool) and I see the same here. Maybe rather name it as follows:

`ServiceSpecMap` => `ServiceSpec`
and then you can use `S` => `Map`

note: and this fits exactly 3 characters.... how nice :-s. but still contains implementation details.

anjannath: I did not name S as something else, because we never directly interact with him..(in outside funtions) and s.S means S inside s.. thats what the `.` means.. :)

We always interact with `s *ServiceSpecMap` which has the functions `Add` and `hasService` for doing this.

I agree that the name of this needs to change..

gbraad: Just arbitrary rules about naming infuriate me (not that bad, just rather see a suggestion instead of merely criticism)

pkg/minishift/openshift/service.go
budhrg: @anjannath , this is creating more confusion and also hard to get proper name. Probably, it is good fit here

I am able to achieve the same result with following:
```
func addToServiceSpecMap(s map[string]*ServiceSpec, service ServiceWeight, routeURL, namespace, nodePort string) {
	_, ok := s[service.Name]
	if ok {
		s[service.Name].URL = append(s[service.Name].URL, genUrl(service, routeURL))
		s[service.Name].Weight = append(s[service.Name].Weight, service.Weight)
	} else {
		s[service.Name] = createServiceSpec(service, routeURL, namespace, nodePort)
	}
}

func filterAndUpdateServiceSpecs(routeSpec map[string][]ServiceWeight, namespace string, serviceNodePort map[string]string) []ServiceSpec {
	var serviceSpecs []ServiceSpec

	serviceSpecMap := make(map[string]*ServiceSpec)
	for routeURL, v := range routeSpec {
		for _, service := range v {
			if hasNodePort(serviceNodePort, service.Name) {
				addToServiceSpecMap(serviceSpecMap, service, routeURL, namespace, serviceNodePort[service.Name])
				delete(serviceNodePort, service.Name)
			} else {
				addToServiceSpecMap(serviceSpecMap, service, routeURL, namespace, "")
			}
		}
	}

	for k, v := range serviceNodePort {
		serviceSpecs = append(serviceSpecs, ServiceSpec{Namespace: namespace, Name: k, URL: nil, NodePort: v, Weight: nil})
	}
	for _, v := range serviceSpecMap {
		serviceSpecs = append(serviceSpecs, *v)
	}

	return serviceSpecs
}
```

anjannath: This was what it looked like.. before.. with a little falvour (the `genUrl` function was not there), but here we are passing around `ServiceSpecMap`, and we are doing operations on it by passing it to functions.. it was a good candidate for making a struct and adding methods.

pkg/minishift/openshift/service.go
```diff
@@ -46,71 +51,79 @@ type Service struct {
...
-type ServiceSpec struct {
+type Service struct {
 	Namespace string
 	Name      string
 	URL       []string
 	NodePort  string
 	Weight    []string
 }
 
+type Services struct {
```
budhrg: This looks fine too me :+1:

LalatenduMohanty: 👍

pkg/minishift/openshift/service.go
```diff
@@ -46,71 +51,79 @@ type Service struct {
...
+		serviceNodePort, err := getNodePorts(namespace)
 		if err != nil {
-			return serviceSpecs, err
+			return services, err
 		}
-		routeSpec, err := getRouteSpec(namespace)
+		route, err := getRoutes(namespace)
 		if err != nil {
-			return serviceSpecs, err
+			return services, err
```
LalatenduMohanty: Till this point services is empty, right? cc @anjannath

anjannath: @LalatenduMohanty Yes..

cmd/minishift/cmd/openshift/service.go
```diff
@@ -33,11 +33,11 @@ import (
...
-	namespace string
-	inbrowser bool
-	https     bool
-	url       bool
-	service   string
+	namespace   string
+	inbrowser   bool
+	https       bool
+	url         bool
+	serviceName string
```
praveenkumar: 👍

Issue #2122, opened 2018-03-14
Determining whether a host folder is mounted is not taking source and target into account


hferentschik:
The code looks like this:

```Go
	procMounts := ""
	if isRunning {
		cmd := fmt.Sprint("cat /proc/mounts")
		procMounts, _ = drivers.RunSSHCommandFromDriver(driver, cmd)
	}

	hostfolders := miniConfig.AllInstancesConfig.HostFolders
	hostfolders = append(hostfolders, miniConfig.InstanceConfig.HostFolders...)
	var mounts []MountInfo
	for _, hostFolder := range hostfolders {

		source := ""
		switch hostFolder.Type {
		case CIFS.String():
			source = hostFolder.Options[config.UncPath]
		case SSHFS.String():
			source = hostFolder.Options[config.Source]
		}

		mounted := false
		if isRunning && strings.Contains(procMounts, hostFolder.MountPoint()) {
			mounted = true
		}

		mount := MountInfo{
			Name:       hostFolder.Name,
			Type:       hostFolder.Type,
			Source:     source,
			MountPoint: hostFolder.MountPoint(),
			Mounted:    mounted,
		}

		mounts = append(mounts, mount)
	}
```

The code only checks whether the output of `cat /proc/mounts` contains the mount point. However, this fails, if for example the mountpoints are foo and foo1. Really the output should be considered line by line and take hostfolder source and target into account.

gbraad:
The above code can be found here:

https://github.com/minishift/minishift/blob/c56f051fa872e8fd15d41544f6096e01f8a6845a/pkg/minishift/hostfolder/hostfolder_manager.go#L103-L136

We might also have to check:

https://github.com/minishift/minishift/blob/c56f051fa872e8fd15d41544f6096e01f8a6845a/pkg/minishift/hostfolder/hostfolder_manager.go#L284-L297

LalatenduMohanty:
> The code only checks whether the output of cat /proc/mounts contains the mount point. However, this fails, if for example the mountpoints are foo and foo1. Really the output should be considered line by line and take hostfolder source and target into account.

+1 . We need to fix it soon as users will find this as a bug.

LalatenduMohanty:
Resolved via #2240

PR #2240, opened 2018-04-04
Issue #2122 Add regex to determine a host folder is mounted

praveenkumar:
How to test:

- Use minishift without this PR.

```
$ minishift hostfolder add test --type sshfs --source <your_source> --target /mnt/sda1/test
$ minishift hostfolder add test1 --type sshfs --source <your_source> --target /mnt/sda1/test1
$ ./minishift hostfolder list
Name    Type    Source                        Mountpoint        Mounted
test    sshfs   /home/prkumar/work/vagrant/   /mnt/sda1/test    N
test1   sshfs   /home/prkumar/work/vagrant/   /mnt/sda1/test1   N
$ ./minishift hostfolder mount test1
$ ./minishift hostfolder list
Name    Type    Source                        Mountpoint        Mounted
test    sshfs   /home/prkumar/work/vagrant/   /mnt/sda1/test    Y => we didn't mount it
test1   sshfs   /home/prkumar/work/vagrant/   /mnt/sda1/test1   Y
$ ./minishift hostfolder mount test
host folder is already mounted => but we never mounted it
```

After this PR
```
$ ./minishift hostfolder list
Name    Type    Source                        Mountpoint        Mounted
test    sshfs   /home/prkumar/work/vagrant/   /mnt/sda1/test    N
test1   sshfs   /home/prkumar/work/vagrant/   /mnt/sda1/test1   Y
$ ./minishift hostfolder mount test
$ ./minishift hostfolder list
Name    Type    Source                        Mountpoint        Mounted
test    sshfs   /home/prkumar/work/vagrant/   /mnt/sda1/test    Y
test1   sshfs   /home/prkumar/work/vagrant/   /mnt/sda1/test1   Y
```

### DIFF
--- a/pkg/minishift/hostfolder/hostfolder_manager.go
+++ b/pkg/minishift/hostfolder/hostfolder_manager.go
@@ -19,12 +19,13 @@ package hostfolder
 import (
 	"errors"
 	"fmt"
+	"regexp"
+
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/state"
 	"github.com/golang/glog"
 	miniConfig "github.com/minishift/minishift/pkg/minishift/config"
 	"github.com/minishift/minishift/pkg/minishift/hostfolder/config"
-	"strings"
 )
 
 type MountInfo struct {
@@ -100,12 +101,6 @@ func (m *Manager) List(driver drivers.Driver) ([]MountInfo, error) {
 		return nil, errors.New("no host folders defined")
 	}
 
-	procMounts := ""
-	if isRunning {
-		cmd := fmt.Sprint("cat /proc/mounts")
-		procMounts, _ = drivers.RunSSHCommandFromDriver(driver, cmd)
-	}
-
 	hostfolders := miniConfig.AllInstancesConfig.HostFolders
 	hostfolders = append(hostfolders, miniConfig.InstanceConfig.HostFolders...)
 	var mounts []MountInfo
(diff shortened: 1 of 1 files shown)


pkg/minishift/hostfolder/hostfolder_manager.go
```diff
@@ -118,9 +119,9 @@ func (m *Manager) List(driver drivers.Driver) ([]MountInfo, error) {
 		case SSHFS.String():
 			source = hostFolder.Options[config.Source]
 		}
-
+		r := regexp.MustCompile(fmt.Sprintf("%s(\\s+)", hostFolder.MountPoint()))
 		mounted := false
-		if isRunning && strings.Contains(procMounts, hostFolder.MountPoint()) {
+		if isRunning && r.MatchString(procMounts) {
```
LalatenduMohanty: We are doing the same thing as isHostFolderMounted() here. So why can not we reuse the function?  The only difference would be that we need to run more number of `cat /proc/mounts` commands but will that make a lot of difference. Also running same regular expression in two different functions is not a ideal thing to do. Ideally it should be done in one place.

LalatenduMohanty: @praveenkumar I was thinking something like this

```
function doesMountMatch(string procMounts, string hostFolderMount) bool {
	r := regexp.MustCompile(fmt.Sprintf("%s(\\s+)", hostFolderMount)

	if !r.MatchString(procMounts) {
		return false 
	}
	return true
}
```

Issue #2234, opened 2018-04-04
Move --logging and --metrics to be part of the experimental features


gbraad:
Although in text we explain `--logging` and `--metrics` are experimenta, we haven't changed the flags to be part of the `MINISHIFT_ENABLE_EXPERIMENTAL` flag. Seen the issues we have in recent releases, I think we need to address this, to reduce the amount of questions related to this topic.

gbraad:
I have seen the related issue: https://github.com/openshift/origin/pull/19209
Adjusting the PR...

PR #2241, opened 2018-04-04
Issue #2234 remove flags for metrics, logging and service catalog

gbraad:
See https://github.com/openshift/origin/pull/19209 for more info
See #2248 for more info

/cc: @deads2k @jcantrill

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -71,11 +71,8 @@ var (
 	HostPvDir         = createConfigSetting("host-pv-dir", SetString, []setFn{validations.IsValidPath}, nil, true, nil)
 	ServerLogLevel    = createConfigSetting("server-loglevel", SetInt, []setFn{validations.IsPositive}, nil, true, nil)
 	OpenshiftEnv      = createConfigSetting("openshift-env", nil, nil, nil, false, nil)
-	Metrics           = createConfigSetting("metrics", SetBool, nil, nil, true, nil)
-	Logging           = createConfigSetting("logging", SetBool, nil, nil, true, nil)
 
 	// future enabled flags
-	ServiceCatalog      = createConfigSetting("service-catalog", SetBool, nil, nil, true, nil)
 	ExtraClusterUpFlags = createConfigSetting("extra-clusterup-flags", SetString, nil, nil, true, nil)
 
 	// Setting proxy
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -605,9 +605,6 @@ func initClusterUpFlags() *flag.FlagSet {
 	clusterUpFlagSet.AddFlag(cmdUtil.HttpsProxyFlag)
 
 	if minishiftConfig.EnableExperimental {
-		clusterUpFlagSet.Bool(configCmd.Metrics.Name, false, "Install metrics (experimental)")
-		clusterUpFlagSet.Bool(configCmd.Logging.Name, false, "Install logging (experimental)")
-		clusterUpFlagSet.Bool(configCmd.ServiceCatalog.Name, false, "Install service catalog (experimental)")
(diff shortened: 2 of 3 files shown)


praveenkumar:
@gbraad Since now your removed the flags so might be need to remove those things from the integration feature file also, currently CI is failing for same.

cmd/minishift/cmd/start.go
```diff
@@ -605,9 +605,6 @@ func initClusterUpFlags() *flag.FlagSet {
 	clusterUpFlagSet.AddFlag(cmdUtil.HttpsProxyFlag)
 
 	if minishiftConfig.EnableExperimental {
-		clusterUpFlagSet.Bool(configCmd.Metrics.Name, false, "Install metrics (experimental)")
-		clusterUpFlagSet.Bool(configCmd.Logging.Name, false, "Install logging (experimental)")
-		clusterUpFlagSet.Bool(configCmd.ServiceCatalog.Name, false, "Install service catalog (experimental)")
```
praveenkumar: We also need to update the integration test case as well, as of now failing for CI.

gbraad: addressed

cmd/minishift/cmd/config/config.go
```diff
@@ -71,11 +71,8 @@ var (
 	HostPvDir         = createConfigSetting("host-pv-dir", SetString, []setFn{validations.IsValidPath}, nil, true, nil)
 	ServerLogLevel    = createConfigSetting("server-loglevel", SetInt, []setFn{validations.IsPositive}, nil, true, nil)
 	OpenshiftEnv      = createConfigSetting("openshift-env", nil, nil, nil, false, nil)
-	Metrics           = createConfigSetting("metrics", SetBool, nil, nil, true, nil)
-	Logging           = createConfigSetting("logging", SetBool, nil, nil, true, nil)
```
praveenkumar: 👍

cmd/minishift/cmd/config/config.go
```diff
@@ -71,11 +71,8 @@ var (
...
 
 	// future enabled flags
-	ServiceCatalog      = createConfigSetting("service-catalog", SetBool, nil, nil, true, nil)
```
praveenkumar: 👍

test/integration/features/cmd-config.feature
```diff
@@ -102,19 +102,19 @@ user defined options which changes default behaviour of Minishift.
...
+    | skip-registry-check | t          | true     |
+    | skip-registration   | T          | true     |
+    | skip-registry-check | 1          | true     |
+    | skip-registration   | false      | false    |
+    | skip-registry-check | False      | false    |
+    | skip-registration   | FALSE      | false    |
+    | skip-registry-check | f          | false    |
+    | skip-registration   | F          | false    |
+    | skip-registry-check | 0          | false    |
+    | skip-registration   | f          | false    |
```
praveenkumar: 👍

test/integration/features/cmd-config.feature
```diff
@@ -27,9 +27,9 @@ user defined options which changes default behaviour of Minishift.
 
   Examples: Wrong values for boolean keys
     | property            | value      |
-    | logging             | TRuE       |
-    | metrics             | positive   |
-    | metrics             | yes        |
+    | skip-registration   | TRuE       |
```
budhrg: :+1:

LalatenduMohanty:
@minishift/minishift-dev AFAIK Users need these features need these from Minishift. So IMO we need to provide them in future. May be after the `oc cluster up` restructure is done. I am not sure if we will provide add-ons or flags with `minishift start` at this point.  Just wanted to make sure we agree to this before merging this.

gbraad:
Since they are experimental, it is possible for us to remove a feature when
we see fit.
At the moment we have to many issues resulting from this, and we have no
support story.

LalatenduMohanty:
> I doubt if WE can ever fix this. it would better to solve this as an addon and pass the responsibility to the actual implementers.

I do not think we have any option here. But we will get the input from concerned people.

Issue #1628, opened 2017-10-27
Preflight: HTTP connectivity check should use proxy settings


gbraad:
It seems the pre-flight check for HTTP connectivity does not use the proxy settings. This is due to libcurl not using httpsx_proxy environment variables. As part of https://github.com/minishift/minishift/issues/1534#issuecomment-335534679 this was highlighted. We likely have to implement the following:

`CURLOPT_PROXY`, `CURLOPT_PROXYUSERPWD` and `CURLOPT_PROXYTYPE` and set like `CURLPROXY_HTTPS` => https://curl.haxx.se/libcurl/c/CURLOPT_PROXYTYPE.html

anjannath:
Fix in #2250

gbraad:
Thanks. Can you ask him to use this issue number instead?

On Saturday, April 7, 2018, Anjan Nath <notifications@github.com> wrote:
> Fix in #2250
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly, view it on GitHub, or mute the thread.<
https://ci5.googleusercontent.com/proxy/6Duy2aHLeDus1PcqeIR-26Iqu0pWVIp2VGWx2AEe3McFW9YHnk1f2MFYa4UW2SmtOySdM4HiprmghatKV3sk5y9I0plCJu2vZwUpmQmxliB7Mn2Eo054WAE305EK63y2WkQzSmVxzFbnQFOnu0MkN5l6-oWH4g=s0-d-e1-ft#https://github.com/notifications/beacon/AAAHZkrHHe0Nk5F4bTVcxflcqqSmwCm4ks5tmLcEgaJpZM4QIgyl.gif
>

-- 

   Gerard Braad | http://gbraad.nl
   [ Doing Open Source Matters ]

svenstaro:
This can now be closed.

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/pull/2250

Issue #2245, opened 2018-04-05
Issue #1628 Pre-flight connectivity check should honor proxy settings


svenstaro:
Duplicate of #1628

---

~~In relation to #2244, the internet connectivity check should honor proxy settings.~~

praveenkumar:
@svenstaro I am not sure how you are running the minishift but check the following logs, for me, it's working. Can you paste log output which might help us to find what causing the issue for you?

```
$ ./minishift start --https-proxy=https://xxx:xxx@x.x.x.x:xxx
-- Starting profile 'minishift'
-- Checking if requested OpenShift version 'v3.9.0' is valid ... OK
-- Checking if requested OpenShift version 'v3.9.0' is supported ... OK
-- Checking if requested hypervisor 'kvm' is supported on this platform ... OK
-- Checking if KVM driver is installed ... 
   Driver is available at /usr/local/bin/docker-machine-driver-kvm ... 
   Checking driver binary is executable ... OK
-- Checking if Libvirt is installed ... OK
-- Checking if Libvirt default network is present ... OK
```

anjannath:
duplicate of #1628

gbraad:
Sorry to be picky. Can we close this issue instead as invalid, as the skip
is available, it was a warning by design for this purpose and it is
documented.

On Saturday, April 7, 2018, Anjan Nath <notifications@github.com> wrote:
> duplicate of #1628
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.<
https://ci5.googleusercontent.com/proxy/dZpR9zFy3hAttMVGi1YkxzrrlJPB1sxC7ouEkaxvkDXb5WaWldkrlmXfdJGRNwJIOcO6uZFmtUgqbTwVaSmGMwXYLzyFQrTf5HYsKB1mk-0mLjiXq0RNnYvYj5tmJtqC9e-Jcz3A0zzR440JGLSDENfMPa2asA=s0-d-e1-ft#https://github.com/notifications/beacon/AAAHZmzgPqMkesK73Xguq_PQi8SDz5yoks5tmLcegaJpZM4TIf8q.gif
>

-- 

   Gerard Braad | http://gbraad.nl
   [ Doing Open Source Matters ]

svenstaro:
I don't think it's technically speaking the same issue. The thing I fixed was only implemented a few days ago. Could you double check?

LalatenduMohanty:
[...]

PR #2250, opened 2018-04-07
Issue #1628 Fix Internet connectivity check to honor proxy settings and use HTTP

svenstaro:
`net.Dial("tcp", address)` did not use the proxy settings which led it to wrongly report a faulty Internet connection when in reality it just didn't know about the proxies.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -168,6 +168,8 @@ func runStart(cmd *cobra.Command, args []string) {
 	// we need to determine whether this is a restart prior to potentially creating a new VM
 	isRestart := cmdUtil.VMExists(libMachineClient, constants.MachineName)
 
+	proxyConfig := handleProxies()
+
 	// preflight check (before start)
 	preflightChecksBeforeStartingHost()
 
@@ -177,8 +179,6 @@ func runStart(cmd *cobra.Command, args []string) {
 	preflightChecksForArtifacts()
 
 	setSubscriptionManagerParameters()
-
-	proxyConfig := handleProxies()
 
 	fmt.Print("-- Starting local OpenShift cluster")
 
--- a/cmd/minishift/cmd/start_preflight.go
+++ b/cmd/minishift/cmd/start_preflight.go
@@ -37,7 +37,7 @@ import (
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/viper"
 
-	cmdUtils "github.com/minishift/minishift/cmd/minishift/cmd/util"
+	cmdUtil "github.com/minishift/minishift/cmd/minishift/cmd/util"
 	minishiftNetwork "github.com/minishift/minishift/pkg/minishift/network"
 	openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"
 	stringUtils "github.com/minishift/minishift/pkg/util/strings"
@@ -45,7 +45,7 @@ import (
 
 const (
 	StorageDisk   = "/mnt/?da1"
-	GithubAddress = "github.com:http"
+	GithubAddress = "https://github.com"
 )
 
 // preflightChecksBeforeStartingHost is executed before the startHost function.
@@ -53,7 +53,22 @@ func preflightChecksBeforeStartingHost() {
 	driverErrorMessage := "See the 'Setting Up the Driver Plug-in' topic (https://docs.openshift.org/latest/minishift/getting-started/setting-up-driver-plugin.html) for more information"
 	prerequisiteErrorMessage := "See the 'Installing Prerequisites for Minishift' topic (https://docs.openshift.org/latest/minishift/getting-started/installing.html#install-prerequisites) for more information"
 
+	proxy := os.Getenv("HTTPS_PROXY")
+
+	// In case we're not running with -v5, we want to obscure the proxy but
+	// still tell the user whether or not a proxy is being used. This is for
+	// security reasons as proxy URLs might contain credentials thay should not
+	// always be shown.
+	if !glog.V(5) {
+		if os.Getenv("HTTPS_PROXY") == "" {
+			proxy = "No"
+		} else {
+			proxy = "Yes"
+		}
+	}
+	fmt.Printf("-- Checking if %s is reachable (using proxy: %q) ... ", GithubAddress, proxy)
 	if network.CheckInternetConnectivity(GithubAddress) {
+		fmt.Printf("OK\n")
 		preflightCheckSucceedsOrFails(
 			configCmd.SkipCheckOpenShiftRelease.Name,
 			checkOriginRelease,
@@ -62,10 +77,8 @@ func preflightChecksBeforeStartingHost() {
 			fmt.Sprintf("%s is not a valid OpenShift version", viper.GetString(configCmd.OpenshiftVersion.Name)),
 		)
 	} else {
+		fmt.Printf("FAIL\n")
 		fmt.Printf("-- Checking if requested OpenShift version '%s' is valid ... SKIP\n", viper.GetString(configCmd.OpenshiftVersion.Name))
-		if glog.V(2) {
-			fmt.Println("github.com is not accessible over internet or host is not connected to internet")
-		}
 	}
 
 	preflightCheckSucceedsOrFails(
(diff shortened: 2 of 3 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -168,6 +168,8 @@ func runStart(cmd *cobra.Command, args []string) {
 	// we need to determine whether this is a restart prior to potentially creating a new VM
 	isRestart := cmdUtil.VMExists(libMachineClient, constants.MachineName)
 
+	proxyConfig := handleProxies()
```
svenstaro: This had to be moved in front of `preflightChecksBeforeStartingHost()` so that it would set the proxy settings in time.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -37,23 +36,25 @@ import (
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/spf13/viper"
 
-	cmdUtils "github.com/minishift/minishift/cmd/minishift/cmd/util"
+	cmdUtil "github.com/minishift/minishift/cmd/minishift/cmd/util"
```
svenstaro: This was inconsistently named with how it was imported everywhere else in the project and so I just fixed it here.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -37,23 +36,25 @@ import (
...
 	minishiftNetwork "github.com/minishift/minishift/pkg/minishift/network"
 	openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"
 	stringUtils "github.com/minishift/minishift/pkg/util/strings"
 )
 
 const (
 	StorageDisk   = "/mnt/?da1"
-	GithubAddress = "github.com:http"
+	GithubAddress = "https://github.com"
```
svenstaro: It's important that we actually try to connect via `https` because that's also what we'll use for all the github.com communication later on. `http` is not enough!

pkg/minishift/network/settings.go
```diff
@@ -312,7 +312,7 @@ func WriteNetworkSettingsToInstance(driver drivers.Driver, networkSettings Netwo
 
 // CheckInternetConnectivity return false if user is not connected to internet
 func CheckInternetConnectivity(address string) bool {
-	_, err := net.Dial("tcp", address)
+	_, err := http.Get(address)
```
svenstaro: This is really the most meat of this commit as `http.Get` actually knows about proxies set in the environment.

pkg/util/proxy.go
```diff
@@ -39,7 +39,10 @@ type ProxyConfig struct {
 // the corresponding environment variable is checked.
 func NewProxyConfig(httpProxy string, httpsProxy string, noProxy string) (*ProxyConfig, error) {
 	if httpProxy == "" {
-		httpProxy = os.Getenv("HTTP_PROXY")
+		httpProxy = os.Getenv("http_proxy")
```
svenstaro: This is technically out of scope but I hope that we can still get this in here: Some tools accept a lower case `http_proxy` env var as well an upper case `HTTP_PROXY` and friends. In curl, lower case env vars take precedence over upper case ones.

anjannath:
add to whitelist

gbraad:
Duplicate issue. Originally filed as an improvement. Please locate, so this
can also he closed ...


-- 

   [ Doing Open Source Matters ]

pkg/util/proxy.go
gbraad: Please address as a separate issue

anjannath:
@svenstaro Please refer to issue #1628 instead of #2245.

svenstaro:
I'll copy this from my comment in issue #2245:

> I don't think it's technically speaking the same issue. The thing I fixed was only implemented a few days ago. Could you double check?

svenstaro:
Specifically, this fixes https://github.com/minishift/minishift/pull/2149 which does not work in a proxy environment despite the comment in the PR telling us that it does!

gbraad:
This fixes issue #1628... please update the description accordingly

cmd/minishift/cmd/start_preflight.go
gbraad: OK

svenstaro:
Ok, done.

gbraad:
The commenst was originally there and it was discussed. Seems we missed this somehow... Thanks!

cmd/minishift/cmd/start_preflight.go
```diff
@@ -37,23 +36,25 @@ import (
...
 )
 
 // preflightChecksBeforeStartingHost is executed before the startHost function.
 func preflightChecksBeforeStartingHost() {
 	driverErrorMessage := "See the 'Setting Up the Driver Plug-in' topic (https://docs.openshift.org/latest/minishift/getting-started/setting-up-driver-plugin.html) for more information"
 	prerequisiteErrorMessage := "See the 'Installing Prerequisites for Minishift' topic (https://docs.openshift.org/latest/minishift/getting-started/installing.html#install-prerequisites) for more information"
 
+	fmt.Printf("-- Checking if %s is reachable (using proxy: %q) ... ", GithubAddress, os.Getenv("HTTPS_PROXY"))
```
praveenkumar: Doesn't this printout every time no matter if the proxy is set or not? We can put this behind an condition where we first check if there is PROXY env set or not.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -62,10 +63,8 @@ func preflightChecksBeforeStartingHost() {
 			fmt.Sprintf("%s is not a valid OpenShift version", viper.GetString(configCmd.OpenshiftVersion.Name)),
 		)
 	} else {
+		fmt.Printf("FAIL\n")
 		fmt.Printf("-- Checking if requested OpenShift version '%s' is valid ... SKIP\n", viper.GetString(configCmd.OpenshiftVersion.Name))
-		if glog.V(2) {
-			fmt.Println("github.com is not accessible over internet or host is not connected to internet")
```
praveenkumar: This we put just in case if internet connectivity is there but `github.com` server is down. I don't this it's good to remove it just in case we want get --v 2 logs from user.

cmd/minishift/cmd/start_preflight.go
svenstaro: Yes, it does but I think that it's good to be explicitly made aware of the fact that no proxy is in use at that point. Now, of course, this is a somewhat selfish use case because I use proxies and I consider it likewise helpful to see which proxy I am using and whether I'm using any proxy at all.

I vote for keeping some time of indication around in this line that will tell you that you are not using a proxy. I think considering that this meant to be an Internet connectivity check, we can justify keeping a proxy message around.

cmd/minishift/cmd/start_preflight.go
svenstaro: The indication is still here. Have a look a few lines around this line. It will now instead append `FAIL` or `OK` to the end of the Internet connectivity check line. This is more consistent with how the other status lines work.

cmd/minishift/cmd/start_preflight.go
praveenkumar: @svenstaro I am all fine with having it part of explicit but then also I would not print out the `HTTPS_PROXY` variable directly to stdout because most of the time it contains the proxy info along with credential and during the demo or presenting to someone might expose this :(

svenstaro: Good point, will fix.

gbraad:
if you want to print the proxy, maybe stripping credential info might be an
option... WDYT?

svenstaro:
I gave it some thought and I will only print whether or not we're using a proxy at all when no -v is given. I think this is the most conservative option that will surprise the least number of people.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -37,23 +37,38 @@ import (
...
+	// still tell the user whether or not a proxy is being used. This is for
+	// security reasons as proxy URLs might contain credentials thay should not
+	// always be shown.
+	if !glog.V(5) {
+		if os.Getenv("HTTPS_PROXY") == "" {
+			proxy = "No"
+		} else {
+			proxy = "Yes"
+		}
+	}
```
praveenkumar: :+1:

svenstaro:
Looks like this is good to go. Can we merge this?

cmd/minishift/cmd/start_preflight.go
```diff
@@ -37,23 +37,38 @@ import (
...
+	fmt.Printf("-- Checking if %s is reachable (using proxy: %q) ... ", GithubAddress, proxy)
 	if network.CheckInternetConnectivity(GithubAddress) {
+		fmt.Printf("OK\n")
```
gbraad: you could use `fmt.Println("OK")` if newline is needed.

svenstaro: I did it like this for consistency's sake as line 81 already does it using `fmt.Printf("\n")`.

Issue #2126, opened 2018-03-14
Write integration test for SSHFS host folder 


hferentschik:
Something along the lines:

* minishift hostfolder list -> nothing there
* minishift hostfolder add ... -> host folder defined
* minishift hostfolder list -> host folder defined, but not mounted
* minishift hostfolder mount -> host folder gets mounted, runs some commands that the file system is mounted and writable, etc
* minishift hostfolder list -> host folder defined and mounted
* minishift hostfolder umount
* minishift hostfolder delete

LalatenduMohanty:
Resolved #2497

PR #2257, opened 2018-04-10
Issue #2123 Key based authentication for sftpd server

### LINKED ISSUE #2123: sftpd server should do proper key based authentication
Right now the code in `sftpd.go` accepts any public key. To make things more secure, we should implement an authorized key mechanism. See also https://github.com/golang/crypto/blob/master/ssh/example_test.go.

We already create a key pair on the VM as part of the sshfs host folder setup. The public key should be copied to a known location within the MINISHIFT_HOME directory. Either as dedicated file, for example _sftp-authorized-keys_ or one adds the keys to the all instances config. 

I guess one way of cleanup is needed as well.

budhrg:
Fix #2123 

I hope integration test will be added as part of https://github.com/minishift/minishift/issues/2126.

### DIFF
--- a/cmd/minishift/cmd/hostfolder/sftpd.go
+++ b/cmd/minishift/cmd/hostfolder/sftpd.go
@@ -17,19 +17,18 @@ limitations under the License.
 package hostfolder
 
 import (
-	"crypto/rand"
-	"crypto/rsa"
-	"crypto/x509"
-	"encoding/pem"
 	"fmt"
 	"github.com/golang/glog"
 	"github.com/minishift/minishift/cmd/minishift/cmd/config"
+	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
+	"github.com/minishift/minishift/pkg/util/filehelper"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/pkg/sftp"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
 	"golang.org/x/crypto/ssh"
 	"io"
+	"io/ioutil"
 	"net"
 	"os"
 	"sync/atomic"
@@ -50,16 +49,18 @@ var (
 		Run:    runSftp,
 		Hidden: true,
 	}
+
+	authorizedKeysMap map[string]bool
 )
 
 func init() {
+	authorizedKeysMap = make(map[string]bool)
 	hostFolderSSHDCmd.Flags().IntVarP(&serverPort, serverPortFlag, "p", 2022, "The server port.")
 	HostFolderCmd.AddCommand(hostFolderSSHDCmd)
 }
 
 func runSftp(cmd *cobra.Command, args []string) {
 	serverConfig := serverConfig()
-
 	port := viper.GetInt(config.HostFoldersSftpPort.Name)
 	if port == 0 {
 		port = serverPort
@@ -82,8 +83,9 @@ func serveConnections(listener net.Listener, serverConfig *ssh.ServerConfig) {
 			glog.Fatal("failed to accept incoming connection", err)
 		}
 
-		// Before use, a handshake must be performed on the incoming
-		// net.Conn.
+		// populate authorized keys for handshake
+		populateAuthorizedKeysMap()
+		// Before use, a handshake must be performed on the incoming net.Conn.
 		_, channels, requests, err := ssh.NewServerConn(nConn, serverConfig)
 		if err != nil {
 			glog.Fatal("failed to handshake", err)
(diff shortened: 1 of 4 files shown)


pkg/minikube/constants/constants.go
```diff
@@ -85,3 +85,13 @@ func GetMinishiftHomeDir() string {
 		return filepath.Join(util.HomeDir(), ".minishift")
 	}
 }
+
+// ProfileAuthorizedKeysPath returns the path of authorized_keys file in profile dir used for authentication purpose
```
gbraad: Why does this have to be in the minikube package?
And these, like previous one's are surely not constant at all...

Can we not have a separate `minishift`-like profile variables function in our own minishift namespace?

gbraad:
I am slightly confused about the 'proper' key based authentication part.
Which key are we using to authenticate?

gbraad:
OK, the key is profile specific...

pkg/minishift/hostfolder/sshfs_hostfolder.go
```diff
@@ -33,7 +35,7 @@ import (
 )
 
 const (
-	keyFile = "/home/docker/.ssh/id_rsa"
+	KeyFile = "/home/docker/.ssh/id_rsa"
```
anjannath: Why export this?

budhrg: Ah, was exporting to be used in another file. Not needed now, reversing now.

budhrg: Done

anjannath:
Uhh.. Appveyor, failing like usual..

gbraad:
AppFailure

budhrg:
> Uhh.. Appveyor, failing like usual..

Re-triggered

anjannath:
> Which key are we using to authenticate?

The key in `/home/docker/.ssh` used to ssh into the vm.

pkg/minishift/constants/constants.go
```diff
@@ -30,3 +35,13 @@ const (
 var (
 	ValidIsoAliases = []string{B2dIsoAlias, CentOsIsoAlias, MinikubeIsoAlias}
 )
+
+// ProfileAuthorizedKeysPath returns the path of authorized_keys file in profile dir used for authentication purpose
+func ProfileAuthorizedKeysPath() string {
+	return filepath.Join(constants.Minipath, "authorized_keys")
```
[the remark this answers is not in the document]
praveenkumar: Even function name can be changed to ` ProfilePublicKeyPath()`

budhrg:
> The key in /home/docker/.ssh used to ssh into the vm.

No. This private key from `/home/docker/.ssh/id_rsa` is being used to authenticate the public key present in `PROFILE_DIR/authorized_keys` file and it is being copied in `PROFILE_DIR/certs/id_rsa` file during mount operation since sftp server doesn't interact with VM.

pkg/minishift/constants/constants.go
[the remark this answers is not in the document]
anjannath: I think this is the `authorized_keys` file which contains the public keys related to users that are authorized to ssh into the server. This can be put inside the certs dir... but the file name should not be changed.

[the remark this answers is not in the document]
budhrg: > Even function name can be changed to ProfilePublicKeyPath()

Function name can't be changed as it will confuse the reader of code since ssh authentication include private and authorized_keys file.

> return filepath.Join(constants.Minipath, "certs" ,"id_rsa.pub")

Sure.

cmd/minishift/cmd/hostfolder/sftpd.go
```diff
@@ -159,47 +161,54 @@ func serveConnections(listener net.Listener, serverConfig *ssh.ServerConfig) {
...
+}
 
-	pem := pem.EncodeToMemory(
-		&pem.Block{
-			Type:  "RSA PRIVATE KEY",
-			Bytes: x509.MarshalPKCS1PrivateKey(key),
-		},
-	)
-	return pem, nil
+func keyAuth(conn ssh.ConnMetadata, key ssh.PublicKey) (*ssh.Permissions, error) {
```
anjannath: What is the purpose of `conn ssh.Connmetadata` in this function, the formation of  `ssh.Permissions` does not seem to be dependent on it, only used for the error??

budhrg:
retest this please

cmd/minishift/cmd/hostfolder/sftpd.go
budhrg: Because, its a callback function and function signature is fixed
https://github.com/golang/crypto/blob/d6449816ce06963d9d136eee5a56fca5b0616e7e/ssh/server.go#L65-L67
```
// PasswordCallback, if non-nil, is called when a user
// attempts to authenticate using a password.
PasswordCallback func(conn ConnMetadata, password []byte) (*Permissions, error)
```

budhrg:
@anjannath , Addressed both the comments. Thanks for the last one where I need to revert the `'StrictHostKeyChecking=no'`

anjannath:
@budhrg Ahh... Happy to be of service :)

cmd/minishift/cmd/hostfolder/sftpd.go
anjannath: Okay..

pkg/minishift/constants/constants.go
```diff
@@ -30,3 +35,13 @@ const (
...
 	ValidIsoAliases = []string{B2dIsoAlias, CentOsIsoAlias, MinikubeIsoAlias}
 )
+
+// ProfileAuthorizedKeysPath returns the path of authorized_keys file in profile dir used for authentication purpose
+func ProfileAuthorizedKeysPath() string {
+	return filepath.Join(constants.Minipath, "certs", "authorized_keys")
+}
+
+// ProfilePrivateKeyPath returns the path of private key of VM present in profile dir which is used for authentication purpose
+func ProfilePrivateKeyPath() string {
```
LalatenduMohanty: @budhrg  I suggest the function name to be something like `RSAPrivateKeyPath()`. We do not need to keep `profile` name in the function name as it is understood that it needs to be per profile as value of constants.Minipath changes per profile.

pkg/minishift/hostfolder/sshfs_hostfolder.go
```diff
@@ -54,13 +56,14 @@ func (h *SSHFSHostFolder) Config() config.HostFolderConfig {
 }
 
 func (h *SSHFSHostFolder) Mount(driver drivers.Driver) error {
-	err := h.ensureSFTPDDaemonRunning()
-	if err != nil {
+	if err := h.ensureRSAKeyExists(driver); err != nil {
+		return err
+	}
+	if err := h.cacheRSAKeys(driver); err != nil {
```
LalatenduMohanty: 👍

pkg/minishift/constants/constants.go
```diff
@@ -30,3 +35,13 @@ const (
 var (
 	ValidIsoAliases = []string{B2dIsoAlias, CentOsIsoAlias, MinikubeIsoAlias}
 )
+
+// ProfileAuthorizedKeysPath returns the path of authorized_keys file in profile dir used for authentication purpose
+func ProfileAuthorizedKeysPath() string {
+	return filepath.Join(constants.Minipath, "certs", "authorized_keys")
+}
```
LalatenduMohanty: Suggestion: RSAAuthorizedKeysPath()

pkg/minishift/constants/constants.go
budhrg: > We do not need to keep profile name in the function name as it is understood that it needs to be per profile as value of constants.Minipath changes per profile.

It will be more readable from different package and make more sense. `RSAPrivateKeyPath` sounds ambiguous. One need to go to function definition to understand it is actually some file under profile dir.

anjannath:
@budhrg Rebase needed..

budhrg:
> Rebase needed..

Not needed. We do `Rebase and merge` now and will be taken care that time.

anjannath:
> Not needed. We do Rebase and merge now and will be taken care that time.

Ohh.. Okay :)

PR #2296, opened 2018-04-16
Issue #2294 Update centos iso from v1.8.0 to v1.9.0

praveenkumar:


budhrg:
We need to run this PR with minikube ISO

LalatenduMohanty:
Deployed an application and it worked fine.

PR #2299, opened 2018-04-17
Fix sed command

svenstaro:
Currently, this sed command uses sudo which is then redirected to `/etc/libvirt/qemu.conf` but that won't work as the sudo is only applied to the sed but not the redirection. However, sed has a built-in way of directly acting on files and that's the `-i` flag.


svenstaro:
retest this please

anjannath:
@svenstaro Please see https://github.com/minishift/minishift/issues/2300#issuecomment-381852621

docs/source/getting-started/setting-up-virtualization-environment.adoc
```diff
@@ -111,6 +111,18 @@ $ sudo sed -r 's/group=".+"/group="kvm"/1' /etc/libvirt/qemu.conf > /etc/libvirt
 $ newgrp libvirt
 ----
 
+. Start the *default* libvirt network:
++
+----
+$ sudo virsh net-start default
+----
+
+. For convenience it's useful to mark the *default* network as autostart:
```
gbraad: convenience? I'd say this is rather essential.. but I thought the standard definition did that...

gbraad: @robin-owen This is a very generic topic... the only difference might be the location of the config file... ? right?

docs/source/getting-started/setting-up-virtualization-environment.adoc
```diff
@@ -102,7 +102,7 @@ $ sudo usermod -a -G kvm,libvirt <username>
 . Update the *libvirt* configuration in *_/etc/libvirt/qemu.conf_* as follows:
 +
 ----
-$ sudo sed -r 's/group=".+"/group="kvm"/1' /etc/libvirt/qemu.conf > /etc/libvirt/qemu.conf
+$ sudo sed -ri 's/.group = ".+"/group = "kvm"/1' /etc/libvirt/qemu.conf
```
robin-owen: Already mentioned in IRC, but I'm wary of the fragility of this `sed` expression. In particular, `.group` will match any single character before 'group', but not zero characters -- likewise, the white space is fragile if it changes again in the future.

:+1: for the in-place modification!

docs/source/getting-started/setting-up-virtualization-environment.adoc
robin-owen: I agree, this is generic information.

In the future, it would be nice if we can make use of AsciiDoc attributes to place these instructions for each distribution with the location of the configuration file, but that lies outside of the scope of this particular issue.

docs/source/getting-started/setting-up-virtualization-environment.adoc
```diff
@@ -350,3 +379,4 @@ If you have already run `minishift start`, ensure that you run `minishift delete
 === Next Steps
 
 {next-steps}
+
```
robin-owen: Is there a reason that this was added?

docs/source/getting-started/setting-up-virtualization-environment.adoc
```diff
@@ -111,6 +111,35 @@ $ sudo sed -r 's/group=".+"/group="kvm"/1' /etc/libvirt/qemu.conf > /etc/libvirt
 $ newgrp libvirt
 ----
 
+[[kvm-driver-networking]]
+=== Configure libvirt networking
+
+On some distributions, the *default* libvirt network will already be started upon installation.
+On others, you'll have to start it manually.
```
gbraad: Feels like repetition... 'on'... 'on'... "Some distributions setup the default network for you, while on others this might have to be done manually. Below you will find instructions to define the network." or somethimg similar.

gbraad:
@svenstaro it is not about me being happy, but about consistency and tone...

anjannath:
Thanks @svenstaro :tada: :balloon:

Issue #2313, opened 2018-04-18
Documentation uses already referenced anchor IDs


robin-owen:
The anchor IDs `installing-minishift` and `adding-cifs-hostfolder` are both used twice in the documentation where unique anchor IDs should be used instead. The `installing-minishift` anchor is used in both **_getting-started/preparing-to-install.adoc_** and **_getting-started/installing.adoc_**. Both uses of `adding-cifs-hostfolder` exist in **_using/host-folders.adoc_**.

LalatenduMohanty:
Resolved via #2314

PR #2314, opened 2018-04-18
Issue #2313 amend duplicate anchors in documentation

robin-owen:


docs/source/getting-started/preparing-to-install.adoc
```diff
@@ -1,6 +1,6 @@
 include::variables.adoc[]
 
-[[installing-minishift]]
+[[preparing-install-minishift]]
```
LalatenduMohanty: @robin-owen What about `preparing-minishift-install`

robin-owen: @LalatenduMohanty I'm trying to keep the anchors similar to the titles they refer to. Would you prefer that the title of this section be changed generally?

LalatenduMohanty: @robin-owen Not really, I think I would prefer to keep them similar. Just that it sound grammatically incorrect leaving bad taste in mouth. If you feel strongly that the current form is fine then I can go ahead and merge this.

robin-owen:
@LalatenduMohanty Changed the anchor name.

LalatenduMohanty:
Thanks @robin-owen

PR #2338, opened 2018-04-26
Issue #2337 Add test step to delete the minishift home dir

### LINKED ISSUE #2337: Add test step to delete the minishift home dir
We need test step as
```
And deleting minishift home directory "test" succeeds
```

budhrg:
Fix #2337

Require for https://issues.jboss.org/browse/CDK-271. Also, potential use case for upstream as well.

### DIFF
--- a/test/integration/testsuite/minishift.go
+++ b/test/integration/testsuite/minishift.go
@@ -26,6 +26,8 @@ import (
 	"time"
 
 	"github.com/minishift/minishift/test/integration/util"
+
+	"os"
 )
 
 var commandOutputs []CommandOutput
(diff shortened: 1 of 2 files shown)


agajdosi:
These steps starts their path directly from integration-test.go pwd, so to delete in out/integration-test you need to specify the path like this:

```
  Scenario: User is able to retrieve host and port of OpenShift registry
    When file from "https://raw.githubusercontent.com/minishift/minishift/master/addons/anyuid/anyuid.addon" is downloaded into location "test"
    When deleting minishift home directory "../../out/integration-test/test" succeeds
    Then minishift home directory "../../out/integration-test/test" shouldn't exist
```

This is however expected right now. LGMT.

budhrg:
@LalatenduMohanty Merging as need to downstream tests.

gbraad:
We should abstract those paths. They shouldn't have to be known to the test feature, only as a simple identifier.

Issue #2323, opened 2018-04-20
Update Hyper-V static IP documentation as it is not an experimental feature


LalatenduMohanty:
Currently the docs are in experimental section of the documentation [1] but the feature is not an experimental feature. 
[1] https://docs.openshift.org/latest/minishift/using/experimental-features.html#hyperv-static-ip

robin-owen:
Right, this feels like it belongs in its own dedicated section under **_using/_**, for example **_using/static-ip.adoc_** (or something along those lines). This is particularly true if the feature will be extended to other hypervisors in the future.

gbraad:
I expect several network related topics to be discussed, like IP, DNS and
possibly how to setup port-forwarding. However, this is mostly future...

... although it does show "networking' is more close to the topic we look
for.

robin-owen:
@gbraad Networking makes sense to me, especially with so many related topics.

LalatenduMohanty:
> using/static-ip.adoc

+1 . We can move docs related to `--set-staic` to this too (in future).

robin-owen:
@LalatenduMohanty @gbraad I think we can go ahead and create **_using/networking.adoc_** with a single topic (setting a static IP for Hyper-V) for the moment. I'd like to create this file early so that we don't have URL changes in the future if at all possible.

LalatenduMohanty:
@gbraad @robin-owen I strongly feel that docs related to static IP and (--set-static in future together ) deserves a dedicated page because it is very important from Minishift users point of view similar to add-ons or profiles.

LalatenduMohanty:
Resolved via #2341

PR #2341, opened 2018-04-26
Issue #2323 Update Hyper-V static IP documentation

robin-owen:
This commit also provides minor changes to **_using/experimental-features.adoc_** for consistency.

### DIFF
--- a/docs/source/using/con_static-ip-overview.adoc
+++ b/docs/source/using/con_static-ip-overview.adoc
@@ -1,0 +1,7 @@
+Most hypervisors do not support extending the lease time when the IP is assigned using DHCP.
+This might lead to a new IP being assigned to the VM after a restart as it will conflict with the security certificates generated for the old IP.
+This will make {project} completely unusable until a new instance is set up by running `minishift delete` followed by `minishift start`.
+
+To prevent this, {project} includes the functionality to set a static IP address to the VM.
+This will prevent the IP address from changing between restarts.
+However, it will not work on all of the driver plug-ins at the moment due to the way the IP address is resolved.
--- a/docs/source/using/static-ip.adoc
+++ b/docs/source/using/static-ip.adoc
@@ -1,0 +1,72 @@
+include::variables.adoc[]
+
+= Assign Static IP Address
+:icons:
+:toc: macro
+:toc-title:
+:toclevels: 1
+
+toc::[]
+
+[[static-ip-overview]]
+== Overview
+
+include::con_static-ip-overview.adoc[]
+
+[NOTE]
+====
+- Assigning a static IP address to the {project} VM is only officially supported for Hyper-V.
+- The {project} VM cannot be assigned a static IP address when using the KVM driver plug-in.
+- An experimental `--set-static` flag is available for the `minishift ip` command for other hypervisors.
+See xref:../using/experimental-features.adoc#set-fixed-ip[Set Fixed IP Address] for more information.
+====
+
+[[static-ip-hyperv]]
+== Assign IP Address to Hyper-V
+
+Since the Internal Virtual Switch for Hyper-V does not provide a DHCP offer option, an IP address needs to be provided in a different way.
+Functionality is provided to assign an IP address on startup using the Data Exchange Service for Hyper-V.
+
+[IMPORTANT]
+====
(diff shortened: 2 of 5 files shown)


docs/source/using/con_static-ip-overview.adoc
```diff
@@ -0,0 +1,7 @@
+Most hypervisors do not support extending the lease time when the IP is assigned using DHCP.
+This might lead to a new IP being assigned to the VM after a restart as it will conflict with the security certificates generated for the old IP.
+This will make {project} completely unusable.
```
gbraad: we might want to include the necessary fix if encountered? `until a minishift delete and minishift start` has been run to setup a completely new instance`.

`To prevent this, {project} includes ...`

budhrg:
retest this please

docs/source/using/con_static-ip-overview.adoc
```diff
@@ -0,0 +1,7 @@
+Most hypervisors do not support extending the lease time when the IP is assigned using DHCP.
+This might lead to a new IP being assigned to the VM after a restart as it will conflict with the security certificates generated for the old IP.
+This will make {project} completely unusable until a new instance is set up by running `minishift delete` followed by `minishift start`.
+
+To prevent this, {project} includes the functionality to set a static IP address to the VM.
+This will prevent the IP address from changing between restarts.
```
gbraad: sounds like duplicate use of `prevent` in a too short timeframe. `This will stop the IP address from ...` although harsh(er) in tone, it is clear in what it does.

docs/source/using/static-ip.adoc
```diff
@@ -0,0 +1,72 @@
...
+WinNAT is limited to one NAT network per host.
+For more details about capabilities and limitations, please see the link:https://blogs.technet.microsoft.com/virtualization/2016/05/25/windows-nat-winnat-capabilities-and-limitations/[WinNAT capabilities and limitations blog].
+====
+
+The following command will attempt to assign an IP address for use on the Internal Virtual Switch 'MyInternal':
+
+----
+PS> $env:HYPERV_VIRTUAL_SWITCH="MyInternal"
+PS> minishift.exe start `
+  --iso-url centos `
```
gbraad: I know this is migrated, but we might not want to mention `centos` here, as this will also end up in the downstream documentation without modifications. WDYT? In a note we have already mentioned it only works on these ISOs

robin-owen: Good point. I briefly thought about this as well while migrating it, but also considered whether the command being run would work upstream without the use of `--iso-url`. Either way is fine with me, whether we make changes here to remove that reference or do that downstream.

gbraad: downstream actually supports the `--iso-url` flag, but it is not necessary as the ISO is included and set as default.

docs/source/using/static-ip.adoc
```diff
@@ -0,0 +1,72 @@
...
+  --network-nameserver 8.8.8.8
+----
+
+If you want to use the 'DockerNAT' network, the following commands are needed to setup the correct NAT networking and assign an IP in the range expected:
+
+----
+PS> New-NetNat -Name SharedNAT -InternalIPInterfaceAddressPrefix 10.0.75.1/24
+PS> $env:HYPERV_VIRTUAL_SWITCH="DockerNAT"
+PS> minishift.exe start `
+  --iso-url centos `
```
gbraad: ditto...

docs/source/using/static-ip.adoc
LalatenduMohanty: @gbraad is right. That means in downstream we should not use the `--iso-url ` flag at all (i.e. examples)

LalatenduMohanty:
@robin-owen Can you please cross check if you are seeing a refernce to `Assign Static IP Address` in the left hand side of the TOC? I can not see it somehow. 

screen shot 2018-05-02 at 5 43 54 pm

robin-owen:
@LalatenduMohanty Just double-checked by running `make clean && make serve_docs`. The reference on the left sidebar appears appropriately for me when checking `localhost:4567`.

LalatenduMohanty:
retest this please

Issue #1157, opened 2017-07-20
When integration test does not end properly it leaves minishift running


agajdosi:
When integration test runner fails - for example on a timeout then minishift.runner.EnsureDeleted() in afterSuite() is not executed and thus the environment is not left clean - minishift instance is running. This can be little impractical in cases of manual use, but is much more in cases of automated tests. In downstream, when nightly tests fail it this manner, then also next nightly tests will fail as minishift start is then not possible.

Would be nice to add check into beforeSuite() to prevent this issue.

gbraad:
In the `BeforeSuite()` you would like to detect if a Minishift instance is still running? I believe there is another problem... I would expect that the timeout should trigger a cleanup action, such as an `OnTimeout()`. I think we should try to keep the integration tests from these startup checks as much as possible,

agajdosi:
@gbraad Often it is because of a network issue. But some fails were also when there were changes with iso building job. It is fine to fail, but then it is wrong to throw away next day's nightly tests. Sometimes I am doing the clean up manually to save the next day, but I should not do that ideally.

I am fine with OnTimeout() if it will work, but I am still curious, what would pre-flight check cost us, or why you want to avoid them? :)

Another possibility would be to have some option to trigger the clean up through some command, so I could do that just on automated tests - but that is the case only if OnTimeout() plan fails.

gbraad:
With CentOS CI, the node is always handed in a clean state. For nightlies
it seems we rerun on the same install? If so, something else needs to
ensure the clean state. It feels weird to me that a build and test it runs
would have to clean up first. This is rather an action that happens at the
end (but I think after the timeout, we are unable to do something). Why not
cronbased? Or a depended task? I am trying to understand the setup. Which
platforms suffer from this, etc

-- 

   Gerard Braad | http://gbraad.nl
   [ Doing Open Source Matters ]

agajdosi:
@gbraad It can happen on all platforms. Yes, this problem can be handled in jenkins job directly, which would lead me to writing a solution for each platform. But another perspective is also: if for any reason, developer, user, qe has some broken forgotten instance of minishift from previous test run still running, can be from terminating the process because you are impatient, what is the meaning of letting her start the tests and watch how they fail.

hferentschik:
@agajdosi Is this still an issue?

agajdosi:
[...]

PR #2340, opened 2018-04-26
Issue #2339 - Chdir to the test-dir as part of the beforeFeature()

agajdosi:
Fix : #2339

PR #2356, opened 2018-05-03
Issue #1157 - If integration test is aborted it leaves minishift running

agajdosi:
This PR makes sure that if test is force quit and there is M left running then on next integration test run the tests will check if M is running and will delete it if needed. Then the M home will be cleaned up.

There is also minor fix for the integration test logger. Which previously had problem with closing, because the logFile was initiated wrongly:
```
logFile, err := func()
```
initiated local copy of logFile and did not used logFile declared at top of the package.


test/integration/testsuite/testsuite.go
```diff
@@ -231,7 +231,10 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^minishift home directory "([^"]*)" shouldn\'t exist$`, MinishiftInstance.minishiftHomeDirectoryShouldntExist)
 
 	s.BeforeSuite(func() {
-		testDir = setUp()
+		testDir = setupTestDirectory()
+		MinishiftInstance.runner.EnsureDeleted()
+		ensureTestDirEmpty()
```
praveenkumar: 👍

test/integration/testsuite/prepare.go
```diff
@@ -102,16 +102,14 @@ func determineIsoFromFile(isoUrl string) string {
 	return isoName
 }
 
-func setUp() string {
+func setupTestDirectory() string {
```
praveenkumar: 👍

praveenkumar:
> The pr is not working on if multiple profile is running for example Interrupt cmd-profile.feature after minishift and foo profile is created. 

@amitkrout was it working before? As per PR concern, I didn't see any change which breaks this kind of functionality.

amitkrout:
> was it working before? As per PR concern, I didn't see any change which breaks this kind of functionality.

@praveenkumar This fix is added first time in minishift through this pr. AFAIU from the description this pr will delete any ```running``` minishift instance (if minishift instance is left behind in the first integration test run due to force stop) in the next integration feature run . As per my test this is working on feature file where we have only one minishift instance but in this case https://github.com/minishift/minishift/pull/2356#pullrequestreview-117560143 its breaking.

How I executed : 

1. make integration GODOG_OPTS="-tags <tagname>"
```
make integration GODOG_OPTS="-tags cmd-profile"
```
2. Interrupt the test once the vm is created or before vm is deleted
```
Force stopped after foo profile is created 
```
3. make integration GODOG_OPTS="-tags <tagname>"
```
make integration GODOG_OPTS="-tags flags"
```

Expected : 
```
Integration test should run without error (VM is already exist)
```
Actual : 
```
Failing because all running minishift instance (minishift and foo) is not deleted in the second integration run 
```

agajdosi:
Good point @amitkrout, what I can do is:

- check all directories in `$MINISHIFT_HOME/profiles`
- get their names
- execute `m profile delete <name-of-folder>`

 WDYT? cc @praveenkumar

agajdosi:
Setting up WIP status as there are some proposals to be considered, also want to merge this after #2340 .

test/integration/testsuite/prepare.go
```diff
@@ -102,16 +102,14 @@ func determineIsoFromFile(isoUrl string) string {
...
 	var err error
 	if testDir == "" {
 		testDir, err = ioutil.TempDir("", "minishift-integration-test-")
 		if err != nil {
 			fmt.Println("Error creating temporary directory:", err)
 			os.Exit(1)
 		}
-	} else {
-		ensureTestDirEmpty()
```
LalatenduMohanty: @agajdosi I think this was a good check to make sure the directory is empty. In case some of the logic to clean up did not work, we can still make sure test should not run if the test directory is not empty.  IMO we should keep it. WDYT?

LalatenduMohanty: @agajdosi Nevermind . my comment was premature.

LalatenduMohanty:
> check all directories in $MINISHIFT_HOME/profiles
> get their names

Why not running `minishift profile list`?

agajdosi:
@LalatenduMohanty +1 for `m profile list`, missed that.

agajdosi:
retest this please

agajdosi:
Suggestions implemented. Integration test in this PR now before and after each test suite run deletes all M instances on all profiles. To secure that also M instances in non-default `out/test-run/.minishift` Minishift home get deleted, the test iterates through all directories in `out/test-run` and if Minishift home is found here, it deletes all instances.

PR is ready for review, ping @minishift/minishift-dev, @budhrg, @amitkrout, @LalatenduMohanty.

test/integration/util/util.go
```diff
@@ -148,13 +151,60 @@ func (m *MinishiftRunner) GetOcRunner() *OcRunner {
...
+
+	for _, line := range lines {
+		re := regexp.MustCompile("- ([.\\S]+)\\s+.+")
+		match := re.FindStringSubmatch(line)
+		if len(match) != 2 {
+			continue
+		}
+
+		profile := match[1]
+		m.RunCommandAndPrintError(fmt.Sprintf("profile set %v", profile))
```
anjannath: @agajdosi Not sure if i understand the purpose of this... why is the profile created here?

anjannath: @agajdosi Understood! Thanks :-)

test/integration/util/util.go
```diff
@@ -148,13 +151,60 @@ func (m *MinishiftRunner) GetOcRunner() *OcRunner {
...
+	}
+}
+
+//EnsureAllProfilesDeleted retrieves all available profiles and deletes all running VMs on them
+func (m *MinishiftRunner) EnsureAllProfilesDeleted() {
+	stdOut, _, _ := m.RunCommandAndPrintError("profile list")
+	lines := strings.Split(stdOut, "\n")
+
+	for _, line := range lines {
+		re := regexp.MustCompile("- ([.\\S]+)\\s+.+")
```
amitkrout: +1

agajdosi:
@LalatenduMohanty ping :)

test/integration/util/util.go
LalatenduMohanty: @agajdosi Why not running `$ minishift profile delete PROFILE_NAME -f` directly instead of `profile set and then delete`.

test/integration/util/util.go
```diff
@@ -148,13 +151,60 @@ func (m *MinishiftRunner) GetOcRunner() *OcRunner {
...
+}
+
+//EnsureDeleted deletes VM instance on currently selected profile
+func (m *MinishiftRunner) EnsureDeleted() error {
 	m.RunCommandAndPrintError("delete --force")
 
 	deleted := m.CheckStatus("Does Not Exist")
 	if deleted == false {
-		fmt.Println("Deletion of minishift instance was not successful! Minishift status is not 'Does Not Exist'.")
+		return errors.New("Deletion of minishift instance was not successful! Minishift status is not 'Does Not Exist'.")
```
LalatenduMohanty: @agajdosi string `Minishift status is not 'Does Not Exist'` is not adding any value IMO. WDYT?

agajdosi: @LalatenduMohanty ok, will remove it and push

Issue #2358, opened 2018-05-04
Update Gitlab CI defintion


gbraad:

anjannath:
Awesome..

gbraad:
https://gitlab.com/gbraad/minishift/-/jobs/66422190

gbraad:
```
--- FAIL: TestNonWritableDirectory (0.00s)
	assertions.go:247: 
                          
	Error Trace:	utils_test.go:229
		
	Error:      	Should be false
		
	Test:       	TestNonWritableDirectory
FAIL
```

gbraad:
Already aware of this, and is being checked for... it is actually the AppVeyor check that was badly formatted.

PR #2359, opened 2018-05-04
[WIP] Issue #2358 Update GitLab CI runner

gbraad:


gbraad:
AppVeyor fails

pkg/util/utils_test.go
```diff
@@ -227,4 +230,11 @@ func TestNonWritableDirectory(t *testing.T) {
 
 	os.Chmod(testDir, 0400) // make dir read-only
 	assert.False(t, IsDirectoryWritable(testDir))
+
+}
+
+func isAppVeyorUser() bool {
+	// Checking for $env:APPVEYOR does not work !
```
praveenkumar: @gbraad Even setting it `APPVEYOR: True` to config file not working? if not then we should remove that addition.

gbraad: I had removed this in a later push... seems it is not at github, but onlu on gitlab. (must have timed out on push)... repushing

pkg/util/utils.go
```diff
@@ -195,3 +196,14 @@ func IsDirectoryWritable(path string) bool {
...
+// IsAdministrativeUser returns true when user is either root or
+// Administrator
+func IsAdministrativeUser() bool {
+	u, _ := user.Current()
+	username := strings.ToLower(u.Username)
+
+	return u.Uid == "1" ||
+		username == "root" ||
+		strings.Contains(username, "administrator")
+}
```
praveenkumar: 👍

gbraad:
Updated to use container from the minishift namespace. Needs https://github.com/CentOS/container-index/pull/331 to be merged and build

.gitlab-ci.yml
```diff
@@ -1,9 +1,8 @@
 build:
-  image: fedora:26
+  image: registry.centos.org/minishift/minishift-build:latest
```
gbraad: Needs https://github.com/CentOS/container-index/pull/331 to be merged

gbraad:
AppVeyor is probably the worst CI we are using... it fails again to just install dependencies...

.gitlab-ci.yml
LalatenduMohanty: @gbraad we should call it `minishift-builder` image , refer https://github.com/CentOS/container-index/pull/331#discussion_r186370023

pkg/util/utils.go
LalatenduMohanty: 👍

gbraad:
how do others feels about this ? @minishift/minishift-dev 

I do not see "builder" really help in naming, as it is not actually a builder, but purely an aid/it provides support for the build process to be possible.

praveenkumar:
> I do not see "builder" really help in naming, as it is not actually a builder, but purely an aid/it provides support for the build process to be possible.

I agree this is not the builder but provide required support to build minishift. To me, the builder is something like s2i which makes the decision around what and how to build. Here, we are just adding the packages which required to build minishift (like go lang and git.).

gbraad:
eventually I will move the `godep` download also into this image... as I
noticed a significant speed increase on gitlab-ci, as the images were
cached during builds.

Issue #2322, opened 2018-04-20
Add integration test step to test experimental feature


LalatenduMohanty:
Currently we do not have any steps in integration test to test experimental features e.g. 
```
When executing "MINISHIFT_ENABLE_EXPERIMENTAL=y minishift start --extra-clusterup-flags '--service-catalog'" succeeds
```

Also adding below code will work in GNU/Linux but not in Windows powershell. 
```code
s.Step(`^executing "MINISHIFT_ENABLE_EXPERIMENTAL=y minishift (.*)" (succeeds|fails)$`,
        ExecutingMinishiftCommandSucceedsOrFails)
```

*We also need to add a test for service catalog. which could not be added in #2319 because of this issue.*

amitkrout:
Thanks @LalatenduMohanty 
Will take it further and make the integration test run on supported terminals

PR #2360, opened 2018-05-04
Issue #2322 Add integration test step to test experimental feature

amitkrout:
Fix : #2322
Executed successfully on bash, powershell and command prompt 
How to test : 
```
make integration GODOG_OPTS="-tags experimental-feature"
```

### DIFF
--- a/test/integration/features/experimental-flags.feature
+++ b/test/integration/features/experimental-flags.feature
@@ -1,0 +1,62 @@
+@experimental-flags
+Feature: Experimental Flags
+  Experimental flag --extra-clusterup-flags will be enabled by setting MINISHIFT_ENABLE_EXPERIMENTAL environment variable,
+  this flag will provide access to some of upcoming feature and experiments.
+
+  Scenario: User cannot start minishift experimental feature directly
+    Given Minishift has state "Does Not Exist"
+     Then executing "minishift start --service-catalog" fails
+      And stderr should contain
+      """
+      Error: unknown flag: --service-catalog
+      """
+
+  Scenario: User cannot use minishift flag --extra-clusterup-flag directly
+    Given Minishift has state "Does Not Exist"
+     Then executing "minishift start --extra-clusterup-flags" fails
+      And stderr should contain
+      """
+      Error: unknown flag: --extra-clusterup-flags
+      """
+
+  Scenario: User can enable and disable minishift experimental flag --extra-clusterup-flags
+    Given Minishift has state "Does Not Exist"
+     When setting up environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" with value "y" succeeds
+     Then executing "minishift start -h" succeeds
+      And stdout should contain
+      """
+      --extra-clusterup-flags string
+      """
+     When unset environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" succeeds
+     Then executing "minishift start -h" succeeds
+      And stdout should not contain
+      """
+      --extra-clusterup-flags string
+      """
+
+  Scenario: User cannot use minishift experimental flag --extra-clusterup-flag without experimental feature name
+    Given Minishift has state "Does Not Exist"
+     When setting up environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" with value "y" succeeds
+     Then executing "minishift start --extra-clusterup-flags" fails
+      And stderr should contain
+      """
+      Error: flag needs an argument: --extra-clusterup-flags
+      """
+      And unset environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" succeeds
+
+  Scenario: User can start minishift experimental feature
+    Given Minishift has state "Does Not Exist"
+      And image caching is disabled
+      And setting up environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" with value "y" succeeds
+     Then executing "minishift start --extra-clusterup-flags --service-catalog" succeeds
+      And stdout should contain
+      """
+      -- Extra 'oc' cluster up flags (experimental) ... 
+         '--service-catalog'
+      """
+      And unset environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" succeeds
+
+  Scenario: Deleting Minishift
+    Given Minishift has state "Running"
+     When executing "minishift delete --force" succeeds
+     Then Minishift has state "Does Not Exist"
--- a/test/integration/testsuite/testsuite.go
+++ b/test/integration/testsuite/testsuite.go
@@ -109,6 +109,12 @@ func FeatureContext(s *godog.Suite) {
 	s.Step(`^scenario variable "(.*)" should not be empty$`,
 		variableShouldNotBeEmpty)
 
+	// steps for set and unset environment variable
+	s.Step(`^setting up environment variable "(.*)" with value "(.*)" succeeds$`,
+		setEnvironmentVariable)
+	s.Step(`^unset environment variable "(.*)" succeeds$`,
+		unSetEnvironmentVariable)
+
 	// steps for rollout check
 	s.Step(`^services? "([^"]*)" rollout successfully$`,
 		MinishiftInstance.rolloutServicesSuccessfully)
@@ -777,4 +783,12 @@ func directoryShouldntExist(dir string) error {
 	}
 
 	return fmt.Errorf("Directory %s exists", dir)
+}
+
+func setEnvironmentVariable(key string, value string) error {
+	return os.Setenv(key, value)
+}
+
+func unSetEnvironmentVariable(key string) error {
+	return os.Unsetenv(key)
 }


test/integration/testsuite/minishift.go
```diff
@@ -350,3 +350,26 @@ func (m *Minishift) minishiftHomeDirectoryShouldntExist(dir string) error {
 
 	return fmt.Errorf("Directory %s exists", dir)
 }
+
+func (m *Minishift) setEnvironmentVariable(variable string, value string) error {
+	os.Setenv(variable, value)
```
praveenkumar: https://golang.org/pkg/os/#Setenv this return the error if not able to set the variable to env. so you need to handle it properly and below block is not required.

```golang
if os.Getenv(variable) != value {
	return fmt.Errorf("Error in setting environment variable %s=%s in the current shell", variable, value)	
}
```

test/integration/testsuite/minishift.go
```diff
@@ -350,3 +350,26 @@ func (m *Minishift) minishiftHomeDirectoryShouldntExist(dir string) error {
...
+
+func (m *Minishift) getEnvironmentVariable(variable string, value string) error {
+	if os.Getenv(variable) != value {
+		return fmt.Errorf("Environment variable value Expected: %s, Actual: %s in the current shell", value, os.Getenv(variable))
+	}
+	return nil
+}
+
+func (m *Minishift) unSetEnvironmentVariable(variable string) error {
+	os.Unsetenv(variable)
```
praveenkumar: Ditto https://golang.org/pkg/os/#Unsetenv need to handle error and don't need below code block.

test/integration/testsuite/minishift.go
```diff
@@ -350,3 +350,26 @@ func (m *Minishift) minishiftHomeDirectoryShouldntExist(dir string) error {
...
+
+func (m *Minishift) setEnvironmentVariable(variable string, value string) error {
+	os.Setenv(variable, value)
+	if os.Getenv(variable) != value {
+		return fmt.Errorf("Error in setting environment variable %s=%s in the current shell", variable, value)
+	}
+	return nil
+}
+
+func (m *Minishift) getEnvironmentVariable(variable string, value string) error {
```
praveenkumar: This method name is misleading `getEnvironmentVariable` sound like I provides the env string and get the value back but here it's more like comparing a env value with given value. Does `compareEnvironmentVariableValue` sound good?

test/integration/features/experimental-feature.feature
```diff
@@ -0,0 +1,47 @@
+@experimental-feature
+Feature: Experimental Feature
+  As a user I can able to enable, run and disable experimental feature
```
agajdosi: As a user I can enable, run and disable experimental feature.

test/integration/features/experimental-feature.feature
```diff
@@ -0,0 +1,47 @@
+@experimental-feature
+Feature: Experimental Feature
```
agajdosi: This should be in plural or specify that the feature tests service-catalog: `service-catalog.feature`.

test/integration/testsuite/minishift.go
```diff
@@ -350,3 +350,26 @@ func (m *Minishift) minishiftHomeDirectoryShouldntExist(dir string) error {
...
+	if os.Getenv(variable) != value {
```
agajdosi: Here I suggest to first retrieve the value, check for error and then compare:
```
actualValue, err := os.Getenv(variable)
if err != nil {
  return err
}

if actualValue != expectedValue {
  return fmt.Errorf("Retrieved environmental variable does not match expected value. Expected: '%s', Actual: '%s'", expectedValue, actualValue)
}
```

amitkrout: @agajdosi I think this won't work as ```os.Getenv(variable)``` returns string otherwise empty if key is not present. See https://golang.org/pkg/os/#Getenv

test/integration/features/experimental-feature.feature
amitkrout: +1, will fix it

test/integration/features/experimental-feature.feature
LalatenduMohanty: @agajdosi agree `s/Feature/Features/`

test/integration/features/experimental-feature.feature
```diff
@@ -0,0 +1,47 @@
...
+  As a user I can able to enable, run and disable experimental feature
+
+  Scenario: User cannot start minishift experimental feature without setting up the environment variable MINISHIFT_ENABLE_EXPERIMENTAL
```
LalatenduMohanty: Suggestion: `User cannot start minishift experimental feature without enabling experimental features`

LalatenduMohanty:
> What I would suggest here is to split the feature file into two features:

I am fine with a single feature file as we should not keep a feature as experimental for long. Ideally it should graduate to  general availability. Also the future of features like `service-catalog` is not certain. Things might change after v3.10.

test/integration/features/experimental-feature.feature
amitkrout: +1, however ```User cannot start minishift experimental feature without enabling experimental flag``` looks more specific.

amitkrout:
Sorry for the delay
Review comments are addressed. 
I updated the feature file name to experimental-flags.feature as it looks more specific.

amitkrout:
> My comments are in line with @agajdosi's, although I still see we can have a single feature (although prefer not to)... 

IMO two feature file for same test will kill the actual feature test as service catalog is the part of experimental feature. so it is unnecessary to create the file for service catalog separately. For example we can not remove ruby app deployment from basic feature to separate file. 

What i think @agajdosi concern is separate feature file help us to keep unblock if service catalog (Experimental) deployment fails however ```--extra-clusterup-flags``` which is added to minishift code should not break as its added form minishift side.

test/integration/features/experimental-flags.feature
```diff
@@ -0,0 +1,47 @@
+@experimental-flags
+Feature: Experimental Flags
+  Experimental flag --extra-clusterup-flags will be enabled by setting MINISHIFT_ENABLE_EXPERIMENTAL environment variable,
+  this flag will provide access to some of upcoming feature and experiments.
+
+  Scenario: User cannot start minishift experimental feature without enabling experimental flag
+    Given Minishift has state "Does Not Exist"
+     Then executing "minishift start --extra-clusterup-flags --service-catalog" fails
```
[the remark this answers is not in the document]
gbraad: Nobody would have to do this: `minishift start --extra-clusterup-flags --service-catalog`, as it contains two or even three failures.

First, he would have used `--service-catalog` to start this feature, but ALSO uses the `--extra-clusterup-flags`? He would use `--extra-clusterup-flags` without setting the experimental enablement... and maybe even third, the actual value for `--extra-clusterup-flags` is empty... this is NOT prevented, but VERY unlikely to be used.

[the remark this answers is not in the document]
gbraad: too much happens at once... while as I said: https://github.com/minishift/minishift/pull/2360#pullrequestreview-118224395

gbraad:
I understand clearly why @agajdosi wants the seperation. It is a completely different user story. It has nothing to do with the actual results of the commands, but the usage... when and why the user would use them.

amitkrout:
All the review comments are addressed. 
@gbraad please comment if it can be improved further or seems ok

amitkrout:
retest this please

amitkrout:
retest this please

amitkrout:
retest this please

test/integration/features/experimental-flags.feature
```diff
@@ -0,0 +1,62 @@
+@experimental-flags
+Feature: Experimental Flags
+  Experimental flag --extra-clusterup-flags will be enabled by setting MINISHIFT_ENABLE_EXPERIMENTAL environment variable,
+  this flag will provide access to some of upcoming feature and experiments.
+
+  Scenario: User cannot start minishift experimental feature directly
+    Given Minishift has state "Does Not Exist"
+     Then executing "minishift start --service-catalog" fails
```
anjannath: What is this scenario testing.. because you can not run `minishift start --service-catalog`... as right now there is no flag `--service-catalog` for `minishift start` command, experimental or otherwise..

test/integration/features/experimental-flags.feature
```diff
@@ -0,0 +1,62 @@
...
+      And setting up environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" with value "y" succeeds
+     Then executing "minishift start --extra-clusterup-flags --service-catalog" succeeds
+      And stdout should contain
+      """
+      -- Extra 'oc' cluster up flags (experimental) ... 
+         '--service-catalog'
+      """
+      And unset environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" succeeds
+
+  Scenario: Deleting Minishift
```
anjannath: Why do we need this scenario in this feature? or is it there in all the features.. to have a clean env for the next feature??

test/integration/testsuite/testsuite.go
```diff
@@ -778,3 +784,11 @@ func directoryShouldntExist(dir string) error {
 
 	return fmt.Errorf("Directory %s exists", dir)
 }
+
+func setEnvironmentVariable(key string, value string) error {
+	return os.Setenv(key, value)
+}
+
+func unSetEnvironmentVariable(key string) error {
```
anjannath: `os.Unsetenv` returns an error when it fails.. is that handled by the `s.Step`?

test/integration/features/experimental-flags.feature
agajdosi: @anjannath `Scenario: Deleting Minishift` is present in all features. It closes the minimal story of user, who starts Minishift, does something, deletes it in the end.

test/integration/testsuite/testsuite.go
agajdosi: @anjannath Yes, s.Step requires and handles the error, when error is not nil, the step fails. Error is reported, scenario ended as failed and the feature continues with next scenario.

test/integration/features/experimental-flags.feature
amitkrout: This is a essential part of feature file. It completes the feature life cycle like ```start -> certain operation -> delete```. moreover the the delete scenario confirms that what ever the changes done in previous set of scenarios on same feature file should not break clean minishift delete.

> to have a clean env for the next feature??

Yes, it only deletes the vm, however the clean environment for the next feature run confirms by the function runCommandsBeforeFeature (https://github.com/minishift/minishift/blob/master/test/integration/testsuite/prepare.go#L154).

test/integration/testsuite/testsuite.go
amitkrout: hmm...Let me check

test/integration/features/experimental-flags.feature
anjannath: I am not  convinced that this a required check.. as `--service-catalog` is passed as argument to `--extra-clusterup-flags`. and is not meant to be passed to `minishift start` command.

> As a user i am going to start experimental feature without knowing the environment setting. This way we can validate the feature functionality such as in which condition it should fail or pass

Are you concerned that people are going to use the `--service-catalog` flag directly and we should prevent them? Then this is something for the docs to make clear.

test/integration/features/experimental-flags.feature
anjannath: Okay

test/integration/testsuite/testsuite.go
anjannath: @amitkrout It seems `s.Step` already takes care of it as @agajdosi mentioned in the above-1 comment.

amitkrout: I did not know that, @agajdosi thanks for the info

amitkrout: i misread the comment https://github.com/minishift/minishift/pull/2360#discussion_r188190110 . yes s.step handles the error

test/integration/features/experimental-flags.feature
amitkrout: I understand your concern 
> I am not convinced that this a required check.. as --service-catalog is passed as argument to --extra-clusterup-flags. and is not meant to be passed to minishift start command.

This is a check nothing much. If user passes it alone it should fail, right ? 

> Are you concerned that people are going to use the --service-catalog flag directly and we should prevent them? Then this is something for the docs to make clear

This is just a abstract form the doc and putting them to the feature file scenario. Mostly scenario contains actual and negative test cases. Theses negative test case comprises wrong data  and can be anything (not necessarily what is exactly there in the doc).

anjannath: > This is a check nothing much. If user passes it alone it should fail, right ?

Right..

>This is just a abstract form the doc and putting them to the feature file scenario. Mostly scenario contains actual and negative test cases. Theses negative test case comprises wrong data and can be anything (not necessarily what is exactly there in the doc).

I get that you are trying to make sure passing that flag to minishift makes it fail to start, but in that case you should use an experimental flag which can be directly passed to minishift, like the `--no-provision` flag.

The appropriate checks for testing the `--service-catalog` flag is there in `Scenario: User cannot use minishift experimental flag --extra-clusterup-flag without experimental feature name`.

Might be i am missing something, lets see what @minishift/minishift-dev thinks.

amitkrout: I feel you are mixing negative scenario step with actual scenario step. Anyway lets wait feedback from @minishift/minishift-dev

LalatenduMohanty: It is not a critical test, but I am fine with it. It is an usual practice to mix positive and negative tests in testing in general.

test/integration/features/experimental-flags.feature
```diff
@@ -0,0 +1,62 @@
...
+      And stdout should contain
+      """
+      --extra-clusterup-flags string
+      """
+     When unset environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" succeeds
+     Then executing "minishift start -h" succeeds
+      And stdout should not contain
+      """
+      --extra-clusterup-flags string
+      """
```
LalatenduMohanty: 👍

test/integration/testsuite/testsuite.go
```diff
@@ -778,3 +784,11 @@ func directoryShouldntExist(dir string) error {
 
 	return fmt.Errorf("Directory %s exists", dir)
 }
+
+func setEnvironmentVariable(key string, value string) error {
+	return os.Setenv(key, value)
```
praveenkumar: :+1:

test/integration/testsuite/testsuite.go
```diff
@@ -778,3 +784,11 @@ func directoryShouldntExist(dir string) error {
...
+}
+
+func unSetEnvironmentVariable(key string) error {
+	return os.Unsetenv(key)
```
praveenkumar: :+1:

PR #2364, opened 2018-05-07
[WIP] Issue #1796 Moved machine config to machine-state config

### LINKED ISSUE #1796: Fix implementation of InstanceConfig (prevent delete, add InstanceState)
The current implementation of `InstanceConfig` is used to store both configuration (`HostFolders`) and state (`OcPath`, `IsRegistered`, `IsRHELBased`). What happens is that this file gets removed on delete of the instance... to aid in the recreation of the instance using the same config, we keep configuration around.

Several things needs to happen to fix this.
  * The current `minishift.json` config file needs to live in a directory higher up in the tree, to prevent it from being deleted. Needs to be confirmed.
  * a `InstanceState` implementation needs to be added
  * The content needs to be split based on config and state information

praveenkumar:
Fix #1796

### DIFF
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -120,7 +120,7 @@ var RootCmd = &cobra.Command{
 		ensureConfigFileExists(constants.ConfigFile)
 
 		// Create MACHINE_NAME.json
-		instanceConfigPath := filepath.Join(constants.Minipath, "machines", constants.MachineName+".json")
+		instanceConfigPath := filepath.Join(constants.Minipath, "machines", constants.MachineName+"-state.json")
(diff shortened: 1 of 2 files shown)


praveenkumar:
I am still not sure should be also take out `cache-image` stuff to it's own config file which can be live under `~/.minishift/config` or it should be part of instance state? As of now this is still part of `config.json` which we want to to avoid. We have to make sure `config.json` only take care of viper flags settings nothing else.

praveenkumar:
retest this please.

budhrg:
retest this please

gbraad:
https://github.com/minishift/minishift/blob/f974f0c6739eaf68d23e248ca2014cd65314a1fa/pkg/minishift/config/instanceconfig.go#L30-L38

praveenkumar:
Closing this one, might be create a proper PR after bit of discussion.

PR #2372, opened 2018-05-09
Issue #2321 Fix output of minishift config view

### LINKED ISSUE #2321: Fix the output of `$ minishift config view`
### General information

  * Minishift version: 1.16.0
  
The output of `$ minishift config view` is not formatted properly

### Steps to reproduce

  1. $ minishift config set skip-check-clusterup-flags true
  2. $ minishift config view

### Expected

```
$ minishift config view
- iso-url                        : centos
- skip-check-clusterup-flags     : true

```

### Actual

```
$ minishift config view
- iso-url              : centos
- skip-check-clusterup-flags: true

```

rhnvrm:
Fixes #2321

### DIFF
--- a/cmd/minishift/cmd/config/view.go
+++ b/cmd/minishift/cmd/config/view.go
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	DefaultConfigViewFormat = "- {{.ConfigKey | printf \"%-21s\"}}: {{.ConfigValue}}"
+	DefaultConfigViewFormat = "- {{.ConfigKey | printf \"%-35s\"}}: {{.ConfigValue}}"
 )
 
 var configViewFormat string


gbraad:
unit test is failing due to mismatch in expectation. please rewrite the test to accomodate the outcome.
perhaps best not to enforce/strictly check on the format/amount of tabs.

cmd/minishift/cmd/config/view.go
```diff
@@ -19,13 +19,14 @@ package config
 import (
 	"bytes"
 	"fmt"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
```
gbraad: Please, do not move this out of place...
We separate on golang provided/framework imports and 3rd party+our own...

cmd/minishift/cmd/config/view_test.go
```diff
@@ -74,6 +77,10 @@ func TestConfigView(t *testing.T) {
 
 		template := determineTemplate(tt.template)
 		configView(configTest, template, tee.StdoutBuffer)
-		assert.Equal(t, tt.expectedString, tee.StdoutBuffer.String())
+		var b bytes.Buffer
+		writer := bufio.NewWriter(&b)
+		formatAndWriteLines(writer, strings.Split(tt.expectedString, "\n"))
```
gbraad: In this case you would also have to test `formatAndWriteLines` in exclusion of the rest of code...

cmd/minishift/cmd/config/view.go
rhnvrm: Updated

cmd/minishift/cmd/config/view_test.go
rhnvrm: Cool, I'll write a test case for this function

rhnvrm:
@gbraad - I have pushed the changes

anjannath:
@rhnvrm For some reason the output of `minishift config view` is broken...
```
[localhost minishift]$ minishift config view
- http-proxy           		: https
- iso-url              		: file
- skip-check-openshift-version	: true
```
---------------------------------------------------------------------------
```
[localhost minishift]$ cat ~/.minishift/config/config.json 
{
    "http-proxy": "https://localhost:80808",
    "iso-url": "file:///home/anath/Downloads/minishfit-centos7.iso",
    "skip-check-openshift-version": true
}
```

gbraad:
@rhnvrm as @anjannath mentioned, please confirm this works well.
perhaps something in the execution flow is altering this...
if so, the unit-test does not cover (which I feared for this part)

rhnvrm:
@anjannath @gbraad - This was happening because I was using Split. I used SplitN now and added a test case for this as well.

cmd/minishift/cmd/config/view_test.go
```diff
@@ -74,6 +77,53 @@ func TestConfigView(t *testing.T) {
...
+				},
+			},
+			wantWriter: "urlKey\t: 'http://minishift.io/'\n",
+		},
+	}
+	for _, tt := range tests {
+		writer := &bytes.Buffer{}
+		formatAndWriteLines(writer, tt.args.lines)
+		gotWriter := writer.String()
+		assert.Equal(t, tt.wantWriter, gotWriter)
```
gbraad: `s/wantWriter/expected`
`s/gotWriter/actual`

No need to state the implementation type...

cmd/minishift/cmd/config/view_test.go
```diff
@@ -74,6 +77,53 @@ func TestConfigView(t *testing.T) {
...
+		writer := bufio.NewWriter(&b)
+		formatAndWriteLines(writer, strings.Split(tt.expectedString, "\n"))
+		expectedString := b.String()
+		assert.Equal(t, expectedString, tee.StdoutBuffer.String())
+	}
+}
+
+func TestFormatAndWriteLines(t *testing.T) {
+	type args struct {
+		lines []string
```
gbraad: I do not think the naming is very intuitive...
Can this be different?

cmd/minishift/cmd/config/view_test.go
rhnvrm: Have updated the naming

cmd/minishift/cmd/config/view_test.go
rhnvrm: Renamed to alignTabbedLines, does this sound fine?

rhnvrm:
Now the go-dog tests are failing because of tabs, 

https://ci.centos.org/job/minishift-pr/4226/consoleFull

For example: 

https://github.com/minishift/minishift/blob/master/test/integration/features/basic.feature#L52
```
 And stdout of command "minishift config view --format {{.ConfigKey}}:{{.ConfigValue}}" contains "<property>:<value>"
```

How should I proceed with this?

rhnvrm:
@gbraad 
One solution I can think of is to modify:

https://github.com/minishift/minishift/blob/8520eabc36811870aca6292b5c969c107b258229/test/integration/util/checks.go#L27

and strip tab characters from actual

LalatenduMohanty:
@rhnvrm The integration tests are failing because it needs change too. Are you comfortable fixing the integration tests [1] or I can do it for you i.e. once we approve your PR, I will add another commit for the integration test long your commit for the issue fix. Also Please squash the commits as we just need one commit for the fix. 

[1] https://github.com/minishift/minishift/blob/master/test/integration/features/cmd-config.feature#L73-L101

rhnvrm:
@LalatenduMohanty - I am open to contribute for modifying the integration tests. :+1: 

For fixing the integration tests, I am having some doubts on how to approach. Should I modify the feature file or I should modify the function that is called when the integration testing is done (`CompareExpectedWithActualContains), or some other way.

anjannath:
@rhnvrm You don't need to modify the godog tests, Try to debug the issue by running the command as @gbraad mentioned in https://github.com/minishift/minishift/pull/2372#issuecomment-387776089 the tabs won't appear then.. but if it appears then you are probably changing the values which you should not just modify them for the display purpose.. not actually modify them by appending tabs.

LalatenduMohanty:
retest this please

rhnvrm:
@gbraad @LalatenduMohanty @anjannath - Hey, based on your inputs I realized the problem was that I was modifying the output after executing the template and that was causing the issue. So I modified the default template to include tabs and modified the unit test template to be same as the new default template.

gbraad:
Not sure, why does the "default" template need to include tabs?

rhnvrm:
(Copying from the discussion in IRC)
I did that as I couldn't figure out any other way to avoid processing after executing the template.
So I changed `DefaultConfigViewFormat` which is used by `minishift config view` without the format is run

cmd/minishift/cmd/config/view.go
```diff
@@ -29,7 +29,7 @@ import (
 )
 
 const (
-	DefaultConfigViewFormat = "- {{.ConfigKey | printf \"%-21s\"}}: {{.ConfigValue}}"
+	DefaultConfigViewFormat = "- {{.ConfigKey | printf \"%-35s\"}}: {{.ConfigValue}}"
```
gbraad: will this not just postpone the issue until longer strings are used?

LalatenduMohanty: @gbraad thats, right. But I do not see any better alternative. Also it is very hard to support the `--format` with tabwriter. So the question is is it really required to put a lot of effort for this. IMO this simple fix which will be enough for us and there is `--format` anyway if someone wants the format in specific way.

LalatenduMohanty: @gbraad I am going to merge it, If you have other suggestions please open an new issue.

LalatenduMohanty:
Thanks @rhnvrm .

PR #2381, opened 2018-05-14
Issue #1802 Add ability to clear a given image from the cache

praveenkumar:


cmd/minishift/cmd/image/prune.go
```diff
@@ -0,0 +1,116 @@
...
+		logFile := createLogFile()
+		defer logFile.Close()
+		out = logFile
+	} else {
+		out = os.Stdout
+	}
+
+	images := imagesToPrune(api, args)
+
+	handler, err := image.NewOciImageHandler(host.Driver, envMap)
```
LalatenduMohanty: This does not look right to me. We should not need docker daemon to delete a cached docker image. Need to look in to it more.

LalatenduMohanty: @praveenkumar Have you used `NewLocalOnlyOciImageHandler`?

```
// NewLocalOnlyOciImageHandler creates  a new ImageHandler which can only interact with the local cache.
// No connection information to the Docker daemon are provided. Functions interacting with the Docker daemon will fail.
func NewLocalOnlyOciImageHandler() (*OciImageHandler, error) {
	return &OciImageHandler{driver: nil, dockerClientSettings: nil}, nil
}
```

praveenkumar: @LalatenduMohanty Check https://github.com/minishift/minishift/pull/2381#discussion_r187993721

cmd/minishift/cmd/image/prune.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minishift/docker/image"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+)
+
+var (
+	imagePruneCmd = &cobra.Command{
+		Use:   "prune [image ...]",
+		Short: "Prunes the specified container images.",
```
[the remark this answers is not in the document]
LalatenduMohanty: @praveenkumar Yes, thats what I meant.

praveenkumar:
@LalatenduMohanty All the changes are now in place.

praveenkumar:
@lala it does delete check the sha blob of the image before and after you
delete.

In your case blob is so small that you didn't notice using df command.

praveenkumar:
@LalatenduMohanty I does delete the blob also.

```
$ ./minishift start --no-provision
[...]
   Pinging 8.8.8.8 ... OK

$ ./minishift image export alpine
Exporting 'alpine:latest'..... OK

$ ./minishift image list
alpine:latest

$ python -mjson.tool ~/.minishift/cache/images/index.json
{
    "manifests": [
        {
            "annotations": {
                "org.opencontainers.image.ref.name": "alpine:latest"
            },
            "digest": "sha256:41229c767e78913442ba2d57886137898ae5b7b968360d0c05c021d9f48626d0",
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            },
            "size": 348
        }
    ],
    "schemaVersion": 2
}

$ ls -l ~/.minishift/cache/images/blobs/sha256/
total 2028
-rw-r--r--. 1 prkumar prkumar     348 May 16 13:20 41229c767e78913442ba2d57886137898ae5b7b968360d0c05c021d9f48626d0
-rw-r--r--. 1 prkumar prkumar     585 May 16 13:20 b1a7f144ece0194921befe57ab30ed1fd98c5950db7996719429020986092058
-rw-r--r--. 1 prkumar prkumar 2065537 May 16 13:20 ff3a5c916c92643ff77519ffa742d3ec61b7f591b6b7504599d95a4a41134e28
[prkumar@local bin]$ ./minishift image prune alpine
Deleting 'alpine:latest' from the local cache. OK
[prkumar@local bin]$ python -mjson.tool ~/.minishift/cache/images/index.json
{
    "manifests": []
}

$ ls -l ~/.minishift/cache/images/blobs/sha256/
total 2024
-rw-r--r--. 1 prkumar prkumar     585 May 16 13:20 b1a7f144ece0194921befe57ab30ed1fd98c5950db7996719429020986092058
-rw-r--r--. 1 prkumar prkumar 2065537 May 16 13:20 ff3a5c916c92643ff77519ffa742d3ec61b7f591b6b7504599d95a4a41134e28
```

LalatenduMohanty:
@praveenkumar I will do another set of test and let you know.

LalatenduMohanty:
@praveenkumar 

> In your case blob is so small that you didn't notice using df command.

As we discussed the issue seems to be in the dependent layers are handled. Somehow while deleting the image all dependent layers are not getting deleted even if it is the last cached image. 

For a old `minishift` user the old caches of container images can be a lot of disk usage e.g. I was using around 5 GB and there is no way except `minishift delete --clear-cache`  which is a bad user experience. For example if I do not have minishift instance running , I have to start and then delete the minishift to delete all cached images. Also It will delete everything in cache , just not the images directory. 

So I think we should have an option to delete all cached images i.e. `minishift image prune --all` which will `rm -rf .minishft/cache/images`. @minishift/minishift-dev WDTY?

gbraad:
It makes sense to have a `prune -a`, as people can expect from https://docs.docker.com/engine/reference/commandline/system_prune/

anjannath:
Yes an option for removing all the images is a nice thing..

praveenkumar:
Added `all` option, centos CI failure not relevant for this.

cmd/minishift/cmd/image/prune.go
```diff
@@ -0,0 +1,110 @@
...
+	if len(images) == 0 {
+		atexit.ExitWithMessage(0, noImageProvided)
+	}
+
+	return images
+
+}
+
+func init() {
+	imagePruneCmd.Flags().BoolVar(&pruneAll, "all", false, "Deletes all images available in the local image cache.")
```
LalatenduMohanty: We should also add short form `-a` here. We have short forms for --force as -f etc in other commands.

cmd/minishift/cmd/image/prune.go
```diff
@@ -0,0 +1,110 @@
...
+	imageCacheConfig := &image.ImageCacheConfig{
+		HostCacheDir:      state.InstanceDirs.ImageCache,
+		CachedImages:      normalizedImageNames,
+		Out:               out,
+		ImageMissStrategy: image.Pull,
+	}
+
+	_, err = handler.PruneImages(imageCacheConfig)
+	if err != nil {
+		msg := fmt.Sprintf("Container image Prune failed:\n%v", err)
```
LalatenduMohanty: `P` in`Prune` is in capital letter which should be small letter. Also IMO we can modify this as `Container image delete failed`

cmd/minishift/cmd/image/prune_test.go
```diff
@@ -0,0 +1,34 @@
...
+
+func Test_no_images_to_prune(t *testing.T) {
+	tee := cli.CreateTee(t, true)
+	defer cli.TearDown("", tee)
+	expectedOut := noImageProvided
+
+	atexit.RegisterExitHandler(cli.VerifyExitCodeAndMessage(t, tee, 0, expectedOut))
+
+	imagesToPrune(nil, nil)
+}
```
LalatenduMohanty: @praveenkumar What is the test doing? I am not able to see how it is adding value.

praveenkumar: This test atleast test if a user doesn't provide any image it show given message and exit code.

PR #2396, opened 2018-05-22
Issue #1259 minishift update should take version into account

### LINKED ISSUE #1259: minishift update should take version into account
Similar to issue #292. Right now, `minishift update` just checks whether there is a newer version of Minishift. It does not consider the actual version. This means the user might do an update to a backward incompatible version. The update mechanism should check the latest version and in case there is a change in major version provide some additional information. 

We could improve this in many ways. For example we could deploy as part of a major change a README with information what an update would entail. The update command could retrieve this information and display to the user.

budhrg:
Fix #1259


cmd/minishift/cmd/update.go
```diff
@@ -73,30 +75,60 @@ func runUpdate(cmd *cobra.Command, args []string) {
...
+	if versionFlag == "" {
+		updateToLatest(localVersion)
+	} else {
+		version, _ := semver.Make(versionFlag)
+		if err := update.Update(version); err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Update failed: %s", err))
+		}
+		fmt.Printf("\nUpdated successfully to Minishift version %s.\n", version)
+	}
+}
```
praveenkumar: @budhrg this is something not look clean, till now what we were doing is updating directly to latest update version but now we just add a verison parameter so for that our base logic doesn't change. And your `updateToLatest` function change to `updateToVersion` which can be latest.

Something like below.

```golang
	currentVersion, err := update.CurrentVersion()
	if err != nil {
		atexit.ExitWithMessage(1, fmt.Sprintf("Update failed: %s", err))
	}
	versionToUpdate, err := update.LatestVersion()
	if versionFlag != "" {
		versionToUpdate, err = semver.Make(versionFlag)
	}
	updateToVersion(currentVersion, versionToUpdate)
```

cmd/minishift/cmd/update.go
```diff
@@ -68,58 +74,26 @@ func runUpdate(cmd *cobra.Command, args []string) {
...
-	localVersion, err := update.CurrentVersion()
+	currentVersion, err := update.CurrentVersion()
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Update failed: %s", err))
 	}
-
-	latestVersion, err := update.LatestVersion()
-	if err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Update failed: %s", err))
+	versionToUpdate, err := update.LatestVersion()
```
praveenkumar: @budhrg, why we are not handling the err here? I think it should be handled properly might be just after Line#84.

LalatenduMohanty:
@budhrg  I tired the PR and updated it from 0.17.0 to 1.17.0 but it did not warn me about the major version change.
```
$ minishift update
Do you want to update from 0.17.0 to 1.17.0 now? [y/N]: y
Downloading https://github.com/minishift/minishift/releases/download/v1.17.0/minishift-1.17.0-darwin-amd64.tgz
 5.64 MiB / 5.64 MiB [=========================================================================================================================================================================] 100.00% 0s
 65 B / 65 B [=================================================================================================================================================================================] 100.00% 0s

Updated successfully to Minishift version 1.17.0.

Current Installed add-ons are locally present at: /Users/lmohanty/.minishift/addons
The add-ons for 1.17.0 available at: https://github.com/minishift/minishift/tree/v1.17.0/addons

Do you want to update the default add-ons? [y/N]: y
Default add-ons will be updated the next time you run any 'minishift' command.
```

The idea is this fix should inform user about a major version change and what it means. For example hypothetically if in 2.0 we change the `.minishift` layout then updating to 2.0 means the layout needs to be changed. Either we inform the user about it and expect user to take the steps or we automate this after asking user confirmation.

budhrg:
@LalatenduMohanty This is how it is looking now: 
```
$ minishift update -f --update-addons
Downloading https://github.com/minishift/minishift/releases/download/v1.17.0/minishift-1.17.0-linux-amd64.tgz
 5.36 MiB / 5.36 MiB [============================================================================================================================================================] 100.00% 0s
 65 B / 65 B [====================================================================================================================================================================] 100.00% 0s

Updated successfully to Minishift version 1.17.0.
Default add-ons will be updated the next time you run any 'minishift' command.
The latest version is not compatible with the current version. Follow the uninstallation procedure at https://docs.openshift.org/latest/minishift/getting-started/uninstalling.html#uninstall-instructions.
```

PR #2399, opened 2018-05-23
Issue #2290 Add minishift architecture diagram as image

### LINKED ISSUE #2290: Rendering of Minishift architecture image is broken
Minishift architecture image rendering is broken in the published docs [1].

[1] https://docs.openshift.org/latest/minishift/using/basic-usage.html

budhrg:
Fix #2290

### DIFF
--- a/docs/source/using/basic-usage.adoc
+++ b/docs/source/using/basic-usage.adoc
@@ -33,41 +33,7 @@ For more details about using {project} to manage your local OpenShift cluster, s
 [[diagram-minishift-architecture]]
 .: Minishift architecture
 
-[ditaa, "minishift-architecture"]
-----
-
- host machine
- +-------------------------------------------------------------------------------------+
- |                      /-------------> Minishift VM                                   |
- |                      |                                                              |
- |                      |         /--->  o boots from pluggable Live ISO               |
- | $PATH                |         |      o runs on VirtualBox, HyperV, xhyve or KVM    |
- | └── minishift -=-----|         |                                                    |
- |                      |         |   /------------------------------------------\     |
- |                      |         |   |c1AB                                      |     |
- |                      |-=-------|=--|-=->  Docker                              |     |
- |                      :         |   |   /-----------------------------\        |     |
- |                      |         |   |   | c1FF                        |        |     |
- |~/.minishift/         \-=-------|=--|-=-|-=---> OpenShift             |        |     |
- | |                              |   |   |  |  /-------------------\   |        |     |
- | +-- cache                      |   |   |  |  |                   |   |        |     |
- |     |                          |   |   |  |  | master   node     |   |        |     |
- |     +-- iso                    |   |   |  |  | +-----+  +-----+  |   |        |     |
- |     |   |                      |   |   |  |  | | {d} |  | {d} |  |   |        |     |
- |     |   +-- minishift_b2d.iso -/   |   |  |  | |     |  |     |  |   |        |     |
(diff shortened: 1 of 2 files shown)


docs/source/using/basic-usage.adoc
```diff
@@ -33,41 +33,7 @@ For more details about using {project} to manage your local OpenShift cluster, s
...
- |         +-- v3.6.0                 |   |  |  | |     |  |     |  |   |        |     |
- |             |                      |   |  |  | |     |  |     |  |   |        |     |
- |             +-- oc  -=-------------|-=-|-=+  | +-----+  +-----+  |   |        |     |
- |                                    |   |     \-------------------/   |        |     |
- |                                    |   \-----------------------------/        |     |
- |                                    \------------------------------------------/     |
- |                                                                                     |
- +-------+-----------------------------------------------------------------------------+
-
-----
```
praveenkumar: I don't think this is the only thing you need to remove, we also have lots of dependencies which use to compile that stuff to convert it to an image. We need to remove those also.

budhrg: Ack

budhrg: Done @praveenkumar

budhrg:
@praveenkumar , reverted the necessary changes required after looking into the implemented PR https://github.com/minishift/minishift/pull/1590 .

gbraad:
I am not sure reverting is the right apptroach.

You would normally try to prevent adding static resources to the repo, especially when you have something that can aid with generating the output.

Since we remove the source-to-output (compile step), we might STILL want to use the process from the command line? If so, documentation needs to be provided (and perhaps keep the resources needed for this in place).

budhrg:
> Since we remove the source-to-output (compile step), we might STILL want to use the process from the command line? 

Since we are mostly agreeing that DITAA image are facing problem in getting rendered consistently, I am not in the favor of keeping the code which support in rendering.

Not sure what is your intention here.

gbraad:
> Not sure what is your intention here.

In future, you will update the picture by hand?

budhrg:
> In future, you will update the picture by hand?

Probably yes. It seems more maintainable to me than having debugging various versions of gems(packages) why the rendering breaks. 

It would have made more sense if we many DITAA images but we have only one image since its implementation. And we are not sure about this count. :smile:

budhrg:
> In future, you will update the picture by hand?

gbraad:
Even for a single image it would make sense to use a DSL or a managable format, however it might need to move to something else... maybe something like http://asciiflow.com/ ?

budhrg:
> however it might need to move to something else... maybe something like http://asciiflow.com/ ?

Moving to completely new thing is like putting our effort till now into waste. 

> Even for a single image it would make sense to use a DSL or a managable format,

This might be possible so thinking on this line now.

gbraad:
> Moving to completely new thing is like putting our effort till now into waste.

Moving from an ASCII format is not that bad of a deal. I provide an alternative, as it seems the toolchain was the issue.

Anyway, look for something that is generally accepted in tools like Pandoc, or at least has converters.

Or find a way to convert to SVG, as this is at least easy to handle from a sourcecode perspective and allows to diff based on test, and is easy to modify with tools.

LalatenduMohanty:
I agree with @gbraad that even if we add the image we need to keep the source and steps to generate the image in the code itself. 

> Moving to completely new thing is like putting our effort till now into waste.

@budhrg it has been proved that the current way is breaking often, so we should look in to other stuff as well.

Issue #2398, opened 2018-05-23
lower case of proxy env also should be part of the VM


praveenkumar:
As of now if you are working on proxy env then we only set the `HTTP_PROXY` and `HTTPS_PROXY` in the VM environment variable which doesn't work for each command. Interestingly it even not work with `curl` :(, looks like curl only respect the lower case of the variable so it is better to export both kinds of the variable instead depending on single one.

cc @xytrams

LalatenduMohanty:
Resolved via #2402

Issue #2368, opened 2018-05-07
nodejs-ex application deployment fails during integration test


LalatenduMohanty:
nodejs-ex application deployment fails sometimes during PR integration test runs. In https://github.com/minishift/minishift/issues/2357 we had increased the timeout but it is not working 100% of time. 

```
myproject.192.168.42.105.nip.io/
09:19:51 HTTP check (31) has failed. Error: Server returned error on url: http://nodejs-ex-myproject.192.168.42.105.nip.io/
09:19:51     Then with up to "30" retries with wait period of "1000ms" the "body" of HTTP request to "/" of service "nodejs-ex" in namespace "myproject" contains "Welcome to your Node.js application on OpenShift" # testsuite.go:641 -> /testsuite.verifyRequestToServiceWithRetry
09:19:51     Server returned error on url: http://nodejs-ex-myproject.192.168.42.105.nip.io/
```
Detailed log https://ci.centos.org/job/minishift-pr/4198/consoleFull

agajdosi:
Problem could be here: https://github.com/minishift/minishift/blob/0a9a21b86d5240bd76cf5128475239ec79973111/test/integration/features/cmd-openshift.feature#L40. We had problem with OpenShift not being ready after restart, which I fixed with the retry statement, but that repeat the oc command until it succeeds, potentially causing other problems. The step however passes, but it consequences are visible once we check if nodejs-ex works. Going to try it today.

LalatenduMohanty:
@agajdosi May be we can increase the interval of retry and see if it helps. `2 seconds` is very less interval.

agajdosi:
@LalatenduMohanty So I think I have found the root of the issue. And surprisingly it seems to be before the actual restart. I have added checks if `docker-registry` and `router` containers are ready and from the logs (https://ci.centos.org/job/minishift-pr/4334/consoleFull) it seems that `router` container is not ready in the moment when tests restarts the OpenShift which later causes an error on the pod. As result we then can't curl application endpoint successfully. Caused by issue: https://github.com/minishift/minishift/issues/2169.

Will try now if adding sleep time before restart and after it resolves the issue with CI until the original cause #2169 is fixed.

LalatenduMohanty:
Resolved #2407

PR #2402, opened 2018-05-24
Issue #2398 Add lower case proxy variable along with upper case

praveenkumar:


pkg/util/proxy.go
```diff
@@ -97,12 +97,15 @@ func (p *ProxyConfig) ProxyConfig() []string {
...
 		config = append(config, fmt.Sprintf("HTTP_PROXY=%s", p.httpProxy))
+		config = append(config, fmt.Sprintf("http_proxy=%s", p.httpProxy))
 	}
 
 	if p.httpsProxy != "" {
 		config = append(config, fmt.Sprintf("HTTPS_PROXY=%s", p.httpsProxy))
+		config = append(config, fmt.Sprintf("https_proxy=%s", p.httpsProxy))
 	}
 	config = append(config, fmt.Sprintf("NO_PROXY=%s", p.NoProxy()))
+	config = append(config, fmt.Sprintf("no_proxy=%s", p.NoProxy()))
```
LalatenduMohanty: @praveenkumar I was expecting that we would also change `ApplyToEnvironment()` function, but we are not.  We should also fix it, isn't it?

```
func (p *ProxyConfig) ApplyToEnvironment() {
	if !p.IsEnabled() {
		return
	}

	if p.httpProxy != "" {
		os.Setenv("HTTP_PROXY", p.httpProxy)
	}
	if p.httpsProxy != "" {
		os.Setenv("HTTPS_PROXY", p.httpsProxy)
	}
	os.Setenv("NO_PROXY", p.NoProxy())
}
```

LalatenduMohanty:
The CI failure is because of #2368

Issue #2169, opened 2018-03-23
"minishift openshift restart" right after "minishift start" make router and docker registry unusable


agajdosi:
### General information

  * Minishift version: current master
  * OS: Linux - CentOS (on centos-ci)
  * Hypervisor: KVM

### Steps to reproduce

  1. minishift start
  2. minishift openshift restart
  3. eval $(minishift oc-env)
  4. oc get pods -n default

### Expected

Pods docker-registry and router should be running. 

### Actual

On some system right after the minishift start the docker-registry and router pods' status is Running, however they are not READY:

```
NAME                            READY     STATUS    RESTARTS   AGE       IP               NODE
docker-registry-1-4np67         0/1       Running   0          15s       172.17.0.5       localhost
docker-registry-1-deploy        1/1       Running   0          21s       172.17.0.4       localhost
persistent-volume-setup-ndlx2   1/1       Running   0          22s       172.17.0.3       localhost
router-1-c9fmh                  0/1       Running   0          15s       192.168.122.40   localhost
router-1-deploy                 1/1       Running   0          21s       172.17.0.2       localhost
```

When OpenShift container is restarted in this situation the docker-registry and router containers can't handle it and fails. The failure is however quiet and thus will cause problems later. 

### Fix?

Docker-registry and router pods should be in Running state and `READY` after startup. Minishift should check this as part of Minishift start.

Also shouldn't we first prepare the local host node for shutdown, restart and then mark the node ready for deployments? (https://docs.openshift.com/container-platform/3.3/admin_guide/manage_nodes.html#rebooting-nodes)

```
oc adm drain localhost
restart origin container (minishift openshift restart)
oc adm uncordon localhost
wait until registry and router are ready
```

anjannath:
Still to do..

>

agajdosi:
@LalatenduMohanty TBH I do not fully understand the concept behind the fact oc provides STATUS and READY for the pods. But it seems that after M start some pods are running, but are not ready sometimes, which causes troubles.

You can try with:
```
oc get pods -n default --as system:admin
```

Issue #2016, opened 2018-02-12
minishift openshift restart is causing regression for origin version > v3.7.0


LalatenduMohanty:
### General information
`minishift openshift restart` for Origin version > v3.7.0 needs extra time to make the OpenShift instance ready to use.

 For example `minishift openshift registry` or `oc new app` commands are failing after `minishift openshift restart`.

  * Minishift version:  v1.13.0
  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

In order to update default Minishift version to v3.7.1 (in PR #1808) we have do modify the test as below in test/integration/features/cmd-openshift.feature

```diff 
  Scenario: User deploys nodejs example application from OpenShift repository
-      When executing "oc new-app https://github.com/openshift/nodejs-ex -l name=myapp" succeeds
+     Given Minishift has state "Running"
+      When executing "oc new-app https://github.com/openshift/nodejs-ex -l name=myapp" retrying 10 times with wait period of 2 seconds
       Then stdout should contain
```
### Steps to reproduce

```
$ minishift openshift restart;minishift openshift registry
Restarting OpenShift
No information found for 'service/docker-registry'

$ minishift openshift restart;sleep 2; minishift openshift registry
Restarting OpenShift
172.30.1.1:5000
```

budhrg:
Similar experience here [minishift openshift restart can break router and Docker registry pods](https://github.com/minishift/minishift/issues/2169).

praveenkumar:
@LalatenduMohanty @agajdosi doesn't this now fixed?

praveenkumar:
@LalatenduMohanty @agajdosi any update on this?

agajdosi:
@praveenkumar From logs on passing PRs it looks fine, no retries reported:
- https://ci.centos.org/job/minishift-pr/4677/consoleFull
- https://ci.centos.org/job/minishift-pr/4674/consoleFull
- https://ci.centos.org/job/minishift-pr/4669/consoleFull
- https://ci.centos.org/job/minishift-pr/4651/consoleFull
- https://ci.centos.org/job/minishift-pr/4648/consoleFull

Can create a PR to remove it, or we can wait for oc v3.10.x becoming default and then see how it goes, wdyt?

LalatenduMohanty:
@agajdosi I think lets keep ot as it is as we are moving towards OpenShift 4.0

PR #2407, opened 2018-05-25
Issue #2368 Nodejs-ex app deployment fails during integration test

agajdosi:


agajdosi:
retest this please

agajdosi:
PR improves debug capabilities of tests: 
- executingRetryingTimesWithWaitPeriodOfSeconds() now reports each failure to the console, thus now we have info whether the step passed on 1st or 15th try,
- verifyHTTPResponse now reports error explicitly.

Also added 2 scenarios to the cmd-openshift.feature to check for status of router and docker-registry pods. This is to make tests to fail on the cause (pods not ready causes restart to break them completely - #2016, #2169) not on consequence (http check of ruby-ex application).

To discuss: 30 seconds sleep before and after restart seems to fix the issue for the tests. Is it ok to keep it until the issues are fixed, or do we want to stay with failing tests and try to fix the issues asap? WDYT?

test/integration/features/cmd-openshift.feature
```diff
@@ -25,6 +25,12 @@ cluster in VM provided by Minishift.
        And stdout should contain "kubernetes"
        And stdout should contain "router"
 
+  Scenario: Pods docker-registry and router are ready before OpenShift restart
+     Given user waits "30" seconds
```
LalatenduMohanty: @agajdosi Why do we need to wait here?

agajdosi: @LalatenduMohanty This is because nodes need some time to get not only running, but be also ready. The second sleep is because right after restart, oc sometimes gives TLS error (the after restart sleep could be probably shorter), but to be sure 30 seconds works best.

test/integration/features/cmd-openshift.feature
```diff
@@ -35,6 +41,12 @@ cluster in VM provided by Minishift.
       Then stdout should contain "OpenShift restarted successfully"
        And stdout of command "minishift ssh -- "docker inspect --format={{.State.FinishedAt}} origin"" is not equal to "0001-01-01T00:00:00Z"
 
+  Scenario: Pods docker-registry and router are ready before OpenShift restart
```
anjannath: s/before/after

agajdosi: thanks, going to update

agajdosi:
@anjannath Yes, the restart should check if all pods are ready, then restart. Otherwise inform user that she can use --force flag to hard restart, which could bring issues. For the start I think we should check that all is running and until then do not say "Minishift is running (but hey, some key components of OpenShift are still not ready)".

Maybe and here I am speculating a little bit we could even implement something like this:
```
[agajdosi@dhcp-24-156 ~]$ minishift status
Minishift:  Running
Profile:    minishift
OpenShift:  Running (openshift v3.9.14)
  - docker-registry-1-4hn2h:     ready, running
  - router-1-84pr2:              ready, running
  - webconsole-fc49b9494-ns2f7:  ready, running
DiskUsage:  9% of 19G
```

anjannath:
>  the restart should check if all pods are ready, then restart. Otherwise inform user that she can use --force flag to hard restart, which could bring issues

for `minishift openshift restart` it restarts the origin container only AFAIR, @praveenkumar am i right?. but it is strange that it needs to be in ready state for restart also, it'll be restarted anyway...

-----------------------------------------

> For the start I think we should check that all is running and until then do not say "Minishift is running (but hey, some key components of OpenShift are still not ready)".

or we can just block till all the containers are ready and only after they are ready, print _Minishift is running.._

test/integration/features/cmd-openshift.feature
anjannath: if #2169 is fixed then this can be removed i guess

Issue #2413, opened 2018-05-29
Fix `minishift oc-env --no-proxy` behavior on Windows


svenstaro:
### General information

  * Minishift version: v1.17.0+c741cbdd
  * OS: Windows
  * Hypervisor: VirtualBox

### Steps to reproduce

1.

    echo $NO_PROXY:
    127.0.0.1,localhost,.company.tld

2.

    echo $no_proxy
    <empty output>

3.

    eval $(minishift oc-env --no-proxy)

### Expected

    echo $NO_PROXY
    127.0.0.1,localhost,.company.tld,192.168.99.101
    echo $no_proxy
    <empty output>

If `NO_PROXY` is already set while `no_proxy` is NOT set then I'd expect `NO_PROXY` to be extended and `no_proxy` shall remain unset. This makes sense since if the user is already using `NO_PROXY` but not `no_proxy`, there's probably a good reason for that.

### Actual

    echo $NO_PROXY
    127.0.0.1,localhost,.company.tld
    echo $no_proxy
    127.0.0.1,localhost,.company.tld,192.168.99.101

svenstaro:
Turns out that this is an issue only on Windows due to its case-insensitivity in regards to env vars.

LalatenduMohanty:
Resolved via #2419

PR #2419, opened 2018-05-29
Issue #2413 Always emit uppercased NO_PROXY on Windows

svenstaro:
I had a fairly involved test with a collegue and this seems to fix the problem well while also being not too bad of an OS-specific hack.

### DIFF
--- a/cmd/minishift/cmd/oc_env.go
+++ b/cmd/minishift/cmd/oc_env.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"text/template"
 
 	"github.com/docker/machine/libmachine"
@@ -115,6 +116,10 @@ var ocEnvCmd = &cobra.Command{
 
 func init() {
 	RootCmd.AddCommand(ocEnvCmd)
-	ocEnvCmd.Flags().BoolVar(&noProxy, "no-proxy", false, "Add the virtual machine IP to the NO_PROXY environment variable.")
+	if runtime.GOOS == "windows" {
+		ocEnvCmd.Flags().BoolVar(&noProxy, "no-proxy", false, "Add the virtual machine IP to the NO_PROXY environment variable.")
+	} else {
+		ocEnvCmd.Flags().BoolVar(&noProxy, "no-proxy", false, "Add the virtual machine IP to the no_proxy/NO_PROXY environment variable.")
+	}
 	ocEnvCmd.Flags().StringVar(&forceShell, "shell", "", "Force setting the environment for a specified shell: [fish, cmd, powershell, tcsh, bash, zsh]. Default is auto-detect.")
 }
--- a/pkg/util/shell/shell.go
+++ b/pkg/util/shell/shell.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"github.com/docker/machine/libmachine/shell"
 	"os"
+	"runtime"
 	"strings"
 )
 
@@ -59,10 +60,22 @@ func FindNoProxyFromEnv() (string, string) {
 	noProxyVar := "no_proxy"
 	noProxyValue := os.Getenv(noProxyVar)
 
-	// otherwise default to allcaps HTTP_PROXY
+	// otherwise default to allcaps NO_PROXY
 	if noProxyValue == "" {
 		noProxyVar = "NO_PROXY"
 		noProxyValue = os.Getenv("NO_PROXY")
+	}
+
+	// On Windows, we've got to cheat a little bit because there is no way to
+	// get a specificially lowercase or uppercase env var. Furthermore, if
+	// both no_proxy and NO_PROXY are set, os.Getenv() will always get
+	// the value for NO_PROXY because it has higher sorting precedence.
+	// Effectively, that means that an uppercase env var on Windows always has
+	// precedence. In addition, it is common practice to write env vars in
+	// Windows in uppercase. All in all, this makes it reasonable to always
+	// uppercase the NO_PROXY var for Windows.
+	if runtime.GOOS == "windows" {
+		noProxyVar = strings.ToUpper(noProxyVar)
 	}
 	return noProxyVar, noProxyValue
 }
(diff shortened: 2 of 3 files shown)


LalatenduMohanty:
ok to test

pkg/util/shell/shell.go
```diff
@@ -59,11 +60,23 @@ func FindNoProxyFromEnv() (string, string) {
...
+	// get a specificially lowercase or uppercase env var. Furthermore, if
+	// both no_proxy and NO_PROXY are set, os.Getenv() will always get
+	// the value for NO_PROXY because it has higher sorting precedence.
+	// Effectively, that means that an uppercase env var on Windows always has
+	// precedence. In addition, it is common practice to write env vars in
+	// Windows in uppercase. All in all, this makes it reasonable to always
+	// uppercase the NO_PROXY var for Windows.
+	if runtime.GOOS == "windows" {
+		noProxyVar = strings.ToUpper(noProxyVar)
+	}
```
praveenkumar: @svenstaro If we are doing this that means we are always going to set `NO_PROXY` env for windows even the user have `no_proxy` env set and that is not get updated. I don't follow why a user wants to have a different value for this env (upper/lower case) when it does the same thing?

svenstaro: This problem is super tricky but I believe this is the best solution. First of all, in Windows itself, it is imossible to set `no_proxy=test1` and `NO_PROXY=test2` globally as the latest entry will always overwrite the former one. That means setting `no_proxy` first and `NO_PROXY` secondly will have you end up with `NO_PROXY` set only.

Bearing that in mind, it IS possible to set `no_proxy` and `NO_PROXY` at the same time using a Windows bash (like mingw's bash). If you set `no_proxy=test1` and `NO_PROXY=test2` on this Bash and then in go do `os.Getenv("no_proxy")` or `os.Getenv("NO_PROXY")` it will in both cases fetch `test2` because the upper case `NO_PROXY` takes precende. go's `os.Getenv()` can not explicitly cased entries correctly because Windows just doesn't work that way.

What this means in turn is that the check a few lines up that checks for whether `os.Getenv("no_proxy")` returns something is ALWAYS true on Windows, even if only `NO_PROXY` is set.

Basically, this will result in this bug:

1. Set NO_PROXY globally using Windows env var tool. Most devs will do it this way natively on Windows, I reckon.
2. Do `oc-env --no-proxy`. This current code will check whether you have set a `no_proxy` variable which will be true even though you actually set `NO_PROXY` because again, `os.Getenv()` can't detect the case.
3. `oc-env --no-proxy` then emits a `no_proxy` varbiable because it detected earlier that this is what you use.
4. This `no_proxy` variable does NOT take effective precedence over the actuall `NO_PROXY` variable that is set in Windows because of the way that mingw and Windows deal with env vars (they are sorted by ASCII value apparently and uppercase wins out) and thus setting eval'ing `oc-env --no-proxy` always gets you a `no_proxy` value which is always overridden by a `NO_PROXY` that is already set.
5. Since `NO_PROXY` which takes precedence is not modified, there is effectively no no-proxy stuff getting done for you by `oc-env --no-proxy`.

I believe what I outlined above leads to my solution to being the only practical one.

praveenkumar: @svenstaro Thanks for details explanation, I understood the code change now and it's windows OS specific. @gbraad Can you please have a look also?

LalatenduMohanty:
retest this please

pkg/util/shell/shell.go
gbraad: This means the uppercase behaviour should be enforced for ALL enviroinment variables on Windows.

LalatenduMohanty: I was testing on Windows and here is the output I got. I was under impresssion that I will see `$Env:NO_PROXY = "192.168.1.5,172.19.89.117"` . What am I missing?
```
PS > $env:NO_PROXY = "192.168.1.5"
PS > $env:no_proxy = "192.168.1.66"
PS > .\minishift.exe oc-env --no-proxy
$Env:PATH = "C:\Users\lmohanty\.minishift\cache\oc\v3.9.0\windows;$Env:PATH"
$Env:NO_PROXY = "192.168.1.66,172.19.89.117"
# & minishift oc-env --no-proxy | Invoke-Expression
```

cmd/minishift/cmd/oc_env.go
```diff
@@ -115,6 +115,6 @@ var ocEnvCmd = &cobra.Command{
 
 func init() {
 	RootCmd.AddCommand(ocEnvCmd)
-	ocEnvCmd.Flags().BoolVar(&noProxy, "no-proxy", false, "Add the virtual machine IP to the NO_PROXY environment variable.")
+	ocEnvCmd.Flags().BoolVar(&noProxy, "no-proxy", false, "Add the virtual machine IP to the no_proxy/NO_PROXY environment variable.")
```
LalatenduMohanty: @svenstaro @gbraad IMO we should put `if runtime.GOOS == "windows" {` here and change it to `Add the virtual machine IP to the NO_PROXY environment variable.` (i.e. only NO_PROXY) and the else can have `Add the virtual machine IP to the no_proxy environment variable.`

pkg/util/shell/shell.go
svenstaro: @gbraad Perhaps, yes, but I consider it out of scope for this issue because it's a very particular issue with a piece of code that is only executed in the `--no-proxy` code path.

@LalatenduMohanty I'm pretty sure that is due to Windows-native tools not adding to the env if the variable is named the same, same as I described above. Can you do this for me (I don't do PowerShell so you gotta bear with my pseudocode):

    Set NO_PROXY=test1
    Set no_proxy=test2
    Set NO_PROXY=test3
    Print no_proxy
    Print NO_PROXY

What are those values respectively?

LalatenduMohanty: My mistake,  the output is correct as windows will only remember NO_PROXY= <value> on powershell.

cmd/minishift/cmd/oc_env.go
svenstaro: Sure, that would make it more correct on Windows.

svenstaro: Same for docker-env I reckon.

svenstaro:
@LalatenduMohanty Windows-specific help is now also in.

LalatenduMohanty:
@svenstaro Thanks for the PR.

Issue #2206, opened 2018-03-30
xpaas addon should use apply and not create


jorgemoralespou:
In openshift, expected/recommended way to create resources in the cluster is via apply and not create. Apply will create or update the resource, while create will fail if the resource already exists.

https://github.com/minishift/minishift/blob/master/addons/xpaas/xpaas.addon#L8 and following lines

anjannath:
Will apply create objects which are not previously defined... like create makes an object ?

jorgemoralespou:
Yes, as I say in the description of the issue.

LalatenduMohanty:
Resolved via #2431

PR #2431, opened 2018-06-01
Issue #2206 use oc apply instead create for file resource

praveenkumar:


LalatenduMohanty:
Thanks @praveenkumar .

PR #2437, opened 2018-06-05
Issue #2370 allow non-ASCII for PowerShell, deprecates 3rd party library

### LINKED ISSUE #2370: Deprecate use of gorillalabs/go-powershell
When writing the pre-flight checks, a simple powershell library was needed to get this working quickly. But when times moves on we might have to reconsider using this. I have filed several issues for the gorillalabs/go-powershell library, but so far no response from the maintainers:
  * https://github.com/gorillalabs/go-powershell/issues/3
  * https://github.com/gorillalabs/go-powershell/issues/10

Also, when I added this library, it had to pull in juju/errors which is not ideal. Someone filed a PR to modify this to use a simple error wrap:
  * https://github.com/gorillalabs/go-powershell/pull/7

and also this PR has been unanswered for almost a month.

In light of this, it might be wise to move to using the docker/machine implementation. This would have to be extracted and publicly exposed. This would be an easy task from our end, as powershell has been abstracted anyway behind a

gbraad:
Fixes #2370

Tested with:

  * static ip
  * assign virtual switch (with modified `docker/machine` lib)

Untested:

  * get network info from host
  * ...

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -325,15 +325,6 @@
   version = "v1.6.1"
 
 [[projects]]
-  name = "github.com/gorillalabs/go-powershell"
-  packages = [
-    ".",
-    "backend",
-    "utils"
-  ]
-  revision = "3bc7a60b1df80a5766820c4a53d59668f4553f75"
-
-[[projects]]
   branch = "master"
   name = "github.com/hashicorp/hcl"
   packages = [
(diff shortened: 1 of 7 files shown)


gbraad:
This PR drops use of gorillalabs/go-powershell due to https://github.com/gorillalabs/go-powershell/issues/10

gbraad:
will do... issue with docker/machine will be opened.

gbraad:
Renamed

Issue #2425, opened 2018-05-31
Minishift - VirtualBox  - Runtime Error


chema-mengibar:
### General information

  * Minishift version: 1.18.0
  * OS:  Windows 8.0
  * Hypervisor:  VirtualBox

### Steps to reproduce

  1. Hyper-V Disabled
  2. minishift config set vm-driver virtualbox

### Actual

https://gist.github.com/chema-mengibar/078a1a6d5bde39fa706f0816bb47781e

praveenkumar:
This error is because of a nil pointer de-reference in the preflight check seems a bug.

```
-- Checking if requested hypervisor 'hyperv' is supported on this platform ... panic: runtime error: invalid memory addr
ess or nil pointer dereference
```

@chema-mengibar meanwhile can you  try following so you will not be blocked.

```
$ minishift config set skip-check-hyperv-driver true
$ minishift start
```

anjannath:
the nill pointer dereference is happening because of the..
```go
defer posh.Close()
```
For some reason the `powershell.New()` is maybe NOT creating new instance so there is nothing to close.

@chema-mengibar Can you please provide the output of this `(Get-WmiObject -Class Win32_OperatingSystem).Caption`.

Also from the logs in the gist, it looks like you are re-starting an old Hyper-V based vm. The config setting for `vm-driver` needs fresh start to take effect. Delete the existing vm do `minishift delete -f` and then start again with the virtualbox driver.

```
$ minishift start --vm-driver virtualbox
or
$ minishift config set vm-driver virtualbox
$ minishift start
```

praveenkumar:
@chema-mengibar Does this issue resolved for you or can you provide the required details?

gbraad:
Windows 8.0 does not come with PowerShell preinstalled

gbraad:
I suggest you to use something like `ver.exe` or `wmic os get Caption /value` instead, as these commands work without the use of PowerShell. Only since Windows 8.1 PowerShell came pre-installed.

anjannath:
@gbraad I was unaware of it, Thanks, will change it.

PR #2444, opened 2018-06-08
Issue #2425 use wmic command instead of posh command to check windows version

anjannath:
Fixes: #2425


pkg/minishift/config/validations.go
```diff
@@ -31,18 +32,19 @@ import (
...
-		checkWindowsName := `(Get-WmiObject -Class Win32_OperatingSystem).Caption`
-		stdOut, _ := posh.Execute(checkWindowsName)
-		if driver == "hyperv" && strings.Contains(stdOut, "Windows 7") {
+		cmd := exec.Command("wmic", "os", "get", "Caption", "/value")
+		stdOut, err := cmd.Output()
+		if err != nil {
+			return err
+		}
+		out := fmt.Sprintf("%s", stdOut)
+		if driver == "hyperv" && strings.Contains(out, "Windows 7") {
```
gbraad: Technically I could start Minishift on Windows XP, right?

pkg/minishift/config/validations.go
```diff
@@ -31,18 +32,19 @@ import (
...
-		defer posh.Close()
-		checkWindowsName := `(Get-WmiObject -Class Win32_OperatingSystem).Caption`
-		stdOut, _ := posh.Execute(checkWindowsName)
-		if driver == "hyperv" && strings.Contains(stdOut, "Windows 7") {
+		cmd := exec.Command("wmic", "os", "get", "Caption", "/value")
+		stdOut, err := cmd.Output()
+		if err != nil {
+			return err
+		}
+		out := fmt.Sprintf("%s", stdOut)
```
gbraad: why not use `stdOut` in the `string.Contains()` ?

anjannath: `stdOut` is of type `[]byte` so need to convert it to be able to use in `strings.Contains()`
```
cannot use stdOut (type []byte) as type string in argument to "strings".Contains
```

pkg/minishift/config/validations.go
anjannath: Yes, will also check if it is windows XP ;-)

pkg/minishift/config/validations.go
gbraad: no `.String()` ?

anjannath: no `.String()` however i haven't tried using `string()`.

anjannath: Weird... if i modify the code as following:
```go
if driver == "hyperv" && strings.Contains(string(stdOut), "Windows 7") || strings.Contains(string(stdOut), "Windows XP") {
			return fmt.Errorf("Driver '%s' is not supported", driver)
		}
```
It throws an error: 
```
pkg/minishift/config/validations.go:46:51: cannot call non-function string (type string)
make: *** [Makefile:86: /home/xxx/work/bin/minishift] Error 2

```

But the same code runs and builds without error if its done like this:
```go
package main

import "fmt"
import "os/exec"
import "strings"


func main() {
	c := exec.Command("echo", "hello", "world")
	out,_ := c.Output()
	fmt.Println(strings.Contains(string(out), "hello"))
}
```
@gbraad What am i doing wrong?

anjannath:
@gbraad We have had to do this earlier in preflight checks..
https://github.com/minishift/minishift/blob/1426bddcbf7aae0b8fab29dbe28b667a9ae218cb/cmd/minishift/cmd/start_preflight.go#L361-L367

anjannath:
@gbraad Will look at this again and make the changes asked for, will also change in the other place, this is to basically avoid using the extra variable...  right?

gbraad:
@anjannath if possible, else it is OK to accept now

Issue #2476, opened 2018-06-20
Add the host interface's IP to the noProxy list


svenstaro:
[Currently, minishift only adds the VM's IP to the noProxy list](https://github.com/minishift/minishift/blob/a039d37ef5e84364a976f3b4b7a07185d202a933/cmd/minishift/cmd/start.go#L205). However, it would also be useful if the machine's host interface IP would be added. We already talked about this on IRC and the minishift team seemed to agree.

gbraad:
You might need:

https://github.com/minishift/minishift/blob/17507025610e89fce899a3fe49d3df5aecbd4d37/pkg/minishift/network/settings.go#L200-L208

gbraad:
Right, there was a single-line implementation for this: https://github.com/minishift/minishift/blob/17507025610e89fce899a3fe49d3df5aecbd4d37/pkg/minishift/hostfolder/cifs_hostfolder.go#L46

gbraad:
OK... been a while since I wrote it.

PR #2477, opened 2018-06-20
Issue #2476 Add the host interface's IP to the noProxy list

svenstaro:


svenstaro:
Please don't merge this right away. I would like to verify with a colleague beforehand.

svenstaro:
@gbraad You're right. Added an explanatory comment.

gbraad:
Added WIP state to prevent merging. Please verify the artifact after build passed ...

gbraad:
add to whitelist

svenstaro:
retest this please

LalatenduMohanty:
@svenstaro PR does not work for me. I can only see the VM's IP address

```
$ minishift docker-env --no-proxy
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.64.3:2376"
export DOCKER_CERT_PATH="/Users/lmohanty/.minishift/certs"
export NO_PROXY="192.168.64.3"
# eval $(minishift docker-env --no-proxy)
```

praveenkumar:
@svenstaro you want to add host interface IP to the VM shell variable instead of host I think, right now what you are doing is adding it to host itself.

gbraad:
@praveenkumar can you make the comment in code and provide the alternative/correct approach?

gbraad:
@svenstaro @praveenkumar It depends on what he needs to achieve.

Either the host or the instance would need this no_proxy?

svenstaro:
I want to make sure that even when a proxy is in use that services on the host can be reached directly by containers run inside minishift without trying to go through the proxy.

For instance, if a database ran on the host, I want to be able to reach that directly from inside a container.

pkg/minikube/cluster/cluster.go
```diff
@@ -354,7 +355,13 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
 			fmt.Println("FAIL")
 			return nil, fmt.Errorf("Error getting VM IP: %s", err)
 		}
+		hostIP, err := minishiftNetwork.DetermineHostIP(h.Driver)
```
praveenkumar: :+1:

pkg/minikube/cluster/cluster.go
```diff
@@ -354,7 +355,13 @@ func createHost(api libmachine.API, config MachineConfig) (*host.Host, error) {
...
+		if err != nil {
+			fmt.Println("FAIL")
+			return nil, fmt.Errorf("Error getting host IP: %s", err)
+		}
 		config.ShellProxyEnv.AddNoProxy(vmIP)
+		config.ShellProxyEnv.AddNoProxy(hostIP)
```
praveenkumar: :+1:

svenstaro:
This is also verified working on Windows.

Issue #2458, opened 2018-06-14
Stopping minishift instance on remote host throws "Error stopping cluster: generic driver does not support stop"


amitkrout:
### General information

  * Minishift version: master
  * OS: macOS
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce

  1. minishift profile set foo
  2. export MINISHIFT_ENABLE_EXPERIMENTAL=y
  3. minishift start --no-provision --vm-driver virtualbox
  4. minishift ip
```
192.168.99.101
```
  5. minishift profile set minishift
  6. minishift start --vm-driver generic --remote-ipaddress 192.168.99.101 --remote-ssh-user docker --remote-ssh-key ~/.minishift/profiles/foo/machines/foo/id_rsa
  7. minishift stop
```
$ ./minishift stop
Stopping local OpenShift cluster...
Error stopping cluster: generic driver does not support stop
```

### Expected
stop should successful

### Actual
Reports error ```Error stopping cluster: generic driver does not support stop```

### Logs

You can start Minishift with `minishift start --show-libmachine-logs -v5` to collect logs.
Please consider posting this on http://gist.github.com/ and post the link in the issue.

gbraad:
I think this is as expected...

praveenkumar:
@amitkrout this is expected because we don't want to a remote machine to shutdown and `oc cluster` doesn't have any option for stop.

`minishift delete` on the other hand will do `oc cluster down` in case of the generic driver.

amitkrout:
> host to shutdown

I am testing the feature on my local setup through multiple profiles so could not realize the cost of shutting down the host. 

> minishift delete on the other hand will do oc cluster down in case of the generic driver. Note: not as reported, but what about minishift delete ?

Working as expected
```
$ ./minishift status
Minishift:  Running
Profile:    foo
OpenShift:  Stopped
DiskUsage:  8% of 19G
CacheUsage: 884.3 MB (used by oc binary, ISO or cached images)
```

Thanks for the info. Will try some more scenarios on remote setup. 

Closing the issue.

gbraad:
@praveenkumar 

well... wouldn't `m stop` have to do `oc cluster down`... while `m delete` should really remove the `/var/lib/minishift` folder on the host ?

amitkrout:
@praveenkumar  I am not sure on how ```m delete``` works on the remote host, however, @gbraad  question looks valid to me

[...]

PR #2479, opened 2018-06-20
Issue #2458 Add stop functionality for generic driver

praveenkumar:
@gbraad @amitkrout @agajdosi  following is what we can have as part of stop when using generic driver. In case of other drivers we halt the VM as part of stop but that something we shouldn't do for `generic` but stopping the local OpenShift cluster make sense.

```
$ ./minishift status
Minishift:  Running
Profile:    minishift
OpenShift:  Running (openshift v3.9.0+b8265d6-31)
DiskUsage:  ‘/mnt/?da1’: of df:
CacheUsage: 1.481 GB (used by oc binary, ISO or cached images)

$ ./minishift stop

$ ./minishift status
Minishift:  Running
Profile:    minishift
OpenShift:  Stopped
DiskUsage:  ‘/mnt/?da1’: of df:
CacheUsage: 1.481 GB (used by oc binary, ISO or cached images)
```

### DIFF
--- a/cmd/minishift/cmd/stop.go
+++ b/cmd/minishift/cmd/stop.go
@@ -55,13 +55,19 @@ func runStop(cmd *cobra.Command, args []string) {
 		atexit.ExitWithMessage(0, fmt.Sprintf("The '%s' VM is already stopped.", constants.MachineName))
 	}
 
-	fmt.Println("Stopping local OpenShift cluster...")
+	fmt.Println("Stopping the OpenShift cluster...")
 
-	// Unregister, allow to be skipped and force deletion is ignored
-	registrationUtil.UnregisterHost(api, true, false)
+	if hostVm.Driver.DriverName() == "generic" {
+		if err := util.OcClusterDown(hostVm); err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		}
+	} else {
+		// Unregister, allow to be skipped and force deletion is ignored
+		registrationUtil.UnregisterHost(api, true, false)
 
-	if err := cluster.StopHost(api); err != nil {
-		atexit.ExitWithMessage(1, fmt.Sprintf("Error stopping cluster: %s", err.Error()))
+		if err := cluster.StopHost(api); err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error stopping cluster: %s", err.Error()))
+		}
 	}
 	fmt.Println("Cluster stopped.")
 }
--- a/cmd/minishift/cmd/util/util.go
+++ b/cmd/minishift/cmd/util/util.go
@@ -23,6 +23,8 @@ import (
 
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/drivers"
+	"github.com/docker/machine/libmachine/host"
+	"github.com/docker/machine/libmachine/provision"
 	"github.com/docker/machine/libmachine/state"
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
@@ -32,6 +34,7 @@ import (
 	cmdState "github.com/minishift/minishift/cmd/minishift/state"
 	"github.com/minishift/minishift/out/bindata"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
 	"github.com/minishift/minishift/pkg/util/shell"
 )
 
@@ -148,4 +151,12 @@ func ValidateGenericDriverFlags(remoteIPAddress, remoteSSHUser, sshKeyToConnectR
 			"--%s string\n--%s string\n--%s string\n", configCmd.RemoteIPAddress.Name, configCmd.RemoteSSHUser.Name, configCmd.SSHKeyToConnectRemote.Name)
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error: %s", msg))
 	}
+}
+
+// OcClusterDown stop Openshift cluster using oc binary inside the remote machine
+func OcClusterDown(hostVm *host.Host) error {
+	sshCommander := provision.GenericSSHCommander{Driver: hostVm.Driver}
+	cmd := fmt.Sprintf("%s/oc cluster down", minishiftConstants.OcPathInsideVM)
+	_, err := sshCommander.SSHCommand(cmd)
+	return err
 }
--- a/cmd/minishift/cmd/delete.go
+++ b/cmd/minishift/cmd/delete.go
@@ -29,7 +29,6 @@ import (
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
-	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
 	pkgUtil "github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/filehelper"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
@@ -76,7 +75,7 @@ func runDelete(cmd *cobra.Command, args []string) {
 	}
 
 	if host.Driver.DriverName() == "generic" {
-		if err := ocClusterDown(host); err != nil {
+		if err := util.OcClusterDown(host); err != nil {
 			atexit.ExitWithMessage(1, err.Error())
 		}
 		if err := deleteExistingDirectory(host); err != nil {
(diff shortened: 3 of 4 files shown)


praveenkumar:
@agajdosi yes that's right because generic driver doen't have a way to stop so this is kind of hack and doing the start it will always check the current status of the minishift. I am not sure if we can get rid of that.

anjannath:
can we not modify the check during start based on the driver used?

praveenkumar:
@agajdosi updated the PR, hope that looks what you expected?

```
$ ./minishift start --vm-driver generic --remote-ipaddress 10.8.x.x --remote-ssh-user root --remote-ssh-key ~/.ssh/id_rsa  --skip-registration 
[...]
    https://10.8.252.96:8443

You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin


$ ./minishift stop

$ ./minishift start --vm-driver generic --remote-ipaddress 10.8.x.x --remote-ssh-user root --remote-ssh-key ~/.ssh/id_rsa  --skip-registration 
[...]
OpenShift server started.

    https://10.8.252.96:8443
```

agajdosi:
@praveenkumar On start from Stopped to Running it ends with strange error-ish message:

```
Could not set oc CLI context for: 'minishift'
```

However console is accessible. Only issue I face is that oc-env gives error that it requires running VM. Status is running M and running OCP.

Interesting part: the start works also without any flag:

```
[agajdosi@dhcp-24-156 ~]$ minishift status
Minishift:  Running
Profile:    minishift
OpenShift:  Stopped
DiskUsage:  10% of 19G
CacheUsage: 578.4 MB (used by oc binary, ISO or cached images)
[agajdosi@dhcp-24-156 ~]$ minishift start
   Driver is available at /home/agajdosi/minishift/docker-machine-driver-kvm ... 
-- Checking if Libvirt default network is present ... OK
-- Checking if Libvirt default network is active ... OK
-- Starting Minishift VM ........ OK
   Pinging 8.8.8.8 ... OK
-- Checking available disk space ... 10% used OK
   Version: v3.9.0
-- Copying oc binary from the OpenShift container image to VM . OK
-- Starting OpenShift cluster ......
Using nsenter mounter for OpenShift volumes
Using public hostname IP 192.168.42.56 as the host IP
Using 192.168.42.56 as the server IP
Starting OpenShift using openshift/origin:v3.9.0 ...
OpenShift server started.

    https://192.168.42.56:8443


Could not set oc CLI context for: 'minishift'
```

budhrg:
@praveenkumar , I am bit confused with following behavior
```
$ minishift stop

$ minishift status
Minishift:  Running
Profile:    minishift
OpenShift:  Stopped
DiskUsage:  10% of 19G
CacheUsage: 890.7 MB (used by oc binary, ISO or cached images)

$ minishift profile list
- foo   Running
- minishift Running   (Active)

$ minishift stop

$ minishift profile list
- foo   Running
- minishift Running   (Active)
```

@LalatenduMohanty @gbraad I think we are mixing the behavior of `minishift stop` here and it is creating confusion for me.
* `minishift stop` in non-remote scenario will stop the VM.
* `minishift stop` in remote scenario will stop the OpenShift.

I feel the need to some flag/switch here to avoid confusion.

Else with the current mind set, I am not able to digest this behavior. 
`I stopped profile -> profile list still showing 'Running'`.

@robin-owen WDYT about it too?

LalatenduMohanty:
> @LalatenduMohanty @gbraad I think we are mixing the behavior of minishift stop here and it is creating confusion for me.
> 
>     minishift stop in non-remote scenario will stop the VM.
>     minishift stop in remote scenario will stop the OpenS

The behavior is definitely not consistent with  rest of the VM drivers. However We can not create a new format to report the status as this format works for rest of the vm-driver. So that means we have very less space to figure out what else we ca do here.

LalatenduMohanty:
One thing we are sure that for `generic driver` we need to change the message which comes out of `minishift stop`. Because that not a local OpenShift cluster.  

```
$ minishift stop
```

* We can modify the output of `minishift status` to say it is Generic driver and Minishift does not control the life cycle of the VM. 

I am fine with  below. Lets see how user is reacting to that one. 
```
$ minishift profile list
- foo   Running
- minishift Running   (Active)
```

budhrg:
> One thing we are sure that for generic driver we need to change the message which comes out of minishift stop. Because that not a local OpenShift cluster.

Yes, that would mitigate the problem.

> I am fine with below. Lets see how user is reacting to that one.

But I am not. As, minishift is not local. We need to change the format as well. May be another task not in this PR.

The profile list should tell whether it is remote or local. 

Approving this PR as the basic functionality is fine but needs improvements.

LalatenduMohanty:
@praveenkumar Looks like there is a merge conflict in this PR.

amitkrout:
Running on a physical rhel remote host

Seems ok
```
$ ./minishift start --vm-driver generic --remote-ipaddress 10.70.45.10 --remote-ssh-user root --remote-ssh-key ~/.ssh/id_rsa
-- Downloading OpenShift binary 'oc' version 'v3.9.0'
 40.58 MiB / 40.58 MiB [==========================================================================================================] 100.00% 0s-- Downloading OpenShift v3.9.0 checksums ... OK
-- Starting local OpenShift cluster using 'generic' hypervisor ...
............. OK
[…]

To login as administrator:
    oc login -u system:admin

-- Exporting of OpenShift images is occuring in background process with pid 32218.
```
Seems ok if excluding @LalatenduMohanty  https://github.com/minishift/minishift/pull/2479#issuecomment-399404647
```
$ ./minishift stop
```
Seems ok, however displayed output does not look promising in case of the remote host. 
```
$ ./minishift delete --force
Deleting the Minishift VM...
Minishift VM deleted.
```

Docker images found after ```m delete``` in the remote rhel host. 
# docker images
REPOSITORY                                   TAG                 IMAGE ID            CREATED             SIZE
docker.io/openshift/origin-web-console       v3.9.0              542196230672        9 days ago          495 MB
docker.io/openshift/origin-docker-registry   v3.9.0              9a88c2195afe        9 days ago          465 MB
docker.io/openshift/origin-haproxy-router    v3.9.0              f14c9be6afe8        9 days ago          1.28 GB
docker.io/openshift/origin-deployer          v3.9.0              22e25a60e7f4        9 days ago          1.26 GB
docker.io/openshift/origin                   v3.9.0              40db23e999d6        9 days ago          1.26 GB
docker.io/openshift/origin-pod               v3.9.0              6718dea05165        9 days ago          223 MB
[root@minishift ~]#

@praveenkumar  if you think cleaning these images pulled by minishift through ```m delete``` is feasible to make a clean ```m delete``` then feel free to create another issue/pr.

praveenkumar:
>  if you think cleaning these images pulled by minishift through m delete is feasible to make a clean m delete then feel free to create another issue/pr.

@amitkrout No, I would not do that as part of delete step. This is not even feasible because we never know what all the images user pulled or was there any images etc.. before just cleaning everything from the docker image side.

LalatenduMohanty:
> We can modify the output of minishift status to say it is Generic driver and Minishift does not control the life cycle of the VM.

@praveenkumar Ignore this for now.

cmd/minishift/cmd/util/util.go
```diff
@@ -149,3 +152,12 @@ func ValidateGenericDriverFlags(remoteIPAddress, remoteSSHUser, sshKeyToConnectR
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error: %s", msg))
 	}
 }
+
+// OcClusterDown down the cluster and if there is any issue during the minishift start before even
+// oc binary cached then it just ignore the error [PK]
```
LalatenduMohanty: @praveenkumar I am able to understand the comment for this function. Can you please improve the explanation?

cmd/minishift/cmd/util/util.go
```diff
@@ -149,3 +152,12 @@ func ValidateGenericDriverFlags(remoteIPAddress, remoteSSHUser, sshKeyToConnectR
...
+func OcClusterDown(hostVm *host.Host) error {
+	sshCommander := provision.GenericSSHCommander{Driver: hostVm.Driver}
+	cmd := fmt.Sprintf("%s/oc cluster down", minishiftConstants.OcPathInsideVM)
+	sshCommander.SSHCommand(cmd)
```
LalatenduMohanty: Why are we not catching the error here? As per function definition  we should return error but we are not doing it.

cmd/minishift/cmd/stop.go
```diff
@@ -55,7 +55,15 @@ func runStop(cmd *cobra.Command, args []string) {
...
-	fmt.Println("Stopping local OpenShift cluster...")
+	fmt.Println("Stopping the OpenShift cluster...")
+
+	if hostVm.Driver.DriverName() == "generic" {
+		if err := util.OcClusterDown(hostVm); err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		}
+		fmt.Println("Cluster stopped.")
+		return
+	}
```
LalatenduMohanty: `fmt.Println("Cluster stopped.")` is there in two places in the function.  One for `generic` driver and another for non generic driver. IMO this is a place to use `else` to keep the logic separated and also helps with one `fmt.Println("Cluster stopped.")`.

Issue #2489, opened 2018-06-21
centos image doesn't start properly when proxies are in use - works on b2d


svenstaro:
### General information

  * Minishift version: 1.19.0
  * OS: Linux
  * Hypervisor: VirtualBox

### Steps to reproduce

Failing:

    minishift start --vm-driver virtualbox --http-proxy=http://51.15.86.88:3128 --iso-url centos

Working:

    minishift start --vm-driver virtualbox --http-proxy=http://51.15.86.88:3128 --iso-url b2d

Also working:

    minishift start --vm-driver virtualbox --iso-url centos

### Expected

Works fine like with b2d.

### Actual

Fails

### Logs

[centos.log which fails](https://paste.xinu.at/Qz6Os/)
[b2d.log which succeeds](https://paste.xinu.at/GN2N/)
[centos-without-proxy.log which also succeeds](https://paste.xinu.at/aJIs/)

As demonstrated above, the problem exists strictly with centos and proxies. I believe this to be a fairly critical problem as enterprises often use proxies and with the new version of minishift defaulting to centos this might lead to many surprises.

svenstaro:
The proxy I provided is a random proxy that you can you in order to reproduce the problem but you can use any proxy, of course.

amitkrout:
Able to reproduce the same error 
```
-- Starting OpenShift cluster ..............Error during 'cluster up' execution: Error starting the cluster. ssh command error:
command : /var/lib/minishift/bin/oc cluster up --use-existing-config --routing-suffix 192.168.99.115.nip.io --host-data-dir /var/lib/minishift/hostdata --host-volumes-dir /var/lib/minishift/openshift.local.volumes --host-pv-dir /var/lib/minishift/openshift.local.pv --host-config-dir /var/lib/minishift/openshift.local.config --http-proxy http://163.172.86.64:3128 --no-proxy localhost,127.0.0.1,172.30.1.1 --public-hostname 192.168.99.115
err     : exit status 1
output  : Using nsenter mounter for OpenShift volumes
Using public hostname IP 192.168.99.115 as the host IP
Using 192.168.99.115 as the server IP
Starting OpenShift using openshift/origin:v3.9.0 ...
-- Starting OpenShift container ...
   Creating initial OpenShift configuration
   Starting OpenShift using container 'origin'
   Waiting for API server to start listening
FAIL
   Error: cannot access master readiness URL https://192.168.99.115:8443/healthz/ready
   Details:
[...]

PR #2493, opened 2018-06-22
Issue #2489 Fix oc cluster up in case of proxy env for centos ISO

praveenkumar:


svenstaro:
I confirm this fix to be working.

amitkrout:
Working as expected on rhel iso
```
$ ./minishift config set iso-url file:///Users/amit/minishift-rhel7.iso 
No Minishift instance exists. New 'iso-url' setting will be applied on next 'minishift start'

$ ./minishift start --vm-driver virtualbox --http-proxy http://163.172.86.64:3128
-- Using proxy for the setup
[...]
-- Starting Minishift VM ..................................-- Setting proxy information ... OK
 OK
[...]
You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin

-- Exporting of OpenShift images is occuring in background process with pid 31426.
$

PR #2500, opened 2018-06-25
[WIP] Issue #1796 Fix implementation of config, instaceconfig and instance-state-config

praveenkumar:

PR #2502, opened 2018-06-26
Issue #1796 Fix implementation of config, instaceconfig and instance-state-config

praveenkumar:


praveenkumar:
Make sure we test the host folder behavior in this PR because now host folder is not the instance state config part.

anjannath:
There's also an instance-only flag for hostfolders..

gbraad:
Ah, you opened a new PR (wondered why #2500 was closed).

praveenkumar:
>  There's also an instance-only flag for hostfolders..

@anjannath yes that what is the instance specific and shouldn't be part of the state as it was before something @gbraad can explain more.

praveenkumar:
> wondered why #2500 was closed

I messed up with that PR earlier :( and original was present in a different system locally so had to close.

cmd/minishift/cmd/image/cache_config_add.go
```diff
@@ -44,17 +45,17 @@ func addConfiguredImage(cmd *cobra.Command, args []string) {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Invalid image name: %v", err))
 	}
 
-	minishiftConfig := getMinishiftConfig()
-	cacheImages := getConfiguredCachedImages(minishiftConfig)
-
+	cacheImages := minishiftConfig.InstanceConfig.CacheImage
```
gbraad: isn't `CacheImage` an array of things? If so, pluralize

cmd/minishift/cmd/image/cache_config_view.go
```diff
@@ -34,8 +35,7 @@ var (
 )
 
 func viewConfiguredImages(cmd *cobra.Command, args []string) {
-	minishiftConfig := getMinishiftConfig()
-	cacheImages := getConfiguredCachedImages(minishiftConfig)
+	cacheImages := minishiftConfig.InstanceConfig.CacheImage
```
gbraad: plural confirmed?

cmd/minishift/cmd/image/util.go
```diff
@@ -121,32 +120,6 @@ func sortImageNames(images map[string]bool) []string {
...
-	var slice []string
-	for _, s := range interfaceSlice {
-		slice = append(slice, s.(string))
-	}
-	return slice
-}
-
-func getConfiguredCachedImages(minishiftConfig config.MinishiftConfig) []string {
-	var cacheImages []string
-	if minishiftConfig[config.CacheImages.Name] == nil {
```
gbraad: It was plural before ...

cmd/minishift/cmd/image/util.go
```diff
@@ -121,32 +120,6 @@ func sortImageNames(images map[string]bool) []string {
...
-		cacheImages = []string{}
-	} else {
-		cacheImages = toStringSlice(minishiftConfig[config.CacheImages.Name].([]interface{}))
```
gbraad: it was even a stringslice

cmd/minishift/cmd/image/cache_config_remove.go
```diff
@@ -45,15 +46,16 @@ func removeConfiguredImage(cmd *cobra.Command, args []string) {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Invalid image name: %v", err))
 	}
 
-	minishiftConfig := getMinishiftConfig()
-	cacheImages := getConfiguredCachedImages(minishiftConfig)
+	cacheImages := minishiftConfig.InstanceConfig.CacheImage
```
gbraad: TBH, I actually think variable should be named `imagesToCache := ....ImagesToCache` or similar.
`CacheImages` would sound like the images that are available inside the cache?

BUT ...
but we also gave it a second meaning, as this same name, as this variable is used to import the images to the VM from the cache`. ... think, think, think ...

cmd/minishift/cmd/start.go
```diff
@@ -384,8 +384,8 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 		RemoteSSHUser:         viper.GetString(configCmd.RemoteSSHUser.Name),
 		SSHKeyToConnectRemote: viper.GetString(configCmd.SSHKeyToConnectRemote.Name),
 	}
-	minishiftConfig.InstanceConfig.VMDriver = machineConfig.VMDriver
-	minishiftConfig.InstanceConfig.Write()
+	minishiftConfig.InstanceStateConfig.VMDriver = machineConfig.VMDriver
```
gbraad: This is a tricky one to call 'state', maybe from our perspective it is... right...

cmd/minishift/cmd/start.go
```diff
@@ -520,7 +520,7 @@ func exportContainerImages(driver drivers.Driver, api libmachine.API, version st
 		return
 	}
 
-	images := viper.GetStringSlice(configCmd.CacheImages.Name)
+	images := minishiftConfig.InstanceConfig.CacheImage
```
gbraad: plural ...

pkg/minishift/config/instance_config.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/minishift/minishift/pkg/minishift/hostfolder/config"
+	"io/ioutil"
+	"os"
+)
+
+var InstanceConfig *InstanceConfigType
+
+type InstanceConfigType struct {
+	FilePath    string   `json:"-"`
+	CacheImage  []string `json:"cache-images"`
```
gbraad: also plural in the json, so better to stay consistent

pkg/minishift/constants/constants.go
```diff
@@ -57,7 +57,12 @@ func GetOpenshiftImageName(openshiftVersion string) string {
...
 }
 
+// GetInstanceStateConfigPath return the path of instance config json file
+func GetInstanceStateConfigPath() string {
+	return filepath.Join(constants.Minipath, "machines", constants.MachineName+"-state.json")
+}
+
 // GetInstanceConfigPath return the path of instance config json file
 func GetInstanceConfigPath() string {
-	return filepath.Join(constants.Minipath, "machines", constants.MachineName+".json")
```
gbraad: we might have an issue here for existing VMs

praveenkumar:
> The current implementation will break existing config for instances AFAICS

@gbraad it will not break if a VM start from scratch or from stop state but if you have a VM running and then use updated minishift binary for an add-on operation that is what going to break. Do you think we should carry both updated state config and old config for 1-2 release and then depreciate old one completely?

gbraad:
It does not break new instances, but existing ones. Right. I am not sure if allowing 1 or 2 releases is the right approach.

The problem is not with state, as that gets recreated. but what if we just read one of the files if it exists in the old location, else ignore and move on to the new one.... but in this case, how about adding a new config option? how does this behave now?

gbraad:
as discussed, checking for an exisitng **old** file, and **moving** it might be a better idea. In this case we have a posible upgrade strategy and no need to handle it within a few releases to really remove the file/check (if ever).

LalatenduMohanty:
> After applying the PR, we only read cache config from \<Profile\>.json, so we will ignore the existing cache-config in config.json.

This is serious IMO as cache-config setting will decide which images will get imported by default to the VM. Because we will ignore the older cache-config, user will be surprised and this can break his workflow. Also we are not providing any way to copy the configs to new `PROFILE.json` .

gbraad:
Have you verified this?
If so, the fix would be simple to duplicate the same move (and/or copy) strategy as we do for the config. Can't verify now

gbraad:
Any config the user made explicitly should be preserved.

LalatenduMohanty:
> The impact is that user would need to download the images from internet even if the images are available in cache.

Expanding the use-case for nodejs-ex application. ` nodejs-ex` needs `centos/nodejs-8-centos7` docker image

```
* $ minishift start 
* $ minishift images cache-config add centos/nodejs-8-centos7:latest
* $ minishift images export centos/nodejs-8-centos7:latest
* $ minishift delete
```
```
* minishift start (It will auto import the `centos/nodejs-8-centos7:latest`
* Deploy the nodejs-ex  (It will download the image from internet
```

Now apply the PR
```
* $ minishift start  ( it wont import `centos/nodejs-8-centos7:latest`)
* Deploy the nodejs-ex  (It will download the image from internet as the cached image is noy auto imported)
```

LalatenduMohanty:
@praveenkumar I think we need to add a method to copy the cache-configs from config.json to PROFILE.json.  After couple of releases we can remove the  method from the code.

cmd/minishift/cmd/root.go
```diff
@@ -108,6 +108,17 @@ var RootCmd = &cobra.Command{
 		// creating all directories for minishift run
 		createMinishiftDirs(state.InstanceDirs)
 
+		// Ensure the viper config file exists.
+		ensureConfigFileExists(constants.ConfigFile)
+
+		// Read the config file and get the details about existing image cache.
+		// This should be removed after 3-4 release of minishift. [PK]
```
LalatenduMohanty: @praveenkumar we do not need `[pk]` as it can be figured out using `git blame`. I will fix it while merging it.

cmd/minishift/cmd/root.go
```diff
@@ -117,14 +128,39 @@ var RootCmd = &cobra.Command{
...
-		// Create MACHINE_NAME.json
+		// Create MACHINE_NAME.json (machine config file)
 		minishiftConfig.InstanceConfig, err = minishiftConfig.NewInstanceConfig(minishiftConstants.GetInstanceConfigPath())
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating config for VM: %s", err.Error()))
 		}
 
+		// Check if there is already something in cacheImages from old config file and write it to new one.
+		// This should be removed after 3-4 release of minishift. [PK]
+		if cacheImages != nil {
```
LalatenduMohanty: Same here.

cmd/minishift/cmd/root.go
praveenkumar: this is just to make sure someone commented it in case retrieve it back.

LalatenduMohanty:
```
$ minishift images cache-config add centos/nodejs-8-centos7:latest

$ cat test.json
{
	"cache-images": [
		"centos/nodejs-8-centos7:latest"
	],
	"HostFolders": []
}

```
`$ minishift image cache-config view` does not show anything
and 
```
$ cat test.json
{
	"cache-images": [],
	"HostFolders": []
}
```

cmd/minishift/cmd/root.go
LalatenduMohanty: @praveenkumar we do not need names in comments.  But as I mentioned no need to fix it as I can change it and then merge.

LalatenduMohanty:
Another issue : The new binary will ignore  ` --instance-only` hostfolders if anything added before moving to the new binary.

On the master branch

```
$ minishift delete

$ minishift hostfolder add myshare3 --source /Users/lmohanty/myshare/ --target /mnt/sda1/myshare --type sshfs  --instance-only

$ minishift hostfolder list
Name       Type    Source                     Mountpoint          Mounted
myshare    sshfs   /Users/lmohanty/myshare/   /mnt/sda1/myshare   N
myshare2   sshfs   /Users/lmohanty/myshare/   /mnt/sda1/myshare   N
myshare3   sshfs   /Users/lmohanty/myshare/   /mnt/sda1/myshare   N
```

Use the PR
```
$ minishift hostfolder list
I am here []string{"centos/nodejs-8-centos7:latest"}
Name       Type    Source                     Mountpoint          Mounted
myshare    sshfs   /Users/lmohanty/myshare/   /mnt/sda1/myshare   N
myshare2   sshfs   /Users/lmohanty/myshare/   /mnt/sda1/myshare   N
```

cmd/minishift/cmd/root.go
```diff
@@ -117,14 +128,40 @@ var RootCmd = &cobra.Command{
...
+		// Create MACHINE_NAME.json (machine config file)
 		minishiftConfig.InstanceConfig, err = minishiftConfig.NewInstanceConfig(minishiftConstants.GetInstanceConfigPath())
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating config for VM: %s", err.Error()))
 		}
 
+		// Check if there is already something in cacheImages from old config file and write it to new one.
+		// This should be removed after 3-4 release of minishift. [PK]
+		if len(cacheImages) != 0 {
+			fmt.Printf("I am here %#v\n", cacheImages)
```
LalatenduMohanty: Unwanted print statement.

LalatenduMohanty:
Tested the PR for `cache-config`, `hostfolder --instance-only`, `config set/view` and everything is working properly now.

cmd/minishift/cmd/root.go
```diff
@@ -117,14 +128,49 @@ var RootCmd = &cobra.Command{
 			}
 		}
 
-		ensureConfigFileExists(constants.ConfigFile)
+		// Check if old instance state config file exists and move it to new one.
+		// Create MACHINE_NAME-state.json (machine state config file)
```
LalatenduMohanty: suggestion: `If older instance state config file exists and copy it to then rename the file else create a the new file`.

cmd/minishift/cmd/root.go
```diff
@@ -117,14 +128,49 @@ var RootCmd = &cobra.Command{
...
+		}
+
+		// Create MACHINE_NAME.json (machine config file)
 		minishiftConfig.InstanceConfig, err = minishiftConfig.NewInstanceConfig(minishiftConstants.GetInstanceConfigPath())
 		if err != nil {
 			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating config for VM: %s", err.Error()))
 		}
 
+		// Check if there is already something in hostfolder from old config file and write it to new one.
+		// This should be removed after 2-3 release of minishift.
```
LalatenduMohanty: `If host folder config exists for an instance then copy it to the new instance state config`. `This change should be removed after few releases as we expect our users will not need this`.

cmd/minishift/cmd/root.go
```diff
@@ -117,14 +128,49 @@ var RootCmd = &cobra.Command{
...
+		hostfolder := minishiftConfig.InstanceStateConfig.HostFolders
+		if len(hostfolder) != 0 {
+			minishiftConfig.InstanceConfig.HostFolders = hostfolder
+			if err != minishiftConfig.InstanceConfig.Write() {
+				atexit.ExitWithMessage(1, fmt.Sprintf("Error coping existing hostfolder to new instance config: %s", err.Error()))
+			}
+		}
+
+		// Check if there is already something in cacheImages from old config file and write it to new one.
+		// This should be removed after 2-3 release of minishift.
```
LalatenduMohanty: `If cache-config exists then copy it to the new instance config. This change should be removed after few releases as we expect our users will not need this.`

LalatenduMohanty:
Merged https://github.com/minishift/minishift/commit/781349690090db8711a30d0f5fd3d84accc27732 . Thanks @praveenkumar

Issue #2128, opened 2018-03-14
Investigate whether there are better sshfs mount options we can use


hferentschik:
Currently, we use:

```golang
		cmd := fmt.Sprintf(
			"sudo sshfs docker@%s:%s %s -o IdentityFile=%s -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p %d",
			ip,
			h.config.Option(config.Source),
			h.config.MountPoint(),
			keyFile,
			SftpPort)
```

There are many more options which might help to improve performance and or file permissions. Some investigation and experiments around this would be nice. 

We could also let the user specify his own options, amending or overriding the default mount options. Technically we should be able to use the `--options` flag of the `hostfolder add` command.

anjannath:
I will do the required investigation about sshfs options.

anjannath:
Still Todo..

anjannath:
There are a couple of options we can use by default such as.. `Ciphers=arcfour`, `Compression=no`, `cache=yes` mostly these will improve performance.

anjannath:
The `--options` flag work will be done as part of #2157

anjannath:
I think we can just close this, WDYT @minishift/minishift-dev ?

LalatenduMohanty:
@anjannath This issue was created to see if we want to add something to default experience of sshfs host folder. Even through users can use these options through `--options flag ` but it wont be part of default behavior. So the question do we want to add any option to the default user experience?

anjannath:
@LalatenduMohanty I have tried other options like these `-o compression=no -o kernel_cache -o large_read -o`(from the manpage of sshfs) that you can see in this [commit](https://github.com/anjannath/minishift/commit/4754ef2e288e94bcd7fc988a3c509343f520788d), but not sure how they impact the performance/stability of sshfs mounts. I don't think we need any additional options for it by default.

anjannath:
The other options were tried and there was no noticeable performance gains, so the current options works for our scenario with sshfs.

PR #2504, opened 2018-06-26
Issue #2128 Additional options for sshfs mount

anjannath:
Fixes: #2128 
<strike>- use `fusermount -u` for unmounting</strike>

### DIFF
--- a/pkg/minishift/hostfolder/sshfs_hostfolder.go
+++ b/pkg/minishift/hostfolder/sshfs_hostfolder.go
@@ -74,8 +74,11 @@ func (h *SSHFSHostFolder) Mount(driver drivers.Driver) error {
 
 	// Mount command seems to fail occasionally. Give it a couple of attempts
 	mount := func() (err error) {
+		// compression=no: do not use compression for the connection
+		// kernel_cache: allow kernel to cache files
+		// large_read: read large byte chunks
 		cmd := fmt.Sprintf(
-			"sudo sshfs docker@%s:%s %s -o IdentityFile=%s -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p %d",
+			"sudo sshfs docker@%s:%s %s -o compression=no -o kernel_cache -o large_read -o IdentityFile=%s -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -p %d",
 			ip,
 			h.config.Option(config.Source),
 			h.config.MountPoint(),


pkg/minishift/hostfolder/sshfs_hostfolder.go
```diff
@@ -75,7 +75,7 @@ func (h *SSHFSHostFolder) Mount(driver drivers.Driver) error {
 	// Mount command seems to fail occasionally. Give it a couple of attempts
 	mount := func() (err error) {
 		cmd := fmt.Sprintf(
-			"sudo sshfs docker@%s:%s %s -o IdentityFile=%s -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p %d",
+			"sudo sshfs docker@%s:%s %s -o compression=no -o kernel_cache -o IdentityFile=%s -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -p %d",
```
anjannath: the default value for `idmap` is none, so removed it.
```
 -o idmap=TYPE          user/group ID mapping, possible types are:
             none             no translation of the ID space (default)
             user             only translate UID of connecting user
             file             translate UIDs/GIDs contained in uidfile/gidfile
```

pkg/minishift/hostfolder/sshfs_hostfolder.go
```diff
@@ -102,7 +102,7 @@ func (h *SSHFSHostFolder) Mount(driver drivers.Driver) error {
 }
 
 func (h *SSHFSHostFolder) Umount(driver drivers.Driver) error {
-	cmd := fmt.Sprintf("sudo umount %s", h.config.MountPoint())
+	cmd := fmt.Sprintf("sudo fusermount -u %s", h.config.MountPoint())
```
praveenkumar: @anjannath What is the advantage of using `fusermount` instead `umount` also does that command available on each iso which we have current support?

anjannath: I am still looking for the advantages of it over umount, just that it was listed on the man page for [SSHFS](https://linux.die.net/man/1/sshfs). It is available in the centos iso and b2d, moreover it is user space cli for the FUSE file system so it should be present on all the isos.

pkg/minishift/hostfolder/sshfs_hostfolder.go
gbraad: have you verified this to be the case on several distros?

pkg/minishift/hostfolder/sshfs_hostfolder.go
gbraad: Can you verify that this command is available? I would expect so, as it is part of our ISO... however, we change something that works and is consistent between all other implementations? Without any benefit, i would not see a need to change this.

pkg/minishift/hostfolder/sshfs_hostfolder.go
gbraad: Also, are these options essential and/or in what way to they provide benefit?

pkg/minishift/hostfolder/sshfs_hostfolder.go
anjannath: @gbraad Yes i have verified it is available in all b2d, centos iso, i haven't checked in the fedora iso though..

pkg/minishift/hostfolder/sshfs_hostfolder.go
anjannath: @gbraad It is like this for the b2d and centos iso, these are not essential, but the issue was to use some _additonal_ options which would improve performance, now i haven't done a benchmark on this but these are supposed to improve the performance, as no compression could speed up the process and kernel_cache would use the cache..

pkg/minishift/hostfolder/sshfs_hostfolder.go
anjannath: Also i think, umount internally calls `fusermount -u` so this might not be necessary.. but since it was mentioned on the SSHFS man page i decided to go with it, we can change back to `umount` as that was also working fine.

pkg/minishift/hostfolder/sshfs_hostfolder.go
gbraad: `compression=no` might speed up on older machines, but it does consume less IOPS for the actual transfer. although this might even be possible to be offloaded on some systems. so, yeah... I am not sure it actually improves without evidence.

other options we have added before, but their effect were instantly noticable... as for instance it enabled SELinux, etc. Here, I am not really convinced

pkg/minishift/hostfolder/sshfs_hostfolder.go
gbraad: well, for mount in that case we could also use `mount.xfs`, `mount,fuse`, etc... but when the abstraction makes sense, please use it. `umount` works for everything as expected; let the abstraction do its' work.

anjannath: @gbraad Okay.. reverting this change.

pkg/minishift/hostfolder/sshfs_hostfolder.go
anjannath: @gbraad I will try to find some benchmarks if there is any for optimizing sshfs.

anjannath:
@gbraad Found this article which does a benchmark/comparison with these options: 
http://www.admin-magazine.com/HPC/Articles/Sharing-Data-with-SSHFS

pkg/minishift/hostfolder/sshfs_hostfolder.go
gbraad: Especially as what is changed here is not in line with the title of the commit. It feels like additional options are added (which are perhaps even configurable).

Note: pre-mature optimization is never a good idea, especially without identifiable improvements. this would be a change without a clear reason. what would `kernel_cache` do? it makes no sense without more details or even an explanation.

anjannath: @gbraad From the help section: `kernel_cache` caches files in kernel. normal caching is by-default turned on, this allows the kernel to also cache, this will speed up subsequent reads.

The other option `large_files` is also for speeding up reads. by reading large chunks we don't have to make frequent read operations, so less `read()` and `open()` calls, i assume.

gbraad:
the article also referred to a Cipher and using large_read. but without any indication, I could not see how this provides us any improvements.

anjannath:
I did not use the `Cipher` option, it is not available in the sshfs we are using (could not find it in the help or man page), however the `large_read` option is there and i have used it. It reduces the number of read operations, and reads large amounts of chunks in one go.

anjannath:
@gbraad Added comments to code!

anjannath:
@minishift/minishift-dev We don't necessarily need these options for sshfs to work, so we can close this and the related issue.

anjannath:
Closing this.

Issue #2464, opened 2018-06-18
Implement a way to specify an empty value for the variable using Var-Defaults for addon


praveenkumar:
As of now, we don't have any capability to specify the empty value for `Var-Defaults` and if I look closely to https://github.com/minishift/minishift-addons/pull/124#pullrequestreview-128642538 conversation looks like it a valid use case and we should have a way to do it. 

I am thinking to define a keyword `NULL` or `null` to do it so during the variable parse we check the value and set it as empty in case of NULL/null/Null.

@minishift/minishift-dev any better idea?

LalatenduMohanty:
> looks like it a valid use case and we should have a way to do it.

+1. Wondering how we have not see this till now.  I think if `# Var-Defaults: PROJECT_USER=`  has no value it should be considered no value.

praveenkumar:
> I think if # Var-Defaults: PROJECT_USER= has no value it should be considered no value.

@LalatenduMohanty We have a `varDefaultsCheck` function which actually there to make sure user put correct convention about the var defaults which will be `key=value` if we disable that check that means if a user misses specifying the value which he should've then his addon might configure something which not intended to do. I think it good to be explicit instead of implicit about the value of variable.

LalatenduMohanty:
Resolved via #2505

PR #2505, opened 2018-06-26
Issue #2464 Addon: Specify empty value for Var-Defaults

praveenkumar:
How to test
--------------
```
$ cat ~/.minishift/addons/test/test.addon
# Name: test
# Description: Test.
# Required-Vars: FOO
# Var-Defaults: FOO=null

echo Out of 'FOO'
echo #{FOO}

$ ./minishift addon apply test
-- Applying addon 'test':
Out of 'FOO'

```

### DIFF
--- a/pkg/minishift/addon/addon_meta.go
+++ b/pkg/minishift/addon/addon_meta.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	minishiftStrings "github.com/minishift/minishift/pkg/util/strings"
 	"regexp"
+	"strings"
 )
 
 var requiredMetaTags = []string{NameMetaTagName, DescriptionMetaTagName}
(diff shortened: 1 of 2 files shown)


pkg/minishift/addon/addon_meta.go
```diff
@@ -95,6 +96,10 @@ func (meta *DefaultAddOnMeta) VarDefaults() ([]RequiredVar, error) {
 		defaults := make([]RequiredVar, 0, len(items))
 		for _, item := range items {
 			varDefault, _ := minishiftStrings.SplitAndTrim(item, "=")
+			// In case of value of varDefault is null/Null/NULL then value converted to empty string.
+			if strings.ToLower(varDefault[1]) == "null" {
+				varDefault[1] = ""
+			}
```
praveenkumar: https://github.com/minishift/minishift/issues/2464#issuecomment-400266767

praveenkumar:
retest this please.

pkg/minishift/addon/addon_meta.go
gbraad: not sure if I like to see `null`... or rather `nil`.
nil has a clearer meaning in language than `null`, as the last one means related to zero or invalid, while nil means non-existant.

does it actually NEED a value? why is `FOO=` not possible?

praveenkumar: > nil has a clearer meaning in language than null, as the last one means related to zero or invalid, while nil means non-existant.

@gbraad sure that something we can modify

> does it actually NEED a value? why is FOO= not possible?

@gbraad https://github.com/minishift/minishift/issues/2464#issuecomment-400266767

LalatenduMohanty: Both `nil` or `null` works for me.

LalatenduMohanty:
ok to test

Issue #2506, opened 2018-06-26
Hostfolder mount fails with sshfs if 2022 port not open on the host firewalld


praveenkumar:
```
$ sudo iptable-save
[...]
-A IN_FedoraWorkstation_allow -p tcp -m tcp --dport 22 -m conntrack --ctstate NEW -j ACCEPT

$ ./minishift hostfolder list
Name   Type    Source   Mountpoint       Mounted
test   sshfs   /home/prkumar/Documents     /mnt/sda1/test   N
```
We should improve the below error message 
```
$ ./minishift hostfolder mount test
error occurred while mounting host folder: error occured while mounting host folder: ssh command error:
command : sudo sshfs docker@192.168.42.1:/home/prkumar/Documents /mnt/sda1/test -o IdentityFile=/home/docker/.ssh/id_rsa -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p 2022
err     : exit status 1
output  : read: Connection reset by peer

```
```
# iptables -A IN_FedoraWorkstation_allow -p tcp -m tcp --dport 2022 -m conntrack --ctstate NEW -j ACCEPT

# iptables-save
$ ./minishift hostfolder mount test
$ ./minishift hostfolder list
Name    Type    Source                    Mountpoint        Mounted
test    sshfs   /home/prkumar/Documents   /mnt/sda1/test    Y
```

anjannath:
You can also change/set the port for the SFTP server, there's a config option for this: `hostfolders-sftp-port` I think the docs already mention this.

praveenkumar:
@anjannath but whatever port you chose should be allowed by the firewall, does docs mention this, if yes link please?

anjannath:
@praveenkumar Not with instructions for how to enable it. Just a _Tip_ mentioning that firewall should allow comm. to this port.. I can't link to it as it is not a section.. it under this section: 
https://docs.openshift.org/latest/minishift/using/host-folders.html#mounting-host-folders

LalatenduMohanty:
Resolved #2517

PR #2517, opened 2018-06-28
Issue #2506 Update error message for sshfs hostfolder mount

praveenkumar:
@gbraad @anjannath this is now error looks like, I am not sure if cifs also need it.

```
$ ./minishift hostfolder mount test

Note: Make sure that your network and firewall settings on the host allows port 2022 to be opened

error occured while mounting host folder: ssh command error:
command : sudo sshfs docker@192.168.42.1:/home/prkumar/Documents /mnt/sda1/test -o IdentityFile=/home/docker/.ssh/id_rsa -o 'StrictHostKeyChecking=no' -o reconnect -o allow_other -o idmap=none -p 2022
err     : exit status 1
output  : read: Connection reset by peer

```

### DIFF
--- a/pkg/minishift/hostfolder/sshfs_hostfolder.go
+++ b/pkg/minishift/hostfolder/sshfs_hostfolder.go
@@ -95,7 +95,8 @@ func (h *SSHFSHostFolder) Mount(driver drivers.Driver) error {
 
 	err = util.Retry(3, mount)
 	if err != nil {
-		return fmt.Errorf("error occurred while mounting host folder: %s", err)
+		errMsg := fmt.Sprintf("\nNote: Make sure that your network and firewall settings on the host allows port %d to be opened\n\n", SftpPort)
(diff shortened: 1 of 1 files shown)


praveenkumar:
> Otherwise it looks unnatural to suggest something before the error is printed.

@LalatenduMohanty We are putting a `Note` before the error which means that error can be caused by it or not but when we put `Note` later that means the error is caused by this so I think it is correct sequence.

LalatenduMohanty:
> @LalatenduMohanty We are putting a Note before the error which means that error can be caused by it or not but when we put Note later that means the error is caused by this so I think it is correct sequence.

I disagree, it looks odd to me as I go through the error because when I do not know what is the error I can not relate to the note and after reading the error, I wanted to read the note again. But if I know what is the error then I can relate to the note better. Also it also gives a feeling that somehow Minishift knows the before it came but the reverse sounds more natural to me.

PR #2527, opened 2018-06-29
Issue #2526 use subscription-manager list to check rhsm status

anjannath:

Issue #2177, opened 2018-03-26
Display registration status in minishift status command


budhrg:
For RHEL ISO, one need to perform registration to leverage RHEL feature inside VM.

Displaying registration status in `minishift status` will help user to know the state of registration.

Note: This should be displayed only for the RHEL ISO.

PR #2533, opened 2018-07-01
Issue #2177 display registration info in status command

anjannath:
#2527 is needed to make this work smoothly


cmd/minishift/cmd/status.go
```diff
@@ -43,6 +44,14 @@ DiskUsage:  {{.DiskUsage}}
 CacheUsage: {{.CacheUsage}} (used by oc binary, ISO or cached images)
 `
 
+var statusFormatWithRegistration = `Minishift:  {{.MinishiftStatus}}
+Profile:    {{.ProfileName}}
+OpenShift:  {{.ClusterStatus}}
+RHSM: 	    {{.Registration}}
```
gbraad: I would place it at the bottom, so it does not affect scripts that might use line numbers

cmd/minishift/cmd/status.go
```diff
@@ -43,6 +44,14 @@ DiskUsage:  {{.DiskUsage}}
 CacheUsage: {{.CacheUsage}} (used by oc binary, ISO or cached images)
 `
 
+var statusFormatWithRegistration = `Minishift:  {{.MinishiftStatus}}
+Profile:    {{.ProfileName}}
+OpenShift:  {{.ClusterStatus}}
+DiskUsage:  {{.DiskUsage}}
+CacheUsage: {{.CacheUsage}} (used by oc binary, ISO or cached images)
+RHSM: 	    {{.Registration}}
```
praveenkumar: @anjannath looks like there is a space needed in between `RHSM:` and `{{.Registration}}` to make it same as other variables in the template.

anjannath: @praveenkumar It appears properly aligned in my editor locally, could you please check locally once, if it is not aligned i'll modify then.

praveenkumar: @anjannath nvm, But why you used a different variable for registration purpose? I am hoping to have it part of `statusFormat` variable only and ISO who doesn't support registration can have `Not Applicable` as the output.

budhrg:
@anjannath Need rebase.

cmd/minishift/cmd/status.go
anjannath: @praveenkumar I thought about it too, but this looks more cleaner, at the cost of an extra string variable so..  :-)

anjannath:
@budhrg Done!

PR #2546, opened 2018-07-03
Issue #2475 Handle proxy specified without scheme

### LINKED ISSUE #2475: Protocol-less proxies are accepted by the command line parser but are not used
### General information

  * Minishift version: 1.19.0
  * OS: Linux
  * Hypervisor: VirtualBox

### Steps to reproduce

    HTTP_PROXY=1.1.1.1 ./minishift start --vm-driver virtualbox

### Expected

Should use the proxy and output

    -- Using proxy for the setup

### Actual

Doesn't use the proxy and doesn't output

    -- Using proxy for the setup

It works when you provide the proxy like 

    HTTP_PROXY=http://1.1.1.1 ./minishift start --vm-driver virtualbox

I can post a full log but I don't think it's necessary as this appears to be a parsing error of some kind. Same thing happens with --http-proxy and respective HTTPS versions.

I believe that 1.1.1.1 should be parsed as http://1.1.1.1 or alternatively throw an error. It should not silently fail, however.

gbraad:
Fixes #2475

Handle proxy specified without scheme

### DIFF
--- a/pkg/util/proxy.go
+++ b/pkg/util/proxy.go
@@ -45,11 +45,11 @@ func NewProxyConfig(httpProxy string, httpsProxy string, noProxy string) (*Proxy
 			httpProxy = os.Getenv("HTTP_PROXY")
 		}
 	}
-	err := ValidateProxyURL(httpProxy)
+	err := ValidateProxyURL(httpProxy, "http")
 	if err != nil {
 		return nil, err
 	}
-	httpProxy = parseProxySpecialChar(httpProxy)
+	httpProxy = parseProxySpecialChar(httpProxy, "http")
 
 	if httpsProxy == "" {
 		httpsProxy = os.Getenv("https_proxy")
@@ -57,11 +57,11 @@ func NewProxyConfig(httpProxy string, httpsProxy string, noProxy string) (*Proxy
 			httpsProxy = os.Getenv("HTTPS_PROXY")
 		}
 	}
-	err = ValidateProxyURL(httpsProxy)
+	err = ValidateProxyURL(httpsProxy, "https")
 	if err != nil {
 		return nil, err
 	}
-	httpsProxy = parseProxySpecialChar(httpsProxy)
+	httpsProxy = parseProxySpecialChar(httpsProxy, "https")
 
 	np := []string{}
 	np = append(np, defaultNoProxies...)
@@ -158,9 +158,13 @@ func (p *ProxyConfig) IsEnabled() bool {
 }
 
 // ValidateProxyURL validates that the specified proxyURL is valid
-func ValidateProxyURL(proxyUrl string) error {
+func ValidateProxyURL(proxyUrl string, scheme string) error {
 	if proxyUrl == "" {
 		return nil
+	}
+
+	if !strings.HasPrefix(proxyUrl, scheme) {
+		proxyUrl = fmt.Sprintf("%s://%s", scheme, proxyUrl)
 	}
 
 	if !govalidator.IsURL(proxyUrl) {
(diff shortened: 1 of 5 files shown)


pkg/util/proxy.go
```diff
@@ -170,13 +170,18 @@ func ValidateProxyURL(proxyUrl string) error {
 }
 
 // parseProxySpecialChar parse the URI and convert special char to hex
-func parseProxySpecialChar(uri string) string {
+func parseProxySpecialChar(uri string, scheme string) string {
```
praveenkumar: :+1:

pkg/util/proxy.go
```diff
@@ -170,13 +170,18 @@ func ValidateProxyURL(proxyUrl string) error {
...
+	if uri == "" {
+		return ""
+	}
+
 	u, _ := url.Parse(uri)
-	scheme := fmt.Sprintf("%s://", u.Scheme)
 	if strings.HasPrefix(uri, scheme) {
+		scheme := fmt.Sprintf("%s://", u.Scheme)
```
praveenkumar: :+1:

gbraad:
@svenstaro does this solve your issue?

svenstaro:
Not quite happy with the fix: `--http-proxy=173.249.48.140` now works but `--http-proxy=173.249.48.140:8080` is an error. Either both of those work or both are errors.

pkg/util/proxy_test.go
```diff
@@ -172,15 +172,16 @@ func Test_parse_special_character_uri(t *testing.T) {
...
-		{"", ""},
-		{"http://foo.com:3128", "http://foo.com:3128"},
-		{"http://user:F@oo!B#ar$@myserver:3128", "http://user:F@oo%21B%23ar$@myserver:3128"},
-		{"https://myuser:my#pass@foo.com:3128", "https://myuser:my%23pass@foo.com:3128"},
-		{"https://newuser:new(pas*)wrd@test.com:3128", "https://newuser:new%28pas%2A%29wrd@test.com:3128"},
+		{"", "", "http"},
+		{"http://foo.com:3128", "http://foo.com:3128", "http"},
+		{"http://user:F@oo!B#ar$@myserver:3128", "http://user:F@oo%21B%23ar$@myserver:3128", "http"},
+		{"https://myuser:my#pass@foo.com:3128", "https://myuser:my%23pass@foo.com:3128", "https"},
+		{"https://newuser:new(pas*)wrd@test.com:3128", "https://newuser:new%28pas%2A%29wrd@test.com:3128", "https"},
```
LalatenduMohanty: @gbraad shouldn't we test for  something like`{"foo.com:3128", "http://foo.com:3128", "http"}` ?

gbraad:
this seems to be the standard behaviour for `!govalidator.IsURL(proxyUrl)`. We will have to add the prefix before the validation happens...

gbraad:
@svenstaro I am not seeing that problem

```
minishift start --http-proxy=10.0.75.101:3128
-- Using proxy for the setup
-- Checking if Libvirt default network is present ... OK
-- Checking if Libvirt default network is active ... OK
-- Starting the OpenShift cluster using 'kvm' hypervisor ...
   Memory:    1 GB
   vCPUs :    1
   Disk size: 5 GB
-- Starting Minishift VM ..............................-- Setting proxy information ... OK
 OK
   Pinging 10.0.75.1 ... OK
   Retrieving http://10.0.75.1:5000/v2/ ... FAIL
   VM cannot connect to external URL with HTTP
   Importing 'openshift/origin:v3.9.0' ....................... OK
   Importing 'openshift/origin-docker-registry:v3.9.0' ........ OK
   Importing 'openshift/origin-haproxy-router:v3.9.0' ........... OK
   Version: v3.9.0
-- Pulling the Openshift Container Image .. EXISTS
-- Copying oc binary from the OpenShift container image to VM ...... OK
-- Starting OpenShift cluster ................
```

gbraad:
@svenstaro settings take precedence, so check what is set with `m config view`

pkg/util/proxy_test.go
gbraad: Added, but this problem does not occur here. Let's see what CI says...

Issue #2457, opened 2018-06-13
Add integration test for --vm-driver generic


agajdosi:
Test should start a VM using --no-provision flag, get its IP and later run `minishift start --vm-driver generic` to the specified IP address.

Test should also focus on some edge cases and possibly problematic aspects of the --vm-driver generic.

agajdosi:
@gbraad Was planning the same, using --no-provisioned VM seems to me like acceptable solution. Created PR #2553 just now, please check if you have time.

PR #2553, opened 2018-07-03
Issue #2457 Add integration test for generic driver

agajdosi:
Fix https://github.com/minishift/minishift/issues/2457

### DIFF
--- a/test/integration/features/generic-driver.feature
+++ b/test/integration/features/generic-driver.feature
@@ -1,0 +1,61 @@
+@generic-driver
+Feature: With generic driver Minishift can provision remote unprovisioned VM 
+
+  Scenario: User creates an unprovisioned VM
+    Given setting up environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" with value "y" succeeds
+     When executing "minishift profile set remote" succeeds
+      And executing "minishift start --no-provision" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: Provisioning remote instance
+    Given unset environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" succeeds
+      And setting scenario variable "REMOTE_VM_IP" to the stdout from executing "minishift ip"
+      And executing "minishift profile set minishift" succeeds
+      And image caching is disabled
+     When executing "minishift start --vm-driver generic --remote-ipaddress $(REMOTE_VM_IP) --remote-ssh-key .minishift/profiles/remote/machines/remote/id_rsa --remote-ssh-user docker" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: Stopping with generic driver
+  Stopping with generic driver puts oc cluster down, remote VM however keeps running.
+     When executing "minishift stop" succeeds
+      And executing "minishift status" succeeds
+     Then stdout should match "OpenShift:\s+Stopped"
+
+  Scenario: VM at remote location is still running
+     When executing "minishift status --profile remote" succeeds
+     Then stdout should match "Minishift:\s+Running"
+      And stdout should match "OpenShift:\s+Stopped"
+
+  Scenario: Starting stopped cluster with generic driver
+     When executing "minishift start" succeeds
+     Then executing "minishift status" succeeds
+      And stdout should match "Minishift:\s+Running"
+      And stdout should match "OpenShift:\s+Running"
+  
+  Scenario: Remote VM is running and cluster is up
+     When executing "minishift status --profile remote" succeeds
+     Then stdout should match "Minishift:\s+Running"
+      And stdout should match "OpenShift:\s+Running"
+
+  Scenario: Deleting remote instance
+     When executing "minishift delete --force" succeeds
+     Then Minishift should have state "Does Not Exist"
+
+  Scenario: VM at remote location is still running
+     When executing "minishift profile set remote" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: OpenShift configuration is not present on remote VM
+  NOTE: Directory openshift.local.volumes will be left at the place due to permision problems.
+  This can be later removed once the issue is fixed. More information at:
+  https://github.com/minishift/minishift/blob/a8c8b37a29119bad216524c1fd067a5a3a00b85f/cmd/minishift/cmd/delete.go#L141-L142
+     When executing "minishift ssh -- ls -a /var/lib/minishift" succeeds
+     Then stdout should not contain "base"
+      And stdout should not contain "bin"
+      And stdout should not contain "hostdata"
+      And stdout should not contain "openshift.local.config"
+      And stdout should not contain "openshift.local.pv"
+
+  Scenario: Deleting remote VM
+     When executing "minishift delete --force" succeeds
+     Then Minishift should have state "Does Not Exist"


test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,49 @@
...
+    When executing "minishift stop" succeeds
+    And executing "minishift status" succeeds
+    Then stdout should match "OpenShift:\s+Stopped"
+
+  Scenario: VM at remote location is still running
+    When executing "minishift status --profile remote" succeeds
+    Then stdout should match "Minishift:\s+Running"
+    And stdout should match "OpenShift:\s+Stopped"
+
+  Scenario: Starting stopped cluster with generic driver
```
amitkrout: I don not think its starting the remote stooped openshift cluster instead its starting a cluster inside minishift profile. I am expecting the steps should be something like
```
minishift start --vm-driver generic --remote-ipaddress $(REMOTE_VM_IP) --remote-ssh-key .minishift/profiles/remote/machines/remote/id_rsa --remote-ssh-user docker" succeeds
```

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,49 @@
...
+    When executing "minishift start" succeeds
+    Then executing "minishift status" succeeds
+    And stdout should match "Minishift:\s+Running"
+    And stdout should match "OpenShift:\s+Running"
+  
+  Scenario: Remote VM is running and cluster is up
+    When executing "minishift status --profile remote" succeeds
+    Then stdout should match "Minishift:\s+Running"
+    And stdout should match "OpenShift:\s+Running"
```
amitkrout: Wondering how this is in running state because nowhere in previous step it is being started

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,49 @@
+@generic-driver
+Feature: With generic driver Minishift can provision remote unprovisioned VM 
+
+  Scenario: User creates an unprovisioned VM
+    Given setting up environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" with value "y" succeeds
+    When executing "minishift profile set remote" succeeds
+    And executing "minishift start --no-provision" succeeds
```
amitkrout: +1

agajdosi:
@budhram Can you point me to the problematic lines? Went through it and I can't find the place.

test/integration/features/generic-driver.feature
agajdosi: @amitkrout It is started at line 7.

amitkrout: You started it on remote profile and then at line no 19 it is stopped

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,49 @@
...
+    Then Minishift should have state "Running"
+
+  Scenario: Provisioning remote instance
+    Given unset environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" succeeds
```
budhram: The alignment is right justified not left justified :) @agajdosi  See below

screenshot from 2018-07-04 18-42-49

test/integration/features/generic-driver.feature
agajdosi: But at line 29 I start the OpenShift again with generic driver, so here both VM (generic does not affect that at all) and the OpenShift are running on the remote machine.

test/integration/features/generic-driver.feature
agajdosi: @amitkrout From my observations it does not need to repeat the flags on start from stopped. These are stored and used, this step does not create new VM locally.

agajdosi:
@budhram formatting fixed

test/integration/features/generic-driver.feature
amitkrout: > But at line 29 I start the OpenShift again with generic driver

How? Ideally, we are not saving any flags like ```--vm-driver```, ```--remote-ipaddress``` ...in the config or even in env. Then how ```minishift start``` will consider these variable.

test/integration/features/generic-driver.feature
amitkrout: It's very hard to believe that it's not creating ```minishift``` vm. I think lets @praveenkumar  share his thoughts on the same

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,49 @@
...
+
+  Scenario: User creates an unprovisioned VM
+    Given setting up environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" with value "y" succeeds
+     When executing "minishift profile set remote" succeeds
+      And executing "minishift start --no-provision" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: Provisioning remote instance
+    Given unset environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" succeeds
+      And setting scenario variable "REMOTE_VM_IP" to the stdout from executing "minishift ip"
```
budhram: This is nice :+1:

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,49 @@
...
+
+  Scenario: Deleting remote instance
+     When executing "minishift delete --force" succeeds
+     Then Minishift should have state "Does Not Exist"
+
+  Scenario: VM at remote location is still running
+     When executing "minishift profile set remote" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: Deleting remote VM
```
budhram: What is the difference between this and `Deleting remote instance` ?

agajdosi: @budhram deleting remote instance deletes it via generic driver, which only deletes information about the remote VM from the profile, however the actual VM is still running, generic driver has no control over it. Here in `Deleting remote VM` scenario we changed to profile `remote` so we operate directly on the Minishift VM, this is normal delete via kvm/xhyve/vbox drivers. These few scenarios are here to also check that there are not any interferences between generic driver and VM created with --no-provision flag.

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,49 @@
...
+  Scenario: Stopping with generic driver
+  Stopping with generic driver puts oc cluster down, remote VM however keeps running.
+     When executing "minishift stop" succeeds
+      And executing "minishift status" succeeds
+     Then stdout should match "OpenShift:\s+Stopped"
+
+  Scenario: VM at remote location is still running
+     When executing "minishift status --profile remote" succeeds
+     Then stdout should match "Minishift:\s+Running"
+      And stdout should match "OpenShift:\s+Stopped"
```
amitkrout: Please add a step like
```
And executing “minishift ssh echo test” succeeds
```
This will ensure that VM is running and can also be accessible despite openshift cluster is down

praveenkumar: This is something we really don't care about when we do `minishift stop` on the remote. This scenario can be removed also.

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,49 @@
...
+  Scenario: Starting stopped cluster with generic driver
+     When executing "minishift start" succeeds
+     Then executing "minishift status" succeeds
+      And stdout should match "Minishift:\s+Running"
+      And stdout should match "OpenShift:\s+Running"
+  
+  Scenario: Remote VM is running and cluster is up
+     When executing "minishift status --profile remote" succeeds
+     Then stdout should match "Minishift:\s+Running"
+      And stdout should match "OpenShift:\s+Running"
```
praveenkumar: :+1:

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,49 @@
...
+
+  Scenario: Deleting remote instance
+     When executing "minishift delete --force" succeeds
+     Then Minishift should have state "Does Not Exist"
+
+  Scenario: VM at remote location is still running
+     When executing "minishift profile set remote" succeeds
+     Then Minishift should have state "Running"
```
praveenkumar: Again for us this scenario doesn't make much sense.

test/integration/features/generic-driver.feature
gbraad: we should verify the configuration on the reote host is gone

test/integration/features/generic-driver.feature
agajdosi: @gbraad I use it sometimes for debugging also, I appreciate this particular step a lot :)

gbraad: good to hear, as there was pushback at the time of implementation. As it does not really honor environment variables. But this is actually the point, as it should be helpful to verify external variables against those we gather in the scenario, and vice versa. but also, to re-use information such as IP address, names, etc

Perhaps there is more we can do with this?

test/integration/features/generic-driver.feature
agajdosi: @gbraad good point, can you please guide/give me to some information about the config? Did quick search in repo today but without results.

agajdosi:
@gbraad Requested changes implemented.
@praveenkumar I have kept the checks for status of non-provisioned machines as they provide some coverage Gerard mentioned and do not cost much time (matter of seconds).

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,60 @@
...
+  Scenario: User creates an unprovisioned VM
+    Given setting up environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" with value "y" succeeds
+     When executing "minishift profile set remote" succeeds
+      And executing "minishift start --no-provision" succeeds
+     Then Minishift should have state "Running"
+
+  Scenario: Provisioning remote instance
+    Given unset environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" succeeds
+      And setting scenario variable "REMOTE_VM_IP" to the stdout from executing "minishift ip"
+      And executing "minishift profile set minishift" succeeds
```
amitkrout: Add ```And image caching is disabled``` before m start

test/integration/features/generic-driver.feature
amitkrout: I am following exactly the scenario steps except I started the VM on virtual box. Then I am executing the steps no 29 with and without vm-driver flag

Just see the difference in the output log. I am using macOS
```
$ ./minishift start
-- Checking if requested hypervisor 'xhyve' is supported on this platform ... OK
-- Checking if xhyve driver is installed ... 
   Driver is available at /usr/local/bin/docker-machine-driver-xhyve
   Checking for setuid bit ... OK
-- Starting the OpenShift cluster using 'xhyve' hypervisor ...
-- Starting Minishift VM ........ OK
   Pinging 8.8.8.8 ... OK
-- Checking available disk space ... 10% used OK
   Version: v3.9.0
-- Copying oc binary from the OpenShift container image to VM ... OK
-- Starting OpenShift cluster .......
Using nsenter mounter for OpenShift volumes
Using public hostname IP 192.168.99.101 as the host IP
Using 192.168.99.101 as the server IP
Starting OpenShift using openshift/origin:v3.9.0 ...
OpenShift server started.

    https://192.168.99.101:8443
```
```
$ ./minishift start --vm-driver generic --remote-ipaddress 192.168.99.101 --remote-ssh-key /Users/amit/.minishift/profiles/remote/machines/remote/id_rsa --remote-ssh-user docker
-- Starting the OpenShift cluster using 'generic' hypervisor ...
-- Starting to provision the remote machine ........ OK
   Version: v3.9.0
-- Copying oc binary from the OpenShift container image to VM ... OK
-- Starting OpenShift cluster .......
Using nsenter mounter for OpenShift volumes
Using public hostname IP 192.168.99.101 as the host IP
Using 192.168.99.101 as the server IP
Starting OpenShift using openshift/origin:v3.9.0 ...
OpenShift server started.

    https://192.168.99.101:8443
```
@agajdosi I am sure you also will be seeing almost the same difference though you are provisioning the vm on default hypervisor. And more surprisingly it is starting the same openshift cluster. @praveenkumar it something we need to address? I feel this is killing the basic rules of minishift start

agajdosi: @amitkrout Which basic rules of Minishift start?

AFAIU this is expected. When --vm-driver is not specified, or is kvm, xhyve, virtualbox etc... then driver checks are executed but then IP address, username and ssh key saved in machines config is used and it all goes as if generic was selected, because the machine is generic.

When --vm-driver generic is specified, then driver checks are skipped as I believe they are skipped in code around generic. In this case I think it does not matter whether you add --remote-ip-address or anything else.

amitkrout: Let me put it in a general way

1. Setup virtual box on macOS or Linux so that user can switch between drivers
2. Run ```./minishift start —vm-driver virtualbox```
3. ```./minishift stop```
4. ```./mishift start```

This will run on default driver as no driver is added to the persistent config I mean to the config.json file. Ideally, this should fail if I am not wrong, however, it's not failing if I consider the same way as in case of generic driver. 

Please try out these steps so that we will be on the same page

test/integration/features/generic-driver.feature
agajdosi: @amitkrout fixed, thanks, I've missed that :)

agajdosi:
@minishift/minishift-dev please take a look when you have time

agajdosi:
@minishift/minishift-dev please take a look when you have time

test/integration/features/generic-driver.feature
```diff
@@ -0,0 +1,61 @@
+@generic-driver
+Feature: With generic driver Minishift can provision remote unprovisioned VM 
+
+  Scenario: User creates an unprovisioned VM
+    Given setting up environment variable "MINISHIFT_ENABLE_EXPERIMENTAL" with value "y" succeeds
+     When executing "minishift profile set remote" succeeds
+      And executing "minishift start --no-provision" succeeds
```
gbraad: so useful ;-)

LalatenduMohanty: @gbraad I think  it is time we make ` --no-provision` non experimental feature.

Issue #2545, opened 2018-07-02
Docker can't be reused after first start of Minishift


krisztiankocsis:
### General information

  * Minishift version: v1.20.0+53c500a
  * OS: macOS High Sierra 10.13.5
  * Hypervisor: VirtualBox 5.2.12
  * Docker: Docker For Mac App 18.03.1-ce Build 9ee9f40

### Steps to reproduce

  1. Install Minishift
  2. Start the instance with VirtualBox, CentOS image, OpenShift 3.7.2
  3. Setup Docker environment to be reused (copy certificates from Minishift to .docker/machine/certs)
  4. Login to Docker inside Minishift (succeed)
  5. Stop Minishift
  6. Start Minishift again
  7. Setup Docker environment to be reused (copy certificates again or not, doesn't make any difference)
  8. Login to Docker inside Minishift (fail with "x509: certificate signed by unknown authority")

### Expected

Able to login and use Docker inside Minishift.

### Actual

TLS verification fails, even with "--tlsverify=false".

### Logs

The full execution with output is available here:
https://gist.github.com/krisztiankocsis/983e6c31ca94e0be6d1576c3e65075d5

Note that the only change I saw is that Minishift always regenerates "server.pem" inside the $MINISHIFT_HOME/machines directory. I don't know whether it affects this or not.

praveenkumar:
@krisztiankocsis when say `Login to Docker inside Minishift (succeed)` you mean login to docker hub or any other registry?

krisztiankocsis:
No, login to OpenShift Docker Registry inside the Minishift VM. Please check the Gist, you will see the commands, in short:

eval $(minishift docker-env)
eval $(minishift oc-env)
docker login -u developer -p `oc whoami -t` docker-registry-default.192.168.99.100.nip.io

gbraad:
Did you `reboot` the VM of did you use `minishift stop; minishift start` ?

krisztiankocsis:
Hi, I’ve used stop and start.

Best regards / Mit freundlichen Grüßen / Üdvözlettel

Krisztian Kocsis
Chief Technology Officer

Optimaster s.r.o.
Druzstevna 8, 94501 Komarno, Slovakia(x-apple-data-detectors://1/1)  

Web: https://www.optimaster.eu  
Mail: krisztian.kocsis@optimaster.eu(mailto:krisztian.kocsis@optimaster.eu)
[...]

PR #2554, opened 2018-07-04
Issue #2545 Addon: make registry route non-fragile for multiple run

praveenkumar:


amitkrout:
From the console log what i can see the failure is happening immediately i.e. it does not sleep for a given time period for the next retry. The only reason i can see here is if the [function](https://github.com/minishift/minishift/blob/master/test/integration/testsuite/minishift.go#L91) returns null then it won't sleep for a given time period. So might be after minishift restart ```docker ps``` returns null. And one more possibility is after restart if image name is completely changed also causes the problem.

test/integration/features/addon-registry-route.feature
```diff
@@ -43,6 +43,20 @@ so that user can able to push their container images directly to registry and de
       And with up to "10" retries with wait period of "500ms" the "status code" of HTTP request to "/" of service "nginx" in namespace "myproject" is equal to "200"
       And with up to "10" retries with wait period of "500ms" the "body" of HTTP request to "/" of service "nginx" in namespace "myproject" contains "Welcome to nginx!"
 
+  Scenario: As a user, I can login to OpenShift integrated docker registry after restart minishift
+    Given Minishift has state "Running"
+     When executing "minishift stop" succeeds
+     Then executing "minishift start --iso-url centos" succeeds
+      And Minishift should have state "Running"
+      And executing "minishift docker-env" in host shell succeeds
```
amitkrout: Remove the duplicate lines from 51 to 54 as host shell is already configured in the previous scenario

test/integration/features/addon-registry-route.feature
```diff
@@ -43,6 +43,20 @@ so that user can able to push their container images directly to registry and de
       And with up to "10" retries with wait period of "500ms" the "status code" of HTTP request to "/" of service "nginx" in namespace "myproject" is equal to "200"
       And with up to "10" retries with wait period of "500ms" the "body" of HTTP request to "/" of service "nginx" in namespace "myproject" contains "Welcome to nginx!"
 
+  Scenario: As a user, I can login to OpenShift integrated docker registry after restart minishift
+    Given Minishift has state "Running"
+     When executing "minishift stop" succeeds
+     Then executing "minishift start --iso-url centos" succeeds
```
amitkrout: As communicated in irc please remove the flag ```--iso-url centos```. In line no 14 ```--iso-url centos``` has been unnoticing since we make centos as default iso. Please remove from both the places.

test/integration/features/addon-registry-route.feature
```diff
@@ -43,6 +43,17 @@ so that user can able to push their container images directly to registry and de
       And with up to "10" retries with wait period of "500ms" the "status code" of HTTP request to "/" of service "nginx" in namespace "myproject" is equal to "200"
       And with up to "10" retries with wait period of "500ms" the "body" of HTTP request to "/" of service "nginx" in namespace "myproject" contains "Welcome to nginx!"
 
+  Scenario: As a user, I can login to OpenShift integrated docker registry after restart minishift
+    Given Minishift has state "Running"
+     When executing "minishift stop" succeeds
+     Then executing "minishift start" succeeds
+      And Minishift should have state "Running"
+      And executing "minishift addon apply registry-route" succeeds
```
amitkrout: Why we are applying the addon again after restart ? AFAIU m stop then start won't affect already applied addon.

praveenkumar: @amitkrout yes it does in case of registry route

PR #2558, opened 2018-07-04
Issue #2028 Fix integration tests to check Openshift 

### LINKED ISSUE #2028: Step required in integration tests to check `OpenShift` status
We need a step in integration tests to check `OpenShift` status in Minishift. 
For example in below test in cmd-openshift.feature we should be using ` Given OpenShift has state "Running"` instead of ` Given Minishift has state "Running"`

```  Scenario: User deploys nodejs example application from OpenShift repository
      When executing "oc new-app https://github.com/openshift/nodejs-ex -l name=myapp" retrying 10 times with wait period of 2 seconds
      Then stdout should contain
       """
       Run 'oc status' to view your app.
       """
```

jemacom:
Closes #2028


test/integration/features/cmd-openshift.feature
```diff
@@ -48,7 +48,7 @@ cluster in VM provided by Minishift.
        And stdout should match "router-\d-\w{5}\s*1\/1\s*Running"
 
   Scenario: User deploys nodejs example application from OpenShift repository
-     Given Minishift has state "Running"
+     Given Openshift has state "Running"
```
budhram: @jemacom This is Minishift only. :)

jemacom: This test scenario requires Openshift (that resides inside minishift ) to be running. Because you can have Minishift running and Openshift stopped and no oc command will work.

gbraad: Minishift needs to be in running state, and with this OpenShift should be running

gbraad:
Also note, you have a merge; please make sure PRs are rebased onto master

gbraad:
In that cxase, this PR is still WIP as the issue has not been addressed as requested.

Issue #2557, opened 2018-07-04
Getting container status in the integration test is fragile


praveenkumar:
- https://github.com/minishift/minishift/blob/master/test/integration/testsuite/minishift.go#L88-L106 

we use as of now to get the status of a container which is fragile because here we check with what image name used to run the container and it is possible to be used the image-id instead name and run the container. The best way is if you already know a partial name of the container which you are interested in you can even use `docker ps -f "name=<name>"` and it will return the name if running.

agajdosi:
@praveenkumar Issue might not be in `docker ps` vs `docker ps -f name=<part-of-name>` itself, but might be in fact that docker command for some reason fails soon after the stop-start.

If I can ask you for something, with your PR please try to add a fmt.Println() here https://github.com/minishift/minishift/blob/c4d3e2535ee83eaa58b68936750448ef5f7b3def/test/integration/util/util.go#L251 to see what we get from `docker ps`, if that's empty, please try to add minute or more sleep right after the stop->start and see if that passes: `And user waits for "60" seconds`.

Error capabilities of functions around the retry for status of containers are just poor/none. I will improve them as part of this issue, but tomorrow and on Friday there are bank holidays, so I could just give above steps ^

amitkrout:
@praveenkumar  Curious to know why image name is changed after a restart, Is it the same container id origin-docker-registry before the restart?

praveenkumar:
@amitkrout Image name doesn't change, it just origin container thought I will just use the Image id instead name and both are same things, but as I told the current implementation of getting container status is fragile and should be improved, I am sending PR soon for this.

amitkrout:
Sure, please go ahead. I will have a look and understand the changes to clear my doubts.

LalatenduMohanty:
@praveenkumar  you mean parsing through `docker ps` output is not a good  way to get status, right?

```
func (m *MinishiftRunner) GetOpenshiftContainers() string {
	cmdOut, _, _ := m.RunCommand("ssh -- docker ps")
	return strings.Trim(cmdOut, " \n")
}
```

praveenkumar:
resolved by #2559

PR #2559, opened 2018-07-05
Issue #2557 Integration: Make container status function non-fragile

praveenkumar:


test/integration/util/util.go
```diff
@@ -247,8 +247,8 @@ func (m *MinishiftRunner) GetProfileList() string {
 	return strings.Trim(cmdOut, " \n")
 }
 
-func (m *MinishiftRunner) GetOpenshiftContainers() string {
-	cmdOut, _, _ := m.RunCommand("ssh -- docker ps")
+func (m *MinishiftRunner) GetOpenshiftContainers(containerName string) string {
+	cmdOut, _, _ := m.RunCommand(fmt.Sprintf(`ssh -- docker ps -f "name=%s"`, containerName))
```
LalatenduMohanty: 👍

test/integration/features/addon-registry-route.feature
```diff
@@ -24,7 +24,7 @@ so that user can able to push their container images directly to registry and de
       And evaluating stdout of the previous command in host shell
       And executing "minishift oc-env" in host shell succeeds
       And evaluating stdout of the previous command in host shell
-     Then with up to "20" retries with wait period of "3" seconds container image "origin-docker-registry" should be "running"
+     Then with up to "20" retries with wait period of "3" seconds container name "registry_docker-registry" should be "running"
```
LalatenduMohanty: The name `registry_docker-registry` looks odd.

praveenkumar: @LalatenduMohanty It is not odd, if you check the container name which openshift creates is something like `k8s_registry_docker-registry-1-vxt5j_default_a720081a-8012-11e8-984c-5254001ed95b_1` for the registry.

LalatenduMohanty: @praveenkumar Why not using the complete name `k8s_registry_docker-registry` ?

praveenkumar: @LalatenduMohanty Has is going to change always so you need to use filter and only use what you for sure not going to change, same trick we are using for origin container status for patch set.

LalatenduMohanty: @praveenkumar right thats why I was suggesting `k8s_registry_docker-registry`
```
$ docker ps -f name=k8s_registry_docker-registry
CONTAINER ID        IMAGE               COMMAND                  CREATED              STATUS              PORTS               NAMES
6bf509f27473        cf67b3158bfb        "/bin/sh -c '/usr/..."   About a minute ago   Up About a minute                       k8s_registry_docker-registry-1-vh5pq_default_dde2153b-8026-11e8-b089-f23a1db36513_0
```

Issue #2561, opened 2018-07-06
Error getting HostIP when used with proxy options [Windows + VirtualBox]


praveenkumar:
```

\minishift.exe start --vm-driver virtualbox --http-proxy http://10.76.99.2:3128 --show-libmachine-logs -v5
[...]
&{[-F /dev/null -o PasswordAuthentication=no -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null -o LogLevel=quiet -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=none docker@127.0.0.1 -o IdentitiesOnly=yes -i C:\Users\lmohanty\.minishift\machines\minishift\id_rsa -p 50045] C:\WINDOWS\System32\OpenSSH\ssh.exe <nil>}
About to run SSH command:
sudo ping -c1 -w1 192.168.99.1
.SSH cmd err, output: exit status 1: PING 192.168.99.1 (192.168.99.1) 56(84) bytes of data.
```

More Info: https://issues.jboss.org/browse/CDK-315

LalatenduMohanty:
Marking it blocker as it is a regression for Virtualbox + Windows setup on proxy.

LalatenduMohanty:
Resolved via #2562

PR #2562, opened 2018-07-06
Issue #2561 Remove IsIPReachable test from DetermineHostIP

praveenkumar:
- Currently with proxy it broken for windows+VB combination

### DIFF
--- a/pkg/minishift/network/utils.go
+++ b/pkg/minishift/network/utils.go
@@ -37,6 +37,11 @@ func GetIP(driver drivers.Driver) (string, error) {
 }
 
 func DetermineHostIP(driver drivers.Driver) (string, error) {
+	// In case of generic driver we don't need to get the bridge network
+	// since it's by default ping able from host IP.
+	if driver.DriverName() == "generic" {
+		return "", nil
+	}
 	instanceIP, err := driver.GetIP()
 	if err != nil {
 		return "", err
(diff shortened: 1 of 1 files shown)


pkg/minishift/network/utils.go
```diff
@@ -37,6 +37,11 @@ func GetIP(driver drivers.Driver) (string, error) {
 }
 
 func DetermineHostIP(driver drivers.Driver) (string, error) {
+	// In case of generic driver we don't need to get the bridge network
+	// since it's by default ping able from host IP.
+	if driver.DriverName() == "generic" {
+		return "", nil
```
gbraad: maybe return an Error here? they aren't in the same network ... This is a tricky one.

pkg/minishift/network/utils.go
```diff
@@ -46,10 +51,13 @@ func DetermineHostIP(driver drivers.Driver) (string, error) {
 
 		if NetworkContains(hostaddr, instanceIP) {
 			hostip, _, _ := net.ParseCIDR(hostaddr)
-			if IsIPReachable(driver, hostip.String(), false) {
+			// This step is not working with Windows + VirtualBox as of now
+			// This test is required for CIFS mount-folder case.
```
gbraad: CIFS will not be broken. It is part of a very limited part of the code in that case; host-based Users folder.

Issue #1360, opened 2017-09-11
 No value is specified in log level for V logs in "minishift start -h"


amitkrout:
Recently i came across this glogal flag ```-v, --v Level```. Mostly we uses a integer value for log level however the value is not specified in ``` minishift start -h```. 

For example:
```-v, --v Level          log level for V logs``` 

Even i am seeing more log level flags such as ```server-loglevel int```. This really improves the user experience if we add something like the default value or the range of values. This values will overwrite the default if passes along with the command. cc_@praveenkumar

LalatenduMohanty:
@amitkrout I think you are suggesting that we need to improve the usage e.g.
` -v, --v Level                          log level for V logs. Level  varies from 1 to 5 (default 1).` WDYT?

amitkrout:
@LalatenduMohanty Exactly

gbraad:
We do not have a real agreement on what these levels represent; we only know 1 is low, and 5 is so far the highest we specify...

blomquisg:
Looks like this change should be in golang/glog. I'll copy the issue up there.

PR #2571, opened 2018-07-08
Issue #1360 Improve help for log level for V logs

blomquisg:
Fixes Issue #1360

Previously:

> $ minishift start -h
> ...
>   -v, --v Level                          log level for V logs

Now:

> $ minishift start -h
> ...
>  -v, --v Level                          log level for V logs. Level varies from 1 to 5 (default 1).


gbraad:
We do not modify files in `/vendor` unless there is a valid reason we can not override, configure or fix it otherwise. Especially we would like to see an upstream issue in that case.

However, this can be solved on our end within `root.go`. WDYT?

blomquisg:
Heh, I should have seen `vendor` in the path!  :)

Might be better to just go after the change in glog and have it trickle down.  I'll report the issue in glog and reference the minishift issue as the basis.

Thanks for the review!

PR #2581, opened 2018-07-10
Issue #2435: Fix asciidoc error message

robin-owen:
Remove the `include` directive. This means we repeat ourselves, but in this instance, that should not be a problem.

**Note:** This is a temporary solution, but since I have not been able to determine the underlying cause, I'm okay with this for the time being.

### DIFF
--- a/docs/source/using/con_static-ip-overview.adoc
+++ b/docs/source/using/con_static-ip-overview.adoc
@@ -1,7 +1,0 @@
-Most hypervisors do not support extending the lease time when the IP is assigned using DHCP.
-This might lead to a new IP being assigned to the VM after a restart as it will conflict with the security certificates generated for the old IP.
-This will make {project} completely unusable until a new instance is set up by running `minishift delete` followed by `minishift start`.
-
-To prevent this, {project} includes the functionality to set a static IP address to the VM.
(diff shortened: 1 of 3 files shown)


robin-owen:
@LalatenduMohanty I've already attempted this using a different, simpler file name. When I triggered a CI build of the documentation using that file name, I noticed no change in the reported error. Considering it's very difficult to determine the underlying cause of the issue, this was the simplest temporary "fix" with reliable results.

While I don't like repeating ourselves in the documentation like this, I'm more concerned with not having an AsciiDoc error displayed in our online docs. :)

LalatenduMohanty:
> When I triggered a CI build of the documentation using that file name, I noticed no change in the reported error. Considering it's very difficult to determine the underlying cause of the issue, this was the simplest temporary "fix" with reliable results.

Fair enough. I was not aware of this.

robin-owen:
> Fair enough. I was not aware of this.

Apologies for that, I believe the discussion was taking place on IRC at the time and I neglected to update the issue with information about what I had attempted. :)

robin-owen:
@LalatenduMohanty Done -- for some reason, I thought I'd been using a colon before. :P

LalatenduMohanty:
@robin-owen Thanks

Issue #2592, opened 2018-07-12
Update all the CI to use golang-1.10


praveenkumar:
We need to update all our CI to use golang-1.10, `openshift/origin` migrated to 1.10 a few days back (5 July).

LalatenduMohanty:
Resolved via #2593

PR #2593, opened 2018-07-12
Issue #2592 CI: update go verion to 1.10.2

praveenkumar:


Makefile
```diff
@@ -229,10 +229,6 @@ integration_pr: $(MINISHIFT_BINARY)
 fmt: ## Format source using gofmt
 	@gofmt -l -s -w $(SOURCE_DIRS)
 
-.PHONY: vet
-vet: ## Vet the source using 'go vet'
-	@go vet $(shell $(PACKAGES))
-
```
praveenkumar: We don't require `vet` as a separate target now since with golang 1.10 it by default run as part of `go test` 

- https://groups.google.com/forum/#!topic/golang-dev/qfa3mHN4ZPA

```
$ go test -h
[..]
	-vet list
	    Configure the invocation of "go vet" during "go test"
	    to use the comma-separated list of vet checks.
	    If list is empty, "go test" runs "go vet" with a curated list of
	    checks believed to be always worth addressing.
	    If list is "off", "go test" does not run "go vet" at all.
```
```

File: centos_ci.sh [hunk 2]
```diff
@@ -56,6 +55,22 @@ function install_core_deps() {
   echo 'CICO: Core dependencies installed'
 }
 
+function install_golang() {
+  # Install virt7-container-common-candidate to install latest golang package
```
praveenkumar: This is the same repo all the container specific ecosystem use like podman, buildah ..etc.

praveenkumar:
@minishift/minishift-dev  anything stopping it to get merge?

Issue #2548, opened 2018-07-03
When generic driver is used, minishift should start the docker if it is not running


gbraad:
When remote deployment using the generic driver failed, it can get stuck on `sudo docker version`. A simple `sudo systemctl start docker` will resolve this, but is never executed.

LalatenduMohanty:
@gbraad You mean `minishift` should start teh docker daemon if not running.

gbraad:
This can happen easily when you want to deploy on a remote instance of CentOS, but you used `/etc/docker/daemon.json` to specify `insecure-registries`, our provisioner will fail... you fix the issue (by removing the file perhaps), but do not restart Docker.
When you redploy/start minishift again it will continually try to use `sudo docker version`, and fails after about 30 times this command (2+ minutes)
If you would do `systemctl start docker` and do a `minishift start` all is OK

praveenkumar:
@gbraad FYI if you use anything as part of `insecure-registries` to `/etc/docker/daemon.json` and then it is not going to consume it for sure because our current provisioner creates `/etc/systemd/system/docker.service.d/10-machine.conf` which is used for execution.

gbraad:
Note: the config actually conflicts and prevents to start

LalatenduMohanty:
Resolved via #2608

PR #2608, opened 2018-07-17
Issue #2548 Provisioner: Start docker daemon if not running by default

praveenkumar:
How to test
=========

```
$ minishift start --no-provision
$ minishift ip
$ minishift ssh -- sudo systemctl stop docker
$ minishift start --profile test --vm-driver generic --remote-ipaddress <IP> --remote-ssh-user docker  --remote-ssh-key <ssh_key>
```

### DIFF
--- a/pkg/minishift/provisioner/minishift_provisioner.go
+++ b/pkg/minishift/provisioner/minishift_provisioner.go
@@ -91,6 +91,13 @@ func (provisioner *MinishiftProvisioner) Package(name string, action pkgaction.P
 func (provisioner *MinishiftProvisioner) dockerDaemonResponding() bool {
 	log.Debug("checking docker daemon")
 
+	// Start the docker daemon service if not running.
+	if out, err := provisioner.SSHCommand("sudo systemctl -f start docker"); err != nil {
+		log.Warnf("Error getting SSH command to check if the daemon is running: %s", err)
+		log.Debugf("'sudo systemctl docker start' output:\n%s", out)
+		return false
+	}
+
 	if out, err := provisioner.SSHCommand("sudo docker version"); err != nil {
 		log.Warnf("Error getting SSH command to check if the daemon is running: %s", err)
 		log.Debugf("'sudo docker version' output:\n%s", out)


LalatenduMohanty:
code looks ok to me. Need to test it.

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -91,6 +91,12 @@ func (provisioner *MinishiftProvisioner) Package(name string, action pkgaction.P
 func (provisioner *MinishiftProvisioner) dockerDaemonResponding() bool {
 	log.Debug("checking docker daemon")
 
+	// Start the docker daemon service if not running.
+	if out, err := provisioner.SSHCommand("sudo systemctl -f start docker"); err != nil {
+		log.Warnf("Error getting SSH command to check if the daemon is running: %s", err)
+		log.Debugf("'sudo systemctl docker start' output:\n%s", out)
```
LalatenduMohanty: @praveenkumar Why we are not failing here. Because if we can not start docker then anyway it is going to fail.

gbraad: the `docker version` would fail anyway afterwards

praveenkumar: @gbraad @LalatenduMohanty yes I did it because it will anyway fail in the later stage but yes I can fail it early.

gbraad: f you look at the whole block of code, maybe returning the `err` would be a better idea anyway. as in that case the `logf` and `warnf` can be isolated to a single occurence in the caller (there), instead of the called function (here)

praveenkumar: @gbraad return err from here is not a good idea because then we are not going to use `mcnutils.WaitFor(provisioner.dockerDaemonResponding)` functionality which actually tries to get proper info after 30 attempts. I do agree can return `false` here so that it will not give user false impression of failing for version but instead it's failing due to service not running.

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -91,6 +91,13 @@ func (provisioner *MinishiftProvisioner) Package(name string, action pkgaction.P
...
+		return false
```
praveenkumar: @LalatenduMohanty @gbraad return added here so will give not give false version alarm.

Issue #2459, opened 2018-06-14
Add a script to perform prerequisites for remote machine in case of generic driver


praveenkumar:
As of now, we have prerequisites from remote machine side before provision it with minishift, such as install Docker, net-tools, and firewalld packages then open required port from firewall those can be easily automated either by ansible or by a script which can run before the provisioner even start.

gbraad:
There is a difference between 'checking pre-requisites' and 'installing pre-requisites'. We need to be careful about what we provision from minishift. As we need to instruct the user we will install packages on the remote machine. In some cases they might not want this. In that case, a check and a fail/warn would be more appropriate?

agajdosi:
@gbraad I am not sure if I got @praveenkumar's issue correctly, but I understand it that we could provide an Ansible playbook as part of Minishift, but it will be the user who will need to execute it. And plus Minishift should check for the packages and configuration needed in beginning of `m start --vm-driver generic`.

cmoulliard:
> who will need to execute it

Ansible playbook (packaged as a docker image) could also be executed using a k8s job included in an minishift addon ;-)

gbraad:
That is the point. We do need to check for pre-requisites, but when
mentioned to install this as part of the provisioner; we have a problem.

Pre-flight checks happen BEFORE the start or before the provisioning. So we
have the be clear what we need to do. I do not think a k8s job would help
here ;-), as this is before standing up OpenShift, but having the user do
it might not be the expected approach either.

praveenkumar:
> Ansible playbook (packaged as a docker image) could also be executed using a k8s job included in an minishift addon ;-)

@cmoulliard But that addon will only be executed once minishift up and running, In our case, this should be part of before even minishift start.

praveenkumar:
> There is a difference between 'checking pre-requisites' and 'installing pre-requisites'. We need to be careful about what we provision from minishift. As we need to instruct the user we will install packages on the remote machine. In some cases, they might not want this. In that case, a check and a fail/warn would be more appropriate?

@gbraad I agree but if a user really wants to provision single node cluster using minishift then those packages are must and should be installed, now we can tell users that these packages are going to install on that remote system and ask for confirmation (one way) or just do it. This is not done by provisioner part because provisioner (as of now) we are consuming is working for both our minishift iso (where those packages already part of our ISO) and Fedora/CentOS/RHEL side. What we can do in case of the generic driver before we hit the provisioner we should have some prerequisites code which should handle it.

[...]

PR #2613, opened 2018-07-18
Issue #2459 Generic: Add steps to pre-configure remote machine

praveenkumar:
Depends on : #2608

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -18,13 +18,12 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"runtime"
 	"strings"
 
-	"os"
-
 	"github.com/asaskevich/govalidator"
-	units "github.com/docker/go-units"
+	"github.com/docker/go-units"
 	"github.com/docker/machine/libmachine"
 	"github.com/docker/machine/libmachine/drivers"
 	"github.com/docker/machine/libmachine/host"
@@ -37,6 +36,7 @@ import (
 	"github.com/minishift/minishift/cmd/minishift/state"
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minikube/sshutil"
 	minishiftCluster "github.com/minishift/minishift/pkg/minishift/cluster"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
@@ -49,6 +49,7 @@ import (
 	openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"
 	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
 	"github.com/minishift/minishift/pkg/minishift/provisioner"
+	"github.com/minishift/minishift/pkg/minishift/remotehost"
 	"github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
 	"github.com/minishift/minishift/pkg/util/progressdots"
@@ -363,8 +364,6 @@ func determineInsecureRegistry(key string) []string {
 }
 
 func startHost(libMachineClient *libmachine.Client) *host.Host {
-	progressDots := progressdots.New()
-
 	// Configuration used for creation/setup of the Virtual Machine
 	machineConfig := &cluster.MachineConfig{
 		MinikubeISO:           determineIsoUrl(viper.GetString(configCmd.ISOUrl.Name)),
@@ -406,8 +405,21 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 
 		fmt.Print("-- Starting Minishift VM ...")
 	} else {
+		s, err := sshutil.NewRawSSHClient(machineConfig.RemoteIPAddress, machineConfig.SSHKeyToConnectRemote, machineConfig.RemoteSSHUser)
+		if err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating ssh client: %v", err))
+		}
+		fmt.Printf("-- Preparing Remote Machine ...")
+		progressDots := progressdots.New()
+		progressDots.Start()
+		if err := remotehost.PrepareRemoteMachine(s); err != nil {
+			atexit.ExitWithMessage(1, err.Error())
+		}
+		progressDots.Stop()
+		fmt.Println(" OK")
 		fmt.Print("-- Starting to provision the remote machine ...")
 	}
+	progressDots := progressdots.New()
 	progressDots.Start()
 	start := func() (err error) {
 		hostVm, err = cluster.StartHost(libMachineClient, *machineConfig)
--- a/docs/source/using/run-against-an-existing-machine.adoc
+++ b/docs/source/using/run-against-an-existing-machine.adoc
@@ -23,12 +23,16 @@ CentOS 7, Red Hat Enterprise Linux 7, and Fedora > 26 are the suggested Linux di
 
 To use an existing machine with {project}, it needs to be configured as follows:
 
-. Establish password-less SSH from the host to the existing remote machine:
-+
+Establish password-less SSH from the host to the existing remote machine:
 ----
 Host$ ssh-copy-id <user>@<remote_IP_address> # Ensure that the user has sudo access without a password or use root
 Host$ ssh <user>@<remote_IP_address>         # Ensure that this login works without a password
 ----
+
+[NOTE]
+====
+Skip following steps if you are using suggested Linux distribution (ie. CentOS 7, Red Hat Enterprise Linux 7, and Fedora > 26).
+====
 
 . Configure the existing remote machine:
 +
--- a/pkg/minikube/sshutil/sshutil.go
+++ b/pkg/minikube/sshutil/sshutil.go
@@ -63,6 +63,29 @@ func NewSSHClient(d drivers.Driver) (*ssh.Client, error) {
 	return client, nil
 }
 
+func NewRawSSHClient(ip, sshKeyPath, username string) (*ssh.Client, error) {
+	h := &sshHost{
+		IP:         ip,
+		Port:       22,
+		SSHKeyPath: sshKeyPath,
+		Username:   username,
+	}
+	auth := &machinessh.Auth{}
+	if h.SSHKeyPath != "" {
+		auth.Keys = []string{h.SSHKeyPath}
+	}
+	config, err := machinessh.NewNativeConfig(h.Username, auth)
+	if err != nil {
+		return nil, err
+	}
+
+	client, err := ssh.Dial("tcp", net.JoinHostPort(h.IP, strconv.Itoa(h.Port)), &config)
+	if err != nil {
+		return nil, err
+	}
+	return client, nil
+}
+
 func TransferFile(f assets.CopyableFile, client *ssh.Client) error {
 	return Transfer(f, f.GetLength(),
 		f.GetTargetDir(), f.GetTargetName(),
--- a/pkg/minishift/remotehost/remotehost.go
+++ b/pkg/minishift/remotehost/remotehost.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2018 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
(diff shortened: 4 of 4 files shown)


praveenkumar:
retest this please.

praveenkumar:
I think test failure for this is because we are trying to install firewalld package on the default ISO which prevents it from doing so for our custom ISO. will investigate on this.

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -88,6 +88,41 @@ func (provisioner *MinishiftProvisioner) Package(name string, action pkgaction.P
 	return errors.New("Custom provisioner does not support package")
 }
 
+func (provisioner *MinishiftProvisioner) remoteHostPrepration() error {
```
LalatenduMohanty: s/remoteHostPrepration/PreparingRemoteHost/

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -107,6 +142,13 @@ func (provisioner *MinishiftProvisioner) Provision(swarmOptions swarm.Options, a
 	provisioner.EngineOptions = engineOptions
 	swarmOptions.Env = engineOptions.Env
 
+	// Pre-configure the remote machine if generic driver in use
+	if provisioner.Driver.DriverName() == "generic" {
+		if err := provisioner.remoteHostPrepration(); err != nil {
+			return err
+		}
+	}
```
LalatenduMohanty: 👍

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -88,6 +88,41 @@ func (provisioner *MinishiftProvisioner) Package(name string, action pkgaction.P
...
 
+func (provisioner *MinishiftProvisioner) remoteHostPrepration() error {
+	packageList := map[string]string{
+		"firewalld": "firewall-cmd",
+		"docker":    "docker",
+		"net-tools": "netstat"}
+
+	// Install required packages if not present.
+	for pkg, cmd := range packageList {
+		if out, err := provisioner.SSHCommand(fmt.Sprintf("which %s || sudo yum install -y %s", cmd, pkg)); err != nil {
```
LalatenduMohanty: Is this fine for Fedora? as we use the same provisioner for Fedora also, right?

gbraad: `yum` is still a fallback option

pkg/minishift/provisioner/minishift_provisioner.go
gbraad: Preparng would also be wrong. `prepareRemoteHost` because `-ing` would suggest that the preparation step is only setup to be finished AFTER running an additional step after this function.

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -88,6 +88,41 @@ func (provisioner *MinishiftProvisioner) Package(name string, action pkgaction.P
...
+		}
+	}
+
+	// 	Start the firewalld service.
+	if _, err := provisioner.SSHCommand("sudo systemctl start firewalld"); err != nil {
+		return fmt.Errorf("Error starting firewalld service: %s", err)
+	}
+
+	firewallCommandsToExecute := []string{
+		"sudo firewall-cmd --permanent --add-port 2376/tcp --add-port 8443/tcp --add-port 80/tcp",
```
gbraad: not so happy with all these commands being everywhere in the file... makes maintenance very hard

pkg/minishift/provisioner/minishift_provisioner.go
```diff
@@ -107,6 +142,13 @@ func (provisioner *MinishiftProvisioner) Provision(swarmOptions swarm.Options, a
 	provisioner.EngineOptions = engineOptions
 	swarmOptions.Env = engineOptions.Env
 
+	// Pre-configure the remote machine if generic driver in use
+	if provisioner.Driver.DriverName() == "generic" {
```
gbraad: Is there a better way to collect information about the driver and OS: https://github.com/praveenkumar/minishift/blob/6ecc79dc2bbd529dbc3666e42a36e5327201336c/pkg/minishift/provisioner/minishift_provisioner.go#L204-L208 in a better way?

We are now adding `generic` or other words VM-driver specific code to our provisioner. Is there not a better way?

praveenkumar:
>  Is not better to split this out in a generic provisioner to separate this better? I dislike the mixing of responsibilities.

provisioner is separated according to the operating system, not from driver specific. Right now we are mixing because of using the same provisioner for Fedora/RHEL/CentOS along with our own ISO or with the remote system.

What we can do is make this prerequisite not the part of provisioner but stand alone but then we need to implement our own way to execute ssh commands (right now we are using the driver ssh commander)

praveenkumar:
@gbraad @LalatenduMohanty I made some changes now so that remote machine preparation happens before provisioner hit.

cmd/minishift/cmd/start.go
```diff
@@ -37,12 +36,14 @@ import (
...
 	"github.com/minishift/minishift/pkg/minikube/cluster"
 	"github.com/minishift/minishift/pkg/minikube/constants"
+	"github.com/minishift/minishift/pkg/minikube/sshutil"
 	minishiftCluster "github.com/minishift/minishift/pkg/minishift/cluster"
 	"github.com/minishift/minishift/pkg/minishift/clusterup"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
 	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
 	"github.com/minishift/minishift/pkg/minishift/docker"
 	"github.com/minishift/minishift/pkg/minishift/docker/image"
+	"github.com/minishift/minishift/pkg/minishift/generic"
```
gbraad: can this package name be `provision` ?

cmd/minishift/cmd/start.go
```diff
@@ -406,8 +405,21 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
...
 		fmt.Print("-- Starting Minishift VM ...")
 	} else {
+		s, err := sshutil.NewRawSSHClient(machineConfig.RemoteIPAddress, machineConfig.SSHKeyToConnectRemote, machineConfig.RemoteSSHUser)
+		if err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating ssh client: %v", err))
+		}
+		fmt.Printf("-- Preparing Remote Machine ...")
+		progressDots := progressdots.New()
+		progressDots.Start()
+		if err := generic.PrepareRemoteMachine(s); err != nil {
```
gbraad: ... changing the package name makes more sense, as the responsibility is clear. Besides, 'generic' is an awkward name to begin with as we mean `remote` everywhere else anyway.

pkg/minikube/sshutil/sshutil.go
```diff
@@ -63,6 +63,29 @@ func NewSSHClient(d drivers.Driver) (*ssh.Client, error) {
 	return client, nil
 }
 
+func NewRawSSHClient(ip, sshKeyPath, username string) (*ssh.Client, error) {
```
gbraad: why do we need this?
It is an SSH client without the driver involved, right?

cmd/minishift/cmd/start.go
praveenkumar: @gbraad instead of `provision` what about `remote` ?

pkg/minikube/sshutil/sshutil.go
praveenkumar: > It is an SSH client without the driver involved, right?

yes, that's right.

praveenkumar:
retest this please.

gbraad:
>
>  @gbraad <https://github.com/gbraad> instead of provision what about
> remote ?
>
Remote means nothing from the perspective of our code. Everything is remote
... This code deals with machine setup. What do the others think?

>

cmd/minishift/cmd/start.go
LalatenduMohanty: I am not sure if putting `remote` everywhere is a good idea because someone might try to use his local machine to start OpenShift.  I also I agree that `generic` is also not a great word as it can confuse people.

LalatenduMohanty: @praveenkumar @gbraad  I could not find any better name. lets go with `remotehost`.

LalatenduMohanty:
I tested the PR and it works fine for me.

cmd/minishift/cmd/start.go
```diff
@@ -406,8 +405,21 @@ func startHost(libMachineClient *libmachine.Client) *host.Host {
 
 		fmt.Print("-- Starting Minishift VM ...")
 	} else {
+		s, err := sshutil.NewRawSSHClient(machineConfig.RemoteIPAddress, machineConfig.SSHKeyToConnectRemote, machineConfig.RemoteSSHUser)
+		if err != nil {
+			atexit.ExitWithMessage(1, fmt.Sprintf("Error creating ssh client: %v", err))
+		}
+		fmt.Printf("-- Preparing Remote Machine ...")
```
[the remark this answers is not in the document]
LalatenduMohanty: This is not an issue. Ignore this comment.

pkg/minishift/generic/generic.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return err
+	}
+	osReleaseInfo, err := provision.NewOsRelease([]byte(osReleaseOut))
+	if err != nil {
+		return err
+	}
+	if osReleaseInfo.ID == "fedora" || osReleaseInfo.ID == "rhel" || osReleaseInfo.ID == "centos" {
+		if err := prepareRHELVariant(s); err != nil {
+			return err
```
LalatenduMohanty: We are calling `PrepareRemoteMachine` before the generic driver kicks in. So if anyone want to add similar support of other operating systems they might need to add code to `PrepareRemoteMachine`. Though it depends on what libmachine has for that os. IMO we should implement an interface here and each OS should have `setupHostForGenericDriver` (or something similar) function which should have actual automation.  WDYT?

praveenkumar: >Though it depends on what libmachine has for that os. IMO we should implement an interface here and each OS should have setupHostForGenericDriver (or something similar) function which should have actual automation. WDYT?

@LalatenduMohanty that is already part of each provisioner here we are explicit doing because libmachine provisioner for RHEL/CentOS and Fedora is outdated once we start consuming the machine-driver/libmachine than we might not need it as this can become part of standard provisioner. I don't want to boated lot of stuff which we may need in future. (If we need we can always refactor). Remember we invested lot of time for registration thinking that might be Suse can implement their own way of subscription but here we are code is there but no outside contribution. Everything shouldn't be solved by interface (it good concept but using it to predict that something can be used in future doesn't make sense to me).

praveenkumar:
> Minishift takes care of installing required packages and starting required services in the host if it is RHEL, CentOS or Fedora.

Yes, that I will take care to update the document.

pkg/minishift/generic/generic.go
LalatenduMohanty: @praveenkumar  Having a flexible code always helps in long term. If you want to do this step by step I am ok with talking this change and we can refactor later.

gbraad: If you look for flexible code, we need a pre-provision `add-on-like` executioner. This will run to deploy packages or setups from files. This is what I see as a refactor...

praveenkumar: @gbraad when you say `add-on-like` executioner you mean have a plain file which user can specify and contain the commands which should run before hand?

gbraad: yep... not Ansible, and not a real sh-bang (#!/bin/sh) shell file...

A file that can be specified to run on a particular target based on meta-data perhaps. But this is food for thought, not to do now...

docs/source/using/run-against-an-existing-machine.adoc
```diff
@@ -23,13 +23,17 @@ CentOS 7, Red Hat Enterprise Linux 7, and Fedora > 26 are the suggested Linux di
...
-+
+Establish password-less SSH from the host to the existing remote machine:
 ----
 Host$ ssh-copy-id <user>@<remote_IP_address> # Ensure that the user has sudo access without a password or use root
 Host$ ssh <user>@<remote_IP_address>         # Ensure that this login works without a password
 ----
 
+[NOTE]
+====
+Skip following steps if you are using suggested Linux distribution (ie. CentOS 7, Red Hat Enterprise Linux 7, and Fedora > 26).
```
LalatenduMohanty: Skip the following steps if you are using entOS 7, Red Hat Enterprise Linux 7, and Fedora > 26.

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/commit/20dd3b33c6e631a5b8f396611ab61aadf775ac8c

Issue #2604, opened 2018-07-15
Applying admin-user addon with --openshift-version 3.10.0-rc.0 results in 'No such file or directory'


donovanmuller:
### General information

  * Minishift version: 

```
$ minishift version
minishift v1.21.0+a8c8b37
```

  * OS: Linux / macOS / Windows

macOS

  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

xhyve

### Steps to reproduce

  1. Start a `minishift` instance with `--openshift-version 3.10.0-rc.0`
  2. Wait for successful start
  3. Try to apply the `admin-user` addon with `minishift addons apply admin-user`

### Expected

`admin-user` addon to be applied successfully

### Actual

```
$ minishift addons apply admin-user
Cannot get the OpenShift master configuration: ssh command error:
command : docker exec -t 282cd17f45d5 cat /var/lib/origin/openshift.local.config/master/master-config.yaml
err     : exit status 1
output  : cat: /var/lib/origin/openshift.local.config/master/master-config.yaml: No such file or directory
```

praveenkumar:
@donovanmuller I tried as following and it works as expected.

```
$ ./minishift start --openshift-version 3.10.0-rc.0
-- Starting profile 'minishift'
-- Check if deprecated options are used ... OK
-- Checking if https://github.com is reachable ... OK
[...]
Login to server ...
Creating initial project "myproject" ...
Server Information ...
OpenShift server started.

The server is accessible via web console at:
    https://192.168.64.3:8443

You are logged in as:
    User:     developer
    Password: <any value>

To login as administrator:
    oc login -u system:admin

$  $ ./minishift status
Minishift:  Running
Profile:    minishift
OpenShift:  Running (openshift v3.10.0-rc.0+745de5b-46)
DiskUsage:  21% of 19G (Mounted On: /mnt/vda1)
CacheUsage: 2.109 GB (used by oc binary, ISO or cached images)

$ ./minishift addon apply admin-user
-- Applying addon 'admin-user':..

 $ oc whoami
developer

10:43 $ oc login -u admin -p admin
Login successful.

You have access to the following projects and can switch between them with 'oc project <projectname>':

    default
    kube-dns
    kube-proxy
    kube-public
    kube-system
  * myproject
    openshift
    openshift-apiserver
    openshift-controller-manager
    openshift-core-operators
    openshift-infra
    openshift-node
    openshift-web-console

Using project "myproject".

10:43 $ oc whoami
admin
[...]

PR #2627, opened 2018-07-24
Issue #2604 Fix addon apply for different profiles

praveenkumar:
- In case of openshift version >= 3.10.x look up path is different.


praveenkumar:
> Ah, no I see what happened. The values were actually static, as they never changed. When the appliction started, they got their value, but not based on the expectation, as this happened later.

Right, our mistake :(

gbraad:
>
>
> Right, our mistake :(
>

How can we deal with these edge cases; I guess addon tests with profiles
involved?

>

PR #2628, opened 2018-07-24
Issue #2624 Support 'latest' tag in openshift-version flag

### LINKED ISSUE #2624: Support "latest" tag in openshift-version flag
The start flag `--openshift-version` should support `latest` tag. This should be the latest tag from the version list `minishift openshift version list`.

budhram:
Fix #2624 
```
$ minishift start --openshift-version latest
-- Checking if requested OpenShift version 'v3.10.0-rc.0' is valid ... OK
-- Checking if requested OpenShift version 'v3.10.0-rc.0' is supported ... OK
[...]
   Version: v3.10.0-rc.0
-- Pulling the Openshift Container Image ................ OK
-- Copying oc binary from the OpenShift container image to VM ... OK
-- Starting OpenShift cluster .........................................
[...]
```

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -177,13 +177,18 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	proxyConfig := handleProxies()
 
+	// Get proper OpenShift version
+	requestedOpenShiftVersion, err := cmdUtil.GetOpenShiftReleaseVersion()
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting OpenShift version: %v", err))
+	}
+
 	// preflight check (before start)
 	if viper.GetString(configCmd.VmDriver.Name) != genericDriver {
 		preflightChecksBeforeStartingHost()
 	}
 
 	// Cache OC binary before starting the VM and perform oc command option check
-	requestedOpenShiftVersion := viper.GetString(configCmd.OpenshiftVersion.Name)
 	ocPath = cmdUtil.CacheOc(requestedOpenShiftVersion)
 	preflightChecksForArtifacts()
 
@@ -220,7 +225,7 @@ func runStart(cmd *cobra.Command, args []string) {
 
 	applyDockerEnvToProcessEnv(libMachineClient)
 
-	err := clusterup.EnsureHostDirectoriesExist(hostVm, getRequiredHostDirectories())
+	err = clusterup.EnsureHostDirectoriesExist(hostVm, getRequiredHostDirectories())
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error creating required host directories: %v", err))
 	}
@@ -626,7 +631,7 @@ func initStartFlags() *flag.FlagSet {
 	startFlagSet.String(configCmd.DiskSize.Name, constants.DefaultDiskSize, "Disk size to allocate to the Minishift VM. Use the format <size><unit>, where unit = MB or GB.")
 	startFlagSet.String(configCmd.HostOnlyCIDR.Name, "192.168.99.1/24", "The CIDR to be used for the minishift VM. (Only supported with VirtualBox driver.)")
 	startFlagSet.Bool(configCmd.SkipPreflightChecks.Name, false, "Skip the startup checks.")
-	startFlagSet.String(configCmd.OpenshiftVersion.Name, version.GetOpenShiftVersion(), fmt.Sprintf("The OpenShift version to run, eg. %s", version.GetOpenShiftVersion()))
+	startFlagSet.String(configCmd.OpenshiftVersion.Name, version.GetOpenShiftVersion(), fmt.Sprintf("The OpenShift version to run, eg. latest or %s", version.GetOpenShiftVersion()))
 
 	startFlagSet.String(configCmd.RemoteIPAddress.Name, "", "IP address of the remote machine to provision OpenShift on")
 	startFlagSet.String(configCmd.RemoteSSHUser.Name, "", "The username of the remote machine to provision OpenShift on")
--- a/cmd/minishift/cmd/start_preflight.go
+++ b/cmd/minishift/cmd/start_preflight.go
@@ -65,30 +65,32 @@ func preflightChecksBeforeStartingHost() {
 		configCmd.WarnDeprecationCheck.Name,
 		"")
 
-	// Conectivity logic
+	requestedOpenShiftVersion, _ := cmdUtil.GetOpenShiftReleaseVersion()
+
+	// Connectivity logic
 	fmt.Printf("-- Checking if %s is reachable ... ", GithubAddress)
 	if network.CheckInternetConnectivity(GithubAddress) {
 		fmt.Printf("OK\n")
 		preflightCheckSucceedsOrFails(
 			configCmd.SkipCheckOpenShiftRelease.Name,
 			checkOriginRelease,
-			fmt.Sprintf("Checking if requested OpenShift version '%s' is valid", viper.GetString(configCmd.OpenshiftVersion.Name)),
+			fmt.Sprintf("Checking if requested OpenShift version '%s' is valid", requestedOpenShiftVersion),
 			configCmd.WarnCheckOpenShiftRelease.Name,
 			"",
 		)
 	} else {
 		fmt.Printf("FAIL\n")
-		fmt.Printf("-- Checking if requested OpenShift version '%s' is valid ... SKIP\n", viper.GetString(configCmd.OpenshiftVersion.Name))
+		fmt.Printf("-- Checking if requested OpenShift version '%s' is valid ... SKIP\n", requestedOpenShiftVersion)
 	}
 	// end of connectivity logic
 
 	preflightCheckSucceedsOrFails(
 		configCmd.SkipCheckOpenShiftVersion.Name,
 		validateOpenshiftVersion,
-		fmt.Sprintf("Checking if requested OpenShift version '%s' is supported", viper.GetString(configCmd.OpenshiftVersion.Name)),
+		fmt.Sprintf("Checking if requested OpenShift version '%s' is supported", requestedOpenShiftVersion),
 		configCmd.WarnCheckOpenShiftVersion.Name,
 		fmt.Sprintf("Minishift does not support OpenShift version %s. "+
-			"You need to use a version >= %s\n", viper.GetString(configCmd.OpenshiftVersion.Name),
+			"You need to use a version >= %s\n", requestedOpenShiftVersion,
 			constants.MinimumSupportedOpenShiftVersion),
 	)
 
@@ -584,9 +586,8 @@ func checkVMDriver() bool {
 }
 
 func validateOpenshiftVersion() bool {
-	requestedVersion := viper.GetString(configCmd.OpenshiftVersion.Name)
-
-	valid, err := openshiftVersion.IsGreaterOrEqualToBaseVersion(requestedVersion, constants.MinimumSupportedOpenShiftVersion)
+	requestedOpenShiftVersion, _ := cmdUtil.GetOpenShiftReleaseVersion()
+	valid, err := openshiftVersion.IsGreaterOrEqualToBaseVersion(requestedOpenShiftVersion, constants.MinimumSupportedOpenShiftVersion)
 	if err != nil {
 		return false
 	}
@@ -601,14 +602,16 @@ func validateOpenshiftVersion() bool {
 func checkOriginRelease() bool {
 	client := github.Client()
 	ctx := context.Background()
-	_, _, err := client.Repositories.GetReleaseByTag(ctx, "openshift", "origin", viper.GetString(configCmd.OpenshiftVersion.Name))
+
+	requestedOpenShiftVersion, _ := cmdUtil.GetOpenShiftReleaseVersion()
+	_, _, err := client.Repositories.GetReleaseByTag(ctx, "openshift", "origin", requestedOpenShiftVersion)
 	if err != nil && github.IsRateLimitError(err) {
 		fmt.Println("\n   Hit github rate limit:", err)
 		return false
 	}
 
 	if err != nil {
-		fmt.Printf("%s is not a valid OpenShift version", viper.GetString(configCmd.OpenshiftVersion.Name))
+		fmt.Printf("%s is not a valid OpenShift version", requestedOpenShiftVersion)
 		return false
 	}
 	return true
--- a/cmd/minishift/cmd/util/util.go
+++ b/cmd/minishift/cmd/util/util.go
@@ -35,7 +35,10 @@ import (
 	"github.com/minishift/minishift/out/bindata"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
+	openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"
 	"github.com/minishift/minishift/pkg/util/shell"
+	"github.com/minishift/minishift/pkg/version"
+	"github.com/spf13/viper"
 )
 
 var (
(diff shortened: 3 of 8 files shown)


test/integration/features/provision-various-versions.feature
```diff
@@ -32,3 +32,11 @@ Feature: Provision all major OpenShift versions
     | v3.7.0        | v3.7.0    |
     | v3.7.1        | v3.7.1    |
     | v3.7.2        | v3.7.2    |
+
+  Scenario Outline: Provision latest OpenShift version
```
amitkrout: +1

budhram:
add to whitelist

budhram:
retest this please

cmd/minishift/cmd/start.go
```diff
@@ -181,8 +181,13 @@ func runStart(cmd *cobra.Command, args []string) {
 		preflightChecksBeforeStartingHost()
 	}
 
+	// Get proper OpenShift version
+	requestedOpenShiftVersion, err := getOpenShiftVersion()
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting OpenShift version: %v", err))
+	}
+
```
praveenkumar: why not put this before L#179 so that you don't need to validate it again for the preflight check?

cmd/minishift/cmd/start_preflight.go
```diff
@@ -65,30 +69,36 @@ func preflightChecksBeforeStartingHost() {
 		configCmd.WarnDeprecationCheck.Name,
 		"")
 
-	// Conectivity logic
+	// Get OpenShift version
+	requestedOpenShiftVersion, err := getOpenShiftVersion()
```
praveenkumar: here you can ignore the err since it already handled in the start if you put that block as I suggested.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -584,9 +594,12 @@ func checkVMDriver() bool {
 }
 
 func validateOpenshiftVersion() bool {
-	requestedVersion := viper.GetString(configCmd.OpenshiftVersion.Name)
+	requestedOpenShiftVersion, err := getOpenShiftVersion()
+	if err != nil {
+		atexit.ExitWithMessage(1, fmt.Sprintf("Error getting OpenShift version: %v", err))
+	}
```
praveenkumar: why don't change the function signature so it will take version also, now you don't need to explicitly call the `getOpenshiftVersion()` multiple places and handle the error?

cmd/minishift/cmd/start_preflight.go
```diff
@@ -601,14 +614,20 @@ func validateOpenshiftVersion() bool {
 func checkOriginRelease() bool {
 	client := github.Client()
 	ctx := context.Background()
-	_, _, err := client.Repositories.GetReleaseByTag(ctx, "openshift", "origin", viper.GetString(configCmd.OpenshiftVersion.Name))
+
+	requestedOpenShiftVersion, err := getOpenShiftVersion()
```
praveenkumar: same here better to change the function signature `checkOriginRelease(openshiftVersion string) bool`

budhram: @praveenkumar `checkOriginRelease()` is a callback function. 
See https://github.com/minishift/minishift/blob/90fa2ae102dbc624acbb31870d3a2e83aca98461/cmd/minishift/cmd/start_preflight.go#L74.

Can't call with arguments.

cmd/minishift/cmd/start_preflight.go
budhram: @praveenkumar What is the expected signature here?

budhram:
retest this please

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -153,3 +142,47 @@ func getGithubReleases() ([]string, error) {
...
+	var releaseList []string
+
+	data, err := getGithubReleases()
+	if err != nil {
+		return []string{}, err
+	}
+
+	for _, releaseTag := range data {
+		if valid, _ := IsGreaterOrEqualToBaseVersion(releaseTag, minSupportedVersion); valid {
+			if valid, _ := IsGreaterOrEqualToBaseVersion(releaseTag, defaultVersion); valid {
```
agajdosi: @budhram Just question, how it comes here that v3.11.0-alpha.0 and v3.8.0 are not appended?

agajdosi: Aaah, GetGithubReleases() does not return those, need to check that.

cmd/minishift/cmd/start.go
budhram: Done :)

pkg/minishift/openshift/version/openshift_versions.go
budhram: Ya @agajdosi and `v3.11.0-alpha.0` is not a proper release with its artifacts. See https://github.com/openshift/origin/releases

agajdosi: @budhram Haha! Completely missed that, was still searching where do we have an exception for those versions, thanks, it is clear now!

praveenkumar:
add to whitelist.

cmd/minishift/cmd/start.go
```diff
@@ -779,3 +784,17 @@ func cacheMinishiftISO(config *cluster.MachineConfig) {
 func shouldPreflightChecksBeSkipped() bool {
 	return viper.GetBool(configCmd.SkipPreflightChecks.Name)
 }
+
+func getOpenShiftVersion() (string, error) {
+	tag := viper.GetString(configCmd.OpenshiftVersion.Name)
+	if tag == "vlatest" {
+		tags, err := openshiftVersion.GetSortedOpenShiftTags(constants.MinimumSupportedOpenShiftVersion, version.GetOpenShiftVersion())
+		if err != nil {
+			return "", nil
```
praveenkumar: This is wrong, you are supposed to return err here but returning nil that means this function never return the err now.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -50,6 +50,10 @@ const (
 	hypervDefaultVirtualSwitchId = "c08cb7b8-9b3c-408e-8e30-5e16a3aeb444"
 )
 
+var (
+	requestedOpenShiftVersion string
+)
+
```
praveenkumar: Why this is now required since now you are already evaluating it for each function?

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -153,3 +142,47 @@ func getGithubReleases() ([]string, error) {
...
+				releaseList = append(releaseList, releaseTag)
+			} else {
+				if !isPrerelease(releaseTag) {
+					releaseList = append(releaseList, releaseTag)
+				}
+			}
+		}
+	}
+
```
praveenkumar: I know that you extracted this because you need the release list to map the latest version but any specific reason for this `sortToSemVerTags` does that was not working before as it should be? Also you need to extract that for downstream also since the code is part of upstream only.

cmd/minishift/cmd/start.go
budhram: Ya missed that thanks

budhram: Done

gbraad: still seeing `return "", nil` ?

cmd/minishift/cmd/start_preflight.go
budhram: Leftover of previous implementation :wink:

pkg/minishift/openshift/version/openshift_versions.go
budhram: @praveenkumar String sort is not working and hence need to do `semver.Sort(versionTags)`.
When I do string sort, I am getting `v3.9.0` at the bottom instead of `v3.10.0-rc.1`.
Hence, I am converting string versions into semver tag and returning the string format only.
Hope it will help.

budhram:
The same logic of sorting the versions are working fine in downstream for its currently available versions list.
IMO, this PR is ready to merge on its own self.

/cc @minishift/minishift-dev

praveenkumar:
@budhram https://github.com/minishift/minishift/blob/master/pkg/minishift/openshift/version/openshift_versions.go#L44-L74 also need to change to adapt refactor you did for getting version info.

LalatenduMohanty:
@budhram we need to update the usage of `$ minishift start --help` for `--openshift-version string         The OpenShift version to run, eg. v3.9.0 (default "v3.9.0")` . We should mention `--openshift-version string         The OpenShift version to run, eg. latest or v3.9.0 (default "v3.9.0") `

cmd/minishift/cmd/start.go
```diff
@@ -779,3 +784,17 @@ func cacheMinishiftISO(config *cluster.MachineConfig) {
 func shouldPreflightChecksBeSkipped() bool {
 	return viper.GetBool(configCmd.SkipPreflightChecks.Name)
 }
+
+func getOpenShiftVersion() (string, error) {
+	tag := viper.GetString(configCmd.OpenshiftVersion.Name)
+	if tag == "vlatest" {
+		tags, err := openshiftVersion.GetSortedOpenShiftTags(constants.MinimumSupportedOpenShiftVersion, version.GetOpenShiftVersion())
```
gbraad: I always dislike names that start with `Sorted` as it means nothing without more context. Which way is it sorted, ascending or descending? if an option would have been added, this is clear... or rename the function to include the actual sort order. `GetOpenShiftVersionByTagsAscending` or similar.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -65,30 +65,33 @@ func preflightChecksBeforeStartingHost() {
 		configCmd.WarnDeprecationCheck.Name,
 		"")
 
-	// Conectivity logic
+	// Get OpenShift version
```
gbraad: does this need comments?

cmd/minishift/cmd/start_preflight.go
```diff
@@ -584,9 +587,8 @@ func checkVMDriver() bool {
 }
 
 func validateOpenshiftVersion() bool {
```
gbraad: maybe provide the version as a parameter, as it is used here as a non-reussable static function. This means we call `getOpenShiftVersion` again. If we provide this as a parameter we put that responsibility somewhere else.

cmd/minishift/cmd/start_preflight.go
```diff
@@ -601,14 +603,16 @@ func validateOpenshiftVersion() bool {
 func checkOriginRelease() bool {
 	client := github.Client()
 	ctx := context.Background()
-	_, _, err := client.Repositories.GetReleaseByTag(ctx, "openshift", "origin", viper.GetString(configCmd.OpenshiftVersion.Name))
+
+	requestedOpenShiftVersion, _ := getOpenShiftVersion()
```
gbraad: ditto... ?

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -76,24 +77,12 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) error
 // PrintUpStreamVersions prints the origin versions which satisfies the following conditions:
 // 	1. Major versions greater than or equal to the minimum supported and default version
 //	2. Pre-release versions greater than default version
-func PrintUpStreamVersions(output io.Writer, minSupportedVersion string, defaultVersion string) error {
-	var releaseList []string
-	data, err := getGithubReleases()
+func PrintUpStreamVersions(output io.Writer, minSupportedVersion, defaultVersion string) error {
```
gbraad: Can we actually change the `PrintUp` to be just `Print...`

pkg/minishift/openshift/version/openshift_versions.go
gbraad: I would even argue if we really should do `getGithubReleases()` is this function, as it prevents you from doing easy unittests on this function. If we can provide the results to `getOpenShiftTags` we can easily write a test against this.

budhram: Change to "getGithubReleases" is out of scope so we prefer to create another task as refactoring task.

LalatenduMohanty:
@budhram Any update on this?

praveenkumar:
@budhram ping, can you update the PR as per suggestions?

pkg/minishift/openshift/version/openshift_versions.go
budhram: Done @gbraad :smile:

budhram:
@gbraad @praveenkumar Please review.

cmd/minishift/cmd/start.go
gbraad: Why those this need to be in start? This is a preparation step that was moved to the pre-flight... right?

agajdosi:
Works as expected for me on Fedora.

pkg/minishift/openshift/openshift.go
```diff
@@ -298,3 +301,22 @@ func getOpenshiftVersion() string {
 	}
 	return openshiftVersion
 }
+
+func GetOpenShiftVersionDuringStart() (string, error) {
+	tag := viper.GetString(configCmd.OpenshiftVersion.Name)
```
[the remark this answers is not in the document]
LalatenduMohanty: This function should be in cmd/minishift/cmd/util

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -76,24 +77,17 @@ func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) error
...
```
LalatenduMohanty: @budhram  The only change I can see here is you have changed `minSupportedVersion string` to `minSupportedVersion`. Any specific reason for this?

pkg/minishift/openshift/openshift.go
```diff
@@ -298,3 +301,22 @@ func getOpenshiftVersion() string {
...
+	if tag == "vlatest" {
```
gbraad: :-s, are really using `vlatest` ?

gbraad: if this is the result of the prefix, please abstract this and combine using `package.VersionPrefix + "latest"`... or document wit comment

pkg/minishift/openshift/version/openshift_versions.go
budhram: @LalatenduMohanty Will revert it, else this is the Go idiom :wink:

gbraad: I also do not see the need to be obscure. please leave code readable.
I do not care what 'Gophers' like... I care about readability and non-guessing.

pkg/minishift/openshift/version/openshift_versions.go
gbraad: the reason why I mention this is the following:> ' `Print up` is a prhrasal verb that means to show something, while you mean to say `Upstream` and ` Downstream` which is a conjoined word (no spaces, no dash).

gbraad: @minishift/minishift-dev ^^ FYI

budhram:
@gbraad ,
> do not use UpStream, but Upstream and Downstream instead.

Agree with you here, but separate issue to discuss over. I feel not in this scope of this PR.

gbraad:
I disagree. You are handling versions here... and the impact is very minimal to do this change. you actually work and improved the code, so.... you touched it already.

budhram:
> I disagree. You are handling versions here... and the impact is very minimal to do this change. you actually work and improved the code, so.... you touched it already.

Others are still yet to know and come into discussion. 
If all agree then only I will do :smile: 
/cc @LalatenduMohanty @praveenkumar @amitkrout @agajdosi @anjannath WDYT ?

gbraad:
> If all agree then only I will do 😄

what is the impact ?!?

LalatenduMohanty:
> do not use UpStream, but Upstream and Downstream instead.

+1 , I would prefer Upstream

praveenkumar:
@budhram you need to rebase it with the master.

cmd/minishift/cmd/util/util.go
```diff
@@ -160,3 +163,22 @@ func OcClusterDown(hostVm *host.Host) error {
 	_, err := sshCommander.SSHCommand(cmd)
 	return err
 }
+func GetOpenShiftVersionDuringStart() (string, error) {
```
LalatenduMohanty: The name `GetOpenShiftVersionDuringStart` sounds weird to me as there is no co-relation between getting OpenShift versions from github and getting the versions during minishift start.

budhram: @LalatenduMohanty any good name?

budhram: @LalatenduMohanty  `GetDefaultOrLatestOpenShiftVersion`?

LalatenduMohanty: @budhram GetLatestOpenShiftVersion sounds fine to me.

budhram:
CI is failing due to known random issue and hence merging it after getting confirmation from @praveenkumar and @agajdosi .

https://ci.centos.org/job/minishift-pr/4797/console

PR #2642, opened 2018-07-31
Issue #2454 Fix check for Default Switch

### LINKED ISSUE #2454: Correct the code for default switch check for HyperV
The environment variable does not exist anymore

https://github.com/minishift/minishift/blob/1426bddcbf7aae0b8fab29dbe28b667a9ae218cb/pkg/minishift/network/settings_windows.go#L43

anjannath:
Fixes #2454 #2885 

One way to test if this is working properly.. start minishift in windows using the `Default Switch` and provide an IP address for it, `minishift` will try to guess the netmask, and if this works as expected the netmask should be `225.255.255.240`

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -158,8 +158,8 @@ var (
 	DnsmasqContainerized  = createConfigSetting("network-dnsmasq-containerized", SetBool, nil, nil, true, false)
 	DnsmasqContainerImage = createConfigSetting("network-dnsmasq-container", SetString, nil, nil, true, nil)
 
-	// Hyper-V
-	HypervVirtualSwitch = createConfigSetting("hyperv-virtual-switch", SetString, []setFn{validations.IsValidHypervVirtualSwitch}, nil, true, nil)
+	// Hyper-V vSwitch set to Default Switch by default
+	HypervVirtualSwitch = createConfigSetting("hyperv-virtual-switch", SetString, []setFn{validations.IsValidHypervVirtualSwitch}, nil, true, "Default Switch")
 
 	// Save start flags to viper config
 	SaveStartFlags = createConfigSetting("save-start-flags", SetBool, nil, nil, true, true)
--- a/cmd/minishift/cmd/start_preflight.go
+++ b/cmd/minishift/cmd/start_preflight.go
@@ -30,6 +30,7 @@ import (
 	configCmd "github.com/minishift/minishift/cmd/minishift/cmd/config"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	validations "github.com/minishift/minishift/pkg/minishift/config"
+	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
 	"github.com/minishift/minishift/pkg/minishift/network"
 	"github.com/minishift/minishift/pkg/minishift/shell/powershell"
 	"github.com/minishift/minishift/pkg/util/github"
@@ -44,10 +45,9 @@ import (
 )
 
 const (
-	StorageDisk                  = "/mnt/?da1"
-	StorageDiskForGeneric        = "/"
-	GithubAddress                = "https://github.com"
-	hypervDefaultVirtualSwitchId = "c08cb7b8-9b3c-408e-8e30-5e16a3aeb444"
+	StorageDisk           = "/mnt/?da1"
+	StorageDiskForGeneric = "/"
+	GithubAddress         = "https://github.com"
 )
 
 // preflightChecksBeforeStartingHost is executed before the startHost function.
@@ -414,9 +414,9 @@ func checkHypervDriverSwitch() bool {
 
 	switchName := viper.GetString(configCmd.HypervVirtualSwitch.Name)
 
-	// check for default switch
-	if switchName == "" {
-		checkIfDefaultSwitchExists := fmt.Sprintf("Get-VMSwitch -Id %s | ForEach-Object { $_.Name }", hypervDefaultVirtualSwitchId)
+	// check for default switch by using the Id
+	if switchName == minishiftConstants.HypervDefaultVirtualSwitchName {
+		checkIfDefaultSwitchExists := fmt.Sprintf("Get-VMSwitch -Id %s | ForEach-Object { $_.Name }", minishiftConstants.HypervDefaultVirtualSwitchId)
 		stdOut, stdErr, _ := posh.Execute(checkIfDefaultSwitchExists)
 
 		if !strings.Contains(stdErr, "Get-VMSwitch") {
--- a/pkg/minishift/constants/constants.go
+++ b/pkg/minishift/constants/constants.go
@@ -39,6 +39,8 @@ const (
 	OcPathInsideVM                 = "/var/lib/minishift/bin"
 	BaseDirInsideInstance          = "/var/lib/minishift/base"
 	ImageNameForClusterUpImageFlag = "openshift/origin-${component}"
+	HypervDefaultVirtualSwitchId   = "c08cb7b8-9b3c-408e-8e30-5e16a3aeb444"
+	HypervDefaultVirtualSwitchName = "Default Switch"
 )
 
 var (
--- a/pkg/minishift/network/settings_windows.go
+++ b/pkg/minishift/network/settings_windows.go
@@ -20,7 +20,6 @@ import (
 	"encoding/base64"
 	"fmt"
 	"net"
-	"os"
 	"strings"
 	"time"
 
@@ -40,13 +39,6 @@ var (
 	successCount = 0
 )
 
-func isUsingDefaultSwitch() bool {
-	switchEnvName := "HYPERV_VIRTUAL_SWITCH"
-	switchEnvValue := os.Getenv(switchEnvName)
-
-	return switchEnvValue == "Default Switch" || switchEnvValue == ""
-}
-
 func getAddressAssignedToDefaultSwitch() string {
 	posh := powershell.New()
 
@@ -57,14 +49,14 @@ func getAddressAssignedToDefaultSwitch() string {
 }
 
 func determineNetmask() string {
-	if isUsingDefaultSwitch() {
+	if IsUsingDefaultSwitch() {
 		return "255.255.255.240" // 28
 	}
 	return "255.255.255.0" // 24
 }
 
 func determineNameservers() []string {
-	if isUsingDefaultSwitch() {
+	if IsUsingDefaultSwitch() {
 		return []string{getAddressAssignedToDefaultSwitch()}
 	}
 
@@ -85,7 +77,7 @@ func determineNameservers() []string {
 }
 
 func determineDefaultGateway(ipaddress string) string {
-	if isUsingDefaultSwitch() {
+	if IsUsingDefaultSwitch() {
 		return getAddressAssignedToDefaultSwitch()
 	}
 
--- a/pkg/minishift/network/utils.go
+++ b/pkg/minishift/network/utils.go
@@ -28,7 +28,12 @@ import (
 	"strings"
 
 	"github.com/docker/machine/libmachine/drivers"
+	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
+	"github.com/minishift/minishift/pkg/minishift/profile"
+	"github.com/minishift/minishift/pkg/minishift/shell/powershell"
 )
+
+var VMSwitch string
 
 // This will return the address as used by libmachine
 func GetIP(driver drivers.Driver) (string, error) {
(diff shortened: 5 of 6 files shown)


pkg/minishift/network/settings_windows.go
```diff
@@ -41,10 +42,8 @@ var (
 )
 
 func isUsingDefaultSwitch() bool {
-	switchEnvName := "HYPERV_VIRTUAL_SWITCH"
-	switchEnvValue := os.Getenv(switchEnvName)
-
-	return switchEnvValue == "Default Switch" || switchEnvValue == ""
+	switchName := viper.GetString(configCmd.HypervVirtualSwitch.Name)
+	return switchName == "Default Switch" || switchName == nil
```
gbraad: the switch should be retrieved with PowerShell, and we should compare the `ID`, since the name is localized.

LalatenduMohanty:
@anjannath Please rebase the PR.

anjannath:
@LalatenduMohanty Done!

cmd/minishift/cmd/start_preflight.go
```diff
@@ -30,6 +30,7 @@ import (
 	configCmd "github.com/minishift/minishift/cmd/minishift/cmd/config"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	validations "github.com/minishift/minishift/pkg/minishift/config"
+	miniConstants "github.com/minishift/minishift/pkg/minishift/constants"
```
budhram: @anjannath , please avoid creating another alias as we already have alias dedicated for such packages.

I can give you a page explanation why it is like [`To put axe on your own foot.`](https://www.urbandictionary.com/define.php?term=To%20put%20axe%20on%20your%20own%20foot.) 😄  

See https://github.com/minishift/minishift/search?q=minishiftConstants&unscoped_q=minishiftConstants

pkg/minishift/network/settings_windows.go
```diff
@@ -20,13 +20,14 @@ import (
...
 	"fmt"
 	"net"
-	"os"
 	"strings"
 	"time"
 
 	hvkvp "github.com/gbraad/go-hvkvp"
 	"github.com/golang/glog"
 	minishiftConfig "github.com/minishift/minishift/pkg/minishift/config"
+	"github.com/minishift/minishift/pkg/minishift/constants"
```
budhram: I would say stick to one format so update this as 
```
minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
```

so that next time user will not do any error by adding another import.

gbraad: the problem is that we RELY on minikube constants, while we should have moved away a long time ago.

pkg/minishift/network/settings_windows.go
```diff
@@ -41,10 +42,13 @@ var (
 )
 
 func isUsingDefaultSwitch() bool {
-	switchEnvName := "HYPERV_VIRTUAL_SWITCH"
-	switchEnvValue := os.Getenv(switchEnvName)
+	posh := powershell.New()
+	switchNameCommand := fmt.Sprintf("$swname=(Get-VM \"%s\" | Get-VMNetworkAdapter | select SwitchName).SwitchName", profile.GetActiveProfile())
```
gbraad: `get-vm`? This would mean this only works AFTER the vm has been created, right?

pkg/minishift/constants/constants.go
```diff
@@ -24,18 +24,20 @@ import (
...
+	OpenshiftApiContainerName    = "k8s_apiserver_openshift-apiserver"
+	OpenshiftExec                = "/usr/bin/openshift"
+	OpenshiftOcExec              = "/usr/bin/oc"
+	DefaultProject               = "myproject"
+	DefaultUser                  = "developer"
+	DefaultUserPassword          = "developer"
+	SkipVerifyInsecureTLS        = "insecure-skip-tls-verify=true"
+	BinaryName                   = "minishift"
+	OcPathInsideVM               = "/var/lib/minishift/bin"
+	HypervDefaultVirtualSwitchId = "c08cb7b8-9b3c-408e-8e30-5e16a3aeb444"
```
gbraad: 👍

budhram:
Functionally it seems to be working fine. I am able to start minishift VM.
```
 .\minishift.exe start --hyperv-virtual-switch "mini_s"
[...]
-- Checking if Hyper-V driver is configured to use a Virtual Switch ...
   'mini_s' ... OK
[...]
-- Starting Minishift VM ............................................................................... OK
   Pinging 8.8.8.8 ... OK
[...]
To login as administrator:
    oc login -u system:admin
```

pkg/minishift/network/settings_windows.go
anjannath: @gbraad Right, this function is used in the process of setting the ip address which happens after the vm has been created so, and also the name suggests, if the VM is using the default switch

cmd/minishift/cmd/start_preflight.go
LalatenduMohanty: Agree with @budhram. It should be `minishiftConstants`.

LalatenduMohanty: Agree with @budhram. It should be minishiftConstants.

pkg/minishift/constants/constants.go
LalatenduMohanty: 👍

pkg/minishift/network/settings_windows.go
LalatenduMohanty: @anjannath I agree with budhram.

pkg/minishift/network/settings_windows.go
```diff
@@ -41,10 +42,13 @@ var (
...
+	_, _, _ = posh.Execute(switchNameCommand)
```
LalatenduMohanty: Why are we running this i.e. `_, _, _ = posh.Execute(switchNameCommand)` as we are ignoring the output anuway? we are also running it again below `id, _, _ := posh.Execute(switchIdComman`.

gbraad: if using `_, _, _` you do not need the actual ignore options... :-s Just doing `posh.Execute(switchNameCommand)` would be enough

gbraad:
@LalatenduMohanty can you describe which scenario you have tested; new instance and/or existing instance?

amitkrout:
@anjannath What i understand is the existing function ```func isUsingDefaultSwitch() bool {``` is refactored. So user experience would be same as it was before with flag ```--hyperv-virtual-switch```. 

Correct me if i am wrong ?   If possible add a quick description of fix in the issue description area.

pkg/minishift/network/settings_windows.go
anjannath: Okay.. i did not know that!

anjannath:
@amitkrout No user experience changes for this PR, i have updated the first comment on how to test this and what it does.

LalatenduMohanty:
> @LalatenduMohanty can you describe which scenario you have tested; new instance and/or existing instance?

I guess we can call it is a new setup as I had to enable HyperV first and then test the PR.

anjannath:
Moving to WIP as the check doesnot work if a VM is not present. Will update the PR..

pkg/minishift/network/settings_windows.go
```diff
@@ -41,10 +42,21 @@ var (
 )
 
 func isUsingDefaultSwitch() bool {
-	switchEnvName := "HYPERV_VIRTUAL_SWITCH"
-	switchEnvValue := os.Getenv(switchEnvName)
+	posh := powershell.New()
+	switchIdCommand := `(Get-VMSwitch -Name $swname).Id.Guid`
+	getSwitchIdForVmCommand := fmt.Sprintf("$swname=(Get-VM \"%s\" | Get-VMNetworkAdapter | select SwitchName).SwitchName;%s",
+		profile.GetActiveProfile(),
```
gbraad: We might end up jumbling arund here with the active profile. but for now, this works, although I believe this might get risky going forward...

anjannath:
@gbraad Still facing a cyclic dependency issue, making a new package didn't solve it.
I am not able to import anything under the `cmd` namespace to the `network` package.

anjannath:
retest this please

anjannath:
@amitkrout Can you try the binary from this PR and check if #2638 is fixed, when you use the `Default Switch` Just pass the ip address in the same network as the ip address of the switch, it should be able to guess the correct gateway and netmask.

amitkrout:
@anjannath thaks for the fix. Will verify the bit as directed

anjannath:
@amitkrout Sure i'll try that tomorrow.

pkg/minishift/network/validations/validations.go
```diff
@@ -0,0 +1,44 @@
...
+
+import (
+	"fmt"
+	"strings"
+
+	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
+	"github.com/minishift/minishift/pkg/minishift/shell/powershell"
+)
+
+var VMSwitch string
```
praveenkumar: @anjannath this doesn't make much sense, if you check your `IsUsingDefaultSwitch` implementation then it would be good to pass the switchName (whatever it will be) then evaluate it.

anjannath: @praveenkumar if you mean instead of using this variable just pass the value we get from `viper.GetString(configCmd.HypervVirtualSwitch.Name)` ?

cmd/minishift/cmd/config/config.go
```diff
@@ -158,8 +158,8 @@ var (
 	DnsmasqContainerized  = createConfigSetting("network-dnsmasq-containerized", SetBool, nil, nil, true, false)
 	DnsmasqContainerImage = createConfigSetting("network-dnsmasq-container", SetString, nil, nil, true, nil)
 
-	// Hyper-V
-	HypervVirtualSwitch = createConfigSetting("hyperv-virtual-switch", SetString, []setFn{validations.IsValidHypervVirtualSwitch}, nil, true, nil)
+	// Hyper-V vSwitch set to Default Switch by default
+	HypervVirtualSwitch = createConfigSetting("hyperv-virtual-switch", SetString, []setFn{validations.IsValidHypervVirtualSwitch}, nil, true, "Default Switch")
```
gbraad: Right...

cmd/minishift/cmd/start_preflight.go
```diff
@@ -410,13 +410,14 @@ func checkLibvirtDefaultNetworkActive() bool {
...
+	const defaultSwitch = "Default Switch"
 	posh := powershell.New()
 
 	switchName := viper.GetString(configCmd.HypervVirtualSwitch.Name)
 
-	// check for default switch
-	if switchName == "" {
-		checkIfDefaultSwitchExists := fmt.Sprintf("Get-VMSwitch -Id %s | ForEach-Object { $_.Name }", hypervDefaultVirtualSwitchId)
+	// check for default switch by using the Id
+	if switchName == defaultSwitch {
```
gbraad: treat as constant, like in the next line

pkg/minishift/constants/constants.go
```diff
@@ -39,6 +39,7 @@ const (
 	OcPathInsideVM                 = "/var/lib/minishift/bin"
 	BaseDirInsideInstance          = "/var/lib/minishift/base"
 	ImageNameForClusterUpImageFlag = "openshift/origin-${component}"
+	HypervDefaultVirtualSwitchId   = "c08cb7b8-9b3c-408e-8e30-5e16a3aeb444"
```
gbraad: also add `HypervDefault...Name`

pkg/minishift/network/utils.go
```diff
@@ -134,3 +139,41 @@ func OverrideProxyForLocalConnections(proxyAddr string) error {
...
+
+// IsUsingDefaultSwitch returns true if the Default Switch is used before creating the VM
+func IsUsingDefaultSwitch() bool {
+	posh := powershell.New()
+
+	checkIfDefaultSwitchExists := fmt.Sprintf("Get-VMSwitch -Id %s | ForEach-Object { $_.Name }", minishiftConstants.HypervDefaultVirtualSwitchId)
+	stdOut, stdErr, _ := posh.Execute(checkIfDefaultSwitchExists)
+
+	// if the config variable is epmty and default switch exists
+	if VMSwitch == "" && !strings.Contains(stdErr, "Get-VMSwitch") {
```
gbraad: Maybe add something along the lines of:  `// If err contains the command, the execution failed`

pkg/minishift/network/utils.go
```diff
@@ -134,3 +139,41 @@ func OverrideProxyForLocalConnections(proxyAddr string) error {
...
+		return true
+	}
+	return false
+}
+
+// GetVMSwitchId returns the ID of the vswitch in use by the running minishift vm on hyperv
+func GetVMSwitchId() string {
+	posh := powershell.New()
+
+	switchNameCommand := fmt.Sprintf("(Get-VMSwitch (Get-VM \"%s\" | Get-VMNetworkAdapter | select SwitchName).SwitchName).Id.Guid", profile.GetActiveProfile())
```
gbraad: `switchIdCommand`

anjannath:
@gbraad Done with the changes..

Issue #2645, opened 2018-07-31
Disable EAP add-on test for release and nightly job


LalatenduMohanty:
We enabled the xpass add-on as the root cause of the original failure was fixed. Refer #2582 . However it still fails.  https://ci.centos.org/job/minishift-release/100/consoleFull , https://ci.centos.org/job/minishift-release/101/consoleFull

https://github.com/minishift/minishift/issues/2644 is created to track the failure. 

```
08:04:23 --- Failed steps:
08:04:23 
08:04:23   Scenario Outline: User deploys, checks out and deletes several templates from XpaaS imagestream # features/addon-xpaas.feature:20
08:04:23     And service "eap-app" rollout successfully within "1200" seconds # features/addon-xpaas.feature:24
08:04:23       Error: Not all services successfully rolled out:
08:04:23 Not all successfully rolled out:
08:04:23 Service: 'eap-app'
08:04:23 -StdOut: Deployment config "eap-app" waiting on image update
08:04:23 Waiting for latest deployment config spec to be observed by the controller loop...
08:04:23 Waiting for latest deployment config spec to be observed by the controller loop...
08:04:23 Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
08:04:23 Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
08:04:23 Waiting for rollout to finish: 0 out of 1 new replicas have been updated...
08:04:23 Waiting for rollout to finish: 0 of 1 updated replicas are available...
08:04:23 
08:04:23  Service build output logs: Cloning "https://github.com/jboss-developer/jboss-eap-quickstarts" ...
08:04:23 	Commit:	465e5a2e7900cdb744740745f3d65a110d6e9962 (Add generated docs)
08:04:23 	Author:	Tomaz Cerar <tomaz.cerar@gmail.com>
08:04:23 	Date:	Thu Dec 14 01:05:08 2017 +0100
08:04:23 Pulling image "registry.access.redhat.com/jboss-eap-7/eap71-openshift@sha256:d5a066bf46835c12cc794dc6481f41efe082afc40b7900a87ff9f881764d903a" ...
08:04:23 Found pom.xml... attempting to build with 'mvn -e -Popenshift -DskipTests -Dcom.redhat.xpaas.repo.redhatga package --batch-mode -Djava.net.preferIPv4Stack=true -Dcom.redhat.xpaas.repo.jbossorg'
[...]

PR #2646, opened 2018-07-31
Issue #2645 Disable addon test from release

praveenkumar:


LalatenduMohanty:
@minishift/minishift-dev @praveenkumar and we had an discussion and agreed that this is better choice than disabling the test in both release and nightly.  Currently `make integration_all` and `make integration_pr` is same after xpass add-on test is disabled. So atleast keeping it enabled for nightly will he helpful to fix the issue proactively as nightly will fail because of this.

Issue #2666, opened 2018-08-06
minishift profile set xxxx reports “…: no such file or directory”


amitkrout:
### General information

  * Minishift version: v1.22.0/master
  * OS: Linux / macOS / Windows
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce

  1. Make sure no previous minishift configuration should present on the system
  2. Make sure no minishift sub command should run which creates the ```.minishift``` dir
  3. 
```
$ ./minishift profile set Test
F0805 23:18:08.339053    1653 root.go:389] Cannot create file '/Users/amit/.minishift/config/global.json': open /Users/amit/.minishift/config/global.json: no such file or directory
```
  4. 

### Expected
Should create ```Test``` profile. However running the same command after profile list set the ```Test``` profile.
```
$ ./minishift profile set Test
F0805 23:18:08.339053    1653 root.go:389] Cannot create file '/Users/amit/.minishift/config/global.json': open /Users/amit/.minishift/config/global.json: no such file or directory
$ ./minishift profile list
- Test		Does Not Exist
- minishift	Does Not Exist	(Active)
$ ./minishift profile set Test
Profile 'Test' set as active profile.
```
### Actual
Creates the ```Test``` profile despite of the weird output, but it does not set it as active profile 
 ```
$ ./minishift profile set Test
F0805 23:18:08.339053    1653 root.go:389] Cannot create file '/Users/amit/.minishift/config/global.json': open /Users/amit/.minishift/config/global.json: no such file or directory
 $ ./minishift profile list
- Test		Does Not Exist
- minishift	Does Not Exist	(Active)
```
### Logs

You can start Minishift with `minishift start --show-libmachine-logs -v5` to collect logs.
Please consider posting this on http://gist.github.com/ and post the link in the issue.

**Comment by LalatenduMohanty** [comment]
@amitkrout @anjannath I can not reproduce this issue on mac. Is it a windows specific issue?

**Comment by LalatenduMohanty** [comment]
It is reproducible when ~/.minishift directory is not present.

``` 
$ mv .minishift .minishift_backup
$ minishift profile set Test
F0806 16:55:03.031042    3686 root.go:389] Cannot create file '/Users/lmohanty/.minishift/config/global.json': open /Users/lmohanty/.minishift/config/global.json: no such file or directory
```

123avi:
Getting the same 
```  ./minishift profile set tutorial ```
result : 
[...]

PR #2669, opened 2018-08-06
Issue #2666 create instance dirs of default profile for global config

anjannath:
Fixes: #2666 

- create instance dirs of default profile for global config irrespective of the profile in use, to store the global config

How to test : 

1. Make sure no previous minishift configuration should present on the system
2. Make sure no minishift sub command should run which creates the ```.minishift``` dir
3. Then ```minishift profile set Test``` successfully creates the ```Test``` profile

### DIFF
--- a/cmd/minishift/cmd/root.go
+++ b/cmd/minishift/cmd/root.go
@@ -92,6 +92,8 @@ var RootCmd = &cobra.Command{
 		// If profile name is 'minishift' then ignore the vaild profile check.
 		if constants.ProfileName != constants.DefaultProfileName {
 			checkForValidProfileOrExit(cmd)
+			//create the default profile dirs for global config
+			createMinishiftDirs(state.NewMinishiftDirs(constants.GetProfileHomeDir(constants.DefaultProfileName)))
 		}
 
 		constants.MachineName = constants.ProfileName


anjannath:
@amitkrout Could you please verify this, and also there are no unexpected side effects..

amitkrout:
@anjannath Ack

cmd/minishift/cmd/root.go
```diff
@@ -92,6 +92,8 @@ var RootCmd = &cobra.Command{
 		// If profile name is 'minishift' then ignore the vaild profile check.
 		if constants.ProfileName != constants.DefaultProfileName {
 			checkForValidProfileOrExit(cmd)
+			//create the default profile dirs for global config
+			createMinishiftDirs(state.NewMinishiftDirs(constants.GetProfileHomeDir(constants.DefaultProfileName)))
```
LalatenduMohanty: This does not look correct as we have `createMinishiftDirs` before `ensureConfigFileExists(constants.GlobalConfigFile)`. Need to find out what actually happening. 

```code
// creating all directories for minishift run
createMinishiftDirs(state.InstanceDirs)

// Ensure the global viper config file exists.
ensureConfigFileExists(constants.GlobalConfigFile)
```

amitkrout:
@anjannath I have added ```how to test``` in the description
Thanks for the fix. Tested it on this pr and it works fine, However it breaks other usecase
```
$ ./minishift profile set test
Profile 'test' set as active profile.

$ ./minishift profile list
- minishift	Does Not Exist
- test		Does Not Exist	(Active)

$ ./minishift addons list
- admin-user	 : disabled	P(0)
- anyuid	 : disabled	P(0)
- registry-route : disabled	P(0)
- xpaas		 : disabled	P(0)

$ ./minishift profile set minishift
Profile 'minishift' set as active profile.

$ ./minishift addons list
$
```
As per the default behavior the last command should list out all the default addons in profile minshift such as 
```
$ ./minishift addons list
- admin-user	 : disabled	P(0)
- anyuid	 : disabled	P(0)
- registry-route : disabled	P(0)
- xpaas		 : disabled	P(0)
```

cmd/minishift/cmd/root.go
LalatenduMohanty: The issue is that when ~/.minishift/config is not present and user runs `minishift profile set  NON_DEFAULT_PROFILE_NAME` , `createMinishiftDirs(state.InstanceDirs)` only created the directory structure in `~./minishift/profiles`  and not the `~/.minishift/config` . Hence `ensureConfigFileExists(constants.GlobalConfigFile)` gives error as it expects that `~/.minishift/config` would be present.

```
$ tree ~/.minishift
/Users/lmohanty/.minishift
├── cache
│   ├── images
│   ├── iso
│   └── oc
└── profiles
    └── Test
        ├── addons
        ├── certs
        ├── config
        ├── logs
        ├── machines
        └── tmp
```

anjannath: @LalatenduMohanty Right, that is why i am adding that line to create the directory structure on the upper level then profile, so that the config directory exists.

anjannath:
Closed in favor of #2670

PR #2670, opened 2018-08-06
Issue #2666 Create directory for global config

LalatenduMohanty:

The root cause of the issue is that when ~/.minishift/config is not present and user runs `minishift profile set  NON_DEFAULT_PROFILE_NAME` , `createMinishiftDirs(state.InstanceDirs)` only created the directory structure in `~./minishift/profiles`  and not the `~/.minishift/config` . Hence `ensureConfigFileExists(constants.GlobalConfigFile)` gives error as it expects that `~/.minishift/config` would be present.

```
$ tree ~/.minishift
/Users/lmohanty/.minishift
├── cache
│   ├── images
│   ├── iso
│   └── oc
└── profiles
    └── Test
        ├── addons
        ├── certs
        ├── config
        ├── logs
        ├── machines
        └── tmp
```

### DIFF
--- a/cmd/minishift/state/instance_dirs.go
+++ b/cmd/minishift/state/instance_dirs.go
@@ -17,42 +17,46 @@ limitations under the License.
 package state
 
 import (
-	"github.com/minishift/minishift/pkg/minikube/constants"
 	"path/filepath"
+
+	"github.com/minishift/minishift/pkg/minikube/constants"
 )
 
 type MinishiftDirs struct {
-	Home       string
-	Config     string
-	Machines   string
-	Certs      string
-	Cache      string
-	IsoCache   string
-	OcCache    string
-	ImageCache string
-	Addons     string
-	Logs       string
-	Tmp        string
+	Home         string
+	Config       string
+	GlobalConfig string
+	Machines     string
+	Certs        string
+	Cache        string
+	IsoCache     string
+	OcCache      string
+	ImageCache   string
+	Addons       string
+	Logs         string
+	Tmp          string
 }
 
 var InstanceDirs *MinishiftDirs
 
 // Constructor for MinishiftDirs
 func NewMinishiftDirs(baseDir string) *MinishiftDirs {
+	minishiftHomeDir := constants.GetMinishiftHomeDir()
 	// We use a global cache, sharing the cache directory of the default 'minishift' instance
-	cacheDir := filepath.Join(constants.GetMinishiftHomeDir(), "cache")
+	cacheDir := filepath.Join(minishiftHomeDir, "cache")
 
 	return &MinishiftDirs{
-		Home:       baseDir,
-		Certs:      filepath.Join(baseDir, "certs"),
-		Machines:   filepath.Join(baseDir, "machines"),
-		Addons:     filepath.Join(baseDir, "addons"),
-		Logs:       filepath.Join(baseDir, "logs"),
-		Tmp:        filepath.Join(baseDir, "tmp"),
-		Config:     filepath.Join(baseDir, "config"),
-		Cache:      cacheDir,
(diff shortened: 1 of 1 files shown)


cmd/minishift/state/instance_dirs.go
```diff
@@ -17,42 +17,46 @@ limitations under the License.
...
-		OcCache:    filepath.Join(cacheDir, "oc"),
-		ImageCache: filepath.Join(cacheDir, "images"),
+		Home:         baseDir,
+		Certs:        filepath.Join(baseDir, "certs"),
+		Machines:     filepath.Join(baseDir, "machines"),
+		Addons:       filepath.Join(baseDir, "addons"),
+		Logs:         filepath.Join(baseDir, "logs"),
+		Tmp:          filepath.Join(baseDir, "tmp"),
+		Config:       filepath.Join(baseDir, "config"),
+		GlobalConfig: filepath.Join(minishiftHomeDir, "config"),
```
LalatenduMohanty: For default profile GlobalConfig and Config would be same but that is not an issue as https://github.com/minishift/minishift/blob/master/cmd/minishift/cmd/root.go#L434-L443 `os.MkdirAll` does not do anything if the directory is present.

cmd/minishift/state/instance_dirs.go
```diff
@@ -17,42 +17,46 @@ limitations under the License.
...
 }
 
 var InstanceDirs *MinishiftDirs
 
 // Constructor for MinishiftDirs
 func NewMinishiftDirs(baseDir string) *MinishiftDirs {
+	minishiftHomeDir := constants.GetMinishiftHomeDir()
 	// We use a global cache, sharing the cache directory of the default 'minishift' instance
-	cacheDir := filepath.Join(constants.GetMinishiftHomeDir(), "cache")
+	cacheDir := filepath.Join(minishiftHomeDir, "cache")
```
anjannath: Why change this, will the env variable `MINISHIFT_HOME_DIR` work if we change this?

cmd/minishift/state/instance_dirs.go
anjannath: :+1:

cmd/minishift/state/instance_dirs.go
anjannath: Sorry, didn't see you added it in the first line

cmd/minishift/state/instance_dirs.go
```diff
@@ -17,42 +17,46 @@ limitations under the License.
...
-	Cache      string
-	IsoCache   string
-	OcCache    string
-	ImageCache string
-	Addons     string
-	Logs       string
-	Tmp        string
+	Home         string
+	Config       string
+	GlobalConfig string
```
praveenkumar: :+1:

cmd/minishift/state/instance_dirs.go
```diff
@@ -17,42 +17,46 @@ limitations under the License.
...
+	Addons       string
+	Logs         string
+	Tmp          string
 }
 
 var InstanceDirs *MinishiftDirs
 
 // Constructor for MinishiftDirs
 func NewMinishiftDirs(baseDir string) *MinishiftDirs {
+	minishiftHomeDir := constants.GetMinishiftHomeDir()
```
praveenkumar: :+1:

LalatenduMohanty:
@amitkrout Can you please check if it fixes the issue.

amitkrout:
Ack, Will verify and update

amitkrout:
@praveenkumar @LalatenduMohanty Works as expected. Tested successfully the same usecase https://github.com/minishift/minishift/pull/2669#issuecomment-410695688. Also did some smoke testing with global ```--profile``` flag without any issue.

Please merge the pr

Issue #2683, opened 2018-08-09
Use PersistentPreRun to stop version, completion commands from creating instance dirs


anjannath:
Right now we are doing a check in `root.go` to stop PersistentPreRun from executing for the two commands, Instead we should use `cobra.Command.PersistentPreRun` to achieve the same.

https://github.com/minishift/minishift/blob/a9f8d156fdb5864fba88baeac98bdce78ac84bf6/cmd/minishift/cmd/root.go#L65-L68
--------------------------------------------------------------------------
https://github.com/minishift/minishift/blob/a9f8d156fdb5864fba88baeac98bdce78ac84bf6/cmd/minishift/cmd/root.go#L81-L85

LalatenduMohanty:
Resolved via https://github.com/minishift/minishift/commit/8ca303ac99a28038b2d606d55df3bc5c5d6fbf4f

PR #2684, opened 2018-08-09
Issue #2683 use PersistentPreRun for version, completion cmd

anjannath:
Fixes: #2683 

* Use `PersistentPreRun` for `version` and `completion` command to stop them from creating instance dirs

### DIFF
--- a/cmd/minishift/cmd/completion.go
+++ b/cmd/minishift/cmd/completion.go
@@ -73,6 +73,9 @@ var completionCmd = &cobra.Command{
 	Use:   "completion SHELL",
 	Short: "Outputs minishift shell completion for the given shell (bash or zsh)",
 	Long:  longDescription,
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		//noop
+	},
 	Run: func(cmd *cobra.Command, args []string) {
 		if len(args) != 1 {
 			fmt.Println("Usage: minishift completion SHELL")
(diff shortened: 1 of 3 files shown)


cmd/minishift/cmd/completion.go
```diff
@@ -73,6 +73,9 @@ var completionCmd = &cobra.Command{
 	Use:   "completion SHELL",
 	Short: "Outputs minishift shell completion for the given shell (bash or zsh)",
 	Long:  longDescription,
+	PersistentPreRun: func(cmd *cobra.Command, args []string) {
+		//noop
```
LalatenduMohanty: @anjannath  can you remove the //noop comments as I do not think we need a comment here or I can change these and merge. Let me know

budhram: @LalatenduMohanty @anjannath I would like to keep it to remind myself that is no operation being performed in this block.
We have such scenarios in other project as well. 
https://github.com/search?q=%2F%2F+noop&type=Code

Update it as `// noop` :wink:

LalatenduMohanty: If you strongly feel like keeping it then please follow the proper indent while making a comment.

LalatenduMohanty:
Merged https://github.com/minishift/minishift/commit/8ca303ac99a28038b2d606d55df3bc5c5d6fbf4f

Issue #650, opened 2017-04-03
Add more explicit mechanism to express dependencies between add-ons


hferentschik:
ATM, the order of addon execution is given by their priority. This feature suggest to add explicit dependency specification via addiotnal metadata, for example:

    # Name: eap
    # Description: Imports EAP templates
    # Depends on: cluster-admin

In this case the _eap_ template has an explicit dependency to cluster-admin. 

If implemented, there is an overlap with the priority functionality and one might want to consider to remove/deprecate it.

hferentschik:
I wam wondering whether it would not be enough to have the ability to specify a default priority in the metadata:

    # Name: eap
    # Description: Imports EAP templates
    # Priority: 5

Assuming for example, that one wants to run a custom addon after provided addons, one would just set a priority higher then the default.

jorgemoralespou:
I would say that the dependency mechanism could be useful when you have 2
separate addons (xpaas templates and docker pull xpaas templates). I would
probably want to add the templates if I explicitly pull down the xpaas
images. (Maybe not the best example) But this would avoid introducing a 3rd
addon which would do both things together.

hferentschik:
I am just wondering whether one gains anything by switching to named dependencies. You can pretty much achieve the same thing with priority, especially if we allow to define a default priority as part of the metadata. In fact this is even more generic and we don't have to determine cyclic dependencies, etc.

jorgemoralespou:
I think that a dependency of an addon to another is required more than priority, as now we can install/apply at addon at any time and that addon might require another that is not installed or applied.

e.g: 
Look into the [set of commands in the comment](https://github.com/minishift/minishift/issues/1111#issuecomment-314402483). When converting this into an addon, it requires anyuid, or at least anyuid for that user. I can do anyuid for that user, or require the anyuid addon to be installed.

Not the best option, since anyuid would be done on a per user basis, but think about intalling a springboot, where I rely on the xpaas images to be installed.

PR #2689, opened 2018-08-10
Issue #650 add add-on metadata to express dependencies

anjannath:
Fixes: #650 
### How to test:
Define an add-on as follows in `test-addon/test-addon.addon`

```
# Name: test-addon
# Description: Demo add-on for checking dependency.
# Url: https://minishift.io
# Depends-On: test, anyuid, demo-addon

echo "Minishift"
```

```
$ minishift addons install ~/test-addon/
Add-on installation failed with the error: The add-on depends on: [test, demo-addon]. Please install these first.
```

### DIFF
--- a/docs/source/using/addons.adoc
+++ b/docs/source/using/addons.adoc
@@ -104,6 +104,25 @@ If the metadata field *Minishift-Version* is not specified in the add-on header,
 *Minishift-Version* only supports versions in the form of <major>.<minor>.<patch>.
 ====
 
+[[addon-defining-addon-dependencies]]
+== Defining Add-on Dependencies
+
+Add-on dependencies can be defined using the _Depends-On_ metadata field.
+Multiple dependencies may be defined using a comma-separated list of add-on names.
+
+[[example-addon-dependency]]
+.Example: Defining add-on dependency in add-on definition file
+
+----
+# Name: example
+# Description: Shows the use of Depends-On tag
+# Depends-On: anyuid, admin-user
+
+# This add-on depends on anyuid and admin-user add-on
+
+echo Depends on anyuid, admin-user add-on, and requires them to be installed
+----
+
 [[addon-commands]]
 == Add-on Commands
 
--- a/pkg/minishift/addon/addon_meta.go
+++ b/pkg/minishift/addon/addon_meta.go
@@ -35,6 +35,7 @@ const (
 	anyOpenShiftVersion      = ""
 	anyMinishiftVersion      = ""
 	varDefaults              = "Var-Defaults"
+	dependsOn                = "Depends-On"
 )
 
 type RequiredVar struct {
@@ -51,6 +52,7 @@ type AddOnMeta interface {
 	GetValue(key string) string
 	OpenShiftVersion() string
 	MinishiftVersion() string
+	Dependency() ([]string, error)
 	Url() string
 }
 
@@ -70,6 +72,9 @@ func NewAddOnMeta(headers map[string]interface{}) (AddOnMeta, error) {
 	}
 	if err := varDefaultsCheck(headers); err != nil {
 		return nil, err
+	}
+	if !checkDependencySemantic(headers) {
+		return nil, fmt.Errorf("The Dependencies should be a comma seperated list of Add-ons.")
 	}
 
 	metaData := &DefaultAddOnMeta{headers: headers}
@@ -145,6 +150,23 @@ func (meta *DefaultAddOnMeta) MinishiftVersion() string {
 		return val
 	}
 	return anyMinishiftVersion
+}
+
+func (meta *DefaultAddOnMeta) Dependency() ([]string, error) {
+	if val, contains := meta.headers[dependsOn].(string); contains {
+		return minishiftStrings.SplitAndTrim(val, ",")
+	}
+	return []string{}, nil
+}
+
+func checkDependencySemantic(headers map[string]interface{}) bool {
+	// Comma seperated list of dependencies
+	if headers[dependsOn] != nil {
+		dependencies := headers[dependsOn].(string)
+		match, _ := regexp.MatchString("^(([a-zA-Z][-[a-zA-Z]*)([,][ ]?[a-zA-Z][-[a-zA-Z]*)*)$", dependencies)
+		return match
+	}
+	return true
 }
 
 func checkVersionSemantic(version string) bool {
--- a/pkg/minishift/addon/manager/addon_manager.go
+++ b/pkg/minishift/addon/manager/addon_manager.go
@@ -213,6 +213,9 @@ func (m *AddOnManager) ApplyAddOn(addOn addon.AddOn, context *command.ExecutionC
 	if err := verifyRequiredVariablesInContext(context, addonMetadata); err != nil {
 		return err
 	}
+	if err := m.verifyRequiredAddons(addonMetadata); err != nil {
+		return err
+	}
 
 	oldDir, err := os.Getwd()
 	if err != nil {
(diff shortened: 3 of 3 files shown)


pkg/minishift/addon/addon_meta.go
```diff
@@ -134,6 +139,23 @@ func (meta *DefaultAddOnMeta) OpenShiftVersion() string {
...
 }
 
+func (meta *DefaultAddOnMeta) Dependency() ([]string, error) {
+	if val, contains := meta.headers[dependsOn].(string); contains {
+		return minishiftStrings.SplitAndTrim(val, ",")
+	}
+	return []string{}, nil
+}
+
+func checkDependencySemantic(headers map[string]interface{}) bool {
```
anjannath: @gbraad The question i asked in IRC about the regex, here it is..  should i be using split here instead?

docs/source/using/addons.adoc
```diff
@@ -76,6 +78,12 @@ If the metadata field *OpenShift-Version* is not specified in the add-on header,
 *OpenShift-Version* only supports versions in the form of <major>.<minor>.<patch>.
 ====
 
+[[addon-depends-on]]
+== Add-on Dependency
```
robin-owen: I'd recommend changing this to the following (or something similar):

```adoc
[[addon-defining-addon-dependencies]]
== Defining Add-on Dependencies
```

docs/source/using/addons.adoc
```diff
@@ -27,19 +27,21 @@ The following example shows the contents of an add-on file, including the name a
...
+oc adm policy add-scc-to-group anyuid system:authenticated                              // <8>
 ----
 <1> (Required) Name of the add-on.
 <2> (Required) Description of the add-on.
 <3> (Optional) Comma separated list of required interpolation variables. See xref:../using/addons.adoc#addon-variable-interpolation[Variable Interpolation]
 <4> (Optional) OpenShift version required to run a specific add-on. See xref:../using/addons.adoc#addon-openshift-version-semantics[OpenShift Version Semantic]
-<5> Run an actual add-on command and store its output in a variable. See xref:../using/addons.adoc#internal-addon-variable[Internal Variable]
-<6> Actual add-on command. In this case, the command executes the *oc* binary using the `mytoken` variable value.
-<7> Actual add-on command. In this case, the command executes the *oc* binary.
+<5> (Optional) Dependencies for the add-on. See xref:../using/addons.adoc#addon-depends-on[Add-on Dependency]
```
robin-owen: The section name cross-referenced here will need to be changed.

docs/source/using/addons.adoc
```diff
@@ -76,6 +78,12 @@ If the metadata field *OpenShift-Version* is not specified in the add-on header,
 *OpenShift-Version* only supports versions in the form of <major>.<minor>.<patch>.
 ====
 
+[[addon-depends-on]]
+== Add-on Dependency
+
+In the add-on metadata section you can specify dependencies for an add-on. If a certain add-on requires one more other add-ons to be installed, this can be
```
robin-owen: Each sentence on its own line, please. :)

I'd also recommend a rewrite like the following:

```adoc
Add-on dependencies can be defined using the _Depends-On_ metadata field.
Multiple dependencies may be defined using a comma-separated list of add-on names.
```

Also, should we include an example here? We shouldn't rely on readers looking back up at the example add-on definition.

anjannath: @robin-owen Removed the example from the original `anyuid` add-on and added an example for only `Depends-On` in its own section.

amitkrout:
@anjannath AFAIU ```minishift addons install ~/test-addon/``` only copy paste the test-addon directory to minishift addon directory. The current pr reports if dependent addons is/are not available in ```minishift addons list``` while installing the addons. Then i am expecting the error message will pop out while applying the addons too. For example

```
$ ./minishift addons apply test-addon
-- Applying addon 'test-addon': Add-on apply failed with the error: The add-on depends on: [test, demo-addon]. Please apply these first.
```

Otherwise the pr does not make more sense for checking the availability of addons in the list. 
WDYT?

anjannath:
@amitkrout We cannot apply an add-on without first installing it, so this way if all the dependencies are not installed, the add-on will not get installed. So checking once during installation is fine IMO.

Now if someone uninstalls a dependency then we run into issue, but in that case the errors from the actual add-on commands should be reaching to the users directly. So i don't think we need a check for it during `addon apply` also, lets see what @minishift/minishift-dev thinks about it.

anjannath:
@praveenkumar So the check should only be performed when the add-on is applied?

anjannath:
@amitkrout @praveenkumar Updated PR to perform the check only when the add-on is applied.

amitkrout:
> Updated PR to perform the check only when the add-on is applied.

This really added value

amitkrout:
@anjannath I think you need to update the doc accordingly

docs/source/using/addons.adoc
```diff
@@ -76,6 +76,25 @@ If the metadata field *OpenShift-Version* is not specified in the add-on header,
...
+----
+# Name: example
+# Description: Shows the use of Depends-On tag
+# Depends-On: anyuid, admin-user
+
+# This add-on depends on anyuid and admin-user add-on
+
+echo Depends on anyuid, admin-user add-on, and requires them to be installed
+----
+
```
praveenkumar: @anjannath Shouldn't we https://github.com/minishift/minishift/pull/2689/files#diff-68958b57e627de2b2b78ab4d47fefd62R29 add it here instead of putting an example and explain in this section. Like we do for minimum openshift version and put it as optional.

anjannath: @praveenkumar This was done as per this comment https://github.com/minishift/minishift/pull/2689#discussion_r209266353

anjannath:
@amitkrout Docs changes are not needed..

docs/source/using/addons.adoc
robin-owen: @praveenkumar I'd like to move to this style more generally. At the moment, the way we have the documentation structured requires the user to scroll back up to the example `anyuid` add-on definition above.

Instead, I'd like to see a generic example add-on _and_ include discrete examples for each add-on command/metadata field with their documentation. This allows the reader to go directly to the command/field they need to use and see a brief example of its use there.

For the time being, we could use an example in both locations -- I just don't want the reader to be confused if they go directly to the documentation for the feature they want to use, thus missing the example.

LalatenduMohanty:
@anjannath You need to resolve the merge conflicts for this PR

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -210,6 +210,20 @@ func (m *AddOnManager) ApplyAddOn(addOn addon.AddOn, context *command.ExecutionC
...
+	if err == nil && len(dependencies) > 0 {
+		for _, dep := range dependencies {
+			if m.IsInstalled(dep) {
+				continue
+			}
+			depsNotInstalled = append(depsNotInstalled, dep)
+		}
+		if len(depsNotInstalled) != 0 {
+			return errors.New(fmt.Sprintf("The add-on depends on: [%s]. Please install these first.", strings.Join(depsNotInstalled, ", ")))
+		}
```
LalatenduMohanty: @anjannath This above logic should be in a function and the function can similar to `verifyRequiredAddons()`, else the code is getting bigger for `ApplyAddOn` which is not good programming practice.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -327,6 +330,23 @@ func verifyRequiredMinishiftVersion(meta addon.AddOnMeta) error {
...
+	dependencies, err := metadata.Dependency()
+	if err == nil && len(dependencies) > 0 {
+		for _, dep := range dependencies {
+			if m.IsInstalled(dep) {
+				continue
+			}
+			depsNotInstalled = append(depsNotInstalled, dep)
+		}
+		if len(depsNotInstalled) != 0 {
+			return errors.New(fmt.Sprintf("The add-on depends on: [%s]. Please install these first.", strings.Join(depsNotInstalled, ", ")))
```
LalatenduMohanty: `Please install these first` is technically wrong as installing some dependent add-on is not enough for making sure these add-ons were applied before applying an add-on. So wondering how can we communicate this properly.

LalatenduMohanty: suggestion `fmt.Sprintf("Dependent add-ons [%s] not found for this instance. Please install and apply them before running this add-on.", strings.Join(depsNotInstalled, ", "))`

pkg/minishift/addon/addon_meta.go
```diff
@@ -147,6 +152,23 @@ func (meta *DefaultAddOnMeta) MinishiftVersion() string {
...
+	// Comma seperated list of dependencies
+	if headers[dependsOn] != nil {
+		dependencies := headers[dependsOn].(string)
+		match, _ := regexp.MatchString("^(([a-zA-Z][-[a-zA-Z]*)([,][ ]?[a-zA-Z][-[a-zA-Z]*)*)$", dependencies)
```
gbraad: maybe add some clarification for the expectation?

gbraad: any error reporting when the format is incorrect?

anjannath: @gbraad No, it does not report what is wrong with the format, just that it is wrong. expected is a comma separated list as in `anyuid,xpaas` or `anyuid, xpaas` (i.e one space is accepted) but it should not end with a comma.

PR #2715, opened 2018-08-20
Issue #1586 Make pre-flight check self-explanatory in case of failure

### LINKED ISSUE #1586: Document usage and configuration options of pre-flight checks
I recently got a question from DevStudio QE if somebody who has IP `8.8.8.8` blocked due to some firm policy could use Minishift/CDK which pings this IP during start.

Great thing is that it can be used, however we lack documentation around these options. I think that we should mention, that a lot of stuff around pre-flight-checks can be set via config and at least add some example. As the names of config keys are quite self-explanatory, this could be enough.

But without a mention, it could be hard for user to realize, that this path is open for her to alter the pre-flight checks.

List of config keys we have:
```
 * skip-check-kvm-driver
 * warn-check-kvm-driver
 * skip-check-xhyve-driver
 * warn-check-xhyve-driver
 * skip-check-hyperv-driver
 * warn-check-hyperv-driver
 * skip-check-instance-ip
 * warn-check-instance-ip
 * skip-check-network-host
 * warn-check-network-host
 * skip-check-network-ping
 * warn-check-network-ping
 * skip-check-network-http
 * warn-check-network-http
 * skip-check-storage-mount
 * warn-check-storage-mount
 * skip-check-storage-usage
 * warn-check-storage-usage
 * check-network-http-host
 * check-network-ping-host
```

praveenkumar:
Fixes #1586 

I think instead of document we should proactive tell user to skip the test when we know can be avoided. I am planning to put this logic to all the test which can be skipped without any major blocker in term of starting the VM.

### DIFF
--- a/cmd/minishift/cmd/start_preflight.go
+++ b/cmd/minishift/cmd/start_preflight.go
@@ -76,7 +76,8 @@ func preflightChecksBeforeStartingHost() {
 			checkOriginRelease,
 			fmt.Sprintf("Checking if requested OpenShift version '%s' is valid", requestedOpenShiftVersion),
 			configCmd.WarnCheckOpenShiftRelease.Name,
-			"",
+			fmt.Sprintf("To warn this test use 'minishift config set %s true'\nTo skip this test use 'minishift config set %s true'",
+				configCmd.WarnCheckOpenShiftRelease.Name, configCmd.SkipCheckOpenShiftRelease.Name),
 		)
 	} else {
 		fmt.Printf("FAIL\n")


cmd/minishift/cmd/start_preflight.go
```diff
@@ -74,7 +74,7 @@ func preflightChecksBeforeStartingHost() {
 			checkOriginRelease,
 			fmt.Sprintf("Checking if requested OpenShift version '%s' is valid", viper.GetString(configCmd.OpenshiftVersion.Name)),
 			configCmd.WarnCheckOpenShiftRelease.Name,
-			"",
+			fmt.Sprintf("Use 'minishift config set %s true' to skip this test", configCmd.SkipCheckOpenShiftRelease.Name),
```
LalatenduMohanty: @praveenkumar  When the tests fail the message should suggest that user can skip the test on his own risk  because the pre-flight checks are there for a reason and skipping the test might not fix the root cause of the issue.

LalatenduMohanty: suggestion: ` fmt.Sprintf("If you want to skip the test, run 'minishift config set %s true'", configCmd.SkipCheckOpenShiftRelease.Name)`

praveenkumar: @LalatenduMohanty I don't think `If you want to skip the test,` needed because this message only come when test failure happen so either user fix it or skip it.

praveenkumar: > When the tests fail the message should suggest that user can skip the test on his own risk because the pre-flight checks are there for a reason and skipping the test might not fix the root cause of the issue.

@LalatenduMohanty right but my plan is only to have this message for tests which might not be blocker for the VM start.

gbraad:
Maybe we would like to suggest warn first?

cmd/minishift/cmd/start_preflight.go
LalatenduMohanty: > @LalatenduMohanty right but my plan is only to have this message for tests which might not be blocker for the VM start.

IMO either we should have this for all preflight checks or no message at all. Being selective about the preflight checks will not be optimal from code maintenance point of view.

LalatenduMohanty:
> Maybe we would like to suggest warn first?

Good point, we should mention about both warn and skip .

cmd/minishift/cmd/start_preflight.go
LalatenduMohanty: > @LalatenduMohanty I don't think If you want to skip the test, needed because this message only come when test failure happen so either user fix it or skip it.

I intentionally put `if` because I wanted to leave the decision on the user. Because when a preflight test fails user should the root cause itself and should not skip the test.

LalatenduMohanty: > To skip this test use 'minishift config set %s true' this will give the user the feedback and it the user choice now.

Sounds better

agajdosi: `To skip this test use 'minishift config set %s true'` :+1:

praveenkumar:
CI failure is nothing to do with this PR.

gbraad:
retest this please

praveenkumar:
retest this please.

praveenkumar:
Closing this PR since we need more brainstorm around this on #1586 and the implement what should be done.

PR #2731, opened 2018-08-23
Issue #2677 Add kube as target for openshift config command

### LINKED ISSUE #2677: Provide a way to modify the kube-apiserver config same as openshift-apiserver.
### General information

  * Minishift version: 
  * OS: Linux / macOS / Windows

With 3.10 now there is now 2 different config one for openshift-apiserver which you can modify with the `minishift config set` command and other one is for kube-apiserver for which we don't have any straightforward way to modify.

We might need to add flag to `openshift config command` which also take `kube-apiserver` changes and then restart the respective container.

praveenkumar:
Fixes #2677 

This PR is should be tested with https://github.com/knative/docs/blob/master/install/Knative-with-OpenShift.md#creating-a-new-openshift-cluster instructions.

### DIFF
--- a/pkg/minishift/openshift/openshift.go
+++ b/pkg/minishift/openshift/openshift.go
@@ -49,6 +49,12 @@ func GetOpenShiftPatchTarget(target string) OpenShiftPatchTarget {
 			getContainerConfigFile("node"),
 			getLocalConfigFile("node"),
 		}
+	case "kube":
+		return OpenShiftPatchTarget{
+			"kube",
+			getContainerConfigFile("kube"),
+			getLocalConfigFile("kube"),
+		}
 	default:
 		return OpenShiftPatchTarget{
 			"unkown",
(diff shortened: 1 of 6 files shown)


pkg/minishift/openshift/openshift.go
```diff
@@ -269,6 +274,8 @@ func getLocalConfigFile(target string) string {
 			return "/var/lib/minishift/openshift.local.config/node-localhost/node-config.yaml"
 		}
 		return "/var/lib/minishift/base/node/node-config.yaml"
+	case "kube":
```
LalatenduMohanty: I feel we are going overboard by fixing it in minishift because there are futher changes to the config.yaml files in 3.11. Ideally this should be fixed in OpenShift side.  
Also I am wondering whether we should call it only kube or kubeapiserver as it is admission webhook is a kube api server thing.

praveenkumar: @LalatenduMohanty we always have to provide a way to user to patch the openshift config or kube config even webhook become the default. We are overboard this because that's the change we have to accommodate from cluster up side.

jorgemoralespou:
@praveenkumar @LalatenduMohanty are we sure we are not getting into deep waters if we start patching things that are changing? Have you aligned with @mfojtik about changes to come?

LalatenduMohanty:
> are we sure we are not getting into deep waters if we start patching things that are changing? Have you aligned with @mfojtik about changes to come?

Right, that is my concern as well. We need sync with  @mfojtik  on this.

praveenkumar:
retest this please.

LalatenduMohanty:
retest this please

praveenkumar:
@agajdosi We are explicitly told long back to remove eap6.x from the template and from the image stream, if you see 3.9 one those are not part of addon.

PR #2738, opened 2018-08-27
Issue #2720 Separate tests into core and extra to be run in CI

### LINKED ISSUE #2720: Separate tests into core/primary and extra/secondary to be run in CI
I think its time now that we need to start thinking on categorizing which tests are our core or primary ones and which are secondary or extra which we care less.

By core tests means
- Unit tests, 
- Fmtcheck
- Basic Integration tests 
- Those integration tests cases which we really care and should be part of PR job as well

By non-core / extra / secondary means
- Those integration tests which we can run in nightly

I would also suggest, we can skip testing core tests during nightly as these are tested regularly during PR or master jobs.

agajdosi:
Fixes #2720 

2. Moves basic.feature and coolstore.feature into directory `stories`
3. Moves addon related features into directory `addons`
4. Adds tag `@quick` to scenarios which does not require a `minishift start` + to elementary.feature
5. Adds tag `@core` to most of the features, except the specialized ones like xpaas.feature. This tags server to PR checks
6. Adds tag `@disabled` to features proxy and coolstore. Any feature, scenario or step tagged as `disabled` will not be run.
 
  1. run `make integration` - would be quick test containing 184 scenarios and only 1 minishift start
  2. run `make integration_pr` - would run all the tests except some more specialized, for example addon-xpaas.feature containing several app deployments. Usable during PR checks.
  3. run `make integration_all` - would run all the tests. Usable during nightly tests.
  4. all test will not run any feature file tagged as `@disabled` - currently the proxy.feature and coolstore.feature. 

@minishift/minishift-dev WDYT?

### DIFF
--- a/docs/source/contributing/developing.adoc
+++ b/docs/source/contributing/developing.adoc
@@ -222,7 +222,7 @@ Features for {project} follow these basic concepts:
 
 User stories::
 Features which follow a happy path of user.
-For example, *basic.feature* or *coolstore.feature*.
+For example, the *coolstore.feature*.
 
 Feature and command coverage::
 Features which focuses on specific fields of {project} functionality or individual commands.
@@ -232,7 +232,7 @@ For example, *proxy.feature* or *cmd-version.feature*.
 ==== Running Integration Tests
 
 By default, tests are run against the binary created by `make build` (*_$GOPATH/bin/minishift_*).
-To run the basic test, use the following command:
+To run the quick test, use the following command:
 
 ----
 $ make integration
@@ -240,10 +240,11 @@ $ make integration
 
 [NOTE]
 ====
-`make integration` only runs tests which are tagged as `@basic` by default.
+By default, `make integration` only runs tests that are tagged as `@quick`.
+These tests contain multiple scenarios, but only start {project} once.
 ====
 
-To run all of the tests, use the following command:
+To run all of the tests (except those tagged as `@disabled`), use the following command:
 
 ----
 $ make integration_all
@@ -251,7 +252,9 @@ $ make integration_all
 
 [NOTE]
 ====
-There is also a `make integration_pr` target which is being run as part of pull request testing.
+There is also a `make integration_pr` target which is run as part of pull request testing.
+This target runs the test on all features tagged as `@core`, which covers the basic functionality of {project}.
+This target is a compromise between total run time and coverage.
 ====
 
 ===== Additional Parameters
@@ -321,11 +324,16 @@ Tags::
 Use `tags` to ensure that scenarios and features containing at least one of the selected tags are executed.
 To select particular feature, you can use its name as a tag.
 For example, the `cmd-config.feature` contains `@cmd-config` tag through which it can be selected and run with the following command: `make integration GODOG_OPTS=--tags=cmd-config`.
-There are also tags covering specific areas and usages of {project}:
+There are also a few special tags used to indicate specific subsets of integration tests.
+These are the following:
 
-- The `@command` tag indicates that the feature file tests commands, for example `cmd-addons.feature`.
-- The `@openshift` tag indicates that the feature file has tests related to OpenShift.
-- The `@addon` tag indicates that the feature file tests particular add-on functionality.
+- `@quick` includes `basic.feature` along with several other features and scenarios which do not require a start of {project}.
+Starting the tests with the `@quick` tag covers a variety of scenarios, but only starts {project} once.
+- `@core` includes feature files which cover all of the essential functionality of {project}.
+This tag is used to test pull requests to the {project} repository.
+- `@disabled` indicates broken feature files.
+Any failures on these features are expected.
+All features, scenarios, and steps tagged as `@disabled` will not be run as part of pull request testing or nightly build testing.
 
 Paths::
 Use `paths` to define paths to different feature files or folders containing feature files.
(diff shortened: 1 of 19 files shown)


LalatenduMohanty:
@agajdosi I do not like the name change as `elementary` . IMO basic is a better name.

amitkrout:
@agajdosi Add these tags to the doc with some explanation I mean when and where to use this. Seems contributor need to be extra cautious in future while adding new scenario/feature

budhram:
@agajdosi What all different test scenarios we have now?

agajdosi:
@LalatenduMohanty Keeping it as basic.feature now and using `@core` to tag the crucial part of tests used for PR checks. Moving to stories directory does not affect how often or how complicated it is to trigger the features. It helps to organize them better, IMO, but I am fine with skipping the stories directory.

@amitkrout I will add docs in second round, I expect some more suggestions now.

@budhram We still have integration, integration_pr and integration_all targets. The `integration` in same way as now will run `basic.feature`, however will also add those scenarios which are in features before the minishift start, so for example `make integration` will now also run some cmd-config and cmd-addons checks. These are tagged as `@quick`.

Ten `make integration_pr` will run features tagged as `@core`, it is like 90% of features. There is not xpaas addon and some other.

Lastly `make integration_all` will run all features. In all cases the test target will skip any feature/scenario/step which is tagged as `@disabled`.

LalatenduMohanty:
retest this please

agajdosi:
@LalatenduMohanty Ack. Will keep only coolstore as part of stories + add needed docs changes.

praveenkumar:
@agajdosi  After going to the PR I don't think other tags for the feature file make more sense now like @basic, @cmd-options ..etc. or you think we might need those in near future or for testing something?

agajdosi:
@praveenkumar These unique per-file tags like cmd-addons, cmd-openshift etc. are important if you want to execute just one feature file. Reason is that there is no way to specify just one file in Godog, you can just specify the whole directory.

However I fully agree that questionable are tags like `cmd`, `addon` and similar. These creates bigger sets of tests, which could be useful but I have never used it. I guess we could remove those.

praveenkumar:
> These unique per-file tags like cmd-addons, cmd-openshift etc. are important if you want to execute just one feature file. Reason is that there is no way to specify just one file in Godog, you can just specify the whole directory.

Right, that I got to know and also used recently. Very handy.

> However I fully agree that questionable are tags like cmd, addon and similar. These creates bigger sets of tests, which could be useful but I have never used it. I guess we could remove those.

Yes, that what I was pointing out, you need to decide that in same PR you can do it or in a different one.

agajdosi:
@praveenkumar It has been few days since I last saw this PR and I must say that the tags after that time looked quite complicated. I have thus removed those like `command` or `openshift`. Keeping the unique feature name tag and then `core`, `quick` and `disabled.` So it is more simple.

If one needs to target just a sub-set, it is easier to specify feature by feature, not to search what is meaning of `openshift` or `command` tags. Now going on docs and will then ping you and other folks.

agajdosi:
Changes implemented. please take a look @LalatenduMohanty, @praveenkumar. There are also some docs changes, could you please check them, @robin-owen? Thank you folks!

agajdosi:
retest this please

docs/source/contributing/developing.adoc
```diff
@@ -321,11 +324,15 @@ Tags::
 Use `tags` to ensure that scenarios and features containing at least one of the selected tags are executed.
 To select particular feature, you can use its name as a tag.
 For example, the `cmd-config.feature` contains `@cmd-config` tag through which it can be selected and run with the following command: `make integration GODOG_OPTS=--tags=cmd-config`.
-There are also tags covering specific areas and usages of {project}:
+There are also few special tags used to indicate specific subsets of integration tests. These are:
 
-- The `@command` tag indicates that the feature file tests commands, for example `cmd-addons.feature`.
-- The `@openshift` tag indicates that the feature file has tests related to OpenShift.
-- The `@addon` tag indicates that the feature file tests particular add-on functionality.
+- `@quick` which incorporates `basic.feature` and several other features and scenarios which do not need a start of {project}.
```
robin-owen:
```adoc
`@quick` includes `basic.feature` along with several other features and scenarios which do not require a start of {project}.
```

docs/source/contributing/developing.adoc
```diff
@@ -350,16 +357,16 @@ Passing any value via `GODOG_OPTS` overrides the default tag definition on each
 Thus in this case `--tags` must be specified manually, otherwise all features will be run.
 ====
 
-For example, to run integration tests on two specific feature files using only `@basic` and `@openshift` tags and without ANSI colors, the following command can be used:
+For example, to run integration tests on two specific feature files using only `@basic` and `@cmd-addons` tags and without ANSI colors, the following command can be used:
```
robin-owen: Missing "the":

```adoc
using only the `@basic` and `@cmd-addons` tags
```

docs/source/contributing/developing.adoc
```diff
@@ -321,11 +324,15 @@ Tags::
...
 To select particular feature, you can use its name as a tag.
 For example, the `cmd-config.feature` contains `@cmd-config` tag through which it can be selected and run with the following command: `make integration GODOG_OPTS=--tags=cmd-config`.
-There are also tags covering specific areas and usages of {project}:
+There are also few special tags used to indicate specific subsets of integration tests. These are:
 
-- The `@command` tag indicates that the feature file tests commands, for example `cmd-addons.feature`.
-- The `@openshift` tag indicates that the feature file has tests related to OpenShift.
-- The `@addon` tag indicates that the feature file tests particular add-on functionality.
+- `@quick` which incorporates `basic.feature` and several other features and scenarios which do not need a start of {project}.
+Starting the tests with `@quick` tag thus covers variety of scenarios while it starts {project} only once.
```
robin-owen:
```adoc
Starting the tests with the `@quick` tag covers a variety of scenarios, but only starts {project} once.
```

agajdosi:
Changes implemented, thank you @robin-owen for all the suggestions :+1:

amitkrout:
@budhram please approve if you think the pr resolves the purpose otherwise request for change

robin-owen:
Documentation looks good to me. Thanks, @agajdosi! :+1:

Issue #2749, opened 2018-08-31
CentOS CI is failing because of invalid external link


LalatenduMohanty:
We have removed the B2D  and minikube nightly [1] but the link to CI still exists in the docs

https://ci.centos.org/job/minishift-pr/4826/console

```
06:47:36 - build/contributing/ci.html
06:47:36   *  External link https://ci.centos.org/job/minishift-nightly-b2d failed: 404 No error
06:47:36 htmlproofer 3.7.2 | Error:  HTML-Proofer found 1 failure!
```
we need to fix the docs . 

[1] https://github.com/minishift/minishift-ci-jobs/commit/389ed1ae4afcf1beb12d2d9d9329631890a78413

PR #2751, opened 2018-08-31
Issue #2732 Use semver comparisons instead via util.VersionOrdinal

### LINKED ISSUE #2732: Remove util.VersionOrdinal since now downstream also use semantic versioning.
Since now downstream follow the semantic versioning so we should remove all the redundant code around this version comparison.

Mostly `VersionOrdinal` function and it's usage.

```
$ git grep -n VersionOrdinal
pkg/minishift/addon/manager/addon_manager.go:342:               if util.VersionOrdinal(currentVersion) < util.VersionOrdinal(strings.TrimPrefix(requiredVersion, ">=")) {
pkg/minishift/addon/manager/addon_manager.go:349:               if util.VersionOrdinal(currentVersion) <= util.VersionOrdinal(strings.TrimPrefix(requiredVersion, ">")) {
pkg/minishift/addon/manager/addon_manager.go:356:               if util.VersionOrdinal(currentVersion) > util.VersionOrdinal(strings.TrimPrefix(requiredVersion, "<=")) {
pkg/minishift/addon/manager/addon_manager.go:363:               if util.VersionOrdinal(currentVersion) >= util.VersionOrdinal(strings.TrimPrefix(requiredVersion, "<")) {
pkg/minishift/openshift/version/openshift_versions.go:59:               if util.VersionOrdinal(version) >= util.VersionOrdinal(minSupportedVersion) {
pkg/util/utils.go:103:func VersionOrdinal(version string) string {
pkg/util/utils.go:128:                  panic("VersionOrdinal: invalid version")
pkg/util/utils_test.go:91:func TestVersionOrdinal(t *testing.T) {
pkg/util/utils_test.go:108:             actualResult := VersionOrdinal(versionTest.MinOpenshiftVersion) >= VersionOrdinal(versionTest.OpenshiftVersion)
```

budhram:
Fixes #2732 

1. cat demoaddon.adon
```
# Name: demoaddon
# Description: Just a demo addon.
# Url: https://docs.okd.io/
# Minishift-Version: <=1.22.0

echo "Hello world"
```
Now, install and apply.
Note: Expect error here.
  
2. Update ~/.minishift/addons/demoaddon/demoaddon.addon
```
# Name: demoaddon
# Description: Just a demo addon.
# Url: https://docs.okd.io/
# Minishift-Version: >=1.22.0

echo "Hello world"
```

Note: Apply should work.

### DIFF
--- a/pkg/minishift/addon/manager/addon_manager.go
+++ b/pkg/minishift/addon/manager/addon_manager.go
@@ -25,13 +25,13 @@ import (
 
 	"strings"
 
+	"github.com/blang/semver"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/addon"
 	"github.com/minishift/minishift/pkg/minishift/addon/command"
 	"github.com/minishift/minishift/pkg/minishift/addon/config"
 	"github.com/minishift/minishift/pkg/minishift/addon/parser"
 	instanceState "github.com/minishift/minishift/pkg/minishift/config"
-	"github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/filehelper"
 	utilStrings "github.com/minishift/minishift/pkg/util/strings"
 	"github.com/minishift/minishift/pkg/version"
@@ -358,33 +358,60 @@ func setStateAndPriority(addOn addon.AddOn, configMap map[string]*config.AddOnCo
 
 func compareComponentVersions(currentVersion, requiredVersion string, componentName string, versionRange string) error {
 	if strings.HasPrefix(requiredVersion, ">=") {
-		// This will work for both upstream and downstream.
-		if util.VersionOrdinal(currentVersion) < util.VersionOrdinal(strings.TrimPrefix(requiredVersion, ">=")) {
+		currentVersion, err := semver.Make(currentVersion)
+		if err != nil {
+			return err
+		}
+		requiredVersion, err := semver.Make(strings.TrimPrefix(requiredVersion, ">="))
+		if err != nil {
+			return err
+		}
+		if currentVersion.GE(requiredVersion) != true {
 			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
 				"You need to use a version %s", componentName, currentVersion, versionRange)
 		}
-		return nil
 	}
 	if strings.HasPrefix(requiredVersion, ">") {
-		if util.VersionOrdinal(currentVersion) <= util.VersionOrdinal(strings.TrimPrefix(requiredVersion, ">")) {
+		currentVersion, err := semver.Make(currentVersion)
+		if err != nil {
+			return err
+		}
+		requiredVersion, err := semver.Make(strings.TrimPrefix(requiredVersion, ">"))
+		if err != nil {
+			return err
+		}
+		if currentVersion.GT(requiredVersion) != true {
 			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
 				"You need to use a version %s", componentName, currentVersion, versionRange)
 		}
-		return nil
 	}
 	if strings.HasPrefix(requiredVersion, "<=") {
-		if util.VersionOrdinal(currentVersion) > util.VersionOrdinal(strings.TrimPrefix(requiredVersion, "<=")) {
+		currentVersion, err := semver.Make(currentVersion)
+		if err != nil {
+			return err
+		}
+		requiredVersion, err := semver.Make(strings.TrimPrefix(requiredVersion, "<="))
+		if err != nil {
+			return err
+		}
+		if currentVersion.LE(requiredVersion) != true {
 			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
 				"You need to use a version %s", componentName, currentVersion, versionRange)
 		}
-		return nil
 	}
 	if strings.HasPrefix(requiredVersion, "<") {
-		if util.VersionOrdinal(currentVersion) >= util.VersionOrdinal(strings.TrimPrefix(requiredVersion, "<")) {
+		currentVersion, err := semver.Make(currentVersion)
+		if err != nil {
+			return err
+		}
+		requiredVersion, err := semver.Make(strings.TrimPrefix(requiredVersion, "<"))
+		if err != nil {
+			return err
+		}
+		if currentVersion.LT(requiredVersion) != true {
 			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
 				"You need to use a version %s", componentName, currentVersion, versionRange)
 		}
-		return nil
 	}
 	if currentVersion != requiredVersion {
 		return fmt.Errorf("\nAddon does not support %s version %s. "+
--- a/pkg/minishift/openshift/version/openshift_versions.go
+++ b/pkg/minishift/openshift/version/openshift_versions.go
@@ -18,13 +18,11 @@ package version
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"regexp"
-	"sort"
 	"strings"
 
 	"github.com/blang/semver"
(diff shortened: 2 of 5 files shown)


budhram:
Also, fixes #2749 but will merge https://github.com/minishift/minishift/pull/2752

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -358,29 +358,36 @@ func setStateAndPriority(addOn addon.AddOn, configMap map[string]*config.AddOnCo
 
 func compareComponentVersions(currentVersion, requiredVersion string, componentName string, versionRange string) error {
 	if strings.HasPrefix(requiredVersion, ">=") {
-		// This will work for both upstream and downstream.
-		if util.VersionOrdinal(currentVersion) < util.VersionOrdinal(strings.TrimPrefix(requiredVersion, ">=")) {
+		currentVersion, _ := semver.Make(currentVersion)
+		requiredVersion, _ := semver.Make(strings.TrimPrefix(requiredVersion, ">="))
+		if currentVersion.LT(requiredVersion) {
```
agajdosi: Maybe it would be more readable if the if-statement was not to catch error, but to catch valid case, so for greater or equal, it would use currentVersion.GE():

```
	if strings.HasPrefix(requiredVersion, ">=") {
		currentVersion, _ := semver.Make(currentVersion)
		requiredVersion, _ := semver.Make(strings.TrimPrefix(requiredVersion, ">="))
		if currentVersion.GE(requiredVersion) {
			return nil
		}
		return fmt.Errorf("\nAdd-on does not support %s version %s. "+
			"You need to use a version %s", componentName, currentVersion, versionRange)
}
```
or
```
if currentVersion.GE(requiredVersion) != true {
	return fmt.Errorf("\nAdd-on does not support %s version %s. "+
		"You need to use a version %s", componentName, currentVersion, versionRange)
}
```

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -358,29 +358,36 @@ func setStateAndPriority(addOn addon.AddOn, configMap map[string]*config.AddOnCo
...
 			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
 				"You need to use a version %s", componentName, currentVersion, versionRange)
 		}
 		return nil
 	}
 	if strings.HasPrefix(requiredVersion, ">") {
-		if util.VersionOrdinal(currentVersion) <= util.VersionOrdinal(strings.TrimPrefix(requiredVersion, ">")) {
+		currentVersion, _ := semver.Make(currentVersion)
+		requiredVersion, _ := semver.Make(strings.TrimPrefix(requiredVersion, ">"))
+		if currentVersion.LE(requiredVersion) {
```
LalatenduMohanty: 👍

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -358,29 +358,36 @@ func setStateAndPriority(addOn addon.AddOn, configMap map[string]*config.AddOnCo
 
 func compareComponentVersions(currentVersion, requiredVersion string, componentName string, versionRange string) error {
 	if strings.HasPrefix(requiredVersion, ">=") {
-		// This will work for both upstream and downstream.
-		if util.VersionOrdinal(currentVersion) < util.VersionOrdinal(strings.TrimPrefix(requiredVersion, ">=")) {
+		currentVersion, _ := semver.Make(currentVersion)
```
praveenkumar: Is there any reason we are not handling the error here or this is handled somewhere else already?

pkg/minishift/openshift/version/openshift_versions.go
```diff
@@ -42,38 +39,6 @@ func GetOpenshiftVersion(sshCommander provision.SSHCommander) (string, error) {
 	return dockerCommander.Exec(" ", minishiftConstants.OpenshiftContainerName, "openshift", "version")
 }
 
-func PrintDownStreamVersions(output io.Writer, minSupportedVersion string) error {
```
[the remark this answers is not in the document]
praveenkumar: :+1:

pkg/minishift/openshift/version/openshift_versions_test.go
```diff
@@ -59,28 +59,6 @@ func EnsureGitHubApiAccessTokenSet(t *testing.T) {
...
-	assert.NoError(t, err)
-	_, err = f.Seek(0, 0)
-	assert.NoError(t, err, "Error setting offset back")
-
-	data, err := ioutil.ReadAll(f)
-	assert.NoError(t, err, "Error reading file")
-
-	actualStdout := string(data)
-	assert.Contains(t, actualStdout, "v3.4.1.10")
-}
```
praveenkumar: :+1:

pkg/minishift/addon/manager/addon_manager.go
budhram: @praveenkumar Here is how I see such least occurrence error. :smile: 
`semver.Make(currentVersion)` is stable enough and even if we get any random string in `currentVersion` or in `requiredVersion`, the equality is definitely going to fail.

I am omitting error checking for the sake of simplicity/readability of code.
Else, three more lines will get added which will break current degree of simplicity/readability.

Hope you understand :) 
Let me know if you have different opinion.

budhram:
@minishift/minishift-dev Pushed the latest changes.

pkg/minishift/addon/manager/addon_manager.go
LalatenduMohanty: `Make func(s string) (Version, error)`  We should handle all errors, because when it comes it is way easier to track it down. So I agree with @praveenkumar to catch the error and handle it.

praveenkumar: >  Here is how I see such least occurrence error.
semver.Make(currentVersion) is stable enough and even if we get any random string in currentVersion or in requiredVersion, the equality is definitely going to fail.

@budhram what makes `semver.Make` stable enough is returning an error and version parsed value. Now if you put arbitrary version value to parse in `Make` then you don't have any `currentVersion` or `requiredVersion` values to compare and it will go through which is not a definite fail but an unexpected behavior which will be hard to debug.

> I am omitting error checking for the sake of simplicity/readability of code.

You also think about the maintainability of the code.

> Let me know if you have different opinion.

Since version check already handled for required version during metadata parse and for current version in the start time so I am OK here to ignore the error.

pkg/minishift/openshift/version/openshift_versions.go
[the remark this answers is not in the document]
LalatenduMohanty: @budhram  I am confused about this. Between this change and CDK-336 , how will downstream going to print the downstream versions (after downstream rebases on this PR)?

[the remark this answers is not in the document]
budhram: Simple answer is, we don't need such code in upstream, we will implement `PrintDownstreamVesions` method in downstream only. Exactly this code. And now, there is single commit approach as introduced by @praveenkumar , it is quite easy.

pkg/minishift/addon/manager/addon_manager.go
budhram: > so I am OK here to ignore the error.

:+1:

LalatenduMohanty: > Now if you put arbitrary version value to parse in Make then you don't have any currentVersion or requiredVersion values to compare and it will go through which is not a definite fail but an unexpected behavior which will be hard to debug.

Thats the exact reason we should catch the error and handle it. 


> Since version check already handled for required version during metadata parse and for current version in the start time so I am OK here to ignore the error.

As a general we can not guarantee how  code changes will happen in future. So if because of code change the check gets compromised the wrong version will fall through.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -358,33 +358,60 @@ func setStateAndPriority(addOn addon.AddOn, configMap map[string]*config.AddOnCo
...
 	}
 	if strings.HasPrefix(requiredVersion, ">") {
-		if util.VersionOrdinal(currentVersion) <= util.VersionOrdinal(strings.TrimPrefix(requiredVersion, ">")) {
+		currentVersion, err := semver.Make(currentVersion)
+		if err != nil {
+			return err
+		}
+		requiredVersion, err := semver.Make(strings.TrimPrefix(requiredVersion, ">"))
+		if err != nil {
+			return err
```
praveenkumar: :+1:

budhram:
Merged as CI passed and got two approvals as it is blocking few other PRs.

PR #2770, opened 2018-09-06
Issue #2769 Integration test timeout need to be revised for nightly

### LINKED ISSUE #2769: Integration test timeout need to be revised for nightly
Current timeout setup ```10800s``` for nightly test is not enough.

Failure log - https://ci.centos.org/job/minishift-nightly-centos/270/consoleFull

amitkrout:
Fixes #2769 

  1. make integration_all
  2. 
  3. 
  4.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -213,7 +213,7 @@ integration: $(MINISHIFT_BINARY)
 integration_all: GODOG_OPTS = --tags=~coolstore\&\&~proxy
 integration_all: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
-	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
+	go test -timeout 18000s $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
 	--run-before-feature="$(RUN_BEFORE_FEATURE)" --test-with-specified-shell="$(TEST_WITH_SPECIFIED_SHELL)" --copy-oc-from="$(COPY_OC_FROM)" $(GODOG_OPTS)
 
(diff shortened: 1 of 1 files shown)


amitkrout:
As usual i have increased the test timeout by ```1hr```, but i would suggest lets increase it to ```5hrs``` and forget the time out issue for atleast 5 releases or so on 

@minishift/minishift-dev What do you guys think ?

Lets fix it quickly before the next nightly fails due the same issue

praveenkumar:
@amitkrout do you think that adding 1 hour will fix it if not then better to have at least 4-5 hour of timeout.

agajdosi:
Agree, lets add 2 hours here and forgot about the timeout for a while. Another reason is that Che tests are coming and htpasswd-identity-provider will be here also soon. Lets have 5hrs aka 18000s.

agajdosi:
Also I would suggest to not change global `TIMEOUT` value but define the time directly in `make integration_all` target. I would be a little afraid of having some PR tests stuck and hanging around for 5 hours, lets keep them limited on current 3 hours.

amitkrout:
@praveenkumar @agajdosi will increase the timeout to 5 hrs. @agajdosi your concern is right, unfortunate stuck in integration_pr will unnecessarily wait till 5hrs. So will directly put the timeout value in integration_all without changing the change global timeout

Makefile
```diff
@@ -213,7 +213,7 @@ integration: $(MINISHIFT_BINARY)
 integration_all: GODOG_OPTS = --tags=~coolstore\&\&~proxy
 integration_all: $(MINISHIFT_BINARY)
 	mkdir -p $(INTEGRATION_TEST_DIR)
-	go test -timeout $(TIMEOUT) $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
+	go test -timeout 18000s $(REPOPATH)/test/integration --tags=integration -v -args --test-dir $(INTEGRATION_TEST_DIR) --binary $(MINISHIFT_BINARY) \
```
praveenkumar: This is 5 hr of timeout for nightly 👍

PR #2772, opened 2018-09-06
Issue #2768 Use origin-cli image if openshift >= 3.10.x

### LINKED ISSUE #2768: Update image name from origin to origin-control-plane when using openshift >= 3.10.x
As of now we are pulling the `openshift/origin` image and then copying the `oc` binary in the VM which used for `oc cluster up`. With 3.10 the image rename happened and the image `origin` changed to `origin-control-plane` which have all the updated bits.

[0] https://hub.docker.com/r/openshift/origin-control-plane/tags/

praveenkumar:
Fixes #2768 

  1. minishift start
  2. docker images => this contain origin-cli image instead origin one for v3.10 and origin for v3.9

### DIFF
--- a/cmd/minishift/cmd/util/oc.go
+++ b/cmd/minishift/cmd/util/oc.go
@@ -35,6 +35,7 @@ import (
 	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
 	"github.com/minishift/minishift/pkg/minishift/docker"
 	"github.com/minishift/minishift/pkg/minishift/oc"
+	openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"
 	profileActions "github.com/minishift/minishift/pkg/minishift/profile"
 	utils "github.com/minishift/minishift/pkg/util"
 	"github.com/minishift/minishift/pkg/util/os/atexit"
@@ -150,14 +151,15 @@ func getOcPathForProfile() (error, string) {
 // PullOpenshiftImageAndCopyOcBinary pull the openshift image if not available and then copy the client binary to Remote/VM machine
 // Any occurring error is also returned.
 func PullOpenshiftImageAndCopyOcBinary(dockerCommander docker.DockerCommander, requestedOpenShiftVersion string) error {
+	isGreaterOrEqualToBaseVersion, _ := openshiftVersion.IsGreaterOrEqualToBaseVersion(requestedOpenShiftVersion, constants.RefactoredOcVersion)
 
 	// We need to make sure if images are already exist from the cache then don't pull it again.
-	imageExist, err := dockerCommander.IsImageExist(minishiftConstants.GetOpenshiftImageName(requestedOpenShiftVersion))
+	imageExist, err := dockerCommander.IsImageExist(minishiftConstants.GetOpenshiftImageToFetchOC(requestedOpenShiftVersion, isGreaterOrEqualToBaseVersion))
 	if !imageExist {
 		fmt.Printf("-- Pulling the Openshift Container Image ")
 		progressDots := progressdots.New()
 		progressDots.Start()
-		_, err = dockerCommander.Pull(minishiftConstants.GetOpenshiftImageName(requestedOpenShiftVersion))
+		_, err = dockerCommander.Pull(minishiftConstants.GetOpenshiftImageToFetchOC(requestedOpenShiftVersion, isGreaterOrEqualToBaseVersion))
 		if err != nil {
 			return fmt.Errorf("Error pulling the openshift container image: %v", err)
 		}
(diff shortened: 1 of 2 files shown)


pkg/minishift/constants/constants.go
```diff
@@ -52,7 +52,10 @@ func ProfilePrivateKeyPath() string {
 	return filepath.Join(constants.Minipath, "certs", "id_rsa")
 }
 
-func GetOpenshiftImageName(openshiftVersion string) string {
+func GetOpenshiftImageName(openshiftVersion string, isGreaterOrEqualToBaseVersion bool) string {
```
praveenkumar: I had to change the function signature if I try to import `openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"` and consume `openshiftVersion.IsGreaterOrEqualToBaseVersion(openShiftVersion, constants.RefactoredOcVersion)` then it become a cyclic dependency and build fail.

cmd/minishift/cmd/start.go
```diff
@@ -265,7 +265,8 @@ func runStart(cmd *cobra.Command, args []string) {
 		fmt.Println("-- OpenShift cluster will be configured with ...")
 		fmt.Println("   Version:", requestedOpenShiftVersion)
 
-		err := cmdUtil.PullOpenshiftImageAndCopyOcBinary(dockerCommander, requestedOpenShiftVersion)
+		valid, _ := openshiftVersion.IsGreaterOrEqualToBaseVersion(requestedOpenShiftVersion, constants.RefactoredOcVersion)
+		err := cmdUtil.PullOpenshiftImageAndCopyOcBinary(dockerCommander, requestedOpenShiftVersion, valid)
```
LalatenduMohanty: @praveenkumar IMO we should move the logic of deciding `isGreaterOrEqualToBaseVersion` inside of `PullOpenshiftImageAndCopyOcBinary`

cmd/minishift/cmd/util/oc.go
```diff
@@ -149,15 +149,15 @@ func getOcPathForProfile() (error, string) {
 
 // PullOpenshiftImageAndCopyOcBinary pull the openshift image if not available and then copy the client binary to Remote/VM machine
 // Any occurring error is also returned.
-func PullOpenshiftImageAndCopyOcBinary(dockerCommander docker.DockerCommander, requestedOpenShiftVersion string) error {
+func PullOpenshiftImageAndCopyOcBinary(dockerCommander docker.DockerCommander, requestedOpenShiftVersion string, isGreaterOrEqualToBaseVersion bool) error {
```
LalatenduMohanty: No need to pass the extra variable, we can decide in the function itself. This way it will more logical IMO.

pkg/minishift/constants/constants.go
gbraad: is this ONLY for the `oc` command or general usage? I am confused about the function name in case of this change...

since we only use this image to copy `oc` out or is it more?

praveenkumar: @gbraad yes this is only for copy oc command nothing more.

cmd/minishift/cmd/start.go
praveenkumar: Changing a function signature is completely OK since this is not something changes to UX side. I don't have any strong option about this to be part of `PullOpenshiftImageAndCopyOcBinary` method or here.

pkg/minishift/constants/constants.go
gbraad: in that case it is better to change the funcgtion name to reflect this better

gbraad: or if this function is used in more place, it is better ot break this out and specify that a choice is needed for the pull of the container for the `oc` command

praveenkumar: @gbraad It only used one place, does `GetOpenshiftCliImage` sound better?

cmd/minishift/cmd/start.go
gbraad: it would make it consolidated

pkg/minishift/constants/constants.go
gbraad: makes more sense

praveenkumar:
retest this please.

praveenkumar:
restest this please.

pkg/minishift/constants/constants.go
```diff
@@ -53,7 +53,10 @@ func ProfilePrivateKeyPath() string {
 	return filepath.Join(constants.Minipath, "certs", "id_rsa")
 }
 
-func GetOpenshiftImageName(openshiftVersion string) string {
+func GetOpenshiftImageToFetchOC(openshiftVersion string, isGreaterOrEqualToBaseVersion bool) string {
+	if isGreaterOrEqualToBaseVersion {
+		return fmt.Sprintf("openshift/origin-control-plane:%s", openshiftVersion)
```
LalatenduMohanty: @praveenkumar I thought we are going to use `origin-cli image`?

praveenkumar: @LalatenduMohanty https://github.com/openshift/origin/issues/20892 but `origin-control-plane` is always updated and used for initial openshift cluster start.

PR #2776, opened 2018-09-07
Issue #2767 implement openshift add command to add component in running cluster

### LINKED ISSUE #2767: Create an feature equivalent of `oc cluster add` functionalility
oc 3.10.0 have a capability to add a component to running cluster using `oc cluster add`. From the minishift point of view, we don't have any such capability and this will be there for 3.11 also. I propose to put an `add` flag behind the experiment in the meantime because from 3.12 (4.0) this might not be available or change completely but the minishift user can consume it till 3.11 release.

```
$ oc cluster -h
[...]
Usage:
  oc cluster ACTION [flags]

Available Commands:
  add         Add components to an 'oc cluster up' cluster
  down        Stop OpenShift on Docker
  status      Show OpenShift on Docker status
  up          Start OpenShift on Docker with reasonable defaults
```

```
$ export MINISHIFT_ENABLE_EXPERIMENTAL=y
$ minishift openshift add <component_name>
```

praveenkumar:
Fixes #2767

  1. minishift start
  2. minishift openshift component add service-catalog => this should success
  3. minishift openshift componet add foobar => this should fail and you will see the component not available and failure
  4. minishift start --openshift-version v3.9.0 => minishift openshift component add service-catalog => this should fail because it not supported and provide the respective error.
 5. minishift openshift component list => this will show valid component list.

### DIFF
--- a/cmd/minishift/cmd/openshift/component.go
+++ b/cmd/minishift/cmd/openshift/component.go
@@ -1,0 +1,34 @@
+/*
+Copyright (C) 2018 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openshift
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var componentCmd = &cobra.Command{
+	Use:   "component [command]",
+	Short: "Interacts with the Components to an OpenShift cluster (Only works openshift version >= 3.10.x)",
+	Long:  `Interacts with the Components to an OpenShift cluster (Only works openshift version >= 3.10.x)`,
+	Run: func(cmd *cobra.Command, args []string) {
+		cmd.Help()
+	},
+}
+
+func init() {
+	OpenShiftCmd.AddCommand(componentCmd)
+}
--- a/cmd/minishift/cmd/openshift/component_list.go
+++ b/cmd/minishift/cmd/openshift/component_list.go
@@ -1,0 +1,43 @@
+/*
+Copyright (C) 2018 Red Hat, Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openshift
+
+import (
+	"fmt"
+	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+// version command represent current running openshift version and available one.
+var componentListCmd = &cobra.Command{
+	Use:   "list [component-name]",
+	Short: "Add component to an OpenShift cluster (Only works openshift version >= 3.10.x)",
+	Long:  `Add component to an OpenShift cluster (Only works openshift version >= 3.10.x)`,
+	Run:   runComponentList,
+}
+
+func runComponentList(cmd *cobra.Command, args []string) {
+	fmt.Fprint(os.Stdout, "The following OpenShift components are available: \n")
+	for _, component := range minishiftConstants.ValidComponents {
+		fmt.Fprintf(os.Stdout, "\t- %s\n", component)
+	}
+}
+
+func init() {
+	componentCmd.AddCommand(componentListCmd)
+}
--- a/pkg/minishift/constants/constants.go
+++ b/pkg/minishift/constants/constants.go
@@ -37,10 +37,12 @@ const (
 	SkipVerifyInsecureTLS       = "insecure-skip-tls-verify=true"
 	BinaryName                  = "minishift"
 	OcPathInsideVM              = "/var/lib/minishift/bin"
+	BaseDirInsideInstance       = "/var/lib/minishift/base"
 )
 
 var (
 	ValidIsoAliases = []string{B2dIsoAlias, CentOsIsoAlias}
+	ValidComponents = []string{"automation-service-broker", "service-catalog", "template-service-broker"}
 )
 
 // ProfileAuthorizedKeysPath returns the path of authorized_keys file in profile dir used for authentication purpose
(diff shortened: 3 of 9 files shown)


cmd/minishift/cmd/openshift/add.go
```diff
@@ -0,0 +1,94 @@
...
+	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
+	openshiftVersion "github.com/minishift/minishift/pkg/minishift/openshift/version"
+	"github.com/minishift/minishift/pkg/util/os/atexit"
+	"github.com/spf13/cobra"
+	"os"
+)
+
+const (
+	componentFlag             = "component"
+	nonExistentComponentError = "You need to specify a component name using -c <component_name>"
```
gbraad: not sure why we would need the `-c` option to specify a component

LalatenduMohanty: The error message looks confusing because instead of saying `you are trying to add a non existent component` we are saying  `You need to specify a component name using -c <component_name>`

praveenkumar:
retest this please.

cmd/minishift/cmd/openshift/add.go
```diff
@@ -0,0 +1,94 @@
...
+
+const (
+	componentFlag              = "component"
+	nonSpecifiedComponentError = "You need to specify a component name using --component <component_name>"
+)
+
+// version command represent current running openshift version and available one.
+var addCmd = &cobra.Command{
+	Use:   "add [command] [flags]",
+	Short: "Add components to an OpenShift cluster (Only works openshift version >= 3.10.x",
```
LalatenduMohanty: Missing parenthesis `(Only works openshift version >= 3.10.x)`.

cmd/minishift/cmd/openshift/add.go
```diff
@@ -0,0 +1,94 @@
...
+	Long:  `Add components to an OpenShift cluster (Only works openshift version >= 3.10.x`,
```
LalatenduMohanty: Missing parenthesis `(Only works openshift version >= 3.10.x)`.

LalatenduMohanty: @praveenkumar How will user know what are the available components for `minishift openshift add`.

cmd/minishift/cmd/openshift/add.go
```diff
@@ -0,0 +1,94 @@
...
+const (
+	componentFlag              = "component"
+	nonSpecifiedComponentError = "You need to specify a component name using --component <component_name>"
+)
+
+// version command represent current running openshift version and available one.
+var addCmd = &cobra.Command{
+	Use:   "add [command] [flags]",
+	Short: "Add components to an OpenShift cluster (Only works openshift version >= 3.10.x)",
+	Long:  `Add components to an OpenShift cluster (Only works openshift version >= 3.10.x)`,
```
LalatenduMohanty: @praveenkumar There is no documentation or hints around what components can be added. So we should add documentation about `valid component names are: "automation-service-broker,centos-imagestreams,persistent-volumes,registry,rhel-imagestreams,router,sample-templates,service-catalog,template-service-broker,web-console"`

LalatenduMohanty: The only downside of the documentation is that `if cluster up changes the componenets we need to update the docs`.

pkg/minishift/constants/constants.go
```diff
@@ -37,10 +37,12 @@ const (
 	SkipVerifyInsecureTLS       = "insecure-skip-tls-verify=true"
 	BinaryName                  = "minishift"
 	OcPathInsideVM              = "/var/lib/minishift/bin"
+	BaseDirInsideInstance       = "/var/lib/minishift/base"
 )
 
 var (
 	ValidIsoAliases = []string{B2dIsoAlias, CentOsIsoAlias}
+	ValidComponents = []string{"automation-service-broker", "service-catalog", "template-service-broker"}
```
praveenkumar: Instead of parsing I put the valid component here because it may happen user just download the minishift and then list the component and if oc binary not available then you never know the what components are vaild, also for `m openshift component list` having oc available shouldn't be required. I put simple documentations so that when there is any component added/removed we just need to change here instead in documents.

praveenkumar:
retest this please.

cmd/minishift/cmd/openshift/component_list.go
```diff
@@ -0,0 +1,43 @@
...
+	Use:   "list [component-name]",
+	Short: "Add component to an OpenShift cluster (Only works openshift version >= 3.10.x)",
+	Long:  `Add component to an OpenShift cluster (Only works openshift version >= 3.10.x)`,
+	Run:   runComponentList,
+}
+
+func runComponentList(cmd *cobra.Command, args []string) {
+	fmt.Fprint(os.Stdout, "The following OpenShift components are available: \n")
+	for _, component := range minishiftConstants.ValidComponents {
+		fmt.Fprintf(os.Stdout, "\t- %s\n", component)
```
LalatenduMohanty: We need to repeat that these are OpenShift components as the command line says it explicitly. Suggestion:

```diff
func runComponentList(cmd *cobra.Command, args []string) {
-       fmt.Fprint(os.Stdout, "The following OpenShift components are available: \n")
+       fmt.Fprint(os.Stdout, "Available components: \n")
        for _, component := range minishiftConstants.ValidComponents {
-               fmt.Fprintf(os.Stdout, "\t- %s\n", component)
+               fmt.Fprintf(os.Stdout, "- %s\n", component)
        }
 }
```
Output
```
$ minishift openshift component list
Available components:
- automation-service-broker
- service-catalog
- template-service-broker
```

praveenkumar: @LalatenduMohanty I did it because we want to be consistent with `minishift openshift version list`, don't want to break this.

```
$ ./minishift openshift version list
	- v3.9.0
	- v3.10.0
```

cmd/minishift/cmd/openshift/component.go
```diff
@@ -0,0 +1,34 @@
...
+
+package openshift
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var componentCmd = &cobra.Command{
+	Use:   "component [command]",
+	Short: "Interacts with the Components to an OpenShift cluster (Only works openshift version >= 3.10.x)",
```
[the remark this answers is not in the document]
praveenkumar: @LalatenduMohanty This can be done during the merge.

cmd/minishift/cmd/openshift/component_list.go
LalatenduMohanty: @praveenkumar I still think we should change it to `fmt.Fprint(os.Stdout, "Available components: \n")`. I feel the current output print a long statement which is not required.

cmd/minishift/cmd/openshift/component.go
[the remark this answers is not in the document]
LalatenduMohanty: Sure.

cmd/minishift/cmd/openshift/component_list.go
LalatenduMohanty: I can fix this during the merge too.

praveenkumar: @LalatenduMohanty one way we are thinking to have same kind of output for different commands and then we are changing here. I don't have any strong option here but doesn't seem right.

LalatenduMohanty:
Merged with https://github.com/minishift/minishift/commit/1b29e07222d6e5a191940f2a030196e8c04c2da4

PR #2779, opened 2018-09-07
Issue #2778 Fix openshift version metadata info for addon

### LINKED ISSUE #2778: OpenShift-Version feature of addons is broken
### General information

  * Minishift version: current master (minishift v1.23.0+fdba8a84)
  * OS: Linux
  * Hypervisor: KVM

### Steps to reproduce

  1. minishift start
  2. minishift addon apply che

### Expected

Addon is applied:
```
Then you should be able the open the Che dashboard here:
http://che-#{NAMESPACE}.#{ip}.nip.io
```

### Actual
```
[agajdosi@dhcp-24-156 ~]$ minishift addons apply che
-- Applying addon 'che':Error applying the add-on: Invalid character(s) found in major number "=3"
```

praveenkumar:
Fixes #2778 

  1. $ minishift start
  2. $ minishift apply che

### DIFF
--- a/addons/admin-user/admin-user.addon
+++ b/addons/admin-user/admin-user.addon
@@ -1,5 +1,7 @@
 # Name: admin-user
 # Description: Create admin user and assign the cluster-admin role to it.
+# OpenShift-Version: >=3.7.0
+# Minishift-Version: >=1.11.0
 
 !oc create user admin --full-name=admin
 oc adm policy add-cluster-role-to-user cluster-admin admin
--- a/pkg/minishift/addon/manager/addon_manager.go
+++ b/pkg/minishift/addon/manager/addon_manager.go
@@ -370,6 +370,7 @@ func compareComponentVersions(currentVersion, requiredVersion string, componentN
 			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
 				"You need to use a version %s", componentName, currentVersion, versionRange)
 		}
+		return err
 	}
 	if strings.HasPrefix(requiredVersion, ">") {
 		currentVersion, err := semver.Make(currentVersion)
@@ -384,6 +385,7 @@ func compareComponentVersions(currentVersion, requiredVersion string, componentN
 			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
 				"You need to use a version %s", componentName, currentVersion, versionRange)
 		}
+		return err
 	}
 	if strings.HasPrefix(requiredVersion, "<=") {
 		currentVersion, err := semver.Make(currentVersion)
@@ -398,6 +400,7 @@ func compareComponentVersions(currentVersion, requiredVersion string, componentN
 			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
 				"You need to use a version %s", componentName, currentVersion, versionRange)
 		}
+		return err
 	}
 	if strings.HasPrefix(requiredVersion, "<") {
 		currentVersion, err := semver.Make(currentVersion)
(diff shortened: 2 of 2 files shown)


addons/admin-user/admin-user.addon
```diff
@@ -1,5 +1,7 @@
 # Name: admin-user
 # Description: Create admin user and assign the cluster-admin role to it.
+# OpenShift-Version: >=3.7.0
+# Minishift-Version: >=1.11.0
```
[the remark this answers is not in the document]
amitkrout: @praveenkumar Question - Any specific reason you are adding the combination of ```3.7.0``` and ```1.11.0``` as lower boundary, why not version above to this ?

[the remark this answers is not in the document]
amitkrout: Or its only for integration test propose and nothing else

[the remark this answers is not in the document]
praveenkumar: @amitkrout this addon worked throughout the release and this is just to make sure integration test works as expected.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -370,6 +370,7 @@ func compareComponentVersions(currentVersion, requiredVersion string, componentN
 			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
 				"You need to use a version %s", componentName, currentVersion, versionRange)
 		}
+		return err
```
LalatenduMohanty: @praveenkumar This should be `return nil`. We have handled all errors before isn't it.

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -384,6 +385,7 @@ func compareComponentVersions(currentVersion, requiredVersion string, componentN
...
```
LalatenduMohanty: same as above

pkg/minishift/addon/manager/addon_manager.go
```diff
@@ -398,6 +400,7 @@ func compareComponentVersions(currentVersion, requiredVersion string, componentN
...
```
LalatenduMohanty: same as above.

praveenkumar:
On Fri 7 Sep, 2018, 7:21 PM Lalatendu Mohanty, <notifications@github.com>
wrote:

> *@LalatenduMohanty* requested changes on this pull request.
> ------------------------------
>
> In pkg/minishift/addon/manager/addon_manager.go
> <https://github.com/minishift/minishift/pull/2779#discussion_r215964944>:
>
> > @@ -370,6 +370,7 @@ func compareComponentVersions(currentVersion, requiredVersion string, componentN
>  			return fmt.Errorf("\nAdd-on does not support %s version %s. "+
>  				"You need to use a version %s", componentName, currentVersion, versionRange)
>  		}
> +		return err
>
> @praveenkumar <https://github.com/praveenkumar> This should be return nil.
> We have handled all errors before isn't it.
>

@*LalatenduMohanty* If you look closely then 'err' at the end is nil so
that's expected.

------------------------------

pkg/minishift/addon/manager/addon_manager.go
gbraad: This is consistent with the rest of the file... plus it is a good practice, as you might want to handle the error in the caller.

LalatenduMohanty: I think returning `nil` when it is not an error is a good programing practice because "explict is better than implicit".  Also I see that we have returned nil in other function in the same file.

LalatenduMohanty: Discussed with @praveenkumar and it is fine.

PR #2783, opened 2018-09-10
Issue #2782 Fix xpaas addon echo command for integration test

### LINKED ISSUE #2782: Nightly failed because of XPaaS addon's echo command.
We made changes to xpaas addon echo command from `XPaaS OpenShift imagestream and templates installed` to `XPaaS imagestream and templates for OpenShift 3.10 installed` and nightly failed because of it https://ci.centos.org/job/minishift-nightly-centos/274/console. I would say that instead of include version info in the echo we should simply put `XPaaS imagestream and templates for OpenShift installed` so we don't need to change the test senario everytime.

praveenkumar:
Fixes #2782


gbraad:
retest this please

PR #2800, opened 2018-09-14
Issue #2792 Add save-start-flags bool option to config

### LINKED ISSUE #2792: Workaround required not to save start flags to persistent config
Recently we have add a feature to save non default flags in `minishift start`[1] to persistent config. 
I am sure many users will like it. However it is really annoying when I am trying out various configs with Minishift. I have to run `minishift config unset ` many times to try a different settings as settings gets saved in to minishift config everytime with `minishift start`. 

So I think we need a way to tell Minishift not to save non default flags to persistent config.  It will be the non default behavior  but user explicitly want it then we should be able to change minishift behaviour e.g. `minishift config set save-to-persistent-config false`.  

[1] https://github.com/minishift/minishift/issues/2611

praveenkumar:
Fixes #2792

  1. ./minishift start --cpus 3 --memory 6GB
  2. ./minishift config view => this will contain cpus and memory value
  3. ./minishift config set save-start-flags false
  4. ./minishift unset cpus
  5. ./minishift unset memory
  6. ./minishift start --cpus 3 --memory 6GB
  7. ./minishift config view => this will now not keep record of start flags.

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -190,7 +190,10 @@ func runStart(cmd *cobra.Command, args []string) {
 		preflightChecksBeforeStartingHost()
 	}
 
-	populateStartFlagsToViperConfig()
+	// Populate start flags to viper config if save-start-flags true in config file
+	if viper.GetBool(configCmd.SaveStartFlags.Name) {
+		populateStartFlagsToViperConfig()
+	}
 
 	// Cache OC binary before starting the VM and perform oc command option check
 	ocPath = cmdUtil.CacheOc(requestedOpenShiftVersion)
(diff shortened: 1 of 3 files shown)


cmd/minishift/cmd/start.go
```diff
@@ -190,7 +190,10 @@ func runStart(cmd *cobra.Command, args []string) {
 		preflightChecksBeforeStartingHost()
 	}
 
-	populateStartFlagsToViperConfig()
+	// Populate start flags to viper config if save-start-flags true in config file
+	if viper.GetBool(configCmd.SaveStartFlags.Name) {
```
agajdosi: :+1:

LalatenduMohanty:
The code works fine for me.

PR #2816, opened 2018-09-20
Issue #2775 Add document for htpasswd-identity-provider addon

### LINKED ISSUE #2775: Add htpasswd-identity-provider information in generic driver documentation.
We do not mention about the `htpasswd-identity-provider` provider in https://docs.okd.io/latest/minishift/using/run-against-an-existing-machine.html#running-against-existing-remote-machine.

praveenkumar:
Fixes #2775

### DIFF
--- a/docs/source/using/run-against-an-existing-machine.adoc
+++ b/docs/source/using/run-against-an-existing-machine.adoc
@@ -84,9 +84,17 @@ Use the following command on the host to run {project} against a remote machine:
 
 ----
 $ minishift start --vm-driver generic --remote-ipaddress <remote_IP_address> --remote-ssh-user <username> --remote-ssh-key <private_ssh_key>
+$ minishift addon apply htpasswd-identity-provider --addon-env USER_PASSWORD=<NEW_PASSWORD>
 ----
 
 [NOTE]
 ====
 The value of the `--remote-ssh-key` flag must be the location of a private SSH key on the host machine.
+====
+
+[NOTE]
+====
+To change the default username and password, you must apply the `htpasswd-identity-provider` add-on with the desired username and password after running the `minishift start` command.
+By default, {{project}} uses the `Allow All` identity provider.
+The `Allow All` identity provider allows any non-empty username and password to log in.
 ====


praveenkumar:
retest this please.

docs/source/using/run-against-an-existing-machine.adoc
```diff
@@ -84,9 +84,16 @@ Use the following command on the host to run {project} against a remote machine:
 
 ----
 $ minishift start --vm-driver generic --remote-ipaddress <remote_IP_address> --remote-ssh-user <username> --remote-ssh-key <private_ssh_key>
+$ minishift addon apply htpasswd-identity-provider
```
LalatenduMohanty: @praveenkumar it should be `$ minishift addon apply htpasswd-identity-provider --addon-env  USER_PASSWORD=<NEW_PASSWORD>`

docs/source/using/run-against-an-existing-machine.adoc
```diff
@@ -84,9 +84,16 @@ Use the following command on the host to run {project} against a remote machine:
...
 ----
 
 [NOTE]
 ====
 The value of the `--remote-ssh-key` flag must be the location of a private SSH key on the host machine.
 ====
+
+[NOTE]
+====
+Make sure you apply `htpasswd-identity-provider` add-on after minishift start, it will allow you to use htpasswd as identity provider.
```
LalatenduMohanty: `as the identity provider`.

LalatenduMohanty: To change the default username/password, you must apply `htpasswd-identity-provider` add-on with the new username/password after minishift start.

docs/source/using/run-against-an-existing-machine.adoc
```diff
@@ -84,9 +84,16 @@ Use the following command on the host to run {project} against a remote machine:
...
+By default minishift uses `Allow All` identity provider, which allows any non-empty user name and password to log in.
```
LalatenduMohanty: The add-on `htpasswd-identity-provide` will allow you to use htpasswd as identity provider instead of `Allow All` identity provider.

docs/source/using/run-against-an-existing-machine.adoc
praveenkumar: @LalatenduMohanty by default if the user doesn't provide the password then it will take `developer` but I think it good to put that in document.

docs/source/using/run-against-an-existing-machine.adoc
praveenkumar: cc @robin-owen please take a look and suggest.

docs/source/using/run-against-an-existing-machine.adoc
praveenkumar: cc @robin-owen take a look and suggest.

docs/source/using/run-against-an-existing-machine.adoc
agajdosi: +1, in this use case it is nice to show how to set non-default password

docs/source/using/run-against-an-existing-machine.adoc
robin-owen: I agree with @LalatenduMohanty's suggestion with one addition -- I'd be more explicit about `minishift start`:

```adoc
To change the default username and password, you must apply the `htpasswd-identity-provider` add-on with the desired username and password after running the `minishift start` command.
```

If "and" is not technically accurate, "or" can be used instead. I'd avoid using "and/or".

docs/source/using/run-against-an-existing-machine.adoc
robin-owen: +1. It would also be worth noting what the default is if no alternative is given.

PR #2858, opened 2018-10-04
Issue #2842 update the image package and build tags

### LINKED ISSUE #2842: Update image package and latest cross build tags
As of now, we are using the older version of `image` package in our `Golang.toml` and also we need to update our cross build tags since some of the tags are now not required and even `podman` have cross build tags which we should have same. 

- https://github.com/containers/libpod/blob/master/Makefile#L22

praveenkumar:
Fixes #2842


praveenkumar:
retest this please.

Issue #2740, opened 2018-08-28
Minishift should use native keychain methods versus endorsing storage of plaintext passwords


LalatenduMohanty:
Copying text from https://issues.jboss.org/browse/CDK-230

The Minishift documentation suggests:

```console
Register permanently: You can export registration information to environment variables so that minishift picks it up automatically each time it starts as shown here: 
```

which then goes on to suggest exporting your Red Hat Developer Login information into .bashrc or .bash_profile.

This strikes me as terrible advice - passwords should not be stored plaintext on the filesystem, let alone encouraged to be.

Other commandline utilities like Git or Docker leverage OS native keychains, like osxkeychain or Seahorse. CDK should do the same.

PR #2859, opened 2018-10-04
Issue #2740 use native keychain for registration password

anjannath:
Fixes: #2740 

Steps to test:
1. start `minishift` with RHEL iso.
2. enter the registration _username_ and _password_ interactively when asked.
    - the password is now stored in keychain
3. start `minishift` again (same profile).
4. This time you just have to enter the _username_, the _password_ will be retrieved from OS native keychain.


anjannath:
OS native key chains can be used to delete the stored password. Password is stored under `minishift` service name.
- For macOS it is the **`Keychain Access`** application, 
- For GNU/Linux gnome-keyring frontend **`Seahorse`** can be used. (`dnf install seahorse`)
- For Windows **`Control Panel > User Accounts > Credential Manager`**

pkg/util/input.go
```diff
@@ -52,3 +53,21 @@ func AskForConfirmation(message string) bool {
 		" Do you want to continue [y/N]?")
 	return strings.ToUpper(userConfirmation) == "Y"
 }
+
+// Retrive password from OS native credential provider/keychain
+func GetPasswordKeyring(username string) string {
+	password, err := keyring.Get("minishift", username)
+	if err != nil {
+		return ""
```
praveenkumar: why `err` is not returned here and we are returning the empty string? if this is because in case user doesn't have keyring software install then it should not brag about it?

if we ignoring the error then why bother to evaluate because in case of error anyway password will be empty string right?

```
password, _ := keyring.Get("minishift", username)
return password
```

pkg/util/input.go
```diff
@@ -52,3 +53,21 @@ func AskForConfirmation(message string) bool {
...
+	}
+	return password
+}
+
+// Store the registration password in OS native keychain
+func SetPasswordKeyring(username, password string) error {
+	err := keyring.Set("minishift", username, password)
```
praveenkumar: Instead of putting `minishift` can you create a constant which should be `ProjectName` and value would be `minishift` and cosume it? you can create it in `pkg/minikube/constants/constants.go`

pkg/util/input.go
```diff
@@ -52,3 +53,21 @@ func AskForConfirmation(message string) bool {
...
+	if err != nil {
+		return err
```
praveenkumar: What if user doesn't have keyring software installed for win or linux, shouldn't this always going to put error? and we are going to ask user to install those?

anjannath: No extra keyring software is required for Windows, for Linux yes, you need to have `gnome-keyring` installed, So if you have a distro with Gnome DE then you already have the keyring software installed.

> shouldn't this always going to put error? 

Yes, that is the idea, if not installed or some other error, and that error is handled by the caller. In our case we just print the error and move on.

pkg/util/input.go
anjannath: Looks like there is already a constant `BinaryName` in `pkg/minishift/constants` going to use that.

pkg/minishift/registration/redhat.go
```diff
@@ -77,8 +78,14 @@ func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) er
 			}
 			// request password (disallow empty value)
 			if param.Password == "" {
+				param.Password = param.GetPasswordKeyring(param.Username)
+
 				for param.Password == "" {
 					param.Password = param.GetPasswordInteractive(RHSMName + " password")
+					err := param.SetPasswordKeyring(param.Username, param.Password)
+					if glog.V(3) {
+						fmt.Println(err)
```
praveenkumar: Gotca so here we are just print the error if log mode is 3 👍

pkg/util/input.go
```diff
@@ -19,11 +19,14 @@ package util
...
 	"fmt"
 	"strings"
+	"syscall"
 
+	"github.com/zalando/go-keyring"
 	"golang.org/x/crypto/ssh/terminal"
-	"syscall"
 )
 
+const projectName = "minishift"
```
praveenkumar: no, not here I was asking to have it part of constants file so in future if something else require it then can be consumed by easily.

anjannath: Yes, i  know but there's a import cycle if the constants (`pkg/minikube/constants` or `pkg/minishift/constants`) package is imported here, that's why put it here itself.
https://github.com/minishift/minishift/pull/2859#discussion_r222702323

pkg/minishift/registration/redhat.go
anjannath: Right..

pkg/util/input.go
praveenkumar: @anjannath can you put some time to actually resolve that cyclic dependency because constant pkg should be used anywhere without that kind of cyclic dependency.

anjannath: @praveenkumar I spent some time on it and looks like we won't be able to take away the cycling dependency, the `pkg/util` package is imported by `pkg/minikube/constants`, so we can't import it back in `pkg/util/input.go`.

Two ways i can think of solving this is put the keyring related new two functions in `pkg/cluster/registration.go` or in `start.go` where it is being used, The second way is we can have a different package under the `pkg/minishift/registration` for these two functions. both feels like an overkill, moreover i suppose if we want to be able to import the `constants` package everywhere then we need to move all the imports that `constants` package has to a standalone package or put them inside the `constants` package itself.

pkg/minikube/constants/constants.go
```diff
@@ -81,6 +81,24 @@ func GetMinishiftHomeDir() string {
 	if ok {
 		return homeEnv
 	} else {
-		return filepath.Join(util.HomeDir(), ".minishift")
+		return filepath.Join(GetHomeDir(), ".minishift")
 	}
 }
+
+// HomeDir returns the home directory for the current user
+func GetHomeDir() string {
```
praveenkumar: :+1: comment still refer to `HomeDir` though :)

pkg/minikube/kubeconfig/config.go
```diff
@@ -52,7 +52,7 @@ type SystemKubeConfig struct {
 }
 
 func GetConfigPath() string {
-	var configPath = filepath.Join(util.HomeDir(), ".kube", "config")
+	var configPath = filepath.Join(constants.GetHomeDir(), ".kube", "config")
```
praveenkumar: :+1:

pkg/minikube/constants/constants.go
anjannath: Ahh.. Fixing that :-)

anjannath: Done!

pkg/minikube/constants/constants.go
```diff
@@ -68,7 +68,7 @@ func GetProfileHomeDir(profile string) string {
 		if ok {
 			return filepath.Join(homeEnv, "profiles", profile)
 		}
-		return filepath.Join(util.HomeDir(), ".minishift", "profiles", profile)
+		return filepath.Join(GetHomeDir(), ".minishift", "profiles", profile)
```
praveenkumar: :+1:

praveenkumar:
retest this please.

anjannath:
retest this please

praveenkumar:
retest this please.

praveenkumar:
@anjannath can you rebase this with master now?

pkg/minikube/constants/constants.go
```diff
@@ -81,6 +81,24 @@ func GetMinishiftHomeDir() string {
...
 	} else {
-		return filepath.Join(util.HomeDir(), ".minishift")
+		return filepath.Join(GetHomeDir(), ".minishift")
 	}
 }
+
+// GetHomeDir returns the home directory for the current user
+func GetHomeDir() string {
+	if runtime.GOOS == "windows" {
+		if homeDrive, homePath := os.Getenv("HOMEDRIVE"), os.Getenv("HOMEPATH"); len(homeDrive) > 0 && len(homePath) > 0 {
```
gbraad: Is this tested?

pkg/minikube/constants/constants.go
```diff
@@ -81,6 +81,24 @@ func GetMinishiftHomeDir() string {
...
+			homeDir := homeDrive + homePath
+			if _, err := os.Stat(homeDir); err == nil {
```
gbraad: I believe the line which constructs the path on Windows is weak (and error prone), as there is no enforcement of format. The stat does check if exists... but there are no tests for this.

anjannath: @gbraad This method was moved here as it is from `pkg/util/homedir.go`

Did you mean the the line `homeDir := homeDrive + homePath` should be formed using `filepath.Join` instead?

gbraad: if possible, yes

anjannath: @praveenkumar Tested in windows working fine!

LalatenduMohanty:
@anjannath I do see any docs around how users can use the OS native key chains. Are you planning to add it.

anjannath:
@LalatenduMohanty In https://github.com/minishift/minishift/pull/2859#issuecomment-427019489 i have mentioned how one can use the keychain software on different platforms, we can certainly add docs on how to remove the password from the keychain in the event the password is changed via the website, but instead i would prefer we link to their original docs for the keychains.

LalatenduMohanty:
@minishift/minishift-dev @anjannath I am concerned whether this will cause issues to existing CDK users. Users who are not familiar with the keychain technology as @anjannath mentioned if the password got changed then user need to delete the old password manually from the keychain.

anjannath:
I think we can add documentation specific to each platform on how to delete
the password from the keychain.

And in the error msg we can specify the link to this doc

pkg/util/input.go
LalatenduMohanty: @anjannath We need to print the error in a way that user can get pointer and take action from the error.

LalatenduMohanty:
> I think we can add documentation specific to each platform on how to delete the password from the keychain. And in the error msg we can specify the link to this doc

+1

LalatenduMohanty:
> * start `minishift` again (same profile).
> 
> * This time you just have to enter the _username_, the _password_ will be retrieved from OS native keychain.

@anjannath We need to add one more line of message in the console that Minishift is retrieving the password from the keyring software 

### Now

```
Red Hat Developers or Red Hat Subscription Management (RHSM) username: lmohanty@redhat.com
   Login to registry.redhat.io in progress ....
```
#### Suggestion

```
Red Hat Developers or Red Hat Subscription Management (RHSM) username: lmohanty@redhat.com
   Retrieving the password from keyring software available in the host .. ok
   Login to registry.redhat.io in progress .... ok
```
```

File: pkg/minishift/registration/redhat.go [hunk 11]
```diff
@@ -77,8 +78,28 @@ func (registrator *RedHatRegistrator) Register(param *RegistrationParameters) er
 			}
 			// request password (disallow empty value)
 			if param.Password == "" {
+				if i == 1 {
+					fmt.Printf("   Retriving password from keychain ...")
```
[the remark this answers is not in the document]
anjannath: Also added docs in the troubleshooting/miscellaneous

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,29 @@ Workaround: Delete the existing {project} VM and start again.
 $ minishift delete --force
 $ minishift start
 ----
+
+[[Remove-password-from-keychain]]
+== Removing registration password from OS native keychain
+
```
robin-owen: I'd suggest specifying that this only applies to the RHEL ISO. (Unless I'm misunderstanding something, these instructions only apply for RHEL.)

Also, when would we expect users to do this? If they've changed their password? Are there any other situations where this would be useful to know? I'm asking so that we can mention those situations here -- this could potentially help users who search the documentation for their particular problem.

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,29 @@ Workaround: Delete the existing {project} VM and start again.
...
+=== Windows
+
+In windows the password is stored using *Credential Manager*.
+To delete the stored password run the following command in command prompt:
+
+----
+CMD> cmdkey /delete:minishift:<username>
+----
+
+=== GNU/Linux
```
robin-owen: I don't personally mind if we keep "GNU/Linux", but it's worth noting that the OpenShift documentation refers to the operating system as "Linux". This is why I've used "Linux" previously in the documentation -- to keep it consistent.

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,29 @@ Workaround: Delete the existing {project} VM and start again.
...
 $ minishift start
 ----
+
+[[Remove-password-from-keychain]]
+== Removing registration password from OS native keychain
+
+=== Windows
+
+In windows the password is stored using *Credential Manager*.
+To delete the stored password run the following command in command prompt:
```
robin-owen: I'd rearrange this sentence to avoid splicing:

```suggestion
Run the following command in Command Prompt to delete the stored password:
```

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,29 @@ Workaround: Delete the existing {project} VM and start again.
 $ minishift delete --force
 $ minishift start
 ----
+
+[[Remove-password-from-keychain]]
+== Removing registration password from OS native keychain
+
+=== Windows
+
+In windows the password is stored using *Credential Manager*.
```
robin-owen:
```suggestion
On Windows, the password is stored using *Credential Manager*.
```

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,29 @@ Workaround: Delete the existing {project} VM and start again.
...
+To delete the stored password run the following command in command prompt:
+
+----
+CMD> cmdkey /delete:minishift:<username>
+----
+
+=== GNU/Linux
+
+In linux the password is stored using *gnome-keyring*.
```
robin-owen:
```suggestion
On Linux, the password is stored using *gnome-keyring*.
```

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,29 @@ Workaround: Delete the existing {project} VM and start again.
...
+To delete the stored password run the following command:
```
robin-owen:
```suggestion
Run the following command to delete the stored password:
```

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,29 @@ Workaround: Delete the existing {project} VM and start again.
...
+
+----
+$ secret-tool clear service minishift username <username>
+----
+
+=== MacOS
```
robin-owen:
```suggestion
=== macOS
```

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,29 @@ Workaround: Delete the existing {project} VM and start again.
...
+
+For macOS use the *Keychain Access* GUI app to delete the stored password, look for an entry  under passwords that says minishift.
```
robin-owen: Unfortunately, I don't use macOS, so I can't see the GUI being described here myself. What we could use, though, is a little bit more instruction about what to do with the entry named `minishift`. I'd also recommend rewording this some, and I'd make referenced GUI elements bold; something like this:

```adoc
For macOS, use the *Keychain Access* application to delete the stored password.
Look for an entry under *Passwords* named `minishift`.
// Describe further actions needed here.
```

docs/source/troubleshooting/troubleshooting-misc.adoc
robin-owen: Note: You're also switching between "GNU/Linux" and "Linux" within the same section -- use one or the other, not both, please. :)

docs/source/troubleshooting/troubleshooting-misc.adoc
anjannath: > I'd suggest specifying that this only applies to the RHEL ISO

Yes, this only applies to RHEL iso

> when would we expect users to do this? If they've changed their password?

Only when they have changed the password, or mistakenly entered a wrong password which got stored in the keychain

docs/source/troubleshooting/troubleshooting-misc.adoc
anjannath: On a second thought i should just mention the command to delete it like the rest

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,35 @@ Workaround: Delete the existing {project} VM and start again.
...
+$ secret-tool clear service minishift username <username>
+----
+
+=== macOS
+
+On macOS, the password is stored using *Keychain Access*.
+Run the following command to delete the stored password:
+
+----
+$ security delete-generic-password -s minishift
```
robin-owen: :+1:

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,35 @@ Workaround: Delete the existing {project} VM and start again.
...
 ----
+
+[[Remove-password-from-keychain]]
+== Removing subscription password from OS native keychain
+
+This only applies if you are using a subscription capable distro like RHEL with {project}. 
+
+=== Windows
+
+On windows, the password is stored using *Credential Manager*.
```
robin-owen: Nitpick: "Windows" should be capitalized. :)

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,35 @@ Workaround: Delete the existing {project} VM and start again.
...
+Run the following command in Command Prompt to delete the stored password:
+
+----
+CMD> cmdkey /delete:minishift:<username>
+----
+
+=== Linux
+
+On linux, the password is stored using *gnome-keyring*.
```
robin-owen: Nitpick: "Linux" should be capitalized. :)

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,35 @@ Workaround: Delete the existing {project} VM and start again.
 $ minishift delete --force
 $ minishift start
 ----
+
+[[Remove-password-from-keychain]]
+== Removing subscription password from OS native keychain
+
+This only applies if you are using a subscription capable distro like RHEL with {project}.
```
LalatenduMohanty: s/This/These steps/ 
s/you are using a subscription capable distro like RHEL with {project}./you are using a subscription capable ISO image like Red Hat Enterprise Linux with {project} which needs registration using username/password./

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,35 @@ Workaround: Delete the existing {project} VM and start again.
...
 ----
+
+[[Remove-password-from-keychain]]
+== Removing subscription password from OS native keychain
+
+This only applies if you are using a subscription capable distro like RHEL with {project}. 
+
+=== Windows
+
+On Windows, the password is stored using *Credential Manager*.
```
LalatenduMohanty: s/the password/by default the password/ . Also it is same for all Windows versions? e.g. between Windows 7 and Windows 10.

docs/source/troubleshooting/troubleshooting-misc.adoc
```diff
@@ -93,3 +93,35 @@ Workaround: Delete the existing {project} VM and start again.
...
+Run the following command in Command Prompt to delete the stored password:
+
+----
+CMD> cmdkey /delete:minishift:<username>
+----
+
+=== Linux
+
+On Linux, the password is stored using *gnome-keyring*.
```
LalatenduMohanty: s/On Linux/On Gnome based Linux systems/

LalatenduMohanty: @anjannath The issue with Linux is there are so many variants of those, so we can not say in all Linux gnome-keyring is used e.g. kde based distributions.

LalatenduMohanty: You should mention `On Linux the feature needs the secret service dbus interface, which is provided by GNOME Keyring`.

LalatenduMohanty: @anjannath sounds good to me.

LalatenduMohanty: @anjannath can you make it `On Linux the feature need to store the password using **D-Bus Secret Service API**, which is provided by **GNOME libsecret**.`

anjannath: Yes, made the change @LalatenduMohanty

pkg/minikube/constants/constants.go
LalatenduMohanty: I also run the PR on Windows 10. `minishift start` worked fine.

PR #2863, opened 2018-10-05
Issue #2862 Test coverage for minishift openshift component command

### LINKED ISSUE #2862: Add more test coverage to minishift openshift component command
```#bash
$ ./minishift openshift component -h
Interacts with the components to an OpenShift cluster (Works only with OpenShift version >= 3.10.x)

Usage:
  minishift openshift component [command] [flags]
  minishift openshift component [command]

Available Commands:
  add         Add component to an OpenShift cluster (Works only with OpenShift version >= 3.10.x)
  list        Add component to an OpenShift cluster (Works only with OpenShift version >= 3.10.x)

Flags:
  -h, --help   help for component

Global Flags:
      --profile string                   Profile name (default "minishift")
      --show-libmachine-logs             Show logs from libmachine.
      --stderrthreshold severity         logs at or above this threshold go to stderr (default 2)
  -v, --v Level                          log level for V logs. Level varies from 1 to 5 (default 1).
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging

Use "minishift openshift component [command] --help" for more information about a command.
```

amitkrout:
Fixes #2862 

  1. make integration GODOG_OPTS="-tags cmd-openshift"
  2. 
  3. 
  4.

### DIFF
--- a/test/integration/features/cmd-openshift.feature
+++ b/test/integration/features/cmd-openshift.feature
@@ -5,143 +5,167 @@ cluster in VM provided by Minishift.
 
   @quick
   Scenario: Trying service command when Minishift is not running
-     Given Minishift has state "Does Not Exist"
-      When executing "minishift openshift service list" succeeds
-      Then stdout should contain
-       """
-       Running this command requires an existing 'minishift' VM, but no VM is defined.
-       """
+    Given Minishift has state "Does Not Exist"
+     When executing "minishift openshift service list" succeeds
+     Then stdout should contain
+      """
+      Running this command requires an existing 'minishift' VM, but no VM is defined.
+      """
 
   Scenario: Minishift start
   Minishift must be started in order to interact with OpenShift via "minishift openshift" command
-
-     Given Minishift has state "Does Not Exist"
-       And image caching is disabled
-      When executing "minishift start" succeeds
+    Given Minishift has state "Does Not Exist"
+      And image caching is disabled
+     When executing "minishift start" succeeds
 
   Scenario: Service list sub-command
-     Given Minishift has state "Running"
-      When executing "minishift openshift service list" succeeds
-      Then stdout should contain "docker-registry"
-       And stdout should contain "kubernetes"
-       And stdout should contain "router"
+    Given Minishift has state "Running"
+     When executing "minishift openshift service list" succeeds
+     Then stdout should contain "docker-registry"
+      And stdout should contain "kubernetes"
+      And stdout should contain "router"
 
   Scenario: Pods docker-registry and router are ready before OpenShift restart
-     Given user waits "30" seconds
-      When executing "oc get pods -n default --as system:admin" succeeds
-      Then stdout should match "docker-registry-\d-\w{5}\s*1\/1\s*Running"
-       And stdout should match "router-\d-\w{5}\s*1\/1\s*Running"
+    Given user waits "30" seconds
+     When executing "oc get pods -n default --as system:admin" succeeds
+     Then stdout should match "docker-registry-\d-\w{5}\s*1\/1\s*Running"
+      And stdout should match "router-\d-\w{5}\s*1\/1\s*Running"
 
   Scenario: Restarting the OpenShift cluster
   Note: This step is based on observation and might be unstable in some environments. It checks for the time when container
         finished last time. When container is new and had never finished then this time value is set to 0001-01-01T00:00:00Z.
         On restart of OpenShift cluster containers are terminated, which sets FinishedAt to actual time. This value persist
         after next start of container.
-     Given stdout of command "minishift ssh -- "docker inspect --format={{.State.FinishedAt}} origin"" is equal to "0001-01-01T00:00:00Z"
-      When executing "minishift openshift restart" succeeds
-      Then stdout should contain "OpenShift restarted successfully"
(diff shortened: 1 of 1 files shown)


test/integration/features/cmd-openshift.feature
```diff
@@ -5,143 +5,164 @@ cluster in VM provided by Minishift.
...
-      When executing "minishift openshift config view" succeeds
-      Then stdout is YAML which contains key "assetConfig.logoutURL" with value matching "http://www\.minishift\.io"
-
-  Scenario: Add a component to running openshift cluster
-      When executing "minishift openshift component add service-catalog" succeeds
-       And stdout should contain
-       """
-       Finished installing "openshift-service-catalog"
-       """
+     When executing "minishift openshift config set --patch '{"assetConfig": {"logoutURL": "http://www.minishift.io"}}'" succeeds
```
agajdosi: Have tried this on Windows? I am suspicious that it will have troubles here.

test/integration/features/cmd-openshift.feature
```diff
@@ -5,143 +5,164 @@ cluster in VM provided by Minishift.
...
+     Then stdout should contain "Patching OpenShift configuration"
+      And user waits "30" seconds
+     When executing "minishift openshift config view" succeeds
```
agajdosi: This should be separate scenario 
scenario: patch was applied
given user waits
when executing succeeds
then stdout is Yaml...

test/integration/features/cmd-openshift.feature
```diff
@@ -5,143 +5,164 @@ cluster in VM provided by Minishift.
...
 
   Scenario: Deleting the Minishift instance
-     Given Minishift has state "Running"
-      When executing "minishift delete --force" succeeds
-      Then Minishift should have state "Does Not Exist"
+    Given Minishift has state "Running"
+     When executing "minishift delete --force" succeeds
+     Then Minishift should have state "Does Not Exist"
+
+  Scenario: Applying component service-catalog whitout running Minishift instance
```
agajdosi: here you can add tag `@quick`

test/integration/features/cmd-openshift.feature
```diff
@@ -5,143 +5,164 @@ cluster in VM provided by Minishift.
...
+  Scenario: List out service-catalog component
+     When executing "minishift openshift component list" succeeds
+     Then stdout should contain
+      """
+      service-catalog
+      """
+
+  Scenario: Add a component to running OpenShift cluster
+     When executing "minishift openshift component add service-catalog" succeeds
+      And stdout should contain
```
agajdosi: It should be `then` here.

test/integration/features/cmd-openshift.feature
LalatenduMohanty: @agajdosi This is not a new scenario as @amitkrout has fixed the formatting, so wondering why we have not reported if it fails in Windows?

agajdosi: @LalatenduMohanty I see, missed that. So there is no problem with this. Thanks for mentioning.

amitkrout: I have used same file for downstream bits validation and never come across with the failure on windows terminal (powershell)

agajdosi: @amitkrout Yes, I have missed that it is already added and thus it works as we tested the feature over and over. Thought it is new scenario...

test/integration/features/cmd-openshift.feature
amitkrout: I did not realize this as it was from the creation of the feature file i guess. Good to have break these scenario to two scenario

test/integration/features/cmd-openshift.feature
amitkrout: Addition of more tags resolves some problem but increases an extra layer of work for contributors for selecting the tag like ```@quick``` which i feel nobody cares. Later down the line i think it will be confusion between quick and normal scenario. 

IMO let keep it as it is and later we can think of assigning the tag if there is hard requirement. WDYT ?

agajdosi: whatever

amitkrout: :)

test/integration/features/cmd-openshift.feature
amitkrout: Done

test/integration/features/cmd-openshift.feature
agajdosi: `@quick` is part of the `make integration`, PR was approved and pulled to the project, no one complaint at that time. If nobody cares and uses it, well ok. Please go ahead and create an issue for removal of the tag from features and makefile and from documentation.

amitkrout: Aha... Might be i misunderstood ```quick``` tag. What i understood is quick comprises basic + those scenario which do not require a running minishift instance.

agajdosi: Exactly and I believe `Scenario: Applying component service-catalog whitout running Minishift instance` does not require a running Minishift instance. So in case of quick the first scenario and this last one will only run. Making it like 2 seconds or something like that.

BTW there is a typo in the scenario name: `whitout`.

amitkrout: Done

PR #2874, opened 2018-10-10
Issue #2474 Image Cache: Add append feature for index file

### LINKED ISSUE #2474: Any other image is deleted if exported earlier before any default images
### General information

  * Minishift version: 
  * OS: Linux / macOS / Windows

### Steps to reproduce

  1. minishift start
```
$ ./minishift start
--
-- Check if depereccated options are used ... OK
[...]
-- Applying addon 'admin-user':..
-- Exporting of OpenShift images is occuring in background process with pid 9687.
```
  2. minishift image list
```
$ ./minishift image list
--
$
```
  3. minishift image export alpine:latest
```
$ ./minishift image export alpine:latest
--
Exporting 'alpine:latest'..... OK
```
  4. minishift image list
```
$ ./minishift image list
--
alpine:latest
```
  5. Wait till all the default images to be exported
  6. Then image list should contain "alpine:latest" but the image is no more in the list
  7. minishift image list
```
$ ./minishift image list
--
registry.access.redhat.com/openshift3/ose-docker-registry:v3.9.30
registry.access.redhat.com/openshift3/ose-haproxy-router:v3.9.30
registry.access.redhat.com/openshift3/ose:v3.9.30
```

### Expected
Exported image should be listed out

### Actual
Exported images is deleted 

### Logs

You can start Minishift with `minishift start --show-libmachine-logs -v5` to collect logs.
Please consider posting this on http://gist.github.com/ and post the link in the issue.

praveenkumar:
Fixes #2474

Make sure you also test this with an old image cache and this shouldn't break any backward compatibility of our `image cache` feature.

```
Terminal 1 $ ./minishift image export fedora
Terminal 2 $ ./minishift image export busybox => do it alomst same time.

Let's wait till both the export complete.

$ ./minishift image list  => This should now show both the images
fedora:latest
busybox:latest
```

### DIFF
--- a/pkg/minishift/docker/image/oci_image_handler.go
+++ b/pkg/minishift/docker/image/oci_image_handler.go
@@ -55,8 +55,11 @@ type dockerClientConfig struct {
 }
 
 type Index struct {
-	Manifests Manifests `json:"manifests"`
+	Manifests     Manifests     `json:"manifests"`
+	SchemaVersion SchemaVersion `json:"schemaVersion"`
 }
+
+type SchemaVersion int
 
 type Manifests []Manifest
 
@@ -216,7 +219,7 @@ func (handler *OciImageHandler) AreImagesCached(config *ImageCacheConfig) bool {
 func (handler *OciImageHandler) GetCachedImages(config *ImageCacheConfig) map[string]bool {
 	cachedImages := make(map[string]bool)
 
-	index, err := handler.getIndex(config)
+	index, err := handler.getIndex(config.HostCacheDir)
 	if index == nil || err != nil {
 		return cachedImages
 	}
@@ -274,8 +277,8 @@ func (handler *OciImageHandler) pullImage(image string, out io.Writer) error {
 	return nil
 }
 
-func (handler *OciImageHandler) getIndex(config *ImageCacheConfig) (*Index, error) {
-	indexPath := filepath.Join(config.HostCacheDir, "index.json")
+func (handler *OciImageHandler) getIndex(cacheDir string) (*Index, error) {
+	indexPath := filepath.Join(cacheDir, "index.json")
 	if !filehelper.Exists(indexPath) {
 		return nil, nil
 	}
@@ -305,7 +308,7 @@ func (handler *OciImageHandler) importImage(image string, config *ImageCacheConf
 		return fmt.Errorf("Invalid image source '%s': %v", image, err)
 	}
 
-	err = handler.copyImage(srcRef, destRef, policyContext)
+	err = handler.copyImage(srcRef, destRef, policyContext, config.HostCacheDir)
 	if err != nil {
 		return err
 	}
@@ -331,13 +334,50 @@ func (handler *OciImageHandler) exportImage(image string, config *ImageCacheConf
 		return fmt.Errorf("Invalid image source '%s': %v", image, err)
 	}
 
-	destRef, err := layout.NewReference(config.HostCacheDir, image)
+	// ImageIndexLocation should be a directory location which will be atomic for each image.
+	// for an image "openshift/origin-control-plane:v3.10.0"
+	// it will be $HOME/.minishift/cache/image/openshift-origin-control-plane-v3.10.0
+	r := strings.NewReplacer(":", "-", "/", "-")
+	ImageIndexLocation := filepath.Join(config.HostCacheDir, r.Replace(image))
+
+	if _, err := os.Stat(ImageIndexLocation); !os.IsNotExist(err) {
+		return fmt.Errorf("Importing %s is already in progress", image)
+	}
+
+	destRef, err := layout.NewReference(ImageIndexLocation, image)
 	if err != nil {
 		return fmt.Errorf("Invalid image destination '%v': %v", destRef, err)
 	}
 
-	err = handler.copyImage(srcRef, destRef, policyContext)
+	err = handler.copyImage(srcRef, destRef, policyContext, config.HostCacheDir)
 	if err != nil {
+		return err
+	}
+
+	// Get index of current export image
+	pulledImageIndex, err := handler.getIndex(ImageIndexLocation)
+	if err != nil {
+		return err
+	}
+
+	// Remove tmpImageIndexLocation
+	if err := os.RemoveAll(ImageIndexLocation); err != nil {
+		return err
+	}
+
+	// Get index of already available image
+	availableImageIndex, err := handler.getIndex(config.HostCacheDir)
+	if err != nil {
+		return err
+	}
+
+	if availableImageIndex == nil {
+		availableImageIndex = &Index{Manifests: Manifests{}, SchemaVersion: 2}
+	}
+
+	availableImageIndex.Manifests = append(availableImageIndex.Manifests, pulledImageIndex.Manifests...)
+
+	if err := handler.updateIndex(config.HostCacheDir, availableImageIndex); err != nil {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


LalatenduMohanty:
@praveenkumar This reduces the chance of overwriting as you are appending to the index and just not overwriting. So there is still a chance of missing information in the index file, right? Though is way better than the previous situation we had of completely overwriting the index file. Did I miss anything?

praveenkumar:
> This reduces the chance of overwriting as you are appending to the index and just not overwriting. So there is still a chance of missing information in the index file, right? 

@LalatenduMohanty No, now there is no chance to of missing information because we are using a temp location and then append what already available on the index so index file is now under our control and we are doing atomic operation only.

pkg/minishift/docker/image/oci_image_handler.go
```diff
@@ -331,21 +334,54 @@ func (handler *OciImageHandler) exportImage(image string, config *ImageCacheConf
 		return fmt.Errorf("Invalid image source '%s': %v", image, err)
 	}
 
-	destRef, err := layout.NewReference(config.HostCacheDir, image)
+	// tmpImageIndexLocation should be a directory location which will be atomic for each image.
+	// for an image "openshift/origin-control-plane:v3.10.0"
+	// it will be $HOME/.minishift/cache/image/openshift-origin-control-plane-v3.10.0
+	r := strings.NewReplacer(":", "-", "/", "-")
+	tmpImageIndexLocation := filepath.Join(config.HostCacheDir, r.Replace(image))
```
LalatenduMohanty: Temporary directory name should not be manually created. It should be auto generated which comes with random number etc. Also we have a location i.e. `~/.minishift/tmp` , so we should use that location for temporary directory/file. 

If you run the same image export command from two shells it will have same directory name, which is not desirable.

pkg/minishift/docker/image/oci_image_handler.go
```diff
@@ -331,21 +334,54 @@ func (handler *OciImageHandler) exportImage(image string, config *ImageCacheConf
...
+	availableImageIndex, err := handler.getIndex(config.HostCacheDir)
+	if err != nil {
+		return err
+	}
+
+	if availableImageIndex == nil {
+		availableImageIndex = &Index{Manifests: Manifests{}, SchemaVersion: 2}
+	}
+
+	availableImageIndex.Manifests = append(availableImageIndex.Manifests, pulledImageIndex.Manifests...)
```
LalatenduMohanty: Is this atomic operation? I do not think so. If two process are exactly at this step at exact same time how will it behave? However I agree the whole PR  reduces the probability of overwriting the index file. But if you want to go one step ahead you should use mutex to make sure only one process can call this step. 

https://stackoverflow.com/questions/46138395/how-make-a-function-thread-safe-in-golang

pkg/minishift/docker/image/oci_image_handler.go
praveenkumar: @LalatenduMohanty No so what we need to do is stop the that operation, if a user try to export same image in 2 different terminal we shouldn't allow it. I will make that change but using the `tmp` module is not what I am after atm.

pkg/minishift/docker/image/oci_image_handler.go
praveenkumar: @LalatenduMohanty what is the probability you think this has to behave as unexpected, check the flow. 

- the image first goes to index on its own.
- Take the existing index in memory and append the index.
- In meanwhile if some other process is running then it will also take index in the memory and append it. We are not overwriting anything at all so I don't know why we want to use a mutex or any other complicated stuff.

praveenkumar:
@LalatenduMohanty  try now with exporting the same image at around same time in 2 different terminals you should see following. Even without it, it will not be going to corrupt the index.json file it will just be appended but I think telling user early is better.

```
Terminal-1 $ ./minishift image export fedora:latest
Exporting 'fedora:latest'................. OK

Terminal-2 $ ./minishift image export fedora:latest
Exporting 'fedora:latest'..... FAIL
Container image export failed:
Importing fedora:latest is already in progress
```

pkg/minishift/docker/image/oci_image_handler.go
```diff
@@ -466,8 +510,12 @@ func (handler *OciImageHandler) progressStatusForError(err error) ProgressStatus
...
 }
 
-func (handler *OciImageHandler) updateIndex(config *ImageCacheConfig, index *Index) error {
-	indexPath := filepath.Join(config.HostCacheDir, "index.json")
+func (handler *OciImageHandler) updateIndex(cacheDir string, index *Index) error {
+	// We only need lock for file write operation.
+	// Check https://play.golang.org/p/2eQmVdxI1LQ for reference.
+	mu.Lock()
+	defer mu.Unlock()
+	indexPath := filepath.Join(cacheDir, "index.json")
```
LalatenduMohanty: @praveenkumar I think this will not solve the issue of race condition. Because when two process have the same copy of `availableImageIndex` from `availableImageIndex, err := handler.getIndex(config.HostCacheDir)` then they are not aware of each other and would have appended the new value which is applicable to the process.  We need to make sure that each process gets the correct copy of `availableImageIndex`. So only one process at one time should get value of `availableImageIndex` -> update -> write it to the disk then other process take control of it. WDYT?

praveenkumar: @LalatenduMohanty Honestly, I am not able to process it from my mind mostly because I am not able to test it out properly so I will just revert it back and if you think that is really we need to do might be create an followup issue.

LalatenduMohanty: >  I will just revert it back and if you think that is really we need to do might be create an followup issue.

+1, lets do that.

LalatenduMohanty:
retest this please

gbraad:
retest this please

pkg/minishift/docker/image/oci_image_handler.go
```diff
@@ -331,21 +334,58 @@ func (handler *OciImageHandler) exportImage(image string, config *ImageCacheConf
...
 	}
 
-	destRef, err := layout.NewReference(config.HostCacheDir, image)
+	// ImageIndexLocation should be a directory location which will be atomic for each image.
+	// for an image "openshift/origin-control-plane:v3.10.0"
+	// it will be $HOME/.minishift/cache/image/openshift-origin-control-plane-v3.10.0
+	r := strings.NewReplacer(":", "-", "/", "-")
+	ImageIndexLocation := filepath.Join(config.HostCacheDir, r.Replace(image))
+
+	if _, err := os.Stat(ImageIndexLocation); !os.IsNotExist(err) {
```
gbraad: `IsExist`... please do not double negate

praveenkumar: @gbraad it's not like this there is major difference when to use `IsExist` and when `IsNotExist`, have a look to https://gist.github.com/mastef/05f46d3ab2f5ed6a6787

praveenkumar:
retest this please.

pkg/minishift/docker/image/oci_image_handler.go
gbraad: It is about readability...

`not not exists` means `exists`. you also avoid double negativity in a spoken language when possible.

gbraad: confusing logic in golang... sounds like a bad api decision >;-/

gbraad:
will merge after CI passes

praveenkumar:
CI failure is known here, merging it now.

Issue #2871, opened 2018-10-09
Provide option to fully manage kubeconfig within MINISHIFT_HOME


hferentschik:
The idea is to let Minishift write the kubeconfig data somewhere into $MINISHIFT_HOME. To stay backward compatible, one could provide an option to enable this feature, eg

```
$ minishift start --minishift-managed-kubeconfig
```

Name to be discussed. Alternatively, it could be a persistent flag only. Once set Minishift will write the config to a $MINISHIFT_HOME internal location. 

If _minishift-managed-kubeconfig_ is enabled, you can configure your shell via which would in this case also export the right KUBECONFIG

```
$ eval $(minishit oc-env)
export PATH="/home/john/.minishift/cache/oc/v1.9.0:$PATH"
export KUBECONFIG=/home/john/.minishift/kube/config"
# Run this command to configure your shell:
# eval $(minishift oc-env)
```

This feature would help to keep better track of the local kubeconfig. Currently, _~/.kube/config_ quickly gets very messy if one creates many Minishift instances. If one also has config data one does not want to delete, one needs to carfully prune kubeconfig.

hferentschik:
Relates to #2871

LalatenduMohanty:
Aren't we using the `~/.kube` because `oc` recognizes the location too. If we change the default location then how will `oc` get the required files for authentication ?

hferentschik:
> This way.

Once you work with multiple clusters, you start to appreciate the possibility of separating the configurations. 

It also means a Minishift instance is even more contained, aka is under full control of us.

jorgemoralespou:
@hferentschik this will mean that you'll need to execute this `eval $(minishit oc-env)` on every shell you open.

The underlying problem is how minishift manages .kube/config and should be fixed there. https://github.com/openshift/origin/issues/18645

hferentschik:
> this will mean that you'll need to execute this eval $(minishit oc-env) on every shell you open.

Only if I care and only if I want to keep a config per Minishift instance. It is an optional feature and it is up to the user to use it the way she seems fit.

Personally, I don't have an issue with that. It's not that I constantly open and close shells. Also, if you always want to make sure that your shell is using a _oc_ version which aligns with the minishift instance you are talking to, you need to run _oc_env_ anyways. 

Last but not least, you could add something to your bashrc to handle this for you. 

> The underlying problem is how minishift manages .kube/config and should be fixed there.

[...]

PR #2881, opened 2018-10-11
Issue #2870 Honour KUBECONFIG env if set

### LINKED ISSUE #2870: Minishift needs to honour KUBECONFIG
ATM `minishift start` does not honour the _KUBECONFIG_ environment variable. For example,

```
$ export KUBECONFIG=/Users/john/tmp/minishift-kube-config
$ minsihift start
```

The certificate data as well as context for system:admin gets still written into _~/.kube/config_. This prevents me from better controlling my various OpenShift configurations.

praveenkumar:
Fixes #2870

cc @hferentschik 

```
$ export KUBECONFIG=/tmp/foo/config
$ minishift start
$ cat /tmp/foo/config
$ oc whoami <= on the same terminal where you have exported the KUBECONFIG file.
```

### DIFF
--- a/pkg/minishift/oc/oc.go
+++ b/pkg/minishift/oc/oc.go
@@ -26,6 +26,7 @@ import (
 	"regexp"
 	"strings"
 
+	"github.com/golang/glog"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	minishiftConstants "github.com/minishift/minishift/pkg/minishift/constants"
 	"github.com/minishift/minishift/pkg/util"
@@ -34,6 +35,7 @@ import (
 	"github.com/pkg/errors"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	"runtime"
 )
 
 const (
@@ -92,24 +94,25 @@ func (oc *OcRunner) AddSudoerRoleForUser(user string) error {
 func (oc *OcRunner) AddSystemAdminEntryToKubeConfig(ocPath string) error {
 	var existingKubeConfig, newKubeConfig *clientcmdapi.Config
 
-	user, err := user.Current()
-	if err != nil {
-		return fmt.Errorf("unable to find current user: %v", err)
-	}
-
 	minishiftKubeConfigPath := oc.KubeConfigPath
-	kubeConfigGlobalPath := filepath.Join(user.HomeDir, ".kube", "config")
-
-	if err := os.MkdirAll(filepath.Join(user.HomeDir, ".kube"), 0755); err != nil {
-		return fmt.Errorf("unable to create kubeconfig dir ($HOME/.kube): %v", err)
+	globalKubeConfigPath, err := getGlobalKubeConfigPath()
+	if err != nil {
+		return err
+	}
+	if glog.V(2) {
+		fmt.Println("Using Kubeconfig Path: ", globalKubeConfigPath)
+	}
+	dir, _ := filepath.Split(globalKubeConfigPath)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return fmt.Errorf("unable to create kubeconfig dir %s: %v", dir, err)
 	}
 
 	// Make sure .kube/config exist if not then this will create
-	os.OpenFile(kubeConfigGlobalPath, os.O_RDONLY|os.O_CREATE, 0600)
+	os.OpenFile(globalKubeConfigPath, os.O_RDONLY|os.O_CREATE, 0600)
 
-	existingKubeConfig, err = clientcmd.LoadFromFile(kubeConfigGlobalPath)
+	existingKubeConfig, err = clientcmd.LoadFromFile(globalKubeConfigPath)
 	if err != nil {
-		return fmt.Errorf("Not able to load %s: %s", kubeConfigGlobalPath, err)
+		return fmt.Errorf("Not able to load %s: %s", globalKubeConfigPath, err)
 	}
 
 	newKubeConfig, err = clientcmd.LoadFromFile(minishiftKubeConfigPath)
@@ -119,7 +122,7 @@ func (oc *OcRunner) AddSystemAdminEntryToKubeConfig(ocPath string) error {
 
 	merged := mergeKubeConfigs([]*clientcmdapi.Config{existingKubeConfig, newKubeConfig})
 
-	return clientcmd.WriteToFile(*merged, kubeConfigGlobalPath)
+	return clientcmd.WriteToFile(*merged, globalKubeConfigPath)
 }
 
 // AddCliContext adds a CLI context for the user and namespace for the current OpenShift cluster. See also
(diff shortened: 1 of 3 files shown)


praveenkumar:
@LalatenduMohanty this PR is only for honor the KUBECONFIG env nothing more. what you suggested is something part of #2871. No 'MINISHIFT_KUBECONFIG` and `KUBECONFIG` 2 different thing and all the folks who works around kubernetes known to `KUBECONFIG`.

LalatenduMohanty:
> No `MINISHIFT_KUBECONFIG` and `KUBECONFIG` 2 different thing and all the folks who works around kubernetes known to `KUBECONFIG`.

You are right about `KUBECONFIG`. I missed it https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#set-the-kubeconfig-environment-variable

pkg/minishift/oc/oc.go
```diff
@@ -226,3 +225,21 @@ func mergeKubeConfigs(configs []*clientcmdapi.Config) *clientcmdapi.Config {
 	}
 	return mergedConfig
 }
+
+func getKubeConfigGlobalPath() (string, error) {
+	kubeConfigGlobalPath := os.Getenv("KUBECONFIG")
+	if kubeConfigGlobalPath != "" {
+		kubeConfigGlobalPath = strings.FieldsFunc(kubeConfigGlobalPath, splitKubeConfig)[0]
```
hferentschik: So the idea is to take the first config in the path, right? Might be worth to document, probably in a godoc:
"getKubeConfigGlobalPath returns the path to the first entry in KUBECONFIG or if not set ~/home/.kube/config" - something along these lines. 

This will help other people to understand the code. I don't think it is so well known that KUBECONFIG can have multiple path elements.

pkg/minishift/oc/oc.go
```diff
@@ -226,3 +225,21 @@ func mergeKubeConfigs(configs []*clientcmdapi.Config) *clientcmdapi.Config {
 	}
 	return mergedConfig
 }
+
+func getKubeConfigGlobalPath() (string, error) {
```
hferentschik: Nice to extract this into a function.

pkg/minishift/oc/oc.go
```diff
@@ -94,16 +94,15 @@ func (oc *OcRunner) AddSudoerRoleForUser(user string) error {
 func (oc *OcRunner) AddSystemAdminEntryToKubeConfig(ocPath string) error {
 	var existingKubeConfig, newKubeConfig *clientcmdapi.Config
 
-	user, err := user.Current()
+	minishiftKubeConfigPath := oc.KubeConfigPath
+	kubeConfigGlobalPath, err := getKubeConfigGlobalPath()
```
LalatenduMohanty: globalKubeConfigPath sounds better as a variable name here.

LalatenduMohanty: Same with function name too i.e. `getGlobalKubeConfigPath`

pkg/minishift/oc/oc.go
praveenkumar: > So the idea is to take the first config in the path, right? Might be worth to document, probably in a godoc

@hferentschik Sure that something we can add.

> I don't think it is so well known that KUBECONFIG can have multiple path elements.

https://kubernetes.io/docs/tasks/access-application-cluster/configure-access-multiple-clusters/#set-the-kubeconfig-environment-variable

pkg/minishift/oc/oc.go
praveenkumar: @LalatenduMohanty it was there before so I didn't change it but can be updated.

LalatenduMohanty:
So minikube has a function for handling KUBECONFIG but we can not use it in its current form https://github.com/minishift/minishift/blob/master/pkg/minikube/kubeconfig/config.go#L54-L6

pkg/minishift/oc/oc_test.go
```diff
@@ -144,6 +146,24 @@ func TestFlagExist(t *testing.T) {
 	assert.False(t, flagExist(expectedOptions, "proxy"))
 }
 
+func Test_get_kubeconfig_global_path(t *testing.T) {
+	testdata := []struct {
+		KubeConfig string
+		Expected   string
+	}{
+		{KubeConfig: "/tmp/john/config", Expected: "/tmp/john/config"},
```
hferentschik: what's about adding some more boundary conditions? What does things like ":/home/john" do or  ";;", etc. The tests are only covering the happy paths.

On another note, maybe the split character should be determined at runtime based on the OS, aka ":" on darwin/linux and ';' on windows. That would be more accurate imo.

LalatenduMohanty:
retest this please

pkg/minishift/oc/oc.go
```diff
@@ -226,3 +225,23 @@ func mergeKubeConfigs(configs []*clientcmdapi.Config) *clientcmdapi.Config {
 	}
 	return mergedConfig
 }
+
+// getGlobalKubeConfigPath returns the path to the first entry in KUBECONFIG environment variable
+// or if KUBECONFIG not set then ~/home/.kube/config
+func getGlobalKubeConfigPath() (string, error) {
```
hferentschik: For what it's worth, why getGlobalKubeConfigPath? What's global about this? Is not a simple (get)KubeConfigPath more fitting?

praveenkumar:
retest this please.

pkg/minishift/oc/oc.go
```diff
@@ -216,3 +216,27 @@ func mergeKubeConfigs(configs []*clientcmdapi.Config) *clientcmdapi.Config {
...
+		usr, err := user.Current()
+		if err != nil {
+			return "", fmt.Errorf("unable to find current user: %v", err)
+		}
+		globalKubeConfigPath = filepath.Join(usr.HomeDir, ".kube", "config")
+	}
+	return globalKubeConfigPath, nil
+}
+
+func splitKubeConfig(r rune) bool {
```
hferentschik: +1, I think this is more accurate

pkg/minishift/oc/oc_test.go
```diff
@@ -128,6 +131,42 @@ func TestFlagExist(t *testing.T) {
...
 
+func Test_get_kubeconfig_global_path(t *testing.T) {
+	type testdata struct {
+		KubeConfig string
+		Expected   string
+	}
+	var tests []testdata
+
+	if runtime.GOOS == "windows" {
+		tests = []testdata{{KubeConfig: "/tmp/john/config", Expected: "/tmp/john/config"},
```
hferentschik: +1, better coverage

pkg/minishift/oc/oc.go
hferentschik: The godocs are still wrong. Either you use $HOME/.kube/config or ~/.kube/config. '~/home' is wrong.

praveenkumar:
> What's about any update to the user docs? Would it not make sense to add something there as well. It might be interesting for the user to which file the minishift config gets written.

@hferentschik Users who actually going to use this, already well aware of how the `KUBECONFIG` variable is used and configs are written.

> Maybe we should also log it somewhere?

log it you mean print it on console that `config written to <kube_config> path` ?

hferentschik:
> Users who actually going to use this, already well aware of how the KUBECONFIG variable is used and configs are written.

So you are saying that it is common knowledge that new cluster configuration is added to the first config found in the path? To me that was new. Personally, I think it is better to err on being a bit more verbose. Anyways,up to you.

> log it you mean print it on console that config written to <kube_config> path 

Maybe. I guess one would have to see how it looks like. It might be quite useful actually. What I initially had in mind is that at least we write some debug trace in an appropriate log level.

praveenkumar:
> Maybe. I guess one would have to see how it looks like. It might be quite useful actually. What I initially had in mind is that at least we write some debug trace in an appropriate log level

@hferentschik That I added as glog level 2 for debug purpose, so we are good now.

pkg/minishift/oc/oc.go
```diff
@@ -92,24 +94,25 @@ func (oc *OcRunner) AddSudoerRoleForUser(user string) error {
...
-		return fmt.Errorf("unable to find current user: %v", err)
+		return err
 	}
-
-	minishiftKubeConfigPath := oc.KubeConfigPath
-	kubeConfigGlobalPath := filepath.Join(user.HomeDir, ".kube", "config")
-
-	if err := os.MkdirAll(filepath.Join(user.HomeDir, ".kube"), 0755); err != nil {
-		return fmt.Errorf("unable to create kubeconfig dir ($HOME/.kube): %v", err)
+	if glog.V(2) {
```
hferentschik: :-)

pkg/minishift/oc/oc.go
```diff
@@ -216,3 +219,29 @@ func mergeKubeConfigs(configs []*clientcmdapi.Config) *clientcmdapi.Config {
 	}
 	return mergedConfig
 }
+
+// getGlobalKubeConfigPath returns the path to the first entry in KUBECONFIG environment variable
+// or if KUBECONFIG not set then $HOME/.kube/config
```
hferentschik: :-)

PR #2929, opened 2018-10-24
Issue #2924 Update docker-machine-kvm driver from 0.7.0 to 0.10.0

### LINKED ISSUE #2924: Update kvm driver plugin version in the doc and CI
Right now we have mentioned `0.7.0` kvm driver plugin in the document and also as part of our CI job.

praveenkumar:
Fixes #2924 

  1. 
  2. 
  3. 
  4.

### DIFF
--- a/docs/source/getting-started/setting-up-virtualization-environment.adoc
+++ b/docs/source/getting-started/setting-up-virtualization-environment.adoc
@@ -34,21 +34,14 @@ See the appropriate section for your hypervisor and operating system:
 [[setting-up-kvm-driver]]
 === Setting Up the KVM Driver
 
-{project} is currently tested against `docker-machine-driver-kvm` version 0.7.0.
-
-. As root, install the KVM binary and make it executable as follows:
-+
-----
-# curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.7.0/docker-machine-driver-kvm -o /usr/local/bin/docker-machine-driver-kvm
-# chmod +x /usr/local/bin/docker-machine-driver-kvm
-----
+{project} is currently tested against `docker-machine-driver-kvm` version 0.10.0.
 
 Proceed to the instructions for your Linux distribution after you have set up the `docker-machine-driver-kvm` binary.
 
 For more information, see the GitHub documentation of the link:https://github.com/dhiltgen/docker-machine-kvm#quick-start-instructions[Docker Machine KVM driver].
 
-[[kvm-driver-debian]]
-=== On Debian/Ubuntu
+[[kvm-driver-ubuntu]]
+=== On Ubuntu
 
 .  Install *libvirt* and *qemu-kvm* on your system:
 +
@@ -66,6 +59,21 @@ $ sudo usermod -a -G libvirtd <username>
 +
 ----
 $ newgrp libvirtd
+----
+
+. As root, install the KVM driver binary and make it executable as follows:
+- For Ubuntu-16.04 and higher
++
+----
+# curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.10.0/docker-machine-driver-kvm-ubuntu16.04 -o /usr/local/bin/docker-machine-driver-kvm
+# chmod +x /usr/local/bin/docker-machine-driver-kvm
+----
+
+- For Ubuntu-14.04
++
+----
+# curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.10.0/docker-machine-driver-kvm-ubuntu14.04 -o /usr/local/bin/docker-machine-driver-kvm
+# chmod +x /usr/local/bin/docker-machine-driver-kvm
 ----
 
 [[kvm-driver-fedora]]
@@ -87,6 +95,13 @@ $ sudo usermod -a -G libvirt <username>
 +
 ----
 $ newgrp libvirt
+----
+
+. As root, install the KVM driver binary and make it executable as follows:
++
+----
+# curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.10.0/docker-machine-driver-kvm-centos7 -o /usr/local/bin/docker-machine-driver-kvm
+# chmod +x /usr/local/bin/docker-machine-driver-kvm
 ----
 
 [[kvm-driver-arch-linux]]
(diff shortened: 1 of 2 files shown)


praveenkumar:
@anjannath don't put retest if it is a known failure.

anjannath:
@praveenkumar Opps, sorry i did not check the console before re-triggering this one..

docs/source/getting-started/setting-up-virtualization-environment.adoc
```diff
@@ -31,14 +31,7 @@ See the appropriate section for your hypervisor and operating system:
 [[setting-up-kvm-driver]]
 == Setting Up the KVM Driver
 
-{project} is currently tested against `docker-machine-driver-kvm` version 0.7.0.
-
```
agajdosi: :+1:

docs/source/getting-started/setting-up-virtualization-environment.adoc
```diff
@@ -84,6 +91,13 @@ $ sudo usermod -a -G libvirt <username>
 $ newgrp libvirt
 ----
 
+. As root, install the KVM driver binary and make it executable as follows:
```
agajdosi: :+1:

docs/source/getting-started/setting-up-virtualization-environment.adoc
```diff
@@ -63,6 +56,20 @@ $ sudo usermod -a -G libvirtd <username>
 $ newgrp libvirtd
 ----
 
+. As root, install the KVM driver binary and make it executable as follows (For Ubuntu-16.04 and higher):
```
LalatenduMohanty: We should not mention `For Ubuntu-16.04 and higher` as this is generic for all distribution. Also we have the specific section for Ubuntu

docs/source/getting-started/setting-up-virtualization-environment.adoc
```diff
@@ -63,6 +56,20 @@ $ sudo usermod -a -G libvirtd <username>
...
++
+----
+# curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.10.0/docker-machine-driver-kvm-ubuntu16.04 -o /usr/local/bin/docker-machine-driver-kvm
+# chmod +x /usr/local/bin/docker-machine-driver-kvm
+----
+
+. For Ubuntu-14.04:
```
LalatenduMohanty: It should be s/Ubuntu-14.04/Ubuntu/. Then add a note that only Ubuntu version 14.04 or higher is supported.

docs/source/getting-started/setting-up-virtualization-environment.adoc
LalatenduMohanty: s/KVM driver binary/ KVM docker machine plugin/

docs/source/getting-started/setting-up-virtualization-environment.adoc
LalatenduMohanty: s/KVM driver binary/ KVM docker machine plugin/

praveenkumar: > We should not mention For Ubuntu-16.04 and higher as this is generic for all distributions.

@LalatenduMohanty I don't understand that, what do you mean this is generic for all distributions?

LalatenduMohanty: @praveenkumar This is common for Debian as well as Ubuntu .Also you have a separate section for Ubuntu anyway. So we do not need to mention Ubuntu specific texts here. You can mention those in the Ubuntu section .

praveenkumar: @LalatenduMohanty I still don't get, here I mentioning ubuntu because driver have different version for different Ubuntu version, what I am missing here.

docs/source/getting-started/setting-up-virtualization-environment.adoc
gbraad: different instructions for different version. Here 14.04, and previously 16.04 or higher

docs/source/getting-started/setting-up-virtualization-environment.adoc
LalatenduMohanty: I got confused by the way instruction for Ubuntu has been written. I will resolve this comments and write new one

docs/source/getting-started/setting-up-virtualization-environment.adoc
LalatenduMohanty: @gbraad I am resolving this comment and will add new one.

docs/source/getting-started/setting-up-virtualization-environment.adoc
```diff
@@ -63,6 +56,20 @@ $ sudo usermod -a -G libvirtd <username>
...
++
+----
+# curl -L https://github.com/dhiltgen/docker-machine-kvm/releases/download/v0.10.0/docker-machine-driver-kvm-ubuntu14.04 -o /usr/local/bin/docker-machine-driver-kvm
+# chmod +x /usr/local/bin/docker-machine-driver-kvm
+----
```
LalatenduMohanty: @praveenkumar This section needs some update. 

1. First I think we can remove `Debian` from the section name i.e. only `On Ubuntu` is fine instead of `On Debian/Ubuntu`.  Because we do not have binary built for Debian and we do not know which binary  (there are two binary or Ubuntu now) will work on which version of Debian. Also, I have not seen any Debian user in Minishift. In case there are any, they can figure that out from Ubuntu versions.
2. The installation text for Ubuntu 16.04 and 14.04 needs to be consistent.  Check the sentences before the `curl` command. 
`As root, install the KVM driver binary and make it executable as follows (For Ubuntu-16.04 and higher):` 
vs 
`For Ubuntu-14.04:`

anjannath: I don't think this is necessary to change, it clearly explains what needs to be done. It first gives the general steps for installing and then goes on to distinguish between the two version based on the binary needed.

```
. As root, install the KVM driver binary and make it executable as follows: 

- For Ubuntu-16.04 and higher:
......................................
- For Ubuntu-14.04:
..................................
```

LalatenduMohanty: @anjannath Why would you want o mention Debian if there are no binary built specifically for Debian and we do no test for Debian? What about consistency in language?

Issue #2926, opened 2018-10-24
File existence check for iso-url


gbraad:
```
$ minishift start --iso-url file:///tmp/minishift-centos7.iso
-- Starting profile 'minishift'
-- Starting local proxy
-- Using proxy for the setup
-- Check if deprecated options are used ... OK
-- Checking if https://github.com is reachable ... OK
-- Checking if requested OpenShift version 'v3.10.0' is valid ... OK
-- Checking if requested OpenShift version 'v3.10.0' is supported ... OK
-- Checking if requested hypervisor 'kvm' is supported on this platform ... OK
-- Checking if KVM driver is installed ... 
   Driver is available at /usr/local/bin/docker-machine-driver-kvm ... 
   Checking driver binary is executable ... OK
-- Checking if Libvirt is installed ... OK
-- Checking if Libvirt default network is present ... OK
-- Checking if Libvirt default network is active ... OK
-- Checking the ISO URL ... OK
-- Checking if provided oc flags are supported ... OK
-- Starting the OpenShift cluster using 'kvm' hypervisor ...
-- Minishift VM will be configured with ...
   Memory:    2 GB
   vCPUs :    1
   Disk size: 5 GB
-- Starting Minishift VM ..... FAIL E1020 16:56:15.331415   94389 start.go:500] Error starting the VM: Error creating the VM. Error creating machine: Error in driver during machine creation: open /tmp/minishift-centos7.iso: no such file or directory. Retrying.
Error starting the VM: Error creating the VM. Error creating machine: Error in driver during machine creation: open /tmp/minishift-centos7.iso: no such file or directory
```

Summary:

  * `minishift start --iso-url file:///tmp/minishift-centos7.iso`
  * `-- Checking the ISO URL ... OK`
  * **no such file or directory** during *Starting Minishift VM*

anjannath:
So the issue is with `filehelper.Exists()`, i thought this was tested..

PR #2930, opened 2018-10-24
Issue #2926 return error when iso file does not exist

anjannath:
Fixes: #2926

### DIFF
--- a/pkg/minishift/config/validations.go
+++ b/pkg/minishift/config/validations.go
@@ -153,16 +153,16 @@ func IsValidISOUrl(_ string, isoURL string) error {
 		if !match {
 			return fmt.Errorf("'%s' url is not valid", isoURL)
 		}
-		if filehelper.Exists(strings.Replace(strings.TrimPrefix(isoURL, "file://"), "/", "\\", -1)) {
-			return nil
+		if !filehelper.Exists(strings.Replace(strings.TrimPrefix(isoURL, "file://"), "/", "\\", -1)) {
+			return fmt.Errorf("'%s' file is not present", isoURL)
 		}
 	} else {
 		match, _ := regexp.MatchString("^file:///.+", isoURL)
 		if !match {
 			return fmt.Errorf("'%s' url is not valid", isoURL)
 		}
-		if filehelper.Exists(strings.TrimPrefix(isoURL, "file://")) {
-			return nil
+		if !filehelper.Exists(strings.TrimPrefix(isoURL, "file://")) {
+			return fmt.Errorf("'%s' file is not present", isoURL)
 		}
 	}
 
--- a/test/integration/features/cmd-config.feature
+++ b/test/integration/features/cmd-config.feature
@@ -174,10 +174,17 @@ user defined options which changes default behaviour of Minishift.
     | property        | value                                                                                            | expected                                                                                        |
     | iso-url         | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso        | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso       |
     | iso-url         | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso  | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso |
-    | iso-url         | file:///home/me/Downloads/my_handmade_b2d.iso                                                    | file:///home/me/Downloads/my_handmade_b2d.iso                                                   |
     | iso-url         | b2d                                                                                              | b2d                                                                                             |
     | iso-url         | centos                                                                                           | centos                                                                                          |
-
+       
+  @quick
+  Scenario: Setting iso-url key to a non existing path
+     When executing "minishift config set iso-url file:///home/Downloads/my_handmade_centos.iso" fails
+     Then stderr should contain
+      """
+      'file:///home/Downloads/my_handmade_centos.iso' file is not present
+      """
+	
   @quick
   Scenario: Unsetting non-existing key
      When executing "minishift config unset i-do-not-exist" succeeds


pkg/minishift/config/validations.go
```diff
@@ -166,7 +166,7 @@ func IsValidISOUrl(_ string, isoURL string) error {
 		}
 	}
 
-	return nil
+	return fmt.Errorf("'%s' file is not present", isoURL)
```
praveenkumar: @anjannath return error at the end of the function doesn't something I expect this should be handled soon and returned.

why not having ?

```golang
 		match, _ := regexp.MatchString("^file://[a-zA-Z]:/.+", isoURL)
 		if !match {
 			return fmt.Errorf("'%s' url is not valid", isoURL)
 		}
 		if !filehelper.Exists(strings.Replace(strings.TrimPrefix(isoURL, "file://"), "/", "\\", -1)) {
 			return fmt.Errorf("'%s' file is not present", isoURL)
 		}
 	} else {
 		match, _ := regexp.MatchString("^file:///.+", isoURL)
 		if !match {
 			return fmt.Errorf("'%s' url is not valid", isoURL)
 		}
 		if !filehelper.Exists(strings.TrimPrefix(isoURL, "file://")) {
 			return fmt.Errorf("'%s' file is not present", isoURL)
		}
	}
return nil
```

anjannath:
Makes sense, will update, thanks

pkg/minishift/config/validations.go
```diff
@@ -153,16 +153,16 @@ func IsValidISOUrl(_ string, isoURL string) error {
 		if !match {
 			return fmt.Errorf("'%s' url is not valid", isoURL)
 		}
-		if filehelper.Exists(strings.Replace(strings.TrimPrefix(isoURL, "file://"), "/", "\\", -1)) {
-			return nil
+		if !filehelper.Exists(strings.Replace(strings.TrimPrefix(isoURL, "file://"), "/", "\\", -1)) {
+			return fmt.Errorf("'%s' file is not present", isoURL)
```
praveenkumar: 👍

LalatenduMohanty:
@anjannath You need to update the test to fix the CI failure https://ci.centos.org/job/minishift-pr/5270/consoleFull  , Preflight check and config validation use the same function  `IsValidISOUrl`

```
11:46:15     Examples: Correct values for iso-url keys
11:46:15       | property | value                                                                                           | expected                                                                                        |
11:46:15       | iso-url  | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso       | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso       |
11:46:15       | iso-url  | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso |
11:46:15       | iso-url  | file:///home/me/Downloads/my_handmade_b2d.iso                                                   | file:///home/me/Downloads/my_handmade_b2d.iso                                                   |
11:46:15         Command 'config set iso-url "file:///home/me/Downloads/my_handmade_b2d.iso"' did not execute successfully. cmdExit: 1, cmdErr: ['file:///home/me/Downloads/my_handmade_b2d.iso' file is not present]
```

anjannath:
@LalatenduMohanty We need to create the file to keep it in the test, but need the file uri, can we get that in the feature, how do i refer it?

LalatenduMohanty:
> @LalatenduMohanty We need to create the file to keep it in the test, but need the file uri, can we get that in the feature, how do i refer it?

I will suggest you create(empty file) the file in /tmp/filename.iso and then use the URI for the test.

anjannath:
The check is not going to boot a VM, it is only checking the behaviour of `minishfit config set iso-url` and not doing a start after that.

test/integration/testsuite/testsuite.go
```diff
@@ -264,6 +264,9 @@ func FeatureContext(s *godog.Suite) {
 		writeToFileInTestDirSucceeds)
 	s.Step(`^file "([^"]*)" should match text "([^"]*)" succeeds$`,
 		readFileForTextMatchSucceeds)
+	s.Step(`^creating file "([^"]*)" in os temporary directory succeeds$`,
+		createFileInTmp)
+
```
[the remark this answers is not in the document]
anjannath: @LalatenduMohanty This creates a file inside the test directory which is of the format `/mnt/minishift-test-xx` now we can not know the full path of the iso file inside this as the directory contains random chars at the end, this is why needed to create the step.

@agajdosi Can you confirm this?

test/integration/features/cmd-config.feature
```diff
@@ -174,7 +175,7 @@ user defined options which changes default behaviour of Minishift.
     | property        | value                                                                                            | expected                                                                                        |
     | iso-url         | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso        | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso       |
     | iso-url         | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso  | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso |
-    | iso-url         | file:///home/me/Downloads/my_handmade_b2d.iso                                                    | file:///home/me/Downloads/my_handmade_b2d.iso                                                   |
+    | iso-url         | file:///tmp/my_handmade_b2d.iso                                                                  | file:///tmp/my_handmade_b2d.iso                                                   |
```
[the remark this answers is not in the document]
anjannath: @praveenkumar The function in validations is not directly unit tested, but the function in preflight checks, `checkISOUrl` is unit tested and also has negative tests. 
It is just calling the `IsValidIsoUrl` from validations..

test/integration/testsuite/testsuite.go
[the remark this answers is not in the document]
amitkrout: Unnecessary atm

test/integration/features/cmd-config.feature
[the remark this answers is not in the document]
amitkrout: @anjannath Please change as per my ```request for change```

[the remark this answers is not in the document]
anjannath: @amitkrout Done! Thanks..

test/integration/features/cmd-config.feature
```diff
@@ -174,10 +174,17 @@ user defined options which changes default behaviour of Minishift.
     | property        | value                                                                                            | expected                                                                                        |
     | iso-url         | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso        | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso       |
```
amitkrout: Remove the row as it contains b2d reference

test/integration/features/cmd-config.feature
```diff
@@ -174,10 +174,17 @@ user defined options which changes default behaviour of Minishift.
...
     | iso-url         | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso  | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso |
-    | iso-url         | file:///home/me/Downloads/my_handmade_b2d.iso                                                    | file:///home/me/Downloads/my_handmade_b2d.iso                                                   |
     | iso-url         | b2d                                                                                              | b2d                                                                                             |
```
amitkrout: Remove  the row

test/integration/features/cmd-config.feature
anjannath: `b2d` still works with minishift, we should remove this when the corresponding code for b2d is also removed

test/integration/features/cmd-config.feature
anjannath: The `iso-url` with file scheme is already removed.

test/integration/features/cmd-config.feature
amitkrout: Its been a long time when we last supported b2d, now no b2d iso is tested with minilsift releases. So keeping it here creates an impression that we have full support for end user like we do for minishift-centos.iso. 

AFAIU b2d support code in master repo is dead and can be removed later probably in a new code clean up issue

test/integration/features/cmd-config.feature
amitkrout: Right, but i am talking about this
| iso-url         | b2d                                                                                              | b2d                                                                                             |

anjannath: Same as above.

test/integration/features/cmd-config.feature
anjannath: > AFAIU b2d support code in master repo is dead and can be removed later probably in a new code clean up issue

Should be done together then, as long as the code lives in master, we should make sure that it works.

anjannath: Anyway this is PR is only for #2926 code clean up for b2d should be done in a separate issue.

amitkrout: > as long as the code lives in master, we should make sure that it works.

@anjannath So you mean b2d is still supported ? AFAIK there is no support for b2d from minishift end

anjannath: It is not supported as in we won't add additional patches to `b2d` as of now. But this test is checking the config option, i think there is value in keeping this test. When we remove `b2d` after deprecating it for one release. Then we should remove this. but yeah let others weigh in..

test/integration/features/cmd-config.feature
```diff
@@ -174,10 +174,17 @@ user defined options which changes default behaviour of Minishift.
...
     | iso-url         | centos                                                                                           | centos                                                                                          |
-
+       
+  @quick
+  Scenario: Setting iso-url key to a non exsisting path
```
agajdosi: typo

test/integration/features/cmd-config.feature
```diff
@@ -174,10 +174,17 @@ user defined options which changes default behaviour of Minishift.
     | property        | value                                                                                            | expected                                                                                        |
     | iso-url         | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso        | https://github.com/minishift/minishift-b2d-iso/releases/download/v1.1.0/minishift-b2d.iso       |
     | iso-url         | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso  | http://github.com/minishift/minishift-centos-iso/releases/download/v1.1.0/minishift-centos7.iso |
-    | iso-url         | file:///home/me/Downloads/my_handmade_b2d.iso                                                    | file:///home/me/Downloads/my_handmade_b2d.iso                                                   |
```
agajdosi: this case of positive test of `file://` format is missing in the PR, however if I consider the time needed to do all the workarounds to get it working, I think we are fine with that. Better to focus on unit tests for these cases.

workarounds would be:
- create file in out/test-run (we have step for this)
- set absolute path of test-run directory into scenario variable (need implementation)
- use step `When executing "minishift config set iso-url "file://$(test-run)/my.iso"" succeeds`

test/integration/features/cmd-config.feature
```diff
@@ -174,10 +174,17 @@ user defined options which changes default behaviour of Minishift.
...
     | iso-url         | b2d                                                                                              | b2d                                                                                             |
     | iso-url         | centos                                                                                           | centos                                                                                          |
-
+       
+  @quick
+  Scenario: Setting iso-url key to a non exsisting path
+     When executing "minishift config set iso-url file:///home/Downloads/my_handmade_centos.iso" fails
+     Then stderr should contain
+      """
+      'file:///home/Downloads/my_handmade_centos.iso' file is not present
```
agajdosi: @anjannath I have tried that locally and the error is in brackets, it seems strange, why is it that way?

```
[agajdosi@dhcp-24-156 ~]$ minishift config set iso-url file:///home/agajdosi/my.iso
['file:///home/agajdosi/my.iso' file is not present]
```

anjannath: @agajdosi I think we are printing a string slice, so the brackets appear, will have to check the code.

agajdosi: @anjannath Yeah, it passes fine, I was just curious.

test/integration/features/cmd-config.feature
anjannath: @agajdosi Fixed it! Sorry for the delay..

test/integration/features/cmd-config.feature
anjannath: @agajdosi Hmm, it is because its a slice.. we append all the errors to an error slice while running validation checks, and return that slice to `config` command which it then prints using `Error()`

agajdosi: @anjannath Ok, thank you.

test/integration/features/cmd-config.feature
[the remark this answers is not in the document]
LalatenduMohanty: @praveenkumar @anjannath The reason the test is important because in the past we broke the URI format for a local ISO file and test like this can catch that issue.

LalatenduMohanty:
Tried https://github.com/minishift/minishift/pull/2980/ but the test is not behaving as expected.  So going to merge this as it is. Maybe worth creating a follow up issue.

PR #2935, opened 2018-10-25
Issue #2933 HypervVirtualSwitch flag default value should be set in start.go

### LINKED ISSUE #2933: Minishift on Windows fails to start with error "Not able to populate hyperv-virtual-switch flag"
### General information

  * Minishift version: 1.26.0
  * OS: Windows 10
  * Hypervisor: VirtualBox

### Steps to reproduce

  1. Run "minishift.exe start --vm-driver virtualbox" on Windows 10

### Expected

Minishift starts

### Actual

Minishift does not start with error message:
"Not able to populate hyperv-virtual-switch flag with Default Switch value" (although virtualbox is used as Hypervisor)

### Logs

```
./minishift start --show-libmachine-logs -v5
-- minishift version: v1.26.0+2fb32c8
-- Checking if requested OpenShift version 'v3.11.0' is valid ... OK
-- Checking if requested OpenShift version 'v3.11.0' is supported ... OK
-- Checking if VirtualBox is installed ... OK
Not able to populate hyperv-virtual-switch flag with Default Switch value
```

praveenkumar:
Fixes #2933 

1. Disable HyperV
2. Restart as prompted
3. minishift start --vm-driver virtualbox

### DIFF
--- a/cmd/minishift/cmd/config/config.go
+++ b/cmd/minishift/cmd/config/config.go
@@ -160,7 +160,7 @@ var (
 	DnsmasqContainerImage = createConfigSetting("network-dnsmasq-container", SetString, nil, nil, true, nil)
 
 	// Hyper-V vSwitch set to Default Switch by default
-	HypervVirtualSwitch = createConfigSetting("hyperv-virtual-switch", SetString, []setFn{validations.IsValidHypervVirtualSwitch}, nil, true, "Default Switch")
+	HypervVirtualSwitch = createConfigSetting("hyperv-virtual-switch", SetString, []setFn{validations.IsValidHypervVirtualSwitch}, nil, true, nil)
 
 	// Save start flags to viper config
 	SaveStartFlags = createConfigSetting("save-start-flags", SetBool, nil, nil, true, true)
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -708,7 +708,7 @@ func initStartFlags() *flag.FlagSet {
 		startFlagSet.String(configCmd.IPAddress.Name, "", "Specify IP address to assign to the instance (Hyper-V only)")
 		startFlagSet.String(configCmd.Netmask.Name, "", "Specify netmask to use for the IP address. Ignored if no IP address specified (Hyper-V only)")
 		startFlagSet.String(configCmd.Gateway.Name, "", "Specify gateway to use for the instance. Ignored if no IP address specified (Hyper-V only)")
-		startFlagSet.String(configCmd.HypervVirtualSwitch.Name, "", "Specify which Virtual Switch to use for the instance (Hyper-V only)")
+		startFlagSet.String(configCmd.HypervVirtualSwitch.Name, "Default Switch", "Specify which Virtual Switch to use for the instance (Hyper-V only)")
 	}
 	startFlagSet.AddFlag(nameServersFlag)
 


amitkrout:
@praveenkumar What i understand from the ```Steps to test the pull request``` is user has enabled Hyper-V and wants to run minishift on VirtualBox driver.

Is it a valid usecase ?

How we test:
1. Disable Hyper-V
2. Restart as prompted 
2. minishift start --vm-driver virtualbox

LalatenduMohanty:
> Fixes the problem on my Windows 7 machine.

@agajdosi You were seeing the same issue on Windows 7 machine?

LalatenduMohanty:
> @praveenkumar What i understand from the `Steps to test the pull request` is user has enabled Hyper-V and wants to run minishift on VirtualBox driver.

You are right, I have updated the steps to reproduce. I could reproduce this on Windows 10.

agajdosi:
@LalatenduMohanty @praveenkumar Yes, I have faced this on my Windows 7 VM.

amitkrout:
@LalatenduMohanty Faced the same issue on win10. Verified the pr bit as commented in https://github.com/minishift/minishift/issues/2933#issuecomment-433022396

cmd/minishift/cmd/config/config.go
```diff
@@ -160,7 +160,7 @@ var (
 	DnsmasqContainerImage = createConfigSetting("network-dnsmasq-container", SetString, nil, nil, true, nil)
 
 	// Hyper-V vSwitch set to Default Switch by default
-	HypervVirtualSwitch = createConfigSetting("hyperv-virtual-switch", SetString, []setFn{validations.IsValidHypervVirtualSwitch}, nil, true, "Default Switch")
+	HypervVirtualSwitch = createConfigSetting("hyperv-virtual-switch", SetString, []setFn{validations.IsValidHypervVirtualSwitch}, nil, true, nil)
```
gbraad: This would mean we can NOT define platform/hypervisor specific default values?
Perhaps this can be solved in a different way... as defining it here seems the more obvious choice.

cmd/minishift/cmd/start.go
```diff
@@ -708,7 +708,7 @@ func initStartFlags() *flag.FlagSet {
 		startFlagSet.String(configCmd.IPAddress.Name, "", "Specify IP address to assign to the instance (Hyper-V only)")
 		startFlagSet.String(configCmd.Netmask.Name, "", "Specify netmask to use for the IP address. Ignored if no IP address specified (Hyper-V only)")
 		startFlagSet.String(configCmd.Gateway.Name, "", "Specify gateway to use for the instance. Ignored if no IP address specified (Hyper-V only)")
-		startFlagSet.String(configCmd.HypervVirtualSwitch.Name, "", "Specify which Virtual Switch to use for the instance (Hyper-V only)")
+		startFlagSet.String(configCmd.HypervVirtualSwitch.Name, "Default Switch", "Specify which Virtual Switch to use for the instance (Hyper-V only)")
```
gbraad: Although... it might be that actually have two ways to deal with this... hmmm...

cmd/minishift/cmd/config/config.go
LalatenduMohanty: We can specify platform/hypervisor  specific default values. But we need to add more if condition to the `func initStartFlags() *flag.FlagSet {`

gbraad: There should be a single way to do this. At the moment it can be done as the config setting, and another way is from the startflags.

It would make more sense if defaults are inside the config section (and only here). Food for thought,

gbraad:
```diff
-..., default string)
+..., default string, appliedOnStart bool)
```

kb7791:
Is this still an ongoing issue?

PR #3019, opened 2018-11-20
[WIP] Issue #2676 enable validating and mutation admission webhook

### LINKED ISSUE #2676: Enable mutating and validating webhooks by default
Mutating and validating webhooks are central to the story of extending Kubernetes. We should ensure the experience of evaluating or running Kubernetes extensions that use these features is as painless as possible in minishift.

When getting [Knative](https://github.com/knative) to run with OpenShift, we found this to be the formula to use to get webhooks enabled:

https://github.com/knative/docs/blob/master/install/Knative-with-OpenShift.md#creating-a-new-openshift-cluster

I'd like to re-write the linked instructions to use minishift instead of oc cluster up; in order to do that we need validating and mutating webhooks to be enabled by default.  See also: https://github.com/openshift/origin/issues/20530

anjannath:
Fixes #2676

### DIFF
--- a/cmd/minishift/cmd/openshift/set.go
+++ b/cmd/minishift/cmd/openshift/set.go
@@ -78,6 +78,8 @@ func runPatch(cmd *cobra.Command, args []string) {
 	sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
 	dockerCommander := docker.NewVmDockerCommander(sshCommander)
 
+	fmt.Println(fmt.Sprintf("Patching OpenShift configuration '%s' with '%s'", patchTarget.LocalConfigFile, patch))
+
 	_, err = openshift.Patch(patchTarget, patch, dockerCommander)
 	if err != nil {
 		atexit.ExitWithMessage(1, fmt.Sprintf("Error patching the OpenShift configuration: %s", err.Error()))
--- a/Makefile
+++ b/Makefile
@@ -278,7 +278,8 @@ integration_all: $(MINISHIFT_BINARY)
 	--run-before-feature="$(RUN_BEFORE_FEATURE)" --test-with-specified-shell="$(TEST_WITH_SPECIFIED_SHELL)" --copy-oc-from="$(COPY_OC_FROM)" $(GODOG_OPTS)
 
 .PHONY: integration_pr ## Run integration tests for pull request (skip more specialized tests)
-integration_pr: GODOG_OPTS = --tags=core\&\&~disabled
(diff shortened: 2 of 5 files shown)


anjannath:
retest this please

anjannath:
CI failure is due to changes in the post cluster up need to look into this.

anjannath:
retest this please

cmd/minishift/cmd/openshift/set.go
```diff
@@ -78,6 +78,8 @@ func runPatch(cmd *cobra.Command, args []string) {
 	sshCommander := provision.GenericSSHCommander{Driver: host.Driver}
 	dockerCommander := docker.NewVmDockerCommander(sshCommander)
 
+	fmt.Println(fmt.Sprintf("Patching OpenShift configuration '%s' with '%s'", patchTarget.LocalConfigFile, patch))
```
LalatenduMohanty: This should be behind Verbose 2.

LalatenduMohanty:
@anjannath Is there anything pending from the code point of view?

cmd/minishift/cmd/openshift/set.go
LalatenduMohanty: Also I am not sure why this Println is present here. Was it for debugging?

anjannath:
@LalatenduMohanty The patch is applied, i need to add code to check if the container is up and then pass on control.

cmd/minishift/cmd/openshift/set.go
anjannath: Yes, this was for debugging.

anjannath:
retest this please

cmd/minishift/cmd/openshift/set.go
anjannath: Nope, this is not for debugging, this println was present in the `pkg/openshift/openshift.go` i moved it to the cmd package, its a refactor

anjannath:
@LalatenduMohanty I am  thinking of closing this in favour of #3044.

There's no way to correctly know if the cluster is up and able to serve, after patching the kube config file, as a result all the features in the integration test that do a `minishift start` and some commands after that needs to be changed to have some wait period/retry in them.

The only way to correctly know if the cluster is up is to run some `oc` commands against it. Adding the waits in post cluster up introduces delays in `minishift start`, which is bad IMO.

/cc @praveenkumar

LalatenduMohanty:
> @LalatenduMohanty I am thinking of closing this in favour of #3044.

Works for me. I am closing this cc @praveenkumar

Issue #3037, opened 2018-11-29
system tray auto start config setting not working


anjannath:
https://github.com/minishift/minishift/blob/5981f99603534104c3b557c6cd610976c88dfc84/cmd/minishift/cmd/start.go#L258-L262

The config value is not checked and tray is always started.

PR #3038, opened 2018-11-29
Issue #3037 Check config value before starting systemtray

anjannath:
Fixes: #3037 

Here's the artefacts for [darwin](https://5673-62728051-gh.circle-artifacts.com/0/ms/darwin-amd64/systemtray/minishift) and [windows](https://5673-62728051-gh.circle-artifacts.com/0/ms/windows-amd64/systemtray/minishift.exe)

/cc @amitkrout

### DIFF
--- a/cmd/minishift/cmd/start.go
+++ b/cmd/minishift/cmd/start.go
@@ -256,9 +256,11 @@ func runStart(cmd *cobra.Command, args []string) {
 	autoMountHostFolders(hostVm.Driver)
 
 	// start the minishift system tray
-	err = startTray()
-	if err != nil {
-		fmt.Println(err)
+	if viper.GetBool(configCmd.AutoStartTray.Name) {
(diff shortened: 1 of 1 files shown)


amitkrout:
Both the bits works fine. Tray icon is not visible with config setting ```minishift config set auto-start-tray false```. Works as expected on value ```true``` aswel.

cmd/minishift/cmd/start.go
```diff
@@ -256,9 +256,11 @@ func runStart(cmd *cobra.Command, args []string) {
 	autoMountHostFolders(hostVm.Driver)
 
 	// start the minishift system tray
-	err = startTray()
-	if err != nil {
-		fmt.Println(err)
+	if viper.GetBool(configCmd.AutoStartTray.Name) {
```
praveenkumar: LGTM :+1:

Issue #3056, opened 2018-12-08
Replace <username> with $(whoami) for easy copy/pasting


gastaldi:
### General information

  * Minishift version: Latest
  * OS: Linux
  * Hypervisor: KVM / xhyve / Hyper-V / VirtualBox

### Steps to reproduce

  1. Go to https://docs.okd.io/latest/minishift/getting-started/setting-up-virtualization-environment.html#kvm-driver-fedora
  2.  Copy/paste the command in there
  3. See it fail because the `<username>` token needs to be replaced

### Expected
Copy/paste and the command works

### Actual
It does not

LalatenduMohanty:
resolved via #3057

PR #3057, opened 2018-12-08
Issue #3056 replace <username> with $(whoami)

gastaldi:
This makes it easier to copy/paste the command

### DIFF
--- a/docs/source/getting-started/setting-up-virtualization-environment.adoc
+++ b/docs/source/getting-started/setting-up-virtualization-environment.adoc
@@ -59,12 +59,12 @@ $ sudo apt install libvirt-bin qemu-kvm
 - For Ubuntu-18.10
 +
 ----
-$ sudo usermod -a -G libvirt <username>
+$ sudo usermod -a -G libvirt $(whoami)
 ----
 - For older Ubuntu versions
 +
 ----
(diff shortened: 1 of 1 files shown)


anjannath:
ok to test

gastaldi:
@LalatenduMohanty I don't have any preference, as long as it works when copying and pasting ;)

robin-owen:
@gastaldi Personally, I think `whoami` is the better choice here. While `$USER` should be set in almost any shell, it's not actually guaranteed -- `whoami` should continue to work in those edge cases. :)

gbraad:
why not `$USER` ?

gbraad:
Although `$(whoami)` is understandable, it needs to execute a command. `$USER` is a variable which can, even when not exisitng, easily be defined from the shell.

```
USER=gbraad usermod -a -G libvirt $USER
```

would even work

anjannath:
> `USER=gbraad usermod -a -G libvirt $USER`

This still can't be copy pasted directly from the docs... cause you have to substitute the username.

gbraad:
@anjannath that is the example for override... `$USER` should be generally defined and available from your shell.

LalatenduMohanty:
Thanks @gastaldi

PR #3058, opened 2018-12-10
Issue #3055 Option to open OpenShift web console from tray

### LINKED ISSUE #3055: Add feature to open openshift console from system tray
A menu item to open the openshift console directly from tray.

anjannath:
Fixes #3055 

  1. `minishift start`
  2.  Click on the _OpenShift Web Console_ menu item.

### DIFF
--- a/pkg/minishift/systemtray/systemtray.go
+++ b/pkg/minishift/systemtray/systemtray.go
@@ -31,18 +31,23 @@ import (
 	"github.com/fsnotify/fsnotify"
 
 	"github.com/anjannath/systray"
+	"github.com/docker/machine/libmachine"
 	"github.com/golang/glog"
+	cmdUtil "github.com/minishift/minishift/cmd/minishift/cmd/util"
+	"github.com/minishift/minishift/cmd/minishift/state"
 	"github.com/minishift/minishift/pkg/minikube/constants"
 	"github.com/minishift/minishift/pkg/minishift/profile"
 	"github.com/minishift/minishift/pkg/minishift/shell/powershell"
 	"github.com/minishift/minishift/pkg/minishift/systemtray/icon"
 	"github.com/minishift/minishift/pkg/util/os"
+	"github.com/pkg/browser"
 )
 
 const (
-	START string = "Start"
-	STOP  string = "Stop"
-	EXIT  string = "Exit"
+	WEB_CONSOLE string = "Web Console"
+	START       string = "Start"
+	STOP        string = "Stop"
+	EXIT        string = "Exit"
 )
 
 const (
@@ -65,8 +70,9 @@ var (
 )
 
 type MenuAction struct {
-	start *systray.MenuItem
-	stop  *systray.MenuItem
+	start   *systray.MenuItem
+	stop    *systray.MenuItem
+	console *systray.MenuItem
 }
 
 func OnReady() {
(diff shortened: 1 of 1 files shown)


amitkrout:
```OpenShift Web Console``` works fine from minishift tray menu. Tested on ```Win10+VirtualBox```

Binary Used - https://5705-62728051-gh.circle-artifacts.com/0/ms/windows-amd64/systemtray/minishift.exe

**Comment by LalatenduMohanty** [comment]
retest this please

**Comment by LalatenduMohanty** [comment]
retest this please

**Comment by LalatenduMohanty** [comment]
@anjannath After I started minishift (default profile) I could see the green color for Minishift, but once I clicked the `OpenShift Web Console` , `Minishift` profile shows greyed out.

**Comment by anjannath** [comment]
> In Minishift we can have multiple profiles running at the same time, so having a single webconsole click link is a no go. We need to bring it in to the profile.

Agreed!

@LalatenduMohanty The profile should still be showing green, haven't touched the code that updates the status as part of this PR changes, need to take a look.

**Comment by anjannath** [comment]
@LalatenduMohanty on a second thought, even if there are multiple profiles only one profile is active at a time, and so in the o/p of `minishift console --url` you get the url of the console from the active profile, if you compare with that the flow is analogous.

But we don't have a way to set the active profile from the tray...

**Comment by LalatenduMohanty** [comment]
> @LalatenduMohanty on a second thought, even if there are multiple profiles only one profile is active at a time, and so in the o/p of `minishift console --url` you get the url of the console from the active profile, if you compare with that the flow is analogous.

Good point. If someone is only using  systemtray and want to access the webconsole , I do not think active profile is so much important for the user.  Lets see what other think about it @minishift/minishift-dev

**Comment by LalatenduMohanty** [comment]
When profile is not started and I click on the web console link it neither gives any error or message to the user. IMO it should tell that console is not available. 
screenshot 2019-02-05 at 1 44 14 pm

**Comment by anjannath** [comment]
> When profile is not started and I click on the web console link it neither gives any error or message to the user. IMO it should tell that console is not available.

@LalatenduMohanty We can't show any error msg or notification from the tray as of now!

The icon changing from green to greay i need to debug, haven't found anything in the code that was making it to change.

**Comment by LalatenduMohanty** [comment]
> @LalatenduMohanty We can't show any error msg or notification from the tray as of now!

Got it, please create an issue to track this as this should be considered as a known issue

**Comment by gbraad** [comment]
estatically this looks 'awful'. This is due to the fact that the `start` and `stop` commands are shorter than `Open web console`. I would prefer to be more descriptive and call it `Start cluster` and `Stop cluster`...

**Comment by gbraad** [comment]
Does it also have to be `OpenShift Web Console` ?

**Comment by anjannath** [comment]
> Does it also have to be `OpenShift Web Console` ?

What would you prefer? i didn't put much thought into it, so open for suggestions.

File: pkg/minishift/systemtray/systemtray.go [hunk 1]
```diff
@@ -31,18 +31,23 @@ import (
...
 	"github.com/minishift/minishift/pkg/minishift/systemtray/icon"
 	"github.com/minishift/minishift/pkg/util/os"
+	"github.com/pkg/browser"
 )
 
 const (
-	START string = "Start"
-	STOP  string = "Stop"
-	EXIT  string = "Exit"
+	WEB_CONSOLE string = "OpenShift Web Console"
```
LalatenduMohanty: I think just `Web Console` is better as OpenShift is kind of understood.

LalatenduMohanty:
@anjannath  The PR lgtm except the required cosmetic change of `Web Console`.

PR #3067, opened 2018-12-12
Issue #2840 Add macOS support to setup command

### LINKED ISSUE #2840: Minishift prerequisite installer for macOS
Enhance `minishift setup` command to work in macOS.

anjannath:
Fixes #2840 

  1. On a clean mac (without prerequisite of minishift set)
  2.  `minishift start` ==> this will fail
  3.  `minishift setup` ==> error msg saying to run command with sudo
  4.  `sudo minishift setup`

### DIFF
--- a/pkg/minishift/setup/hypervisor/virtualization_darwin.go
+++ b/pkg/minishift/setup/hypervisor/virtualization_darwin.go
@@ -17,15 +17,109 @@ limitations under the License.
 package hypervisor
 
 import (
+	"errors"
 	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"os/exec"
+	"os/user"
+)
+
+const (
+	driverBinaryDir   = "/usr/local/bin"
+	driverBinaryPath  = driverBinaryDir + "/docker-machine-driver-xhyve"
+	driverDownloadUrl = "https://github.com/machine-drivers/docker-machine-driver-xhyve/releases/download/v0.3.3/docker-machine-driver-xhyve"
 )
 
 func CheckHypervisorAvailable() error {
-	fmt.Println("Checking Hypervisor Availability ...")
 	return nil
 }
 
 func CheckAndConfigureHypervisor() error {
-	fmt.Println("Checking and Configuring Hypervisor  ...")
+	if isRoot() {
+		fmt.Println("Configuring Xhyve Hypervisor ...")
+		err := downloadXhyveDriver(driverBinaryPath, driverDownloadUrl)
+		return err
+	}
+	return errors.New("This command needs to be executed as administrator or with sudo.")
+}
+
+func isXhyveConfigured() bool {
+	//Following check is also present in cmd/start_preflight.go
+	path, err := exec.LookPath("docker-machine-driver-xhyve")
+
+	if err != nil {
+		return false
+	}
+
+	fi, _ := os.Stat(path)
+	// follow symlinks
+	if fi.Mode()&os.ModeSymlink != 0 {
+		path, err = os.Readlink(path)
+		if err != nil {
+			return false
+		}
+	}
+	fmt.Println("Driver is available at", path)
+
+	fmt.Print("Checking for setuid bit ... ")
+	if fi.Mode()&os.ModeSetuid == 0 {
+		print("FAIL")
+		return false
+	}
+	print("OK")
+	return true
+}
+
+func downloadXhyveDriver(filepath string, url string) error {
+	fmt.Println("Checking if docker-machine-driver-xhyve is already present and configured ... ")
+	if isXhyveConfigured() {
+		return nil
+	}
+	os.MkdirAll(driverBinaryDir, 0751)
+
+	out, err := os.Create(filepath)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
+	fmt.Printf("Downloading docker-machine-driver-xhyve binary to %s ... ", filepath)
+	resp, err := http.Get(url)
+	if err != nil {
+		return err
+	}
+	defer resp.Body.Close()
+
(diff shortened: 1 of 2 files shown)


pkg/minishift/setup/hypervisor/virtualization_darwin.go
```diff
@@ -17,15 +17,80 @@ limitations under the License.
...
+	if isRoot() {
+		fmt.Print("Configuring Xhyve Hypervisor.")
+		progressDots := progressdots.New()
+		progressDots.Start()
+		err := downloadXhyveDriver(driverBinaryPath, driverDownloadUrl)
+		progressDots.Stop()
+		fmt.Print("OK")
+		return err
+	}
+	return errors.New("You need to execute this command with sudo.")
```
LalatenduMohanty: `This command needs to be executed as administrator or with sudo`

pkg/minishift/setup/hypervisor/virtualization_darwin.go
```diff
@@ -17,15 +17,80 @@ limitations under the License.
...
+	err = out.Chown(0, 0)
+	if err != nil {
+		return err
+	}
+
+	err = out.Chmod(os.ModeSetuid | 0751)
+	if err != nil {
+		return err
+	}
+
```
LalatenduMohanty:
```
$ sudo minishift setup 
-- minishift version: v1.28.0+7bbed531
Configuring Xhyve Hypervisor.....OK

Consider checking getting started guide at https://docs.okd.io/latest/minishift/getting-started/index.html
```

The above output is not telling about what are the steps are being done in the background. We should  be transparent about changes we are doing to the system if user needs to know them. Because if user wants to cleanup the installation then at-least he would be aware of what changes were done to the system for Minishift setup.  Also that would make it consistent with `minishift setup` output in Windows.

LalatenduMohanty: Also when I run it when `the docker-machine-driver-xhyve` already present with correctly filepermission we should say so.e.g. `docker-machine-driver-xhyve` exists in the PATH with right permissions`.

pkg/minishift/setup/hypervisor/virtualization_darwin.go
```diff
@@ -17,15 +17,80 @@ limitations under the License.
...
+	err = out.Chown(0, 0)
+	if err != nil {
+		return err
+	}
+
+	err = out.Chmod(os.ModeSetuid | 0751)
+	if err != nil {
+		return err
+	}
```
LalatenduMohanty: We definitely need to add more logging in this function. We should add to the verbose logs `-v2` if something might not be ideal to be shown on the console by default. Also We need to make sure when failure happens, user should be able to pinpoint the failure, so that he can fix it or get better understanding about the failure.

pkg/minishift/setup/hypervisor/virtualization_darwin.go
anjannath: @LalatenduMohanty I deliberately did this because, we have the preflight check which already checks for this, my understanding is that, the user will only run the `minishift setup` command if `minishift start` fails for him in the first place, so checking again if the requirements are met felt like an redundant task. So `minishift setup` will setup the system again even if the requirements are already met.

pkg/minishift/setup/hypervisor/virtualization_darwin.go
anjannath: WIll add logging to the console, for the user to get an idea of what's being done by the command.

pkg/minishift/setup/hypervisor/virtualization_darwin.go
LalatenduMohanty: > my understanding is that, the user will only run the `minishift setup` command if `minishift start` fails for him in the first place

I do not agree. Once `minishift setup` is ready we will update our documentation to suggest running `minishift setup` first then `minishift start`. Same applicable when vscode openshift plugin suggest users to install minishift or even automates the install part.

pkg/minishift/setup/hypervisor/virtualization_darwin.go
```diff
@@ -17,15 +17,80 @@ limitations under the License.
...
+	"os"
+	"os/user"
+
+	"github.com/minishift/minishift/pkg/util/progressdots"
+)
+
+const (
+	driverBinaryDir   = "/usr/local/bin"
+	driverBinaryPath  = driverBinaryDir + "/docker-machine-driver-xhyve"
+	driverDownloadUrl = "https://github.com/machine-drivers/docker-machine-driver-xhyve/releases/download/v0.3.3/docker-machine-driver-xhyve"
```
praveenkumar: @anjannath The only concern I have and discussed (with @gbraad ) is this pre-compiled binary might not work across all the MacOS versions :( I don't know what would be the best solution. Might be we host this on somewhere according to mac version and then fetch it?

anjannath: @praveenkumar My thoughts as well, although i tried this binary in macOS mojave and it worked out of the box. But we should recompile this for specific versions of macOS and do the release in _machine-drivers_ org.

Then we can fetch the respective driver based on version of macOS.

anjannath:
@LalatenduMohanty I think i have addressed all your comments, PTAL when you have time.

pkg/minishift/setup/hypervisor/virtualization_darwin.go
LalatenduMohanty: @anjannath Lets create a follow up issue to handle this issue. I think this an important issue else we will end up getting issues for `minishift setup`.

Issue #3091, opened 2018-12-20
Fail when configuration is invalid


gbraad:
Currently when the ocnfiguration file has been hand edited for instance, the Read will not fail and just return an empty configuration.

praveenkumar:
fixed by #3092

Issue #2819, opened 2018-09-21
Add codecov.io into make tests and Travis


kpiwko:
### General information

Add codecov.io integration into .travis.yml so that code coverage results are computed on each PR in travis and team would have additional check on PR to see whether a particular PR is not missing unit tests.

Additionally add test coverage generation into `make test` target - just adding `go test -cover` might be enough.

  * Minishift version: N/A
  * OS: N/A
  * Hypervisor: N/A

Some links:
* https://gist.github.com/jonfriesen/1c64bebdd2b30ba1b2465c259f070459
* https://github.com/codecov/example-go
* https://blog.golang.org/cover

### Steps to reproduce

  1. `make test`
  
### Expected

Code coverage results are generated.

### Actual

Code coverage results are not generated.

praveenkumar:
@kpiwko thanks for the issue. This looks interesting, will have a look or you can also send a PR, most welcome :)

kpiwko:
Tried but I was not able to have a clean run of `make test` on macOS:

```
# github.com/minishift/minishift/pkg/util/os/atexit
pkg/util/os/atexit/atexit.go:77: Exitf call has arguments but no formatting directives
?   	github.com/minishift/minishift	[no test files]
?   	github.com/minishift/minishift/cmd/minishift	[no test files]
=== RUN   TestDisplayConsoleInMachineReadable
--- PASS: TestDisplayConsoleInMachineReadable (0.00s)
=== RUN   Test_clear_cache_user_confirms
--- PASS: Test_clear_cache_user_confirms (0.00s)
=== RUN   Test_clear_cache_user_aborts
--- PASS: Test_clear_cache_user_aborts (0.00s)
=== RUN   Test_clear_cache_forced
--- PASS: Test_clear_cache_forced (0.00s)
=== RUN   Test_delete_succeeds_for_non_existing_vm
--- PASS: Test_delete_succeeds_for_non_existing_vm (0.00s)
=== RUN   Test_unix_oc_path
--- PASS: Test_unix_oc_path (0.00s)
=== RUN   TestPreRunDirectories
--- PASS: TestPreRunDirectories (0.00s)
=== RUN   TestViperConfig
--- PASS: TestViperConfig (0.00s)
=== RUN   TestViperAndFlags
--- PASS: TestViperAndFlags (0.00s)
[...]

PR #3092, opened 2018-12-20
Issue #3091 Fail when json configuration is invalid

gbraad:
edit `allinstances.json` and put in an erroneous character.

### DIFF
--- a/pkg/minishift/config/allinstancesconfig.go
+++ b/pkg/minishift/config/allinstancesconfig.go
@@ -18,6 +18,7 @@ package config
 
 import (
 	"encoding/json"
+	"errors"
 	"github.com/minishift/minishift/pkg/minishift/hostfolder/config"
 	"io/ioutil"
 	"os"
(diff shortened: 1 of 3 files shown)


gbraad:
Codecov is so handy, that all PRs are now turned red.
And what benefit do we actually get? If no actions are taken, these 'integrations' are useless.

praveenkumar:
@gbraad allinstance config is something we create and not expect user to change anything by hand :) I think that's why till now we didn't have this kind of bug.

pkg/minishift/config/allinstancesconfig.go
```diff
@@ -83,6 +84,10 @@ func (cfg *GlobalConfigType) read() error {
 		return err
 	}
 
+	if !json.Valid(raw) {
+		return errors.New("Invalid JSON")
+	}
+
```
praveenkumar: :+1:

agajdosi:
@gbraad Issue with codecov was fixed, so it is not true that all PRs are red because of Codecov. But I understand you, I took a look and it seems we could define threshold for the commit status (https://docs.codecov.io/docs/commit-status), so if the coverage drops a little it won't fail the PR status.

agajdosi:
Created #3093.

agajdosi:
Oh, and on #2819 it looked like nobody was against it. If even -2% or 5% drops are ok, we can turn the reports completely off and keep just the comments about coverage. Or we can even remove it completely, lets discuss the actions we take on https://github.com/minishift/minishift/issues/3093.

PR #3253, opened 2019-05-07
WIP Update Che addon default parameter value

garagatyi:
Update Che addon default parameter value that sets Che plugin registry location.
It was changed recently and no longer compatible with the old value of the parameter.

1. Start minishift
2. Enable Che addon
3. Create Che 7 workspace
4. Start Che 7 workspace


gbraad:
Can only be merged AFTER the Che Plugin Registry has been updated to v3.

gbraad:
Also, please rebase as there is currently a conflict in the `che.addon` file.

PR #3260, opened 2019-05-13
Update Eclipse Che addon

### LINKED ISSUE #3259: Minishift loses all Che workspaces after restart
### General information

  * Minishift version: v1.33.0+ba29431
  * OS: Fedora 29 (Workstation Edition)
  * Hypervisor: VirtualBox 6.0.4

### Steps to reproduce

  1. Start minishift: `minishift start --vm-driver=virtualbox --cpus=4 --memory=16192 --disk-size=50GB --docker-opt userland-proxy=false`
  2. Deplay Che with `chectl`
  3. Create workspace, start it, clone a project
  4. Stop minishift with `minishift stop` command
  5. Start minishift again with `minishift start` command

### Expected
Che workspace is in place

### Actual
All Che workspaces with all the data are gone

sleshchenko:
Fixes #3259

  1. Install Che addon.
  2. Create and start a workspace and then try to use terminal there. It should work fine.
  3. Rollout Che deployment with a started workspace. The workspace should exist and be running after rollout finished.

### DIFF
--- a/addons/che/templates/che-server-template.yaml
+++ b/addons/che/templates/che-server-template.yaml
@@ -55,6 +55,18 @@ objects:
     port:
       targetPort: http
 - apiVersion: v1
+  kind: PersistentVolumeClaim
+  metadata:
+    labels:
+      app: che
+    name: che-data-volume
+  spec:
+   accessModes:
+   - ReadWriteOnce
+   resources:
+     requests:
+       storage: 1Gi
+- apiVersion: v1
   kind: DeploymentConfig
   metadata:
     name: che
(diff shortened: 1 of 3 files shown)


addons/che/templates/che-server-template.yaml
```diff
@@ -54,6 +54,18 @@ objects:
...
+  metadata:
+    labels:
+      app: che
+    name: che-data-volume
+  spec:
+   accessModes:
+   - ReadWriteOnce
+   resources:
+     requests:
+       storage: 1Gi
```
praveenkumar: 👍

praveenkumar:
@sleshchenko can you also check #3253 if that is also needed from che side?

sleshchenko:
@praveenkumar

> @sleshchenko can you also check #3253 if that is also needed from che side?

I've checked and it should be applied but later when Che Plugin Registry will be updated to v3.
Now it's not updated yet https://che-plugin-registry.openshift.io/v3/plugins/
Only https://che-plugin-registry.openshift.io/v2/plugins/ is available.

sleshchenko:
I tested it locally and everything worked as expected.

PR #3384, opened 2019-12-06
Issue #3375 Add deprecation message for xhyve driver

### LINKED ISSUE #3375: Removal of the xhyve driver implementation
We have decided to remove the xhyve driver support from Minishift. There are several motivations behind this; we have limited time to work on fixes, so we prefer to work on native and current hypervisors. In the case of xhyve, we have a replacement available to works without issues, namely hyperkit. The current implementation would need a fix to work with the latest version as can be seen in https://github.com/minishift/minishift/issues/3338. In the upcoming release we will add a deprecation notice with a link to instructions in the documentation on how to proceed (upgrade path to hyperkit), after which the subsequent release will remove support for xhyve.

anjannath:
Fixes #3375

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
     macos:
       xcode: "9.0"
     environment:
-        GOVERSION: 1.11.4
+        GOVERSION: 1.13.3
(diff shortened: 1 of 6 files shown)


anjannath:
retest this please

PR #3487, opened 2020-10-07
remove -F and -p flags on ssh command line

### LINKED ISSUE #3470: Is it possible to exclude -F and -p flags while calling external ssh?
### General information

  * Minishift version: 1.34.2
  * OS: Windows 10
  * Hypervisor: Hyper-V

I am trying to run minishift on a work laptop where Cisco AnyConnect Secure Mobility Client is installed and while logging in to work VPN. I can connect only to VM's ssh with port forwarding (with help of vmware vmnet8 adapter), so I have my localhost:30022 forwarded to myvmip:22, it all works fine except the -F flag makes external ssh ignore my ~/.ssh/config file, -p flag forces me to setup a forward from localhost:22 to myvmip:22.

Also would you be able to explain current use case for adding -F and -p flags?

### Steps to reproduce

  1. Make sure you an external ssh client.
  2. Run '`minishift start -v=7 --show-libmachine-logs`' or '`minishift ssh -v=7 --show-libmachine-logs -- echo hello`'
  3. You will see that external ssh command is being called with -F /dev/null and -p 22 flags.

### Expected
Configurable flag to remove flags while calling external ssh (and do not add these by default?).

### Actual
These ssh flags are always added while calling external ssh.

### Workaround
I currently have a shim executable for ssh that calls actual ssh client without these flags.

### Logs

```
...\ssh.exe -F /dev/null -o ConnectionAttempts=3 -o ConnectTimeout=10 -o ControlMaster=no -o ControlPath=none -o LogLevel=quiet -o PasswordAuthentication=no -o ServerAliveInterval=60 -o StrictHostKeyChecking=no -o UserKnownHostsFile=/dev/null docker@myvmip -o IdentitiesOnly=yes -i ~\.minishift\machines\minishift\id_rsa -p 22 echo hello
```

vmvmpvm:
fixes #3470

### DIFF
--- a/vendor/github.com/docker/machine/libmachine/ssh/client.go
+++ b/vendor/github.com/docker/machine/libmachine/ssh/client.go
@@ -67,7 +67,6 @@ const (
 
 var (
 	baseSSHArgs = []string{
-		"-F", "/dev/null",
 		"-o", "ConnectionAttempts=3", // retry 3 times if SSH connection fails
 		"-o", "ConnectTimeout=10", // timeout after 10 seconds
 		"-o", "ControlMaster=no", // disable ssh multiplexing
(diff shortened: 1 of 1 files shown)


vmvmpvm:
I do not see why the build failed, I wonder if it is not due to my changes.

vmvmpvm:
I do not think this is going anywhere, and I suspect it is due to an error in build (not in code), so I will resubmit this.

PR #3575, opened 2023-03-26
fix(addons): backward compatibility with OCP4 +

abdennour:
Fixing apiVersion of Template

Fixes <place holder for #issue number>

  1. 
  2. 
  3. 
  4.


praveenkumar:
minishift doesn't work with OCP-4.x so template update is not going to work.

